PR #13, opened 2015-11-11
Lp spaces, p-norms and weighting

kohr-h:
SO the Lp spaces and corresponding p-norms are ready now, together with additional vector (diagonal matrix weighting) in CUDA spaces. Some things need a review and some other things future optimizations, mostly marked with TODO.

- I enabled the possibility to pass a `CudaNtuples.Vector` into `CudaNtuples.element()`, which basically only checks data type and then re-wraps.
- p-norms with non-integer powers are not supported in CUDA since the `__pow__` implementation does not. Maybe it's too expensive and we never want that feature.
- Vector weighting is implemented via vector multiplication, which creates a new array and is slow. We probably want a native C++ implementation where the weights are a `CudaFn.Vector`, and the products and sums are calculated on the fly without creating an intermediate array.
- Some further choices may be debatable, e.g. the `dist_using_inner` default settings, which are a bit different from the NumPy spaces since there is a native `dist()` for CUDA vectors which is used by default.


odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+
+    :math:`||a||_{w, p} := ||w^{1/p} * a||_p`.
+
+    The vector may only have positive entries, otherwise it does not
+    define an inner product or norm, respectively. This is not checked
+    during initialization.
+    """
+
+    def __init__(self, vector, exponent=2.0, dist_using_inner=False,
+                 copy_to_gpu=False):
```
kohr-h: Having thought about this again, it's probably better to not allow this copy (and remove the option) but to require giving the initializer directly a `CudaFn.Vector`. Otherwise, the behavior of the `__eq__()` is weird: If a NumPy array is provided, and `copy_to_gpu` is `True`, two weightings initialized with the identical array will still not be equal.
The same is true a `CudaFn.Vector` is used in init with wrong data type, but that can be better justified (and documented).

adler-j:
I'll get to this sometime later today or this weekend.

examples/simple_operator.py
```diff
@@ -18,7 +18,7 @@
 """An example of a very simple operator on Rn."""
 
 # Imports for common Python 2/3 codebase
-from __future__ import print_function, division
+from __future__ import print_function, division, absolute_import
```
adler-j: Why this change?

examples/tomography.py
```diff
@@ -21,7 +21,7 @@
 """
 
 # Imports for common Python 2/3 codebase
-from __future__ import print_function, division
+from __future__ import print_function, division, absolute_import
```
adler-j: It seems this is a general change, I wont comment on more than this one, but an explanation please.

odl/discr/grid.py
```diff
@@ -26,7 +26,7 @@
 
 from future import standard_library
 standard_library.install_aliases()
-from builtins import range, super, zip
+from builtins import range, super, str, zip
```
adler-j: Where is this needed?

odl/discr/lp_discr.py
```diff
@@ -144,11 +166,15 @@ def interp(self):
         """Interpolation type of this discretization."""
         return self._interp
 
-    def points(self):
-        return self.grid.points(order=self.order)
+    def _inner(self, x1, x2):
+        """Raw inner product of two vectors."""
+        if self.exponent != 2.0:
```
adler-j: I dont think we should have these tests, easier to simply pass them to `super()` since the `Rn` space should complain later anyway.

odl/operator/operator.py
```diff
@@ -240,7 +240,7 @@ def __init__(self, domain, range, linear=False):
         self._is_linear = bool(linear)
         self._is_functional = isinstance(range, Field)
 
-        if self.is_linear:
+        if self._is_linear:
```
adler-j: Is this an optimization? I think we need a policy on this. I in general try to call the property since this makes the code more robust, but there is a performance penalty. You seem to like to call the private variable.

odl/set/__init__.py
```diff
@@ -15,119 +15,9 @@
...
------------------------------------
-
-+----------------------+-----------------------------------------------+
-|Name                  |Description                                    |
-+======================+===============================================+
-|`L2`                  |Square-integrable functions taking real or     |
-|                      |complex values                                 |
-+----------------------+-----------------------------------------------+
-"""
+# TODO: write an introduction
```
adler-j: Shouldnt merge with this todo still here.

odl/space/__init__.py
```diff
@@ -15,59 +15,9 @@
...
-=================================
-
-+----------------------+-----------------------------------------------+
-|Name                  |Description                                    |
-+======================+===============================================+
-|`L2`                  |Square-integrable functions taking real or     |
-|                      |complex values                                 |
-+----------------------+-----------------------------------------------+
-"""
+# TODO: write an introduction
```
adler-j: This should be done before merge else we'll simply forget to do it.

odl/space/cu_ntuples.py
```diff
@@ -647,9 +698,9 @@ def _multiply(self, x1, x2, out):
         >>> x2 = rn.element([1, 2, 3])
         >>> out = rn.element()
         >>> rn.multiply(x1, x2, out)  # out is returned
-        CudaRn(3).element([5.0, 6.0, 6.0])
+        CudaRn(3, 'float32').element([5.0, 6.0, 6.0])
```
adler-j: Why do we write `'float32'`?

odl/space/cu_ntuples.py
```diff
@@ -807,6 +820,122 @@ def fun(x, out=None):
...
+        elif weight_.ndim == 2:
+            raise NotImplementedError('matrix weighting not implemented '
+                                      'for CUDA spaces.')
+#            weighting = CudaFnMatrixWeighting(
+#                weight_, exponent, dist_using_inner=dist_using_inner)
+        else:
+            raise ValueError('array-like weight must have 1 or 2 dimensions, '
+                             'but {} has {} dimensions.'
+                             ''.format(weight, weight_.ndim))
+    return getattr(weighting, attr)
```
adler-j: i dont realy get this line, what happens?

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
     return x.data.norm()
 
 
+def _pnorm_default(x, p):
+    if p == float('inf'):
+        raise NotImplementedError('inf-norm not implemented.')
+    elif int(p) != p:
+        raise NotImplementedError('non-integer powers not implemented')
+    # TODO: optimized version in C++ code?
```
adler-j: this is an absolute must, IMO, before we even commit this.

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+    xp = abs(x)
+    xp **= p
+    return sum(xp)**(1/p)
```
adler-j: `sum()` uses the iterator interface. At least use `np.sum`.

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+
+
+def _pnorm_diagweight(x, p, w):
```
adler-j: see comments on `_pnorm_default`.

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
     return x.data.norm()
 
 
+def _pnorm_default(x, p):
+    if p == float('inf'):
+        raise NotImplementedError('inf-norm not implemented.')
+    elif int(p) != p:
+        raise NotImplementedError('non-integer powers not implemented')
+    # TODO: optimized version in C++ code?
+    xp = abs(x)
```
adler-j: Is `abs` even defined for cuda vectors?

odl/space/cu_ntuples.py
adler-j: I agree.

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+
+    def vector_is_valid(self):
+        """Test if the vector is a valid weight, i.e. positive.
+
+        Note
+        ----
+        This operation copies the vector to the CPU memory and uses
+        `numpy.all`, which can be very time-consuming in total.
+        """
+        return np.all(self.vector > 0)
```
adler-j: `self.vector > 0` should not be valid, does this code run?

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+
+        See also
+        --------
+        equiv : test for equivalent inner products
+        """
+        if other is self:
+            return True
+
+        return (isinstance(other, CudaFnVectorWeighting) and
+                self.vector is other.vector and
```
adler-j: id be fine with a `self.vector == other.vector` since this is quite fast and doesnt require any copies.

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+        -------
+        inner : float or complex
+            The inner product of the two provided vectors
+        """
+        if self.exponent != 2.0:
+            raise NotImplementedError('No inner product defined for '
+                                      'exponent != 2 (got {}).'
+                                      ''.format(self.exponent))
+        else:
+            inner = _inner_default(x1 * self.vector, x2)
```
adler-j: Where do we support weighted `l2` type norms?

adler-j: The case with weighted l2 norms is the most common and should be optimized.

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+
+        Returns
+        -------
+        norm : float
+            The norm of the provided vector
+        """
+        if self.exponent == 2.0:
+            return super().norm(x)
+        elif self.exponent == float('inf'):
+            return _pnorm_default(x, float('inf'))
```
adler-j: better to pass `self.exponent` here, no need to make new object.

odl/space/cu_ntuples.py
```diff
@@ -943,8 +1318,14 @@ def norm(self, x):
         norm : float
             The norm of the vector
         """
-        from math import sqrt
-        return sqrt(self.const) * float(_norm_default(x))
+        if self.exponent == 2.0:
+            from math import sqrt
+            return sqrt(self.const) * float(_norm_default(x))
+        elif self.exponent == float('inf'):  # Weighting irrelevant
```
adler-j: Shouldnt this be linear in the `constant`?

odl/space/fspace.py
```diff
@@ -503,6 +502,11 @@ def product(arg):
             return x_old(arg) * y_old(arg)
         y._function = product
 
+    def _inner(self, x1, x2):
```
adler-j: tbh simply let this pass through, it will fail in `LinearSpace._inner` anyway. This way causes more code at marginal gain.

odl/space/ntuples.py
```diff
@@ -455,9 +456,10 @@ def _blas_is_applicable(*args):
     if len(args) == 0:
         return False
 
-    return (all(x.dtype == args[0].dtype for x in args) and
-            all(x.dtype in _BLAS_DTYPES for x in args) and
-            all(x.data.flags.contiguous for x in args))
+    return (all(x.dtype == args[0].dtype and
```
adler-j: Nice improvement!

odl/space/ntuples.py
```diff
@@ -499,19 +501,19 @@ def fallback_copy(x1, x2, n):
         _lincomb(a+b, x1, 0, x1, out, dtype)
     elif out is x1 and out is x2:
         # All the vectors are aligned -> out = (a+b)*out
-        scal(a+b, out.data, out.size)
+        scal(a+b, out.data, native(out.size))
```
adler-j: Having casts everywhere he is kindof against the idea of this beeing optimized. Not 100% convinced.

odl/space/ntuples.py
```diff
@@ -1320,7 +1349,7 @@ def adjoint(self):
                                       ''.format(self.domain.field,
                                                 self.range.field))
         return MatVecOperator(self.range, self.domain,
-                              np.asarray(np.asmatrix(self.matrix).H))
+                              self.matrix.conj().T)
```
adler-j: I think we should change `self.matrix` to `self.array` simply to stop the implication that this is a matrix.

odl/space/ntuples.py
```diff
@@ -1426,9 +1461,11 @@ def weighted_dist(weight, use_inner=False):
         evaluate to exactly zero for equal (but not identical)
         `x` and `y`.
 
+        Can only be used if `exponent` is 2.0.
+
     Returns
     -------
-    inner : callable
+    dist : callable
```
adler-j: I've been doing this on the doc branch, but anytime you write callable, you can do `callable`, that way you get a link to the python doc, in case someone doesnt underdand.

odl/space/ntuples.py
```diff
@@ -1767,55 +1939,103 @@ def inner(self, x1, x2):
...
+        x : `Fn.Vector`
+            Vector whose norm is calculated
+
+        Returns
+        -------
+        norm : float
+            The norm of the provided vector
+        """
+        if self.exponent == 2.0:
+            return super().norm(x)
```
adler-j: This does not use the weighting, is there a test that runs this?

test/discr/l2_discr_test.py
```diff
@@ -29,146 +28,145 @@
 
 import odl
 
-from odl.util.testutils import almost_equal, all_almost_equal, skip_if_no_cuda
+from odl.util.testutils import all_equal, skip_if_no_cuda
+
+
+# TODO: Lp for p != 2
```
adler-j: We can't accept such a big commit without some tests, at least basic ones.

test/operator/operator_test.py
```diff
@@ -643,7 +642,7 @@ def test_nonlinear_functional_operators():
 
     assert not C.is_linear
     assert almost_equal(C(x), A(x * 2.0))
-
```
adler-j: In general such changes shouldnt be in Pull requests, I know we both do it, but we should try to avoid it. This causes some extra trubble in the review.

test/space/cu_ntuples_test.py
```diff
@@ -18,70 +18,151 @@
...
 # TODO:
-# * weighted spaces
 # * custom dist/norm/inner
 
 
 def _array(fn):
     # Generate numpy vectors, real or complex or int
     if np.issubdtype(fn.dtype, np.floating):
-        return np.random.randn(fn.size).astype(fn.dtype)
+        return np.random.uniform(-1, 1, fn.size).astype(fn.dtype)
```
adler-j: Why this change?

test/space/cu_ntuples_test.py
```diff
@@ -18,70 +18,151 @@
...
+
+def test_init_cudantuples_dtypes():
+    # verify that the code runs
+    for dtype in odl.CUDA_DTYPES:
+        odl.CudaNtuples(3, dtype=dtype).element()
+        odl.CudaFn(3, dtype=dtype).element()
+
+
+def test_init_exponent(exponent):
+    for dtype in odl.CUDA_DTYPES:
```
adler-j: fixture for cuda dtypes?

test/space/fspace_test.py
```diff
@@ -29,38 +29,8 @@
 
 # ODL imports
 import odl
-from odl import L2
 from odl.util.testutils import almost_equal
 
 
-def test_interval():
```
adler-j: Why remove these test? Simply rename `L2` -> `FunctionSpace`

test/space/ntuples_test.py
```diff
@@ -112,7 +139,7 @@ def test_init():
 
     # Rn
     Rn(3, float)
-    Rn(3, int)
+    Rn(3, int)  # TODO: do we want this??
```
adler-j: I'd say no, we dont.

test/space/ntuples_test.py
```diff
@@ -300,15 +362,6 @@ def idiv_aliased(x):
     _test_unary_operator(fn, lambda x: x / x)
 
 
-def test_norm(fn):
```
adler-j: Why remove this? Do we have any tests for the weighted norms?

adler-j:
Added some comments. Still havnt built the code and ran it, nor have I done a real design overview. Abit to much so I'll do that tomorrow or later.

examples/simple_operator.py
kohr-h: Admittedly not the best idea to bundle with an unrelated pull request. Generally, I once in a while run the modules through the `futurize` and `pasteurize` tools from the `future` package and add (some of) the recommended changes. An obvious exception are the `unicode_literals` which can cause trouble but would still be good to have since it would make the code behave the same for both Py2 and Py3.

examples/tomography.py
kohr-h: The import prevents implicit relative imports, see [here](http://python-future.org/compatible_idioms.html#imports-relative-to-a-package). It can (and should IMO) safely be added in any module.

odl/discr/grid.py
kohr-h: [here](https://github.com/odlgroup/odl/blob/issue-246__l2discr/odl/discr/grid.py#L129). It's a safe import and makes the code slightly more predictable.

odl/discr/lp_discr.py
kohr-h: That's correct. I'll change it.

odl/operator/operator.py
kohr-h: Before I saw it in action, I didn't even know that it was possible to use the property at that point. But why shouldn't it? The instance is already created at that point, so the property exists and works once the "private" attribute is in place.
So no, it isn't an optimization, more something like "I just set the private attribute so I know I can use that, not so sure about the public property." I'll change it back and (incrementally) also in other places where I encounter similar code.

odl/set/__init__.py
kohr-h: Sure. Btw, now that we have a proper API doc, the tables are not really that useful any longer. I'd rather write an introductory text about the contents and the usefulness of the submodule.

odl/space/__init__.py
kohr-h: See above. Just wanted to check how we should proceed here. Some words rather than tables, right?

odl/space/cu_ntuples.py
```diff
@@ -434,17 +458,35 @@ def __init__(self, size, dtype, **kwargs):
...
+
+                float : Weighting by a constant
 
-                float : Use functions weighted by a constant
+                array-like : Weighting by a vector (1-dim. array,
+                corresponds to a diagonal matrix). Note that the array
+                is stored in main memory, which results in slower
+                space functions due to a copy during evaluation.
+
+                `CudaFn.Vector` : same as 1-dim. array-like, except
```
kohr-h: I'll fix the quotation marks BTW.

examples/simple_operator.py
adler-j: Okay, agree that this should be done on master.

examples/tomography.py
adler-j: Okay, makes sense.

odl/discr/grid.py
adler-j: :+1:

odl/discr/lp_discr.py
adler-j: :+1:

odl/space/cu_ntuples.py
kohr-h: I don't exactly know. It shouldn't be printed out, right? I noted though that this doctest failed and fixed it. Need to look at it.

odl/set/__init__.py
adler-j: I agree, these tables simply give a lot of duplication.

odl/space/__init__.py
adler-j: Agree!

odl/space/cu_ntuples.py
kohr-h: Well, the point of the free functions below was to get the weighted inner product, norm and dist without having to to through the creation of the weighting class explicitly. Pure convenience. Code-wise this is just DRY since for each of the three free functions, the procedure is the same - create the class and fish out the corresponding attribute.

odl/space/cu_ntuples.py
kohr-h: Sorry for committing :-)
I should have used an exclamation mark instead. However, it will take me a while to find out how to do this. I'll make an issue in odlpp.

odl/space/cu_ntuples.py
kohr-h: [yes](https://github.com/odlgroup/odl/blob/issue-246__l2discr/odl/space/cu_ntuples.py#L818)

odl/space/cu_ntuples.py
kohr-h: Big OOPS. For some reason I thought `sum` was a unary CUDA function, too. Then we need to add it (anyway useful).

odl/space/cu_ntuples.py
```diff
@@ -847,36 +976,271 @@ def _norm_default(x):
...
+    xp **= p
+    return sum(xp)**(1/p)
+
+
+def _pnorm_diagweight(x, p, w):
+    if p == float('inf'):
+        raise NotImplementedError('inf-norm not implemented.')
+    elif int(p) != p:
+        raise NotImplementedError('non-integer powers not implemented')
+    # TODO: optimized version in C++ code?
```
kohr-h: So yes also here?

odl/space/cu_ntuples.py
kohr-h: Oh dear, copy-and-paste. And obviously there's no test for it. Any simple way to solve this or do we have to dive into the CUDA code?

odl/space/cu_ntuples.py
kohr-h: Err, no. The vector has the same size as the data (potentially GB), and even "quite fast" is not sufficient since this check is done in *every* space comparison, which in turn is done in each `vec in space` check. This check must be very fast, not just quite fast.

odl/space/cu_ntuples.py
kohr-h: Correct. Imagine the same `# TODO` line as in the norm functions.

odl/space/cu_ntuples.py
kohr-h: Sure.

odl/space/cu_ntuples.py
adler-j: Change this line to `return weighting`, and change `def cu_weighted_norm` to `return _weighted(weight, exponent=exponent).norm`?

odl/space/cu_ntuples.py
adler-j: Its quite simple, I could do it for you. Simply give me the interface and I'll do it. (unless you want to learn how `odlpp` works ofc).

I clearly should have meant pulled, not commited :D

odl/space/cu_ntuples.py
adler-j: Thats called by `odl.space.cuda.abs`, is that the function called here?

odl/space/cu_ntuples.py
adler-j: `np.greater_than(self.vector, 0)` should do.

odl/space/cu_ntuples.py
adler-j: Oh it's for every space comparsion? I agree with you then ofc.

odl/space/cu_ntuples.py
kohr-h: No, that's not possible. It's not even the case if `p==2`.

The mathematical behavior we want is that `p==inf` is the limiting case for `p->inf`. In the case of constant weighting, it is easy to see what happens since
```
||x||_{c, p} = sum(c * abs(x)**p)**(1/p) = c**(1/p) ||x||_p
```
so for `p->inf`, the factor `c**(1/p)` goes to 1.

Even in the matrix-weighted case, one can write
```
||x||_{G, p} = ||G**(1/p) x||_p
```
so, by continuity in `p` and since `G**(1/p) -> I` for `p->inf`, the weighted norm goes to the unweighted one in the limiting case. (I've also tested it numerically to be sure :-) )

odl/space/fspace.py
kohr-h: :+1: To also be honest, I simply copied it and prettified the message a little, at little gain.

odl/space/cu_ntuples.py
adler-j: Hmm ok, but you could still define a norm as `max(x * w)`, which is probably what people would expect, right?

odl/space/ntuples.py
kohr-h: I don't see a performance hit by casting a single integer. And so far, the BLAS functions seem to be the only ones so far which are "stupid" enough to screw up if the input size is not a native integer. I wonder what happens if one passes a `numpy.int32`.

Other than that, it's once again about making code more predictable across Py2/3. With integers, it's mostly about printing (with or w/o `L` for long ones), but also about instance checks (which we don't have so far). See [here](http://python-future.org/compatible_idioms.html#long-integers).

odl/space/ntuples.py
kohr-h: But it represents a matrix. Or do you mean the implication that this might be a `numpy.matrix`?

A bit of `matrix` would still be nice in the name, what about `mat_array`?

odl/space/ntuples.py
kohr-h: I'll get it fixed.

odl/space/ntuples.py
adler-j: Yeah i mean that it would be a `numpy.matrix`, maybe some name of what it does instead of what it is?

odl/space/cu_ntuples.py
adler-j: Agree

odl/space/cu_ntuples.py
adler-j: Indeed it shouldnt be printed, try checking the `__repr__`

odl/space/ntuples.py
kohr-h: Yes it does, and ~~yes there is a test for it~~ crap, I forgot the test. `super` magic :-)

What happens is that the [norm implementation of `_FnWeightingBase`](https://github.com/odlgroup/odl/blob/issue-246__l2discr/odl/space/base_ntuples.py#L471-L487) is called, which refers to `inner`, which in turn uses weighting. I'll write a test and see if this is really what happens :-)

test/discr/l2_discr_test.py
kohr-h: :cry: So much work

test/operator/operator_test.py
kohr-h: Right. However, in this case, most things happen by me saving the file, which removes all the trailing whitespaces *which should not be there in the first place*. I'll be more angry about those and other avoidable PEP8 violations in the future then :angry:

test/space/cu_ntuples_test.py
kohr-h: Yeah, why not?

test/space/fspace_test.py
kohr-h: Okay, I was a a bit too quick here.

test/space/ntuples_test.py
kohr-h: Alright, so I'll adapt the check in `Rn.__init__()`

test/space/ntuples_test.py
kohr-h: That one only got reshuffled. Regarding weighting, I'd say tests for the `FnWeighting` classes should cover this, no need to double up weighting tests here (beyond init and equality tests, of course). The tests for the weighting classes are desperately needed, though.

test/space/cu_ntuples_test.py
kohr-h: I changed that in my first attempt to make a [positive array](https://github.com/odlgroup/odl/pull/13/files#diff-7005bf93daa4498c5a50638aa9347a75R72) but then decided on a different implementation. I'll change it back.

odl/space/cu_ntuples.py
kohr-h: Sure, that's possible. Slightly less fancy ;-)

odl/space/cu_ntuples.py
kohr-h: I'd like to know how it works. I actually had a quick look into the C++ code, but there was no obvious place to steal code from, so I decided not to invest more time searching around. Are there any `thrust` transform functions one can use or do we need a custom functor as last time (don't remember which function it was for)?

odl/space/cu_ntuples.py
kohr-h: Yes, the builtin `abs` is shadowed by the definition in the same file.

odl/space/cu_ntuples.py
kohr-h: [Not big OOPS after all](https://github.com/odlgroup/odl/blob/issue-246__l2discr/odl/space/cu_ntuples.py#L967-L968). `sum` is also shadowed, that's why I have to [re-import the built-in one](https://github.com/odlgroup/odl/blob/issue-246__l2discr/odl/space/cu_ntuples.py#L543-L544) at one place.

odl/space/cu_ntuples.py
kohr-h: Sure. At least something similar :wink:

odl/space/cu_ntuples.py
kohr-h: At least in `Fn`. I'll add it in `CudaFn`, too, which I should have done already.

odl/space/cu_ntuples.py
kohr-h: That one could do in principle it would define a valid norm, but the question is what we prioritize - mathematical consistency in the sense I sketched above, with the potentially surprising implication that the inf-norm does not care about weighting - or to save users from surprises. Not really sure.

odl/operator/operator.py
adler-j: :+1:

odl/space/cu_ntuples.py
adler-j: Yeah but easier to understand, I realy had to stop and think abit here, which shouldnt be needed. It wouldnt change any interface aswell :)

odl/space/cu_ntuples.py
adler-j: You can use a `thrust::transform_reduce`, but since you will have three input vectors (at least in the `inner` case), you will need to wrap the data using `thrust::zip_iterator` and use a custom functor I think.

We could do this come monday, ok?

odl/space/cu_ntuples.py
adler-j: Thus kindoff indicates that we need to treat `sum` in a somewhat smarter way, doesnt affect this pull however.

odl/space/cu_ntuples.py
kohr-h: If that's hard to understand then look at [this class](https://github.com/odlgroup/odl-solvers/blob/issue-5__steplen/odlsolv/util/state.py) :grin:. I felt like doing some crazy shit on Friday...

odl/space/cu_ntuples.py
adler-j: :+1:

odl/space/cu_ntuples.py
adler-j: See above comment, we'll get this fixed.

odl/space/cu_ntuples.py
adler-j: I'm not sure either. How about we make it throw an error in the `inf` case for now so people dont get used to a behaviour, and we make an issue on this to discuss it?

odl/space/cu_ntuples.py
kohr-h: Sounds good!

odl/space/cu_ntuples.py
kohr-h: Maybe, maybe not. It's confined to the module, so it doesn't really hurt much. And when you get an error, it will point to the `sum` function in the module, so it's also easy to debug. Anyway, not in this pull request.

odl/space/ntuples.py
adler-j: Not quite sure I like the use of inheritance here then, this is somewhat confusing (what if someone reimplements `super.norm` to be optimized?). How about simply writing it out to be more explicit?

test/discr/l2_discr_test.py
adler-j: Kindoff now or never though. If we dont we'll never get it done.

test/space/cu_ntuples_test.py
adler-j: If you havn't done it, make an issue and we can fix it somewhere else.

odl/space/cu_ntuples.py
adler-j: yes!

odl/space/cu_ntuples.py
kohr-h: That sounds like a good idea. I'd @mentioned Ozan, but he first needs a bloody account here.

odl/space/ntuples.py
kohr-h: Not sure I understand. Where should the optimized `super().norm()` come from?

Apart from that, it is this kind of slightly spooky inheritance for which `super()` is made. I'll come back with a better explanation.

test/space/cu_ntuples_test.py
kohr-h: Well, it's not that much work, but maybe we want to take care of that in the context of unified `FnBase` tests? Seems a bit remote though.

odl/space/ntuples.py
adler-j: What I mean is that currently `_FnWeightingBase` implements `norm` using `inner`. To me this isnt obviously the way to go, for example, one might expect the base one to be optimized properly. In the `norm` case this doesnt realy matter, but currently `dist` requires a copy. What if we decided to optimize it in `_FnWeightingBase`? Then this code would break (in the `dist` case).

I would in this case assume that `super().norm()` gave me the "standard R^n norm", not some magic depending on how `inner` is implemented in this function.

test/space/cu_ntuples_test.py
adler-j: Couldnt be made in the context of a unified `FnBase` test since this tests the initialization of these `Fn`'s, but sure, that branch may be a good place to clean all of this up.

test/space/cu_ntuples_test.py
```diff
@@ -18,70 +18,151 @@
...
+        odl.CudaFn(3, dtype='float32', exponent=exponent)
```
adler-j: Shouldnt it be `dtype=dtype`?

odl/space/ntuples.py
kohr-h: Ah you mean that we optimize the base class `dist` and `norm`? Okay, in that case you're right, although I don't see how to do that with enough generality to not break future implementations of `Fn`-like spaces.



> I would in this case assume that `super().norm()` gave me the "standard R^n norm", not some magic depending on how inner is implemented in this function.

Okay, admittedly this usage is a bit abuse of `super()` in the way that I assume that `inner()` is called. In general, however, your expectations of `super()` are not exactly in line with reality. Despite the name, it's not referring to a "superclass", but that's another topic. I make an explicit call here instead.

test/space/cu_ntuples_test.py
kohr-h: Oh yeah. This is one of the dangers of just going for 100% coverage..

kohr-h:
Note to self: TODO

- [ ] Intro to `discr`
- [ ] Intro to `operator`
- [ ] Intro to `set`
- [ ] Intro to `space`
- [ ] Optimize CUDA p-norm
- [ ] Optimize CUDA vector weighted norm

odl/space/ntuples.py
kohr-h: Given the fact that it can be either an `np.ndarray` or a `sp.sparse.spmatrix`, I don't really know how to call it other than `matrix`. That what it very much stands for, after all it's an operator of matrix multiplication with a vector. And even if somebody gives a `numpy.matrix`, it's internally converted to `np.ndarray`.

adler-j: Okay, ill yield this one!

odl/space/cu_ntuples.py
adler-j: Agree

kohr-h:
I'm wrapping this stuff up, but I can't get rid of the SciPy warnings in intersphinx. It does not find the `scipy.sparse.spmatrix` class (I also tried the actual location `scipy.sparse.base.spmatrix`).

I'll do another round of merge, then it's time to merge this stuff.

adler-j:
Simply make it ``scipy.sparse.spmatrix`` then, our users have google :)

kohr-h:
Alright, down to zero warnings again. Time for a new review.

We obviously haven't done the C++ work yet, and I've also only written one out of 5 intros. I'll make a separate issue out of that, it's not really related to this pull request.

kohr-h:
Bumping this.

adler-j:
Ok, I'm fine with this, simply solve the merge conflicts and merge to master. Also make a new issue to optimize the weighted cuda norms. We'll do that later.

Also make issues for the "introduction to ..." things.

kohr-h:
The CUDA issue exists [here](https://github.com/odlgroup/odlpp/issues/1)

PR #27, opened 2015-11-15
Issue 26  travis

adler-j:
Added travis-CI to get some simple CI for ODL, this is free since we are open source :)

For obvious reasons travis does not support the cuda part, and we dont get any regression tests on performance from it.

Questions

* What python versions should we test? Currently test 2.7 and 3.5, should we ad `PyPy`?
* Any smarter way to get our dependencies than miniconda? This is abit slow (a build takes about 4 min)


kohr-h:
Cool!

> What python versions should we test? Currently test 2.7 and 3.5, should we ad PyPy?

I think we should add Python 3.3. and 3.4, to be sure people with not-brand-new (distro) Python versions can us the code. Python 2.7 is sufficient, honestly, who uses older versions?

> Any smarter way to get our dependencies than miniconda? This is abit slow (a build takes about 4 min)

Maybe there is a possibility to cache downloaded packages? Otherwise, maybe we need to live with that slowness? Once we have our own local CI server we don't need to push here and wait that often.

This push history is a good candidate to cleanup. Try `git rebase -i <hash-of-last-good-commit>`.

kohr-h:
What about additional buttons like "coverage 100%" (:wink:)? Since our tests include a coverage report anyway, can one use that in travis?

adler-j:
It seems [coveralls](coveralls.io) is the way to go for coverage. The issue will be that it will report 0% coverage for CUDA. Ill look into it.

Issue #11, opened 2015-11-11
Operators on product spaces


adler-j:
There are a few `Operator`'s on product spaces that we could standardize and that would be usefull, see #239 for example.

The ones i can think of are the following (naming obviously TBD).

##### Distributive
The distributive operator, which calls a series of "sub" operators given the same argument. Think of the matrix:

```
[A,
 B] 
```

`DistributiveOperator(A,B)(x)  = [A(x), B(x)]`

##### Diagonal
The "diagonal" operator, which calls a series of operators "per space", think of the matrix

```
[A, 0,
 0, B]
```

`DiagonalOperator(A,B)((x,y))  = [A(x), B(y)]`

##### Sum
Finally we can think of calling a sequence of operators with given (different) arguments and somehow reducing the result. Reduction by summation is the obvious choice, but other options could work, think of the matrix:

```
[A, B]
```

`SumOperator(A,B)((x, y)) = A(x) + B(y)`

All of these operators could obviously be generalized to tuples of `n` operators, for example:

`DistributiveOperator(A, B, C, D)(x)  = [A(x), B(x), C(x), D(x)]`

I think we should add these to `odl.operator.default_ops`, but first I need some input on the names.

##### References

The first one ("Distributive") is called [vstack](http://docs.scipy.org/doc/numpy/reference/generated/numpy.vstack.html) in numpy

The second one seems to be [scipy.linalg.block_diag](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.linalg.block_diag.html)

and the last one is [hstack](http://docs.scipy.org/doc/numpy/reference/generated/numpy.hstack.html) 

The more I think of it, it would maybe be a good idea to not implement three separate classes for this, but instead to create one base class and then (maybe) have these as special cases. Something like a sparse matrix class (but for operators). That way the user could do complicated shit like

```
op = GeneralProductOperator([[A   , None, None, B   ],
                             [A   , None, C   , D   ],
                             [A   , C   , None, None]])
```

which would have the obvious meaning:

```
op([x, y, z, w]) == [A(x) + B(w), A(x) + C(z) + D(w), A(x) + C(y)]
```

This is similar to the [scip.sparse.bmat](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.sparse.bmat.html) method.

PR #61, opened 2015-11-19
Issue 11  pspace ops

adler-j:
Making a preliminary pull request here, could use some review. Doc needs to be written and style fixed.


odl/operator/pspace_ops.py
```diff
@@ -0,0 +1,190 @@
...
+        # Validate input data
+        if dom is not None and not isinstance(dom, ProductSpace):
+            raise TypeError('space {!r} not a ProductSpace instance.'
+                            ''.format(dom))
+        if ran is not None and not isinstance(ran, ProductSpace):
+            raise TypeError('space {!r} not a ProductSpace instance.'
+                            ''.format(dom))
+
+        # Convert ops to sparse representation
+        self.ops = sp.sparse.coo_matrix(operators)
```
kohr-h: With this we could definitely make good use of `dot()`, see #7.

kohr-h: Given that Scipy sparse matrices are smart enough to support blocks.

adler-j: Not really, dot would not be able to handle zero rows, nor would it be optimized like this for the case where there is only one operator per element in the range (in which case this makes no copies).

odl/operator/pspace_ops.py
```diff
@@ -0,0 +1,189 @@
...
+        --------
+        todo
+        """
+        has_evaluated_row = np.zeros(self.range.size, dtype=bool)
+        for i, j, op in zip(self.ops.row, self.ops.col, self.ops.data):
+            if not has_evaluated_row[i]:
+                op(x[j], out=out[i])
+            else:
+                # TODO: optimize
+                out[i] += op(x[j])
```
kohr-h: This optimization would require operators to be able to use an "in-place lincomb" instead of just overwriting, right?

kohr-h:
Very nice stuff in general. And the `ComponentProjection` would work directly with slices if I'm not mistaken. Could use a test.

odl/operator/pspace_ops.py
adler-j: Well that would be the best, a GEMM style call. But thats only useful in some cases, so maybe not worth it.

A more simple would be reusable temporaries.

adler-j:
It would not. We need the change to Productspace.Vector and a minor fix to _apply here. Not to far fetched tho.

adler-j:
Oh and there are tests available! Not that extensive though...

We could also use the special cases discussed in in #11

kohr-h:
I saw the tests, sure. Meant the slices thing.

kohr-h:
So I think this is really good and useful. I suggest that before merge, we need

- [ ] a bit more extensive tests
- [ ] slices support (also in `CoordinateProjection`)
- [ ] projection adjoint
- [ ] optimization in `_apply()`
- [ ] make it past the angry style checker

Question: Can one multiply two such operators using the matrix multiplication? Or is it not worth the trouble? From a user perspective, it would actually be nice to be able to multiply two such "matrices", possibly with an extremely stupid multiplication method.

adler-j:
I disagree on the optimization, agree on the others.

adler-j:
Any further ideas on tests?

kohr-h:
> I disagree on the optimization

What in particular? You write a TODO in the code

adler-j:
It will not be that important. I don't forsee us having "dense" `ProductSpaceOperator`'s) anytime soon, this is mainly as a general framework for the once we saw in the thread, we can do performance improvements when someone starts having performance issues (in special cases like this at least).

adler-j:
Only need some doc now, other than that I'm fine.

adler-j:
Doc massively improved, however, there is some issue with `operator` causing any import from it to cause a warning.

adler-j:
It would seem all of these warnings are present on master as well, so not this branch's fault.

odl/operator/pspace_ops.py
kohr-h: I fiddled around a bit with sparse matrix, and it seems impossible to initialize a CSR matrix without running into some C library code. It's simply not made for `dtype==object`, and when one creates an empty matrix with that type, there is an explicit warning about this.

That does not stop us from trying to use these operators a bit like matrices, especially when doing arithmetic. With dense matrices, it works out of the box, in fact, so for that part I guess (unless we're talking about huge operator arrays) we can simply use that, while for evaluation we store them in a sparse matrix as now. How about that?

adler-j: These can be quite large (think projection of a ct data set with 1000 angles, that would be 1000^2 elements or O(1000^3) for a dense mult) so i think we should use dense in all cases. Worse case we could roll our own later on.

adler-j:
Now also added `adjoint`, propose merge.

odl/operator/pspace_ops.py
kohr-h: Yes, for those sizes we cannot make dense matrices, I agree. Let's leave it for now, but maybe some explicit interface like `ProductSpaceOperator.matmul()` can be added in the future for this purpose, together with a warning that this can be very slow. Not a high prio anyway.

kohr-h:
The implementation of these new operators assumes no weighting of the product spaces. This can be fixed easily mathematically with two diagonal matrices. In code we'll have to see.

kohr-h:
Actually, this should be easy. The adjoint in the weighted case (with vector, i.e. diagonal matrix) can be written as the element-wise product of the transposed matrix of operator adjoints with the outer product matrix of the weighting vectors. That last one can be precomputed (in a sparse manner, of course), and then only a slight adaption in the `adjoint` method is required.

kohr-h:
Or, alternatively, the application of this outer product could be delayed until the adjoint is actually evaluated, i.e. pre-multiplying the input and post-multiplying the result with the (then separately stored) weighting matrices. This may be slightly advantageous with respect to pretty-printing (avoids `OperatorLeftScalarMult(OperatorRightScalarMult(...))`), but on the other hand it would duplicate the functionality of those operator scalar multiplications.

kohr-h:
I added some documentation for the math people :-). And I fixed the bad `import odl` in `wavelet.py`.

Otherwise we can look at the weighting thing tomorrow.

adler-j:
Nice doc, I suggest we simply disallow weighted spaces for now.

kohr-h:
Okay, I'll make a separate issue for that then. We're going to need it.

kohr-h:
Done. If Travis agrees, I'll merge.

Issue #49, opened 2015-11-19
Add function that gives matrix representation of operator


aringh:
I know that for large problems this is a BAD idea. However, for small test problems this might be a good things, especially to check things like conditioning of the problem (in my case it would be a 510x510 matrix, so it should be possible to handle).

What lead to the thought was a discussion with @ozanoktem. I have a nonlinear operator and want to solve `f(x) = 0`. I'm trying to do this with [Newton's method](https://github.com/odlgroup/odl/blob/master/odl/solvers/vector/newton.py), however the CG used there does not seem to find a solution to `gradf(x)p = -f(x)`, although  `gradf(x)` is a positive definite operator for all `x`. It would therefore be useful to be able to investigate the matrix, to see if it is ill-conditioned.

adler-j:
I have a implementation of this laying around

```python
def getMatrixForm(op):
    n = op.range.dimension
    m = op.domain.dimension

    matrix = np.zeros([n,m])
    for i in range(m):
        v = op.domain.zero()
        v[i] = 1.0

        matrix[:,i] = op(v)[:]
    return matrix
```

adler-j:
With that said, we could maybe add this to `utils`?

ozanoktem:
A good idea to add the getMatrixForm type of utility to utils. Who does that?

aringh:
@adler-j With the code you have written, would it work for discretized function spaces?

adler-j:
Yes, it works for any `FnBase` type space.

kohr-h:
Nice idea! Now the function just needs a proper (not-C++ style :wink:) name.

kohr-h:
Oh, and one should add a check for `op.is_linear`.

adler-j:
```python
def matrix_representation(op):
    assert op.is_linear
    assert isinstance(op.domain, FnBase)
    assert isinstance(op.range, FnBase)

    n = op.range.size
    m = op.domain.size

    matrix = np.zeros([n, m])
    v = op.domain.element()
    tmp = op.range.element()
    for i in range(m):
        v.set_zero()
        v[i] = 1.0
        matrix[:,i] = op(v, out=tmp)

    return matrix
```

aringh:
I tried this on my local code, and it does not work for product operators as it is written now (first I get assertion error for not FnBase, then n and m takes value 2, since my operator goes from a product space of two spaces into a product space of two other spaces). I will have a look at it.

kohr-h:
[...]

PR #73, opened 2015-11-23
Issue 49  matrix representation of linear operator

aringh:
Added the function for matrix representation of a linear operator; see issue #49


aringh:
There seems to be two errors in this build, both being more or less them same. However I'm not sure of what the problem is... where is the test supposed to be put? I thought that the structure of the tests was supposed to be the same as for the implementation, and thus I put it in test/discr/utility_test.py.

Can either @adler-j or @kohr-h have a look at this, whenever you have time?

kohr-h:
It obviously doesn't like the two identical file names `utility_test.py` under `discr/` and `util`. Why don't you put the code in `discr_mappings.py`? It wasn't intended for that from the beginning, but I think it fits there. As there is no test file for that module (I think), you can just rename your test file.

aringh:
I renamed it to operator_utilities. Looks ok?

odl/discr/operator_utilities.py
```diff
@@ -0,0 +1,78 @@
...
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
```
adler-j: A short module header goes here. Basically a few rows about what this module contains.

odl/discr/operator_utilities.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    if not op.is_linear:
+        print('WARNING: The operator is not linear; cannot produce matrix',
+              'representation of it.')
+        return
+
+    if not (isinstance(op.domain, FnBase) or
+            isinstance(op.domain, ProductSpace)):
```
adler-j: I dislike the productspace case here, would prefer if we got the coerce code to work instead.

test/discr/operator_utilities_test.py
```diff
@@ -0,0 +1,70 @@
...
+import pytest
+import numpy as np
+
+# ODL imports
+import odl
+from odl.discr.operator_utilities import matrix_representation
+from odl.util.testutils import almost_equal
+
+
+class MultiplyOp(odl.Operator):
```
adler-j: Use `MatVecOp` instead, it is tested and saves some space.

odl/discr/operator_utilities.py
```diff
@@ -0,0 +1,78 @@
...
+
+    n = op.range.size
+    m = op.domain.size
+
+    matrix = np.zeros([n, m])
+    v = op.domain.element()
+    tmp = op.range.element()
+    for i in range(m):
+        v.set_zero()
+        v[i] = 1.0
```
adler-j: Should fail for product spaces?

odl/discr/operator_utilities.py
aringh: Ok, then I will wait a while and see if we can resolve it later :)

odl/discr/operator_utilities.py
aringh: In this version, yes it must fail here if the range is a product space. Otherwise `n = op.range.size` will return the "size of the product space", i.e., how many spaces it is the product of. Therefore `matrix` will get the wrong size.

With that said, the current code should probably fail also for product spaces in the domain as well, otherwise you will get the same kind of problem... that was what I was trying to fix with [this suggestion](https://github.com/odlgroup/odl/issues/49#issuecomment-158406356)

test/discr/operator_utilities_test.py
aringh: Thanks, I was looking for it but I didn't know what it was called and I don't know in which module it is :) I took the code [from this test](https://github.com/odlgroup/odl/blob/master/test/operator/operator_test.py#L187-L206), so maybe this should be changed too?

aringh: I don't find the `MatVecOp` operator... where is it located?

kohr-h: It's called `MatVecOperator` and [located in `ntuples.py`](https://github.com/odlgroup/odl/blob/master/odl/space/ntuples.py#L1289)

kohr-h: > Thanks, I was looking for it but I didn't know what it was called and I don't know in which module it is :) I took the code from this test, so maybe this should be changed too?

Yes, that one should also use the shipped guy.

adler-j:
Almost done, still need `discreate` -> `discrete` in quite a few places, and this should be in `odl/operator` imo, not `odl/discr`.

kohr-h:
> Almost done, still need discreate -> discrete in quite a few places, and this should be in odl/operator imo, not odl/discr.

Maybe you're right. I suggested `discr` when I was starting to think that the operator alone would not suffice as information to calculate the matrix representation, but up to implementation errors (not consistent), this is possible. So it can go to operators. Just needs a better name. I'm not too creative now, cannot think of anything.

aringh:
I moved it to `operator` (and moved the test accordingly), changed the typos, and changed to use the `MatVecOperator` instead (also in the `operator_test` file).

The code works for `FnBase` only, so for product operators one need to use the methods suggested in #49. The following code works also for `ProducSpace` (as long as it is not a `ProductSpace` where one component is another `ProductSpace`...) Last call: should we keep the simple one or switch for this more complicated code?

```python
def my_matrix_representation(op):
    assert op.is_linear
    assert (isinstance(op.domain, odl.space.base_ntuples.FnBase) or
            isinstance(op.domain, odl.set.pspace.ProductSpace))
    assert (isinstance(op.range, odl.space.base_ntuples.FnBase) or
            isinstance(op.range, odl.set.pspace.ProductSpace))

    # Get the size of the range, and handle ProductSpace
    n = []
    op_ran = op.range
    if isinstance(op_ran, odl.set.pspace.ProductSpace):
        num_ran = op_ran.size
        for i in range(num_ran):
            n.append(op_ran[i].size)
    else:
        num_ran = 1
        n.appned(op_ran.size)

    # Get the size of the domain, and handle ProductSpace
    m = []
    op_dom = op.domain
    if isinstance(op_dom, odl.set.pspace.ProductSpace):
        num_dom = op_dom.size
        for i in range(num_dom):
            m.append(op_dom[i].size)
    else:
        num_dom = 1
        m.appned(op_dom.size)

    # Generate the matrix
    matrix = np.zeros([np.sum(n), np.sum(m)])
    tmp = op_ran.element()
    v = op_dom.element()
    index = 0
    for i in range(num_dom):
        for j in range(m[i]):
            v.set_zero()
            v[i][j] = 1.0
            tmp_two = op(v, out=tmp)
            tmp_three = []
            for k in range(num_ran):
                tmp_three = np.concatenate((tmp_three, tmp_two[k].asarray()))
            matrix[:, index] = tmp_three
            index += 1

    return matrix
```

adler-j:
Quite sure that code fails for `FnBase`, what would `v[0][0]` be in that case?

Anyway I also see the purpose of having this for `ProductSpace`, and since we have put some effort in it, why not. Just make sure to add a test for it.

aringh:
I have rebased this, in order to be able to use `ProductSpaceOperator`. This new implementation can take operators defined from `ProductSpace` (or `FnBase`) to `ProductSpace` (or `FnBase`), and there are a few test to try the code in different cases. More complicated code, but easier for the end-user.

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+    stacking the output as a matrix.
+
+    """
+
+    if not op.is_linear:
+        print('WARNING: The operator is not linear; cannot produce matrix',
+              'representation of it.')
+        return
+
+    if not (isinstance(op.domain, FnBase) or
```
kohr-h: This can be one instance check: `isinstance(op.domain, (FnBase, ProductSpace))`

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Notes
+    ----------
+    The algorithm works by letting the operator act on all unit vectors, and
+    stacking the output as a matrix.
+
+    """
+
+    if not op.is_linear:
+        print('WARNING: The operator is not linear; cannot produce matrix',
```
kohr-h: Printing and returning `None` will screw up code depending on the result. Raise an exception: `TypeError` if the type is wrong and `ValueError` if the operator is not linear. The sentence "Cannot produce [..]" is then redundant since the exception is raised in the `matrix_representation` function, so it's clear what goes wrong and where.

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+              'representation of it.')
+        return
+
+    if not (isinstance(op.domain, FnBase) or
+            isinstance(op.domain, ProductSpace)):
+        print('WARNING: The operator domain is not discrete or produc space;',
```
kohr-h: Don't use commas in print statements like these, but apply auto-concatenation:
```python
>>> raise ValueError('you gave me {}, '  # No comma at the end!
                     'but I expected {}!'.format(1, 2))
...
ValueError: you gave me 1, but I expected 2!
```
will work nicely because the concatenation (triggered by writing 'a' 'b') either re-triggers or even precedes the `format` method, so that `format` is evaluated on the concatenated string. Compare:
```python
>>> print('{}' '{}'.format(1, 2))
12

>>> print('{}', '{}'.format(1, 2))
{} 1

>>> print('{}' + '{}'.format(1, 2))
{}1
```

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+              'cannot produce matrix representation of it.')
+        return
+
+    if not (isinstance(op.range, FnBase) or
+            isinstance(op.range, ProductSpace)):
+        print('WARNING: The operator range is not discrete; cannot produce',
+              'matrix representation of it.')
+
+    # Get the size of the range, and handle ProductSpace
```
kohr-h: I can see why you store this instance check result. Maybe a short comment why? Something like `# Store for reuse in loop`

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+    op_ran = op.range
```
kohr-h: This is, however, a bit unnecessary to store. Nothing wrong with `op.range`.

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+    op_ran_is_prod_space = isinstance(op_ran, ProductSpace)
+    if op_ran_is_prod_space:
+        num_ran = op_ran.size
+        n = [op_ran[i].size for i in range(num_ran)]
```
kohr-h: This is ~~not pythonic~~ only half-pythonic :-). List comprehension is good, but don't iterate over an index if you don't have to. More elegant is to write
```python
n = [ran.size for ran in op.range]
```

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+    if op_dom_is_prod_space:
+        num_dom = op_dom.size
+        m = [op_dom[i].size for i in range(num_dom)]
+    else:
+        num_dom = 1
+        m = [op_dom.size]
+
+    # Generate the matrix
+    matrix = np.zeros([np.sum(n), np.sum(m)])
+    tmp_1 = op_ran.element()
```
kohr-h: I guess both are temporaries? In `operators` they have the name `tmp_dom` and `tmp_ran`.

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+    index = 0
+
+    for i in range(num_dom):
+        for j in range(m[i]):
+            v.set_zero()
+            if op_dom_is_prod_space:
+                v[i][j] = 1.0
+            else:
+                v[j] = 1.0
+            tmp_2 = op(v, out=tmp_1)
```
kohr-h: `tmp_2` is superfluous. In fact, after this call, `tmp_1 is tmp_2` is `True`.

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,114 @@
...
+            tmp_3 = []
+            if op_ran_is_prod_space:
+                for k in range(num_ran):
+                    tmp_3 = np.concatenate((tmp_3, tmp_2[k].asarray()))
```
kohr-h: This is very inefficient. `np.concatenate` is really only a last resort when you don't know beforehand or it is very hard to calculate how large your final array will be. Otherwise, it's always faster to allocate once and then fill the array by slicing. Quick speed test:
```python
>>> x = np.ones(5000)
>>> y = np.empty(500000)
>>> def assign(x, y):
...     for i in range(100):
...         y[i * 5000: (i + 1) * 5000] = x   # Slicing -> fast and memory-friendly

>>> def concat1(x):
...     y = []
...     for i in range(100):
...         y = np.concatenate((y, x))  # Your version
        
>>> def concat2(x):
...     lst = [x] * 100   # Need to store all arrays -> bad
...     y = np.concatenate(lst)  # Fast, only one call

>>> %timeit assign(x, y)
1000 loops, best of 3: 214 µs per loop

>>> %timeit concat1(x)
10 loops, best of 3: 17 ms per loop

>>> %timeit concat2(x)
1000 loops, best of 3: 191 µs per loop

Comment (kohr-h) [inline odl/operator/operator_utilities.py]: Of course you're waiting for #83 to happen :-)

File: odl/operator/operator_utilities.py [hunk 14]
```diff
@@ -0,0 +1,114 @@
...
+
+    """
+
+    if not op.is_linear:
+        print('WARNING: The operator is not linear; cannot produce matrix',
+              'representation of it.')
+        return
+
+    if not (isinstance(op.domain, FnBase) or
+            isinstance(op.domain, ProductSpace)):
```
adler-j: If it is a productspace, it has to be a productspace where all the subspaces are FnBase:

```
isinstance(op.domain, ProductSpace) and all(isinstance(spc, FnBase) for spc in op.domain)
```

test/operator/operator_test.py
```diff
@@ -213,7 +191,9 @@ def test_linear_Op():
     x = np.random.rand(3)
     out = np.random.rand(3)
 
-    Aop = MultiplyOp(A)
+    r3 = odl.Rn(3)
+
+    Aop = MatVecOperator(r3, r3, A)
```
adler-j: A note for later is that this syntax for `MatVecOperator` isnt optimal. I think automatic deduction of domain and range is very nice, especially when writing tests.

test/operator/operator_utilities_test.py
```diff
@@ -0,0 +1,146 @@
...
+    m = 2
+    rm = odl.Rn(m)
+    B = np.random.rand(m, m)
+    Bop = MatVecOperator(rm, rm, B)
+
+    ran_and_dom = ProductSpace(rn, rm)
+
+    AB_matrix = np.vstack([np.hstack([A, np.zeros((n, m))]),
+                          np.hstack([np.zeros((m, n)), B])])
+    ABop = ProductSpaceOperator([[Aop, None], [None, Bop]],
```
adler-j: The syntax

```python
ABop = ProductSpaceOperator([[Aop, 0], 
                             [0, Bop]])
```

is often a-bit nicer.

test/operator/operator_test.py
kohr-h: While I agree on the usefulness for tests, I don't think we should change to operator interface. Why should one be forced to use a complex matrix between complex spaces when a real one would do?

IMO, in general, class constructors should not try to be too clever and guess what the user wants. They should be low-level with explicit arguments, maybe a slight level of redundancy, but most of all flexible. If a class constructor is annoying to use over time, it's easy to write a convenience function. That has the positive side effect that users see functions instead of class constructors.

adler-j: Well having default values with optional selection is (imo) almost always better then forcing the users to always pick the obvious choice, see for example if users had to write `Rn(3, dtype=float)`.

aringh:
I think I have made changes according to all comments on the code (except the two comments that are not outdated). However now the build fails for python 2.7 since `TypeError: super() takes at least 1 argument (0 given)`. I don't see the difference in my call to `super` compared to, for example the `ProductSpaceOperator`: `super().__init__(domain=dom, range=ran, linear=linear)`. Can someone help me out on this one?

adler-j:
[from odl doc](http://odl.readthedocs.org/guide/faq.html#errors-related-to-python-2-3) :)

aringh:
Thank you :)

test/operator/operator_test.py
kohr-h: Valid point.

kohr-h:
> 11 code smells

@aringh your code stinks! ;-)

aringh:
Apparently it does :-p but the health still increases 0.10% ! ;-)  I just saw this landscape-bot, I had no idea that this existed.

kohr-h:
Neither did I. @adler-j must have added it in secret.

test/operator/operator_test.py
kohr-h: I implemented this now in e98cdc06f66ff31eadaaa7c09dc1a3de7820eb50. I suggest you rebase and, if you want, remove the `rn`s from the class initializer. If not, the matrix must come first now.

adler-j:
Added long time ago, never thought it would give angry messages here :O do we keep it?

aringh:
In principle I think that it is a good idea, like a complement to the build server in order to write nice and healthy code. However I guess we should take the output with a pinch of salt. I started looking at what it complains about (to see if I could go through and fix some); on example is [here](https://landscape.io/github/odlgroup/odl/111/modules/examples/simple_operator.py#L44). In the last line it complains that `op` is not callable, however I'm pretty sure that you can write `op(x)` also for an instance of the `AddOp` operator :p

test/operator/operator_test.py
aringh: Fixed in commit 872df47

test/operator/operator_utilities_test.py
aringh: Also fixed in 872df47

adler-j:
Its quite obnoxious, is it not? Maybe disabling the comment function (no idea why it was enabled suddenly).

kohr-h:
Yes, if it gives one comment per push, that's a bit too much. In principle, it can be handy, so we can keep it, I guess. Or maybe [codacy](https://github.com/integrations/codacy) is an alternative.

adler-j:
Fixed the most un-needed warnings. Now it's actually quite useful, if only it wouldn't write actual messages.

odl/operator/operator_utilities.py
```diff
@@ -0,0 +1,116 @@
...
+
+    """
+
+    if not op.is_linear:
+        raise ValueError('The operator is not linear')
+
+    if not (isinstance(op.domain, FnBase) or
+            (isinstance(op.domain, ProductSpace) and
+            all(isinstance(spc, FnBase) for spc in op.domain))):
+        raise TypeError('Operator domain {} is not FnBase, nor ProductSpace'
```
kohr-h: Almost good, just one blank missing - it will glue like "[...] nor ProductSpacewith only [...]"

test/operator/operator_utilities_test.py
```diff
@@ -0,0 +1,201 @@
...
+standard_library.install_aliases()
+from builtins import super
+
+# External module imports
+import pytest
+import numpy as np
+
+# ODL imports
+import odl
+from odl.operator.operator_utilities import matrix_representation
```
kohr-h: A bit too long a name for my gusto. `oputils` should do.

kohr-h:
> Fixed the most un-needed warnings. Now it's actually quite useful, if only it wouldn't write actual messages.

Comment on an old diff? I don't see any more `print` statements.

adler-j:
This is on the somewhat parallel landscape messages :)

kohr-h:
Got it! ;-)

odl/operator/operator_utilities.py
aringh: Done!

test/operator/operator_utilities_test.py
aringh: And done!

aringh:
Ok to merge?

kohr-h:
Looks good, let's do it!

kohr-h:
Let me fix two small things.

kohr-h:
Got a nice speedup by 3.3 with these optimizations. Basically trashed the `tmp_result` and instead of setting the complete `tmp_dom` back to zero, I only reset the one value I know I changed in the last step.

Issue #71, opened 2015-11-21
Make LinearSpace.Vector a non-member


adler-j:
I propse the simple change of:

```python
class SomeSpace(odl.LinearSpace):
    class Vector(odl.LinearSpace.Vector):
        pass
```

to

```python
class SomeSpace(odl.LinearSpace):
    @property
    def vector_type:
        return SomeSpaceVector

class SomeSpaceVector(odl.LinearSpaceVector):
    pass
```

The main reason is that sphinx and other tools really dont support nested classes, and with this approach each vector class gets a unique name. Any use of `SomeSpace.Vector` could still be replaced to `SomeSpace.vector_type`.

This also ties in nicely with issue #70.

kohr-h:
This is good. Python internally handles nested classes anyway this way (sort of) without giving them proper names. With the suggested change, things become explicit. I like that.

Issue #70, opened 2015-11-21
Add space.Vector to all


adler-j:
The vector classes, while not fully public are still part of the public interface and should be available to users. They need to be added to `__all__`

Issue #69, opened 2015-11-21
Autogenerate rst files


adler-j:
Currently there is some noticeable overhead in handling the documentation, for example adding a new submodule or class. The .rst files we currently have are all very similar and violate DRY quite badly. I suggest we make a python script to automatically generate these.

adler-j:
It seems the way to do this is to use `autosummary`. Sadly, it only seems to support class level summaries, so we may have to write a script to solve the issue on a module level, but this should be quite doable. Another required change seems to be to drop inner classes, e.g. issue #71.

I'll make a branch for this changes.

adler-j:
Attempt at this available in the new branch `issue-69__auto_doc`, still not getting everything to work, but it is much improved imo.

PR #74, opened 2015-11-23
Issue 69  auto doc

adler-j:
Initializing a pull request for this, but it is not done yet.

Now only has a few warnings left.


adler-j:
This branch also contains the fixes of #69, #70 and #71. A huge improvement is that we now will be able to document using 

    `target` 

instead of 

    :class:`odl.my.lib.target`

adler-j:
I actually recommend merge of this now.

kohr-h:
First of all, this is really good stuff. I like the auto-generation of the doc files and the fact that you don't need to add `:class:` everywhere.

Speaking of which, what is different now in the way one documents stuff? The answer to that, plus some general guide, should be added to our documentation.

adler-j:
It is already added in [document.rst](https://github.com/odlgroup/odl/blob/issue-69__auto_doc/doc/source/dev/document.rst).

kohr-h:
I fixed some bad LaTeX style causing warnings in the `pspace_ops` module and added / fixed some stuff in the "How to document" thing. Looks good (except the complaint by Coveralls).

Issue #81, opened 2015-11-25
Optimize weighted cuda reductions


adler-j:
The cuda `inner`, `norm` etc are not optimized at all, this should be done.

kohr-h:
There's [an issue on that in odlpp](https://github.com/odlgroup/odlpp/issues/1), not exactly as broad but in the same spirit.

adler-j:
Pull request submitted, looks good.

Issue #80, opened 2015-11-24
UFuncs for FnBase elements


adler-j:
It is possible (using the `__array_wrap__` interface) to define ufuncs for any `FnBaseVector`. How about we do this, and optionally allow overloading? Then users could write `my_vector.sin()` and it always works, but it COULD be overloaded. Specifically this would help with cuda.

kohr-h:
Hm, this doesn't immediately make sense to me, so I have to think about it properly. I guess you want to be able to use the same interface for NumPy and CUDA implementations and get rid of the module-level ufuncs in `cu_ntuples.py`?

adler-j:
Yes that is the idea. Basically only allow member ufuncs:

```python
>>> x = CudaRn(5)
>>> x = X.element([1, 2, 3, 4, 5])
>>> y = x.sin()
>>> z = X.element()
>>> x.sin(out=z)
```

PR #82, opened 2015-11-25
ENH: add optimized cuda reductions, see #81

adler-j:
Added the optimized versions of these, give the same result as before. Some notes:

* `dist_using_inner` is redundant in these cases now, removed.
* `Removed the option of using numpy vectors as weights since this cannot be properly optimized. Now requires a proper `CudaRnVector`. This causes some obvious issues if one wants to initialize a space using this (how to create the vector without the space?). Either users will need to create a convenience space for this, or we need to add something.
* `dist` and `norm` never worked for `exponent=inf`, and this is not fixed here either.
* Non integer powers were however added.

### DIFF
--- a/odl/space/cu_ntuples.py
+++ b/odl/space/cu_ntuples.py
@@ -1362,7 +1362,7 @@ class CudaFnConstWeighting(FnWeightingBase):
         self._const = float(constant)
         if self.const <= 0:
             raise ValueError('constant {} is not positive.'.format(constant))
-        if not np.isfinite(self.const):  # TODO: we actually support this
+        if not np.isfinite(self.const):
             raise ValueError('constant {} is invalid.'.format(constant))
 
     @property
@@ -1444,8 +1444,10 @@ class CudaFnConstWeighting(FnWeightingBase):
         norm : `float`
             The norm of the vector
         """
-        if self.exponent == float('inf'):  # Weighting irrelevant
-            return self.const * float(_pnorm_default(x, self.exponent))
+        if self.exponent == float('inf'):
+            raise NotImplementedError
+            # Example impl
+            # return self.const * float(_pnorm_default(x, self.exponent))
         else:
             return (self.const ** (1 / self.exponent) *
                     float(_pnorm_default(x, self.exponent)))
@@ -1464,7 +1466,9 @@ class CudaFnConstWeighting(FnWeightingBase):
             The distance between the vectors
         """
         if self.exponent == float('inf'):
-            return self.const * float(_pdist_default(x1, x2, self.exponent))
+            raise NotImplementedError
+            # Example impl
+            # return self.const * float(_pdist_default(x1, x2, self.exponent))
         else:
             return (self.const**(1 / self.exponent) *
                     _pdist_default(x1, x2, self.exponent))
--- a/odl/space/cu_ntuples.py
+++ b/odl/space/cu_ntuples.py
@@ -938,20 +938,20 @@ sign = _make_unary_fun('sign')
 sqrt = _make_unary_fun('sqrt')
 
 
-def _weighting(weight, exponent, dist_using_inner=False):
+def _weighting(weight, exponent):
     if np.isscalar(weight):
         weighting = CudaFnConstWeighting(
             weight, exponent)
     elif isinstance(weight, CudaFnVector):
         weighting = CudaFnVectorWeighting(
-            weight, exponent=exponent, dist_using_inner=dist_using_inner)
+            weight, exponent=exponent)
     else:
         weight_ = np.asarray(weight)
         if weight_.dtype == object:
             raise ValueError('bad weight {}'.format(weight))
         if weight_.ndim == 1:
             weighting = CudaFnVectorWeighting(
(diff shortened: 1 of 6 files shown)


odl/space/cu_ntuples.py
```diff
@@ -1366,12 +1358,11 @@ def __init__(self, constant, exponent=2.0):
...
         """
-        super().__init__(impl='cuda', exponent=exponent,
-                         dist_using_inner=False)
+        super().__init__(impl='cuda', exponent=exponent)
         self._const = float(constant)
         if self.const <= 0:
             raise ValueError('constant {} is not positive.'.format(constant))
-        if not np.isfinite(self.const):
+        if not np.isfinite(self.const):  # TODO: we actually support this
             raise ValueError('constant {} is invalid.'.format(constant))
```
kohr-h: Not so useful, though. I don't think anybody will miss it.

odl/space/cu_ntuples.py
```diff
@@ -1453,10 +1444,7 @@ def norm(self, x):
         norm : `float`
             The norm of the vector
         """
-        if self.exponent == 2.0:
-            from math import sqrt
-            return sqrt(self.const) * float(_norm_default(x))
-        elif self.exponent == float('inf'):  # Weighting irrelevant
+        if self.exponent == float('inf'):  # Weighting irrelevant
```
kohr-h: The comment is not valid any more, I changed the behavior after our discussions. Why `if` instead of `elif`?

adler-j: No reason (imo) to single out the exponent == 2 case. If someone uses this one we can kinda assume it is not 2, and if it is, the performance loss is minimal.

kohr-h:
> `dist_using_inner` is redundant in these cases now, removed.

Good. That was only an optimization if you don't have something better.



> Removed the option of using numpy vectors as weights since this cannot be properly optimized. Now requires a proper `CudaRnVector`. This causes some obvious issues if one wants to initialize a space using this (how to create the vector without the space?). Either users will need to create a convenience space for this, or we need to add something.

Yes, that's an obvious issue. We need some simple helper, like
```python
def make_cu_weight(arr, dtype):
    # Make Numpy array, check dimension
    return CudaFn(len(arr), dtype=dtype).element(arr)
```
and use that in the `CudaFn.__init__()`.

> dist and norm never worked for exponent=inf, and this is not fixed here either.

Hard to implement? I guess there are `abs()` and `max()` available?

odl/space/cu_ntuples.py
kohr-h: Ah, of course, I didn't see that the whole section on exponent 2 was deleted. Makes much more sense then.

adler-j:
> Yes, that's an obvious issue. We need some simple helper, like (...)

Quite honestly we could use a general convenience class for odl:

```python
def vector(inp, dtype=None, impl=None):
    if impl == 'numpy':
         return Rn(len(inp), dtype=dtype).element(inp)
    elif impl == 'cuda':
         return CudaRn(len(inp), dtype=dtype).element(inp)
```

> Hard to implement? I guess there are `abs()` and `max()` available?

Neither `max` nor `min` is available. Thus could (imo) be added as part of the greater UFunc push if we decide to do it.

kohr-h:
> Quite honestly we could use a general convenience class for odl: [...]

That's a good idea in general. The only question is where to put it due to the obvious CUDA dependency. It would almost make most sense in `odl/space/__init__py`.

> Neither max nor min is available. Thus could (imo) be added as part of the greater UFunc push if we decide to do it.

Sure, but that would be slow if I didn't understand #80 wrong. Is it possible to use [`thrust::max_element`](https://thrust.github.io/doc/group__extrema.html#gadd477c531dc4a4bf847f70ce73d01464)? No idea if one can sneak in the absolute value into the comparison.

adler-j:
> That's a good idea in general. The only question is where to put it due to the obvious CUDA dependency. It would almost make most sense in `odl/space/__init__py`.

I would prefer a new `odl/space/space_utils.py` or similar class. Keeping `__init__.py` clean is quite important imo.

> Sure, but that would be slow if I didn't understand #80 wrong. Is it possible to use [`thrust::max_element`](https://thrust.github.io/doc/group__extrema.html#gadd477c531dc4a4bf847f70ce73d01464)? No idea if one can sneak in the absolute value into the comparison.

It seems you missunderstand #80. What I mean is that we should have default implementations using numpy, then we could have optimizations of these (for example) for cuda. But Suppose we opt to not implement `left_shift`, then the system would fall-back to numpy for an implementation. That way, users can always assume that `FnBaseVector.left_shift` will work.

kohr-h:
> It seems you missunderstand #80. What I mean is that we should have default implementations using numpy, then we could have optimizations of these (for example) for cuda. But Suppose we opt to not implement left_shift, then the system would fall-back to numpy for an implementation. That way, users can always assume that FnBaseVector.left_shift will work.

Sure, that far I understood it correctly (just wrote it badly :-) ) that it's only a fallback. But for the inf-norm guy that would be the only one we have as for now.

adler-j:
Implemented the suggested changes, and by misstake some changes meant for master. Pulling this.

Issue #83, opened 2015-11-25
asarray() for power spaces


adler-j:
This could be implemented as calling asarray() on all subspaces, creating a N+1:d array. Another option is to create a linear array.

kohr-h:
Or both, using an optional boolean flag `contiguous` or `linear_arr` to differentiate.

adler-j:
That is definitely a decent middle way.

kohr-h:
Depends a bit on what one expects when asking a power space vector to become an array. Syntactically, one could index the vector with a single integer, `vec[0]`, which then gave the component vector. The corresponding array would be the n-dim. one so you can use the same syntax `arr[0]` there to get the same as `vec[0].asarray()`. Hence, by default, that makes most sense IMO. An option (maybe called `ravel` or `flat`) to flatten the array would be good to have for certain applications, I guess.

adler-j:
I agree, but we should also note that the `ndarray` version only works for "PowerSpace" type spaces, what about unmatched pairs? `ProductSpace(Rn(3), Rn(5))`?

kohr-h:
Correct, so then it's clear that only the flat version will fly. Any reason to make an optional n-dim array version for power spaces? If not, we don't differentiate.

adler-j:
Users could anyway themselves:

```python
>>> x.asarray().reshape(...)
```

If they want nd version.

Another argument for the linear version is that we can make it recursive, so it will work on productspaces of productspaces.

kohr-h:
That's an interesting issue. Using the `find_common_type` sounds like a good idea. Tests will reveal if that's true.

kohr-h:
This could be solved as part of #857 by going through the equivalent tensor-based power space.

adler-j:
Well that is part of the problem, but what about

```python
>>> pspace = odl.ProductSpace(odl.rn(5), odl.rn(6))
>>> array = pspace.zero().asarray()
```

what is `array` in this case?

kohr-h:
Well, the issue was about power spaces from the beginning, so that's strictly speaking another issue.

Not sure what to do in the general case. Is a single array ever needed for general product spaces or is it rather so that different things will happen in different components anyway? Strongly depends on the set of use cases.

adler-j:
Well any user would expect some form of continuity, i.e if we make different choices we would get weird situations like

```
>>> pspace = odl.ProductSpace(odl.rn(2), odl.rn(3))
[...]

Issue #59, opened 2015-11-19
Coercion operator(s)


adler-j:
There are many cases in mathematics where a user "inteprets an X as an Y", often very subtly or unstated. Nonetheless, this needs to be made explicit at times. Some examples in odl include

* Overlook `FnBase` details, 
    * `Rn(10)` as `CudaRn(10)`
    * `Rn(10)` as `Rn(10, dtype='float32')
* Interpret elements in `ProductSpace(X, 1)` as in `X`, or the inverse of this
* Overlooking discretization details. `uniform_discr(FunctionSpace(Interval(0, 1)), 10)` as `uniform_discr(FunctionSpace(Interval(0, 3.1415)), 10)`
* Overlook dspace of `Discretization` according to the above

To solve this we should implement some kind of `CoerceOperator`, example below for most simple `FnBase` case

```python
class CoerceOperator(odl.Operator):
    def __init__(self, dom, ran):
        super().__init__(dom, ran, True)

    def _call(self, x):
        return self.ran.element(x.asarray())

    def _apply(self, x, out):
        out[:] = x
```

kohr-h:
This is definitely useful. "Coerce" is more of a computer science notion AFAIK, this makes sense if changing only data type within the same "family", or changing implementation. In math one would rather speak of an embedding.

adler-j:
From the discussion in #49 i found another case:

* ProductSpaces of `FnBase` spaces (possibly only of the same kind) as a single space:
    `ProductSpace(Rn(3), Rn(2))  <--> Rn(5)`
    this would solve many of the problems in #49 aswell.

adler-j:
Suggested programming of this:

* Make a new file under `operator`, call it `coercion` or `embedding` if we prefer that. (@ozanoktem ?)
* Add classes for the cases we have above.
* make a `coerce`/`embedding` "entry point" function, so users can call

```
>>> coerce(Rn(10), CudaRn(10))
FnBaseCoercion(Rn(10), CudaRn(10))
>>> coerce(uniform_discr(FunctionSpace(Interval(0, 3.1415)), 10), 
           uniform_discr(FunctionSpace(Interval(0, 1)), 10))
USpaceCoercion(uniform_discr(FunctionSpace(Interval(0, 3.1415)), 10), 
               uniform_discr(FunctionSpace(Interval(0, 1)), 10))
```

etc...

adler-j:
Followup, it seems `coerce` is already taken, `embed` it is then!

PR #95, opened 2015-11-26
Issue 59  coerce

adler-j:
Initial attempt at issue #59. Not ready for pull.

Some points of discussion/additions
- Verify `DiscreteLp` <-> `DiscreteLp` especially if they have different data orders.
- Look into optimizations (should be able to be zero overhead in some situations).
- I was forced to add a method `FnBaseVector.asflatarray`. I highly dislike this, but it shows a problem we have in `DiscreteLpVector.asarray`, namely `x.asarray()[index] != x[index]`, imo this doesnt make much sense. I would suggest we move the current behaviour to a method called `DiscreteLpVector.asvoxels()` (or preferably some better name), and leave the `asarray` behaviour to match that of `Discretization`.

Edit: Also solved issue #83 since it was "needed" for this.

### DIFF
--- a/odl/operator/embedding.py
+++ b/odl/operator/embedding.py
@@ -1,0 +1,105 @@
+# Copyright 2014, 2015 Jonas Adler
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Operators to cast between spaces."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import super
+
+from odl.operator.operator import Operator
+
+__all__ = ('Embedding', 'EmbeddingFnInFn', 'EmbeddingPowerSpaceInFn',
+           'EmbeddingFnInPowerSpace')
+
+
+class Embedding(Operator):
+    """An operators to cast between spaces."""
+    def __init__(self, target, origin):
+        super().__init__(origin, target)
+
+
+class EmbeddingFnInFn(Embedding):
+    """ Embed a `Fn` space into another """
+
+    def __init__(self, target, origin):
+        # TODO: tests goes here
+
+        super().__init__(target, origin)
+
+    def _apply(self, x, out):
+        out[:] = x.asflatarray()
+
+    def _call(self, x):
+        return self.range.element(x.asflatarray())
+
+    @property
+    def adjoint(self):
+        return EmbeddingFnInFn(self.domain, self.range)
+
+
+class EmbeddingPowerSpaceInFn(Embedding):
+    """ Embed a `PowerSpace` of `Fn` space into a `Fn` """
+
+    def __init__(self, target, origin):
+        # TODO: tests goes here
+
+        super().__init__(target, origin)
+
+    def _apply(self, x, out):
+        out[:] = x.asflatarray()
+
+    def _call(self, x):
+        return self.range.element(x.asflatarray())
+
+    @property
+    def adjoint(self):
+        return EmbeddingFnInPowerSpace(self.domain, self.range)
+
+
+class EmbeddingFnInPowerSpace(Embedding):
+    """ Embed a `Fn` into `PowerSpace` of `Fn`'s"""
+
+    def __init__(self, target, origin):
+        # TODO: tests goes here
+
+        super().__init__(target, origin)
+
+    def _apply(self, x, out):
+        index = 0
+        for sub_vec in out:
+            sub_vec[:] = x.asflatarray(index, index + sub_vec.size)
+            index += sub_vec.size
+
+    def _call(self, x):
+        out = self.range.element()
+        index = 0
+        for sub_vec in out:
+            sub_vec[:] = x.asflatarray(index, index + sub_vec.size)
+            index += sub_vec.size
+        return out
+
+    @property
+    def adjoint(self):
+        return EmbeddingPowerSpaceInFn(self.domain, self.range)
+
+
+if __name__ == '__main__':
+    from doctest import testmod, NORMALIZE_WHITESPACE
+    testmod(optionflags=NORMALIZE_WHITESPACE)
--- a/odl/operator/embedding.py
+++ b/odl/operator/embedding.py
@@ -23,6 +23,9 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import super
 
+from odl.space.base_ntuples import FnBase
+from odl.discr.lp_discr import DiscreteLp
+from odl.set.pspace import ProductSpace
 from odl.operator.operator import Operator
 
 __all__ = ('Embedding', 'EmbeddingFnInFn', 'EmbeddingPowerSpaceInFn',
@@ -31,17 +34,21 @@ __all__ = ('Embedding', 'EmbeddingFnInFn', 'EmbeddingPowerSpaceInFn',
 
 class Embedding(Operator):
     """An operators to cast between spaces."""
-    def __init__(self, target, origin):
+
+    # TODO: is this needed?
+
+    def __init__(self, origin, target):
         super().__init__(origin, target)
 
 
 class EmbeddingFnInFn(Embedding):
-    """ Embed a `Fn` space into another """
+    """ Embed a `FnBase` space into another """
 
-    def __init__(self, target, origin):
-        # TODO: tests goes here
+    def __init__(self, origin, target):
+        assert isinstance(origin, FnBase)
+        assert isinstance(target, FnBase)
 
-        super().__init__(target, origin)
+        super().__init__(origin, target)
 
     def _apply(self, x, out):
         out[:] = x.asflatarray()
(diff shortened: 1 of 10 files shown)


odl/operator/embedding.py
```diff
@@ -0,0 +1,105 @@
...
+
+from odl.operator.operator import Operator
+
+__all__ = ('Embedding', 'EmbeddingFnInFn', 'EmbeddingPowerSpaceInFn',
+           'EmbeddingFnInPowerSpace')
+
+
+class Embedding(Operator):
+    """An operators to cast between spaces."""
+    def __init__(self, target, origin):
```
kohr-h: Why do you switch order here? IMO it makes much more sense to mimic `Operator`.

adler-j: This has plagued me. Basically it "feels good" to write (if `op` has `Rn` as domain)

```python
op * embed(rn, cudarn)
```

This is because you can read it nicely from the right. In the end I'll have to agree however, keeping things "as is" is probably better than having a somewhat more convenient special case here.

kohr-h: I would always read this as "embed `rn` into `cudarn`", but I'm spoiled by math :-)

kohr-h: And the "feel bad" vs "feel good" is the same as in math, where you have to think twice when your read `A o B` and need to map domains and ranges. I can see your point, reading as a "feed" from right to left is somewhat charming, but it interferes too much with the "math common sense".

adler-j: Will get this fixed.

kohr-h:
> was forced to add a method FnBaseVector.asflatarray. I highly dislike this, but it shows a problem we have in DiscreteLpVector.asarray, namely x.asarray()[index] != x[index], imo this doesnt make much sense. I would suggest we move the current behaviour to a method called DiscreteLpVector.asvoxels() (or preferably some better name), and leave the asarray behaviour to match that of Discretization.

I see the problem. The *proper* solution to this would be multi-indexing, but that's a large step to take on the CUDA side for such a relatively small fix. The second best solution is what you did. We can for sure adapt the names, maybe `asmultiarray()` would fit the picture. But in the long run, I think we should keep the pain to switch to multi-indexing in `DiscretizationVector` by default as painless as possible.

kohr-h:
> Verify DiscreteLp <-> DiscreteLp especially if they have different data orders.

Probably some options needed, definitely (imo) `strict=True` which would disallow embedding `Lp` into `Lq` when `p > q`. For the math nazis.
The reordering issue could be controlled with a `allow_reorder=False` switch.

adler-j:
Multi-indexing would also require any further implementations to provide full slicing, I dislike that.(what if a user wanted to use ASTRA storage, would it need slicing?). We'll have to think abit more about this.

kohr-h:
It's not a super pressing issue anyway. Linear arrays are good since anybody supports them, while multiarrays are fancy and in 90 % of the cases require flattening to interact with other libraries. Let's postpone that.

odl/operator/embedding.py
```diff
@@ -0,0 +1,133 @@
...
+        return EmbeddingFnInFn(self.range, self.domain)
+
+
+class EmbeddingDiscreteLpInDiscreteLp(Embedding):
+    """ Embed a `DiscreteLp` space into another """
+
+    def __init__(self, origin, target):
+        assert isinstance(origin, DiscreteLp)
+        assert isinstance(target, DiscreteLp)
+        assert origin.grid.shape == target.grid.shape
```
kohr-h: This should only be possible if `target.exponent <= origin.exponent` for mathematical reasons.

odl/operator/embedding.py
```diff
@@ -0,0 +1,133 @@
...
+class Embedding(Operator):
+    """An operators to cast between spaces."""
+
+    # TODO: is this needed?
+
+    def __init__(self, origin, target):
+        super().__init__(origin, target)
+
+
+class EmbeddingFnInFn(Embedding):
```
kohr-h: Small thing: `Into` instead of `In`

odl/operator/embedding.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Embed a `FnBase` space into another """
+
+    def __init__(self, origin, target):
+        assert isinstance(origin, FnBase)
+        assert isinstance(target, FnBase)
```
kohr-h: I'm wondering if we could replace all our obvious type checks with asserts? At least when the docstring says what you can provide, the error message is easy to relate to the error source.

odl/operator/embedding.py
```diff
@@ -0,0 +1,133 @@
...
+        return self.range.element(x.asflatarray())
+
+    @property
+    def adjoint(self):
+        return EmbeddingFnInFn(self.range, self.domain)
+
+
+class EmbeddingDiscreteLpInDiscreteLp(Embedding):
+    """ Embed a `DiscreteLp` space into another """
+
```
kohr-h: This operator could also handle extension by zero (constant? nearest value?) to a larger domain.

odl/set/pspace.py
```diff
@@ -458,6 +466,85 @@ def size(self):
...
+
+        index = 0
+        for part in self.parts:
+            if isinstance(part, ProductSpaceVector):
+                out[index:index + part.total_size] = part.asarray()
+                index += part.total_size
+            else:
+                out[index:index + part.size] = part.asarray()
+                index += part.size
+
```
kohr-h: I'm trying to understand how this works. If `part` is some linear vector, then this is all fine. What happens if `part` is a `DiscreteLpVector` on some higher-dimensional grid? Then the statement in 529 would fail. I guess it should be `asflatarray()` there?

adler-j: yes, it should be `asflatarray()`, still kinda dislike that name tho.

odl/operator/embedding.py
adler-j: I thought about this before, but I'm not quite sure. Anytime a user could give the wrong input we should give an error imo. Asserts are to make sure we havnt made any algorithmic error.

odl/operator/embedding.py
adler-j: Only with some strict constraints, for example the values need to coicide in some points at least approximately if we want to do this on a `dspace` level. Another option would be to walk the long walk with restriction and extensions, but I fear for the performance.

kohr-h:
Well, for now it seems like the only way of getting away with it. I don't think it's so bad. There's still the alternative suggestion to make flat default and call the other one `asmultiarray`.

odl/operator/embedding.py
kohr-h: I just tried it in the console. `assert` is really very short on information, you almost never get away without looking into the code. Better proper messages.

odl/operator/embedding.py
adler-j: Not sure about that, embedding (to me) means, "Do as your told and ignore the consequences". Kindoff like how mathematicians often "identify" things with others. Perhaps the vectors given to this will only be in a subspace where it still works? I think we need a way for users to "force" this through.

odl/operator/embedding.py
adler-j: Sure!

adler-j:
How about `asndarray`? Kinda double meaning, both tells you nd and tells you its a `ndarray`.

odl/operator/embedding.py
kohr-h: This guy could take a parameter `strict=True` which would check the mathematical condition. To me, an embedding is the identity mapping from a space to a superspace, not simply closing your eyes. If users want to go for it, they can set `strict=False`. We can also call it `force=False`.

adler-j: >This guy could take a parameter strict=True which would check the mathematical condition.

Sure, but I'd default it to False.

>To me, an embedding is the identity mapping from a space to a superspace, not simply closing your eyes.

Then this needs to be renamed. Very few of these are actually to sub-spaces, and the adjoint would then never be defined (which obviously isn't that useful).

odl/operator/embedding.py
kohr-h: On a `dspace` level, only some restricted operations are possible, that's for sure. The best you can hope for is that your old data stays together as a chunk in the new flat array.

But in general, it's hard to get around using restriction and extension, and then one can implement a much wider range of operations going beyond pure embedding, e.g. subsample by interpolation. With vectorization, this should not be too inefficient. The good part here is also that the interpolation comes along with some options for how to handle values outside.

Conclusion: alright, keep it simple and implement the fancy stuff in a different operator.

odl/operator/embedding.py
kohr-h: > Then this needs to be renamed. Very few of these are actually to superspaces, and the adjoint would then never be defined (which obviously isn't that useful).

It's a projection then. In that case I'm good with not checking the exponent.

odl/operator/embedding.py
adler-j: Agree.

adler-j:
Ok, things left to consider here:

- [ ] Naming of `asflatarray`, is it good?
- [ ] Adjoint handling (it should be correct)
- [ ] Add inverse
- [ ] Naming, should we change `Embedding` to something else?

kohr-h:
> Naming of `asflatarray`, is it good?

I think it's actually better this way, that we internally have to use the less intuitive (`asflatarray`) method while users can go with `asarray()` and get a nice reshaped multiarray.



> Naming, should we change `Embedding` to something else?

Mathematically, the functionality defines a projection. Embedding is not correct, and I don't think we should callit like that then. 
Projection unfortunately collides with the notion of projection in tomography, and we should avoid that. Maybe `ProjectionMapping`?

adler-j:
> I think it's actually better this way, that we internally have to use the less intuitive (`asflatarray`) method while users can go with `asarray()` and get a nice reshaped multiarray.

OK, But I think numpy has some of these around, I'll see if they have a naming convention.

> Mathematically, the functionality defines a projection. Embedding is not correct, and I don't think we should callit like that then. 
Projection unfortunately collides with the notion of projection in tomography, and we should avoid that. Maybe `ProjectionMapping`?

You are correct but I don't quite like the sound of that (a bit too long when you start adding cases). I'll think about it some more.

adler-j:
This has to be taken up at some point, Ill see if I can get to it. Meanwhile regarding `asflatarray`, numpy has a reference in `nonzero` and `flatnonzero`:

https://github.com/numpy/numpy/blob/183fdb290cd46b1f01303d24ac0c9fc3ff24fe05/numpy/core/numeric.py#L791

adler-j:
Should we close this for now? I guess this is so old a new effort would be needed if we want it. I wont delete the branch, for now.

kohr-h:
There doesn't seem to be an urgent need for it right now, so I guess we can have the code lying around in this branch and close the issue for now, as you suggested.

Issue #64, opened 2015-11-20
Profiler run


adler-j:
In regards to #63 Test runtimes, we should at some point do a profiler run on different platforms and remove any obvious bottlenecks.

Issue #63, opened 2015-11-20
Test runtimes


adler-j:
Currently the full test suite takes me 10 seconds, we should try to keep this down and lot have it explode. Tests should be ran all the time and increasing this much more would hinder it.

PR #102, opened 2015-11-27
Issue 64  profile

adler-j:
I've done some of this, issues #64 and #63. Tests now run more than twice as fast.

* We had a few leftover `super()` in ntuples. Since we create a new space anytime we use `__getitem__`, anything where it is used can be considered "performance critical" and should use the named ancestor. Shaved off roughly 30% run-time on the tests by removing this.
* Roughly 20% of the test time is in `all_almost_equal`. I did a few optimizations here that should help (reduced from ~50%).
* In the vector classes we do `isinstance` checks on the input. I propose that we document that the vector classes are not intended to be used separately and that users should use `element`. That way we can remove a few `isinstance` calls here (that will always be true).
* There seems to be a `__instancecheck__` called for every ABCMeta inheriting class on construction. This uses about 20% of all runtime. There could also be a overhead in our `isinstance` calls since they seem to use a un-optimized python version instead of C code. We may actually want to consider totally dropping ABC and simply going for `raise NotImplementedError` in the methods.
* We used to spend about 20% of all run-time doing `int(indices)` as in:

    ```python
    try:
        return self.data[int(indices)]
    except TypeError:
        ...
    ```
    
    Since `int` is very expensive (if we want to use `future` ints), it is faster to do
    
    ```python
    if isinstance(indices, Integral):
        return self.data[indices]
    else:
        ...
    ```

### DIFF
--- a/test/space/ntuples_test.py
+++ b/test/space/ntuples_test.py
@@ -23,7 +23,6 @@ standard_library.install_aliases()
 from builtins import range
 
 # External module imports
-from math import ceil
 import numpy as np
 import pytest
 import scipy as sp
@@ -49,12 +48,12 @@ from odl.util.testutils import almost_equal, all_almost_equal, all_equal
 def _array(fn):
     # Generate numpy vectors, real or complex or int
     if np.issubdtype(fn.dtype, np.floating):
-        return np.random.rand(fn.size).astype(fn.dtype)
+        return np.random.rand(fn.size).astype(fn.dtype, copy=False)
     elif np.issubdtype(fn.dtype, np.integer):
-        return np.random.randint(0, 10, fn.size).astype(fn.dtype)
+        return np.random.randint(0, 10, fn.size).astype(fn.dtype, copy=False)
     elif np.issubdtype(fn.dtype, np.complexfloating):
         return (np.random.rand(fn.size) +
-                1j * np.random.rand(fn.size)).astype(fn.dtype)
+                1j * np.random.rand(fn.size)).astype(fn.dtype, copy=False)
     else:
         raise TypeError('unable to handle data type {!r}'.format(fn.dtype))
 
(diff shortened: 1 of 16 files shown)


adler-j:
Added some further commits, not much more to add than the previous message. Running the test-suite now takes 2.28 seconds, compared to 5.38 on master. Shaved of 58% of the time.

Biggest culprits is still `isinstance` with abstract classes, and I would also say typechecking in the vector classes (which should be done away with since they shouldn't be used that way.

test/space/ntuples_test.py
```diff
@@ -80,35 +80,26 @@ def _pos_array(fn):
...
-    coo_c = np.random.randint(0, fn.size, size=nnz)
+def _dense_matrix(fn):
+    """Create a dense positive definite Hermitian matrix for `fn`."""
+
     if np.issubdtype(fn.dtype, np.floating):
-        values = np.random.rand(nnz).astype(fn.dtype)
+        mat = np.random.rand(fn.size, fn.size).astype(fn.dtype)
     elif np.issubdtype(fn.dtype, np.integer):
-        values = np.random.randint(0, 10, nnz).astype(fn.dtype)
+        mat = np.random.randint(0, 10, (fn.size, fn.size)).astype(fn.dtype)
```
kohr-h: `astype` can take a `copy=False` option which optimizes the case of the data type returned by the `np.random` functions.

kohr-h:
I give up insisting on `super()`, at least in the case of `__init__()` and as long as we use the backported Python 2 thing. The native Py3 `super` should be fast. `ABCmeta` can go, I won't miss it. Agree also on the element type classes.

test/space/ntuples_test.py
```diff
@@ -218,7 +209,7 @@ def test_vector_class_init(fn):
     arr = _array(fn)
 
     FnVector(fn, arr)
-
+"""
```
kohr-h: Is this test removed for a reason or just forgotten to be put back in?

kohr-h:
I added `copy=False` to the `astype()` calls, cutting off another 10 % in the Ntuples tests. ABCMeta will be dropped in the operator branch. Merge?

adler-j:
Fine with merging then.

test/space/ntuples_test.py
adler-j: Was removed when I removed the tests in FnVector, but when I re-added that it shoulda been added back in.

Issue #90, opened 2015-11-25
Diagnostics extremely slow on large data


adler-j:
It seems the examples are not vectorized anymore, possibly related to the bug in issue #89. This causes the generation of realistic size example to take way to long. We need to fix this vectorization.

kohr-h:
Yes, it's on my extended agenda until the beginning of December. I noticed that vectorization is completely off currently, but I didn't bother to make a quick fix for now. The large overhaul lags a bit behind, but I'll take it up asap.

adler-j:
This would be quite nice to have since this bug renders the whole diagnostics package moot untill fixed.

kohr-h:
Closed by #116

Issue #89, opened 2015-11-25
Bug in vector_examples for nd data


adler-j:
The [step function](https://github.com/odlgroup/odl/blob/676f8f43cd8fd785797ab67b4eabafb563da5d3a/odl/diagnostics/examples.py#L112) example fails with an error:

```python
  File "/home/jonasadler/Programming/ODL/odl/diagnostics/examples.py", line 112, in _step_fun
    z[:space.grid.shape[0] // 2, ...] = 1

IndexError: too many indices for array
```

adler-j:
Fixed by vectorization branch.

PR #107, opened 2015-11-30
Issue 97  stir

adler-j:
STIR bindings up and running, can reconstruct using landweber with no problems.

Obviously needs a review and possibly some added features. We currently support any projection using a input file, but this should be extendable to the general case. The current method also uses a projection matrix, but we could change this.

@niinimaki could perhaps give some input.
@ozanoktem would also be interest in this.


odl/tomo/backends/stir_bindings.py
```diff
@@ -0,0 +1,148 @@
...
+from builtins import super
+
+from odl.operator.operator import Operator
+
+try:
+    import stir
+    STIR_AVAILABLE = True
+except ImportError:
+    STIR_AVAILABLE = False
+
```
kohr-h: I'm never absolutely sure about this, but it may be necessary to make a mock `stir = None` so you don't get `NameError` later.

adler-j: I dont see where this could cause problems, its only used inside functions not in declarations.

kohr-h: Yes, right. Travis didn't complain either. I think there was a problem in CUDA tests before we had the skip decorator, then it would complain about name errors. No longer relevant as it seems.

kohr-h:
I don't think there is anything wrong with this implementation, it's a good starting point.

I'm not so happy with the names, however. `Stir` and `FromFile` are implementation details and don't have anything to do with the transform. IMO it should be called `ExponentialXrayTransform` and have an `impl` parameter (here e.g. `'stir_file'`) rather than carrying implementation details as ballast in the name.

One thing we should keep in mind is a mathematical one: what are sensible domain and range spaces for this transform? Need to look this up in the literature.

adler-j:
> I don't think there is anything wrong with this implementation, it's a good starting point.

Great! Perhaps @niinimaki can take it further.

> I'm not so happy with the names, however. `Stir` and `FromFile` are implementation details and don't have anything to do with the transform. IMO it should be called `ExponentialXrayTransform` and have an `impl` parameter (here e.g. `'stir_file'`) rather than carrying implementation details as ballast in the name.

One option would be to make the class general and then make a factory function. Another is as you say, a parameter, if that doesn't make the interface cluttered.

I disgree with `ExponentialXrayTransform`, the best would be to actually call this something ridicolous that says what it does in stir STIR:

`ForwardProjectorByBinUsingProjMatrixByBinWrapper`

If we later want, we can make a "on top" wrapper/factory function that allows users to specify what they want, but in the short run we will likely get several of these, and it will be easier for users to understand if we simply tell them what we are wrapping.

> One thing we should keep in mind is a mathematical one: what are sensible domain and range spaces for this transform? Need to look this up in the literature.

Agree. Quite sure the current ones are wrong (but cannot run the diagnostics until #89, #90 are solved)

kohr-h:
> I disgree with `ExponentialXrayTransform`, the best would be to actually call this something ridicolous that says what it does in stir STIR:
> 
> `ForwardProjectorByBinUsingProjMatrixByBinWrapper`
> 
> If we later want, we can make a "on top" wrapper/factory function that allows users to specify what they want, but in the short run we will likely get several of these, and it will be easier for users to understand if we simply tell them what we are wrapping.

I don't see how your suggestion would be "best" in any way. And it goes into the wrong direction in my opinion. We're not building a wrapper package. We're building a math library (at least something more similar to that than a wrapper), and in math we're talking about the exponential X-ray (or Radon) transform. That's what the operator should compute, no matter which package it uses, if it needs a matrix, what technique the projector implements etc. That's all irrelevant a the highest level.

That said, I admit that it may be too much asked to wrap everything in one single class since there are just too many options. But we should at least try to map variants in implementation to concepts we already have or, alternatively, make them an implementation flag. For example, the `ByBin` sounds very much like a discretization and should be mapped to what we have. Using a matrix is an implementation detail, and I don't think this requires a new class. Communicating with STIR via Interfile is also an implementation detail, and I'd rather not have that as a part in a class name.

A compromise can be to make all these ridiculous wrapper classes in some extra module and NOT clutter the `odl.tomo` namespace with them, but rather create them in a factory function (or class) which has a nice name and just a few parameters.

adler-j:
> I don't see how your suggestion would be "best" in any way. And it goes into the wrong direction in my opinion. We're not building a wrapper package. We're building a math library (at least something more similar to that than a wrapper), and in math we're talking about the exponential X-ray (or Radon) transform. That's what the operator should compute, no matter which package it uses, if it needs a matrix, what technique the projector implements etc. That's all irrelevant a the highest level.

Best may be a strong word, more like currently the best choice. I agree that it is irrelevant *at the highest level*, but many users will want to use this intermediate level, at least until we have a very good general geometry description, which is not in place yet, but also after that if they have some wierd geometry they may still want to use these "deep" wrappers, probably more so than we'd prefer.

> That said, I admit that it may be too much asked to wrap everything in one single class since there are just too many options. But we should at least try to map variants in implementation to concepts we already have or, alternatively, make them an implementation flag. For example, the ByBin sounds very much like a discretization and should be mapped to what we have. Using a matrix is an implementation detail, and I don't think this requires a new class. Communicating with STIR via Interfile is also an implementation detail, and I'd rather not have that as a part in a class name.

I'm quite fine with having it as part of the class name if it is integral to how you use the class. In this case you construct it using file data, which sets it apart from some other which may use parameter data. As said above, this is only used in the initialization, so i suppose we could make a `ForwardProjectorByBinUsingProjMatrixByBinWrapper` and then have sub-classes or parameters or a factory depending on how you instantiate these.

Regarding copying names it is for users that simply want to use their earlier knowledge of STIR (old STIR users, for example) and perhaps old input scripts they have hanging around, but use the power of odl to do tomography and algorithm development, this is a decent middle way and will help adoption, which imo is very important.

> A compromise can be to make all these ridiculous wrapper classes in some extra module and NOT clutter the odl.tomo namespace with them, but rather create them in a factory function (or class) which has a nice name and just a few parameters.
    
Agree that this is the best *once we have a PET geometry description*, then we can have "nice" wrappers/factories taking some parameters (like a `PET-geometry` object) and giving a `StirProjector` object or similar, we could then simply not add these ridiculously named classes to `odl.tomo.backends.stir_bindings.__all__` so users need to explicitly look for them.

kohr-h:
Alright, we seem to be back in the same boat again. So this time, we follow more a bottom-up approach from very thin wrapping to the final nice top-level object. Sure that last part is a bit theoretical, we need a PET geometry first (and one for SPECT while we're at it). Just that we keep that in mind and don't stop with the `StirBlablablabla` classes.

kohr-h:
Are we ready to pull yet? I don't see a reason to have this hanging around in a branch with limited usefulness.

adler-j:
Nope, still needs to do the renaming properly.

adler-j:
Getting this fixed for a final pull later today.

adler-j:
Did some changes here, should be improved in structure. Sadly failed to build on my elekta computer:

https://github.com/UCL/STIR/issues/6

Ill, try to run it tomorrow. Is the new "structure" better?

kohr-h:
I'll have a proper look at it tomorrow.

adler-j:
Added some further improvements now, looking quite ok I'd say.

examples/tomo/stir_project.py
```diff
@@ -0,0 +1,64 @@
...
+volume_file = str(pth.join(base, 'initial.hv'))
+volume = stir.FloatVoxelsOnCartesianGrid.read_from_file(volume_file)
+
+projection_file = str(pth.join(base, 'small.hs'))
+proj_data_in = stir.ProjData.read_from_file(projection_file)
+proj_data = stir.ProjDataInMemory(proj_data_in.get_exam_info(),
+                                  proj_data_in.get_proj_data_info())
+
+# Create ODL spaces
+recon_sp = odl.uniform_discr(odl.FunctionSpace(odl.Cuboid([0, 0, 0],
```
kohr-h: Is it possible to extract this information from the interfile? I guess the volume file just defines the grid? Not necessary to do that immediately, but that would be the next logical step.

odl/tomo/backends/stir_bindings.py
```diff
@@ -0,0 +1,293 @@
...
+        Parameters
+        ----------
+        dom : `DiscreteLp`
+            Volume of the projection. Needs to have the same shape as
+            ``volume.shape()``.
+        ran : `DiscreteLp`
+            Projection space. Needs to have the same shape as
+            ``proj_data.to_array().shape()``.
+        volume : `stir.FloatVoxelsOnCartesianGrid`
```
kohr-h: Is this an actual array with grid info in memory or just an empty template to set up the grid? `volume` suggests a proper data structure (same with `proj_data`).

kohr-h:
I think we can pull this now and take it from here in new feature branches or make small adjustments on master.

BTW, for the long lines due to long class names of STIR, you don't have to stick your code to the right side of the text area. You can avoid that by (in accordance with PEP8) breaking the line immediately after the opening parenthesis and starting the next line with an indentation of only 4 - check the example of [hanging indents](https://www.python.org/dev/peps/pep-0008/#indentation)  in the PEP8 text.

odl/tomo/backends/stir_bindings.py
adler-j: Actual data structure that we use as temporary and write to.

examples/tomo/stir_project.py
adler-j: Yes, see the wrapper function `stir_projector_from_file`, wanted to keep this short.

Regarding the range, it is much harder. I'll need to get Kris to get us a description of how the space actually looks, with weight and all.

PR #112, opened 2015-12-01
Issue 109  grad div

moosmann:
Implementation of spatial gradient and divergence operator.

To do:
* Convert helper function `discrete_part_deriv` into an operator `DiscretePartDeriv`
* Implement forward and backward differences on the interior. So far only central differences can be used.


odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    except IndexError:
+        raise IndexError("Axis paramater ({0}) exceeds number of dimensions "
+                         "({1}).".format(axis, ndim))
+
+    out = np.empty_like(f_data)
+
+    # Numerical differentiation: 2nd order interior
+    slice1[axis] = slice(1, -1)
+    slice2[axis] = slice(2, None)
+    # noinspection PyTypeChecker
```
adler-j: We shouldnt riddle our code with these, it's not used by all tools and makes it harder to read. Imo, remove them.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    ...    print(e)
+    Axis paramater (2) exceeds number of dimensions (2).
+    """
+
+    f_data = np.asanyarray(f)
+    ndim = f_data.ndim
+
+    # create slice objects --- initially all are [:, :, ..., :]
+    # noinspection PyTypeChecker
+    slice1 = [slice(None)] * ndim
```
adler-j: These should have better names `slice_out` or something.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    # noinspection PyTypeChecker
+    slice2 = [slice(None)] * ndim
+    # noinspection PyTypeChecker
+    slice3 = [slice(None)] * ndim
+    # noinspection PyTypeChecker
+    slice4 = [slice(None)] * ndim
+
+    try:
```
adler-j: This can be done without a try catch

```python
if not -f_data.ndim <= axis < f_data.ndim:
    raise IndexError("Axis paramater ({0}) exceeds number of dimensions "
                     "({1}).".format(axis, ndim))
    
if f_data.shape[axis] < 2:
    raise ValueError("Shape of array too small to calculate a "
                     "numerical gradient, at least two elements are "
                     "required.")
    
```

adler-j: Another comment is that this test should be done before you create the slices `slice1`, ...

Test as early as possible.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    # one-side differences
+    else:
+        # Numerical differentiation: 1st order edges
+        if f_data.shape[axis] == 2 or edge_order == 1:
+
+            slice1[axis] = 0
+            slice2[axis] = 1
+            slice3[axis] = 0
+            # 1D equivalent -- out[0] = (y[1] - y[0])
+            out[slice1] = (f_data[slice2] - f_data[slice3])
```
adler-j: np.minus(f_data[slice2], f_data[slice3], out[slice1])

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    out[slice1] = (f_data[slice2] - f_data[slice3]) / 2.0
+
+    # central differences
+    if zero_padding:
+        # Assume zeros for indices outside the volume
+
+        # 1D equivalent -- out[0] = (y[1] - 0)/2.0
+        slice1[axis] = 0
+        slice2[axis] = 1
+        out[slice1] = f_data[slice2] / 2.0
```
adler-j: `np.divide(f_data[slice2], 2.0, out[slice1])`

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+
+        # 1D equivalent -- out[-1] = (0 - y[-2])/2.0
+        slice1[axis] = -1
+        slice3[axis] = -2
+        out[slice1] = - f_data[slice3] / 2.0
```
adler-j: `np.divide(f_data[slice2], -2.0, out[slice1])`

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+            slice2[axis] = 1
+            slice3[axis] = 0
+            # 1D equivalent -- out[0] = (y[1] - y[0])
+            out[slice1] = (f_data[slice2] - f_data[slice3])
+
+            slice1[axis] = -1
+            slice2[axis] = -1
+            slice3[axis] = -2
+            # 1D equivalent -- out[-1] = (y[-1] - y[-2])
+            out[slice1] = (f_data[slice2] - f_data[slice3])
```
adler-j: np.minus(f_data[slice2], f_data[slice3], out[slice1])

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+
+    out = np.empty_like(f_data)
+
+    # Numerical differentiation: 2nd order interior
+    slice1[axis] = slice(1, -1)
+    slice2[axis] = slice(2, None)
+    # noinspection PyTypeChecker
+    slice3[axis] = slice(None, -2)
+    # 1D equivalent -- out[1:-1] = (y[2:] - y[:-2])/2.0
+    out[slice1] = (f_data[slice2] - f_data[slice3]) / 2.0
```
adler-j:
```python
np.minus(f_data[slice2], f_data[slice3], out[slice1])
out[slice1] /= 2.0
```

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+         An N-dimensional array
+    axis : `int`, optional
+        The axis along which the partial derivative is evaluated. Default: 0
+    dx : `float`, optional
+        Scalars specifying the sample distances in each dimension `axis`.
+        Default distance: 1.
+    edge_order : {1, 2}, optional
+        Partial derivative is calculated using Nth order accurate
+        differences at the boundaries. Default edge order: 2
+    zero_padding : `bool`, optional
```
adler-j: Should document and throw error if user tries using both zero_padding and edge_order.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
+# Copyright 2014, 2015 Jonas Adler
```
adler-j: Says jonas adler, even if im honored, im quite sure we should change this.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    Assuming (implicit) zero padding central differences are used on the
+    interior and on endpoints. Otherwise one-sided differences differences
+    are used. Then first-order accuracy can be triggered on endpoints with
+    parameter `edge_order`.
+
+    The returned array has the same shape as the input array `f`.
+
+    Parameters
+    ----------
+    f : `numpy.ndarray`
```
adler-j: Since you do a cast, shouldnt array-like be ok?

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+
+    # divide by step size
+    out /= dx
+
+    return out
+
+
+# noinspection PyAbstractClass
+class DiscreteGradient(Operator):
+    """Gradient operator for any number of dimensions `n`.
```
adler-j: `n` is a link with no target. Should be ``n``

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        """Initialize a `DiscreteGradient` operator instance.
+
+        Zero padding is required for the negative `DiscreteDivergence`
+        operator to match the adjoint of the `DiscreteGradient` operator.
+
+        Parameters
+        ----------
+        space : `FunctionSpace`
+            The space of elements which the operator is acting on
+        voxel_size : `n`-tuple of `floats`
```
adler-j: See comment on `n` above.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+class DiscreteGradient(Operator):
+    """Gradient operator for any number of dimensions `n`.
+
+    Calls helper function `discrete_part_deriv` to calculate each component
+    of the resulting product space vector. For the adjoint of the `Gradient`
+    operator to match the negative `DiscreteDivergence` operator requires
+    `zero_padding`.
+    """
+
+    def __init__(self, space, voxel_size=(1,), edge_order=2,
```
adler-j: voxel_size is a 1-tuple, doc below says n-tuple

odl/operator/fspace_ops.py
adler-j: also, optional

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+            `n`-tuple of scalars specifying the sample distances in each
+            dimension. Default distance is 1 in each direction
+        edge_order : {1, 2}, optional
+            Partial derivative is calculated using Nth order accurate
+            differences at the boundaries. Default edge order: 2
+        zero_padding : `bool`, optional
+            Implicit zero padding. Assumes values outside of `space` to be
+            zero. Default: True
+        """
+        self.voxel_size = voxel_size
```
adler-j: We should verify some thins here, type of space, length of voxel_size, edge order = 1 or 2

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+
+        Parameters
+        ----------
+        rhs : ``domain`` element
+            Input vector to which the `DiscreteGradient` operator is applied to
+        out : ``range`` element
+            Output vector to which the result is written
+
+        Examples
+        --------
```
adler-j: This example is i.m.o too big. Perhaps split it? Also move adjoint example to adjoint,

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        ...         vol = vol * vol.reshape(s)
+        ...     return vol
+        >>> ndim = 3
+        >>> vsize = 100
+        >>> disc = uniform_discr(FunctionSpace(IntervalProd(
+        ... [0.]*ndim, [vsize]*ndim)), [vsize]*ndim)
+        >>> f = disc.element(ndvolume(vsize, ndim, np.int32))
+        >>> A = DiscreteGradient(disc, (1.,)*ndim, zero_padding=True)
+        >>> Af = A(f)
+        >>> g = A.range.one()
```
adler-j: Gradient of a constant is not so interesting. How about lower dimension and a explicit example?

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+    `DiscreteDivergence` operator to match the negative `DiscreteGradient`
+    operator requires `zero_padding`.
+    """
+
+    def __init__(self, space, voxel_size=(1,), edge_order=2,
+                 zero_padding=True):
+        """Initialize a `DiscreteDivergence` operator instance.
+
+        Zero padding is required for the negative `DiscreteGradient`
+        operator to match the adjoint of the `DiscreteDivergence` operator.
```
adler-j: Does this mean if you create a DiscreteGradient op with zero_padding=False, the adjoint will be wrong?

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        ...     return vol
+        >>> ndim = 4
+        >>> vsize = 6
+        >>> disc = uniform_discr(FunctionSpace(IntervalProd(
+        ... [0.]*ndim, [vsize]*ndim)), [vsize]*ndim)
+        >>> f = disc.element(ndvolume(vsize, ndim, np.int32))
+        >>> B = DiscreteDivergence(disc, (1.,)*ndim, zero_padding=True)
+        >>> g = B.domain.one()
+        >>> Bg = B(g)
+        >>> Bdf = B.adjoint(f)
```
adler-j: Doctests should show what it does, Id prefer actual numbers showing. This is more fit for the actual tests.

odl/operator/operator.py
```diff
@@ -640,6 +640,9 @@ def __str__(self):
     # vector properly
     __array_priority__ = 2000000.0
 
+    def __neg__(self):
```
adler-j: This already exists [here on master](https://github.com/odlgroup/odl/blob/04010e33eb87e133c93716aff67b5bcc15e00d9b/odl/operator/operator.py#L608)

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        >>> f.inner(Bg) - g.inner(Bdf)
+        0.0
+        """
+
+        tmp = np.zeros_like(rhs[0].asarray())
```
adler-j: zeros_like(out.asarray())

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        >>> f.inner(Bg) - g.inner(Bdf)
+        0.0
+
+        """
+        rhs_data = np.asanyarray(rhs)
+        ndim = rhs_data.ndim
+
+        dx = self.voxel_size
+        if np.size(dx) == 1:
+            dx = [dx for _ in range(ndim)]
```
adler-j: [dx] * ndim

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        >>> g.inner(Af) - f.inner(Adg)
+        0.0
+        >>> B = DiscreteDivergence(disc, (1.,)*ndim, zero_padding=True)
+        >>> Bg = B(g)
+        >>> Bdf = B.adjoint(f)
+        >>> f.inner(Bg) - g.inner(Bdf)
+        0.0
+
+        """
+        rhs_data = np.asanyarray(rhs)
```
adler-j: Any reason for asanyarray istead of rhs.asarray?

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+        """
+        self.voxel_size = voxel_size
+        self.edge_order = edge_order
+        self.zero_padding = zero_padding
+        super().__init__(domain=space,
+                         range=ProductSpace(space, len(voxel_size)),
+                         linear=True)
+
+    def _apply(self, rhs, out):
+        """Apply gradient operator to `rhs` and store result in `out`.
```
adler-j: \`rhs\` and \`out\` are dead links.

adler-j:
Very nice functionality, gonna be really useful, added a few comments.

adler-j:
Also: consider rebasing to remove many of the not so informative commits (merges, refactors).

odl/operator/fspace_ops.py
kohr-h: Definitely! So you don't have a chance to run away with the code

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Default operators defined on any `FunctionSpace`."""
```
kohr-h: Here it should obviously say something else.

odl/operator/fspace_ops.py
```diff
@@ -0,0 +1,412 @@
...
+
+# ODL imports
+from odl.operator.operator import Operator
+from odl.set.pspace import ProductSpace
+
+
+__all__ = ('DiscreteGradient', 'DiscreteDivergence')
+
+
+def discrete_part_deriv(f, axis=0, dx=1.0, edge_order=2,
```
kohr-h: This should be an operator IMO. And as such, it should not be defined on NumPy arrays immediately but rather for `DiscreteLp` vectors. If we really need a pure array version (name TBD, not `discrete_part_deriv`, rather `divided_diffs` or `finite_diff`) that one could be a helper function similar to this guy, which is then used internally by the operator.

adler-j: Not quite sure I agree. Having this as a helper makes the code easier to read. Perhaps simply make it private? `_discrete_part_deriv`?

adler-j: Second thought. Yes we should have an operator, like "PartialDerivative". I dont see the shame in keeping this still though, it is more efficient to implement GradientOperator using this than using repeated PartialDerivative.

kohr-h: There's no contradiction in what you say. I'm arguing for an operator with this function as a helper for implementation. If it's useful in its own right, we don't need to make it private.

odl/operator/fspace_ops.py
kohr-h: And we should scrap all occurrences of `rhs` where there is no equation which can have a right hand side. If it's an argument to an operator, it's just `x`, at least in 90 % in the code now.

odl/operator/fspace_ops.py
adler-j: Agree.

kohr-h: Okay, so agreement. When it comes to naming, I think we should make a difference between genuinely **discrete** concepts and **discretized continuous** concepts, i.e. **divided/finite differences** vs. **discrete partial derivative**. That's why I'm arguing for a `DiscretePartDeriv` operator and a `finite_diff` helper function.

odl/operator/fspace_ops.py
adler-j: Muhahahaha. I'll hold the code ransom for

https://youtu.be/jTmXHvGZiSY?t=33

odl/operator/fspace_ops.py
kohr-h: A fundamental difference here would e.g. be the way the boundary is handled. The helper `finite_diff` can implement all flexibility there is, while the operator has to be consistent with the choice of discretization.

odl/operator/fspace_ops.py
adler-j: Yeah I did a huge renaming of everything to `x` a while ago. Primarily because this matches numpy usage.

odl/operator/fspace_ops.py
moosmann: I suppose you mean `np.subtract`. Is this more efficient than a `-`?

odl/operator/fspace_ops.py
moosmann: Yeah, I know. For the moment I was too lazy to create tests.I just wanted to get it quickly on github since Kati wants to use the gradient operator for TV regulariztation. But I'll do it.

odl/operator/fspace_ops.py
kohr-h: Yes, it writes directly to `out` instead of allocating a new `out`.

adler-j: Yeah, its `np.subtract`. Using the `out` parameter you dont need to make a copy. Both faster and more memory efficient.

Your current code is:

```python
tmp = f_data[slice2] - f_data[slice3]
out[slice1] = tmp
```

odl/operator/fspace_ops.py
moosmann: `disc` is not constant, or did I miss something here?

odl/operator/fspace_ops.py
adler-j: No problem!

Some of these comments arnt "hold the pull" worthy, more like, we should fix this at some point.

odl/operator/fspace_ops.py
adler-j: No, but you take the adjoint of a one, which should be zero?

Id prefer a example in one dimension that is very simple to follow. These types of examples should go into the test suite.

odl/operator/fspace_ops.py
moosmann: Yes, without zero padding the surface term after partial integration won't be zero any more and has to be handled properly.

odl/operator/fspace_ops.py
moosmann: Not really. That what used in the original numpy code for the gradient. I'll change it.

odl/operator/fspace_ops.py
adler-j: In that case I think we should raise `NotImplementedError` in the adjoint aswell? Any chance we can solve this later?

odl/operator/fspace_ops.py
moosmann: The code is already a bit older. I'll change it.

odl/operator/fspace_ops.py
kohr-h: This operator should act on `DiscreteLp` functions (maybe this file needs to be called `discr_ops` then?), and the chosen discretization should determine how the boundary is handled. Currently, it always assumes zero boundary. Basically all the options in `__init__()` should go away and be handled internally.

odl/operator/fspace_ops.py
adler-j: Try to keep branches short lived for this reason, diverging branches cause lots of trouble in the long run.

moosmann:
Thanks for the comments. I gonna implement them.

kohr-h:
Not to forget: this is good work and nice to have, it will be very useful. It just needs to get up to speed.

I'm going to do a pen and paper exercise and come back with instructions on how to handle the formulas.

adler-j:
Agree with holger, this will likely be a backbone of most solvers we use.

kohr-h:
I crunched the expressions, and now I have a mapping from the discretizations to the endpoint formulas, together with a couple of new discretizations we need to handle :-)

I'll write it up properly until tomorrow.

moosmann:
Implemented most comments. Please check if further changes are necessary, especially regarding the first bullet point below.

Still to do:
* optimize `_apply` method of `DivergenceOperator`
* implement forward / backward differences
* implement other boundary conditions including adjoint
* implement adjoint for `DiscretePartDeriv`

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+    zero_padding : `bool`, optional
+        Implicit zero padding. Assumes values outside the domain of `f` to be
+        zero. Default: False
+
+    Returns
+    -------
+    out : `numpy.ndarray`
+        Returns an N-dimensional array of the same shape as `f`
+
+    Examples
```
adler-j: Very nice examples!

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+        First-order accurate differences (1) can be used at the boundaries
+        if no zero padding is used. Default edge order: 2
+    zero_padding : `bool`, optional
+        Implicit zero padding. Assumes values outside the domain of `f` to be
+        zero. Default: False
+
+    Returns
+    -------
+    out : `numpy.ndarray`
+        Returns an N-dimensional array of the same shape as `f`
```
adler-j: \`f\` is a dead link. Try building the doc so you find all of these.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+                            "shape of input array `f` ({1})".format(out.shape,
+                                                                    f.shape))
+
+    if edge_order not in {1, 2, None}:
+        raise ValueError("edge order ({0}) not valid".format(edge_order))
+
+    if edge_order is None:
+        edge_order = 2
+
+    dx = float(dx)
```
adler-j: Complex arrays?

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+            slice_node2[axis] = 0
+            # 1D equivalent: out[0] = (f[1] - f[0])
+            np.subtract(f_data[slice_node1], f_data[slice_node2],
+                        out[slice_out])
+
+            slice_out[axis] = -1
+            slice_node1[axis] = -1
+            slice_node2[axis] = -2
+            # 1D equivalent: out[-1] = (f[-1] - f[-2])
+            np.subtract(f_data[slice_node1], f_data[slice_node2],
```
adler-j: We need tests to make sure that subtract works properly here.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+            slice_out[axis] = -1
+            slice_node1[axis] = -1
+            slice_node2[axis] = -2
+            slice_node3[axis] = -3
+            # 1D equivalent: out[-1] = (3*f[-1] - 4*f[-2] + f[-3]) / 2.0
+            out[slice_out] = (3.0 * f_data[slice_node1] - 4.0 * f_data[
+                slice_node2] + f_data[slice_node3]) / 2.0
+
+    # divide by step size
+    out /= dx
```
adler-j: Possibly check `dx != 0`?

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+
+
+class DiscretePartDeriv(Operator):
+    """Calculate the discrete partial derivative along a given axis.
+
+        Calls helper function `finite_diff` to calculate finite difference.
+        Preserves the shape of the underlying grid.
+    """
+    # TODO: implement adjoint
+    # TODO: better name for `dx`
```
adler-j: I dont think `dx` is that bad, everyone understands what you mean instantly, which is the most important.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+        >>> disc = uniform_discr(space, data.shape)
+        >>> par_div = DiscretePartDeriv(disc)
+        >>> f = par_div.domain.element(data)
+        >>> par_div_f = par_div(f)
+        >>> print(par_div_f)
+        [[0.0, 1.0, 2.0, 3.0, 4.0],
+         [0.0, 1.0, 2.0, 3.0, 4.0]]
+        """
+
+        finite_diff(x.asarray(), out.asarray(), axis=self.axis, dx=self.dx,
```
adler-j: out.asarray() only works if the underlying space is `Rn`, if it is a `CudaRn`, `asarray` does not return a `view`. We may need some way to check this.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+        >>> g = grad.range.element((data, data ** 2))
+        >>> adj = grad.adjoint
+        >>> adj_g = adj(g)
+        >>> print(adj_g)
+        [[-2.5, -11.0, -22.0, -33.0, 18.5],
+         [-10.0, -39.5, -79.0, -118.5, 92.0]]
+        >>> g.inner(grad_f) - f.inner(adj_g)
+        0.0
+        """
+        x_data = np.asarray(x)
```
adler-j: better to do `x.asarray()`.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+        ----------
+        x : ``domain`` element
+            `ProductSpaceVector` to which the divergence operator is applied to
+        out : ``range`` element
+            Output vector to which the result is written
+
+        Examples
+        --------
+        >>> from odl import Rectangle, uniform_discr, FunctionSpace
+        >>> data = np.arange(5, dtype=float) * np.arange(1, 3).reshape(2,1)
```
adler-j: Prefer explicit data (even a 3x3 example is fine)

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+        -3248.4
+        >>> f.inner(adj_g)
+        -3248.4
+        """
+
+        ndim = self.range.grid.ndim
+        dx = self.range.grid.stride
+
+        # TODO: can the following code be optimized?
+        tmp = np.zeros_like(out.asarray())
```
adler-j: Without calling `asarray()`
```python
tmp = np.zeros(out.shape, out.dtype, order=out.order)
```

I'd write this:

```python
arr = out.asarray()
tmp = np.zeros(out.shape, out.dtype, order=out.order)
for axis in range(ndim):
    arr += finite_diff(x[axis], out=tmp, axis=axis, dx=dx[axis],
                       edge_order=2, zero_padding=True)
out[:] = arr # self assignment in case asarray is a view, thus no overhead.
```

This is a pattern you could use in general.

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+                                    DiscreteGradient, DiscreteDivergence)
+from odl.space.ntuples import Rn
+from odl.set.domain import Rectangle
+
+
+def test_finite_diff():
+    """Finite differences test. """
+
+    ndim = 10
+    f = np.arange(ndim, dtype=float)
```
adler-j: I.m.o using functions such as these does not help, simply doing a very simple `f = [0, 1, 2, 3]` should do splendidly.

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+
+    df = finite_diff(f)
+    assert all(df == np.ones(ndim))
```
adler-j: `assert all(df == 1)`

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+    assert df1[-1] is not df2[-1]
+    with pytest.raises(ValueError):
+        finite_diff(f, edge_order=3)
+
+    # axis parameter
+    ndim = 5
+    f = np.arange(ndim, dtype=float)
+    f = f * f.reshape((ndim, 1))
+    df = finite_diff(f, axis=0)
+    f0 = np.array([[-0., 1., 2., 3., 4.],
```
adler-j: -0?

odl/operator/discr_ops.py
moosmann: How to do that most efficiently? `sphinx-build ...`?

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Adjoint operator
+    ran_vec = div.range.element(ndvolume(vsize, ndim, np.float32)**2)
+    adj = div.adjoint
+    adj_vec = adj(ran_vec)
+    lhs = ran_vec.inner(div_vec)
+    rhs = dom_vec.inner(adj_vec)
+    assert not lhs == 0
+    assert not rhs == 0
+    assert lhs == rhs
```
adler-j: Exact float equality is a harsh condition, maybe `almost_equal`?

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+    vsize = 3
+    intvl = IntervalProd([0.] * ndim, [vsize] * ndim)
+    space = FunctionSpace(intvl)
+    discr_space = uniform_discr(space, [vsize] * ndim)
+    # dom = ProductSpace(discr_space, ndim)
+    # dom_vec0 = dom.element([ndvolume(vsize, ndim, np.float32)] * ndim)
+
+    # Divergence
+    div = DiscreteDivergence(discr_space)
+    dom_vec = div.domain.element([ndvolume(vsize, ndim, np.float32)] * ndim)
```
adler-j: Why float32?

odl/operator/discr_ops.py
adler-j: Simply `make html` in the doc folder. You may need to run the `generate_doc.py` script in doc/source first to populate the files etc.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,446 @@
...
+    else:
+        if not out.shape == f.shape:
+            raise TypeError("shape of `out` array ({0}) does not match the "
+                            "shape of input array `f` ({1})".format(out.shape,
+                                                                    f.shape))
+
+    if edge_order not in {1, 2, None}:
+        raise ValueError("edge order ({0}) not valid".format(edge_order))
+
+    if edge_order is None:
```
adler-j: Why not simply `edge_order=2` in the signature?

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+    par_div_f1 = par_div(f)
+
+    # operator
+    par_div = DiscretePartDeriv(discr_space, axis=1, dx=0.2, edge_order=2,
+                                zero_padding=True)
+
+    # partial derivative
+    par_div_f2 = par_div(f)
+
+    assert par_div_f1 is not par_div_f2
```
adler-j: I dont get this test, do you mean `==`?

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # at lease two elements are required
+    with pytest.raises(ValueError):
+        finite_diff(np.array([0.0]))
+
+    # in-place evaluation
+    ndim = 10
+    f = np.arange(ndim, dtype=float)
+    out = np.zeros_like(f)
+    assert out is not finite_diff(f)
```
adler-j: There is no way this assert could fail, `finite_diff` doesn't know about `out`, so Id say this is redundant.

test/operator/discr_ops_test.py
```diff
@@ -0,0 +1,242 @@
...
+    assert out is finite_diff(f, out)
+    assert all(out == finite_diff(f))
+    assert out is not finite_diff(f)
+    out = np.zeros(2)
+    with pytest.raises(TypeError):
```
adler-j: If the error is that `out` is too small, I'd raise a ValueError.

odl/operator/discr_ops.py
moosmann: What to do in the case of a `CudaRn`? 
```
        if isinstance(self.domain.dspace, Rn):
            finite_diff(x.asarray(), out.asarray(), axis=self.axis,
                        dx=self.dx, edge_order=self.edge_order,
                        zero_padding=self.zero_padding)
        elif isinstance(self.domain.dspace, Rn):
            ??
```
I cannot test CUDA cases since installation of odlpp already fails when trying to update the submodules:
```
jmoosmann@moosmann:~/git/LCR/odl$ git submodule update --init --recursive
fatal: reference is not a tree: 7eb19f623f0eaf6be0e953b2e0bc826564fcef4c
Unable to checkout '7eb19f623f0eaf6be0e953b2e0bc826564fcef4c' in submodule path 'doc/sphinxext'
```
Shall I make an issue in the odlpp repo?

odl/operator/discr_ops.py
moosmann: Good question. Probably a relict. Changed.

test/operator/discr_ops_test.py
moosmann: You're right, should have been `==` to test that `axis` argument works.

test/operator/discr_ops_test.py
moosmann: No reason. Due to copy and paste.

adler-j:
Looking very good now, certainly pullable imo.

odl/operator/discr_ops.py
moosmann: Thanks. I will check this in the future.

odl/operator/discr_ops.py
adler-j: We can have the CudaRn discussion in person.

Regarding second fail, have you tried going into sphinxext and `git checkout master` `git pull`?

If your CUDA installation fails this is something we need to fix ASAP. Make a proper issue in the odlpp repo.

odl/operator/discr_ops.py
```diff
@@ -0,0 +1,453 @@
...
+        [[0.0, 1.0, 2.0, 3.0, 4.0],
+         [0.0, 1.0, 2.0, 3.0, 4.0]]
+        """
+
+        if isinstance(self.domain.dspace, Rn):
+            finite_diff(x.asarray(), out.asarray(), axis=self.axis,
+                        dx=self.dx, edge_order=self.edge_order,
+                        zero_padding=self.zero_padding)
+        elif isinstance(self.domain.dspace, CudaRn):
+            pass
```
adler-j: At least raise `NotImplementedError` :)

Easiest would be to implement 

```python
out[:] = finite_diff(x.asarray(), axis=self.axis,
                     dx=self.dx, edge_order=self.edge_order,
                     zero_padding=self.zero_padding)
```

moosmann:
Implemented final changes, rebased, rewrote commit history, checks passed. I think it is now ready to merge.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+                         "gradient, at least two elements are required.")
+    if out is None:
+        out = np.empty_like(f_data)
+    else:
+        if not out.shape == f.shape:
+            raise ValueError(
+                "shape of `out` array ({0}) does not match the shape of "
+                "input array `f` ({1}).".format(out.shape, f.shape))
+
+    if edge_order not in {1, 2}:
```
adler-j: `{1, 2}` creates a `set`, `[1, 2]` is standard.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        raise ValueError("edge order ({0}) not valid".format(edge_order))
+
+    if dx == 0:
```
adler-j: Do we allow negative steps?

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+
+# TODO: CUDA support
+
+class DiscretePartDeriv(Operator):
+    """Calculate the discrete partial derivative along a given axis.
+
+    Calls helper function `finite_diff` to calculate finite difference.
+    Preserves the shape of the underlying grid.
+    """
+    # TODO: implement adjoint
```
adler-j: Make a new issue for this, it will be needed.

odl/discr/discr_ops.py
moosmann: Fixed.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        out : ``range`` element
+            Output vector to which the result is written
+
+        Examples
+        --------
+        >>> from odl import uniform_discr_fromspace, FunctionSpace, Rectangle
+        >>> data = np.array([[ 0.,  1.,  2.,  3.,  4.],
+        ...                  [ 0.,  2.,  4.,  6.,  8.]])
+        >>> space = FunctionSpace(Rectangle([0, 0], [2, 1]))
+        >>> disc = uniform_discr_fromspace(space, data.shape)
```
adler-j: ~~space = FunctionSpace(Rectangle([0, 0], [2, 1]))~~
~~disc = uniform_discr_fromspace(space, data.shape)~~

`disc = uniform_discr([0, 0], [2, 1], data.shape`

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        """Apply gradient operator to ``x`` and store result in ``out``.
+
+        Parameters
+        ----------
+        x : ``domain`` element
+            Input vector to which the operator is applied to
+        out : ``range`` element
+            Output vector to which the result is written
+
+        Examples
```
adler-j: Nice example!

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+                self.domain.dspace))
+
+    @property
+    def adjoint(self):
+        """Return the adjoint."""
+        raise NotImplementedError('adjoint not implemented')
+
+
+class DiscreteGradient(Operator):
+    """Spatial gradient operator for `DiscreteLp` spaces with any number of
```
adler-j: Top doc should fit inside one row:

`Spatial gradient operator for `DiscreteLp` spaces.`

or similar.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        out : ``range`` element
+            Output vector to which the result is written
+
+        Examples
+        --------
+        >>> from odl import uniform_discr_fromspace, FunctionSpace, Rectangle
+        >>> data = np.array([[ 0.,  1.,  2.,  3.,  4.],
+        ...                  [ 0.,  2.,  4.,  6., 8.]])
+        >>> space = FunctionSpace(Rectangle([0,0], [2,1]))
+        >>> disc = uniform_discr_fromspace(space, data.shape)
```
adler-j: `uniform_discr` as above.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        >>> f = disc.element(data)
+        >>> grad = DiscreteGradient(disc)
+        >>> grad_f = grad(f)
+        >>> print(grad_f[0])
+        [[0.0, 1.0, 2.0, 3.0, 4.0],
+         [-0.0, -0.5, -1.0, -1.5, -2.0]]
+        >>> print(grad_f[0])
```
adler-j: Exactly the same as row above?

odl/discr/discr_ops.py
moosmann: I interpret it as step length, thus no. Changed it o `dx <= 0`

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        [[0.0, 1.0, 2.0, 3.0, 4.0],
+         [-0.0, -0.5, -1.0, -1.5, -2.0]]
+        >>> g = grad.range.element((data, data ** 2))
+        >>> adj = grad.adjoint
+        >>> adj_g = adj(g)
```
adler-j: Simply `grad.adjoint(g)`

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        elif isinstance(self.domain.dspace, CudaRn):
+            raise NotImplementedError('support for `CudaRn` to be '
+                                      'implemented.')
+        else:
+            raise TypeError('invalid data space {}.'.format(
+                self.domain.dspace))
+
+    @property
+    def adjoint(self):
+        """Return the adjoint operator given by the negative of the
```
adler-j: Top doc string should fit in one row.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+
+        Note that the ``space`` argument of the `DiscreteDivergence`
+        operator is not the range but the domain of the `DiscreteGradient`
+        operator.
+        """
+        return - DiscreteDivergence(self.domain)
+
+
+class DiscreteDivergence(Operator):
+    """Divergence operator for `DiscreteLp` spaces with any number of
```
adler-j: Should fit in one row.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+            Output vector to which the result is written
+
+        Examples
+        --------
+        >>> from odl import Rectangle, uniform_discr_fromspace, FunctionSpace
+        >>> data = np.array([[0., 1., 2., 3., 4.],
+        ...                  [1., 2., 3., 4., 5.],
+        ...                  [2., 3., 4., 5., 6.]])
+        >>> space = FunctionSpace(Rectangle([0, 0], [3, 5]))
+        >>> disc = uniform_discr_fromspace(space, data.shape)
```
adler-j: `uniform_discr`

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        >>> div = DiscreteDivergence(disc)
+        >>> f = div.domain.element([data, data])
+        >>> div_f = div(f)
+        >>> print(div_f)
+        [[1.0, 2.0, 2.5, 3.0, 1.0],
+         [2.0, 2.0, 2.0, 2.0, -1.0],
+         [1.0, 0.0, -0.5, -1.0, -5.0]]
+        >>> g = div.range.element(data ** 2)
+        >>> adj = div.adjoint
+        >>> adj_g = adj(g)
```
adler-j: `div.adjoint(g)`

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+                if axis == 0:
+                    arr[:] = tmp
+                else:
+                    arr += tmp
+
+            # self assignment in case asarray is a view, thus no overhead
+            out[:] = arr
+
+        elif isinstance(self.domain.dspace, CudaRn):
+            raise NotImplementedError('support for `CudaRn` to be '
```
adler-j: This should work, how about write a TODO: write tests, and create an issue on this.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,477 @@
...
+                                      'implemented.')
+        else:
+            raise TypeError('invalid data space {}.'.format(
+                self.domain.dspace))
+
+    @property
+    def adjoint(self):
+        """Return the adjoint operator given by the negative of the
```
adler-j: Top doc one row.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    assert out is not finite_diff(f)
+
+    # axis
+    f = np.array([[0., 1., 2., 3., 4.],
+                  [1., 2., 3., 4., 5.]])
+    df0 = finite_diff(f, axis=0)
+    df1 = finite_diff(f, axis=1)
+    assert all_equal(df0, df1)
+
+    # complex arrays
```
adler-j: Nice :)

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    with pytest.raises(TypeError):
+        DiscretePartDeriv(discr_space)
+
+    # phantom data
+    data = np.array([[1.2, 0, 3, 5, 7],
+                     [4, -1, 3, -2.1, -4]])
+
+    # discretized space
+    space = FunctionSpace(Rectangle([0, 0], [2, 1]))
+    discr_space = uniform_discr_fromspace(space, data.shape)
```
adler-j: uniform_discr

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    par_div_f1 = par_div(f)
+
+    # operator
+    par_div = DiscretePartDeriv(discr_space, axis=1, dx=0.2, edge_order=2,
+                                zero_padding=True)
+
+    # partial derivative
+    par_div_f2 = par_div(f)
+
+    assert not par_div_f1 == par_div_f2
```
adler-j: `assert par_div_f1 != par_div_f2` should work?

odl/discr/discr_ops.py
moosmann: Changed in all doctests.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    discr_space = Rn(1)
+    with pytest.raises(TypeError):
+        DiscreteGradient(discr_space)
+
+    for ndim in range(1, 6):
+        # DiscreteLp Vector
+        vsize = 3
+        intvl = IntervalProd([0.] * ndim, [vsize] * ndim)
+        space = FunctionSpace(intvl)
+        discr_space = uniform_discr_fromspace(space, [vsize] * ndim)
```
adler-j: `uniform_discr`

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+        # Gradient
+        grad = DiscreteGradient(discr_space)
+        grad(dom_vec)
+
+    # DiscreteLp Vector
+    ndim = 2
+    vsize = 3
+    intvl = IntervalProd([0.] * ndim, [vsize] * ndim)
+    space = FunctionSpace(intvl)
+    discr_space = uniform_discr_fromspace(space, [vsize] * ndim)
```
adler-j: `uniform_discr`

odl/discr/discr_ops.py
moosmann: Done.

odl/discr/discr_ops.py
kohr-h: I always use tuples for literal sets, but that's a matter of taste.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    dom_vec = discr_space.element(ndvolume(vsize, ndim))
+
+    # Gradient
+    grad = DiscreteGradient(discr_space)
+    grad_vec = grad(dom_vec)
+    assert len(grad_vec) is ndim
```
adler-j: Do not compare integers using `is`, see [stockoverflow](http://stackoverflow.com/questions/306313/pythons-is-operator-behaves-unexpectedly-with-integers) for more info.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+
+    # Adjoint operator
+    ran_vec = grad.range.element(
+        [ndvolume(vsize, ndim) ** 2] * ndim)
+    adj = grad.adjoint
+    adj_vec = adj(ran_vec)
+    lhs = ran_vec.inner(grad_vec)
+    rhs = dom_vec.inner(adj_vec)
+    assert not lhs == 0
```
adler-j: lhs != 0

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+
+    # Adjoint operator
+    adj_div = div.adjoint
+    ran_vec = div.range.element(data ** 2)
+    adj_div_ran_vec = adj_div(ran_vec)
+
+    # Adjoint condition
+    lhs = ran_vec.inner(div_dom_vec)
+    rhs = dom_vec.inner(adj_div_ran_vec)
+    assert not lhs == 0
```
adler-j: lhs != 0

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    assert not rhs == 0
+    assert almost_equal(lhs, rhs)
+
+    # Higher dimensional arrays
+    for ndim in range(1, 6):
+
+        # DiscreteLp Vector
+        lin_size = 3
+        space = FunctionSpace(IntervalProd([0.] * ndim, [lin_size] * ndim))
+        discr_space = uniform_discr_fromspace(space, [lin_size] * ndim)
```
adler-j: uniform_discr

odl/discr/discr_ops.py
kohr-h: I wouldn't generally disallow it *if* it makes sense. It may be useful, at least in this low-level helper. But that's something which can be done later.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Divergence
+        div = DiscreteDivergence(discr_space)
+        dom_vec = div.domain.element([ndvolume(lin_size, ndim)] * ndim)
+        div(dom_vec)
```
adler-j: We don't do anything with the result?

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    # Apply operator
+    dom_vec = div.domain.element([data, data])
+    div_dom_vec = div(dom_vec)
+
+    # computation of divergence with helper function
+    dx0, dx1 = discr_space.grid.stride
+    df0 = finite_diff(data, axis=0, dx=dx0, zero_padding=True, edge_order=2)
+    df1 = finite_diff(data, axis=1, dx=dx1, zero_padding=True, edge_order=2)
+
+    assert all_equal(df0 + df1, div_dom_vec.asarray())
```
adler-j: `all_equal(df0 + df1, div_dom_vec)` should do?

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,287 @@
...
+    dfe[0] = f[1] - f[0]  # 1st-order accurate forward difference
+    dfe[-1] = f[-1] - f[-2]  # 1st-order accurate backward difference
+    assert all_equal(df, dfe)
+
+    # different edge order really differ
+    df1 = finite_diff(f, edge_order=1)
+    df2 = finite_diff(f, edge_order=2)
+    assert all_equal(df1[1:-1], dfe[1:-1])
+    assert all_equal(df2[1:-1], dfe[1:-1])
+    assert not df1[0] == df2[0]
```
adler-j: df1[0] != df2[0]

odl/discr/discr_ops.py
kohr-h: Please don't use the name `disc` (I guess it's stolen from @adler-j 's examples, we should change it there, too). For me, a disc is something round and gives the totally wrong impression of what is meant.

adler-j: If i use `disc` you are indeed correct, it should be `discr` across the board.

adler-j:
Minor stylistic things, fix those and add new issues for the TODO's. Looking great (one of our better pulls :D). When you are done with these you can pull (imo).

kohr-h:
Agree, this should go into master today. We look into the remaining TODOs and the math stuff in a separate issue.

odl/discr/discr_ops.py
moosmann: Should have been `f[1]`. Fixed.

odl/discr/discr_ops.py
moosmann: I cannot test it, since currently `odlpp` is not working. I'll make an issue.

test/discr/discr_ops_test.py
moosmann: True. Changed everywhere.

test/discr/discr_ops_test.py
moosmann: No, it doesn't. This raises:
```
E           ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

kohr-h: At which line in `testutils.py`? This should have been fixed.

odl/discr/discr_ops.py
moosmann: A forward difference would then become a backward difference for negative `dx`. Wouldn't this be confusing?

test/discr/discr_ops_test.py
adler-j: Indeed seems to be a problem with `testutils`.

adler-j: No it is not actually. Nice catch. It is line 93 in `testutils` and the error is because `df0 + df1` is `2 x n` while `div_dom_vec` is `2n`. We need to try catch the statement `return iter1 == iter2` and return `false` on catch.

kohr-h: I suspected that line, too. And yes, it should return `False` instead of bailing out.

moosmann: Line 102

odl/discr/discr_ops.py
kohr-h: I don't know. Maybe it should reverse the array along the specified axis? Leave it as it is now, if we see a good use in the future, we may think about it again.

test/discr/discr_ops_test.py
moosmann: Well, yes. But it also doesn't raise an error. Some more checks would be nice of course.

test/discr/discr_ops_test.py
kohr-h: There the function calls itself recursively. The last failure should be in line 93, that's the only place where an unfenced comparison by `==` is done.

moosmann: Output from pytest:
```
    def all_equal(iter1, iter2):
        # Sentinel object used to check that both iterators are the same length
        diff_length_sentinel = object()
    
        try:
            if iter1 == iter2:
                return True
        except ValueError:
            pass
    
        if iter1 is None and iter2 is None:
            return True
    
        try:
            i1 = iter(iter1)
            i2 = iter(iter2)
        except TypeError:
            return iter1 == iter2
    
        for [ip1, ip2] in zip_longest(i1, i2,
                                      fillvalue=diff_length_sentinel):
            # Verify that none of the lists has ended (then they are not the
            # same size)
            if ip1 is diff_length_sentinel or ip2 is diff_length_sentinel:
                return False
    
>           if not all_equal(ip1, ip2):
E           ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

odl/util/testutils.py:102: ValueError
```

adler-j: This is due to recursion, I'm like 99% sure it is line 93.

test/discr/discr_ops_test.py
kohr-h: That's a test of the sort "runs at all", which is valid in a `test_<class>_init()` type test, just for functions. 

Speaking of which, I think the test functions do a bit too much at a time. In theory, each test function should test exactly one scenario and one parameter. This is probably a bit too fine-grained, but having small and isolated tests is a big plus when you need to pin down an error source. I suggest that you split the tests at least into 1D and nD and also test different schemes separately. It may be worth making a fixture for that.

adler-j: Imo: get this pulled to master and make an issue "clean up tests in ...", we can all agree this probably raises the quality of our unit-tests already :)

kohr-h: Sure, that's what I meant.

moosmann:
Fixed all comments. However, now the docs don't build any more:
```
jmoosmann@moosmann:~/git/LCR/odl/doc$ make html
sphinx-build -b html -d _build/doctrees   source _build/html
Running Sphinx v1.3.1
making output directory...
loading pickled environment... not yet created
[autosummary] generating autosummary for: ./index.rst, ./odl.diagnostics.examples.rst, ./odl.diagnostics.operator.rst, ./odl.diagnostics.rst, ./odl.diagnostics.space.rst, ./odl.discr.discr_mappings.rst, ./odl.discr.discr_ops.rst, ./odl.discr.discretization.rst, ./odl.discr.grid.rst, ./odl.discr.lp_discr.rst, ..., ./odl.tomo.backends.stir_bindings.rst, ./odl.tomo.rst, ./odl.trafos.rst, ./odl.trafos.wavelet.rst, ./odl.util.graphics.rst, ./odl.util.phantom.rst, ./odl.util.rst, ./odl.util.testutils.rst, ./odl.util.ufuncs.rst, ./odl.util.utility.rst
loading intersphinx inventory from http://python.readthedocs.org/en/latest/objects.inv...
loading intersphinx inventory from http://docs.scipy.org/doc/scipy/reference/objects.inv...
loading intersphinx inventory from http://numpy.readthedocs.org/en/latest/objects.inv...
loading intersphinx inventory from http://matplotlib.sourceforge.net/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1977 source files that are out of date
updating environment: 1977 added, 0 changed, 0 removed
reading sources... [ 33%] generated/odl.set.domain.Cuboid                                                              
Exception occurred:
  File "/home/jmoosmann/git/LCR/odl/doc/sphinxext/numpydoc/docscrape.py", line 524, in __init__
    raise ValueError("Expected a class or None, but got %r" % cls)
ValueError: Expected a class or None, but got <function Cuboid at 0x2ac8416e60c8>
The full traceback has been saved in /tmp/sphinx-err-Tq8LkZ.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 1
```
Any idea?

adler-j:
I think it may be because you have not removed the old `doc/source/generated` folder. This is not properly removed by `make clean`. I'll update the `make` script to do this later, simply remove it manually now.

moosmann:
Indeed I only did a `make clean`. Removing the folder fixed the problem.

adler-j:
I'm fixing this problem, a few commits will soon appear on master.

moosmann:
I've added CUDA support (and improved the tests a bit). Thus another review is required.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,457 @@
...
+        if out.shape != f.shape:
+            raise ValueError(
+                "shape of `out` array ({0}) does not match the shape of "
+                "input array `f` ({1}).".format(out.shape, f.shape))
+
+    if edge_order not in [1, 2]:
+        raise ValueError("edge order ({0}) not valid".format(edge_order))
+
+    if dx <= 0:
+        raise ValueError("step length is zero.")
```
kohr-h: Check and message out of sync :-)

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,457 @@
...
+    array([ 0.5,  1. ,  1. ,  1. ,  1. ,  1. ,  1. ,  1. ,  1. , -4. ])
+    >>> finite_diff(f, zero_padding=False, edge_order=1)
+    array([ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.])
+    >>> out = finite_diff(f)
+    >>> out is finite_diff(f)
+    False
+    >>> out is finite_diff(f, out)
+    True
+    """
+    # TODO: implement forward/backward differences
```
adler-j: Make an issue on this.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,457 @@
...
+    # TODO: implement alternative boundary conditions
```
adler-j: Make an issue on this.

odl/discr/discr_ops.py
```diff
@@ -0,0 +1,457 @@
...
+        -119.0
+        >>> f.inner(adj_g)
+        -119.0
+        """
+
+        ndim = self.range.grid.ndim
+        dx = self.range.grid.stride
+
+        arr = out.asarray()
+        tmp = np.zeros(out.shape, out.dtype, order=out.space.order)
```
adler-j: `empty` instead of `zeros`

test/discr/discr_ops_test.py
adler-j: Is this fixed or do we fix it in some other place?

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,421 @@
...
+
+    # Adjoint operator
+    adj_div = div.adjoint
+    ran_vec = div.range.element(data ** 2)
+    adj_div_ran_vec = adj_div(ran_vec)
+
+    # Adjoint condition
+    lhs = ran_vec.inner(div_dom_vec)
+    rhs = dom_vec.inner(adj_div_ran_vec)
+    assert lhs != 0
```
adler-j: What do we test here?

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,421 @@
...
```
adler-j: Same question as above.

adler-j:
Looks good, extremely minor things, just make the issues before we merge so we dont forget.

test/discr/discr_ops_test.py
moosmann: This was rather more used as a check not to use trivial arrays (as I already did) than a real test.

test/discr/discr_ops_test.py
moosmann: I didn't fixed it. I haven't got yet how this errors caused.

moosmann:
Implemented latest comments. Issues are created. Rebased. Builds are working. Time to merge?

adler-j:
Sure!

PR #115, opened 2015-12-03
Issue 80  ufuncs

adler-j:
Fixed the things suggested in issue #80.

Basically we can call (almost) all [UFuncs](http://docs.scipy.org/doc/numpy-1.10.0/reference/ufuncs.html) using the following syntax

```python
>>> x = vector([1.0, 2.0])
>>> x.ufunc.sin()
Rn(2).element([0.8414709848078965, 0.90929742682568171])
>>> y = x.copy()
>>> z = x.ufunc.sin(out=y)
Rn(2).element([0.8414709848078965, 0.90929742682568171])
>>> z is y
True
```

This works for any element that inherits from `NtuplesBaseVector`. 

Encapsulation
-------------
I opted to encapsulate this feature inside `x.ufunc` so that we wont have collisions such as `x.multiply` and also so that users know exactly what is happening.

With this users can do things like broadcasting addition etc:

```python
>>> x.ufunc.add(1, out=x)
Rn(2).element([2.0, 3.0])
```

Since we don't know the data storage, the `out` parameter is (by default) handled by calculating the result and then using assignment. However, for the spaces inheriting from `Ntuples` I have added a optimization, so this is as fast as numpy. I'm planning on adding this optimization for `CudaNtuples` as well, but users will only notice it in performance, the functionality is already there.

Missing ufuncs
--------------
Most ufuncs are in, the only ones missing are:

* `isreal` -> requires bool output, we don't really support it.
* `iscomplex` -> requires bool output, we don't really support it.
* `ldexp` -> requires integer input.
* `frexp` ->requires integer output.

TODO:
-----

- [ ] Possibly add missing ufuncs (if anyone needs them)
- [ ] Add CUDA support
- [x] Add optimization in `DiscreteLp`


kohr-h:
I definitely like the encapsulation. The syntax `x.sin()` wasn't really love on first sight. This is really cool stuff and makes our spaces even more usable.

Regarding boolean output, what's wrong with `Ntuples(n, dtype='bool')`? Or do you mean CUDA support? The support for the underlying set is missing, of course, but that's an easy one.

Optimizations for `DiscreteLp`: done?

adler-j:
> I definitely like the encapsulation. The syntax x.sin() wasn't really love on first sight. This is really cool stuff and makes our spaces even more usable.

Agree, this looks very nice.

> Regarding boolean output, what's wrong with Ntuples(n, dtype='bool')? Or do you mean CUDA support? The support for the underlying set is missing, of course, but that's an easy one.

Theoretically yes, but currently `__array_wrap__` returns a element in the same space as the input, since this all relies on it, we'd have to add some intelligence there. I dont really want to do that for these rather obscure UFuncs for now.

> Optimizations for DiscreteLp: done?

Yes

kohr-h:
> Theoretically yes, but currently `__array_wrap__` returns a element in the same space as the input, since this all relies on it, we'd have to add some intelligence there. I dont really want to do that for these rather obscure UFuncs for now.

Stupid idea: use the new `vector()` convenience function in `__array_wrap__`?

adler-j:
> Stupid idea: use the new vector() convenience function in __array_wrap__?

Not directly at least, this would cause

```python
>>> x = odl.vector([1.0, 2.0], impl='cuda')
>>> x
CudaRn(2).element([1.0, 2.0])
>>> x.ufunc.sin()
Rn(2).element([1.0, 2.0])
```

Even if we had a weighted space one would expect the weight to remain after calling a ufunc. We'd need to add quite some information to `vector`.

kohr-h:
The first issue would require to pass the `impl` parameter to `vector` in `__array_wrap__`, the second one is more complicated, of course.

adler-j:
Is this pullable? Anything to add would be an extension and can be added later.

kohr-h:
What about the CUDA support? If it just amounts to writing another variant of `wrap_method_...` - which I assume without knowing better - I think this should be done first. If it's harder than that, I'd say we merge first and make an issue for the CUDA support.

adler-j:
CUDA support would require properly adding the functions with out parameters in CUDA, this hasn't been done yet. I will do it at a later time (when I see it is really needed). The good thing here is that the interface still works so users can program "as if" it exists.

kohr-h:
Okay, you meant *proper* CUDA support. Well if it's already supported but not with native CUDA Ufuncs but through NumPy, then why not pull now and optimize later?

adler-j:
Yeah works on *any* FnBase space that satisfies the (somewhat unspoken requirement) that this works:

```python
>>> # x = FnBaseSpace element
>>> y = FnBaseSpace.element(x.asarray())
>>> x == y
True
```

I'll see if I can add proper cuda support first.

adler-j:
I added the already existing ufuncs to follow the protocol. A comparsion between `sin` which is optimized, and `tan` which is not, shows runtimes (when used in-place)

```
                           sin :      0.021 
                           tan :      2.891 
```

And out of place:

```
                           sin :      0.057 
                           tan :      3.026 
```

Quite an improvement. Im pulling now.

Issue #111, opened 2015-12-01
More simple uniform_discr


adler-j:
Currently `uniform_discr` looks like:

```python
recon_sp = odl.uniform_discr(odl.FunctionSpace(odl.Cuboid([-1, -1, -1],
                                                          [1, 1, 1])),
                             [15, 64, 64])
```

I propose we simplify this to allow users to do:

```python
recon_sp = odl.uniform_discr([-1, -1, -1], [1, 1, 1], [15, 64, 64])
```

We could still allow the current syntax by making it a member function.

PR #118, opened 2015-12-04
Issue 111  simple uniform discr

adler-j:
Implemented this from issue #111.

All tests run, examples need a polish.


adler-j:
I'd say this is ready for pull now. The main discussion would be the `uniform_discr` vs `uniform_discr_space` "debate". Perhaps we could make `uniform_discr_space` a member function of FunctionSpace, but I dislike adding such features to the functionspaces (also gives circular dependencies).

kohr-h:
A somewhat parallel question is "which is which"? By that I mean the question of how to read the function name. In my view `uniform_discr` needs an argument which "gives the function something to discretize", while a `uniform_discr_space` makes a uniformly discretized space out of something. So I would actually exchange the roles of these two functions. 
I know you're going to argue that the simplest name should be used for the most useful function, but it may still be worth to think about this for a second.

kohr-h:
Other than that, it's definitely in a pullable state. Free functions "feel" better for this purpose, I definitely agree.

adler-j:
> A somewhat parallel question is "which is which"? By that I mean the question of how to read the function name. In my view uniform_discr needs an argument which "gives the function something to discretize", while a uniform_discr_space makes a uniformly discretized space out of something. So I would actually exchange the roles of these two functions. 

My view is that `uniform_discr` doesn't say much more than that it creates a uniform discretization, while `uniform_discr_space` creates an uniform discretization from a space. Cleraly there is no obvious way to interpret this. One (very satisfying from an interface standpoint) way would be to somehow remove one, but I dont really see how except for scrapping `uniform_discr_space` fully and having users call `DiscreteLp` themselves, but I dislike that.

> I know you're going to argue that the simplest name should be used for the most useful function,

Yes

> but it may still be worth to think about this for a second.

Agree, theres not obvious solution.

adler-j:
Should we pull, or do we need to think about this naming first?

kohr-h:
> My view is that `uniform_discr` doesn't say much more than that it creates a uniform discretization, while `uniform_discr_space` creates an uniform discretization from a space.

I always read function names such that they return what their names say. So `uniform_discr_space` says it returns a space, which is technically true but not the point that appendix wants to make. More correct would be `uniform_discr_fromspace` (too verbose IMO) or `uniform_space_discr` which keeps the emphasis on `discr`.
With some clarification in this direction, I would be ok with the layout.



> Should we pull, or do we need to think about this naming first?

Let's sort this out first, it's an easy change.

adler-j:
I'd probably prefer `uniform_discr_fromspace` actually, mostly because it will end up close to `uniform_discr` in autocomplete lists (which shouldn't not be neglected). It wont be used that often I think so it being slightly longer should be OK.

kohr-h:
Alright, that sounds good in my ears.

PR #120, opened 2015-12-04
Fourier transform implementation

kohr-h:
### TODO

- [x] Implement `FourierTransformInverse` and write doc
- [x] Really, really check if the implementation is correct
- [x] Write math documentation
- [x] Zero padding -> new issue
- [x] Numpy based FFT as default
- [x] Add possibility to choose sign
- [x] Variants with placement of `2*pi` -> new issue

EDIT: Update TODO list, remove outdated initial text.

See below for more comments.


odl/trafos/fourier.py
```diff
@@ -0,0 +1,596 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Discrete wavelet transformation on :math:`L^p` spaces."""
```
kohr-h: Oops

adler-j:
Without spending any significant time on this. Is the interface of this guy synced with the interface of the wavelet trafo? It would be nice if they behaved similarly.

kohr-h:
Well, it's an `Operator`. What else is there to say?

Seriously, I don't think there is very much similarity to be expected. The Fourier transform has, apart from some practical decisions, no real parameters in contrast to the wavelet transform where you have to specify a wavelet and the number of detail levels. Pretty much everything in the FT is fixed and, in contrast to the wavelet transform, it makes (more) sense for the FT to regard the result as an image. For the wavelets, we need to define some sensible range (Besov space I suppose), but that needs to be well synced with the discretization stuff.

So apart from both being `Operators` with domain `DiscreteLp`, there is not so much similarity.

adler-j:
> Well, it's an Operator. What else is there to say?

Obviously transformations are a special case of operators with a special structure. As an (rather arbitrary) example the wavelet transform implements

```python
@property
def is_orthogonal(self):
    """Whether or not the wavelet basis is orthogonal."""
```

this would make sense for fourier transform as well.


> Seriously, I don't think there is very much similarity to be expected. The Fourier transform has, apart from some practical decisions, no real parameters in contrast to the wavelet transform where you have to specify a wavelet and the number of detail levels. Pretty much everything in the FT is fixed and, in contrast to the wavelet transform, it makes (more) sense for the FT to regard the result as an image. For the wavelets, we need to define some sensible range (Besov space I suppose), but that needs to be well synced with the discretization stuff.

The Fourier transform could take some parameter such as the sign convention, scaling etc. These parameters should have names that match the wavelet case (if it makes sense). For example the `DiscreteWaveletTrafoInverse` has a parameter `mode` which apparently means "Signal extension mode". If the Fourier transform also uses some extension (say zero padding), then this parameter should have a name that matches these.

> So apart from both being Operators with domain DiscreteLp, there is not so much similarity.

There is some, not that much in the mathematical sense but more in the programming sense that they both should "feel" similar to work with.

kohr-h:
> The Fourier transform could take some parameter such as the sign convention, scaling etc.

The FT in our philosophy is a discretization of the Fourier integral, so shifting and scaling is all determined by the function domain etc. What would make sense is to have the flexibility to choose between the "mathematician" version with the isometry property and the "engineer" version with `2*pi` up in the exponent. Zero padding can also be an option, but I'm trying to avoid it by having a boundary correction term. We'll see.
Periodization should be implemented as a separate operator IMO since the domain is interpreted in a completely different way. There it would actually be better to have native support for periodic functions on bounded domains.

kohr-h:
> this would make sense for fourier transform as well.

I don't see that. The Fourier basis is not integrable, so speaking of orthogonality does not make too much sense there.

niinimaki:
Just a note on wavelets and Besov spaces: all wavelets do not provide a valid basis for Besov spaces, which wavelets then do provide a theoretically valid basis for Besov spaces depends on the choices of Besov parameters. (This is not of course directly/at all related to the Fourier transform, but since the aspect of Besov space & wavelets was raised by @kohr-h I thought of mentioning  this...)

kohr-h:
Okay, so we have to think about this properly sometime. What about mapping properties `L^p --> L^q` for the wavelet transform? Let's discuss this in person before we get even more off-topic :-)

niinimaki:
I agree let's discuss this in person.

adler-j:
> The FT in our philosophy is a discretization of the Fourier integral, so shifting and scaling is all determined by the function domain etc. What would make sense is to have the flexibility to choose between the "mathematician" version with the isometry property and the "engineer" version with 2*pi up in the exponent. 

I agree in principle, but we shouldn't neglect practicalities, for example what Fourier integral? As you mention there are scaling choices in the exponent, but also a scaling of the transform itself.

> Zero padding can also be an option, but I'm trying to avoid it by having a boundary correction term. We'll see.

Mathematically it would be nice to avoid, but for practical reasons we may need it, perhaps some user requires it.

> Periodization should be implemented as a separate operator IMO since the domain is interpreted in a completely different way. There it would actually be better to have native support for periodic functions on bounded domains.

While you are correct in principle, I'm afraid of an explosion of operators and spaces. Implementation wise periodic functions are *very* similar to the usual `Rn`.

kohr-h:
PyFFTW low-level functions are in place now, operators next.

odl/trafos/fourier.py
```diff
@@ -0,0 +1,943 @@
...
+        # TODO: Implement and update doc
+        # TODO: Handle returned plan, maybe an init flag?
+        # TODO: Implement zero padding
+
+
+class DiscreteFourierTransform(Operator):
+
+    """Discretized Fourier transform between discrete Lp spaces.
+
+    The Fourier transform is defined as the linear operator
```
adler-j: To me all this math makes this impossible to read, how about a mathy introduction in the doc and keep this in pseudocode?

kohr-h: Yes, that's in fact also on my TODO list, even if I didn't write it explicitly. I realized that this is unreadable.

odl/discr/lp_discr.py
```diff
@@ -117,6 +117,16 @@ def exponent(self):
         """The exponent ``p`` in ``L^p``."""
         return self._exponent
 
+    @property
+    def conj_exponent(self):
```
adler-j: This should be a utility and not a member function?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+
+def _pyfftw_to_local(flag):
+    return flag.lstrip('FFTW_').lower()
+
+
+def _local_to_pyfftw(flag):
+    return 'FFTW_' + flag.upper()
+
+
+def _pyfftw_destroys_input(flags, direction, halfcomplex, ndim):
```
adler-j: Dramatic!

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+                        'FFTW_DESTROY_INPUT')):
+        return True
+    elif (direction in ('backward', 'FFTW_BACKWARD') and halfcomplex and
+          ndim != 1):
+        return True
+    else:
+        return False
+
+
+def _pyfftw_check_in_out(arr_in, arr_out, axes, halfcomplex, direction):
```
adler-j: `_pyfftw_check_args`?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        using all axis and is equivalent to ``np.arange(ndim)``.
+    halfcomplex : `bool`, optional
+        If `True`, calculate only the negative frequency part along the
+        last axis. If `False`, calculate the full complex FFT.
+        This option can only be used with real input data.
+    fftw_plan : ``pyfftw.FFTW``, optional
+        Use this plan instead of calculating a new one. If specified,
+        the options ``planning_effort``, ``planning_timelimit`` and
+        ``threads`` have no effect.
+    planning_effort : {'estimate', 'measure', 'patient', 'exhaustive'}
```
adler-j: Default?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        Flag for the amount of effort put into finding an optimal
+        FFTW plan. See the `FFTW doc on planner flags
+        <http://www.fftw.org/fftw3_doc/Planner-Flags.html>`_.
+    planning_timelimit : `float`, optional
+        Limit planning time to roughly this amount of seconds.
+        Default: `None` (no limit)
+    threads : `int`, optional
+        Number of threads to use. Default: 1
```
adler-j: Why not default to all threads?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+    normalise_idft : `bool`, optional
+        If `True`, the backward transform is normalized by
+        ``1 / N``, where ``N`` is the total number of points in
+        ``array_in[axes]``. This ensures that the IDFT is the true
+        inverse of the forward DFT.
+        Default: `False`
+    import_wisdom : `str`, optional
+        File name to load FFTW wisdom from. If the file does not exist,
+        it is ignored.
+    export_wisdom : `str`, optional
```
adler-j: I think we should follow the `numpy.savetxt` pattern here:

> fname : filename or file handle

That way users could provide a file handle for some file in an arbitrary folder.

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+    planning_effort = _pyfftw_to_local(kwargs.pop('planning_effort',
+                                                  'estimate'))
+    planning_timelimit = kwargs.pop('planning_timelimit', None)
+    threads = kwargs.pop('threads', 1)
+    normalise_idft = kwargs.pop('normalise_idft', False)
+    wimport = kwargs.pop('import_wisdom', '')
+    wexport = kwargs.pop('export_wisdom', '')
+
+    # Cast input to complex if necessary
+    array_in_copied = False
```
adler-j: This should be a flag that users can provide as `True` in case they have done this externally.

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        """Implementation of ``self(x[, out])``.
+
+        Parameters
+        ----------
+        x : domain element
+            Input vector to be transformed
+        out : range element, optional
+            Output vector storing the result
+        direction : {'FFTW_FORWARD', 'FFTW_BACKWARD'}, optional
+            Direction of the transform
```
adler-j: Default?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+            ran_dspace = ran_dspace_type(ran_grid.size, dtype=ran_dtype,
+                                         exponent=conj_exp)
+            ran = DiscreteLp(ran_fspace, ran_grid, ran_dspace,
+                             exponent=conj_exp)
+
+        super().__init__(dom, ran, linear=True)
+        self._fftw_plan = None
+
+    def _call(self, x, out=None, **kwargs):
+        """Implementation of ``self(x[, out])``.
```
adler-j: self(x[, out, **kwargs])

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        Returns
+        -------
+        out : `DiscreteLpVector`
+            Result of the transform. If ``out`` was given, the returned
+            object is a reference to it.
+        """
+        if out is None:
+            out = self.range.element()
+
+        self._fftw_plan = pyfftw_call(
```
adler-j: Regarding `array_in_copied` it could be checked here by checking:

```python
arr = x.asarray()
array_in_copied = arr.flags.owndata
```

and perhaps as a kwarg

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        # TODO: Implement zero padding
+
+
+class FourierTransform(Operator):
+
+    """Discretized Fourier transform between discrete Lp spaces.
+
+    The Fourier transform is defined as the linear operator
+
+        :math:`\mathcal{F}: L^p(\mathbb{R}^d) \\to L^q(\mathbb{R}^d)`,
```
adler-j: Code instead

adler-j: Note: this can live in the doc

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        """
+        # TODO: variants wrt placement of 2*pi
+        # TODO: impl flag for Numpy vs. PyFFTW
+
+        if not isinstance(dom, DiscreteLp):
+            raise TypeError('domain {!r} is not a `DiscreteLp` instance.'
+                            ''.format(dom))
+        if not isinstance(dom.dspace, Ntuples):
+            raise NotImplementedError(
+                'Only Numpy-based data spaces are supported, got {}.'
```
adler-j: Why?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+        # TODO: Implement zero padding
+
+        # We pop some kwargs options here so that we always use the ones
+        # given during init or implicitly assumed.
+        kwargs.pop('axes', None)
+        kwargs.pop('halfcomplex', None)
+        kwargs.pop('normalise_idft', None)
+
+        # We're always modifying the input, so a copy is unavoidable
+        x_cpy = x.copy()
```
adler-j: Here we get two copies, one here and one inside the call, needed?

odl/trafos/fourier.py
```diff
@@ -0,0 +1,1024 @@
...
+                                      'not {}.'.format(self.domain.exponent))
+
+    @property
+    def inverse(self):
+        """The inverse Fourier transform."""
+        # TODO: add appropriate arguments
+        raise NotImplementedError
+
+
+class InverseFourierTransform(Operator):
```
adler-j: Do we really need a separate class for the inverse?

test/trafos/fourier_test.py
```diff
@@ -0,0 +1,786 @@
...
+# TODO: add reciprocal test with axes
+pytestmark = pytest.mark.skipif("not odl.trafos.PYFFTW_AVAILABLE")
+
+
+exp_params = [2.0, 1.0, float('inf'), 1.5]
+exp_ids = [' p = {} '.format(p) for p in exp_params]
+exp_fixture = pytest.fixture(scope="module", ids=exp_ids, params=exp_params)
+
+
+@exp_fixture
```
adler-j: No need to create a `exp_fixture` and only use it once

test/trafos/fourier_test.py
```diff
@@ -0,0 +1,786 @@
...
+def exponent(request):
+    return request.param
+
+
+if platform.system() == 'Linux':
```
adler-j: Perhaps add a utility for this instead?

test/trafos/fourier_test.py
```diff
@@ -0,0 +1,786 @@
...
+    dtype_params = ['float32', 'float64', 'float128',
+                    'complex64', 'complex128', 'complex256']
+else:
+    dtype_params = ['float32', 'float64', 'complex64', 'complex128']
+dtype_ids = [' dtype = {} '.format(dt) for dt in dtype_params]
+dtype_fixture = pytest.fixture(scope="module", ids=dtype_ids,
+                               params=dtype_params)
+
+
+@dtype_fixture
```
adler-j: Simply use the fixture as is instead of creating a variable

test/trafos/fourier_test.py
```diff
@@ -0,0 +1,786 @@
...
+plan_fixture = pytest.fixture(scope="module", ids=plan_ids,
+                              params=plan_params)
+
+
+@plan_fixture
+def planning(request):
+    return request.param
+
+
+def _random_array(shape, dtype):
```
adler-j: We have this in every file, perhaps move to testutils?

test/trafos/fourier_test.py
```diff
@@ -0,0 +1,786 @@
...
+                np.exp(-1j * x[1] * 3 / 2) * sinc(x[1] / 2) /
+                (2 * np.pi))
+
+    discr = odl.uniform_discr([-2] * 2, [2] * 2, (65,) * 2, impl='numpy',
+                              field=odl.ComplexNumbers())
+    dft = FourierTransform(discr)
+    func_true_ft = dft.range.element(fshift_char_rect_ft)
+    func_dft = dft(fshift_char_rect)
+    assert (func_dft - func_true_ft).norm() < 0.001
+
```
adler-j: Did not review tests yet.

test/util/utility_test.py
```diff
@@ -62,5 +64,104 @@ def test_is_complex_floating_dtype():
...
+    test_arr = np.ones(shape)
+    fast_1d_tensor_mult(test_arr, [y, z], axes=(1, 2))
+    assert all_equal(test_arr, true_result)
+
+    test_arr = np.ones(shape)
+    fast_1d_tensor_mult(test_arr, [z, y], axes=(2, 1))
+    assert all_equal(test_arr, true_result)
+
+
+def test_fast_1d_tensor_mult_error():
```
adler-j: `error` -> `exception`

adler-j:
Looking good, just added some minor review comments for you.

odl/trafos/fourier.py
kohr-h: Okay, I can just remove that line.

odl/trafos/fourier.py
kohr-h: Here we need it, yes. The input is first preprocessed in place, so we need the copy. An then it's destroyed during planning if we don't use 'estimate'. In the latter case, we don't copy, otherwise yes. I don't quite know how to circumvent this.

odl/trafos/fourier.py
kohr-h: TBD

test/trafos/fourier_test.py
kohr-h: Ah that's good!

test/trafos/fourier_test.py
kohr-h: There may be more cases later.

test/trafos/fourier_test.py
kohr-h: Make sense

odl/discr/lp_discr.py
kohr-h: Correct, that would be a better fit

odl/trafos/fourier.py
kohr-h: You need some drama sometimes :-)

odl/trafos/fourier.py
kohr-h: Why not?

odl/trafos/fourier.py
kohr-h: But it's actually the same wording as in the FFTW doc, so not my own creation.

odl/trafos/fourier.py
kohr-h: Hm, must have been a non-kwarg before. Will change.

odl/trafos/fourier.py
kohr-h: I'm not aware of an option to simply say "use all". You need to give the number explicitly. And on my machine, 3 threads were faster than 4 but I have 8 cores?? Perhaps there's some clever way to find out the number of cores and use a reasonable number based on that.

odl/trafos/fourier.py
kohr-h: Okay, can do.

odl/trafos/fourier.py
kohr-h: OK

odl/trafos/fourier.py
kohr-h: Shouldn't be an argument at all.

odl/trafos/fourier.py
kohr-h: Sure.

odl/trafos/fourier.py
kohr-h: The thing is that if `pyfftw_call` does FFT planning internally, the arrays are zeroed out. So once the plan is there and can be reused we're on the safe side. Otherwise, just checking if the data was once copied doesn't help much. Or am I getting something wrong?

odl/trafos/fourier.py
kohr-h: It's on the extended agenda.

odl/trafos/fourier.py
kohr-h: The best I can think of is to just use empty arrays for the planning, but they have to be of the same size and dtype. So I don't know how much would be gained.

adler-j: Guess its minor then, but does planning really need a full array? Cant we do that once for the operator? Perhaps on first call or during creation?

odl/trafos/fourier.py
adler-j: I thought it was to make sure the input vector vas not destroyed (we require the input arg to be untouched by operators), if that is the case the difference is more minor, but see my other comment on reusing this.

odl/trafos/fourier.py
kohr-h: Oh but that's how it is (I hope :-) ). The FFTW plan is returned by the `pyfftw_call` function and in the next operator call, the same plan is fed back into `pyfftw_call`, which then skips all that. So you plan only the very first time the operator is evaluated.

odl/trafos/fourier.py
kohr-h: One can use `multiprocessing.cpu_count()` as a default. I'll do some tests later, maybe a note for the user would be helpful. The [FFTW doc on this](http://www.fftw.org/doc/How-Many-Threads-to-Use_003f.html) basically says nothing.

adler-j: > Typically, the problem will have to involve at least a few thousand data points before threads become beneficial

So basically set this to 1 if transform is smaller than say 1000 elements?

kohr-h: OK, did that

kohr-h:
I'll make a new TODO list here:

- [ ] Implement `FourierTransformInverse`
- [ ] Shoot annoying bug when using complex input
- [ ] Zero padding
- [ ] Numpy based FFT as default
- [ ] Variants with placement of `2*pi` -> new issue

The current bug does not occur if the transformed function is purely real or a real function times a complex number. Only if real and imaginary parts are independent function, maybe also for purely imaginary input. The result looks kind of correct, but not quite. The error is constant at 5 % maybe, and it does not decrease with growing number of samples. I'll have to look closer into it.

Boundary correction turns out to be prohibitively costly. It would require addition of `M` full-sized arrays to the calculated FT, where `M` is the number of lower-dimensional boundary sets (i.e. 2 in 1d due to 2 boundary points, 8 in 2d due to 4 edges and 4 corners, 22 in 3d due to 6 faces, 8 edges and 8 corners, etc.). We'll stick with zero padding I guess. Maybe someone has a better idea, like a sinc series type approach?

adler-j:
> I'll make a new TODO list here:

Looking good, but editing the main post causes the status to show in the pulls tab :)

> The current bug does not occur if the transformed function is purely real or a real function times a complex number.

Perhaps simply add an exception/warning for that and make a further issue. Do we need complex atm?

> Boundary correction turns out to be prohibitively costly.

Sounds quite bad, yeah. Perhaps we can skip it.

> Maybe someone has a better idea, like a sinc series type approach?

Is there an application where we need this?

kohr-h:
> Looking good, but editing the main post causes the status to show in the pulls tab :)

Aha, I didn't know that. I'll shove it to the top.

> Perhaps simply add an exception/warning for that and make a further issue. Do we need complex atm?

I wouldn't exclude it, but usually the input is real. I'll add some more tests so I know exactly when the error occurs. If it's a non-standard situation, a temporary raise and a new issue would be a solution.

> Is there an application where we need this?

Not really. It was more thought as a way to handle partitions with the last grid points at the boundary in a consistent manner. But it looks like zero padding is much more feasible.

kohr-h:
Some more work on this. Now all tests with known pairs succeed when I use more sampling points. I'm not sure if I believe this completely yet, but for the rather non-smooth examples I used in the tests, one actually needs a certain amount of samples to cover enough of the Fourier space. The FTs of these guys only decay slowly, so one misses information. And the error is not monotone in the number of samples in the region of, say, 100. Probably if you hit the oscillations in Fourier space in a fortunate manner, you get less error than with more samples, but worse distribution.

I declare the annoying bug as no longer existing for now and go on with my life.

kohr-h:
I finally had enough of all the guesswork and wrote a test which calculates the FT of 2 functions step by step and compares intermediate results with hand-calculated ones. The examples are characteristic functions in a discretization with 4 points only, which nonetheless is enough to exactly represent those functions. More precisely, the interval is [-2, 2] and the functions are the characteristic functions of [-1, 1] and [0, 1], respectively. And since we know that the discretization is exact, the FT must be, too. That test looks all good now.

The pure DFT operators low look more like the discretized "integral FT" operator in their init syntax, and I think those now have a decent quality.

Numpy is now the default backend, works fine. Now zero padding and the FT inverse remains to be done, seems manageable.

kohr-h:
So I'm finally done with this. Perhaps another review would be good, but I can summarize the changes since last time:

- Documentation on the mathematics now resides in doc/source/math/trafos/ as a proper math introduction into the topic and the basis for the algorithms. Docstrings are now more console-friendly and shorter.
- Addition of `real`, `imag` (with setters) and `conj` to `DiscreteLpVector`
- New helper `complex_space` for creating a complex space counterpart of a real space; `real_space` for the opposite
- All transforms now have a `sign` parameter where one can choose the sign in the complex exponential
- The inverses are exact inverses (up to arithmetic error) of the discrete guys. Especially the inverse of `FourierTransform` is not the discretization of the operator with opposite sign.
- The helper functions are rewritten for array-like input rather than `DiscreteLpVector` which makes them more general-purpose.
- All transforms have methods to initialize FFTW plans so the subsequent runs profit from the cached plan.
- A small example has been added.

kohr-h:
The code [here](https://github.com/odlgroup/odl/blob/95c28c6d4734b09c76bfeabe586a70b99317f1d7/odl/trafos/fourier.py#L676) and at other places with similar functionality is preliminary and relies on a solution of #284.

kohr-h:
I'm going to merge this if there are no objections.

kohr-h:
Coverage could be a bit better, I'll check it out. Mostly trivial misses from what I remember.

odl/discr/lp_discr.py
```diff
@@ -446,8 +528,16 @@ def __setitem__(self, indices, values):
             shape is allowed as ``values``.
         """
         if values in self.space:
-            self.ntuple.__setitem__(indices, values.ntuple)
+            # For RawDiscretizationVector of the same type, use ntuple directly
+            return self.ntuple.__setitem__(indices, values.ntuple)
+        elif (isinstance(self.space.dspace, Ntuples) and
+              equiv_views(self.asarray(), values) and
```
adler-j: This cannot be merged. It makes some operations exponentially slower, e.g.

```
>>> X = odl.CudaRn(10**7)
>>> x = X.element()
>>> x[0] = 2
```

currently this is almost instant, with this change this would be an O(n) operation.

kohr-h: I don't get it. What exactly would be O(n) here? The instance check?

adler-j: The asarray call

adler-j: Dont we have benchmarks for these things?

kohr-h: Which is never done for CUDA spaces?

adler-j: Ahh and I'm slow. We do a `Ntuples` check first. :/

Still I dislike these very specfic optimizations, is there no way to simply push this to the dspace?

odl/discr/lp_discr.py
```diff
@@ -446,8 +528,16 @@ def __setitem__(self, indices, values):
             shape is allowed as ``values``.
         """
         if values in self.space:
-            self.ntuple.__setitem__(indices, values.ntuple)
+            # For RawDiscretizationVector of the same type, use ntuple directly
+            return self.ntuple.__setitem__(indices, values.ntuple)
```
adler-j: I suppose it would be more pythonic to do

```python
self.ntuple[indices] = values.ntuple
```

odl/discr/lp_discr.py
adler-j: Also, if we know that `self.space.dspace` is a `Ntuples`, we should use `self.data` instead since this is both faster and more clear.

kohr-h: I can add a TODO, we probably want to use a context manager once that is in place. This was more like a quick fix.

kohr-h: > Also, if we know that self.space.dspace is a Ntuples, we should use self.data instead since this is both faster and more clear.

I'll change that. Otherwise, ok with a TODO?

adler-j: Do you have a before-after benchmark of this? I still feel like numpy should optimize this properly.

kohr-h: I'll benchmark it.

examples/fourier_trafo.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Internal
+import odl
+
+
+# Create a uniformly discretized space of functions on the rectangle
+# [-1, 1] x [-1, 1] with 512 samples per dimension. Using a complex data
+# type so one naturally gets the full Fourier transform (see below for real
+# data).
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (512, 512), dtype='complex')
```
adler-j: [-1, -1], [1, 1], [512, 512]

kohr-h: I actually prefer tuples for the shape, since that's what you get back if you ask for it.

doc/source/math/trafos/index.rst
```diff
@@ -0,0 +1,10 @@
+###############
+Transformations
+###############
+
+This section contains the mathematical descriptions of (integral) transforms implemented in ODL.
+
+.. toctree::
+    :maxdepth: 3
+    
+    fourier_transform
```
adler-j: Looks good, we could use one for wavelets aswell :)

examples/fourier_trafo.py
```diff
@@ -0,0 +1,59 @@
...
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+# Internal
+import odl
+
+
+# Create a uniformly discretized space of functions on the rectangle
```
adler-j: This is way to long for one of the most standard operations in ODL

odl/discr/lp_discr.py
```diff
@@ -605,6 +695,108 @@ def show(self, method='', title=None, indices=None, show=False, fig=None,
                                   show=show, fig=fig, **kwargs)
 
 
+def uniform_discr_frompartition(partition, exponent=2.0, interp='nearest',
```
adler-j: This has to be on the wrong branch?

odl/discr/lp_discr.py
```diff
@@ -626,9 +818,12 @@ def uniform_discr_fromspace(fspace, nsamples, exponent=2.0, interp='nearest',
             'nearest' : use nearest-neighbor interpolation (default)
 
             'linear' : use linear interpolation (not implemented)
+
     impl : {'numpy', 'cuda'}
         Implementation of the data storage arrays
 
+    Other Parameters
```
adler-j: Is this a valid header in numpydoc?

odl/discr/lp_discr.py
kohr-h: The lack of this became obvious in this branch, so that's how it ended up here. I can cherry-pick the commit to master if that helps.

odl/discr/lp_discr.py
```diff
@@ -785,6 +997,60 @@ def uniform_discr(min_corner, max_corner, nsamples,
                                    **kwargs)
 
 
+def sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
+    """Return an object mimicing the sequence space ``l^p(R^d)``.
```
adler-j: How is this different froma R^d space?

odl/discr/lp_discr.py
kohr-h: Yes. From the numpydoc doc:

    7. Other Parameters

    An optional section used to describe infrequently used parameters.
    It should only be used if a function has a large number of keyword parameters,
    to prevent cluttering the Parameters section.

Perhaps an over-use here, but sometimes it may be worth separating the most important parameters from the advanced, extra ones. AFAIK they end up further down in the Sphinx doc also.

odl/discr/lp_discr.py
kohr-h: Right. In practice, the only difference is the multi-dimensionality of the indices.

odl/discr/lp_discr.py
kohr-h: Oops, no significant time difference :-)

I'll remove the code :innocent:

odl/space/cu_ntuples.py
```diff
@@ -536,19 +536,24 @@ def __init__(self, size, dtype, **kwargs):
...
             Only scalar data types are allowed.
 
-        weight : `array-like`, `CudaFnVector` or `float`, optional
-            Use weighted inner product, norm, and dist.
+        weight : optional
+            Use weighted inner product, norm, and dist. The following
+            types are supported as ``weight``:
 
-            `float`:
+            `FnWeightingBase` :
```
adler-j: type?

odl/space/fspace.py
```diff
@@ -500,6 +500,11 @@ def element(self, fcall=None, vectorized=True):
 
             return self.element_type(self, fcall)
 
+    @property
+    def out_dtype(self):
+        """Output data type of functions in this space."""
+        return 'complex128' if self.field == ComplexNumbers() else 'float64'
```
adler-j: np.dtype('complex128') would probably be expected

adler-j: And perhaps this should be customize-able?

odl/space/ntuples.py
```diff
@@ -390,6 +390,9 @@ def __setitem__(self, indices, values):
         >>> x
         Ntuples(2, 'int8').element([0, 0])
         """
+        if ((values is self or values is self.data) and
```
adler-j: Should be redundant?

odl/util/utility.py
```diff
@@ -163,6 +185,145 @@ def is_complex_floating_dtype(dtype):
     return np.issubsctype(dtype, np.complexfloating)
 
 
+def default_dtype(impl, field):
```
adler-j: Shouldnt this be possible to do in some smarter way, perhaps as a static function on `Ntuples` and `CudaNtuples`? That would be easier to extend.

adler-j: We have to think about what will happen if a user wants to implement for example a MPI space.

odl/util/utility.py
```diff
@@ -163,6 +185,145 @@ def is_complex_floating_dtype(dtype):
...
+            raise NotImplementedError('complex data types not supported in '
+                                      'CUDA.')
+            # dtype = np.dtype('complex64')
+    else:
+        raise ValueError("'impl '{}' not understood.".format(impl))
+
+    return dtype
+
+
+def complex_space(space):
```
adler-j: Should also (imo) be a member function so it is easier to extend.

test/space/cu_ntuples_test.py
```diff
@@ -38,9 +38,6 @@
 
 pytestmark = pytest.mark.skipif("not odl.CUDA_AVAILABLE")
 
-# TODO:
-# * custom dist/norm/inner
```
adler-j: We have tests for these?

test/space/ntuples_test.py
```diff
@@ -39,11 +39,6 @@
 from odl.util.testutils import almost_equal, all_almost_equal, all_equal
 from odl.util.ufuncs import UFUNCS, REDUCTIONS
 
-# TODO: add tests for:
```
adler-j: We have tests here too?

test/util/utility_test.py
```diff
@@ -62,5 +85,74 @@ def test_is_complex_floating_dtype():
...
+
+    if real_float_dtype == np.dtype('float16'):
+        return
+
+    discr = odl.uniform_discr([0, 0], [1, 2], (10, 20),
+                              dtype=real_float_dtype)
+
+    cdiscr = odl.uniform_discr([0, 0], [1, 2], (10, 20),
+                               dtype=_TYPE_MAP_R2C[real_float_dtype])
+
```
adler-j: Id also

```python
assert real_space(discr) == discr
assert complex_space(cdiscr) == cdiscr
```

adler-j: Damn you did above. -.-

odl/space/cu_ntuples.py
kohr-h: type of what?

adler-j: Was expecting a type after the :, notice that it may be un-needed here

odl/space/ntuples.py
kohr-h: Can also go away.

odl/util/utility.py
kohr-h: This is already a bit smarter than the old copy-paste solution. But I agree, this still feels a bit clumsy.

Static method sounds good.

test/space/cu_ntuples_test.py
kohr-h: Yes

test/space/ntuples_test.py
kohr-h: Yes, all of it. This comment was just never removed.

test/util/utility_test.py
kohr-h: ^^

odl/space/cu_ntuples.py
kohr-h: Yeah it's already the second level, so we're explaining the behavior for different types of input here.

odl/space/fspace.py
```diff
@@ -721,9 +724,9 @@ def ipow_posint(x, n):
         def power_call_out_of_place(x):
             """The power out-of-place evaluation function."""
             if p == int(p) and p >= 1:
-                return pow_posint(x_call_oop(x), int(p))
+                return pow_posint(x_call_oop(x), int(p)).astype(self.out_dtype)
```
adler-j: what if the result is not a numpy array?

kohr-h: Hm, I have a hard time imagining a situation where this would happen, but to be sure, one could use `np.asarray(..., dtype=self.out_dtype)` instead.

examples/fourier_trafo.py
kohr-h: I don't agree, and I don't see the problem with longer explanations in examples. This may be the very first example somebody looks at, and if I instead write `# Create discretization` as in a number of examples we have, that's not helpful at all.

adler-j: I guess you are right, but we should strive for uniformity. If we want this level of doc we should make an effort in the other examples as well.

kohr-h: I went through the examples in the conebeam branch but I [managed to be more concise](https://github.com/odlgroup/odl/blob/22afa9852ff08752c54735068be15e241a456f9d/examples/tomo/xray_trafo_circular_cone.py#L29-L33). Perhaps this is a viable middle ground.

kohr-h: See #290.

odl/discr/lp_discr.py
adler-j: Personally I would like to think of sequence spaces as a function space defined on the (positive) integers. Could we think of another name here?

odl/util/utility.py
kohr-h: I have implemented the static methods. With those, the code would look like this:
```python
dspc_type = dspace_type(fspace, impl)
dtype = dspc_type.default_dtype(fspace.field)
```
which is more intelligent and extensible. We need that additional step to infer the class first and then get the default type. Maybe that's okay.

kohr-h: I mean the code [here](https://github.com/odlgroup/odl/pull/120/files#diff-c53e5a0e4ff98988cce249116ccaaabcR882)

kohr-h: Actually, the code I posted is wrong since it can return `Rn`, for example, which is a function and does not have a `default_dtype` method. Instead we must hard-code
```python
dspc_type = CudaFn if impl == 'cuda' else Fn
dtype = dspc_type.default_dtype(fspace.field)
```

odl/discr/lp_discr.py
kohr-h: Isn't that exactly what you get? After slight adaptions:
```python
>>> spc = odl.sequence_space((4, 2))
>>> spc.grid.points()
array([[ 0.,  0.],
       [ 0.,  1.],
       [ 1.,  0.],
       [ 1.,  1.],
       [ 2.,  0.],
       [ 2.,  1.],
       [ 3.,  0.],
       [ 3.,  1.]])
```

odl/util/utility.py
adler-j: Why does it return Rn? It is just a wrapper for Fn with some flags. Would be more simple to just do Fn.

odl/discr/lp_discr.py
adler-j: I mean the space of functions on ALL the integers (similar to how our FunctionSpace(IntervalProd) is a space of functions on ALL reals in some set). Discretization would then be perhaps truncation.

odl/util/utility.py
kohr-h: We could change it to return `Fn` and the data type.

odl/discr/lp_discr.py
kohr-h: That's a different story, of course. If we need that at all. Then this is the simplest truncated version of that.

odl/space/fspace.py
kohr-h: It is now since fb2e8edb20972072531dea5e529a1aa916f699c0.

odl/discr/lp_discr.py
kohr-h: I removed the over-uses of this section. Just a few left.

kohr-h:
Some more improvements:

- `FunctionSet` can now be customized with an `out_dtype` parameter. That simplifies also some internal code. If not given, the "lazy" data type selection is done.
- Functions in `numerics.py` have `out` parameters now.
- `sequence_space` uses integer grid points. Suggestions for a name?
- The `dspace_type` helper in `discretization.py` returns space type and data type, and the returned space type is `Fn` or `CudaFn` for all scalar cases, not `Rn` and the likes (not types). If data type is not given as input, the default is inferred from the space type via, e.g. `Fn.default_dtype(field)`. If the field is `None`, the returned data type is also `None`. OK?
```python
>>> fset = odl.FunctionSet(odl.Interval(0, 1), odl.Strings(2))
>>> dspace_type(fset, 'numpy')
(odl.space.ntuples.Ntuples, None)
>>> dspace_type(fset, 'numpy', np.int)
(odl.space.ntuples.Ntuples, int)

>>> rspc = odl.FunctionSpace(odl.Interval(0, 1), field=odl.RealNumbers())
>>> dspace_type(rspc, 'numpy')
(odl.space.ntuples.Fn, dtype('float64'))
>>> dspace_type(rspc, 'cuda')
(odl.space.cu_ntuples.CudaFn, dtype('float32'))
```

odl/discr/lp_discr.py
kohr-h: How shall we call this guy now? It's isometrically isomorphic to any discretization of a sequence space with the same number of samples, so it's not wrong to call it sequence space (maybe with `discr_` prefix). An alternative would be simply call it what it is, namely a uniform discretization with unit cells, `uniform_discr_unitcells` maybe?

odl/space/fspace.py
kohr-h: So now this is fully customizable. In `FunctionSet` [you can give the output data type](https://github.com/odlgroup/odl/blob/6fbc2e3cbb8f0d239aa295806ec3c9b12ae99e32/odl/space/fspace.py#L84-L90), and if you don't, the ["lazy" approach](https://github.com/odlgroup/odl/blob/6fbc2e3cbb8f0d239aa295806ec3c9b12ae99e32/odl/space/fspace.py#L62-L64) is used.
For `FunctionSpace`, again, you can give a data type explicitly, but if you don't, the `field` argument [is used to determine the default](https://github.com/odlgroup/odl/blob/6fbc2e3cbb8f0d239aa295806ec3c9b12ae99e32/odl/space/fspace.py#L484-L518) (`'float64'` if both are `None`). Then there are some consistency checks.
If the field is not `RealNumbers` or `ComplexNumbers`, you need to give the data type explicitly. This could be changed to setting it to `None` as for `FunctionSet` and rely on lazy evaluation, now that I think of it.

odl/util/utility.py
kohr-h: `dspace_type` now returns `Fn` or `CudaFn`.

odl/util/utility.py
kohr-h: These are now member functions of `FnBase`, `FunctionSpace` and `DiscreteLp`. I thought about putting it into `Discretization` directly, but there you can have any `LinearSpace` as `uspace`, and it's not possible to turn that into a complex space (unknown init signature).

odl/space/fspace.py
kohr-h: I did that change.

kohr-h:
All comments implemented. I did one subtle change to `uniform_discr_fromspace` which is quite significant. It uses the `out_dtype` of the given function space as default `dtype` instead of the default for the field. I think this is sensible and avoids mismatch (e.g. function space returns in double precision but data is stored in single precision). If that's okay, how about merging?

odl/discr/lp_discr.py
adler-j: Id add the discr prefix to this name. It would be better to not call it unitcells IMO since this can be thought of as more general than that.

kohr-h: I guess you're right. I'll add the `discr_` prefix.

kohr-h: Done.

adler-j:
I'm quite sure its good, but will do a quick read tonight. Need an overall look especially at the interface etc.

conftest.py
```diff
@@ -30,3 +30,11 @@
     collect_ignore.append('odl/space/cu_ntuples.py')
 if not PYWAVELETS_AVAILABLE:
     collect_ignore.append('odl/trafos/wavelet.py')
+
+
+def pytest_addoption(parser):
+    parser.addoption('--largescale', action='store_true',
```
adler-j: How did this end up here?

examples/fourier_trafo.py
```diff
@@ -0,0 +1,57 @@
...
+phantom_ft = ft_op(phantom)
+phantom_ft.show(title='Full Fourier Transform', show=True)
+
+# Calculate the FT only along the first axis
+ft_op_axis0 = odl.trafos.FourierTransform(discr_space, axes=(0,))
+phantom_ft_axis0 = ft_op_axis0(phantom)
+phantom_ft_axis0.show(title='Fourier transform along axis 0', show=True)
+
+# If a real space is used, the Fourier transform can be calculated in the
+# "half-complex" mode. This means that along the last axis of the transform,
```
adler-j: Shouldnt this be the default? I think it is what most "computer side" people would expect. But perhaps not the math crowd?

odl/discr/discretization.py
```diff
@@ -214,11 +214,6 @@ def domain(self):
         return self.uspace.domain
 
     @property
-    def dtype(self):
```
adler-j: Why?

odl/discr/lp_discr.py
```diff
@@ -36,15 +36,18 @@
 from odl.discr.partition import RectPartition, uniform_partition_fromintv
 from odl.set.sets import RealNumbers, ComplexNumbers
 from odl.set.domain import IntervalProd
+from odl.space.base_ntuples import _TYPE_MAP_R2C, _TYPE_MAP_C2R
```
adler-j: we should cleanup these by using the "topmost" import we can use.

odl/discr/lp_discr.py
```diff
@@ -204,6 +207,11 @@ def element(self, inp=None):
             return self.element_type(self, self.dspace.element())
         elif inp in self.dspace:
             return self.element_type(self, inp)
+        elif isinstance(inp, self.element_type):
+            raise TypeError('input {!r} not an element of {}.'
```
adler-j: This logic feels wierd. Should there be a `not` in the elif statement?

odl/discr/lp_discr.py
```diff
@@ -221,18 +229,88 @@ def element(self, inp=None):
                         ''.format(arr.shape, self.shape))
             arr = arr.ravel(order=self.order)
             return self.element_type(self, self.dspace.element(arr))
-        except TypeError as err:
+        except (TypeError, IndexError) as err:
             if str(err.args[0]).startswith('output contains points outside'):
```
adler-j: Is this the "best" pattern to do this? Shouldnt we create a new error for this case that inherits from the correct one?

odl/discr/lp_discr.py
```diff
@@ -221,18 +229,88 @@ def element(self, inp=None):
...
                 raise_from(TypeError('unable to create an element of {} from '
-                                     '{!r}.'.format(self, inp)), err)
+                                     '{!r}: {}'.format(self, inp, err)), err)
 
     @property
     def interp(self):
         """Interpolation type of this discretization."""
         return self._interp
 
+    def as_complex_space(self, dtype=None):
```
adler-j: Not quite sure i want different functions here. I'd prefer one big `astype` that takes `dtype`, `impl` etc.

odl/discr/lp_discr.py
```diff
@@ -427,6 +511,73 @@ def cell_volume(self):
         """Cell volume of an underlying regular grid."""
         return self.space.cell_volume
 
+    @property
+    def real(self):
+        """Real part of this element."""
+        return self.space.as_real_space().element(self.asarray().real)
```
adler-j: Maybe cache this real space somehow? Afraid of all these creations in these.

odl/discr/lp_discr.py
```diff
@@ -427,6 +511,73 @@ def cell_volume(self):
...
+
+    @real.setter
```
adler-j: Why do we need setters? Shouldnt users be calling `x.real[:] = ...` or similar?

odl/discr/lp_discr.py
```diff
@@ -733,22 +1003,25 @@ def uniform_discr(min_corner, max_corner, nsamples,
...
         Data type for the discretized space
 
-            Default for 'numpy': 'float64' / 'complex128'
+            Default for 'numpy': 'float64'
 
-            Default for 'cuda': 'float32' / TODO
+            Default for 'cuda': 'float32'
 
-    weighting : {'simple', 'consistent'}
+    order : {'C', 'F'}, optional
```
adler-j: We really should have a glossary entry on this.

odl/space/base_ntuples.py
```diff
@@ -448,6 +437,66 @@ def is_cn(self):
         """
         return (not self._is_real) and self._is_floating
 
+    def as_complex_space(self, dtype=None):
```
adler-j: Same comments as above apply here. Dislike the possible explosion of functions, better have one unified.

odl/space/cu_ntuples.py
```diff
@@ -668,6 +676,39 @@ def exponent(self):
         """Exponent of the norm and distance."""
         return self._space_funcs.exponent
 
+    @property
+    def weighting(self):
```
adler-j: Has this been added consistently?

test/largescale/trafos/fourier_slow_test.py
```diff
@@ -0,0 +1,132 @@
...
+              odl.uniform_discr(-2, 2, 10 ** 5, dtype='complex'),
+              odl.uniform_discr([-2, -2, -2], [2, 2, 2], [200, 200, 200],
+                                dtype='complex')]
+
+dom_ids = [' {!r} '.format(dom) for dom in dom_params]
+
+
+impl_params = ['numpy']
+if odl.trafos.PYFFTW_AVAILABLE:
+    impl_params += ['pyfftw']
```
adler-j: These should say skipped otherwise I think, not sure how to do it.

test/largescale/trafos/fourier_slow_test.py
```diff
@@ -0,0 +1,132 @@
...
+
+def test_dft_forward(domain, impl):
+
+    halfcomplex = np.issubdtype(domain.dtype, np.floating)
+    dft = odl.trafos.DiscreteFourierTransform(domain, halfcomplex=halfcomplex)
+    one = domain.one()
+    out = dft.range.element()
+
+    dft(one, out=out)
+    assert out.ntuple[0] == domain.size
```
adler-j: out[0]?

conftest.py
kohr-h: I added a slow test.

examples/fourier_trafo.py
kohr-h: For real data, it is the default.

odl/discr/discretization.py
kohr-h: Defined in `NtuplesBase` already, this was a duplicate.

odl/discr/lp_discr.py
kohr-h: OK

odl/discr/lp_discr.py
kohr-h: The logic is "if in space, ok; otherwise, if `DiscreteLpVector` from foreign space, bad". This keeps vectors from other spaces reaching the sequence-type input part. Maybe one could clarify by writing
```python
elif isinstance(inp, self.element_type) and inp not in space:
    # Same kind, but different space -> error
    raise ...
elif inp in space:
    return ...
```

odl/discr/lp_discr.py
kohr-h: `astype`, eh? I think the current guys could be generalized to that, ok.

odl/discr/lp_discr.py
kohr-h: Kind of clumsy, I know. Maybe an `OpDomainError` would be sensible?

odl/discr/lp_discr.py
kohr-h: Could, be done during init, ok.

odl/discr/lp_discr.py
kohr-h: This mimics Numpy's behavior to make `real` and `imag` writable properties. Note that `x.real` may create a copy so you need a setter to make it effectively writable.

odl/discr/lp_discr.py
kohr-h: Hm, agreed.

odl/space/cu_ntuples.py
kohr-h: I'll check. I *think* it's also present in `DiscreteLp`, but it would actually make sense in `Discretization` directly.

test/largescale/trafos/fourier_slow_test.py
kohr-h: ok

test/largescale/trafos/fourier_slow_test.py
kohr-h: Probably [like this](http://pytest.org/latest/skipping.html#skip-xfail-with-parametrize).

odl/discr/lp_discr.py
kohr-h: Second thought, the error raised in `fspace` is a `ValueError` anyway and will propagate through, so I'll just delete this part. Alternatively, I could add `ValueError` and append the error message just as in the other two cases.

odl/discr/lp_discr.py
kohr-h: Second thought also here: `astype` becomes quite difficult to handle if you include `impl`. Then you also need to translate the weighting if you switch implementation, and we don't have an easy way of doing this. 
Implementing, in a first step, only `astype(dtype)` is easier, though. I'll do that first, and when we really need the `impl` option, we can make a new issue.

kohr-h: But then, the `astype` is trivial, at least for `FnBase` spaces:
```python
def astype(self, dtype):
    return type(self)(self.size, dtype=dtype, weight=self.weighting)
```
It's more complicated for, e.g., `DiscreteLp` so it may well be worth a method. On the other hand, the nice thing with the real <-> complex translation is that you don't have to manually type all those if's and else's for the 4 possible cases (r2r, r2c, c2r, c2c) in all situations it pops up. I'd argue that this is worth considering as an addition (when you deal with Fourier transforms, these conversions are *everywhere*).

odl/discr/lp_discr.py
kohr-h: Caching now, but it's troublesome to do it during init. For example, in `FnBase`, the weighting is not yet defined when needed. I decided to store the trivial one (`self`) immediately and the calculated counterpart in the first invocation of the corresponding method.

odl/discr/lp_discr.py
kohr-h: Done.

odl/space/cu_ntuples.py
kohr-h: Done that. One slight remark about this is that `FnBase` does not guarantee existence of a `weighting` attribute, even if both its implementations do. I wrote the properties with `gettattr` and default `None` and `False`, resp.

test/largescale/trafos/fourier_slow_test.py
kohr-h: I implemented the conditional skips with parametrized tests. However, there seems to be a bug in pytest since the global `skip_if_no_largescale` is ignored if I try to parametrize the test here in the same way as in `fourier_test.py`. The unconditional `'numpy'` case is skipped but the conditional `'pyfftw'` one runs anyway.

kohr-h:
Any more comments on this? What about the real/complex space thing? See comments above.

odl/discr/lp_discr.py
adler-j: How about we simply rename this to astype and give it only a dtype parameter? That should do for now.

kohr-h:
Implementing the `astype` thing currently. One question: do we want to allow this:
```
>>> discr = odl.uniform_discr(0, 1, 10, dtype=complex)
>>> x = discr.element()
>>> x.real = 0  # error??
```

adler-j:
Why would that give an error?

kohr-h:
Because we cannot create an `FnBase` element from a single number:
```
unable to create an element of uniform_discr(0.0, 1.0, 10) from 0: input shape (1,)
not broadcastable to shape (10,).
```

kohr-h:
After piping the input to `real_space.element()` I should add.

kohr-h:
Now we have `astype` for `FnBase`, `DiscreteLp` and `FuctionSpace`. Implemented also `real` and `imag` for `FunctionSpaceVector` while I was at it.

kohr-h:
I think all comments are implemented how about merging? Maybe tomorrow, two releases the same day is not so great :-)

kohr-h:
Ping

adler-j:
Was OK last I checked. Go ahead! Also bump version to 0.2 and make release notes, perhaps on this branch already?

kohr-h:
Sure. Do you mean a file in the docs? We could have a link on the online doc pointing to a cumulative release notes site.

adler-j:
Not quite sure, I simply added some on GitHub but we should perhaps have them with the code, perhaps have a look at how numpy does it?

kohr-h:
They have files lying around rather prominently where contributors are asked to put a note on changes they made (as part of pull requests). I'm just writing up a file for our doc tree. I guess we can live with a slightly more "hidden" file and point our contributors to that one. Having a running document seems like a good idea in general, I'd say.

PR #124, opened 2015-12-07
ENH: made Rn, Cn and CudaRn classes factories

adler-j:
Rather simple pull, removes 3 classes. Add `is_rn` and `is_cn` to general `FnBase` spaces to replace the old `isinstance(x, Rn)` functionality.


adler-j:
Is this pullable?

kohr-h:
Well, apart from the slightly decreased coverage I think we can pull. The naming issue of class factories is broader than this pull request, let's take that separately.

PR #126, opened 2015-12-07
ENH: made field a property of LinearSpace

adler-j:
Removed the condition that `LinearSpace`'s inplement `field` property, instead call `LinearSpace.__init__` for it to be initialized.


adler-j:
Quite a trivial branch, ready to pull.

PR #148, opened 2015-12-11
ENH: add ufuncs for pspace

adler-j:
Rather trivial branch.


kohr-h:
Can you use the `out` parameter by keyword in the tests? Otherwise it's hard to understand what `x.ufunc.absolute(y)` is supposed to do.

adler-j:
Added the suggested features as well as doctests, will merge after travis is happy.

Issue #131, opened 2015-12-08
Forward and backward differences in finite_diff


moosmann:
Implementation of forward and backward difference is needed for helper function `finite_diff` in module `odl.discr.discr_ops`.

PR #149, opened 2015-12-11
Issue 131  forward diff

adler-j:
In response to issue #131, should be OK.

@moosmann?

### DIFF
--- a/odl/discr/discr_ops.py
+++ b/odl/discr/discr_ops.py
@@ -67,7 +67,7 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
         Implicit zero padding. Assumes values outside the domain of ``f`` to be
         zero. Default: `False`
     method : {'central', 'forward', 'backward'}, optional
-        What method that should be used.
+        The method that should be used.
 
     Returns
     -------
@@ -127,9 +127,10 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
     else:
         dx = float(dx)
 
-    method = str(method).lower()
+    method, method_in = str(method).lower(), method
     if method not in ('central', 'forward', 'backward'):
-        raise ValueError('method {} has to be central, forward or backward')
+        raise ValueError("method '{}' has to be central, forward or backward"
+                         ''.format(method_in))
 
     # create slice objects: initially all are [:, :, ..., :]
     # current slice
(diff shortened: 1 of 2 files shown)


odl/discr/discr_ops.py
```diff
@@ -66,6 +66,8 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
     zero_padding : `bool`, optional
         Implicit zero padding. Assumes values outside the domain of ``f`` to be
         zero. Default: `False`
+    method : {'central', 'forward', 'backward'}, optional
+        What method that should be used.
```
kohr-h: Bad English.

odl/discr/discr_ops.py
```diff
@@ -126,6 +127,10 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
     else:
         dx = float(dx)
 
+    method = str(method).lower()
+    if method not in ('central', 'forward', 'backward'):
+        raise ValueError('method {} has to be central, forward or backward')
```
kohr-h: This will print the curly brackets literally. Print the original argument, not the sanitized one. And use quotes in the string since the possible values are strings.

test/discr/discr_ops_test.py
```diff
@@ -63,6 +63,9 @@ def test_finite_diff():
         finite_diff(arr, out)
     with pytest.raises(ValueError):
         finite_diff(arr, dx=0)
+    # wribg method
```
kohr-h: ??

kohr-h:
Okay for now. Fix the small issues, after that it can be merged.

odl/discr/discr_ops.py
```diff
@@ -134,30 +139,71 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
...
+    if method == 'central':
+        # Numerical differentiation: 2nd order interior
+        slice_out[axis] = slice(1, -1)
+        slice_node1[axis] = slice(2, None)
+        slice_node2[axis] = slice(None, -2)
+        # 1D equivalent: out[1:-1] = (f[2:] - f[:-2])/2.0
+        np.subtract(f_data[slice_node1], f_data[slice_node2], out[slice_out])
+        out[slice_out] /= 2.0
+    elif method == 'forward':
+        # Numerical differentiation: 2nd order interior
```
moosmann: Should be 1st order accurate differences.

odl/discr/discr_ops.py
```diff
@@ -134,30 +139,71 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
...
+        slice_out[axis] = slice(1, -1)
+        slice_node1[axis] = slice(2, None)
+        slice_node2[axis] = slice(1, -1)
+        # 1D equivalent: out[1:-1] = (f[2:] - f[1:-1])
+        np.subtract(f_data[slice_node1], f_data[slice_node2], out[slice_out])
+    elif method == 'backward':
+        # Numerical differentiation: 2nd order interior
```
moosmann: 1st order accurate differences.

odl/discr/discr_ops.py
```diff
@@ -134,30 +139,71 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
...
+        slice_out[axis] = slice(1, -1)
+        slice_node1[axis] = slice(1, -1)
+        slice_node2[axis] = slice(None, -2)
+        # 1D equivalent: out[1:-1] = (f[1:-1] - f[:-2])
+        np.subtract(f_data[slice_node1], f_data[slice_node2], out[slice_out])
 
     # central differences
```
moosmann: Should be removed.

adler-j:
Updated according to feedback, merging after travis.

odl/discr/discr_ops.py
```diff
@@ -134,30 +139,71 @@ def finite_diff(f, out=None, axis=0, dx=1.0, edge_order=2,
...
+            slice_node1[axis] = -1
+            slice_node2[axis] = -2
+            # 1D equivalent: out[-1] = (f[-1] - f[-2])
+            out[slice_out] = f_data[slice_node1] - f_data[slice_node2]
 
     # one-sided differences
     else:
+        if method != 'central':
+            raise NotImplementedError('Can only use zero padding with non-'
+                                      'central differences')
```
moosmann: Shouldn't we raise a `ValueError` for `edge_order=2` and `method in ('forward','backward')` instead? Thus allowing to use forward and backward differences (in the interior) also without zero padding similar to when using central differences.

adler-j: Technically we can allow both, the word "central" doesnt really hold at the edges either, I'll let this parameter "override" that keyword.

moosmann: We could also make the arguments a bit more structured i.e. instead of `zero_padding` using `boundary_method` with `zero_padding` or `one-sided` as options?

PR #160, opened 2015-12-14
Issue155  1dwavelet transform

niinimaki:
Implement 1D wavelet transforms, update the helper functions and the tests accordingly. 
Removed redundant inputs from 3D helper functions.

### DIFF
--- a/odl/trafos/wavelet.py
+++ b/odl/trafos/wavelet.py
@@ -46,14 +46,14 @@ _SUPPORTED_IMPL = ('pywt',)
 def coeff_size_list(shape, nscales, wbasis, mode):
     """Construct a size list from given wavelet coefficients.
 
-    Related to 2D and 3D multidimensional wavelet transforms that utilize
+    Related to 1D, 2D and 3D multidimensional wavelet transforms that utilize
     `PyWavelets
     <http://www.pybytes.com/pywavelets/>`_.
 
     Parameters
     ----------
     shape : `tuple`
-        Number of pixels/voxels in the image. Its length must be 2 or 3.
+        Number of pixels/voxels in the image. Its length must be 1, 2 or 3.
     nscales : `int`
         Number of scales in the multidimensional wavelet
         transform.  This parameter is checked against the maximum number of
@@ -105,8 +105,8 @@ dwt-discrete-wavelet-transform.html#maximum-decomposition-level\
 
         ``N`` = number of scaling levels = nscales
     """
-    if len(shape) not in (2, 3):
-        raise ValueError('shape must have length 2 or 3, got {}.'
+    if len(shape) not in (1, 2, 3):
+        raise ValueError('shape must have length 1, 2 or 3, got {}.'
                          ''.format(len(shape)))
 
     max_level = pywt.dwt_max_level(shape[0], filter_len=wbasis.dec_len)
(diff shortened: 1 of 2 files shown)


odl/trafos/wavelet.py
```diff
@@ -136,6 +136,133 @@ def coeff_size_list(shape, nscales, wbasis, mode):
...
+
+
+def array_to_pywt_coeff1d(coeff, size_list):
+    """Convert a flat array into a `pywt
+    <http://www.pybytes.com/pywavelets/>`_ coefficient list.
+    For multilevel 1D discrete wavelet transform
+
+    Parameters
+    ----------
+    coeff : `DiscreteLp.Vector`
```
adler-j: DiscreteLpVector

niinimaki:
Merged helper functions `array_to_pywt_coeff1d`, `array_to_pywt_coeff2d` and `array_to_pywt_coeff3d` into `array_to_pywt_coeff`. Similarly for `pywt_coeff_to_array` (`pywt_coeff_to_array1d` still remains).
Updated tests accordingly to the changes made in the helper functions.

niinimaki:
Merged helper function `pywt_coeff_to_array1d` into  `pywt_coeff_to_array`.
Updated tests accordingly to the changes made in the helper functions.

odl/trafos/wavelet.py
```diff
@@ -136,123 +136,26 @@ def coeff_size_list(shape, nscales, wbasis, mode):
...
+def pywt_coeff_to_array(coeff, size_list):
     """Convert a `pywt
     <http://www.pybytes.com/pywavelets/>`_ coefficient into a flat array.
 
-    Related to 2D multilevel discrete wavelet transform.
+    Related to 2D and 3D multilevel discrete wavelet transform.
 
     Parameters
     ----------
+    1D:
```
kohr-h: You need to structure this a bit differently due to the way the doc parser reads the text. It will interpret `1D` as a parameter, so you cannot insert extra headers between the parameters. 

Instead make sections in the text under `coeff`.

odl/trafos/wavelet.py
```diff
@@ -344,36 +217,43 @@ def pywt_coeff_to_array3d(coeff, size_list, nscales):
...
         be transformed and on the chosen wavelet basis.
-        If the size of the input image is 2^n x 2^n, the length of the
-        wavelet coefficient array is the same.
-    """
-    size_flatCoeff = np.prod(size_list[0])
-    for kk in range(1, nscales + 1):
-        size_flatCoeff += 7 * np.prod(size_list[kk])
+      """
+    flat_sizes = [np.prod(shp) for shp in size_list[:-1]]
+    if np.size(size_list[0]) == 1:
```
kohr-h: Just use `len`, it's a built-in function.

odl/trafos/wavelet.py
```diff
@@ -344,36 +217,43 @@ def pywt_coeff_to_array3d(coeff, size_list, nscales):
...
+        const = 1
+    if np.size(size_list[0]) == 2:
```
kohr-h: `elif`

odl/trafos/wavelet.py
```diff
@@ -344,36 +217,43 @@ def pywt_coeff_to_array3d(coeff, size_list, nscales):
...
+        const = 3
+    elif np.size(size_list[0]) == 3:
```
kohr-h: Use `else` as the last case.

odl/trafos/wavelet.py
```diff
@@ -344,36 +217,43 @@ def pywt_coeff_to_array3d(coeff, size_list, nscales):
...
-        If the size of the input image is 2^n x 2^n, the length of the
-        wavelet coefficient array is the same.
-    """
-    size_flatCoeff = np.prod(size_list[0])
-    for kk in range(1, nscales + 1):
-        size_flatCoeff += 7 * np.prod(size_list[kk])
+      """
+    flat_sizes = [np.prod(shp) for shp in size_list[:-1]]
+    if np.size(size_list[0]) == 1:
+        const = 1
```
kohr-h: `const` is only used once later, so you can give it a really nice name, like `dcoeffs_per_scale`.

odl/trafos/wavelet.py
```diff
@@ -648,14 +560,17 @@ def __init__(self, dom, nscales, wbasis, mode):
 
         ran_size = np.prod(self.size_list[0])
 
-        if len(dom.grid.shape) == 2:
+        if len(dom.grid.shape) == 1:
```
kohr-h: `dom.ndim`

odl/trafos/wavelet.py
```diff
@@ -648,14 +560,17 @@ def __init__(self, dom, nscales, wbasis, mode):
...
+                            self.size_list[1:-1])
+        elif len(dom.grid.shape) == 2:
             ran_size += sum(3 * np.prod(shape) for shape in
                             self.size_list[1:-1])
         elif len(dom.grid.shape) == 3:
             ran_size += sum(7 * np.prod(shape) for shape in
                             self.size_list[1:-1])
         else:
-            raise NotImplementedError('ndim {} not 2 or 3'
+            raise NotImplementedError('ndim {} not 1, 2 or 3'
```
kohr-h: `dom.ndim`

odl/trafos/wavelet.py
```diff
@@ -685,16 +600,20 @@ def _call(self, x):
             The length of the array depends on the size of input image to
             be transformed and on the chosen wavelet basis.
         """
+        if len(x.shape) == 1:
```
kohr-h: `x.space.ndim`

odl/trafos/wavelet.py
```diff
@@ -767,14 +683,17 @@ def __init__(self, ran, nscales, wbasis, mode):
                                          self.wbasis, self.mode)
 
         dom_size = np.prod(self.size_list[0])
-        if len(ran.grid.shape) == 2:
+        if len(ran.grid.shape) == 1:
```
kohr-h: `ran.ndim`

odl/trafos/wavelet.py
```diff
@@ -767,14 +683,17 @@ def __init__(self, ran, nscales, wbasis, mode):
...
+                            self.size_list[1:-1])
+        elif len(ran.grid.shape) == 2:
             dom_size += sum(3 * np.prod(shape) for shape in
                             self.size_list[1:-1])
         elif len(ran.grid.shape) == 3:
             dom_size += sum(7 * np.prod(shape) for shape in
                             self.size_list[1:-1])
         else:
-            raise NotImplementedError('ndim {} not 2 or 3'
+            raise NotImplementedError('ndim {} not 1, 2 or 3'
```
kohr-h: `ran.ndim`

kohr-h:
A couple of *very* minor comments, mostly pointers to more adequate properties and style things. Generally this is a very nice simplification, and we should pull when you're done with the amendments.

odl/trafos/wavelet.py
adler-j: unless there is a strong guarantee here, failing at else is also an option.

niinimaki: Would something of the sort
`else:
        raise NotImplementedError('ndim {} not 1, 2 or 3'
                                  ''.format(len(size_list[0])))`

do? Since only 1d, 2d and 3d are implemented...

kohr-h: Yes, except that you need to be more explicit than `ndim` because it does not appear anywhere, so the message would be rather cryptic to the user. Maybe `"{}-dimensional conversion not supported..format(len(size_list[0]))"`. And since it appears already 4 times, it's worth giving it a name, e.g. `ndim` :-)

niinimaki: Actually, going to get rid of these (`if, elif, elif, else`) checks (from both helper funcs) since can be done more compactly by

`ndim = len(size_list[0])`
`dcoeffs_per_scale = 2**ndim - 1`

kohr-h: Well noted!

niinimaki:
Implemented latest comments.

odl/trafos/wavelet.py
```diff
@@ -344,36 +215,39 @@ def pywt_coeff_to_array3d(coeff, size_list, nscales):
...
-        If the size of the input image is 2^n x 2^n, the length of the
-        wavelet coefficient array is the same.
-    """
-    size_flatCoeff = np.prod(size_list[0])
-    for kk in range(1, nscales + 1):
-        size_flatCoeff += 7 * np.prod(size_list[kk])
+      """
+    flat_sizes = [np.prod(shp) for shp in size_list[:-1]]
+    ndim = len(size_list[0])
+    dcoeffs_per_scale = 2**ndim - 1
```
kohr-h: Here our angry PEP8 checker will complain. `2 ** ndim - 1`.

kohr-h:
Looks good, merge. We look into the test running times a bit closer later.

Issue #10, opened 2015-11-11
Add possibility to refresh same figure in DiscreteLp.Vector.show


adler-j:
Currently the following gives you 50 figures:

```python
for i in range(50):
    #do stuff
    x.show()
```

a syntax like

```python
fig=None
for i in range(50):
    #do stuff
    fig=x.show(figure=fig)
```

or possibly something better, would be very nice.

kohr-h:
I'll see what I can do. However, `matplotlib` has its quirks, not trivial to get this running smoothly.

adler-j:
Don't waste to much time on it then, we can wait.

moosmann:
The only thing that worked for me to refresh a figure was to use `matplotlib.pyplot.pause` instead of `matplotlib.pyplot.show`. However, it's an experimental function and probably won't work for all backends.

PR #163, opened 2015-12-15
Issue 10  show same fig

adler-j:
Ancient issue (#10), should be solved by this pull request. Syntax is very simple:

```python
# Create figure
fig = phantom.show()

# Reuse figure
phantom.show(fig=fig)
```

The current syntax overwrites the old plot values so you can get plots like this:

```python
fig = phantom.show(indices=np.s_[:, n//2])

def fun(x):
    x.show(indices=np.s_[:, n//2], fig=fig)
partial = odl.solvers.util.ForEachPartial(fun)

odl.solvers.landweber(A, d.zero(), rhs, 10, 0.05, partial)
```

plt

updated continuously, making it easy to watch progress of your algorithm. As a side note, the "partial" syntax clearly needs to be improved, we could shave of a bit of complexity here.


kohr-h:
Nice, I overlooked that the figure has an `axes` attribute when I tried this. Can you update the summary line in the docstring also? Safe to merge after that I guess.

Issue #166, opened 2015-12-16
Improve wavelet transform


niinimaki:
A couple of improvements to the wavelet transform:
The docstrings of wavelet transform class could be more informative (for example the signal extension modes are not described there, only in the docstrings of  helper functions).
Fix linearity check.

adler-j:
:+1: sounds good. Btw, you could make a checklist like this:
```
- [ ] The docstrings of wavelet transform class could be more informative (for example the signal extension modes are not described there, only in the docstrings of helper functions).
- [ ] Fix linearity check.
```

- [ ] The docstrings of wavelet transform class could be more informative (for example the signal extension modes are not described there, only in the docstrings of helper functions).
- [ ] Fix linearity check.

PR #167, opened 2015-12-16
 Improve wavelet transform #166 

niinimaki:
Made the improvements of the docstrings and the fix of linearity check.


adler-j:
All good, pulled.

Issue #173, opened 2015-12-17
Diagonal operator


wjp:
It would be nice to have an operator for multiplication with a diagonal matrix.

Currently it can be written as

    y = some_vector
    D = odl.OperatorLeftVectorMult ( odl.IdentityOperator(y.space), y )

But this is rather unwieldy.

adler-j:
You could also do

```
y = some_vector
D = y * odl.IdentityOperator(y.space)
```

But I agree, we need this. Ill add something.

adler-j:
Now added, syntax is

```python
y = some_vector
D = odl.MultiplyOperator(y)
```

PR #176, opened 2015-12-17
Issue 173  diagonal op

adler-j:
In reference to #173, changes the behavior of `MultiplyOperator` to be similar to that of `InnerProductOperator`, which is likely the expected. Also allows `MultiplyOperator` the be overloaded to mean multiplication with vector or scalar, this allows the removal of the clumsy `InnerProductAdjointOperator`

### DIFF
--- a/odl/operator/default_ops.py
+++ b/odl/operator/default_ops.py
@@ -330,28 +330,56 @@ class MultiplyOperator(Operator):
 
         >>> op = MultiplyOperator(x)
         >>> out = r3.element()
-        >>> op(x)
+        >>> op(x, out)
         Rn(3).element([1.0, 4.0, 9.0])
 
         Multiply by scalar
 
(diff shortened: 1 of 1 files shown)


odl/operator/default_ops.py
```diff
@@ -294,20 +324,66 @@ def _apply(self, x, out):
...
+        else:
+            out.multiply(x, self.y)
+
+    @property
+    def adjoint(self):
+        """ The adjoint operator.
+
+        Returns
+        -------
+        adjoint : `InnerProductOperator` or `MultiplyOperator`
```
kohr-h: For two choices `a` and `b` we write `{a, b}`, not `a or b`

kohr-h:
Much better, this would actually be expected from a multiplication operator. Without looking closely, if the tests run (and all branches have a test), merge.

Issue #170, opened 2015-12-16
Examples for show()


adler-j:
We have many methods in show, scatter, imshow, real, complex, etc. Is there some example for all of these laying around? Would be nice to have a "test" if you change something.

kohr-h:
No, those have not been tested in a while. I checked all of them in the initial implementation, but having an example lying around which checks those would of course be better.

adler-j:
If you have them somewhere, could you add them to examples?

kohr-h:
No, but I can write some simple ones.

Issue #168, opened 2015-12-16
Optional plt.show in show method


moosmann:
Unless IPython/Spyder or a non-interactive backend (or probably something else) is used, calling `plt.show` in `odl.util.graphics.show_discrete_function()` blocks the execution of subsequent code until the window which has popped up is closed. Calling `plt.show` should be made optional, thus that one can call `plt.show` explicitly at the end of the script to display the previously created images.

adler-j:
Do you have the latest version? The "show" is now blocking=False, does that help?

moosmann:
The block=False option never worked properly for me. I already tried it several times. When I'm using
```
plt.show(block=False)
```
or
```
plt.show(block=False)
plt.draw()
```
then empty figures pop up which when using Pycharm do not respond anymore and require a force quit or which close immediately when executing a script directly. If I use
```
plt.show(block=False)
plt.draw()
plt.pause(0.1)
```
then windows pop up and the figures are displayed properly, but just for the time specified in plt.pause. 

There's also the option to enable interactive mode with plt.ion() which, however, never worked for me. I think it would easiest to allow for an additional argument to call plt.show or not. This at least works for me without problems.  Using interactive plotting one can in general not guarantee that plotting works as expected, see [matplotlib interactive mode](http://matplotlib.org/faq/usage_faq.html?highlight=usage#what-is-interactive-mode). The situation also differs if you are displaying a graph with plt.plot or an image with plt.imshow: the examples in [matplotlib  interactive mode](http://matplotlib.org/faq/usage_faq.html?highlight=usage#interactive-example) only worked for me as described when using plt.plot but not with plt.imshow.

adler-j:
Ok, I guess we will need to do some thinking on the interactive plots, they are very useful in some situations.

kohr-h:
Definitely! I suspect that global commands like `ion` don't help, and in my experience any of those global commands like `plt.figure` or global `axis` instances cause problems when one tries to change stuff. Maybe it is possible to do something similar locally, via class methods on the objects which are created along the way, in our case `figure`, which can be reused. We need to have a closer look.

niinimaki:
Calling plt.show() fails
`TypeError: show() got an unexpected keyword argument 'block'`

adler-j:
What version of matplotlib do you use? This keyword should exist.

niinimaki:
Seems that I have version 1.3.1 ... too old?

adler-j:
[...]

PR #187, opened 2015-12-21
Issue 168  show

adler-j:
Several updates to the visualization ecosystem, now all examples work "as expected" at my local machine. This should close both #168 and #170 (except for the "scatter" and complex examples).

### DIFF
--- a/examples/visualization/show_1d.py
+++ b/examples/visualization/show_1d.py
@@ -33,7 +33,7 @@ vec = spc.element(np.sin(spc.points()))
 vec.show()
 (vec * 2).show()
 
-# Plotting is defered untill show() is called
+# Plotting is deferred until show() is called
 plt.show()
 
 # Can also force "instant" plotting
--- a/examples/visualization/show_1d.py
+++ b/examples/visualization/show_1d.py
@@ -1,0 +1,40 @@
+# Copyright 2014, 2015 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Examples on using the vector.show() syntax
+
+NOTES
+-----
+The behaviour of blocking shows etc in matplotlib is experimental and can cause
+issues with these examples.
+"""
+
+import matplotlib.pyplot as plt
+import odl
+import numpy as np
+
+spc = odl.uniform_discr(0, 5, 100)
+vec = spc.element(np.sin(spc.points()))
+
+vec.show()
+(vec * 2).show()
+
+# Plotting is defered untill show() is called
+plt.show()
+
+# Can also force "instant" plotting
+vec.show(show=True)
--- a/examples/visualization/show_productspace.py
+++ b/examples/visualization/show_productspace.py
@@ -1,0 +1,42 @@
+# Copyright 2014, 2015 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Example on using show with ProductSpace's """
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+# Internal
+import odl
+
+n = 100
+m = 10
+spc = odl.uniform_discr([0, 0], [1, 1], [n, n])
+pspace = odl.ProductSpace(spc, m)
+
+vec = pspace.element([odl.util.shepp_logan(spc)] * m)
+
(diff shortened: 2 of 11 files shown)


examples/visualization/show_1d.py
```diff
@@ -0,0 +1,40 @@
...
+import odl
+import numpy as np
+
+spc = odl.uniform_discr(0, 5, 100)
+vec = spc.element(np.sin(spc.points()))
+
+vec.show()
+(vec * 2).show()
+
+# Plotting is defered untill show() is called
```
kohr-h: deferred, until

examples/visualization/show_productspace.py
```diff
@@ -0,0 +1,48 @@
...
+""" Example on using show with ProductSpace's """
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+import odl
+import numpy as np
+
+n = 100
```
kohr-h: We're gathering tons of Pylint warnings with these names. I don't think names like `n` and `m` are necessarily bad, but here at least the `m` is a bit uncreative and lazy. A better name would be `nfactors`.

Generally, since we have legit uses of `n` and `m`, why don't we add them to the `.pylintrc` as accepted names?

odl/discr/lp_discr.py
```diff
@@ -148,8 +148,10 @@ def element(self, inp=None):
         # Sequence-type input
         arr = np.asarray(inp, dtype=self.dtype, order=self.order)
         if arr.ndim > 1 and arr.shape != self.shape:
-            raise ValueError('input shape {} does not match grid shape {}'
-                             ''.format(arr.shape, self.shape))
+            arr = np.squeeze(arr)  # Squeeze could solve the problem
```
kohr-h: :+1:

examples/visualization/show_productspace.py
adler-j: >We're gathering tons of Pylint warnings with these names. I don't think names like n and m are necessarily bad, but here at least the m is a bit uncreative and lazy. A better name would be nfactors.

I agree.

>Generally, since we have legit uses of n and m, why don't we add them to the .pylintrc as accepted names?

[The list](https://github.com/odlgroup/odl/blob/e6fd57615387d0b8d7634876b7e12b16934cccd7/.pylintrc#L124) is huge, adding a few more wouldnt hurt :)

odl/discr/lp_discr.py
```diff
@@ -433,6 +436,10 @@ def show(self, method='', title='', indices=None, fig=None, **kwargs):
         title : `str`, optional
             Set the title of the figure
 
+        show : `bool`, optional
+            If the plot should be showed now or defered untill later.
```
kohr-h: deferred, until

odl/set/pspace.py
```diff
@@ -617,13 +617,24 @@ def show(self, *args, **kwargs):
...
+            else:
+                indices = np.linspace(0, self.size - 1, 4, dtype=int)
         else:
-            show_parts = self.parts[:2] + self.parts[-2:]
-            indices = [0, 1, len(self) - 2, len(self) - 1]
-        for i, part in zip(indices, show_parts):
+            if isinstance(indices, tuple):
+                indices, kwargs['indices'] = indices[0], indices[1:]
+
+            if isinstance(indices, slice):
```
kohr-h: elif

odl/util/graphics.py
```diff
@@ -71,6 +71,10 @@ def show_discrete_function(dfunc, method='', title=None, indices=None,
         two axes at the "middle" along the remaining axes is shown
         (semantically ``[:, :, shape[2:] // 2]``).
 
+    show : `bool`, optional
+        If the plot should be showed now or defered untill later.
```
kohr-h: see above

kohr-h:
It took me a while to understand what happened in the default case if `indices` is `None` since the example shows the first three and then the last, skipping just the fourth (I thought it was a bug). Choosing 7 parts and then showing 0, 2, 4, 6 would be easier to understand.

Otherwise it looks good on my machine.

odl/set/pspace.py
adler-j: Actually no, consider `indices = (slice(None), 2)`, which is valid and here we expect to unpack the first one and do stuff with it. There should be an elif on the next if however.

adler-j:
Fixed according to comments, pull?

odl/set/pspace.py
kohr-h: You're right.

examples/visualization/show_update_1d.py
```diff
@@ -0,0 +1,41 @@
...
+n = 100
+m = 20
+spc = odl.uniform_discr(0, 5, n)
+vec = spc.element(np.sin(spc.points()))
+
+# Pre-create a plot and set some property
+fig = plt.figure()
+plt.ylim(-m, m)
+
+# Reuse the figure indefinately
```
kohr-h: indefinitely

examples/visualization/show_update_2d.py
```diff
@@ -0,0 +1,39 @@
...
+
+n = 100
+m = 20
+spc = odl.uniform_discr([0, 0], [1, 1], [n, n])
+vec = odl.util.shepp_logan(spc)
+
+# Create a figure by saving the result of show
+fig = None
+
+# Reuse the figure indefinately, values are overwritten.
```
kohr-h: see above

odl/util/graphics.py
```diff
@@ -71,6 +71,9 @@ def show_discrete_function(dfunc, method='', title=None, indices=None,
         two axes at the "middle" along the remaining axes is shown
         (semantically ``[:, :, shape[2:] // 2]``).
 
+    show : `bool`, optional
+        If the plot should be showed now or deferred until later.
```
kohr-h: No punctuation at the end of a one-liner

kohr-h:
Some tiny remaining things, pull when you're done.

examples/visualization/show_update_1d.py
adler-j: My English needs some work, kinda shameful this. Damn spellcheck in all tools make you forget how to spell.

odlpp
```diff
@@ -1 +1 @@
-Subproject commit d3352e6e1d7e5a505654272e880c5fea3995b29e
+Subproject commit 7f9e2b793abb588df52ac331f1a776d008caf8e7
```
adler-j: Not sure about this, I'll make sure it is correct.

examples/visualization/show_update_1d.py
kohr-h: Sort of my only possibility for suggested changes :-)

odlpp
kohr-h: Right, I forgot to mention that the build with odlpp failed. It points to the pybind11 branch probably.

Issue #159, opened 2015-12-13
Interpolation scheme per axis


kohr-h:
The interpolation flag in `DiscreteLp` and the convenience functions are global. We should make it possible to define flags per axis by accepting a sequence of flags. The implementation of this should not be too hard.

adler-j:
Agree!

kohr-h:
Done by #190. Integration into `DiscreteLp` is a separate issue #201.

Issue #158, opened 2015-12-13
More interpolation schemes


kohr-h:
There's only nearest neighbor interpolation properly implemented, and of that only one variant. We need some more, like linear interpolation and cubic splines.

PR #190, opened 2015-12-26
Issue 158  interpolation

kohr-h:
This PR simplifies the interpolator structures significantly by using one class for both arrays and meshgrids and specifying the variant as init argument instead. The principle is to keep the private interpolator classes simple and implement the decision logic in the callers (the interpolation operators).

Secondly, linear interpolation (#158) now works as specified in our internal discretization schemes. This means that all interpolators assume that the next "virtual" node outside the region has value zero. Hence, the value at the boundary is half the value at the last node in the region for linear interpolation. Proper boundary values are yet to be implemented.

Thirdly, interpolation schemes can now be specified per axis with the `PerAxisInterpolation` operator (#159). In fact, linear interpolation is just a special case of this operator. Nearest neighbor, however, has its own implementation for performance reasons (about 3x faster for arrays, about 2^(dim+1) times for meshes).

All new features and some old ones come with new tests.

Remaining TODOs: make new issues for
- [x] Cubic spline interpolation
- [x] Boundary values (grids with "half basis functions" at the ends), including periodic interpolation
- [x] Add per-axis support to `DiscreteLp` and adapt norm, inner product etc.
- [x] proper tessellations 

EDIT: new issues for the checklist points


adler-j:
> Cubic spline interpolation -> new issue?

New issue

> Boundary values (grids with "half basis functions" at the ends)

Also new issue

> Periodic interpolation

Also new issue, perhaps same as "Boundary values" since they are very related.

> Add per-axis support to DiscreteLp
 
New issue
 
> Adapt norm, inner product etc.

Same issue as above

> New issue: proper tessellations

:+1:

I'll try to give you a review of this.

.pylintrc
```diff
@@ -42,11 +42,15 @@ load-plugins=
 # Disable all messages caused solely by 'future' imports and functions
 # and also backslash to enable :math: and not callable since it fails with inherited __call__.
 disable=redefined-builtin,
-  anomalous-backslash-in-string
+  anomalous-backslash-in-string,
```
adler-j: Why these in such a branch?

.pylintrc
```diff
@@ -121,7 +125,7 @@ indent-after-paren=4
 bad-functions=map,filter,apply,input
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_,a,b,x,y,z,xp,yp,zp,s,q,p,I,T,dx,dy,x1,x2
+good-names=i,j,k,ex,Run,_,a,b,x,y,z,xp,yp,zp,s,q,p,I,T,dx,dy,x1,x2,xi,yi,n,m,ax
```
adler-j: Wrong branch but we'll let it slide.

odl/discr/discr_mappings.py
```diff
@@ -15,19 +15,22 @@
...
+operators.
+"""
 
 # Imports for common Python 2/3 codebase
 from __future__ import print_function, division, absolute_import
 
 from future import standard_library
 standard_library.install_aliases()
-from builtins import str, zip
+from builtins import str, super, zip
```
adler-j: We kindoff decided to be careful with super for performance reasons (slow as hell in the python2 backport).

odl/discr/discr_mappings.py
```diff
@@ -35,17 +38,22 @@
...
+from odl.util.vectorization import (is_valid_input_meshgrid,
+                                    out_shape_from_meshgrid)
 
 
 __all__ = ('FunctionSetMapping',
-           'GridCollocation', 'NearestInterpolation', 'LinearInterpolation')
+           'GridCollocation', 'NearestInterpolation', 'LinearInterpolation',
+           'PerAxisInterpolation')
+
+SUPPORTED_INTERP_SCHEMES = ['nearest', 'linear']
```
adler-j: Perhaps if this is internal it should have a `_`?

odl/discr/discr_mappings.py
```diff
@@ -35,17 +38,22 @@
...
 
 
 class FunctionSetMapping(Operator):
 
     """Abstract base class for function set discretization mappings."""
 
+    # pylint: disable=abstract-method
```
adler-j: Sceptical to having these in to many places. Any chance to modify this externally somehow?

odl/discr/discr_mappings.py
```diff
@@ -353,14 +372,14 @@ def _call(self, x, out=None):
 
         Examples
         --------
-        >>> from __future__ import unicode_literals, print_function
-        >>> from odl import Rectangle, Strings
+        We test nearest neighbor interpolation with a non-scalar
+        data type in 2d:
+
+        >>> from __future__ import print_function
```
adler-j: No way to get around this?

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
         def linear(arg, out=None):
             """Interpolating function with vectorization."""
             if is_valid_input_meshgrid(arg, self.grid.ndim):
-                # TODO: check if this works for 'F' ordering
-                interp = _LinearMeshgridInterpolator(
+                interp = _LinearInterpolator(
                     self.grid.coord_vectors,
-                    x.data.reshape(self.grid.shape, order=self.order))
+                    x.data.reshape(self.grid.shape, order=self.order),
+                    typ='meshgrid')
```
adler-j: `typ`, dunno what i think about this name. How about `grid` or `layout`?

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
             else:
-                # TODO: check if points are ordered. If not, use
-                # LinearNDInterpolator
-                interp = _LinearPointwiseInterpolator(
+                interp = _LinearInterpolator(
```
adler-j: Tbh writing out `interpolator` should be done here.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+    def __init__(self, ip_fspace, grid, dspace, schemes,
+                 nn_variants=None, order='C'):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        fspace : `FunctionSpace`
+            The undiscretized (abstract) space of functions to be
+            discretized. Its field must be the same as that of data
+            space. Its `Operator.domain` must be an
```
adler-j: `FunctionSpace.domain` (if it does not exist, we should probably add it)

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            `IntervalProd`.
+        grid :  `TensorGrid`
+            The grid on which to evaluate. Must be contained in
+            the domain of the function set.
+        dspace : `FnBase`
+            Data space providing containers for the values of a
+            discretized object. Its size must be equal to the
```
adler-j: \`FnBase.size\`

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            total number of grid points. Its field must be the same
+            as that of the function space.
+        schemes : `str` or `list` of `str`
```
adler-j: Is list correct, shouldnt it be sequence?

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            Indicates which interpolation scheme to use for which axis.
+            A single string is interpreted as a global scheme for all
+            axes.
+        nn_variants : `str` or `list` of `str`, optional
```
adler-j: sequence of str

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            Which variant ('left' or 'right') to use in nearest neighbor
```
adler-j: Since the use of this kw depends on the choice of `schemes` this could be rendered unused. I'd thus make it part of `kwargs`, or at the least put it last.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+        FunctionSetMapping.__init__(self, 'extension', ip_fspace, grid, dspace,
+                                    order, linear=True)
+
+        try:
+            schemes_ = str(schemes + '').lower()  # pythonic string check
+            schemes_ = [schemes_] * self.grid.ndim
+        except TypeError:
+            schemes_ = [str(scm).lower() for scm in schemes]
+
+        if nn_variants is None:
```
adler-j: Complain if used with non nearest neighbor

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            variants_ = ['left'] * self.grid.ndim
+        else:
+            try:
+                variants_ = str(nn_variants + '').lower()  # pythonic str check
+                variants_ = [variants_] * self.grid.ndim
+            except TypeError:
+                variants_ = [str(var).lower() for var in nn_variants]
+
+        for i, (scm, var) in enumerate(zip(schemes, variants_)):
```
adler-j: Dislike using `variants_` if `scheme` is not `nn`.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            if scm == 'nearest' and var not in ('left', 'right'):
+                raise ValueError("Nearest neighbor variant '{}' at index {} "
+                                 "not understood.".format(var, i))
+
+        self._schemes = schemes_
+        self._nn_variants = variants_
+
+    @property
+    def schemes(self):
+        """List of interpolation schemes, one for each axis."""
```
adler-j: Why not simply return a single value if they are all the same?

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+        return self._schemes
+
+    @property
+    def nn_variants(self):
+        """List of nearest neighbor variants, one for each axis."""
```
adler-j: Same as above.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+        Returns
+        -------
+        out : `FunctionSpaceVector`
+            Per-axis interpolator for the grid of this operator. If
+            ``out`` was provided, the returned object is a reference
+            to it.
+        """
+        def interpolator(arg, out=None):
+            """Interpolating function with vectorization."""
+            if is_valid_input_meshgrid(arg, self.grid.ndim):
```
adler-j: Why not simply:

```python
if is_valid_input_meshgrid(arg, self.grid.ndim):
    typ = 'meshgrid'
else:
    typ = 'array'
    
interp = _PerAxisInterpolator(
    self.grid.coord_vectors,
    x.data.reshape(self.grid.shape, order=self.order),
    schemes=self.schemes, nn_variants=self.nn_variants,
    typ=typ)
```

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+                    x.data.reshape(self.grid.shape, order=self.order),
+                    schemes=self.schemes, nn_variants=self.nn_variants,
+                    typ='array')
+
+            return interp(arg, out=out)
+
+        return self.range.element(interpolator, vectorized=True)
+
+
+class _Interpolator(object):
```
adler-j: Shouldnt this come before the `_PerAxisInterpolator`?

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+    implementations.
+    """
 
-        if not hasattr(values, 'ndim'):
-            # allow reasonable duck-typed values
-            values = np.asarray(values)
+    def __init__(self, coord_vecs, values, typ):
+        """Initialize a new instance.
+
+        coord_vecs : `tuple` of `numpy.ndarray`
```
adler-j: sequence

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
                                                        values.shape[i], i))
 
-        # TODO: use the following code to determine upstream if a different
-        # interpolator needs to be used
-
-#        for i, p in enumerate(points):
-#            if not np.all(np.diff(p) > 0.):
-#                raise ValueError("The points in dimension {} must be strictly"
-#                                 " ascending".format(i))
         self.grid = tuple([np.asarray(p) for p in coord_vecs])
```
adler-j: `self.grid = tuple(np.asarray(p) for p in coord_vecs)` should also work

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
 
     def __call__(self, x, out=None):
         """Do the interpolation.
 
-        Modified for in-place evaluation support and without method
-        choice. Evaluation points are to be given as an array with
-        shape (n, dim), where n is the number of points.
+        Parameters
+        ----------
+        x : meshgrid or `numpy.ndarray`
```
adler-j: Do we have meshgrid written in the glossary?

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
-        if x.ndim != 2:
-            raise ValueError('x has {} axes instead of 2.'.format(x.ndim))
-
-        if x.shape[0] != ndim:
-            raise ValueError('x has axis 1 with length {} instead '
-                             'of the grid dimension {}.'
-                             ''.format(x.shape[0], ndim))
+        if self.type == 'array':
+            # Make a (1, n) array from one with shape (n,)
+            ndim = len(self.grid)
```
adler-j: Move outside of if statement and use `self.grid.ndim` instead.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            if ndim == 1 and x.ndim == 1:
+                x = x[None, :]
```
adler-j: I would use `x = x.reshape([ndim, -1])` as it communicates the goal better. Then you could also remove the two following tests since this would fail and we can rely on numpys testing instead of writing our own.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+            if x.ndim != 2:
+                raise ValueError('x has {} axes instead of 2.'.format(x.ndim))
+            if x.shape[0] != ndim:
+                raise ValueError('x has axis 1 with length {} instead '
+                                 'of the grid dimension {}.'
+                                 ''.format(x.shape[0], ndim))
+            out_shape = (x.shape[1],)
+        else:
+            if len(x) != len(self.grid):
```
adler-j: Use `ndim` as moved outside of the if statement above.

odl/discr/discr_mappings.py
```diff
@@ -467,219 +482,307 @@ def _call(self, x, out=None):
...
+                raise ValueError('number of vectors in x is {} instead of '
+                                 'the grid dimension {}.'
+                                 ''.format(len(x), len(self.grid)))
+            if len(x) == 1:
```
adler-j: This is done all the time, how about we make a utility for it? (and make an issue on numpy that `broadcast` should work with a single argument.)

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
 
-class _NearestMeshgridInterpolator(_MeshgridInterpolator,
-                                   _NearestPointwiseInterpolator):
+def _compute_nearest_weights_edge(idcs, ndist, variant):
+    """Helper for nearest interpolation mimicing the linear case."""
+    # Get out-of-bounds indices from the norm_distances. Negative
+    # means "too low", larger than or equal to 1 means "too high"
+    lo = np.where(ndist < 0)
+    hi = np.where(ndist > 1)
+    ones = np.ones_like(ndist)
```
adler-j: `np.where` supports broadcasting, these shouldnt be needed.

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
+        w_lo = np.where(ndist <= 0.5, ones, zeros)
+    else:
+        w_lo = np.where(ndist < 0.5, ones, zeros)
+
+    w_lo[lo] = 0
+    w_lo[hi] = 1
+
+    # For "too high" nodes, the upper neighbor gets weight zero;
+    # "too low" gets 1.
+    if variant == 'left':
```
adler-j: Not so sure about this. Numerical stability tells me that this will have some real issues and since floating point behaviour is not 100% defined, we could even get implementation dependent behavior. I cant think of many cases where we could guarantee this feature to work as people expect it to:

```python
>>> np.array([0.2 + 0.1, 0.3]) <= 0.3
array([ False, True], dtype=bool)
```

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
             return self.values[idx_res]
 
 
-class _NearestMeshgridInterpolator(_MeshgridInterpolator,
-                                   _NearestPointwiseInterpolator):
+def _compute_nearest_weights_edge(idcs, ndist, variant):
+    """Helper for nearest interpolation mimicing the linear case."""
+    # Get out-of-bounds indices from the norm_distances. Negative
+    # means "too low", larger than or equal to 1 means "too high"
+    lo = np.where(ndist < 0)
```
adler-j: Why do we use `where` here?

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
+
+    # For "too low" nodes, the lower neighbor gets weight zero;
+    # "too high" gets 2 - yi (since yi >= 1)
+    w_lo = (1 - ndist)
+    w_lo[lo] = 0
+    w_lo[hi] += 1
+
+    # For "too high" nodes, the upper neighbor gets weight zero;
+    # "too low" gets 1 + yi (since yi < 0)
+    w_hi = np.array(ndist, copy=True)
```
adler-j: `numpy.copy()` is more clear

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
+    # For upper/lower out-of-bounds nodes, we need to set the
+    # lower/upper neighbors to the last/first grid point
+    edge = [idcs, idcs + 1]
+    edge[0][hi] = -1
+    edge[1][lo] = 0
+
+    return w_lo, w_hi, edge
+
+
+def _create_lists(indices, norm_distances, schemes, variants):
```
adler-j: Perhaps a more specific name

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
+        if out is None:
+            out = np.zeros(out_shape, dtype=out_dtype)
+        else:
+            out[:] = 0.0
+
+        # Weights and indices (per axis)
+        low_weights, high_weights, edge_indices = _create_lists(
+            indices, norm_distances, self.schemes, self.nn_variants)
+
+        # Iterate over all possible combinations of [i, i+1] for each
```
adler-j: Not the most obvious logic, make sure this is tested for corner cases.

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
+        coord_vecs : `tuple` of `numpy.ndarray`
+            Coordinate vectors defining the interpolation grid
+        values : array-like
+            Grid values to use for interpolation
+        typ : {'array', 'meshgrid'}
+            Type of expected input values in ``__call__``
+        """
+        super().__init__(coord_vecs, values, typ,
+                         schemes=['linear'] * len(coord_vecs),
+                         nn_variants=[None] * len(coord_vecs))
```
adler-j: Should be omit-able.

odl/discr/discr_mappings.py
```diff
@@ -687,42 +790,185 @@ def _evaluate(self, indices, norm_distances, out=None):
...
+
+        coord_vecs : `tuple` of `numpy.ndarray`
+            Coordinate vectors defining the interpolation grid
+        values : array-like
+            Grid values to use for interpolation
+        typ : {'array', 'meshgrid'}
+            Type of expected input values in ``__call__``
+        """
+        super().__init__(coord_vecs, values, typ,
+                         schemes=['linear'] * len(coord_vecs),
```
adler-j: `schemes = 'linear'`

odl/discr/discretization.py
```diff
@@ -33,11 +33,8 @@
 from odl.set.sets import Set, RealNumbers, ComplexNumbers
 from odl.set.space import LinearSpace
 from odl.space import CUDA_AVAILABLE
-if CUDA_AVAILABLE:
-    from odl.space.cu_ntuples import CudaNtuples, CudaFn, CudaRn
-    CudaCn = type(None)  # TODO: add CudaCn to imports once it is implemented
-else:
-    CudaRn = CudaCn = CudaFn = CudaNtuples = type(None)
+from odl.space.cu_ntuples import CudaNtuples, CudaFn, CudaRn
+CudaCn = type(None)  # TODO: add CudaCn to imports once it is implemented
```
adler-j: I guess this is getting more and more requested, I'll look at it.

odl/discr/discretization.py
```diff
@@ -186,6 +183,7 @@ def element(self, inp=None):
         if inp is None:
             return self.element_type(self, self.dspace.element())
         try:
+            # pylint: disable=not-callable
```
adler-j: Dislike these disables. pylint seems to fail us here, both for `element_type` and all `Operator`'s, id say disable globally.

odl/discr/discretization.py
```diff
@@ -235,14 +233,8 @@ class RawDiscretizationVector(NtuplesBaseVector):
...
-        if not isinstance(space, RawDiscretization):
-            raise TypeError('space {!r} not a `RawDiscretization` '
-                            'instance.'.format(space))
-
-        if not isinstance(ntuple, space.dspace.element_type):
-            raise TypeError('n-tuple {!r} not an `{}` vector.'
-                            ''.format(ntuple,
-                                      space.dspace.__class__.__name__))
+        assert isinstance(space, RawDiscretization)
+        assert isinstance(ntuple, space.dspace.element_type)
```
adler-j: `assert ntuple in space.dspace`?

odl/discr/discretization.py
```diff
@@ -416,26 +408,32 @@ def one(self):
 
     def _lincomb(self, a, x1, b, x2, out):
         """Raw linear combination."""
+        # pylint: disable=protected-access
```
adler-j: arghh, lets just use the public method instead then? Less error prone I guess.

odl/discr/discretization.py
```diff
@@ -450,9 +448,7 @@ class DiscretizationVector(RawDiscretizationVector, FnBaseVector):
 
     def __init__(self, space, data):
         """Initialize a new instance."""
-        if not isinstance(space, Discretization):
-            raise TypeError('space {!r} not a `Discretization` '
-                            'instance.'.format(space))
+        assert isinstance(space, Discretization)
```
adler-j: Using asserts in these cases is totally fine IMO, these shouldnt realy be used by users and this makes the code more simple to read.

odl/discr/discretization.py
```diff
@@ -450,9 +448,7 @@ class DiscretizationVector(RawDiscretizationVector, FnBaseVector):
...
         super().__init__(space, data)
```
adler-j: Perhaps `FnBaseVector.__init__` instead? Performance and lets you see which of the super-classes is called at a glance.

odl/discr/grid.py
```diff
@@ -1165,6 +1185,7 @@ def uniform_sampling(intv_prod, num_nodes, as_midp=True):
         raise ValueError('number of nodes {} has non-positive entries.'
                          ''.format(num_nodes))
 
+    # pylint: disable=protected-access
```
adler-j: Make `ideg` public instead.

odl/discr/lp_discr.py
```diff
@@ -101,7 +96,11 @@ def __init__(self, fspace, grid, dspace, exponent=2.0, interp='nearest',
         if self.interp == 'nearest':
             extension = NearestInterpolation(fspace, grid, dspace,
                                              order=self.order)
+        elif self.interp == 'linear':
+            extension = LinearInterpolation(fspace, grid, dspace,
+                                            order=self.order)
         else:
+            # Should not happen
```
adler-j: Is `raise NotImplementedError` correct then? Perhaps `assert False`?

odl/space/ntuples.py
```diff
@@ -31,13 +31,12 @@
...
 
 # External module imports
-# pylint: disable=no-name-in-module
 import ctypes
 from numbers import Integral
 from functools import partial
 from math import sqrt
 import numpy as np
-import scipy as sp
+import scipy.linalg as linalg
```
adler-j: Avoid importing `a.b.c`, it harms mock and is somewhat irritating to deal with in code since it is not obvious what `linalg` is intended. Simply use `sp.linalg` instead.

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+from odl.discr.discr_mappings import (
+    GridCollocation, NearestInterpolation, LinearInterpolation,
+    PerAxisInterpolation)
+
+
+pytestmark = pytest.mark.benchmark
+
+
+def initialize():
+    global shape, values, hcube, grid, space, dspace32, dspace64
```
adler-j: ugh, no better way?

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+
```
adler-j: Description should be added.

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+    global shape, values, hcube, grid, space, dspace32, dspace64
+    global points, mesh
+    shape = (100, 100, 100, 100)
+    values = np.zeros(np.prod(shape))
+    hcube = odl.IntervalProd([0] * 4, [1] * 4)
+    grid = odl.uniform_sampling(hcube, shape, as_midp=True)
+    space = odl.FunctionSpace(hcube)
+    dspace32 = odl.Rn(grid.ntotal, dtype='float32')
+    dspace64 = odl.Rn(grid.ntotal, dtype='float64')
+    points = np.random.rand(4, 60 ** 4)
```
adler-j: 60? See above comment.

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+    PerAxisInterpolation)
+
+
+pytestmark = pytest.mark.benchmark
+
+
+def initialize():
+    global shape, values, hcube, grid, space, dspace32, dspace64
+    global points, mesh
+    shape = (100, 100, 100, 100)
```
adler-j: I'd write this:

```python
n = 100
ndim = 4

shape = [100] * ndim
values = np.zeros(np.prod(shape))
hcube = odl.IntervalProd([0] * ndim, [1] * ndim)
grid = odl.uniform_sampling(hcube, shape, as_midp=True)
space = odl.FunctionSpace(hcube)
dspace32 = odl.Rn(n ** ndim, dtype='float32')
dspace64 = odl.Rn(n ** ndim, dtype='float64')
points = np.random.rand(ndim, n ** ndim)
mesh = sparse_meshgrid(*([np.random.rand(n)] * ndim))
```

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+initialize()
+
+
+def func(x):
+    return np.sin(x[0] * x[1]) - np.cos(x[2] * x[3])
+
+
+dspace_params = [dspace32, dspace64]
+dspace_ids = [' Rn, dtype = {} '.format(spc.dtype) for spc in dspace_params]
```
adler-j: `['{!r}'.format(spc) for spc in dspace_params]`

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+dspace_fixture = pytest.fixture(scope="module", ids=dspace_ids,
```
adler-j: Defined and used on next row, simply use it instantly instead.

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+@dspace_fixture
+def dspace(request):
+    return request.param
+
+
+input_params = [points, mesh]
+input_ids = [' Array input ', ' Meshgrid input ']
+input_fixture = pytest.fixture(scope="module", ids=input_ids,
```
adler-j: As above

test/benchmark/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_per_axis_interpolation(dspace, input_values):
+    schemes = ['nearest', 'linear', 'linear', 'nearest']
+    pa_interp_op = PerAxisInterpolation(space, grid, dspace, schemes=schemes)
+    function = pa_interp_op(values)
+    function(input_values)
+
+
+if __name__ == '__main__':
+    pytest.main(str(__file__.replace('\\', '/')) + ' -v')
```
adler-j: Perhaps we could rewrite this using `sys.path`. Thats for later i presume.

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+
```
adler-j: Docstring.

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+    # 'right' variant
+    interp_op = NearestInterpolation(space, grid, dspace, variant='right')
+    values = np.arange(5)
+    function = interp_op(values)
+
+    # Testing two midpoints and the extreme values
+    pts = np.array([0.4, 0.8, 0.0, 1.0])
+    true_arr = [2, 4, 0, 4]
+    assert all_equal(function(pts), true_arr)
```
adler-j: Pytest allows you to specify better error messages for these, perhaps we should look into it.

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+    true_arr = [1, 3, 0, 4]
+    assert all_equal(function(pts), true_arr)
+
+    # 'right' variant
+    interp_op = NearestInterpolation(space, grid, dspace, variant='right')
+    values = np.arange(5)
+    function = interp_op(values)
+
+    # Testing two midpoints and the extreme values
+    pts = np.array([0.4, 0.8, 0.0, 1.0])
```
adler-j: Does pts need to be a `np.array`?

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+    # Coordinate vectors are:
+    # [0.1, 0.3, 0.5, 0.7, 0.9]
+
+    space = odl.FunctionSpace(intv)
+    dspace = odl.Fn(grid.ntotal, dtype='float64')
+
+    # 'left' variant
+    interp_op = NearestInterpolation(space, grid, dspace, variant='left')
+    values = np.arange(5, dtype='float64')
+    function = interp_op(values)
```
adler-j: Dont we have a `interp_op([1.0, 2.0, 3.0, 4.0, 5.0])` syntax?

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+def test_nearest_interpolation_1d_variants():
+    intv = odl.Interval(0, 1)
+    grid = odl.uniform_sampling(intv, 5, as_midp=True)
+    # Coordinate vectors are:
+    # [0.1, 0.3, 0.5, 0.7, 0.9]
+
+    space = odl.FunctionSpace(intv)
+    dspace = odl.Fn(grid.ntotal, dtype='float64')
```
adler-j: I propose renaming `grid.ntotal` -> `grid.size` for consistency reasons.

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+    # Coordinate vectors are:
+    # [0.125, 0.375, 0.625, 0.875], [0.25, 0.75]
+
+    space = odl.FunctionSpace(rect)
+    dspace = odl.Fn(grid.ntotal, dtype='float64')
+    interp_op = NearestInterpolation(space, grid, dspace)
+    values = np.arange(8, dtype='float64')
+    function = interp_op(values)
+
+    # Evaluate at single point
```
adler-j: More spacing here for readability.

test/discr/discr_mappings_test.py
```diff
@@ -0,0 +1,392 @@
...
+    coll_op_c = GridCollocation(space, grid, dspace, order='C')
+    coll_op_f = GridCollocation(space, grid, dspace, order='F')
+    interp_ops_c = [NearestInterpolation(space, grid, dspace, order='C',
+                                         variant='left'),
+                    NearestInterpolation(space, grid, dspace, order='C',
+                                         variant='right'),
+                    LinearInterpolation(space, grid, dspace, order='C'),
+                    PerAxisInterpolation(space, grid, dspace, order='C',
+                                         schemes=['linear', 'nearest'])]
+    interp_ops_f = [NearestInterpolation(space, grid, dspace, order='F',
```
adler-j: Why one `_c` and one `_f`?

adler-j:
Some general comments, structure looks good. Since we added more abstraction, we should make sure we didnt make anything slow. Else fix and pull imo.

.pylintrc
kohr-h: Bad old habit :-)

odl/discr/discr_mappings.py
kohr-h: You're right, I'll have a look into its use and if it's potentially slowing down things.

odl/discr/discr_mappings.py
kohr-h: :+1:

odl/discr/discr_mappings.py
kohr-h: I know and agree, but I chose this over not disabling (because the warnings are nonsense) and disabling globally since warnings can actually help to find bugs. Maybe not so much warnings like `abstract-method` and `private-access`.

odl/discr/discr_mappings.py
kohr-h: Should not be needed at all, actually.

odl/discr/discr_mappings.py
kohr-h: Or `in_type`. I think I saw `typ` as a replacement for `type` as argument somewhere. Not so great, yeah.

odl/discr/discr_mappings.py
kohr-h: ok

odl/discr/discr_mappings.py
kohr-h: :+1:

odl/discr/discr_mappings.py
kohr-h: :+1:

odl/discr/discr_mappings.py
kohr-h: Oh, you're right.

odl/discr/discr_mappings.py
kohr-h: That's where it should have been. Numpy does it this way: make it an item of the same kind, just at the end.

odl/discr/discr_mappings.py
kohr-h: :+1:

odl/discr/discr_mappings.py
kohr-h: I'll disallow it.

odl/discr/discr_mappings.py
kohr-h: Can do.

odl/discr/discr_mappings.py
kohr-h: Yes

odl/discr/discr_mappings.py
kohr-h: Hm, the code must have been more different before. Can be easily unified, right.

odl/discr/discr_mappings.py
kohr-h: It does

odl/discr/discr_mappings.py
kohr-h: Yep

odl/discr/discr_mappings.py
kohr-h: Right, I think I just copied the code.

odl/discr/discr_mappings.py
kohr-h: Good one, I'll add it.

odl/discr/discr_mappings.py
kohr-h: I'll do that change.

odl/discr/discr_mappings.py
kohr-h: Actually we have a utility, which I forgot to use. I'll look into the NumPy issue tracker if there is something about this topic (e.g. why they don't implement it).

odl/discr/discr_mappings.py
kohr-h: I see the point. Maybe the situation can be rescued by changing the perspective in that case by checking for `ndist > 0.5` and change the implementation of `_evaluate`.

odl/discr/discr_mappings.py
kohr-h: ok

odl/discr/discr_mappings.py
kohr-h: In an earlier version I needed the indices, too. We should be able to do without now.

odl/discr/discr_mappings.py
kohr-h: Not too great creativity here, I admit.

odl/discr/discr_mappings.py
kohr-h: Sure. This should be robust, however, since it's equivalent to [the original scipy inmplementation](https://github.com/scipy/scipy/blob/master/scipy/interpolate/interpolate.py#L1773-L1786).

odl/discr/discr_mappings.py
kohr-h: This is a private class, I think we don't need convenience here.

odl/discr/discr_mappings.py
kohr-h: Same as above. The logic to allow for this is in the public operator, I deliberately kept this class as simple as possible.

odl/discr/discretization.py
kohr-h: :+1:

odl/discr/discretization.py
kohr-h: Ok

odl/discr/discretization.py
kohr-h: That was my thought, too.

odl/discr/discretization.py
kohr-h: Okay, but that does not save us everywhere, e.g. in `OperatorSum`.

odl/discr/discretization.py
kohr-h: Sure

odl/discr/grid.py
kohr-h: Ok

odl/discr/lp_discr.py
kohr-h: We could also raise `RuntimeError`, I think you did that at some place.

odl/discr/discr_mappings.py
adler-j: Possibly, and I can certainly see where this comes from mathematically, but I have a hard time seeing this working in practice since we always need to assume floats are approximate.

odl/discr/discr_mappings.py
adler-j: Great!

odl/discr/discr_mappings.py
adler-j: Ok that's totally fine.

odl/space/ntuples.py
kohr-h: That was the first version, which failed. The problem is the following (you need to store the script as a file and run it in a new interpreter, not in Spyder):
```
>>> import scipy as sp
>>> sp.linalg.blas.get_blas_funcs('axpy')
...
AttributeError: 'module' object has no attribute 'linalg'
```
Whereas this works:
```
>>> import scipy.linalg as linalg
>>> linalg.blas.get_blas_funcs('axpy')  # ok
```
I know it's harder to mock, but the original code triggered an error when I removed some scipy imports in the `discr_mappings.py` module.

test/benchmark/discr/discr_mappings_test.py
kohr-h: Alternative would be to just not make it a function. But I thought it might be useful to have a figure on the setup time, too. Or maybe the function could return all that stuff.

test/benchmark/discr/discr_mappings_test.py
kohr-h: Sure.

test/benchmark/discr/discr_mappings_test.py
kohr-h: Alright

test/benchmark/discr/discr_mappings_test.py
kohr-h: Why not? 60 is a good number in terms of prime number decomposition. If we want an example with `ndim=5` and `ndim=6` with the same number of nodes, it's easily possible.
And, btw, this does not in any way have to do with the 100 in `shape`.

test/benchmark/discr/discr_mappings_test.py
kohr-h: Can do

test/benchmark/discr/discr_mappings_test.py
kohr-h: Ok

test/benchmark/discr/discr_mappings_test.py
kohr-h: :+1:

test/benchmark/discr/discr_mappings_test.py
adler-j: Noticed that when reading on, perhaps we can split them or add some comment here. Tests are also somewhat intended as tests and should (hopefully) be readable to outsiders (not you/me that is).

test/benchmark/discr/discr_mappings_test.py
kohr-h: It should have a default implementation somewhere, right.

test/discr/discr_mappings_test.py
kohr-h: Ok

test/discr/discr_mappings_test.py
kohr-h: Which would be implemented in `all_equal`, I guess?

adler-j: Yeah, it was just a random note I had when reading.

test/discr/discr_mappings_test.py
kohr-h: Currently yes. The problem is that general sequences can be interpreted both ways, i.e. is
```python
>>> x = [[1, 2],
         [3, 4]]
```
an array of two points or a meshgrid in 2D? It can be both. We need some way to distinguish, so here [I'm strict](https://github.com/odlgroup/odl/blob/issue-158__interpolation/odl/util/vectorization.py#L41-L42) and require an input array to be `ndarray`, while meshgrids are [a bit more flexible](https://github.com/odlgroup/odl/blob/issue-158__interpolation/odl/util/vectorization.py#L49-L66).

test/discr/discr_mappings_test.py
kohr-h: Should work. It follows the same rules as `Operator` in general, so any input understood by `Ntuples.element` is okay. I can use your version to improve clarity.

test/discr/discr_mappings_test.py
kohr-h: Then we must rename [the current `size` method](https://github.com/odlgroup/odl/blob/issue-158__interpolation/odl/discr/grid.py#L292-L294) to something else. Maybe `extent`? In the long run, this functionality will go to a new `Tessellation` class, I guess.

test/discr/discr_mappings_test.py
adler-j: Ok im fine with this for now. Better use the strict sense until we know better.

kohr-h:
We have benchmarks on Jenkins, look there if you want to see numbers. I wrote some lines in the beginning why I think the relations are reasonable.
What I could do is add a check if our implementation is about as fast as the original Scipy one. That one could fail if we're off by a too large factor.

test/discr/discr_mappings_test.py
adler-j: Options:

* extent
* sides
* edges
* sidelen(gth)
* edgelen(gth)



Inspiration:

https://en.wikipedia.org/wiki/Square
Uses "sides" seemingly everywhere.

test/discr/discr_mappings_test.py
adler-j: If `dtype='float64'` isnt necessary we could use `interp_op([1, 2, 3, 4, 5])`

kohr-h: The `dtype` is only needed for `arange` since it produces an `int` array by default. The operator would convert, but maybe the test code relying on the array would have undesired results. I'll just change to the simple list.

test/discr/discr_mappings_test.py
kohr-h: Right. And for the record, meshgrids are the intended default use, at least as long as we have a tensor grid to evaluate on. They're much more efficient. Arrays as input are intended for arbitrary point collections, which we don't use internally at the moment, but that may change over time.

odl/discr/discr_mappings.py
kohr-h: These are used only in the private classes, once per interpolator creation, i.e. once per operator evaluation.

odl/discr/discr_mappings.py
kohr-h: Second thought, not sure about this. I rather return the same type regardless of input during init. Otherwise, each time we use this value for iteration, we need to differentiate.

odl/discr/discr_mappings.py
kohr-h: See above.

odl/discr/discr_mappings.py
adler-j: Ok, valid point. We should however have a repr method on this, and in that one, we can use the shortened version.

kohr-h: That's true. I'm gonna add it.

odl/discr/discr_mappings.py
kohr-h: There's an [issue in NumPy](https://github.com/numpy/numpy/pull/6905) about this now. We'll use our utility in the meanwhile.

kohr-h:
All suggested changes applied. The only open question is whether we want a left and a right variant of nearest neighbor or not. The disadvantage in not having it is that there is no discretization which naturally leads to backward differences in derivative discretizations. But maybe this is not so bad, one can still explicitly use backward difference even if it's not natural for the discretization. We could simply remove the `nn_variants` option and always go for the left version.

kohr-h:
I added now also a speed test against Scipy for our linear interpolation. A quick check says that our version is about 20 % faster than the one in Scipy, so I added a test which fails if we are slower than Scipy.

adler-j:
> All suggested changes applied. The only open question is whether we want a left and a right variant of nearest neighbor or not. The disadvantage in not having it is that there is no discretization which naturally leads to backward differences in derivative discretizations. But maybe this is not so bad, one can still explicitly use backward difference even if it's not natural for the discretization. We could simply remove the nn_variants option and always go for the left version.

I certainly see where this comes from, and since we have it, I assume we can keep it, but with some disclaimers in the doc. In 98% of situations users will select the default anyway.

> I added now also a speed test against Scipy for our linear interpolation. A quick check says that our version is about 20 % faster than the one in Scipy, so I added a test which fails if we are slower than Scipy.

A 20% margin cannot hold on all platforms? Seems to tight, I'd relax it to say we are about as fast as scipy (say < 2 * scipy) to avoid un-needed fails.

adler-j:
Also, it seems this is the first benchmark we have (at all), I'd leave it for later and make a branch with some benchmarks so we can add them at once, then we get to consider the structure some more before we settle.

kohr-h:
> Also, it seems this is the first benchmark we have (at all), I'd leave it for later and make a branch with some benchmarks so we can add them at once, then we get to consider the structure some more before we settle.

OK, I'll remove the code from the branch and merge the rest. There will be a new issue branch with the benchmark then.

Issue #188, opened 2015-12-22
Add optional support for Numba


kohr-h:
There's a very interesting project called [Numba](http://numba.pydata.org/) which adds JIT functionality to core python with a focus on mathematical functions. A very brief but still interesting summarizing talk can be watched in [this video from this year's SciPy conference](https://www.youtube.com/watch?v=eYIPEDnp5C4&index=23&list=PLYx7XA2nY5Gcpabmu61kKcToLz0FapmHu).

What is good about it compared to other approaches like PyPy is that it sets on top of CPython instead of defining a new language, so we don't have to make compromises wrt functionality.

The most interesting use cases for us would be
- Proper, fast vectorization instead of our own shitty one. Numba has its own [`vectorize` decorator](http://numba.pydata.org/numba-doc/0.21.0/user/vectorize.html#the-vectorize-decorator) 
- Possibility to define more general Numpy Ufuncs in the Python layer with [the `guvectorize` decorator](http://numba.pydata.org/numba-doc/0.21.0/user/vectorize.html#the-guvectorize-decorator)
- Crank down execution time of functions with repeated Ufunc evaluation as [in this issue](https://github.com/odlgroup/odl/pull/178#discussion_r48008288). Here, Numba detects the repeated evaluations and aggregates them into one loop if possible. The list of [supported Python features](http://numba.pydata.org/numba-doc/0.21.0/reference/pysupported.html) and [NumPy features](http://numba.pydata.org/numba-doc/0.21.0/reference/numpysupported.html) seems quite exhaustive.

Adding support for Numba would not be a big deal since it basically consists of a couple of decorators only. We just decorate the functions we think (or know after profiling) will be worth it, and if Numba isn't installed, all decorators are just identity.

adler-j:
An interesting idea, and I have done some simple tests with this. However, nIn general I am skeptical to optional features like this, mainly because we get more cases to test and support. Issues like "ODL crashing on arch-linux when using numba 1.337 in german locale" come to mind.

With that in mind, these are my comments to your use-cases

> Proper, fast vectorization instead of our own shitty one. Numba has its own vectorize decorator

For all inherent implementations we should try to use numpys vectorization, for user defined operators they are free to use Numba to vectorize their functions, and we should make sure that we properly support this, maybe even provide an example.

> Possibility to define more general Numpy Ufuncs in the Python layer with the guvectorize decorator

Which ones would we standardize?

[...]

PR #198, opened 2016-01-04
Issue 188 numba

adler-j:
Initial addition, currently fails in the numba example with:

```python
Running numba example
  ...
  File "D:/GitHub/ODL/examples/vectorization.py", line 70, in numba_example
    f_numba = X.element(my_vectorized_func)
  File "d:\github\odl\odl\space\fspace.py", line 509, in element
    return self.element_type(self, fcall)
  File "d:\github\odl\odl\space\fspace.py", line 780, in __init__
    FunctionSetVector.__init__(self, fspace, fcall)
  File "d:\github\odl\odl\space\fspace.py", line 221, in __init__
    bound_call=fcall.__call__)
  File "d:\github\odl\odl\operator\operator.py", line 231, in _dispatch_call_args
    raise TypeError('{} is not a bound method.'.format(call))
TypeError: <method-wrapper '__call__' of DUFunc object at 0x0000000019C27308> is not a bound method.
```

This is a "deficiency" in our vectorization implementation and should perhaps be fixed as discussed in #188. Perhaps @kohr-h could look at it?

### DIFF
--- a/examples/vectorization.py
+++ b/examples/vectorization.py
@@ -63,14 +63,14 @@ def numba_example():
         else:
             return x[0] + x[1]
 
-    @vectorize(vectorizer=numba.vectorize, nin=2,
-               vec_args=[['float64(float64, float64)']])
+    # Numba expects functions f(x1, x2, x3, ...), while we have the
+    # convention f(x) with x = (x1, x2, x3, ...). Therefore we need
+    # to wrap the Numba-vectorized function.
+    vectorized = numba.vectorize(lambda x, y: x - y if x > y else x + y)
+
     def myfunc_vec(x):
         """Return x - y if x > y, otherwise return x + y."""
-        if x[0] > x[1]:
-            return x[0] - x[1]
-        else:
-            return x[0] + x[1]
+        return vectorized(*x)
 
     def myfunc_native_vec(x):
         """Return x - y if x > y, otherwise return x + y."""
--- a/examples/vectorization.py
+++ b/examples/vectorization.py
@@ -1,0 +1,85 @@
+# Copyright 2014, 2015 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example showing how to use vectorization of FunctionSpaceVector's."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+import numpy as np
+import odl
+import timeit
+
+
+def performace_example():
+    # Create a function space
+    X = odl.FunctionSpace(odl.Interval(0, 1))
+
+    # Functions default to vectorized
+    f_vec = X.element(lambda x: x**2)
+
+    # If 'vectorized=False' is used, odl automatically vectorizes
+    f_novec = X.element(lambda x: x**2, vectorized=False)
+
+    # Example of runtime, expect vectorized to be much faster
+    points = np.linspace(0, 1, 10000)
+
+    print('Vectorized runtime:     {:5f}'
+          ''.format(timeit.timeit(lambda: f_vec(points), number=100)))
+    print('Non-vectorized runtime: {:5f}'
+          ''.format(timeit.timeit(lambda: f_novec(points), number=100)))
+
+
+def numba_example():
+    # Some functions are not easily vectorized,
+    # here we can use numba to improve performance.
+
+    try:
+        import numba
+    except ImportError:
+        print('Numba not installed, skipping.')
+        return
+
+    def myfunc(x):
+        "Return a-b if a>b, otherwise return a+b"
+        if x[0] > x[1]:
+            return x[0] - x[1]
+        else:
+            return x[0] + x[1]
+
+    my_vectorized_func = numba.vectorize(myfunc)
+
+    # Create functions
+    X = odl.FunctionSpace(odl.Rectangle([0, 0], [1, 1]))
+    f_default = X.element(myfunc, vectorized=False)
+    f_numba = X.element(my_vectorized_func)
+
+    # Create points
+    points = odl.uniform_sampling(X.domain, [100, 100]).points().T
+
+    print('Vectorized runtime:     {:5f}'
+          ''.format(timeit.timeit(lambda: f_default(points), number=100)))
+    print('Non-vectorized runtime: {:5f}'
+          ''.format(timeit.timeit(lambda: f_numba(points), number=100)))
+
+if __name__ == '__main__':
+    print('Running performance example')
+    performace_example()
+
+    print('Running numba example')
+    numba_example()
--- a/examples/vectorization.py
+++ b/examples/vectorization.py
@@ -23,21 +23,24 @@ from future import standard_library
 standard_library.install_aliases()
 import numpy as np
 import odl
-from odl.util.vectorization import vectorize
 import timeit
 
 
 def performace_example():
-    # Create a function space
-    X = odl.FunctionSpace(odl.Interval(0, 1))
+    # Create a space of functions on the interval [0, 1].
+    fspace = odl.FunctionSpace(odl.Interval(0, 1))
 
-    # Functions default to vectorized
-    f_vec = X.element(lambda x: x ** 2)
+    # Simple function, already supports vectorization.
+    f_vec = fspace.element(lambda x: x ** 2)
 
-    # If 'vectorized=False' is used, odl automatically vectorizes
-    f_novec = X.element(lambda x: x ** 2, vectorized=False)
+    # If 'vectorized=False' is used, odl automatically vectorizes with
+    # the help of numpy.vectorize. This will be very slow, though, since
+    # the implementation is basically a Python loop.
+    f_novec = fspace.element(lambda x: x ** 2, vectorized=False)
 
-    # Example of runtime, expect vectorized to be much faster
+    # We test both versions with 10000 evaluation points. The natively
+    # vectorized version should be much faster than the one using
+    # numpy.vectorize.
     points = np.linspace(0, 1, 10000)
 
     print('Vectorized runtime:     {:5f}'
@@ -47,8 +50,9 @@ def performace_example():
 
 
 def numba_example():
-    # Some functions are not easily vectorized,
-    # here we can use numba to improve performance.
+    # Some functions are not easily vectorized, here we can use Numba to
+    # improve performance.
+    # See http://numba.pydata.org/
 
     try:
         import numba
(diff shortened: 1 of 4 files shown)


kohr-h:
This is a case of limiting design. We support python functions, `FunctionSetVector` and Numpy ufuncs, all of it explicitly, because we want to be able to understand the signature of the function. This excludes unknown callable objects and is clearly undesirable.

A solution would be to implement a very simple `wrap_external_callable` decorator which returns a Python function with one of the three possible call patterns. The returned wrapper could then be inspected as usual. Something like this:
```python
def wrap_external_callable(outarg):
    def decorator(func):
        if outarg == 'none':
            def wrapper(x, **kwargs):
                return func(x, **kwargs)

        elif outarg == 'positional':
            def wrapper(x, out, **kwargs):
                return func(x, out, **kwargs)
            
        elif outarg == 'optional':
            def wrapper(x, out=None, **kwargs):
                return func(x, out=out, **kwargs)

        return wrapper
    return decorator
```
We would need to instruct users to decorate their functions according to the underlying implementation then.

An alternative (or last resort instead of failing) would be to "play it safe" and assume out-of-place by default, but as only implementation, this would make Numba vectorizations less useful.

kohr-h:
The last push comes with a new `vectorize` decorator which can use `numba.vectorize` internally. The example is rewritten and works, but Numba performance is not so great, not sure why. It's a bit faster than the `np.vectorize` solution, but native vectorized code is still about 10x faster.

I'm not sure if the implemented layout is optimal for the internals of Numba. The first version vectorized at function call time, hence each time the decorated function is called, which is, of course, not so great. But the "vectorize once, call often" pattern does not seem to help much.

Anyway, we do support Numba with this decorator now, although the "dynamic ufuncs" don't work as of now. And additionally, this rewritten vectorizer is a clear win for the `FunctionSetVector` class.

kohr-h:
As a side note, the code looks more complicated than it actually is. Much of it is due to the special cases up to 5 input arguments. Numba does not work (at least I didn't get it to work) with variable arguments.
Other than that:
- The `__new__` trick implements the decorator without argument, `@vectorize`.
- With arguments, a class is created and the wrapper is implemented in `__call__`.
- The rest of the code takes care of the "single-arg-with-components" vs. "multiple-args" translation between our approach and the vectorizers.
- Another split is with respect to `kwargs`. We could remove that and instead try to inspect and decide, but that may be complicated, too, as we saw before. In addition, Numba does not support `kwargs` at all, so I wanted to make a case completely without them.
- And finally, if `nin`, the number of input arguments is unknown, it must be determined by the wrapper function, which then vectorizes and calls the vectorized function directly. For given `nin`, the vectorized function is created, and then wrapped before the first call. This should be faster, but surprisingly, running times are not really better. Maybe a mistake at another place. In any case another branch.

kohr-h:
@adler-j you're good at optimization, maybe you spot a performance glitch.

adler-j:
The code does not vectorize at all according to profiling, however, the following works fine:

```python
def myfunc_vec(x):
    @numba.vectorize(['float64(float64, float64)'])
    def inner(x, y):
        if x > y:
            return x - y
        else:
            return x + y

    return inner(x[0], x[1])
```

with that, I get:

```
Non-Vectorized runtime (points):     1.093183
Non-Vectorized runtime (meshgrid):     0.911517
Numba vectorized runtime (points): 0.089119
Numba vectorized runtime (meshgrid): 0.019631
Native vectorized runtime (points): 0.091807
Native vectorized runtime (meshgrid): 0.020120
```

The new code is honestly quite complicated and after an hour of looking and profiling I honestly cant find the issue, how about simply providing the above example and let users vectorize themselves?

kohr-h:
You're right, we should just give that simple example and let the users generalize it to their own purposes. I guess we can just use the `numpy_vectorize` directly then, which could be renamed to `vectorize`. I'll test it.

kohr-h:
OK so now we have a much simplified vectorizer using always `numpy.vectorize`. I also changed the example so it uses your manual vectorization code. Looks much more reasonable to me now, too.

odl/space/fspace.py
```diff
@@ -504,7 +506,7 @@ def element(self, fcall=None, vectorized=True):
                 else:
                     dtype = 'complex128'
 
-                fcall = vectorize(dtype=dtype)(fcall)
+                fcall = vectorize(otypes=[dtype])(fcall)
```
adler-j: why change to otypes?

kohr-h: That's what `numpy.vectorize` takes.

adler-j: Good by me.

odl/util/vectorization.py
```diff
@@ -124,169 +125,240 @@ def out_shape_from_array(arr):
...
-                try:
-                    return func(x)
-                except Exception as err:
-                    raise_from(
-                        TypeError('invalid vectorized input type.'), err)
-
-        return vect_wrapper
-    return vect_decorator
+    def decorator(func):
+        if not use_kw:
```
adler-j: Do we really need this distinction? Isnt a empty **kwargs ignored?

```python
>>> def f(x): print(x)
>>> def g(x, **kwargs): f(x, **kwargs)
>>> g(5)
5
```

odl/util/vectorization.py
```diff
@@ -124,169 +125,240 @@ def out_shape_from_array(arr):
...
+class vectorize(object):
+
+    """Decorator class for function vectorization.
+
+    This vectorizer expects a function with exactly one positional
+    argument (input) and optional keyword arguments. The decorated
+    function has an optional ``out`` parameter for in-place evaluation.
+    """
+
+    def __new__(cls, *args, **kwargs):
```
adler-j: This is a great pattern, but isnt there some standard class to do this? Else we should standardize it to use with other decorators.

odl/util/vectorization.py
kohr-h: That's right. I wrote this decorator first and intended to use it like this:
```python
@combine_input_args
@np.vectorize
@split_input_args
def f(x):
    # bla
```
As in the docstring example as I see now. But This doesn't work with arguments to `np.vectorize`, so this stuff is somehow obsolete. The question is whether we want to keep it anyway, maybe some users would find such a decorator useful.

Anyway, the `use_kw` thing can go.

adler-j: I'm against keeping anything we dont use, this will simply lead to us having HEAPS of legacy code after a while. If we can remove, please do.

examples/vectorization.py
```diff
@@ -0,0 +1,108 @@
...
+            return x[0] + x[1]
+
+    # Numba expects functions f(x1, x2, x3, ...), while we have the
+    # convention f(x) with x = (x1, x2, x3, ...). Therefore we need
+    # to wrap the Numba-vectorized function.
+    vectorized = numba.vectorize(lambda x, y: x - y if x > y else x + y)
+
+    def myfunc_vec(x):
+        """Return x - y if x > y, otherwise return x + y."""
+        return vectorized(*x)
```
adler-j: Perhaps we should write `x[0], x[1]` since this is easier for new users to understand.

examples/vectorization.py
```diff
@@ -0,0 +1,108 @@
...
+    fspace = odl.FunctionSpace(odl.Rectangle([0, 0], [1, 1]))
+    f_default = fspace.element(myfunc, vectorized=False)
+    f_numba = fspace.element(myfunc_vec)
+    f_native = fspace.element(myfunc_native_vec, vectorized=True)
+
+    # Create points
+    points = odl.uniform_sampling(fspace.domain, [2000, 2000]).points().T
+    mesh = odl.uniform_sampling(fspace.domain, [2000, 2000]).meshgrid()
+
+    print('Non-Vectorized runtime (points):     {:5f}'
```
adler-j: spacing is wrong here, should fix so they all align.

examples/vectorization.py
```diff
@@ -0,0 +1,108 @@
...
+        return np.where(x[0] > x[1], x[0] - x[1], x[0] + x[1])
+
+    # Create functions
+    fspace = odl.FunctionSpace(odl.Rectangle([0, 0], [1, 1]))
+    f_default = fspace.element(myfunc, vectorized=False)
+    f_numba = fspace.element(myfunc_vec)
+    f_native = fspace.element(myfunc_native_vec, vectorized=True)
+
+    # Create points
+    points = odl.uniform_sampling(fspace.domain, [2000, 2000]).points().T
```
adler-j: Comment on .T

odl/util/vectorization.py
kohr-h: I don't think there is anything standard doing this, but I haven't searched very intensely either. This functionality could also be implemented with the help of the [wrapt](https://pypi.python.org/pypi/wrapt) module, but I find its syntax slightly confusing, and we don't need its full power.

I'll factor out the pattern.

examples/vectorization.py
```diff
@@ -0,0 +1,108 @@
...
+    # Numba expects functions f(x1, x2, x3, ...), while we have the
+    # convention f(x) with x = (x1, x2, x3, ...). Therefore we need
+    # to wrap the Numba-vectorized function.
+    vectorized = numba.vectorize(lambda x, y: x - y if x > y else x + y)
+
+    def myfunc_vec(x):
+        """Return x - y if x > y, otherwise return x + y."""
+        return vectorized(*x)
+
+    def myfunc_native_vec(x):
```
adler-j: Add a comment on this

odl/util/vectorization.py
kohr-h: Sure can do.

odl/util/vectorization.py
```diff
@@ -124,169 +125,240 @@ def out_shape_from_array(arr):
...
+        """
+        # Decorating without arguments: return wrapper w/o args directly
+        instance = super().__new__(cls)
+
+        if (not kwargs and
+                len(args) == 1 and
+                callable(args[0])):
+            func = args[0]
+
+            return instance._wrapper(func)
```
adler-j: How is `instance._wrapper` set to anything here?

odl/util/vectorization.py
adler-j: Great!

examples/vectorization.py
kohr-h: Absolutely. Actually I was lazy and just copied your code. I didn't even test it since I don't have Numba on my machine here :grin:

examples/vectorization.py
kohr-h: OK

examples/vectorization.py
kohr-h: Yeah

examples/vectorization.py
kohr-h: Copy-paste as usual

odl/util/vectorization.py
kohr-h: Well, the instance is created before, so it has a `_wrapper` method. It's somewhat dirty since the instance is not yet initialized, so anything relying on a state of the instance would fail. But `_wrapper` doesn't - in fact, it doesn't even use its `self` argument, so the proper solution would be to make it static.

kohr-h: Or alternatively, make a free function.

adler-j: Now I see. The easiest would clearly be to simply make the `_wrapper` a static/free method indeed.

odl/util/vectorization.py
```diff
@@ -124,169 +125,240 @@ def out_shape_from_array(arr):
...
+        return self._wrapper(func, *self.vect_args, **self.vect_kwargs)
+
+    def _wrapper(self, func, *vect_args, **vect_kwargs):
+        """Return the vectorized wrapper function."""
+        return wraps(func)(_VectWrapper(func, *vect_args, **vect_kwargs))
+
+
+class _VectWrapper(object):
+
+    """Class for vectorization wrapping."""
```
kohr-h: I should comment on its purpose, I guess.

odl/util/vectorization.py
kohr-h: Yes. I prefer to make it static since there is hardly any other class making use of this guy.

adler-j:
Any chance to get the last things done so we can merge?

kohr-h:
Yeah, I'll do it right away.

kohr-h:
Done all the changes and Travis is happy. Merge?

adler-j:
Sure!

PR #199, opened 2016-01-04
STY: Fix many pylint warnings, add good names and remove local disables

### LINKED ISSUE #194: Reduce Pylint warnings
We have close to 1000 Pylint errors and warnings now. Some are just a result of Pylint not being able to understand the code properly. Others are things we don't want to change like variable names `x` or `n`. But then there are also relevant warnings where we should take a closer look.

I suggest that we
- globally disable errors and warnings where Pylint simply screws up,
- add variable names we consider okay to the list of accepted ones and change others to something better,
- try to resolve the remaining relevant warnings.

kohr-h:
Reducing the errors and warnings as far as possible. Remaining stuff mainly consists in
- TODOs
- too many arguments / branches / local variables / return statements
- Imports at the wrong place
- Warnings in `__init__.py` files

Total number of warnings is now ~500, which is still a lot, but acceptable given the list above, which contributes over 90 %, I'd say.

Closes #194

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -47,9 +47,10 @@ disable=redefined-builtin,
   no-member,
   missing-super-argument,
   abstract-method,
-  protected-access
+  protected-access,
+  not-callable
 
-# redefined-builtin: 'object', 'range', 'str', 'super' are redefined, 
+# redefined-builtin: 'object', 'range', 'str', 'super' are redefined,
 #                    that's okay
 # no-name-in-module, no-member: triggered by scipy and numpy, let's ignore them
 # missing-super-argument: importing new super, pylint doesn't get it
@@ -57,6 +58,8 @@ disable=redefined-builtin,
 #                  abstract methods. This warning does not really prevent bugs, so we're disabling it.
 # protected-access: We use private attributes of "internalized" objects. Since this check is not really
 #                   a bug catcher, we're disabling it.
+# not-callable: pylint tries to determine if a class attribute (e.g. element_type) is
+#               callable or not - and fails to do so
 
 [REPORTS]
 
@@ -131,10 +134,10 @@ indent-after-paren=4
 bad-functions=map,filter,apply,input
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_,a,b,x,y,z,xp,yp,zp,s,q,p,I,T,dx,dy,x1,x2,xi,yi,n,m,ax
+good-names=a,b,c,d,f,g,h,i,j,k,m,n,p,q,r,s,t,u,v,w,x,y,z,xp,yp,zp,dx,dy,dz,x0,x1,x2,y0,y1,y2,z0,z1,z2,xi,yi,zi,ax,rn,cn,op,ex,_,I,T,nx,ny,nz,fx
 
 # Bad variable names which should always be refused, separated by a comma
-bad-names=foo,bar,baz,toto,tutu,tata
+bad-names=foo,baz,toto,tutu,tata,l
 
 # Colon-delimited sets of names that determine each other's naming style when
 # the name regexes allow several styles.
--- a/odl/diagnostics/operator.py
+++ b/odl/diagnostics/operator.py
@@ -35,6 +35,7 @@ __all__ = ('OperatorTest',)
 
 
 class OperatorTest(object):
+
     """Automated tests for `Operator` implementations.
 
     This class allows users to automatically test various
@@ -78,7 +79,7 @@ class OperatorTest(object):
         print('\n== Calculating operator norm ==\n')
 
         operator_norm = 0.0
-        for [n_x, x] in samples(self.operator.domain):
+        for _, x in samples(self.operator.domain):
             result = self.operator(x)
             x_norm = x.norm()
             estimate = 0 if x_norm == 0 else result.norm() / x_norm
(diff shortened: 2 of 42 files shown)


.pylintrc
```diff
@@ -121,10 +134,10 @@ indent-after-paren=4
 bad-functions=map,filter,apply,input
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_,a,b,x,y,z,xp,yp,zp,s,q,p,I,T,dx,dy,x1,x2
+good-names=a,b,c,d,f,g,h,i,j,k,m,n,p,q,r,s,t,u,v,w,x,y,z,xp,yp,zp,dx,dy,dz,x0,x1,x2,y0,y1,y2,z0,z1,z2,xi,yi,zi,ax,rn,cn,op,ex,_,I,T,nx,ny,nz,fx
 
 # Bad variable names which should always be refused, separated by a comma
-bad-names=foo,bar,baz,toto,tutu,tata
```
adler-j: Why is bar removed?

odl/diagnostics/operator.py
```diff
@@ -167,18 +167,18 @@ def _adjoint_of_adjoint(self):
 
         with FailCounter('error = ||Ax - (A^T)^T x|| / '
                          '||A|| ||x||') as counter:
-            for [n_x, x] in vector_examples(self.operator.domain):
-                A_result = self.operator(x)
-                ATT_result = self.operator.adjoint.adjoint(x)
+            for [name_x, x] in vector_examples(self.operator.domain):
+                opx = self.operator(x)
+                transp_transp_x = self.operator.adjoint.adjoint(x)
```
adler-j: we should use `adj`/`adjoint` consistently, helps users if we stick to one.

`op_adj_adj_x`

odl/discr/lp_discr.py
```diff
@@ -17,10 +17,6 @@
 
 """:math:`L^p` type discretizations of function spaces."""
 
-# TODO: write some introduction doc
```
adler-j: Status on this?

odl/set/sets.py
```diff
@@ -210,16 +209,21 @@ def __init__(self, length):
...
+        self._length = length_
 
     @property
     def length(self):
         """The length attribute."""
         return self._length
 
+    @property
+    def ndim(self):
+        """Fake dim 1, needed for this to work as domain."""
```
adler-j: Better than saying that it is fake, please add a work on why it should be 1 for this to work.

odl/util/phantom.py
```diff
@@ -30,7 +30,10 @@
 
 
 def _shepp_logan_ellipse_2d():
-    # Modified Shepp Logan
+    """Create a 2d Shepp-Logan phantom.
```
adler-j: Create is wrong, id say "Return ellipses for a 2d Shepp-Logan phantom."

odl/util/phantom.py
```diff
@@ -44,7 +47,11 @@ def _shepp_logan_ellipse_2d():
 
 
 def _modified_shepp_logan_ellipse_2d():
-    # Modified Shepp Logan
+    """Create a 2d modified Shepp-Logan phantom.
```
adler-j: See above

odl/util/phantom.py
```diff
@@ -58,9 +65,10 @@ def _modified_shepp_logan_ellipse_2d():
 
 
 def _derenzo_sources_2d():
-    """ A popular phantom in spect/pet.
+    """Create a 2d Derenzo sources phantom.
```
adler-j: Above

odl/util/phantom.py
```diff
@@ -144,6 +152,10 @@ def _derenzo_sources_2d():
 
 
 def _shepp_logan_ellipse_3d():
+    """Create a 3d Shepp-Logan phantom.
```
adler-j: Above

odl/util/phantom.py
```diff
@@ -157,6 +169,11 @@ def _shepp_logan_ellipse_3d():
 
 
 def _modified_shepp_logan_ellipse_3d():
+    """Create a 3d Shepp-Logan phantom.
```
adler-j: Above

odl/util/phantom.py
```diff
@@ -372,9 +391,9 @@ def derenzo_sources(space):
 
 
 def shepp_logan(space, modified=True):
-    """ Create a Shepp-Logan phantom.
+    """Create a Shepp-Logan phantom.
 
-    The shepp-logan phantom
+    The standard Shepp-Logan phantom in 2 or 3 dimensions.
```
adler-j: standard is missleading since we default to modified. Perhaps we should change that instead?

odl/util/testutils.py
```diff
@@ -153,10 +150,12 @@ def all_almost_equal(iter1, iter2, places=None):
 
 
 def is_subdict(subdict, dictionary):
+    """`True` if all items of ``subdict`` are in ``dictionary``."""
     return all(item in dictionary.items() for item in subdict.items())
 
 
 def _pass(function):
+    """Trivial decorator."""
```
adler-j: used for ...

test/operator/difference_test_cuda.py
```diff
@@ -74,8 +74,8 @@ def adjoint(self):
 
 
 class ForwardDiff2DAdjoint(odl.Operator):
```
adler-j: Related note: This code is age old and should be moved and updated.

adler-j:
Truly doing gods work here!

Some minor stuff, fix and pull.

.pylintrc
kohr-h: Because it's used in `ProgressBar` :-)

odl/diagnostics/operator.py
kohr-h: :+1:

.pylintrc
adler-j: So these apply also to parts of words? Remove them all for now then.

odl/discr/lp_discr.py
kohr-h: There is a math-style introduction, but implementation guide is still missing. I'll make an issue.

.pylintrc
kohr-h: No, they only apply to full names. `ProgressBar` has a `bar` attribute.

odl/set/sets.py
kohr-h: Actually, this should not be part of the PR. I'll remove it.

.pylintrc
adler-j: Ahh makes sense, fine then.

odl/util/phantom.py
kohr-h: Yes, I think so in fact. Modified is not standard.

PR #210, opened 2016-01-11
'Submarine' phantom

kohr-h:
Trivial PR, just adding a nice new phantom.

### DIFF
--- a/odl/util/phantom.py
+++ b/odl/util/phantom.py
@@ -401,17 +401,25 @@ def shepp_logan(space, modified=False):
 def submarine_phantom(discr, smooth=False):
     """Return a 'submarine' phantom consisting in an ellipsoid and a box.
 
+    This phantom is used in [1]_ for shape-based reconstruction.
+
     Parameters
     ----------
     discr : `Discretization`
-        Space in which the phantom is supposed to be created
+        Discretized space in which the phantom is supposed to be created
     smooth : `bool`, optional
-        If `True`, the boundaries are smoothed out, otherwise, the
+        If `True`, the boundaries are smoothed out. Otherwise, the
         function steps from 0 to 1 at the boundaries.
 
     Returns
     -------
     phantom : `LinearSpaceVector`
+
+    References
+    ----------
+    .. [1] Oktem, Ozan. Mathematics of electron tomography. In:
+       Handbook of Mathematical Methods in Imaging. Scherzer, Otmar,
+       Ed. Springer, 2015, pp 937--1031.
     """
     if discr.ndim == 2:
         if smooth:
--- a/odl/util/phantom.py
+++ b/odl/util/phantom.py
@@ -355,7 +355,7 @@ def _phantom_3d(space, ellipses):
 
 
 def phantom(space, ellipses):
-    """ Return a phantom given by ellipses. """
+    """Return a phantom given by ellipses."""
 
     if space.ndim == 2:
         return _phantom_2d(space, ellipses)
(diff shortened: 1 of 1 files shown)


odl/util/phantom.py
```diff
@@ -398,6 +398,98 @@ def shepp_logan(space, modified=False):
...
+
+    def blurred_ellipse(x):
+        sq_ndist = np.zeros_like(x[0])
+        radii = (2, 0.7)
+        center = (0.5, -1)
+
+        for xi, rad, cen in zip(x, radii, center):
+            sq_ndist = sq_ndist + ((xi - cen) / rad) ** 2
+
+        out = np.sqrt(sq_ndist)
```
adler-j: No need to optimize this unless you optimize the for loop really

odl/util/phantom.py
```diff
@@ -398,6 +398,98 @@ def shepp_logan(space, modified=False):
...
+        for xi, rad, cen in zip(x, radii, center):
+            sq_ndist = sq_ndist + ((xi - cen) / rad) ** 2
+
+        return np.where(sq_ndist <= 1, 1, 0)
+
+    def rect(x):
+        out = np.ones_like(x[0])
+        xlower, xupper = (.3, -.5), (1.3, .5)
+        for xi, low, upp in zip(x, xlower, xupper):
+            out = out * np.where(np.logical_and(xi >= low, xi <= upp), 1, 0)
```
adler-j: Well short note:

```python
out = out * np.where(np.logical_and(xi >= low, xi <= upp), 1, 0)
```

I would probably use

```python
out *= (xi >= low) & (xi <= upp)
```

odl/util/phantom.py
```diff
@@ -398,6 +398,98 @@ def shepp_logan(space, modified=False):
...
+        return out
+
+    out = discr.element(ellipse)
+    out += discr.element(rect)
+    return out.ufunc.minimum(1)
```
adler-j: Since you are somewhat careful with performance you could:

```python
 return out.ufunc.minimum(1, out=out)
```

odl/util/phantom.py
```diff
@@ -410,6 +502,10 @@ def shepp_logan(space, modified=False):
     shepp_logan(discr, modified=False).show()
     derenzo_sources(discr).show()
 
+    discr = odl.uniform_discr([-2.5, -2.5], [2.5, 2.5], [n, n])
```
adler-j: This is prolly the only "real" complaint. These phantoms should (imo) work on arbitrary domains, i.e. not only ones that fit in `[-2.5, -2.5] x [2.5, 2.5]`. You may want to rescale the points so this works. This is how the other phantoms work.

odl/util/phantom.py
```diff
@@ -398,6 +398,98 @@ def shepp_logan(space, modified=False):
...
+    ----------
+    discr : `Discretization`
+        Discretized space in which the phantom is supposed to be created
+    smooth : `bool`, optional
+        If `True`, the boundaries are smoothed out. Otherwise, the
+        function steps from 0 to 1 at the boundaries.
+
+    Returns
+    -------
+    phantom : `LinearSpaceVector`
```
adler-j: `DiscretizationVector`?

odl/util/phantom.py
```diff
@@ -398,6 +398,98 @@ def shepp_logan(space, modified=False):
...
+    else:
+        raise ValueError('Phantom only defined in 2 dimensions, got {}.'
+                         ''.format(discr.dim))
+
+
+def _submarine_phantom_2d_smooth(discr):
+    """Return a 2d smooth 'submarine' phantom."""
+
+    # Smaller value = more blurring
+    c = 20
```
adler-j: Make public parameter?

adler-j:
Minor comments

odl/util/phantom.py
kohr-h: That won't work.

odl/util/phantom.py
kohr-h: That I can do, sure.

odl/util/phantom.py
adler-j: Why?

kohr-h: The size of `out`, i.e. the number of "fleshed out" dimensions, grows over time. [See here](https://github.com/odlgroup/odl/blob/master/odl/discr/discr_mappings.py#L981-L990) for the same principle. At least for the interpolation code, this version is about 1.5 times faster.

kohr-h: Alternatively I would have to allocate a big array from the beginning and work with that. Here I start with the size of the first element, which is essentially one-dimensional, so operations are basically for free.

odl/util/phantom.py
kohr-h: OK

odl/util/phantom.py
kohr-h: Right.

odl/util/phantom.py
adler-j: Sure, `out = out * ((xi >= low) & (xi <= upp))` then.

kohr-h: Which doesn't keep me from simplifying the right hand side, of course.

odl/util/phantom.py
kohr-h: We can do that, sure. Then it gets a nicer name.

adler-j:
Please add submarine_phantom to `__all__`

kohr-h:
All done.

adler-j:
Should be fine then, merge!

Issue #195, opened 2016-01-04
FunctionSetVector vectorization not working in all situations


kohr-h:
The vectorization input checks rely on `domain.ndim` to exist, which is, of course, the case for `NtuplesBase` spaces, but not in other cases. The same is true for output checks with `range.ndim`.

Either we work around this issue by equipping our `Set`s with `ndim` by default, which is a bit ugly and adds unnecessary complexity. Or we change the checks to something like
```python
# In FunctionSetVector.__call__
ndim = getattr(domain, 'ndim', None)


# in odl.util.vectorization.py
def is_valid_input_array(x, ndim=None):  # ndim optional
    if not isinstance(x, np.ndarray):
        return False
    if ndim is None or ndim == 1:  # treat None like 1d
        return x.ndim == 1 or x.ndim == 2 and x.shape[0] == 1
    else:
        return x.ndim == 2 and x.shape[0] == ndim
```
Meshgrids don't really make much sense for non-scalar types, so I wouldn't necessarily adapt that helper.

A failing test looks like this:
```python
def test_fset_vector_eval():
    str3 = odl.Strings(3)
    ints = odl.Integers()
    fset = FunctionSet(str3, ints)
    strings = np.array(['aa', 'b', 'cab', 'aba'])
    out_vec = np.empty((4,), dtype=int)

    # Vectorized for arrays only
    f_vec = fset.element(lambda s: np.array([str(si).count('a') for si in s]))
    true_vec = [2, 0, 1, 2]

    assert f_vec('abc') == 1
    assert all_equal(f_vec(strings), true_vec)
    f_vec(strings, out=out_vec)
    assert all_equal(out_vec, true_vec)
```
The failure comes with this message:
```
...
        # Check for input type and determine output shape
>       if is_valid_input_array(x, self.domain.ndim):
E       AttributeError: 'Strings' object has no attribute 'ndim'
```

adler-j:
> Either we work around this issue by equipping our Sets with ndim by default, which is a bit ugly and adds unnecessary complexity.

I agree, especially since dimension would need some more information in general (C is 2-dimensional over the reals).

> Or we change the checks to something like ...

Agree with this approach.

PR #215, opened 2016-01-13
BUG: Fix broken vectorized raw FunctionSetVector, see #195

kohr-h:
This fixes the issue with the vectorization of `FunctionSetVector`'s which are not `FunctionSpaceVector`'s. I followed the suggested fix in #195. Additionally I had to handle the case `ndim=None` also in the `is_valid_input_meshgrid` function so that it doesn't crash. Now it just returns `False`.

Just saw that the test has a duplication. Removing that.


adler-j:
Looks good to me.

Issue #222, opened 2016-01-16
Tests for STIR bindings


kohr-h:
We currently don't have any tests for the STIR bindings. The interface is still rudimentary, but at least some very simple tests should be possible. Our Jenkins CI has a build job for it which successfully compiles STIR (at least on Python 2.7, on 3.4 their Numpy detection is broken), so it's integrated into our CI flow.

adler-j:
We do have an example, perhaps adding a slow test for STIR that makes sure it can be used to reconstruct some volume to decent precision would be nice?

kohr-h:
That would indeed be nice. Maybe it's even possible to generate a very small example which runs fast enough for a unit test? Otherwise, a test mock would help. Right now STIR bindings have zero test coverage, which is somewhat retarded.

adler-j:
A mock could do but thats damn complicated, Ill likely just make a slow test for this.

adler-j:
Added some, but made it a pull into conebeam since we may want to do this for astra and others as well.

PR #224, opened 2016-01-18
Add some largescale tests

adler-j:
Add some slow tests for STIR and ASTRA. See #222.

This is a merge into conebeam since it is our de-facto master until it gets merged.

TODO: 

- [ ] Handle largescale tests properly
- [ ] Perhaps add the previously removed largescale test.


examples/tomo/stir_reconstruct.py
```diff
@@ -26,13 +26,13 @@
...
 # Set path to input files
-base = pth.join(pth.join(pth.dirname(pth.abspath(__file__)), 'data'), 'stir')
+base = pth.join(pth.dirname(pth.abspath(__file__)), 'data', 'stir')
 volume_file = str(pth.join(base, 'initial.hv'))
 projection_file = str(pth.join(base, 'small.hs'))
 
 # Create a STIR projector from file data.
-proj = odl.tomo.stir_bindings.stir_projector_from_file(volume_file,
-                                                       projection_file)
+proj = odl.tomo.backends.stir_bindings.stir_projector_from_file(volume_file,
```
kohr-h: Break immediately after opening parenthesis, then indent by four.

examples/tomo/xray_trafo.py
```diff
@@ -151,7 +151,8 @@ def conebeam():
     dgrid = odl.uniform_sampling(dparams, [558, 558])
     geom = odl.tomo.CircularConeFlatGeometry(angle_intvl, dparams,
                                              src_radius=1000, det_radius=100,
-                                             agrid=agrid, dgrid=dgrid)
+                                             agrid=agrid, dgrid=dgrid,
```
kohr-h: See above, not necessary here, but would not look as crammed as now.

test/largescale/tomo/astra_slow_test.py
```diff
@@ -0,0 +1,142 @@
...
+    # Calculate operator norm for landweber
+    op_norm_est_squared = projector.adjoint(projections).norm() / vol.norm()
+    omega = 0.5 / op_norm_est_squared
+
+    # Reconstruct using ODL
+    recon = projector.domain.zero()
+    odl.solvers.landweber(projector, recon, projections, niter=100,
+                          omega=omega)
+
+    assert recon.dist(vol) < vol.norm() / 3.0
```
kohr-h: Where does the magic 3.0 come from?

adler-j: Its a number larger than 1, i.e. the error is somehow decreased. 3.0 seems to be about what we get to so I picked that. This is mostly to see if it somehow becomes worse. Perhaps add a comment.

adler-j:
Added some more to this in large scale tests, primarily one for cu_ntuples. I guess we can add more.

examples/tomo/stir_reconstruct.py
adler-j: I'll probably do a `from odl.tomo.backends.stir_bindings import stir_projector_from_file` instead.

kohr-h:
All good implementation-wise. If you want to integrate the benchmark for interpolation, look in the `issue-128__performance_tests` branch.

I'm wondering how to integrate this into our Jenkins CI landscape. Currently a push to a branch (not master) triggers a bunch of builds, where the most time-consuming ones are those which involve compiling stuff:

- ~6 min for odlpp
- ~10 min for astra (which includes odlpp)
- ~18 min for stir (no odlpp)

Multiplied by 2 due to multiple Python versions, of course. Adding large-scale and benchmark tests to the regular work flow could easily choke the whole server with jobs if multiple people are pushing to different branches.

Maybe it's better to keep the slow tests in the queue until some time during night and let them run when no other jobs are in the way. Or not trigger them by push events at all but just build them on a regular basis like the nightly builds. Maybe we need to put up a second machine to share jobs. What do you think?

test/largescale/tomo/astra_slow_test.py
kohr-h: Yeah I guessed that, a comment would definitely help.

adler-j:
> All good implementation-wise. If you want to integrate the benchmark for interpolation, look in the issue-128__performance_tests branch.

Will do

> I'm wondering how to integrate this into our Jenkins CI landscape. Currently a push to a branch (not master) triggers a bunch of builds, where the most time-consuming ones are those which involve compiling stuff:
    ~6 min for odlpp
    ~10 min for astra (which includes odlpp)
    ~18 min for stir (no odlpp)
    Multiplied by 2 due to multiple Python versions, of course. Adding large-scale and benchmark tests to the regular work flow could easily choke the whole server with jobs if multiple people are pushing to different branches.
    
Why do we need to recompile these if they are pointing to the same hash, as they are like 90% of the time? Perhaps a quick check for that?

> Maybe it's better to keep the slow tests in the queue until some time during night and let them run when no other jobs are in the way. Or not trigger them by push events at all but just build them on a regular basis like the nightly builds. Maybe we need to put up a second machine to share jobs. What do you think?

Yes we should have a separate build for this, imo simply make it part of the nightly build.

kohr-h:
> Why do we need to recompile these if they are pointing to the same hash, as they are like 90% of the time? Perhaps a quick check for that?

Yeah, I think that's possible. I already do something similar for the `master` branch, just without the hash check. That would be the cleaner solution, now that there is one build per branch. I'll have a look into it.

> Yes we should have a separate build for this, imo simply make it part of the nightly build.

I'll do that anyway, but on the other hand, separate builds for slow tests are needed at least for feature branches, e.g. if you want to have a performance figure for your new feature on the branch right away. You could then simply trigger the build manually and look at the data some time later (or just run the stuff at your machine, assuming everything is installed).
It's not going to be much extra work, so what the heck.

adler-j:
> Yeah, I think that's possible. I already do something similar for the master branch, just without the hash check. That would be the cleaner solution, now that there is one build per branch. I'll have a look into it.

Great

> I'll do that anyway, but on the other hand, separate builds for slow tests are needed at least for feature branches, e.g. if you want to have a performance figure for your new feature on the branch right away. You could then simply trigger the build manually and look at the data some time later (or just run the stuff at your machine, assuming everything is installed).
It's not going to be much extra work, so what the heck.

Sure, but dont overdo it. I came home to 20 mails from Jenkins now, adding more wont help I think.

kohr-h:
>> Yeah, I think that's possible. I already do something similar for the master branch, just without the hash check. That would be the cleaner solution, now that there is one build per branch. I'll have a look into it.
> 
> Great

Mostly done now, just need to adapt the current jobs, seems to work. 


>> I'll do that anyway, but on the other hand, separate builds for slow tests are needed at least for feature branches, e.g. if you want to have a performance figure for your new feature on the branch right away. You could then simply trigger the build manually and look at the data some time later (or just run the stuff at your machine, assuming everything is installed).
It's not going to be much extra work, so what the heck.
>
> Sure, but dont overdo it. I came home to 20 mails from Jenkins now, adding more wont help I think.

Just delete them, who cares? Seriously, there are lots of mails now since I'm redoing stuff, and *all* currently existing branches are being built, some of them failing for various reasons. And in addition, I'm testing stuff and mostly forget to turn off mail notification :wink:.

adler-j:
Did some more work, improved performance of the `all_almost_equal` thing when both parts have `__array__`, this also made the regular test suite about 1 second faster.

adler-j:
Added the option to do `py.test --largescale` to run the largescale tests, otherwise they are skipped. Quite happy about this now, how about a merge?

odl/util/phantom.py
```diff
@@ -161,6 +161,16 @@ def _derenzo_sources_2d():
...
 
 
+def _make_3d_cylinders(ellipses2d):
+    """ Create 3d cylinders from ellipses """
+    ellipses2d = np.asarray(ellipses2d)
+    ellipses3d = np.zeros((ellipses2d.shape[0], 10))
+    ellipses3d[:, [0, 1, 2, 4, 5, 7]] = ellipses2d
+    ellipses3d[:, 3] = 100000.0
+
+    return ellipses3d
```
kohr-h: Did you update the doc of the Derenzo phantom accordingly?

odl/util/testutils.py
```diff
@@ -115,6 +115,21 @@ def all_equal(iter1, iter2):
...
+        v1 = v1.__array__().ravel(v1.order)
+    else:
+        v1 = v1.__array__()
+
+    if hasattr(v2, 'order'):
+        v2 = v2.__array__().ravel(v2.order)
+    else:
+        v2 = v2.__array__()
+
+    return np.all(np.isclose(v1, v2, rtol=10 ** (-places), equal_nan=True))
```
kohr-h: If `equal_nan=True` is not essential, you could use `np.allclose`.

test/largescale/space/fn_base_slow_test.py
```diff
@@ -0,0 +1,306 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+
```
kohr-h: One-line docstring

test/largescale/space/fn_base_slow_test.py
```diff
@@ -0,0 +1,306 @@
...
+    x0 = np.arange(fn.size)
+    x = fn.element(x0)
+
+    for index in indices:
+        x[index] = -index
+        assert x[index] == -index
+
+
+@pytest.mark.xfail  # Expected to fail since inner scaling is not public
```
kohr-h: Use `xfail(reason='...')` instead of comment

test/largescale/tomo/astra_slow_test.py
```diff
@@ -0,0 +1,168 @@
...
+        angle_intvl = odl.Interval(0, 2 * np.pi)
+        dparams = odl.Interval(-30, 30)
+        agrid = odl.uniform_sampling(angle_intvl, 200)
+        dgrid = odl.uniform_sampling(dparams, 200)
+        geom = tomo.FanFlatGeometry(angle_intvl, dparams,
+                                    src_radius=200, det_radius=100,
+                                    agrid=agrid, dgrid=dgrid)
+
+        # X-ray transform
+        return tomo.DiscreteXrayTransform(discr_reco_space, geom,
```
kohr-h: Slightly off-topic, but we shouldn't use this operator for non-parallel cases.

test/largescale/tomo/astra_slow_test.py
```diff
@@ -0,0 +1,168 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test reconstruction with stir."""
```
kohr-h: ASTRA

test/largescale/tomo/stir_slow_test.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test reconstruction with stir."""
```
kohr-h: Maybe we should capitalize STIR (and ASTRA) if we mean the package in general.

test/largescale/space/fn_base_slow_test.py
```diff
@@ -0,0 +1,306 @@
...
+# ODL imports
+import odl
+
+from odl.util.testutils import all_almost_equal, almost_equal
+
+pytestmark = pytest.mark.skipif(
+    "not pytest.config.getoption('--largescale')",
+    reason='Need --largescale option to run'
+)
+
```
kohr-h: I like this option thing. In fact, I think it's so useful that this decorator (along with its benchmark counterpart) should be put into `testutils`.

odl/util/testutils.py
adler-j: It is essential since we rely on it in some tests. This is how allclose is implemented anyway.

test/largescale/tomo/astra_slow_test.py
adler-j: Make a comment on it in the branch then!

odl/util/phantom.py
adler-j: now did!

test/largescale/tomo/astra_slow_test.py
kohr-h: Done

odl/util/phantom.py
```diff
@@ -369,9 +379,14 @@ def derenzo_sources(space):
     """Create the PET/SPECT Derenzo sources phantom.
 
     The Derenzo phantom contains a series of circles of decreasing size.
+
+    In 3d the phantom is simply the 2d phantom extended in the z direction as
+    ellipses.
```
kohr-h: cylinders?

odl/util/phantom.py
```diff
@@ -161,6 +161,16 @@ def _derenzo_sources_2d():
             [1.0, 0.023968, 0.023968, 0.88528, -0.11791, 0.0]]
 
 
+def _make_3d_cylinders(ellipses2d):
+    """ Create 3d cylinders from ellipses """
```
kohr-h: Remove spaces, add dot

kohr-h:
Two slight things, merge when done.

Issue #175, opened 2015-12-17
Switch argument order of insert methods


kohr-h:
The index parameter of the `insert` methods of `IntervalProd` and `TensorGrid` comes second, whereas the corresponding method of core Python lists take the index parameter first. We should mirror that signature.

adler-j:
Should be a simple change, please go ahead. Some additional suggestions, probably fit for the same branch, would be

Append and extend
-----------------

Currently the following is valid

```python
>>> rbox = IntervalProd(0, 1)
>>> rbox.insert(IntervalProd(2, 3))
IntervalProd([0.0, 2.0], [1.0, 3.0])
```

This would clearly fail with the proposed change. To remedy this, I propose that we add `IntervalProd.append` and `IntervalProd.extend` defined as

```python
def append(self, item):
    return self.insert(self.ndim, item)
    
def extend(self, seq):
    result = self
    for item in seq:
        result = result.append(item)
    return result
```

this would give us similar syntax to `list`.

Simplified insert
-----------------

Currently we have an auto-casting syntax given by [casting the input argument](https://github.com/odlgroup/odl/blob/1a0c8385563783807531f63ef11c691663081854/odl/set/domain.py#L539-L540)

```python
if not isinstance(other, IntervalProd):
    other = IntervalProd(other, other)
```

Imo, this is not very useful and would probably not be used much. I propose we make the following small change:

```python
if not isinstance(other, IntervalProd):
    other = IntervalProd(*other)
```

This would allow the following syntax:

```python
>>> rbox = IntervalProd(0, 1)
>>> rbox.insert(1, [2, 3])
IntervalProd([0.0, 2.0], [1.0, 3.0])
```

PR #235, opened 2016-01-26
ENH: fix order of insert, add append method in grid and intervalprod,…

adler-j:
Very simple pull request, fixes an inconsistiency given by issue #175.

### DIFF
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -417,7 +417,7 @@ class TensorGrid(Set):
 
         return True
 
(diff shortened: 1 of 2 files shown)


kohr-h:
Looks good. I did something similar on the partitions branch, but I'll just merge then :-)

adler-j:
Small branches! We need a mantra about that.

kohr-h:
Feels like a lot of work done? The build server has to work hard for it, though :-)

PR #237, opened 2016-01-26
Astra scaling

moosmann:
Pull request into the conebeam branch. ASTRA scaling issue are (hopefully) all fixed.


adler-j:
Great initiative, will review!

examples/tomo/astra_parallel_2d_fp.py
```diff
@@ -54,8 +54,8 @@ def phantom(x):
 geom = odl.tomo.Parallel2dGeometry(angle_range, det_range, angles, det_pixels)
 
 # Initialize the operator
-xray_trafo = odl.tomo.DiscreteXrayTransform(discr_reco_space, geom,
-                                        backend='astra')
+xray_trafo = odl.tomo.XrayTransform(discr_reco_space, geom,
```
adler-j: Good change, perhaps we do the same to gradient, divergence?

odl/__init__.py
```diff
@@ -54,3 +54,4 @@
 from . import solvers
 from . import tomo
 from . import trafos
+from . import util
```
adler-j: Not sure here, if so we need to remove testutils from utils.__all__ at least

odl/tomo/backends/astra_cpu.py
```diff
@@ -109,9 +111,13 @@ def astra_cpu_forward_projector_call(vol_data, geometry, proj_space, out=None):
     algo_id = astra_algorithm('forward', ndim, vol_id, sino_id, proj_id,
                               impl='cpu')
 
-    # Run algorithm and delete it
+    # Run algorithm
     astra.algorithm.run(algo_id)
 
+    # Fix scaling issue
+    # 2D CPU does not scale with line integration weight
+    # out *= float(vol_data.space.grid.stride[0])  # isotropic voxel size
```
adler-j: dead code?

odl/tomo/backends/astra_cpu.py
```diff
@@ -191,6 +197,18 @@ def astra_cpu_backward_projector_call(proj_data, geometry, reco_space,
     # Run algorithm and delete it
     astra.algorithm.run(algo_id)
 
+    # Fix scaling issue
+    # parallel2d and fanflat do scale quadratically with linear voxel size
+    # and linearly with inverse pixel size
+    sf = float(geometry.det_grid.stride[0])
+    sf /= float(reco_space.grid.stride[0]) ** 2
+    # magnification factor
```
adler-j: What about the cone beam case?

odl/tomo/backends/astra_cpu.py
```diff
@@ -191,6 +197,18 @@ def astra_cpu_backward_projector_call(proj_data, geometry, reco_space,
...
+    if isinstance(geometry, FanFlatGeometry):
+        src_radius = geometry.src_radius
+        det_radius = geometry.det_radius
+        sf /= ((src_radius + det_radius) / src_radius)
+    out *= sf
```
adler-j: Good implementation

odl/tomo/backends/astra_cuda.py
```diff
@@ -123,6 +123,11 @@ def astra_cuda_forward_projector_call(vol_data, geometry, proj_space,
         else:
             out.assign(tmp)
 
+    # Fix scaling issue
+    if isinstance(geometry, Parallel2dGeometry):
+        # cuda parallel2d scales linearly with linear pixel stride
```
adler-j: What about non square pixels?

odl/tomo/backends/astra_cpu.py
```diff
@@ -191,6 +197,18 @@ def astra_cpu_backward_projector_call(proj_data, geometry, reco_space,
     # Run algorithm and delete it
     astra.algorithm.run(algo_id)
 
+    # Fix scaling issue
+    # parallel2d and fanflat do scale quadratically with linear voxel size
+    # and linearly with inverse pixel size
+    sf = float(geometry.det_grid.stride[0])
```
adler-j: non square pixels?

odl/tomo/operators/xray_trafo.py
```diff
@@ -107,35 +112,44 @@ def __init__(self, discr_dom, geometry, backend='astra', **kwargs):
...
-                                           weight=weight,
-                                           dtype=discr_dom.dspace.dtype)
-
-        ran_interp = kwargs.pop('range_interpolation', 'nearest')
-        discr_ran = DiscreteLp(ran_uspace, geometry.grid, ran_dspace,
-                               interp=ran_interp, order=geometry.grid.order)
-        super().__init__(discr_dom, discr_ran, linear=True)
-
-        self._adjoint = DiscreteXrayTransformAdjoint(self)
+        if isinstance(geometry, HelicalConeFlatGeometry):
```
adler-j: Why do we have weighting code in two places? Not sure If I agree with this, this would always be implementation dependent, no?

adler-j: Another comment. What about the general `DivergentBeam` case, it would give different results from the helical case, even though they should be equivalent. Can we solve that somehow?

odl/tomo/operators/xray_trafo.py
```diff
@@ -210,12 +230,16 @@ def _call(self, inp):
...
         back, impl = self.backend.split('_')
         if back == 'astra':
+            # angle interval weight
+            weight = float(self.forward.geometry.motion_grid.stride)
+            if weight <= 0:
+                weight = 1.0
             if impl == 'cpu':
-                return astra_cpu_backward_projector_call(
-                    inp, self.forward.geometry, self.range)
+                return weight * astra_cpu_backward_projector_call(
```
adler-j: `weight * result` causes a copy. How about `result = ...`, `result *= weight`

odl/tomo/operators/xray_trafo.py
```diff
@@ -210,12 +230,16 @@ def _call(self, inp):
         """
         back, impl = self.backend.split('_')
         if back == 'astra':
+            # angle interval weight
+            weight = float(self.forward.geometry.motion_grid.stride)
+            if weight <= 0:
```
adler-j: Stride is strictly positive?

test/tomo/backends/astra_cpu_test.py
```diff
@@ -31,9 +31,7 @@
 from odl.tomo.util.testutils import skip_if_no_astra
 
 
-# TODO: test scaling of ASTRA projectors
```
adler-j: Where are these tests? :)

test/tomo/operator/xray_trafo_test.py
```diff
@@ -30,28 +30,261 @@
...
+mag = (src_radius + det_radius) / src_radius
+
+# Slice index to print
+z_vol = np.round(discr_vol_space_3d.shape[2] / 2)
+y_proj = np.floor(dgrid_3d.shape[1] / 2)
+
+# Precision for adjoint test
+precision = 4
+
+# Print section
```
adler-j: Guess this is temporary code?

test/tomo/operator/xray_trafo_test.py
```diff
@@ -30,28 +30,261 @@
...
 from odl.util.testutils import almost_equal
+from odl.tomo import ASTRA_CUDA_AVAILABLE
+from odl.tomo.util.testutils import skip_if_no_astra_cuda
 
 
-@pytest.mark.xfail  # Expected to fail since scaling of adjoint is wrong.
-@pytest.mark.skipif("not odl.tomo.ASTRA_CUDA_AVAILABLE")
-def test_xray_trafo():
-    """Test discrete X-ray transformt using ASTRA with CUDA."""
+# TODO: increase test coverage and improve tests
```
adler-j: Look into the largescale tests, we have pretty much exactly this except we do tomography at the end. Simply change that part to an adjoint check and you should be fine?

adler-j:
Looks good, some minor things. Guess test needs a cleanup before merge.

examples/tomo/astra_parallel_2d_fp.py
moosmann: I thought about the same and would change it.

odl/tomo/backends/astra_cpu.py
moosmann: Yes, obsolete.

odl/__init__.py
moosmann: I can remove it again.

odl/tomo/backends/astra_cuda.py
moosmann: Good question. I realized that scaling behaviour at the very end and forgot to check that.

odl/__init__.py
adler-j: I think we should remove it tbh. util should be "internal" things. Public stuff should be in the others.

odl/tomo/operators/xray_trafo.py
moosmann: Right, that's redundant.

odl/tomo/operators/xray_trafo.py
moosmann: I'll change it.

test/tomo/backends/astra_cpu_test.py
moosmann: I moved the scaling tests to the Xray trafo tests in order to have them all (cpu + cuda) in one place. Nevertheless the tests have to be improved.

test/tomo/operator/xray_trafo_test.py
moosmann: Yes. This was to figure out the scaling behaviour. Once the rescaling is correct, I want to change the print statements into tests.

examples/tomo/astra_parallel_2d_fp.py
kohr-h: Yes, I did the same with FT and Kati with the Wavelets.

odl/tomo/backends/astra_cpu.py
```diff
@@ -36,12 +36,14 @@
                                            astra_volume_geometry, astra_data,
                                            astra_projector, astra_algorithm)
 from odl.tomo.geometry import Geometry
+from odl.tomo.geometry.fanbeam import FanFlatGeometry
 
 __all__ = ('astra_cpu_forward_projector_call',
            'astra_cpu_backward_projector_call')
 
 
 # TODO: Fix inconsistent scaling of ASTRA projector with pixel size
```
kohr-h: Done by this PR?

odl/tomo/backends/astra_cpu.py
kohr-h: What is `sf`? Better name?

odl/tomo/backends/astra_cpu.py
```diff
@@ -191,6 +197,18 @@ def astra_cpu_backward_projector_call(proj_data, geometry, reco_space,
     # Run algorithm and delete it
     astra.algorithm.run(algo_id)
 
+    # Fix scaling issue
+    # parallel2d and fanflat do scale quadratically with linear voxel size
+    # and linearly with inverse pixel size
```
kohr-h: I don't understand this sentence. Is it me? :-)

odl/tomo/backends/astra_cuda.py
kohr-h: What is linear about the pixel stride?

odl/tomo/backends/astra_cuda.py
```diff
@@ -209,6 +214,14 @@ def astra_cuda_backward_projector_call(proj_data, geometry, reco_space,
     # Run algorithm
     astra.algorithm.run(algo_id)
 
+    # Fix scaling issues
+    if isinstance(geometry, (FanFlatGeometry, Parallel2dGeometry)):
+        # cuda fanflat and parallel2d scale linearly with cell volume
+        out *= float(reco_space.cell_volume)
+    elif isinstance(geometry, (HelicalConeFlatGeometry, Parallel3dGeometry)):
+        # cuda cone and parallel3d scale linearly with linear voxel size
```
kohr-h: Same here. Linear?

odl/tomo/backends/astra_cpu.py
moosmann: Hmm. I hoped so and thus it should have been removed, but as Jonas realized I forgot to check non-isotropic pixels. So it's maybe still an issue.

odl/tomo/backends/stir_bindings.py
```diff
@@ -170,7 +168,7 @@ def adjoint(self):
 
 class BackProjectorByBinWrapper(Operator):
 
-    """A backprojector using STIR."""
+    """A back projector using STIR."""
```
kohr-h: I'm never sure which one is right, you see both often. Shall we agree on "back projector"? Let's be consistent then (also in class names!).

odl/tomo/geometry/conebeam.py
```diff
@@ -46,7 +46,7 @@ class HelicalConeFlatGeometry(DivergentBeamGeometry, AxisOrientedGeometry):
     the source and moves on a spiral with radius ``R`` in the azimuthal
     plane and pitch ``P``.
 
-    The motion parameter is the (1d) rotation angle parametrizing source and
+    The motion parameter is the (1d) rotation angle parameterizing source and
```
kohr-h: Old version was correct, too. But never mind.

odl/tomo/geometry/detector.py
```diff
@@ -240,7 +240,7 @@ def __init__(self, params, detector_axis, grid=None):
 
     @property
     def detector_axis(self):
-        """The directions of the principial axes of the detector."""
+        """The directions of the principal axes of the detector."""
```
kohr-h: Call it `detector_axes` then?

odl/tomo/backends/astra_cpu.py
moosmann: Should be clearer. That sentence changed a few times until I figured out how the rescaling should be done.

odl/tomo/operators/xray_trafo.py
```diff
@@ -39,19 +41,22 @@
 
 _SUPPORTED_BACKENDS = ('astra', 'astra_cpu', 'astra_cuda')
 
-__all__ = ('DiscreteXrayTransform', 'DiscreteXrayTransformAdjoint')
+__all__ = ('XrayTransform', 'XrayTransformAdjoint',)
```
kohr-h: Same question as several times before. Is the adjoint an operator in its own right? I think so. We should call it `XrayBackProjection` or something similar.

odl/tomo/operators/xray_trafo.py
```diff
@@ -39,19 +41,22 @@
...
 
 
-class DiscreteXrayTransform(Operator):
+# TODO: Check scaling with magnification
+# TODO: DivergentBeamTransform
+
+class XrayTransform(Operator):
 
     """The discrete X-ray transform between :math:`L^p` spaces."""
```
kohr-h: ~~:math:~~

odl/tomo/backends/stir_bindings.py
moosmann: I agree on "back projector". The concatenation of words is rather German thingy.

odl/tomo/operators/xray_trafo.py
```diff
@@ -59,7 +64,7 @@ def __init__(self, discr_dom, geometry, backend='astra', **kwargs):
             the operator range. It needs to have a sampling grid for
             motion and detector parameters.
         backend : {'astra', 'astra_cuda', 'astra_cpu'}, optional
-            Implementation backend for the transform. Supported backends:
+            Implementation back-end for the transform. Supported backends:
```
kohr-h: Good you corrected that. Do it also at the end of the line.

odl/tomo/geometry/conebeam.py
moosmann: I know. I just wanted to get it off my spell checker list.

odl/tomo/operators/xray_trafo.py
kohr-h: We need to go back and look into the stuff I wrote up. The weighting in this place is only with respect to the **parametrization**, i.e. if we happen to choose the parameter rectangle to coincide with the size of the (flat) detector, we get weight equal to the cell size in that parameter space. The exact same is true for curved detectors with the only difference that one or two of the parameters are angles - but they still lie on a rectangle. The shape of the detector is handled by the weight function defined in `Detector` (`surface_measure` I think it is). Finally, geometric relations like acquisition curve, radii etc. are handled by the back projector.

So, in conclusion: push this geometry related weighting to the back projector and (for later) weightings due to curvature to the `surface_measure` of the detector.

odl/tomo/operators/xray_trafo.py
```diff
@@ -70,9 +75,9 @@ def __init__(self, discr_dom, geometry, backend='astra', **kwargs):
                 Interpolation type for the discretization of the
```
kohr-h: We always abbreviate `interp` for interpolation, should be fine here, too.

test/tomo/backends/astra_cpu_test.py
```diff
@@ -117,7 +115,6 @@ def test_astra_cpu_projector_fanflat():
                                                    dtype='float32')
 
     # forward
-    discr_data = odl.util.phantom.cuboid(discr_vol_space, 0.5, 1)
     proj_data = odl.tomo.astra_cpu_forward_projector_call(discr_data, geom,
                                                           discr_proj_space)
     assert proj_data.shape == proj_shape
```
kohr-h: Cannot comment further below, doing it here.

The test `assert proj_data > 0` makes no sense at all and fails on Py3. Python 2 somehow makes sense of this comparison in some nonsense way.

test/tomo/backends/astra_setup_test.py
```diff
@@ -50,11 +50,11 @@ def _discrete_domain(ndim, interp):
     space : `DiscreteLp`
         Returns a `DiscreteLp` instance
     """
-    minpt = np.arange(-1, -ndim - 1, -1)  # -1, -2 [, -3]
-    maxpt = np.arange(1, ndim + 1, 1)  # 1, 2 [, 3]
+    min_pt = np.arange(-1, -ndim - 1, -1)  # -1, -2 [, -3]
+    max_pt = np.arange(1, ndim + 1, 1)  # 1, 2 [, 3]
```
kohr-h: You felt the need to comment in order to clarify, which is a red flag. What about `max_pt = np.arange(1, ndim + 1)` and `min_pt = -max_pt`?

test/tomo/backends/astra_setup_test.py
```diff
@@ -73,11 +73,11 @@ def _discrete_domain_anisotropic(ndim, interp):
     space : `DiscreteLp`
         Returns a `DiscreteLp` instance
     """
-    minpt = [-1] * ndim
-    maxpt = [1] * ndim
+    min_pt = [-1] * ndim
+    max_pt = [1] * ndim
     nsamples = np.arange(1, ndim + 1, 1) * 10  # 10, 20 [, 30]
```
kohr-h: Omit the third argument, it's default. And here I think its readable, no need for the extra comment.

test/tomo/backends/astra_setup_test.py
```diff
@@ -204,47 +204,47 @@ def test_astra_projection_geometry():
...
+    assert astra_geom['type'] == 'cone_vec'
 
-    # HELICAL CONEFLAT
+    # Helical conebeam flat
     spiral_pitch_factor = 1
     geom_hcf = odl.tomo.HelicalConeFlatGeometry(angle_intvl, dparams, src_rad,
                                                 det_rad,
                                                 spiral_pitch_factor,
                                                 agrid=angle_grid,
                                                 dgrid=det_grid)
```
kohr-h: You can write
```python
geom_hcf = odl.tomo.HelicalConeFlatGeometry(
    angle_intvl, dparams, src_rad, det_rad, spiral_pitch_factor,
    agrid=angle_grid, dgrid=det_grid)
```
which looks less squeezed (break line immediately after parenthesis and indent).

test/tomo/operator/xray_trafo_test.py
```diff
@@ -30,28 +30,261 @@
...
+
+# DiscreteLp volume / reconstruction space
+xx = 5
```
kohr-h: Don't like these Matlab-style variables. Do we need to store at all here?

test/tomo/operator/xray_trafo_test.py
```diff
@@ -30,28 +30,261 @@
...
+
+    dparams = dparams_2d
+    dgrid = dgrid_2d
+    discr_vol_space = discr_vol_space_2d
+
+    # Geometry
+    geom = odl.tomo.Parallel2dGeometry(angle_intvl, dparams, agrid, dgrid)
+
+    # X-ray transform
+    A = odl.tomo.XrayTransform(discr_vol_space, geom,
```
kohr-h: No capitals in variable names except constants which are all capitalized.

odl/tomo/backends/astra_cpu.py
adler-j: ScalingFactor. Perhaps simply `scaling_factor`?

odl/tomo/backends/stir_bindings.py
adler-j: Well its Swedish as well and odl is basically a German Swedish joint venture. (Finland is basically [East-Sweden](http://sv.urbandictionary.com/define.php?term=East-Sweden))

odl/tomo/operators/xray_trafo.py
adler-j: Currently it can only be created as `XrayTransform.adjoint` I say we keep this until that is fixed?

test/tomo/operator/xray_trafo_test.py
adler-j: This convention is my fault, we should go for `op` if it is general or `projector` in this case, imo.

kohr-h: `projector` is perfect. Then we can also exchange `f` for `phantom` or `reco`, depending on its role, and `g` for `data` or `proj_data`.

odl/tomo/geometry/detector.py
moosmann: I guess it's a copy-and-paste error. It's the 1d detector and it should be singular.

adler-j:
Do we have a progress plan for getting this in?

moosmann:
I first need to check what you have changed, because all of my old test in `xray_trafo_test` fail.

moosmann:
All comments are implemented including to make the back-projector an operator on its own. Another quick review and we can merge.

odl/tomo/backends/astra_cpu.py
```diff
@@ -191,6 +191,17 @@ def astra_cpu_backward_projector_call(proj_data, geometry, reco_space,
     # Run algorithm and delete it
     astra.algorithm.run(algo_id)
 
+    # Fix inconsistent scaling
+    # angle interval weight by approximate cell volume
```
adler-j: Perhaps a todo here to fix properly?

odl/tomo/backends/astra_cuda.py
```diff
@@ -123,6 +122,11 @@ def astra_cuda_forward_projector_call(vol_data, geometry, proj_space,
         else:
             out.assign(tmp)
 
+    # Fix inconsistent scaling
+    if isinstance(geometry, Parallel2dGeometry):
+        # parallel2d scales with pixel stride
```
adler-j: Non quadratic is still fine? Edit comment if it is.

odl/tomo/operators/xray_trafo.py
```diff
@@ -176,57 +183,145 @@ def _call(self, inp):
...
+        discr_range : `odl.DiscreteLp`
+            Reconstruction space, the range of the back-projector
+        geometry : `Geometry`
+            The geometry of the transform, contains information about
+            the operator domain. It needs to have a sampling grid for
+            motion and detector parameters.
+        backend : {'astra', 'astra_cuda', 'astra_cpu'}, optional
+            Implementation back-end for the transform. Supported back-ends:
+            'astra': ASTRA toolbox, uses CPU or CUDA depending on the
+            underlying data space of ``discr_range``
```
adler-j: Never quite understood this, giving a domain/range that is cuda makes the code throw an error anyway?

odl/tomo/operators/xray_trafo.py
```diff
@@ -176,57 +183,145 @@ def _call(self, inp):
...
+            'astra_cpu': ASTRA toolbox using CPU, only 2D
+            'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
+            Default: 'astra'
+        kwargs : {'interp'}
+            'interp' : {'nearest', 'linear', 'cubic'}
+                Interpolation type for the discretization of the
+                operator range. Default: 'nearest'
```
adler-j: A word on when the options are available.

test/tomo/operator/xray_trafo_test.py
```diff
@@ -15,58 +15,171 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Test for X-ray transforms."""
+"""Test reconstruction with ASTRA."""
```
adler-j: Well the astra thing is only temporary, once others come they will be added.

test/tomo/operator/xray_trafo_test.py
```diff
@@ -15,58 +15,171 @@
...
-    assert almost_equal(Af.inner(g), f.inner(Adg), 3)
+    # Adjoint definition <Ax, Ax> = <x, A*A x>
+    result_AxAx = proj.inner(proj)
+    result_xAtAx = vol.inner(projector.adjoint(proj))
+    assert almost_equal(result_AxAx, result_xAtAx, places=places)
 
 
 if __name__ == '__main__':
-    pytest.main(str(__file__.replace('\\', '/') + ' -v'))
+    pytest.main(str(__file__.replace('\\', '/') + ' -v --largescale'))
```
adler-j: largescale is a misstake

odl/tomo/backends/astra_cpu.py
```diff
@@ -191,6 +191,17 @@ def astra_cpu_backward_projector_call(proj_data, geometry, reco_space,
...
+    extent = float(geometry.motion_grid.extent())
+    size = float(geometry.motion_grid.size)
+    scaling_factor = extent / size
+    # parallel2d & fanflat scale with (voxel stride)**2 / (pixel stride)
```
adler-j: Comment on that we only have square voxels

odl/tomo/backends/astra_cuda.py
```diff
@@ -123,6 +122,11 @@ def astra_cuda_forward_projector_call(vol_data, geometry, proj_space,
         else:
             out.assign(tmp)
```
adler-j: This is likely slower than simply not calling `proj_space.element(` above and instead using the `__setitem__` operator here:

```
np.rollaxis(astra.data3d.get(sino_id), 0, 3)
out[:] = tmp
```

odl/tomo/geometry/detector.py
```diff
@@ -410,7 +419,13 @@ def surface(self, param):
...
-        """The partial derivative(s) of the surface parametrization."""
+        """The partial derivative(s) of the surface parametrization.
+
+        Parameters
+        ----------
+        param : element of `params`
+            The parameter value where to evaluate the function
+        """
         if (param in self.params or
                 (isinstance(param, np.ndarray) and
```
adler-j: Should perhaps use contains_all?

odl/tomo/geometry/detector.py
```diff
@@ -410,7 +419,13 @@ def surface(self, param):
                              '{}.'.format(param, self.params))
 
     def surface_deriv(self, param):
-        """The partial derivative(s) of the surface parametrization."""
+        """The partial derivative(s) of the surface parametrization.
+
+        Parameters
+        ----------
+        param : element of `params`
+            The parameter value where to evaluate the function
```
adler-j: Says basically nothing imo.

odl/tomo/backends/astra_cpu.py
moosmann: What is the proper way to do this?

odl/tomo/backends/astra_cuda.py
moosmann: It's only the 2D case where this rescaling is needed. So there are no anisotropic pixels.

odl/tomo/operators/xray_trafo.py
moosmann: I think this was because of the intention to decide on to use cpu or cuda depending on the provided space and because I've anticipated, back then when Willem-Jan was here, that we soon will have an ASTRA version which can use a shared CUDA memory. But, since we don't when this will happen, I'm gonna remove the 'astra' option. Or what do you think?

odl/tomo/backends/astra_cpu.py
adler-j: Depends on what we mean by the projection, if it is the radon transform, this should be independent of the interval length?

odl/tomo/operators/xray_trafo.py
adler-j: Agree, we can add it later, also we can allow users to pick another than the one they are providing, perhaps they prefer using the CPU projector with CUDA data for some reason, then we should allow that.

odl/tomo/operators/xray_trafo.py
```diff
@@ -176,57 +183,145 @@ def _call(self, inp):
...
+        extent = float(geometry.grid.extent().prod())
+        size = float(geometry.grid.size)
+        weight = extent / size
+
+        domain_dspace = discr_range.dspace_type(
+            geometry.grid.size, weight=weight, dtype=discr_range.dspace.dtype)
 
-    def _call(self, inp):
-        """Call the adjoint transform on an input, producing a new vector.
+        domain_interp = kwargs.pop('interp', 'nearest')
```
adler-j: `pop` removes it from the dict, since you save kwargs you lose this option. Use `get` instead.

odl/tomo/operators/xray_trafo.py
```diff
@@ -176,57 +183,145 @@ def _call(self, inp):
...
+            'astra': ASTRA toolbox, uses CPU or CUDA depending on the
+            underlying data space of ``discr_range``
+            'astra_cpu': ASTRA toolbox using CPU, only 2D
+            'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
+            Default: 'astra'
+        kwargs : {'interp'}
+            'interp' : {'nearest', 'linear', 'cubic'}
+                Interpolation type for the discretization of the
+                operator range. Default: 'nearest'
         """
```
adler-j: See also the backprojector.

odl/tomo/operators/xray_trafo.py
```diff
@@ -33,46 +33,47 @@
...
             'astra_cpu': ASTRA toolbox using CPU, only 2D
             'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
             Default: 'astra'
-        kwargs : {'range_interpolation'}
-            'range_interpolation' : {'nearest', 'linear', 'cubic'}
+        kwargs : {'interp'}
+            'interp' : {'nearest', 'linear', 'cubic'}
                 Interpolation type for the discretization of the
                 operator range. Default: 'nearest'
         """
```
adler-j: See also backprojector.

odl/tomo/operators/xray_trafo.py
moosmann: I'm not sure what you mean. This option only affects the operator range. The interpolation scheme used by ASTRA is inferred from the space of the provided data.

kohr-h: Quick one from my side: we don't write "kwargs" explicitly any longer but list the options just like the other ones. That means: Delete the line containing "kwargs", dedent the rest below by 4 and remove the single quotes from the parameter name.

moosmann:
Merged.

Issue #231, opened 2016-01-22
Complex dtype in uniform_discr does not adapt field correctly


kohr-h:
Currently we have
```python
>>> odl.uniform_discr(-2, 2, 64, impl='numpy', dtype='complex64')
...
TypeError: complex floating data type 'complex64' requires space field to be of type `ComplexNumbers`, got <class 'odl.set.sets.RealNumbers'>.
```
This means that the `field` argument takes precedence over `dtype` and defaults to `Realnumbers`. Instead, the function should check `field` and `dtype` for consistency and then select the correct data space based on *both*.

PR #239, opened 2016-01-27
Issue 231 uniform discr dtype

adler-j:
Should close issue #231. Changes:

* Remove field parameter from `uniform_discr`, now deduced from `dtype`. The previous behaviour violated "there should be one obvious way to do it". 
* Field is (uniquely) determined by dtype
* Change repr to match these changes
* Add doctest

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -224,7 +224,7 @@ class DiscreteLp(Discretization):
             arg_fstr = '{}, {}, {}'
             if self.exponent != 2.0:
                 arg_fstr += ', exponent={exponent}'
-            if not self.dtype == default_dtype:
+            if self.dtype != default_dtype:
                 arg_fstr += ', dtype={dtype}'
             if self.interp != 'nearest':
                 arg_fstr += ', interp={interp!r}'
@@ -576,8 +576,7 @@ def uniform_discr_fromspace(fspace, nsamples, exponent=2.0, interp='nearest',
 
 
 def uniform_discr(min_corner, max_corner, nsamples,
-                  exponent=2.0, field=None,
-                  interp='nearest', impl='numpy', **kwargs):
+                  exponent=2.0, interp='nearest', impl='numpy', **kwargs):
     """Discretize an Lp function space by uniform sampling.
 
     Parameters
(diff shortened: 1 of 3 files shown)


odl/discr/lp_discr.py
```diff
@@ -592,8 +592,6 @@ def uniform_discr(min_corner, max_corner, nsamples,
     exponent : positive `float`, optional
         The parameter :math:`p` in :math:`L^p`. If the exponent is not
         equal to the default 2.0, the space has no inner product.
-    field : `Field`, optional
-        The field of the `FunctionSpace`, default `RealNumbers`.
```
kohr-h: Doc removed but argument still there?

kohr-h:
Makes a lot of sense to remove field from `uniform_discr` (you seem to have forgotten to actually do it).

Slightly off-topic but related: I always thought that `field=ComplexNumbers()` was ugly due to instance creation just to define a default. And it's slightly annoying to always have to type those brackets even though the class is completely static. Perhaps one could make all their (RealNumbers, ComplexNumbers, Integers, ...) methods static or class methods and just use the class, not an instance? Just a thought. (Wouldn't work for strings, though.)

odl/discr/lp_discr.py
adler-j: Fixed

adler-j:
> Slightly off-topic but related: I always thought that field=ComplexNumbers() was ugly due to instance creation just to define a default. And it's slightly annoying to always have to type those brackets even though the class is completely static. Perhaps one could make all their (RealNumbers, ComplexNumbers, Integers, ...) methods static or class methods and just use the class, not an instance? Just a thought. (Wouldn't work for strings, though.)

I think that would violate the "set" design, but we could do the c++ manner and define a `ComplexNumbers_t` (obviously with a more pythonic name), i.e.

```python
class ComplexNumbers_t():
    ...
# Create instance
ComplexNumbers = ComplexNumbers_t()
```

This would work for all of the simple ones.

kohr-h:
With something like `_ComplexNumbers` as class name, sounds doable. Pylint will complain about the constant name, but that's a minor issue.

PR #241, opened 2016-01-28
Issue 79  fnbase show

### LINKED ISSUE #79: LinearSpaceVector.show in general
For any cases it makes sense to talk about `LinearSpaceVector.show`, for example:

- [x] `ProductSpaceVector` : make an image with subplots
- [ ] `FnBaseVector` : Line plot. Possibly some statistics on the values?

### LINKED ISSUE #219: Title for second subfigure gone in DiscreteLp.show for complex data
For plots of complex-valued functions, the subplots should have the labels "Real part" and "Imaginary part", which was the case until some internal change. Now the title for the first subfigure is still there, but the second one is gone.

adler-j:
Generalized the `show` syntax, now works for `NtuplesBaseVectors` as well.

* closes #79 
* closes #219


odl/util/graphics.py
```diff
@@ -259,6 +224,7 @@ def show_discrete_function(dfunc, method='', title=None, indices=None,
             plt.colorbar(csub_re, orientation='horizontal',
                          ticks=ticks_re, format='%.4g')
 
+        print(len(fig.axes))
```
kohr-h: Debug?

odl/util/graphics.py
```diff
@@ -78,6 +72,12 @@ def show_discrete_function(dfunc, method='', title=None, indices=None,
         given by this function. The most common usecase is that fig is the
         return value from an earlier call to this function.
 
+    interp : {'nearest', 'linear'}
+        Interpolation method to use.
```
kohr-h: Good you added that, I was thinking the same these days

odl/set/pspace.py
```diff
@@ -619,10 +619,34 @@ def __repr__(self):
...
+
+        Parameters
+        ----------
+        indices : index expression, optional
+            If a single index (``indices=0``) displays that part
+            If a slice, displays those parts (``indices=slice(None)``)
+            If `list` of indices, display those (``indices=[0, 1, 3]``)
+            If a tuple is given, interprets the first element as an index for
+            this product space vector and passes the rest on.
+            (``indices=np.s_[0, :, :]``).
```
kohr-h: Somewhat bumpy language here, and I don't really understand what is being said here. From the code it appears to me that lists and tuples are equivalent.
It would be good to clarify what is differentiated between at all. For example like this 
```
Indices can refer to parts of a `ProductSpaceVector` and slices
in the parts in the following way:

Single index (``indices=0``) => display that part
Single slice (``indices=slice(None)``) or index list (``indices=[0, 1, 3]``) => display those parts
Sequence of slices  (``indices=np.s_[0, :, :]``): take the first axis to select the parts
and pass the rest on.
```

adler-j: The specific handling tuples is standard in python, numpy considers these different:

```python
x = array...
x[[1, 2, 3]] # Get elements 1, 2 and 3
x[(1, 2, 3)] # Get the element at index (1, 2, 3)
```

We simply follow this convention.

With that said, I'll try to improve this.

adler-j: Now given by:

```python
            Indices can refer to parts of a `ProductSpaceVector` and slices
            in the parts in the following way:

            Single index (``indices=0``)
            => display that part

            Single slice (``indices=slice(None)``), or
            index `list` (``indices=[0, 1, 3]``)
            => display those parts

            Any `tuple`, for example:
            Created by numpy.s_ ``indices=np.s_[0, :, :]`` or
            Using a tuple ``indices=([0, 3], slice(None))``
            => take the first elements to select the parts and
            pass the rest on to the underlying show methods.
```

is that OK?

kohr-h: Much easier to understand, good.

odl/util/graphics.py
```diff
@@ -269,7 +234,7 @@ def show_discrete_function(dfunc, method='', title=None, indices=None,
             else:
                 sub_im.set_ylabel('value')
         else:
-            sub_im = fig.axes[1]
+            sub_im = fig.axes[2]
```
kohr-h: Was that the bug causing the scrapped complex title?

adler-j: Nope it was calling `plt.title` below, added an if statement there.

PR #243, opened 2016-01-28
Issue 242 laplacian

### LINKED ISSUE #242: Laplacian operator
The laplacian operator can currently be obtained by `- gradient.adjoint * gradient`, but this is somewhat unintuitive, and also somewhat slower. We should add a dedicated operator for this.

adler-j:
Closes #242 by adding laplacian operator. Not optimized currently, but way better than calling `Gradient.adjoint * Gradient`.

### DIFF
--- a/odl/discr/discr_ops.py
+++ b/odl/discr/discr_ops.py
@@ -31,7 +31,7 @@ from odl.set.pspace import ProductSpace
 from odl.discr.lp_discr import DiscreteLp
 
 
-__all__ = ('PartialDerivative', 'Gradient', 'Divergence')
+__all__ = ('PartialDerivative', 'Gradient', 'Divergence', 'Laplacian')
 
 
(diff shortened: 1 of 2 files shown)


odl/discr/discr_ops.py
```diff
@@ -553,6 +553,102 @@ def adjoint(self):
...
+
+    def __init__(self, space):
+        """Initialize a `Laplacian` operator instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp`
+            The space of elements which the operator is acting on.
+        method : {'central', 'forward', 'backward'}, optional
+            What method that should be used.
```
kohr-h: Which

kohr-h: ~~that~~
~~.~~

kohr-h:
Cool

Issue #247, opened 2016-01-29
astra_projection_geometry is slow


adler-j:
It seems we call `astra_projection_geometry` in each call to `astra` and this method is not the fastest, in a 200^3 -> 200^3 problem, it takes about 30% of the total runtime. In a 100^3 -> 100^3 problem it takes 50% of the runtime.

We could solve this by doing one or more of the following:

* Storing the result and reusing it in each iteration.
* Adding "discrete" properties for some of the continuous functions that are computed once and for all using the angles of the geometry.
    * `rotation_matrix()`  ->  `rotation_matrices`
    * `src_position()`  ->  `src_positions`
    * `det_refpoint()`  ->  `det_refpoints`
* Optimizing `IntervalProd.__contains__` which seems to be really slow atm.
    Suggest optimizing for the case `tol=0` and removing some of the currently rendundant(?) checks.

This should not stop the `conebeam` branch from going through, but needs to be fixed at some point.

adler-j:
This is solved

PR #250, opened 2016-02-01
Issue 247  astra geometry slow

adler-j:
Minor changes according to #247 

* Add a cache to geometries where implementations can be stored. Running a 50^3 -> 50^3 problem went from 0.08s per iteration to 0.027s.
* Improve performance of `IntervalProd.__contains__`, reduce runtime by ~70%.

### DIFF
--- a/odl/set/domain.py
+++ b/odl/set/domain.py
@@ -30,8 +30,8 @@ standard_library.install_aliases()
 import numpy as np
 
 # ODL imports
-from odl.set.sets import Set, RealNumbers
-from odl.util.utility import array1d_repr
+from odl.set.sets import Set
+from odl.util.utility import array1d_repr, is_real_dtype
 from odl.util.vectorization import (
     is_valid_input_array, is_valid_input_meshgrid, meshgrid_input_order,
     vecs_from_meshgrid)
@@ -154,9 +154,37 @@ class IntervalProd(Set):
         """The maximum value in this interval product"""
         return self.end
 
-    def element(self):
-        """An arbitrary element, the midpoint."""
-        return self.midpoint
+    def element(self, inp=None):
+        """Create element in this set.
+
+        Parameters
+        ----------
+        inp : `float` or array-like, optional
+            Point to be cast to an element in self
+
+        Returns
+        -------
+        element
+            Returns ``inp`` if given, else ``self.midpoint``
+
+        Raises
+        ------
+        TypeError
+            If ``inp`` is not a valid element.
+
+        Examples
+        --------
+        >>> interv = IntervalProd(0, 1)
+        >>> interv.element(0.5)
+        0.5
+        """
+        if inp is None:
+            return self.midpoint
+        elif inp in self:
+            return inp
+        else:
+            raise TypeError('inp {!r} not a valid element in {!r}.'
+                            ''.format(inp, self))
 
     # Overrides of the abstract base class methods
     def approx_equals(self, other, tol):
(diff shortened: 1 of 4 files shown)


odl/set/domain.py
```diff
@@ -154,9 +154,37 @@ def max(self):
...
+        >>> interv.element(0.5)
+        0.5
+        """
+        if inp is None:
+            return self.midpoint
+        elif inp in self:
+            return inp
+        else:
+            raise TypeError('inp {!r} not a valid element in {!r}.'
+                            ''.format(inp, self))
```
kohr-h: Reason for this change?

odl/set/domain.py
```diff
@@ -218,23 +246,44 @@ def approx_contains(self, point, tol):
...
+        >>> interv = IntervalProd(0, 1)
+        >>> 0.5 in interv
+        True
+        >>> 2 in interv
+        False
+        >>> 'string' in interv
+        False
+        """
+        try:
+            # Ducktyped check
```
kohr-h: Duck-typed I think it is

odl/set/domain.py
adler-j: Consistency, the definition of `element`says it should take an `inp`:

https://github.com/odlgroup/odl/blob/eec730cd0128a77a4315bf2303da755878bc346d/odl/set/sets.py#L126-L131

odl/tomo/backends/astra_setup.py
```diff
@@ -396,6 +401,10 @@ def astra_projection_geometry(geometry):
         raise NotImplementedError('unkown ASTRA geometry type {}.'.format(
             geometry))
 
+    if 'astra' not in geometry.implementations:
+        # Shortcut, reuse already computed value.
```
kohr-h: Other way around

odl/set/domain.py
kohr-h: OK

odl/set/domain.py
```diff
@@ -218,23 +246,44 @@ def approx_contains(self, point, tol):
...
+        return self.dist(point, ord=np.inf) <= tol
 
     def __contains__(self, other):
-        """Return ``other in self``."""
-        return self.approx_contains(other, tol=0)
+        """Return ``other in self``.
+
+        Parameters
+        ----------
+        other
```
kohr-h: With or without colon at the end? Have you tested the doc?

adler-j: Nope, but this should work I think, will do a big test of doc before we merge conebeam

odl/tomo/geometry/geometry.py
```diff
@@ -154,6 +155,18 @@ def ndim(self):
         return self._ndim
 
     @property
+    def implementations(self):
```
kohr-h: Why "implementations"? What about "cache"?

adler-j: I was thinking something complicated, so users dont go for it. How about `implementation_cache`?

kohr-h: What about not making it public then?

adler-j: Well its used by our other classes, private means only for this class.

kohr-h: Alright, we should actually not use private attributes too much externally. Then s.th. like `impl_cache` would sound even more mysterious.

kohr-h:
We have a failing test, actually for a while, in the conebeam branch:

```
Error Message

TypeError: unorderable types: DiscreteLpVector() > int()

Stacktrace

@skip_if_no_astra
    def test_astra_cpu_projector_fanflat():
        """ASTRA CPU forward and back projection for fanflat geometry."""
    
        # `DiscreteLp` space for volume data
        vol_shape = (4, 5)
        discr_vol_space = odl.uniform_discr([-4, -5], [4, 5], vol_shape,
                                            dtype='float32')
    
        # Create an element in the `DiscreteLp` space
        discr_data = odl.util.phantom.cuboid(discr_vol_space, 0.5, 1)
    
        # Angles
        angle_intvl = odl.Interval(0, 2 * np.pi)
        angle_grid = odl.uniform_sampling(angle_intvl, 8)
    
        # Detector
        dparams = odl.Interval(-6, 6)
        det_grid = odl.uniform_sampling(dparams, 6)
    
        # Distances for fanflat geometries
        src_rad = 100
        det_rad = 10
    
        # 2D geometry instances for parallel and fan beam with flat line detector
        geom = odl.tomo.FanFlatGeometry(angle_intvl, dparams, src_rad, det_rad,
                                        angle_grid, det_grid)
    
        # Projection space
        proj_space = odl.FunctionSpace(geom.params)
    
        # `DiscreteLp` projection space
        proj_shape = geom.grid.shape
        discr_proj_space = odl.uniform_discr_fromspace(proj_space, proj_shape,
                                                       dtype='float32')
    
        # forward
        discr_data = odl.util.phantom.cuboid(discr_vol_space, 0.5, 1)
        proj_data = odl.tomo.astra_cpu_forward_projector_call(discr_data, geom,
                                                              discr_proj_space)
        assert proj_data.shape == proj_shape
>       assert proj_data > 0
E       TypeError: unorderable types: DiscreteLpVector() > int()

test/tomo/backends/astra_cpu_test.py:124: TypeError
```

I wrote a comment in another place some time ago, but nobody fixed it.

odl/tomo/geometry/geometry.py
adler-j: Went for `implementation_cache` since its long and cumbersome :)

adler-j:
Is that failing test really due to this branch?

kohr-h:
No, it's from upstream. But as I wrote, there was no reaction.

adler-j:
In general I think those tests are waiting for the correct scalings before we can do them. The `astra_scaling` branch contains fixes to this I think.

kohr-h:
Yes, but this particular test fails because it's nonsense, regardless of scaling :-)

odl/tomo/geometry/geometry.py
kohr-h: Go for it!

moosmann:
The test should be
```
assert proj_data.norm() > 0
```
and simply check that the projection data is positive. I already fixed this test in the `astra_scaling` branch but not in the cone-beam branch.

kohr-h:
You're lucky that this dates back to the time when I was remodeling our Jenkins CI. Otherwise you would have gotten angry e-mail :-)

Issue #262, opened 2016-02-05
Add pitch_offset


adler-j:
Currently there is no way to give a "offset" along the axis for helical geometries, at angle 0, the source/detector is in a fixed hight (at z=0, by default). I propose we add a `pitch_offset` variable that allows users to specify how far along the axis the source should be when ``angle=0``.

This is redundant, since users could theoretically move the volume such that it matches, but this requires a specific choice of coordinate system which may not be natural.

kohr-h:
You may even have to rotate the whole system since your new starting point may not lie on the old helix. Makes sense

PR #263, opened 2016-02-05
Issue 262 pitch offset

adler-j:
Solves #262


odl/tomo/geometry/conebeam.py
```diff
@@ -50,7 +50,8 @@ class HelicalConeFlatGeometry(DivergentBeamGeometry, AxisOrientedGeometry):
     detector positions.
     """
 
-    def __init__(self, angle_intvl, dparams, src_radius, det_radius, pitch,
+    def __init__(self, angle_intvl, dparams, src_radius, det_radius,
+                 pitch, pitch_offset=0,
                  agrid=None, dgrid=None, axis=[0, 0, 1], src_to_det=None,
                  detector_axes=None):
```
kohr-h: Make `pitch_offset` and `detector_axes` part of `kwargs`, too many arguments here (even w/o the intervals)

odl/tomo/geometry/conebeam.py
```diff
@@ -155,7 +169,8 @@ def det_refpoint(self, angle):
         circle_component = self.rotation_matrix(angle).dot(origin_to_det)
 
         # Increment by pitch
-        pitch_component = self.axis * self.pitch * angle / (np.pi * 2)
+        pitch_component = self.axis * (self.pitch_offset +
+                                       self.pitch * angle / (np.pi * 2))
```
kohr-h: So we have to do that in all places, right? Makes it more complicated, but ok

adler-j: Its basically in these two places, right?

odl/tomo/geometry/conebeam.py
adler-j: Good suggestion

odl/tomo/geometry/conebeam.py
kohr-h: I don't know. Probably. But also possibly more to come. Anyway, it's a sensible addition.

adler-j: Honestly cant see where else, the geometries are uniquely defined by these two points, any utility  would likely just call them and get it for free?

kohr-h: Yes, I guess so.

adler-j:
Will fix all these commits, travis has changed something and now it fails.

kohr-h:
Can you clean up with `git rebase -i` when you're done here? Would be good not to have this commit mess in the final git log.

adler-j:
Rebase done, merging.

PR #265, opened 2016-02-08
ENH: made apply_on_boundary use out parameter

adler-j:
Did the change I proposed to `apply_on_boundary` now conforms to 

Profiling:

```python
arr = np.zeros([10] * 5)

# old
for i in range(10**6):
    apply_on_boundary(arr, lambda x: x / 2)
    
# new
for i in range(10**6):
    apply_on_boundary(arr, lambda x: x / 2, out=arr)
```

Time:

old: 21.015s
new: 21.115s

Also simplifies the implementation of `_norm` etc notably.

### DIFF
--- a/odl/util/numerics.py
+++ b/odl/util/numerics.py
@@ -24,20 +24,20 @@ from __future__ import print_function, division, absolute_import
 from future import standard_library
 standard_library.install_aliases()
 
-# External module imports
-
 
 __all__ = ('apply_on_boundary',)
 
 
 def apply_on_boundary(array, func, only_once=True, which_boundaries=None,
-                      axis_order=None):
+                      axis_order=None, out=None):
     """Apply a function of the boundary of an n-dimensional array.
+
+    All other values are preserved as is.
 
     Parameters
     ----------
     array : `numpy.ndarray`
-        Modify the boundary of this array in place
+        Modify the boundary of this array
     func : `callable` or `sequence`
         If a single function is given, assign
         ``array[slice] = func(array[slice])`` on the boundary slices,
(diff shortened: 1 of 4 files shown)


odl/util/numerics.py
```diff
@@ -109,6 +125,11 @@ def apply_on_boundary(array, func, only_once=True, which_boundaries=None,
         raise ValueError('axis_order has length {}, expected {}.'
                          ''.format(len(axis_order), array.ndim))
 
+    if out is None:
+        out = array.copy()
+    else:
+        out[:] = array  # Self assignment is free, in case out is array
```
kohr-h: Interesting, didn't know that.

odl/util/numerics.py
```diff
@@ -66,31 +68,45 @@ def apply_on_boundary(array, func, only_once=True, which_boundaries=None,
         to process the axes. If combined with ``only_once`` and a
         function list, this determines which function is evaluated in
         the points that are potentially processed multiple times.
+    out : `numpy.ndarray`
+        Location in which to store the result, can be the same as ``array``.
+        Default: copy of ``array``
```
kohr-h: Aha so this is how you meant it. That makes sense, sure.

odl/util/numerics.py
```diff
@@ -66,31 +68,45 @@ def apply_on_boundary(array, func, only_once=True, which_boundaries=None,
         to process the axes. If combined with ``only_once`` and a
         function list, this determines which function is evaluated in
         the points that are potentially processed multiple times.
+    out : `numpy.ndarray`
```
kohr-h: , optional

kohr-h:
Najs, go ahead

PR #275, opened 2016-02-10
Add biorthogonal wavelets

niinimaki:
Please review the implementations.
Especially the importing of `jos_wavelets`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ script:
   # Invoke also the alternative way of running the unit tests
   - python -c "import odl; odl.test()"
   # Build the Sphinx doc (only for Python 3.5 and the master branch)
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "3.5" && "$TRAVIS_BRANCH" == "master" ]]; then
+  - if [[ "$TRAVIS_PYTHON_VERSION" == "3.5" && "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]]; then
       cd $TRAVIS_BUILD_DIR/doc/source && python generate_doc.py && cd -;
       travis-sphinx -n -s $TRAVIS_BUILD_DIR/doc/source build;
     fi
@@ -64,6 +64,6 @@ after_success:
   - coveralls
   # Deploy the Sphinx doc to gh-pages (only for Python 3.5 and the master branch)
   # See https://github.com/Syntaf/travis-sphinx
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "3.5" && "$TRAVIS_BRANCH" == "master" ]]; then
+  - if [[ "$TRAVIS_PYTHON_VERSION" == "3.5" && "$TRAVIS_BRANCH" == "master" && "$TRAVIS_PULL_REQUEST" == "false" ]]; then
       travis-sphinx deploy;
     fi
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,31 +1,33 @@
 package:
     name: odl
-    version: "0.5.1"
+    version: "0.5.2"
 
 source:
     git_url: https://github.com/odlgroup/odl
     # git_rev: master  # for testing, put any branch here
-    git_rev: v0.5.1  # release
+    git_rev: v0.5.2  # release
+    # git_rev: a542c12d23da7fa5b92b360a51ea14e4804c58f6  # intermediate bugfix revision
 
 build:
-    noarch_python: True
+    number: 0
+    # noarch_python: True
 
 requirements:
     build:
-        - python
-        - nomkl
+        - python 2.7|>=3.4
         - setuptools
+        - nomkl # [not win]
         - future >=0.14
         - numpy >=1.9
         - scipy >=0.14
     run:
-        - python
-        - nomkl
+        - python 2.7|>=3.4
         - future >=0.14
+        - nomkl # [not win]
         - numpy >=1.9
         - scipy >=0.14
         - matplotlib
-        - pytest >=3.0
+        - pytest >=3.0.3
 
 test:
     imports:
--- a/doc/source/getting_started/installing.rst
+++ b/doc/source/getting_started/installing.rst
@@ -54,14 +54,19 @@ Alternative 1: Installing a release using ``conda`` (recommended for users)
 
 TL;DR
 =====
+Instructions for the impatient:
+
 - Download and install `Miniconda`_
 - Create conda environment:
 
   .. code-block:: bash
 
-    $ conda create -c odlgroup -n odl-py35 python=3.5 odl odlcuda matplotlib pytest scikit-image spyder
+    $ conda create -c odlgroup -n odl-py35 python=3.5 odl matplotlib pytest scikit-image spyder
 
 - Activate the conda enviroment and start working!
+
+
+.. _installing_anaconda:
 
 Installing Anaconda
 ===================
@@ -92,15 +97,24 @@ This is a very convenient way to have several "ecosystems" of Python packages in
 
     $ conda create --name odl-py35 python=3.5
 
-Run ``source activate odl-py35`` to enter/exit the newly created conda environment with name ``odl-py35``.
+Enter the newly created conda environment by running ``source activate odl-py35`` (Linux/MacOS) or ``activate odl-py35`` (Windows).
+If you want to exit later on, run ``source deactivate`` (Linux/MacOS) or ``deactivate`` (Windows), respectively.
 See the `Managing conda environments`_ documentation for further information.
 
-If you use `Spyder`_ as integrated development environment (IDE, see `Development enviroment`_), you should also install it in the new conda environment and run it from there.
-Otherwise, Spyder is not able to use the packages in the conda environment:
+.. note::
+    If you want to use `Spyder`_ as integrated development environment (IDE, see `Development enviroment`_) on Linux or MacOS, you should also install it in the new conda environment and run it from there.
+    Otherwise, Spyder may not able to use the packages in the conda environment:
 
-.. code-block:: bash
+    .. code-block:: bash
 
-    $ conda install spyder
+        $ conda install spyder
+
+    On Windows, you can install Spyder in the root conda environment (run ``deactivate`` to get there), but you need to change its default Python interpreter.
+    To do this, open Spyder and use the navigation bar to open "Tools -> Preferences".
+    Click on "Python interpreter" and change the first setting "Select the Python interpreter for all Spyder consoles" from the default setting to "Use the following Python interpreter:".
+    In the text field, fill in the path to the Python executable in your newly created conda environment.
+    For example, if you installed Miniconda (or Anaconda) in ``C:\Programs\Miniconda3``, then the environment's Python interpreter is ``C:\Programs\Miniconda3\envs\odl-py35\bin\python.exe``.
+    You can use the file system browser (symbol to the right of the text field) to find the interpreter on your system.
 
 
 Installing ODL and its dependencies
@@ -184,6 +198,8 @@ Alternative 2: Installing a release using ``pip``
 
 TL;DR
 =====
+Instructions for the impatient:
+
 - Install `pip`_
 - Install ODL and dependencies:
 
@@ -194,7 +210,7 @@ TL;DR
 Installing a Python interpreter
 ===============================
 Open a terminal and type ``python`` + Enter.
-If a Python prompt appears, you already have an interpreter installed and can skip this step (exit by pressing Ctrl+D).
+If a Python prompt appears, you already have an interpreter installed and can skip this step (exit by running ``exit()``).
 Otherwise, you need to install it.
 
 On Linux:
@@ -219,7 +235,7 @@ Pick the latest release for your favorite version (2 or 3).
 Installing ODL and its dependencies
 ===================================
 You may need to `install pip`_ to be able to install ODL and its dependencies from the `Python Package Index`_ (PyPI).
-If running ``pip`` (alternatively: ``pip2`` or ``pip3``) an shows a help message, it is installed -- otherwise you need to install it first.
+If running ``pip`` (alternatively: ``pip2`` or ``pip3``) shows a help message, it is installed -- otherwise you need to install it first.
 
 For basic installation without extra dependencies, run
 
@@ -278,13 +294,21 @@ No GitHub account is required for this step.
 
 In a conda environment
 ======================
-This part assumes that you have run ``source activate <your_conda_env>`` before.
+This part assumes that you have activated a conda environment before (see :ref:`installing_anaconda`).
 
 You can choose to install dependencies first (optional ones in square brackets):
+
+**On Linux/MacOS:**
 
 .. code-block:: bash
 
     $ conda install nomkl numpy scipy future [matplotlib]
+
+**On Windows:**
+
+.. code-block:: bash
+
+    $ conda install numpy scipy future [matplotlib]
 
 After that, enter the top-level directory of the cloned repository and run
 
@@ -306,7 +330,7 @@ Enter the top-level directory of the cloned repository and run
 
 .. note::
     We recommend the ``--editable`` option (can be shortened to ``-e``) since it installs a link instead of copying the files to your Python packages location.
-    This way, local changes to the code (e.g. after a ``git pull``) take immediate effect without reinstallation.
+    This way, local changes to the code (e.g. after a ``git pull``) take immediate effect after reloading the package, without requiring reinstallation.
 
 
 Further developer information
@@ -326,10 +350,17 @@ They can be run either from within ``odl`` or by invoking ``pytest`` directly.
 
 Installing testing dependencies
 ===============================
+If you installed an ODL release using ``conda`` or ``pip``, respectively, you should install ``pytest`` using the same method.
+For source installations, you can choose your favorite method below.
 
 Using ``conda``:
 ----------------
-``pytest`` is already a dependency, nothing to do.
+If you installed ODL using conda, ``pytest`` is already installed as dependency, so there should not be anything left to do.
+Otherwise, you can install it by running
+
+.. code-block:: bash
+
+    $ conda install pytest
 
 Using ``pip``:
 --------------
@@ -347,9 +378,9 @@ Now you can check that everything was installed properly by running
    $ python -c "import odl; odl.test()"
 
 .. note::
-    Don't run this command in the top-level directory of an ODL clone, since in that case, the tests in the repository are run, not the ones in the installed package.
+    Don't run this command in the top-level directory of an ODL clone, since in that case, the tests in the repository may be run, not the ones in the installed package.
 
-If you have installed ODL from source, you can also use ``pytest`` directly:
+If you have installed ODL from source, you can also use ``pytest`` directly in the root of your ODL clone:
 
 .. code-block:: bash
 
@@ -367,13 +398,14 @@ Some of them can be installed using ``conda``, others require manual compilation
 
 CUDA backend for linear arrays
 ==============================
-The `odlcuda`_ backend for fast array calculations on CUDA requires the `CUDA toolkit`_ (on Linux: use your distro package manager) and a CUDA capable graphics card with compute capability of at least 5.0.
+The `odlcuda`_ backend for fast array calculations on CUDA requires the `CUDA toolkit`_ (on Linux: use your distro package manager) and a CUDA capable graphics card with compute capability of at least 3.0.
 Search `this table <https://en.wikipedia.org/wiki/CUDA#GPUs_supported>`_ for your model.
 
 Installation using ``conda``
 ----------------------------
 .. note::
-    In ``conda``, the ``odlcuda`` package is currently only available for Linux 64-bit and Python 3.5.
+    In conda, the ``odlcuda`` package is currently available only for Linux 64-bit and Python 3.5.
+    Furthermore, you may experience failures due to "invalid device function" -- this is a known issue, and we're trying to fix it.
 
 If you have installed an ODL release, simply run (in a directory of your choice)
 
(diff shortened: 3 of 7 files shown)


adler-j:
I would personally put all code associated with `jos_wavelets` into his library and add `odl` as a dependency to it, as it is now, I doubt it will have many external users until it stabilizes. Adding very young codes straight into ODL will cause us some headaches in the long run since we then risk becoming the point of support of these.

Much of the code could be moved "as is" to that library, so I'll add some comments on this.

odl/trafos/wavelet.py
```diff
@@ -653,22 +704,29 @@ def _call(self, x):
...
-                                                 self.nscales)
-            coeff_arr = pywt_coeff_to_array(coeff_dict, self.size_list)
-
-            return self.range.element(coeff_arr)
+        if not isinstance(self.wbasis, pywt.Wavelet):
+            Op = BiorthWaveletTransform(dom=self.domain, nscales=self.nscales,
+                                        wbasis=self.wbasis)
+            coeff = Op(x)
+            return self.range.element(coeff)
+        elif isinstance(self.wbasis, pywt.Wavelet):
```
adler-j: this could have been written `else:`

odl/trafos/wavelet.py
```diff
@@ -653,22 +704,29 @@ def _call(self, x):
...
+                coeff_list = pywt.wavedec(x, self.wbasis, self.mode,
+                                          self.nscales)
+                coeff_arr = pywt_coeff_to_array(coeff_list, self.size_list)
+                return self.range.element(coeff_arr)
+
+            if x.space.ndim == 2:
+                coeff_list = pywt.wavedec2(x, self.wbasis, self.mode,
+                                           self.nscales)
+                coeff_arr = pywt_coeff_to_array(coeff_list, self.size_list)
+                return self.range.element(coeff_arr)
```
adler-j: This is automatically done by odl, simply do:

```python
coeff_list = pywt.wavedec2(x, self.wbasis, self.mode,
                           self.nscales)
return pywt_coeff_to_array(coeff_list, self.size_list)
```

odl/trafos/wavelet.py
```diff
@@ -653,22 +704,29 @@ def _call(self, x):
...
-            coeff_dict = wavelet_decomposition3d(x, self.wbasis, self.mode,
-                                                 self.nscales)
-            coeff_arr = pywt_coeff_to_array(coeff_dict, self.size_list)
-
-            return self.range.element(coeff_arr)
+        if not isinstance(self.wbasis, pywt.Wavelet):
+            Op = BiorthWaveletTransform(dom=self.domain, nscales=self.nscales,
+                                        wbasis=self.wbasis)
+            coeff = Op(x)
+            return self.range.element(coeff)
```
adler-j: What is the range of `Op`, does this need to be called?

odl/trafos/wavelet.py
```diff
@@ -599,45 +630,65 @@ def __init__(self, dom, nscales, wbasis, mode):
...
+            return self.wbasis.orthogonal
+        else:
+            return False
 
     @property
     def is_biorthogonal(self):
         """Whether or not the wavelet basis is bi-orthogonal."""
-        return self.wbasis.biorthogonal
+        if isinstance(self.wbasis, pywt.Wavelet):
+            return self.wbasis.biorthogonal
```
adler-j: else?

odl/trafos/wavelet.py
```diff
@@ -25,17 +25,25 @@
 
 # External
 import numpy as np
+#import jos_wavelets as bwt
 try:
     import pywt
     PYWAVELETS_AVAILABLE = True
+    import jos_wavelets as bwt
```
adler-j: You need two try catches to do things like this:

```python
try:
    import lib1
    LIB1_AVAILABLE = True    
except ImportError:
    LIB1_AVAILABLE = False
    
try:
    import lib2
    LIB2_AVAILABLE = True    
except ImportError:
    LIB2_AVAILABLE = False
```

odl/trafos/wavelet.py
```diff
@@ -532,26 +541,37 @@ def __init__(self, dom, nscales, wbasis, mode):
...
+            calles J.-O. Stromberg's wavelet library.
+
             Describes properties of a selected wavelet basis.
             See PyWavelet `documentation
             <http://www.pybytes.com/pywavelets/ref/wavelets.html>`_
 
             Possible wavelet families are:
 
-            Haar (``haar``)
+            Haar (``'haar'``)
```
adler-j: These changes to doc are all good

odl/trafos/wavelet.py
```diff
@@ -676,18 +734,28 @@ def adjoint(self):
...
     def inverse(self):
         """The inverse wavelet transform."""
-        return WaveletTransformInverse(
-            ran=self.domain, nscales=self.nscales, wbasis=self.wbasis,
-            mode=self.mode)
+        if not isinstance(self.wbasis, pywt.Wavelet):
+            return InverseBiorthWaveletTransform(ran=self.domain,
+                                                 nscales=self.nscales,
+                                                 wbasis=self.wbasis)
+        elif isinstance(self.wbasis, pywt.Wavelet):
```
adler-j: Just else

odl/trafos/wavelet.py
```diff
@@ -750,11 +828,26 @@ def __init__(self, ran, nscales, wbasis, mode):
...
-        self.wbasis = wbasis
         self.mode = str(mode).lower()
 
+        if isinstance(wbasis, pywt.Wavelet):
+            self.wbasis = wbasis
+        else:
+            try:
+                wbasis = wbasis + ''
+                if wbasis.startswith('JOS'):
+                    self.wbasis = wbasis
```
adler-j: Here `self.wbasis` can be either a `str` or a `pywt.Wavelet` such type mixings are in general not good since any consumer of this property would need to check which one it is at all times.

odl/trafos/wavelet.py
```diff
@@ -763,44 +856,66 @@ def __init__(self, ran, nscales, wbasis, mode):
...
+            return self.wbasis.orthogonal
+        else:
+            return False
 
     @property
     def is_biorthogonal(self):
         """Whether or not the wavelet basis is bi-orthogonal."""
-        return self.wbasis.biorthogonal
+        if isinstance(self.wbasis, pywt.Wavelet):
+            return self.wbasis.biorthogonal
```
adler-j: else?

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        if not isinstance(self.wbasis, pywt.Wavelet):
+            return BiorthWaveletTransform(dom=self.range, ran=self.domain,
+                                          nscales=self.nscales,
+                                          wbasis=self.wbasis)
+        elif isinstance(self.wbasis, pywt.Wavelet):
+            return WaveletTransform(dom=self.range, nscales=self.nscales,
+                                    wbasis=self.wbasis, mode=self.mode)
+
+
+class BiorthWaveletTransform(Operator):
```
adler-j: Why would we have a separate `BiorthWaveletTransform` if `WaveletTransform` can already be biorthonormal? It even has a `is_biorthogonal` property. This needs to be documented.

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+            raise TypeError('domain {!r} is not a `DiscreteLp` instance.'
+                            ''.format(dom))
+
+        if dom.exponent != 2.0:
+            raise ValueError('domain Lp exponent is {} instead of 2.0.'
+                             ''.format(dom.exponent))
+
+        ndim = dom.ndim
+        if ndim == 1:
+            max_level = np.log2(dom.grid.shape[0])
```
adler-j: log2 returns a float, how is rounding done here?

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+            raise NotImplementedError('ndim {} not 1, 2 or 3'
+                                      ''.format(len(dom.ndim)))
+
+        max_level = np.ceil(max_level)
+        if nscales > max_level:
+            raise ValueError('Maximum useful number of scaling levels is {}, '
+                             'got {}.'.format(max_level, self.nscales))
+            #raiseETC ('Setting {} to {}' .format(self.nscales, max_level))
+
+        filterlength = int(wbasis[9])
```
adler-j: `filterlength = int(wbasis[-1])` would perhaps be more clear

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        if filterlength not in (1, 3, 5, 7, 9):
+            raise NotImplementedError('Filterlength {} not 1, 3, 5, 7 or 9'
+                                      ''.format(filterlength))
+
+        ran_size = dom.size
```
adler-j: redundant variable

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        x : `DiscreteLpVector`
+
+        Returns
+        -------
+        arr : `numpy.ndarray`
+            Flattened and concatenated coefficient array
+            The length of the array depends on the size of input image to
+            be transformed and on the chosen wavelet basis.
+        """
+        filterlength = int(self.wbasis[9])
```
adler-j: If this is used repeatedly perhaps we can save it

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        Parameters
+        ----------
+        ran : `DiscreteLpVector` ("image domain")
+
+        nscales : `int`
+            Number of scaling levels in the wavelet transform.
+
+        wbasis : `str`
+            (``'JOSbiorthN'``) J.-O. Stomberg's biorthogonal wavelets
+            where ``N`` refers to the filter length.
```
adler-j: Why does `nscales` get its own parameter but `N` does not?

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        -------
+        arr : `DiscreteLpVector`
+        """
+        filterlength = int(self.wbasis[9])
+        if len(self.range.grid.shape) == 1:
+            nx = self.range.grid.shape[0]
+            #A copy of coeff is unavoidable since bwt writes over
+            #the given input
+            coeff_in = coeff.copy()
+            coeff_in = coeff_in.asarray()
```
adler-j: simply do this on one line:

``` 
coeff_in = coeff.copy().asarray()
```

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+            x = np.asarray(self.range.element())
```
adler-j: Decide on either `vector.asarray()` or `np.asarray(vector)`, but try to be consistent.

adler-j: `x` does not realy convey what it is intended to do

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+
+        if not isinstance(ran, DiscreteLp):
+            raise TypeError('domain {!r} is not a `DiscreteLp` instance.'
+                            ''.format(ran))
+
+        if ran.exponent != 2.0:
+            raise ValueError('domain Lp exponent is {} instead of 2.0.'
+                             ''.format(ran.exponent))
+
+        ndim = ran.ndim
```
adler-j: This piece of code (following 10 lines) is written many times. Perhaps make a utility?

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+                                                wbasis=self.wbasis)
+
+    @property
+    def inverse(self):
+        """The inverse of the adjoint of biorthogonal wavelet transform."""
+        return BiorthWaveletTransform(dom=self.range, nscales=self.nscales,
+                                      wbasis=self.wbasis)
+
+
+class InverseAdjBiorthWaveletTransform(Operator):
```
adler-j: These four classes are all terribly similar, is there no chance to change them to one class?

niinimaki:
> I would personally put all code associated with jos_wavelets into his library and add odl as a dependency to it, as it is now, I doubt it will have many external users until it stabilizes. Adding very young codes straight into ODL will cause us some headaches in the long run since we then risk becoming the point of support of these.Much of the code could be moved "as is" to that library, so I'll add some comments on this.

It is OK for me to put these codes into `jos_wavelets` library, if it is preferred. 
So, how do we proceed then?
 But anyway I'll first do the modifications according to the review.

> Also, the name of the commit is somewhat non-descriptive.

Sorry about that, will be more descriptive in the future.

odl/trafos/wavelet.py
niinimaki: The range of `BiorthWaveletTransform` doesn't need to be called, the range is determined inside the class. Now the user only needs to know the "image domain" and thus the calling of  `BiorthWaveletTransform` is uniform with  calling of `WaveletTransform`.

odl/trafos/wavelet.py
niinimaki: I suppose you refer mainly to ` wbasis = wbasis + ''`
This was a fix suggested by @kohr-h, previously I had
 
 `if isinstance(wbasis, pywt.Wavelet):`
            `self.wbasis = wbasis`
        `elif isinstance(wbasis, str):`
            `if wbasis[0:6] == 'biorth':`
                `self.wbasis = wbasis`
            `else:`
                `self.wbasis = pywt.Wavelet(wbasis)`

but this did not work (and would have certainly needed some improve). 

As you said `self.wbasis` can be either a `str` or a `pywt.Wavelet`. When it is a `str` one needs to distinguish if it refers to the wavelet families within pyWavelets (quite a many combinations with all possible filters in all wavelet families)  of to the Jan-Olov's wavelets. If it refers to Jan-Olov's then it will further be passed as a `str` else transformed to `pywt.Wavelet`. As I said my previous attempt passed all `str` to `BiorthWaveletTranform` even when they referred to some of pyWavelets wavelet families...

Would appreciate a comment how to better formulate this.

odl/trafos/wavelet.py
niinimaki: The separate `BiorthWaveletTransform` is due to the Jan-Olov's wavelet library.
The biorthogonal wavelets that are in pyWavelets have even filter lengths, Jan-Olov's wavelets have odd filter length. Also for Jan-Olov's wavelets there is own classes for the adjoint transform and the inverse adjoint transform, these we don't have for biorthogonal pyWavelets (with orthogonal wavelets they are not necessary). _Maybe_ Jan-Olov will write adjoint and inverse adjoint codes for even length filters at some point, but that remains to be seen...

I'll add some comment to the code related to this

odl/trafos/wavelet.py
niinimaki: This is again due to making the calling of `WaveletTransform` and `BiorthWaveletTransform` uniform. Also `nscales` refers to the number of scaling levels that can be chosen independently from the choice of wavelets (i.e. almost independently, up to the maximum useful level that is determined from the size of the image and from the chosen wavelet basis). `N` refers to the filter length that are implemented in Jan-Olov's wavelet library and the values are fixed to `1, 3, 5, 7` or `9`. `nscales` is used for both `WaveletTransform` and `BiorthWaveletTransform` but `N` would only be needed for  `BiorthWaveletTransform`

odl/trafos/wavelet.py
kohr-h: The simplest possibility, and I guess that's what @adler-j means, is to always expect a string and only differentiate between `wbasis.startswith('jos')` (lowercase preferred). For the pywavelets strings, you would use the string to initialize a `Wavelet` object, otherwise you call the corresponding function in Jan-Olov's library.

In the long run, I suggest to use a flag like `'backend'` with choices `pywavelets` or `jos` and separate the wavelet families according to that. Then add a property `backend` to the classes, that makes checking for implementation easy. You can also split the `_call` code accordingly, i.e. make a `_call_pyfftw` and a `_call_jos` and just call those from the `_call` depending on backend.

odl/trafos/wavelet.py
kohr-h: How about adding this to the standard `WaveletTransform` and just raise an error if the filter length is not supported? The init patterns are quite similar, so no worries there. You only have to make `mode` optional (which would be good anyway), expect `wbasis` to be a string, and we're all good I guess.

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        Parameters
+        ----------
+        dom : `DiscreteLp`
+            Domain of the biorthogonal wavelet transform (the "image domain").
+            The exponent :math:`p` of the discrete :math:`L^p`
+            space must be equal to 2.0.
+        nscales : `int`
+            Number of scaling levels in the wavelet transform.
+        wbasis : `str`
+            (``'JOSbiorthN'``) J.O. Stomberg's biorthogonal wavelets
```
kohr-h: "r" in Strömberg missing.

odl/trafos/wavelet.py
adler-j: I.m.o calls to two different libraries should not be in the same file, this makes the imports etc confusing. I think the way it has been done with `XrayTransform` is ok, make a general class and make it call an `...impl` in some other file which contains all the dirty inner workings.

odl/trafos/wavelet.py
adler-j: I think the main line of thought here still should be to move all of this JOS things outside of ODL and into a fully separate operator, then it could be called `BiorthWaveletTransform` without problems.

If we wanted to, we could make an abstract "WaveletTransform" class that both of these inherit from (we could even overload its `__new__` to return a proper one depending on inputs).

odl/trafos/wavelet.py
kohr-h: Right, we could actually just clone that pattern with a `backends` directory and push the `size_list` and stuff to the `pywavelets` part and `jos` to  another file there. And then we would just need three operators here, `WaveletTransform`, `WaveletTransformAdjoint` and `WaveletTransformInverse`. Regarding names, we have somehow settled for this pattern since it will put the names together in auto-completion lists. Sorry @niinimaki, I know I said differently a while ago. I do the restructuring.

odl/trafos/wavelet.py
niinimaki: After the `if` statements the rounding is done by

`max_level = np.ceil(max_level)`

kohr-h: You can simply do `max_level = int(np.ceil(np.log2(np.max(dom.shape))))`. The dimension check can happen independently.

odl/trafos/wavelet.py
```diff
@@ -834,14 +957,511 @@ def adjoint(self):
...
+        ndim = dom.ndim
+        if ndim == 1:
+            max_level = np.log2(dom.grid.shape[0])
+        elif ndim == 2:
+            max_level = np.log2(max(dom.grid.shape[0], dom.grid.shape[1]))
+        elif ndim == 3:
+            max_level = np.log2(max(dom.grid.shape[0], dom.grid.shape[1],
+                                dom.grid.shape[2]))
+        else:
+            raise NotImplementedError('ndim {} not 1, 2 or 3'
```
kohr-h: Since the user never specified something like `ndim`, this error message will look cryptic. Better say something like 
`'dimension of input is {}, supported are 1, 2 and 3.'.format(dom.ndim)`. `len(dom.ndim)` will raise an error BTW.

odl/trafos/wavelet.py
kohr-h: > to return a proper one depending on inputs

What does that mean?

odl/trafos/wavelet.py
adler-j: :+1:

odl/trafos/wavelet.py
adler-j: Something like:

```python
class WaveletTransform(object):
    def __new__(cls, arg=None):
        if arg == 'jos':
            cls = JOSWaveletTransform
        elif arg == 'pywt':
            cls = pywaveletWaveletTransform
        return object.__new__(cls)
        
    def general_wavelet_method(self):
        pass


class JOSWaveletTransform(WaveletTransform):
    def __repr__(self):
        return 'JosWaveletTransform()'
        
    def JOS_specific_method(self):
        pass


class pywaveletWaveletTransform(WaveletTransform):
    def __repr__(self):
        return 'pywaveletWaveletTransform()'
```        

with calls working like this

```python
>>> print(WaveletTransform('jos'))
JosWaveletTransform()
>>> print(JOSWaveletTransform())
JosWaveletTransform()
>>> print(WaveletTransform('pywt'))
pywaveletWaveletTransform()
>>> print(pywaveletWaveletTransform())
pywaveletWaveletTransform()
```

kohr-h: Oh, you're dusting off the old idea with the top-level factory class. Yeah, perhaps.

Side note: I know that `__repr__` is just an example, but one should clearly use `self.__class__.__name__` there.

niinimaki:
The following modifications have been made:

- Replaces `elif`s with `else` where applicable and added missing/clarifying `else` statements where necessary.

- removed redundant variables (`ndim` ->`dom.ndim` or `ran.ndim` accordingly, removed redundant `ran_size` `dom_size` where applicable, new deffinition of `max_level` related jos_wavelets removed a lot of redundancies)

- improved error messaged according to @kohr-h suggestion

- changes to lowercase letters `JOS` -> `jos` in code and in the test (and corrected the spelling error/erros)

- `filterlength = int(wbasis[9])` -> `filterlength = int(wbasis[-1])` and saved the variable in `BiorthWaveletTransform` classes

- `try` - `import` - `except` also modified in `wavelet.py`

Test seem to fail due to ` NameError: global name 'pywt' is not defined`...

I suppose that the next (big) thing would then be the restructuring @kohr-h mentioned...

kohr-h:
> The following modifications have been made:

Being picky now: please don't use `ENH` in all commit messages. That one is not intended for general "improvement" of code, which could be anything, but rather real improvements, e.g. speed optimizations, feature additions etc.
What you did was mostly refactoring, removal of redundant code, logic simplifications (these could possibly be ENH), spelling fixes, which are MAINT. If you fix a bug, use BUG. If you add a test, use TST. If you modify/add documentation, use DOC.

> Test seem to fail due to NameError: global name 'pywt' is not defined...

A bunch of tests fail since `bwt` is not skipped if not present. You need to add a skip decorator similar to the `skip_if_no_pywavelets`. If there are also tests with `pywavelets` failing, check if all those are decorated properly.

> I suppose that the next (big) thing would then be the restructuring @kohr-h mentioned...

I'll look into it.

niinimaki:
> What you did was mostly refactoring, removal of redundant code, logic simplifications (these could possibly be ENH), spelling fixes, which are MAINT. If you fix a bug, use BUG. If you add a test, use TST. If you modify/add documentation, use DOC.

Thanks for the reminder on proper commits :)

> You need to add a skip decorator similar to the skip_if_no_pywavelets.

Besides addind `@skip_if_no...` in the wavelet_test.py I then need to add something like

`    skip_if_no_joswavelets = pytest.mark.skipif(`
`        "not odl.trafos.wavelet.JOSWAVELETS_AVAILABLE",`
`        reason='jos_wavelet not available'`
    )

in odl.utils.testutils, right? Are there other places where I should add lines related to skipping if `bwt` not available??

kohr-h:
> Besides addind @skip_if_no... in the wavelet_test.py I then need to add something like
> 
> skip_if_no_joswavelets = pytest.mark.skipif(
> "not odl.trafos.wavelet.JOSWAVELETS_AVAILABLE",
> reason='jos_wavelet not available'
> )
> 
> in odl.utils.testutils, right? 

Yes. The tests and this decorator will later go to JO's repo, together with the rest.

> Are there other places where I should add lines related to skipping if bwt not available??

Only in the tests that are using it.

niinimaki:
Actually I think that we do not need

`    @property`
`    def is_biorthogonal(self):`
`...`

since it is not used afterward unlike `is_orthogonal`
This could partly reduce the unclearness of having biorthogonal wavelets in pyWavelet but still a separate library for (another type of) biorthogonal wavelets...

Opinions?

adler-j:
In general we prefer a `is_...` method over adding a new class (the removal of `LinearOperator` ages ago is an example of this). The reason is that if you make a new class, users have to use it and they have to do more work do distinguish cases:

```python
if operator.is_linear:
    pass
```

vs

```python
if isinstance(operator, odl.LinearOperator):
    pass
```

And you also get some real issues for corner cases where it is conditionally linear:

```python
class AdditionOperator(odl.Operator)
    """ Calculates ax + b, is linear of b=0"""
    def __init__(self, a, b):
        super().__init__(RealNumbers(), RealNumbers(), linear=(b==0))
    def _call(x):
        return a * x + b 
```

vs:

```python
class AdditionOperator(odl.Operator)
    """ Calculates ax + b, is linear of b=0"""
    def __new__(cls, a, b):
        # This is one way to implement it, the other would be to create two classes and a factory
        if b==0:
            # Create a subclass that is linear
            return type('AdditionOperatorLinear', (AdditionOperator, odl.LinearOperator), {})(a, b)
        else:
            return odl.Operator.__new__(cls, a, b)
    def __init__(self, a, b):
        super().__init__(RealNumbers(), RealNumbers())
    def _call(x):
        return a * x + b 
```

I suppose many of these issues could pop up for the biorthonromal case aswell, someone builds a function that takes an arbitrary `WaveletTransform` and returns something, then he would need to make two classes, one for `Biorthonormal` and one for the other.

niinimaki:
The following modifications have been made:

- added skips for test if `jos_wavelet` not available

- make `mode` to have a default value in `WaveletTransform` and `InverseWaveletTransform`

- Added more comments about Jan-Olov's wavelets into the docstrings of biorthogonal wavelet classes

kohr-h:
Okay, I'll see what I can do regarding the code splitting and refactoring.

adler-j:
So what's happening here?

kohr-h:
Not so much lately. Code needs to be moved from one repo to the other.

adler-j:
I guess we close this? Any chance this comes in?

kohr-h:
Well I did some of the work here but honestly I don't quite see the point of putting more time into it. The least we can do is to move the corresponding code to the other repo, albeit in non-working state.

adler-j:
If it can be moved "cleanly" then sure go ahead.

adler-j:
Did you make a new repo somewhere? Then link here if someone manages to google their way here.

adler-j:
https://github.com/odlgroup/jos_wavelets

kohr-h:
Oh, it's been around for a while. I just pushed over the stuff from here now.

Issue #273, opened 2016-02-10
Implement linear deformations using reproducing kernels


adler-j:
We need to implement linear deformations of the form

```
g(x) = f(x + v(x))
```

where `f(x)` is a `DiscreteLp` vector and `v(x)` is given by

```
v(x) = sum_i K(x, y_i) alpha_i
```

Where `K` is a n x n matrix (n dimension of space) and `alpha_i` are coefficients in positions `y_i`.

adler-j:
Will this truly be in core ODL? Close?

chongchenmath:
Close please.

PR #276, opened 2016-02-10
Issue 273 add linear deformations

adler-j:
This is an initial commit, work in progress on #273, example of using it [available in odl-examples](https://github.com/odlgroup/odl-examples/blob/master/linear_deformation.py).

### DIFF
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -316,19 +316,63 @@ class RawDiscretizationVector(NtuplesBaseVector):
         else:
             self.ntuple.__setitem__(indices, values)
 
-    @property
-    def restriction(self):
-        """The restriction operator associated with this vector.
+    def restriction(self, ufunc):
+        """Restrict a continuous function and assign to this vector
+
+        Parameters
+        ----------
+        ufunc : ``self.space.uspace`` element
+            The continuous function that should be restricted.
+
+        Examples
+        --------
+        >>> import odl
+        >>> import numpy as np
+
+        Create discretization
+
+        >>> X = odl.uniform_discr(0, 1, 5)
+        >>> x = X.element()
+
+        Assign x according to continuous vector
+
+        >>> x.restriction(lambda x: x)
+        >>> print(x) # Print values at gridpoints (which are centered)
+        [0.1, 0.3, 0.5, 0.7, 0.9]
 
         See Also
         --------
         RawDiscretization.restriction : For full description
         """
-        return self.space.restriction(self.ntuple)
+        self.space.restriction(ufunc, out=self.ntuple)
 
     @property
     def extension(self):
         """The extension operator associated with this vector.
+
+        Returns
+        -------
+        extension_op : `FunctionSetMapping`
+            Operatior representing a continuous extension of this vector.
+
+        Examples
+        --------
+        >>> import odl
+        >>> import numpy as np
+
+        Create continuous extension of 1d function with nearest neighbour
+
+        >>> X = odl.uniform_discr(0, 1, 3, nodes_on_bdry=True)
+        >>> x = X.element([0, 1, 0])
+        >>> x.extension(np.array([0.24, 0.26]))
+        array([ 0.,  1.])
+
+        Create continuous extension of 1d function wiht linear interpolation
+
+        >>> X = odl.uniform_discr(0, 1, 3, nodes_on_bdry=True, interp='linear')
+        >>> x = X.element([0, 1, 0])
+        >>> x.extension(np.array([0.24, 0.26]))
+        array([ 0.48,  0.52])
 
         See Also
         --------
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -316,6 +316,26 @@ class RawDiscretizationVector(NtuplesBaseVector):
         else:
             self.ntuple.__setitem__(indices, values)
 
+    @property
+    def restriction(self):
+        """The restriction operator associated with this vector.
+
+        See Also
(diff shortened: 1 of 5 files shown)


adler-j:
Example:

template function:

template

deformation, x direction

xdir

y direction

ydir

and the result:

result

looking decent to me, I guess someone could continue this.

odl/discr/discretization.py
```diff
@@ -316,6 +316,26 @@ def __setitem__(self, indices, values):
...
 
+    @property
+    def restriction(self):
+        """The restriction operator associated with this vector.
+
+        See Also
+        --------
+        RawDiscretization.restriction : For full description
+        """
+        return self.space.restriction(self.ntuple)
```
kohr-h: How does that make sense? According to the docstring you want to return the operator. It cannot be applied to the `ntuple`.

odl/discr/discretization.py
```diff
@@ -316,6 +316,26 @@ def __setitem__(self, indices, values):
...
+
+    @property
+    def extension(self):
+        """The extension operator associated with this vector.
+
+        See Also
+        --------
+        RawDiscretization.extension : For full description
+        """
+        return self.space.extension(self.ntuple)
```
kohr-h: See above. While this makes sense, the docstring isn't quite correct. If you apply it to the `ntuple` I wouldn't make it a property since this could possibly be expensive.

odl/space/base_ntuples.py
```diff
@@ -271,8 +276,8 @@ def size(self):
 
     @property
     def shape(self):
-        """Shape of this vector, equals ``(size,)``."""
-        return (self.size,)
+        """Shape of this vector."""
```
kohr-h: Rephrase? Number of entries per axis, for example.

odl/space/ntuples.py
```diff
@@ -132,6 +132,30 @@ def element(self, inp=None, data_ptr=None):
...
+        """Create a vector of ones.
+
+        Examples
+        --------
+        >>> c3 = Cn(3)
+        >>> x = c3.one()
+        >>> x
+        Cn(3).element([(1+0j), (1+0j), (1+0j)])
+        """
+        return self.element(np.ones(self.size, dtype=self.dtype))
```
kohr-h: Are we okay with these methods to return (basically) garbage for non-scalar types?

odl/space/ntuples.py
```diff
@@ -1018,6 +1017,10 @@ def __repr__(self):
         inner_str += _repr_space_funcs(self)
         return '{}({})'.format(class_name, inner_str)
 
+    # Copy these to handle bug in ABCmeta
```
kohr-h: How about scrapping ABCMeta completely?

odl/space/ntuples.py
adler-j: Well, yeah I'll agree it is garbage, but if you insist on calling them, why not? At least the `zero` guy returns something remotely decent:

```
>>> np.zeros(5, dtype='U1')
array([u'', u'', u'', u'', u''], 
      dtype='<U1')
```

odl/space/base_ntuples.py
adler-j: Good idea

odl/discr/discretization.py
adler-j: My fault here, I simply copied from `extension`. I suppose having one of these for `restriction` does not make much sense.

odl/discr/discretization.py
adler-j: Very initial stuff, but I quite like this one:

```python
>>> vec.extension(point)
```

In general users will want to use the vectorized one anyway:

```python
>>> vec.extension(array_or_meshgrid)
```

and that one should be quite fast, no?

odl/space/ntuples.py
kohr-h: Sure, the methods never die, even for `dtype=object` you get *something*. For strings the `zeros` makes sense, for `bool` both `zeros` and `ones`. Perhaps it's really just corner cases where stuff can get bizarre. Probably we can live with them.

adler-j: Yeah for ones I dont know what to expect but you get:

```python
>>> np.ones(5, dtype='U3')
array([u'1', u'1', u'1', u'1', u'1'], 
      dtype='<U3')
```

odl/discr/discretization.py
kohr-h: > and that one should be quite fast, no?

For the current extensions we have, this is instantaneous since it just creates the interpolator and returns it. I was more thinking of possible future applications, where this extension can involve significant computations (e.g. calculating a tessellation?). The syntax is nice, admittedly, but we may have to change it later and should consider the slightly more ugly non-property syntax.

adler-j: I personally find this quite similar to our `adjoint` syntax. Most guys are quick, but some could end up slow. The gains of the syntax is kindoff worth it anyway.

adler-j:
This branch has evolved to do other things than what it was originally intended to do. I propose merge and make a new branch. Fixes according to commits have been made.

kohr-h:
Yeah, merge it and make a new branch for the original task.

odl/discr/discretization.py
```diff
@@ -316,6 +316,70 @@ def __setitem__(self, indices, values):
         else:
             self.ntuple.__setitem__(indices, values)
 
+    def restriction(self, ufunc):
```
kohr-h: `ufunc` sounds lik `numpy.ufunc`

Issue #217, opened 2016-01-13
DiscreteLp.show() displays two identical figures


kohr-h:
As the title says, when calling `show()` without arguments, two figures appear.

kohr-h:
Update: that must have something to do with the interactive mode. If I additionally specify `show=False`, it still shows a plot - in the console if I plot inline, otherwise in a window.

adler-j:
It is because show returns the fig object, and apparently that causes it to be shown in IPython.

kohr-h:
> It is because show returns the fig object

Do we always have to do that?

adler-j:
> Do we always have to do that?

No, but it is convenient to get the image when you plot.

PR #278, opened 2016-02-12
Issue 277  show

adler-j:
Some fixes, now works with (at least)

- [x] Windows
  - [x] IPython in spyder
  - [x] Python in spyder
  - [x] Python commandline

Please verify for other platforms and we can finally merge. The following should be tested:

* All tests in examples/visualization.
* Some test using ShowPartial, for example `odl-examples/pde_laplacian.py`.

Ping @kohr-h, @moosmann

### DIFF
--- a/examples/visualization/show_ntuple.py
+++ b/examples/visualization/show_ntuple.py
@@ -22,4 +22,4 @@ import odl
 spc = odl.Rn(5)
 vec = spc.element([1, 2, 3, 4, 5])
 
-vec.show()
+vec.show(show=True)
--- a/examples/visualization/show_update_2d.py
+++ b/examples/visualization/show_update_2d.py
@@ -21,6 +21,7 @@
 from __future__ import print_function, division, absolute_import
 from future import standard_library
 standard_library.install_aliases()
+
 import odl
 import matplotlib.pyplot as plt
 
--- a/odl/solvers/util/partial.py
+++ b/odl/solvers/util/partial.py
@@ -242,42 +242,40 @@ class ShowPartial(Partial):
 
     """Show the partial result."""
 
-    def __init__(self, *args, **kwargs):
-        """ Create a show partial
+    def __init__(self, **kwargs):
+        """Create a show partial
 
-        parameters are passed through to the vectors show method. Additional
+        Parameters are passed through to the vectors show method. Additional
         parameters include
 
         Parameters
         ----------
+        kwargs
+            Arguments to the show method
         display_step : positive `int`
             Number of iterations between plots. Default: 1
         args, kwargs :
             Optional arguments passed on to ``x.show``
         """
-        self.args = args
         self.kwargs = kwargs
-        self.fig = None
+        self.fig = kwargs.pop('fig', None)
         self.display_step = kwargs.pop('display_step', 1)
         self.iter = 0
 
     def __call__(self, x):
         """Show the current iterate."""
         if (self.iter % self.display_step) == 0:
-            self.fig = x.show(fig=self.fig, show=True,
-                              *self.args, **self.kwargs)
+            self.fig = x.show(fig=self.fig, **self.kwargs)
 
         self.iter += 1
 
-    def __str__(self):
-        """Return ``str(self)``"""
-        return '{}(*{}, **{})'.format(self.__class__.__name__, self.args,
-                                      self.kwargs)
-
     def __repr__(self):
         """Return ``repr(self)``"""
-        return '{}(*{!r}, **{!r})'.format(self.__class__.__name__, self.args,
-                                          self.kwargs)
+        return '{}(display_step={}, fig={}, **{!r})'.format(
+            self.__class__.__name__,
+            self.display_step,
+            self.fig,
+            self.kwargs)
 
 
 if __name__ == '__main__':
--- a/odl/util/graphics.py
+++ b/odl/util/graphics.py
@@ -191,7 +191,7 @@ def show_discrete_data(values, grid, method='', title=None,
             updatefig = False
         else:
             # Set current figure to given input
-            plt.figure(fig.number)
+            fig = plt.figure(fig.number)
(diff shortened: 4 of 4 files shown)


moosmann:
Add `plt.show()` at the end of `odl-examples/pde_laplacian.py`, otherwise the figure window freezes after execution and I have to force quit.

Using spyder with ipython and inline plotting, all examples which use figure updates (`odl-examples/pde_laplacian.py`, `show_update_*d.py`) create an extra inline image in each iteration. Plotting into the same figure doesn't work in `show_1d.py`. Instead of that three single images are created inline. Not using inline plotting, the examples work normally.

Using PyCharm or ipython/python command line, examples with `show=True` (or `plt.show()`) block code execution until the figure window is close unless I'm using `plt.ion()` or `show=False` with `plt.show(block=False)` or Spyder with ipython and inline off.

Except that, all examples in examples/visualization and `odl-examples/pde_laplacian.py` are working on my machine (Ubuntu 14.04) when using

- PyCharm
- python command line
- ipython command line
- Spyder ipyhton inline off

adler-j:
> Add plt.show() at the end of odl-examples/pde_laplacian.py, otherwise the figure window freezes after execution and I have to force quit.

Will do.

> Using spyder with ipython and inline plotting, all examples which use figure updates (odl-examples/pde_laplacian.py, show_update_*d.py) create an extra inline image in each iteration. Plotting into the same figure doesn't work in show_1d.py. Instead of that three single images are created inline. Not using inline plotting, the examples work normally.

Do you mean that a new image is created with the next iterate, or that the old image is updated and an empty new image is created? What three images are shown in the 1d case? How is the 1d case different from the 2d case?

> Using PyCharm or ipython/python command line, examples with show=True (or plt.show()) block code execution until the figure window is close unless I'm using plt.ion() or show=False with plt.show(block=False) or Spyder with ipython and inline off.

This is the expected behaviour of matplotlib, so that will stay.

> Except that, all examples in examples/visualization and odl-examples/pde_laplacian.py are working on my machine (Ubuntu 14.04) when using

> - PyCharm
> - python command line
> - ipython command line
> - Spyder ipyhton inline off
    
Great work, thanks!

kohr-h:
> Using spyder with ipython and inline plotting, all examples which use figure updates (odl-examples/pde_laplacian.py, show_update_*d.py) create an extra inline image in each iteration. Plotting into the same figure doesn't work in show_1d.py. Instead of that three single images are created inline. Not using inline plotting, the examples work normally.

This is the same behavior I pointed out before, see #217. It will help to adapt the examples such that the returned figure is stored somewhere, which suppresses immediate inline plotting (hopefully).

kohr-h:
I'll check this later today, but since I use roughly the same configuration as @moosmann, results won't be too different I guess. I also have the newest versions of spyder (2.3.8) and matplotlib (1.5.1) installed.

moosmann:
> 

    Using spyder with ipython and inline plotting, all examples which use figure updates (odl-examples/pde_laplacian.py, show_update_*d.py) create an extra inline image in each iteration. Plotting into the same figure doesn't work in show_1d.py. Instead of that three single images are created inline. Not using inline plotting, the examples work normally.

Do you mean that a new image is created with the next iterate, or that the old image is updated and an empty new image is created? What three images are shown in the 1d case? How is the 1d case different from the 2d case?

    Using PyCharm or ipython/python command line, examples with show=True (or plt.show()) block code execution until the figure window is close unless I'm using plt.ion() or show=False with plt.show(block=False) or Spyder with ipython and inline off.

This is the expected behaviour of matplotlib, so that will stay.

    Except that, all examples in examples/visualization and odl-examples/pde_laplacian.py are working on my machine (Ubuntu 14.04) when using

        PyCharm
        python command line
        ipython command line
        Spyder ipyhton inline off

Great work, thanks!

moosmann:
Oops. Misclicked. Reopened.

adler-j:
> A new image is created not updated. In the 1d case, three images with a single plot are created. But I think the behavior is also related to which matplotlib version, backends, etc ist used. I just ran the show_1d.py example on Chong's machine in Spyder with inline plotting and it worked properly. In the 2d case the same behavior is observed i.e. in each iterate a new image is created.

I think we will have to live with that, IPython seems to invalidate old fig handles once shown so we cant update them in any sensible manner.

Waiting for @kohr-h to test this then I'll pull.

moosmann:
> I think we will have to live with that, IPython seems to invalidate old fig handles once shown so we cant update them in any sensible manner.

I think so, too. If it doesn't work properly in one case or the other, one still can get it working using proper arguments or pyplot functions. Anyway, very nice functionality.

kohr-h:
Some observations:

- Plotting in native Python console is much faster than in IPython
- When the Laplacian example is done and you close the figure, a new figure is opened with the final result. Only after you close that, the program is done with execution. That doesn't happen with the `show_update_2d` example, so probably the final `plt.show()` command suppresses that.
- I don't have any inline figures during update, that works fine for me.

Nice otherwise.

kohr-h:
I should add that I tested Spyder with both native Python console and IPython.

adler-j:
With what setup do we observe:

> Add plt.show() at the end of odl-examples/pde_laplacian.py, otherwise the figure window freezes after execution and I have to force quit.

kohr-h:
> so probably the final plt.show() command suppresses that.

It does. Maybe that's connected to the freezing issue?

moosmann:
> With what setup do we observe:

> > Add plt.show() at the end of odl-examples/pde_laplacian.py, otherwise the figure window freezes after execution and I have to force quit.

Using PyCharm on Ubuntu. Using python/ipython command line there's no freeze, but without `plt.show()` the window closes as soon as code execution is finished. Using Spyder with python there is no freeze and the window stays open. Using Spyder with ipython and inline plotting there's no freezing (but additional images are created in each loop).

PR #287, opened 2016-02-22
ENH: add aswriteablearray to ntuples_base

adler-j:
Could use a look on the interface, usage is very simple as demonstrated in the doctest

### DIFF
--- a/odl/space/base_ntuples.py
+++ b/odl/space/base_ntuples.py
@@ -205,6 +205,44 @@ class NtuplesBaseVector(with_metaclass(ABCMeta, object)):
             Numpy array of the same type as the space.
         """
 
+    @property
+    def aswriteablearray(self):
+        """Extract the data of this array as a context managed numpy array.
+
+        Parameters
+        ----------
+        args, kwargs
+            Arguments to `asarray`
+
+        Returns
+        -------
+        context_manager
+            A context manager class that manages the writable array.
+
+        Examples
+        --------
+
+        >>> import odl
+        >>> X = odl.CudaRn(3)
+        >>> x = X.zero()
+        >>> with x.aswriteablearray() as arr:
+        ...     arr[:] = [1, 2, 3]
+        >>> x
+        CudaRn(3).element([1.0, 2.0, 3.0])
+        """
+        class ArrayContextManager(object):
+            def __init__(cm_self, *args, **kwargs):
+                cm_self.array = self.asarray(*args, **kwargs)
+
+            def __enter__(cm_self):
+                return cm_self.array
+
+            def __exit__(cm_self, type, value, traceback):
+                # TODO: handle exceptions as expected
+                self[:] = cm_self.array
+
+        return ArrayContextManager
+
     @abstractmethod
     def __getitem__(self, indices):
         """Access values of this vector.
--- a/odl/space/cu_ntuples.py
+++ b/odl/space/cu_ntuples.py
@@ -25,6 +25,7 @@ from builtins import int, super
 
 # External module imports
 import numpy as np
+from contextlib import contextmanager
 
 # ODL imports
 from odl.set.space import LinearSpaceVector
@@ -331,12 +332,26 @@ class CudaNtuplesVector(NtuplesBaseVector, LinearSpaceVector):
         array([1, 2, 3], dtype=uint8)
         >>> result is out
         True
+
+        Can also be used with a context manager to become writable
+
+        >>> x = uc3.element([0, 0, 0])
+        >>> with x.aswriteablearray() as arr:
+        ...     arr[:] = [1, 2, 3]
+        >>> x
+        CudaNtuples(3, 'uint8').element([1, 2, 3])
+
+        See Also
+        --------
+        aswriteablearray : gives a writable array
         """
+
         if out is None:
-            return self.data.get_to_host(slice(start, stop, step))
+            out = self.data.get_to_host(slice(start, stop, step))
         else:
             self.data.copy_device_to_host(slice(start, stop, step), out)
-            return out
+
+        return out
 
     def __getitem__(self, indices):
         """Access values of this vector.


odl/space/cu_ntuples.py
```diff
@@ -25,6 +25,7 @@
 
 # External module imports
 import numpy as np
+from contextlib import contextmanager
```
adler-j: leftover, remove.

odl/space/base_ntuples.py
```diff
@@ -205,6 +205,44 @@ def asarray(self, start=None, stop=None, step=None, out=None):
             Numpy array of the same type as the space.
         """
 
+    @property
+    def aswriteablearray(self):
```
adler-j: dont quite like this name

odl/space/base_ntuples.py
```diff
@@ -205,6 +205,44 @@ def asarray(self, start=None, stop=None, step=None, out=None):
...
+
+        >>> import odl
+        >>> X = odl.CudaRn(3)
+        >>> x = X.zero()
+        >>> with x.aswriteablearray() as arr:
+        ...     arr[:] = [1, 2, 3]
+        >>> x
+        CudaRn(3).element([1.0, 2.0, 3.0])
+        """
+        class ArrayContextManager(object):
```
adler-j: Could be simplified? using https://docs.python.org/2/library/contextlib.html:

```python
@contextmanager
def array_context(*args, **kwargs):
    arr = self.asarray(*args, **kwargs)
    yield arr
    self[:] = arr
```

adler-j: This would make exception handling etc somewhat more complex

kohr-h:
How about simply making `NtuplesBaseVector` a context manager *directly* (supposes we have one obvious way of interpreting it)? This would simply mean taking the `__enter__` and `__exit__` methods to `NtuplesBase` and then writing
```python
>>> x = CudaRn(6).one()
>>> with x as arr:
...     arr.reshape((2, 3))[0, :] = -1
>>> print(x)
[-1, -1, -1, 1, 1, 1]
```

kohr-h:
Other outcomes of `as` could then be implemented as dedicated properties with more complicated names. The easiest and most obvious case could be reserved for Numpy arrays.

adler-j:
> How about simply making `NtuplesBaseVector` a context manager *directly* ...

Nice suggestion, but I'd still say no for two reasons. First, it implies the operation is fast, while it could mean two copies (even as much as a gather operation if we ever implement MPI support). Second, suppose our users want do something else with the vector as a context manager, perhaps a file type vector, maybe memory mapped. Then they might want to

```python
>>> with MemoryMappedRn.element('data/file.dat') as vec:
... do stuff to vec
```

by implementing your suggestion we would lock users out of doing this, or at least risk clashes.

kohr-h:
Regarding the second point, you can always make extra properties for additional, less obvious contexts. I certainly see the first point. Naming can be improved, what about `view`?

kohr-h:
Or perhaps `expose_array`? `with x.expose_array() as array: ...`

adler-j:
> Regarding the second point, you can always make extra properties for additional, less obvious contexts.

Sure, but do we realy want to enforce that our interpretation of "less obvious" is the correct one? Perhaps it would be ok here, the syntax is indeed very nice.

> Or perhaps `expose_array`? `with x.expose_array() as array: ...`

I suppose you dislike the very long name? I do too. Another option would be to make a subclass of `ndarray` and return it from `asarray()`

```python
def asarray(self):
    class context_ndarray(ndarray):
        def __begin__(self2):
            pass
        def __exit(self2):
            self[:] = self2
    return GET_DATA_TO_ARRAY().view(type=context_ndarray)
```

the issue usere is that we would need to rename `asarray` in all subclasses to `GET_DATA_TO_ARRAY`, or write this for each subclass. Another is that `repr` on such an array would show `context_ndarray(...)`.

kohr-h:
> Sure, but do we realy want to enforce that our interpretation of "less obvious" is the correct one? Perhaps it would be ok here, the syntax is indeed very nice.

Well, currently it's the only one, but maybe there will be one even more obvious. Hard to say.

> I suppose you dislike the very long name? I do too.

It's not bad, just a bit long-ish and doubles up the "as". If we have a method, a verb like `expose` sounds good since it fits well into the "sentence" `with x.expose() as array`. A substantive like `view` also works, but "as" shouldn't be part of it since it breaks the flow.

I currently slightly tend to having a good method name and the current functionality. Explicit is better than implicit. The second best, with nicest syntax but "closing doors", would be to make `NtuplesBaseVector` itself a context manager.

adler-j:
Another (similar) feature I just thought about (since my program is taking forever to run on the CPU) would be to have the exact equivalent to this but between implementations:

```
with x.ascudavector() as x2:
    # do computations with cuda
```

This could be easily generalized to any ntuplesbase.

kohr-h:
> This could be easily generalized to any ntuplesbase.

Like the idea! How would we treat type conversions? With a `cast` flag, similar to Numpy?

adler-j:
Well now we're talking about generalizations, I'm quite into this idea. How about we try unifying these into some `astype('numpy')` or similar?

Basically we could do these:

* astype('numpy'): the ndarray
* astype('cpu'): cast a vector with cuda backend to cpu backend
* astype('gpu): inverse of above

And possibly more

kohr-h:
That's a very interesting idea. We could simply start by renaming your implementation to `cast` and take `'array'` as parameter, and then add further options later. Add also `dtype` to the list, then we get quite far with this.

kohr-h:
Shall we try to get a first version (e.g. only with `type='ndarray'`) working for now? There are a couple of places where current code silently doesn't write to CUDA arrays where it should.

How about we start with just the "standard" case of getting a writable NumPy array? As far as I see, this would only amount to changing the current implementation to 

```python
class NtuplesBase(...):
    ...
    def cast(self, type='ndarray', **kwargs):
        """Return a context manager for write access to this vector's data.

        Parameters
        ----------
        type_ : {'ndarray'}, optional
            Data structure type returned by the context manager
        kwargs :
            Additional arguments passed to `asarray`

        Returns
        -------
        context_manager
            A context manager class that manages the writable array.

        Examples
        --------
        >>> import odl
        >>> X = odl.CudaRn(3)
        >>> x = X.zero()
        >>> with x.cast() as arr:
        ...     arr[:] = [1, 2, 3]
        >>> x
        CudaRn(3).element([1.0, 2.0, 3.0])
        """
        # TODO: handle type argument
        vector = self

        class ArrayContextManager(object):
            def __init__(self, *args, **kwargs):
                self.array = vector.asarray(*args, **kwargs)

            def __enter__(self):
                return self.array

            def __exit__(self, type, value, traceback):
                # TODO: handle exceptions as expected
                vector[:] = self.array

        return ArrayContextManager
```

adler-j:
Solved using a functional approach in #524.

PR #289, opened 2016-02-22
Proximal operator for the Kullback-Leibler data divergence

moosmann:
Added the proximal operator of the convex conjugtae of the Kullback-Leibler data divergence including test and example.


examples/solvers/denoising_with_entropy_type_regularization
```diff
@@ -0,0 +1,148 @@
...
+print('Norm of the product space operator: {}'.format(prod_op_norm))
+
+# Proximal operator related to the primal variable
+
+# Non-negativity constraint
+proximal_primal = odl.solvers.proximal_nonnegativity(op.domain)
+
+# Proximal operators related to the dual variable
+
+# l2-data matching
```
adler-j: l2?

examples/solvers/denoising_with_entropy_type_regularization
```diff
@@ -0,0 +1,148 @@
...
+prox_convconj_kl = odl.solvers.proximal_convexconjugate_kl(discr_space,
+                                                           lam=1, g=noisy)
+
+# TV-regularization: l1-semi norm of grad(x)
+prox_convconj_l1 = odl.solvers.proximal_convexconjugate_l1(gradient.range,
+                                                           lam=1/5)
+
+# Combine proximal operators: the order must match the order of operators in K
+proximal_dual = odl.solvers.combine_proximals([prox_convconj_kl,
```
adler-j: Perhaps `combine_proximals` should be `proximal_combined` to match the naming of the others?

examples/solvers/denoising_with_entropy_type_regularization
```diff
@@ -0,0 +1,148 @@
...
+
+# Run algorithms (and display intermediates)
+odl.solvers.chambolle_pock_solver(
+    op, x, tau=tau, sigma=sigma, proximal_primal=proximal_primal,
+    proximal_dual=proximal_dual, niter=niter, partial=partial)
+
+# Display images
+orig.show(title='original image')
+noisy.show(title='noisy image')
+x.show(title='reconstruction')
```
adler-j: by doing show=True here you can skip plt.show() in the next row and thus avoid the import up top. This allows us to change backend later on without pain here.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -375,6 +377,113 @@ def _call(self, x, out):
     return _ProximalConvConjL1
 
 
+# TODO: move notes to ODL doc
+def proximal_convexconjugate_kl(space, lam=1, g=None):
```
adler-j: is `kl` unique enough?

odl/solvers/advanced/proximal_operators.py
```diff
@@ -375,6 +377,113 @@ def _call(self, x, out):
...
+            sig = self.sigma
+
+            # 1 / 2 (lam_X + x - sqrt((x - lam_X) ^ 2 + 4; lam sigma g)
+
+            # TODO: optimize
+            # out = x - lam_X
+            out.lincomb(1, x, -lam, space.one())
+
+            # (out)^2
+            out.ufunc.square(out=out)
```
adler-j: possibly slower than  `out.multiply(out, out)`.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -254,46 +266,126 @@ def test_proximal_factory_convconj_l1_product_space():
...
+    assert isinstance(prox, odl.Operator)
+
+    # Allocate an output vector
+    x_opt = space.element()
+
+    # Apply the proximal operator returning its optimal point
+    prox(x, x_opt)
+
+    # Explicit computation:
+    # 1 / 2 * (lam_X + x - sqrt((x - lam_X)^2 + 4 * lam * sigma * g)
```
adler-j: this is quite obvious from the row below

examples/solvers/denoising_with_entropy_type_regularization
kohr-h: KL-data matching for Poisson noise?

examples/solvers/denoising_with_entropy_type_regularization
kohr-h: or `combined_proximals`

adler-j: Part of the idea here was to force it to begin with `proximal` so it shows up properly in autocomplete.

kohr-h: Ah that was what you meant. `proximal_combination` is also possible.

adler-j:
Whos responsible for getting this in?

kohr-h:
Somebody needs to check if the implementation is correct. Stylistic stuff, tests, etc. I can give it a shot.

adler-j:
Any chance we could try to get this out of the way?

kohr-h:
Not today at least, got some other stuff to do. Just realized that I signed up for this branch, oh well. I'll put it on top of the stack (kind of).

adler-j:
I'll try to get this out of the way.

examples/solvers/denoising_with_entropy_type_regularization.py
```diff
@@ -0,0 +1,112 @@
...
+# Read test image:
+# convert integer values to float, and rotate to get the image upright
+image = np.rot90(scipy.misc.ascent()[::2, ::2], 3).astype('float')
+shape = image.shape
+
+# Rescale
+image *= 100 / image.max()
+
+# Add noise
+noisy_image = np.random.poisson(1 + image)
```
kohr-h: Why `1 + image`? Is it to avoid 0?

odl/solvers/advanced/proximal_operators.py
kohr-h: I don't think so.
On the other hand, shortening `convexconjugate` to `convexconj` wouldn't hurt.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -375,6 +377,112 @@ def _call(self, x, out):
...
+
+        F_cc(p) = sum_i (-g ln(pos(1_X - p))_i + ind_P(1_X - p)
+
+    where p is the variable dual to x, and 1_X is a vector in the space X with
+    all components set to 1.
+
+    The proximal operator of the convex conjugate of F is
+
+        prox_sigma[F_cc](x) = 1/2 (lam_X + x - sqrt((x - lam_X)^2 + 4 lam
+        sigma g)
```
kohr-h: Regarding math explanations in docstrings: I think we should consider doing it in the NumPy/SciPy style, namely to give only a short explanation in the class docstring (hardly any math) and then use the `Notes` section to introduce the math. There one can go wild and use LaTeX commands and all that. See [numpydoc](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections), points 3. and 10.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -42,41 +43,47 @@ def test_proximal_zero():
...
+    out = space.element()
 
-    # Factory function for the proximal operator
-    make_prox = proximal_nonnegativity(x_space)
+    # Factory function returning the proximal operator
+    make_prox = proximal_nonnegativity(space)
 
     # Initialize proximal operator of G with sigma
-    prox_op = make_prox(None)
+    prox = make_prox(None)
```
kohr-h: The comment doesn't make much sense.

examples/solvers/denoising_with_entropy_type_regularization.py
adler-j: I guess so, I'll add a comment to this.

odl/solvers/advanced/proximal_operators.py
adler-j: I basically removed all of this, noone will read it anyway. We should however have a math part of the doc where we do this in the general case and refer to these as special cases.

test/solvers/advanced/proximal_operator_test.py
adler-j: :D

kohr-h:
Looking good to me

PR #292, opened 2016-03-08
DOC: add faq on add error, glossary on dtype, close #271

### LINKED ISSUE #271: Bad error message for e.g. addition of elements from different spaces
This is what happens currently:
```python
>>> discr1 = odl.uniform_discr(0, 1, 10)
>>> discr2 = odl.uniform_discr(0, 1, 11)
>>> discr1.one() - discr2.one()
...
TypeError: unsupported operand type(s) for -: 'DiscreteLpVector' and 'DiscreteLpVector'
```
For users unfamiliar with ODL, this is quite devastating since it's obviously wrong and gives no clue how to fix it.

The problem is that we return `NotImplemented` in all situations we cannot do anything. But aren't there situations where we can be sure that the operation will never make sense and we should say exactly why?

If there is no way to reliably fix this behavior, we definitely need an FAQ entry.

adler-j:


adler-j:
All comments fixed, pull?

kohr-h:
Alrightey!

PR #293, opened 2016-03-08
ENH: change order of show to have title first in all cases, close #252

### LINKED ISSUE #252: Change argument of show
Currently the first argument is in general `method`, this is basically never used, and could be changed to something more simple. Imo, there are two options:

* Change to `title` since this is quite intuitive:

    ```python
    x.show('plot name')
    ```
    
    Pros:
    * Easy to understand from code what is meant in most cases
    * Common usecase when using ODL
    
* Change to `indices` since this is most commonly used:

    ```python
    x.show(np.s_[:, 50, 50])
    ```
        
    Pros:
    * Easily the most commonly used parameter, especially with `ShowPartial`.

Thoughts?

adler-j:
Changed title to come first since this is the only real standard we can apply to all cases, and it makes sense somewhat.

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -608,7 +608,7 @@ class DiscreteLpVector(DiscretizationVector):
         grid = self.space.grid[indices].squeeze()
         values = self.asarray()[indices].squeeze()
 
-        return show_discrete_data(values, grid, title=title, method=method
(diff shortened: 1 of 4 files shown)


kohr-h:
Agree, I often ended up getting errors when leaving out the parameter name.

kohr-h:
Looking good

PR #304, opened 2016-03-10
Issue 303  spect geometry

niinimaki:
Added a class for SPECT parallel hole geometry and tests for it.

### DIFF
--- a/odl/tomo/geometry/parallel.py
+++ b/odl/tomo/geometry/parallel.py
@@ -57,8 +57,7 @@ class ParallelGeometry(Geometry):
         detector : `Detector`
             The detector to use in this geometry
         det_init_pos : `array-like`
-            Initial position of the detector reference point. The zero
-            vector is not allowed.
+            Initial position of the detector reference point
         """
         super().__init__(ndim, apart, detector)
 
--- a/odl/tomo/geometry/spect.py
+++ b/odl/tomo/geometry/spect.py
@@ -1,0 +1,84 @@
+# Copyright 2014 - 2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Single-photon emission computed tomography (SPECT) geometry."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import super
+
+import numpy as np
+
+from odl.tomo.geometry.geometry import AxisOrientedGeometry
+from odl.tomo.geometry.parallel import Parallel3dAxisGeometry
+from odl.tomo.util.utility import perpendicular_vector
+
+__all__ = ('ParallelHoleCollimatorGeometry', )
+
+
+class ParallelHoleCollimatorGeometry(Parallel3dAxisGeometry):
+
+    """Geometry for SPECT Parallel hole collimator."""
+
+    def __init__(self, apart, dpart, det_rad, axis=[0, 0, 1],
+                 **kwargs):
+        """Initialize a new instance.
+        Parameters
+        ----------
+        apart : 1-dim. `RectPartition`
+            Partition of the angle interval
+        dpart : 2-dim. `RectPartition`
+            Partition of the detector parameter rectangle
+        det_rad : positive `float`
+            Radius of the circular detector orbit.
+        axis : `array-like`, shape ``(3,)``, optional
+            Fixed rotation axis.
+        orig_to_det_init : `array-like`, shape ``(3,)``, optional
+            Vector pointing towards the detector reference point in
+            the initial position.
+            Default: a `perpendicular_vector` to ``axis``.
+        det_init_axes : 2-tuple of `array-like` (shape ``(3,)``), optional
+            Initial axes defining the detector orientation.
+            Default: the normalized cross product of ``axis`` and
+            ``orig_to_det_init`` is used as first axis and ``axis`` as second.
+        """
+        self._det_radius = float(det_rad)
+        if self.det_radius <= 0:
+            raise ValueError('expected a positive radius, got {}.'
+                             ''.format(det_rad))
+
+        orig_to_det_init = kwargs.pop('orig_to_det_init',
+                                      perpendicular_vector(axis))
+
+        init_pos_norm = np.linalg.norm(orig_to_det_init)
+        if init_pos_norm > 1e-10:
+            orig_to_det_init *= self.det_radius / init_pos_norm
+        else:
+            raise ValueError('initial detector position {} is too close '
+                             'to zero.'.format(orig_to_det_init))
+        kwargs['det_init_pos'] = orig_to_det_init
+        super().__init__(apart, dpart, axis, **kwargs)
+
+    @property
+    def det_radius(self):
+        """Radius of the detector orbit."""
+        return self._det_radius
+
+    # Fix for bug in ABC thinking this is abstract
+    rotation_matrix = AxisOrientedGeometry.rotation_matrix
--- a/test/tomo/geometry/spect_geometry_test.py
+++ b/test/tomo/geometry/spect_geometry_test.py
@@ -1,0 +1,54 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test ODL geometry objects for SPECT."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+
+standard_library.install_aliases()
+
+# External
+import pytest
+import numpy as np
+
+# Internal
+import odl
+from odl.util.testutils import all_equal
+from odl.tomo.geometry.spect import ParallelHoleCollimatorGeometry
+
+
+def test_spect():
+    det_nx_pix = 64
+    det_ny_pix = 64
+    det_nx_mm = 4
+    det_radius = 200
+    n_proj = 180
+    det_param = det_nx_mm * det_nx_pix
+    dpart = odl.uniform_partition([-det_param, -det_param],
+                                  [det_param, det_param],
+                                  [det_nx_pix, det_ny_pix])
+
+    apart = odl.uniform_partition(0, 2 * np.pi, n_proj)
+    geom = ParallelHoleCollimatorGeometry(
+        apart, dpart, det_rad=det_radius)
+    assert isinstance(geom.detector, odl.tomo.Flat2dDetector)
+    assert all_equal(geom.det_radius, det_radius)
+
+if __name__ == '__main__':
+    pytest.main(str(__file__.replace('\\', '/')) + ' -vs')
--- a/odl/tomo/geometry/__init__.py
+++ b/odl/tomo/geometry/__init__.py
@@ -39,3 +39,6 @@ __all__ += fanbeam.__all__
 from . import conebeam
 from .conebeam import *
 __all__ += conebeam.__all__
+
+from .spect import *
+__all__ += spect.__all__


odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+        center_to_det = perpendicular_vector(axis)
+        self._center_to_det = (np.array(center_to_det) /
+                               np.linalg.norm(center_to_det))
+
+        det_init_axis_0 = np.cross(self.axis, self._center_to_det)
+        det_init_axes = (det_init_axis_0, axis)
+        detector = Flat2dDetector(dpart, det_init_axes)
+
+        self._detector = detector
+        self._motion_params = odl.Interval(apart.points()[0],
```
adler-j: apart.begin, apart.end

adler-j: Tbh why not simply apart.set?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+                                           apart.points()[-1])
+        self._motion_part = apart
+
+    @property
+    def motion_partition(self):
```
adler-j: I dont see why we re-run these? You should call Geometry.__init__ in the constructor, which will give you all of these for free.

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        return self.partition.set
+
+    @property
+    def grid(self):
+        """Joined sampling grid for motion and detector.
+
+        By convention, the motion grid comes before the detector grid.
+        """
+        return self.partition.grid
```
adler-j: All the way down here can be removed.

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+import odl
+from odl.tomo.geometry.detector import Flat2dDetector
+from odl.tomo.geometry.geometry import (Geometry,
+                                        AxisOrientedGeometry)
+from odl.tomo.util.utility import perpendicular_vector
+
+__all__ = ('ParallelHole')
+
+
+class ParallelHole(AxisOrientedGeometry, Geometry):
```
adler-j: Perhaps some longer name?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+import numpy as np
+
+# Internal
+import odl
+from odl.tomo.geometry.detector import Flat2dDetector
+from odl.tomo.geometry.geometry import (Geometry,
+                                        AxisOrientedGeometry)
+from odl.tomo.util.utility import perpendicular_vector
+
+__all__ = ('ParallelHole')
```
adler-j: This creates an all that looks like:

```
__all__ = ('P', 'a', 'r', 'a', 'l', 'l', 'e', 'l', 'H', 'o', 'l', 'e')
```

you need an extra `,` at the end to make this proper:

```
__all__ = ('ParallelHole',)
```

test/tomo/geometry/spect_geometry_test.py
```diff
@@ -0,0 +1,57 @@
...
+    dpart = odl.uniform_partition([-a, -a], [a, a],
+                                  [det_nx_pix, det_ny_pix])
+
+    apart = odl.uniform_partition(0, 2 * np.pi, n_proj, nodes_on_bdry=True)
+    geom = ParallelHole(apart, dpart, det_radius)
+    assert isinstance(geom.detector, odl.tomo.Flat2dDetector)
+    assert all_equal(geom.det_radius, det_radius)
+
+    angles = geom.motion_grid.points()
+    assert all_equal(float(angles[0]), 0)
```
adler-j: assert angles[0] = 0

test/tomo/geometry/spect_geometry_test.py
```diff
@@ -0,0 +1,57 @@
...
+    assert all_equal(float(angles[-1]), 2*np.pi)
```
adler-j: asset angles[-1] == 2 * np.pi

But be careful with numerics here...

test/tomo/geometry/spect_geometry_test.py
```diff
@@ -0,0 +1,57 @@
...
+from odl.tomo.geometry.spect_geometry import ParallelHole
+
+
+def test_spect():
+    det_nx_pix = 64
+    det_ny_pix = 64
+    det_nx_mm = 4
+    det_radius = 200
+    n_proj = 180
+    a = det_nx_mm * det_nx_pix
```
adler-j: rename `a` to something else

adler-j:
Did a short review

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+                 detector_axes=None):
+        """Initialize a new instance.
+        Parameters
+        ----------
+        apart : 1-dim. `RectPartition`
+            Partition of the angle interval
+        dpart : 2-dim. `RectPartition`
+            Partition of the detector parameter rectangle
+        det_radius : `float`
+            Radius of the detector circle (mm), must be positive
```
kohr-h: We shouldn't mix units. Not sure what the rest of the geometries assume. I guess SI units (i.e. meters)?

And write
    positive `float`

Perhaps there are a couple of places in the rest of the geometries where we should adapt notation in the doc.

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+
+        AxisOrientedGeometry.__init__(self, axis)
+
+        self._det_radius = float(det_radius)
+
+        # Perpendicular vector from the center of rotation to the detector
+        center_to_det = perpendicular_vector(axis)
```
kohr-h: This can be the default, but we should allow users to specify this vector. Check one of the cone beam geometries for reference.

odl/tomo/geometry/spect_geometry.py
kohr-h: Both `_detector` and `_motion_params` will be set by `super().__init__(apart, dpart)`. For this, `Geometry` must be the first parent class (see above).

odl/tomo/geometry/spect_geometry.py
kohr-h: Parent classes the other way around.

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __init__(self, apart, dpart, det_radius, axis=[0, 0, 1],
+                 detector_axes=None):
+        """Initialize a new instance.
+        Parameters
+        ----------
+        apart : 1-dim. `RectPartition`
+            Partition of the angle interval
+        dpart : 2-dim. `RectPartition`
+            Partition of the detector parameter rectangle
```
kohr-h: What do we do if we have two (non-identical) detectors?

odl/tomo/geometry/spect_geometry.py
adler-j: We never use units in any class, we simply require the user to be consistent, we should try to follow this standard imo.

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,156 @@
...
+        -------
+        point : `numpy.ndarray`, shape (3,)
+            The reference point on a circle in the transaxial plane with
+            radius ``R`` and on at a axial position ``z``.
+            Transaxial point given by the rotation angle ``phi``.
+        """
+        origin_to_det = self.det_radius * self._center_to_det
+        circle_component = self.rotation_matrix(angle).dot(origin_to_det)
+
+        return circle_component
```
kohr-h: Having read this, I don't see a substantial difference to `Parallel3dAxisGeometry` except that there is a radius for the detector. Perhaps easiest is to just subclass that one and add the radius?

niinimaki: BTW, for the SPECT projector, I need also to define the PSF (its FWHM for example). Should this be set using the geometry class or set when creating the SPECT projector?

odl/tomo/geometry/spect_geometry.py
niinimaki: Maybe a new geometry class for that if it is needed??

odl/tomo/geometry/spect_geometry.py
niinimaki: All parameters (from what I know) are in mm, but I can remove that (mm requirement info) to be consistent with the other classes.

odl/tomo/geometry/spect_geometry.py
kohr-h: Yes, perhaps, my question was not rhetoric. I don't exactly know how to solve this most easily.

One basic way would be to allow the `dpart` to be 3-dimensional with `k` grid points (= k subintervals) in the third axis, to get `k` detectors.

odl/tomo/geometry/spect_geometry.py
kohr-h: Yes, do so. We should write somewhere that all parameters are assumed to be given in the same unit system, regardless which one.

odl/tomo/geometry/spect_geometry.py
kohr-h: Is this PSF due to the dectector (imperfect collimation) or due to finite source sizes? In the former case, the way to go is to create a new detector class (subclassing `Flat2dDetector`) and add the PSF to it (given by the user either as a function, an array or a parameter for a standard PSF model, Gaussian or whatever). In the latter case, the projector needs to handle it.

adler-j: If it is in the detector, I'd finally implement it by composing an idela detector with a convolution, easier separation of responsibilities.

kohr-h: Yes, but that convolution could get its kernel from the detector. The SPECT forward operator could then choose to use it or not (compose the line projection with a convolution).

niinimaki:
Modified the code and test according to the reviews:

- changed the order of the parent classes and did the related changes

- removed redundancy in code

- changed the name from `ParallelHole` to `ParallelHoleCollimatorGeometry`

- The initial vector pointing from the center of  rotation to the detector reference point is perpendicular by default but is allowed to be determined by the user

- modified the test accordingly

Remains TODO:

- proper handing if we have two (non-identical) detectors

- add PSF to the geometry class

niinimaki:
Added parameters to the initialization of spect geometry that are related to defining the PSF.

niinimaki:
Added a way to handle situation when there are several radii for SPECT. This would apply for cases when the detector head does not move on a perfect circular trajectory but follows the "contours" of a patient. 
All checks seem to have passed. Could you review?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self, apart, dpart, det_radius, axis=[0, 0, 1],
+                 **kwargs):
+        """Initialize a new instance.
+        Parameters
+        ----------
+        apart : 1-dim. `RectPartition`
+            Partition of the angle interval
+        dpart : 2-dim. `RectPartition`
+            Partition of the detector parameter rectangle
+        det_radius : `list` of positive `float`
```
kohr-h: `list` -> `sequence`
Also, `det_rad` would be a better name now since it's undetermined if singular or plural.

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""SPECT geometry."""
```
kohr-h: File name should be only `spect.py`.

niinimaki: OK. To be consistent, should the test for spect geometry also be simply added to the `geometry_test.py`?, now they are in `spect_gemetery_test.py`

adler-j: A different file is definitely preferable

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+import odl
+from odl.tomo.geometry.detector import Flat2dDetector
+from odl.tomo.geometry.geometry import (Geometry,
+                                        AxisOrientedGeometry)
+from odl.tomo.util.utility import perpendicular_vector
+
+__all__ = ('ParallelHoleCollimatorGeometry', )
+
+
+class ParallelHoleCollimatorGeometry(Geometry, AxisOrientedGeometry):
```
kohr-h: I'm not quite happy with the name. What is this "hole" about anyway? We should think about the different geometry types that are used in SPECT and what distinguishes them. We don't have to repeat all kinds of technical details.

So what is characteristic of this geometry? Mathematically, it's (correct me if I'm wrong) in the simplest case just a standard single-axis parallel 3D geometry. To take the imperfect collimation into account, one needs the additional information of the detector position and some PSF parameters. In the full-blown and most generic setting, each detector pixel has its own field of view, which means that there is not a unique direction of incoming rays, but a bunch of directions which are integrated over. Our general geometry model would allow that, but I don't think any forward projector supports that.

So my conclusion is that

- this geometry should subclass `Parallel3dAxisGeometry` and by default be identical to it,
- offer *optional* parameters for detector positioning,
- offer *optional* parameters for the collimators.

And we should make an attempt at understanding what the SPECT forward model really is about. Where does this PSF come from and on which parameters does it depend?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        ----------
+        apart : 1-dim. `RectPartition`
+            Partition of the angle interval
+        dpart : 2-dim. `RectPartition`
+            Partition of the detector parameter rectangle
+        det_radius : `list` of positive `float`
+            A list of radii for every position of the detector
+            Individual radii must be positive.
+            If all radii are equal, `list` of one radius is sufficient,
+            otherwise a radius per angular position is needed.
```
kohr-h: Ah, I think you misunderstood me. We should accept both a list and a single float, but *store* it internally always as a list. You can use the following code (assuming `length` is known):

```python
try:
    rad_list = [float(r) for r in det_rad]
    if len(rad_list) != length:
        raise ValueError('Expected {} entries in radius sequence, got {}.'
                         ''.format(length, len(rad_list)))
except TypeError:
    # Triggered when we have a single float
    rad_list = [float(det_rad)] * length
```

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        for ii in range(len(det_radius)):
+            bool_list.append((odl.util.all_equal(det_radius[0],
+                                                 det_radius[ii])))
+        if False not in bool_list:
+            self._equal_radii = True
+        else:
+            if len(det_radius) != len(apart.grid.points()):
+                raise ValueError('detector radii has to be defined for all '
+                                 'angular positions, got {} angular positions.'
+                                 ''.format(len(apart.grid.points())))
```
kohr-h: Some clean-up needed. If you use the code I posted above, you have the list of radii. Then you can simply store
```python
self._all_radii_equal = all(r == rad_list[0] for r in rad_list)
```

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        self._fwhm = kwargs.pop('fwhm', None)
+        self._depth_dep = kwargs.pop('depth_dep', None)
+        self._n_pix = kwargs.pop('n_pix', None)
+        super().__init__(3, apart, detector)
+        self._motion_part = apart
+
+    @property
+    def det_radius(self):
+        """Detector circle radius of this geometry."""
+        return self._det_radius
```
kohr-h: This should then be named `det_radii`. Add also a property
```python
@property
def det_radii_all_equal(self):
    """`True` if all detector radii are equal, `False` otherwise."""
    return self._all_radii_equal

File: odl/tomo/geometry/spect_geometry.py (hunk 15 above)
Comment (kohr-h) [inline odl/tomo/geometry/spect_geometry.py]: I'm getting more and more convinced that it is actually the **Detector** which is special in this kind of SPECT geometry, and that it would make sense to create a `Flat2dCollimatedDetector` subclass of `Flat2dDetector`. That new guy would then allow to define the field of view of an individual detector pixel (I think it is safe to assume that it's the same for all pixels).

File: odl/tomo/geometry/geometry.py [hunk 19]
```diff
@@ -164,26 +164,6 @@ def det_refpoint(self, mpar):
             The reference point, an `ndim`-dimensional vector
         """
 
-    @abstractmethod
```
adler-j: Were did this go?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+            The reference point on a circle in the transaxial plane with
+            radius ``R`` and on at a axial position ``z``.
+            Transaxial point given by the rotation angle ``phi``.
+        """
+        if self._equal_radii:
+            origin_to_det = self.det_radius[0] * self._center_to_det
+            circle_component = self.rotation_matrix(angle).dot(origin_to_det)
+        else:
+            angles = self.motion_grid.points()
+            ind = int(np.where(angles == angle)[0])
```
adler-j: What if `angle not in angles`?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+            rotation to the detector reference point.
+            The zero vector is not allowed.
+            By default, a `perpendicular_vector` to ``axis`` is used.
+
+        """
+        AxisOrientedGeometry.__init__(self, axis)
+
+        bool_list = []
+        for ii in range(len(det_radius)):
+            bool_list.append((odl.util.all_equal(det_radius[0],
```
adler-j: double parentesis

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        center_to_det : `array-like`, shape ``(2,)``, optional
+            Initial state of the vector pointing from the center of
+            rotation to the detector reference point.
+            The zero vector is not allowed.
+            By default, a `perpendicular_vector` to ``axis`` is used.
+
+        """
+        AxisOrientedGeometry.__init__(self, axis)
+
+        bool_list = []
```
adler-j: Sure, its a list of bools, but what are these bools?

adler-j: This whole thing could be written (something like)

```python
self._equal_radii = all(rad == det_radius[0] for rad in det_radius)
```

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        for ii in range(len(det_radius)):
+            bool_list.append((odl.util.all_equal(det_radius[0],
+                                                 det_radius[ii])))
+        if False not in bool_list:
+            self._equal_radii = True
+        else:
+            if len(det_radius) != len(apart.grid.points()):
```
adler-j: `len(det_radius) != apart.size`

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        if np.linalg.norm(center_to_det) <= 1e-10:
+            raise ValueError('initial center to detector vector {} is too '
+                             'close to zero.'.format(center_to_det))
+        self._center_to_det = (np.array(center_to_det) /
+                               np.linalg.norm(center_to_det))
+
+        det_init_axis_0 = np.cross(self.axis, self._center_to_det)
+        det_init_axes = (det_init_axis_0, axis)
+        detector = Flat2dDetector(dpart, det_init_axes)
+        self._fwhm = kwargs.pop('fwhm', None)
```
adler-j: is fwhm used somewhere?

odl/tomo/geometry/spect_geometry.py
```diff
@@ -0,0 +1,129 @@
...
+        self._depth_dep = kwargs.pop('depth_dep', None)
```
adler-j: this is not documented in the interface

odl/tomo/geometry/spect_geometry.py
niinimaki: It would be used in forward operator related to the PSF. We had a discussion about this with @kohr-h  and @ozanoktem at some point and I included it in the geometry. 
But after a new discussion with @kohr-h, I will change this so that `fwhm` is rather included in the forward operator.

odl/tomo/geometry/spect_geometry.py
niinimaki: Yes because it is just a leftover from my previous version and was intended to be removed from this one :)

test/tomo/geometry/spect_geometry_test.py
```diff
@@ -0,0 +1,58 @@
...
+    dpart = odl.uniform_partition([-det_param, -det_param],
+                                  [det_param, det_param],
+                                  [det_nx_pix, det_ny_pix])
+
+    apart = odl.uniform_partition(0, 2 * np.pi, n_proj, nodes_on_bdry=True)
+    geom = ParallelHoleCollimatorGeometry(apart, dpart, det_radius)
+    assert isinstance(geom.detector, odl.tomo.Flat2dDetector)
+    assert all_equal(geom.det_radius, det_radius)
+
+    angles = geom.motion_grid.points()
```
adler-j: General note:

This call is made sufficiently often that we should add a shorthand to `AxisOrientedGeometry`

```python
@property
def angles(self):
    return geom.motion_grid.coord_vectors[0]
```

I'll do this.

adler-j: See #369

odl/tomo/geometry/geometry.py
kohr-h: Scrapped. In the most general case, you don't necessarily have a rotation matrix. For example in PET, the reference point would be stationary.

adler-j: Not quite sure i agree on removing it, by the same argument we could remove the `motion_params` as well, no?

kohr-h: In principle that's true. It would be possible to make the hierarchy a bit more fine-grained (perhaps by duck-typing) by differentiating between geometries where the detector moves relative to the object (CT, SPECT) and ones where it does not (PET, MRI, anything with an array of fixed detectors).

The case with the rotation matrix (i.e. a rotating detector in general) could encompass all cases where the detector movement can be parametrised in the in polar coordinates `x = r(phi) * d(phi)`, where `r(phi)` is the distance from a reference point (e.g. 0) and `d(phi)` is a direction vector. The circular acquisition curves would be special cases of that, but we also need more general cases e.g. in SPECT, where `r(phi)` is given as a list of numbers. Probably the detector movement is determined during acquisition so I don't know if the points will lie on any reasonably regular curve.

The axis-oriented case is similar, except that the polar coordinates only apply perpendicular to the axis, while a third function `g(phi)` governs the movement along the axis.

Anyway, this is probably better suited for a separate issue. The removal of the matrix was anyway motivated by the imminent PET geometry implementation, which took a turn towards SPECT in the end, so we can keep the matrix in this PR.

adler-j: > In principle that's true. It would be possible to make the hierarchy a bit more fine-grained (perhaps by duck-typing) by differentiating between geometries where the detector moves relative to the object (CT, SPECT) and ones where it does not (PET, MRI, anything with an array of fixed detectors).

Certainly possible to do and may be a good idea. Having `motion_param` and the things that come along, namely the idea of a detector separate from the rest of the geometry description, `rotation_matrix`, `det_ref_point`, etc is probably confusing in quite some cases.

> The case with the rotation matrix (i.e. a rotating detector in general) could encompass all cases where the detector movement can be parametrised in the in polar coordinates x = r(phi) * d(phi), where r(phi) is the distance from a reference point (e.g. 0) and d(phi) is a direction vector. The circular acquisition curves would be special cases of that, but we also need more general cases e.g. in SPECT, where r(phi) is given as a list of numbers. Probably the detector movement is determined during acquisition so I don't know if the points will lie on any reasonably regular curve.
> The axis-oriented case is similar, except that the polar coordinates only apply perpendicular to the axis, while a third function g(phi) governs the movement along the axis.

This can represent any geometry where a detector with fixed shape moves and rotates according to some parameter in an arbitrary interval product. The current approach assumes that any point on the detector can be given by:

```
point = det_ref_point(motion_param) + rotation_matrix(motion_param).dot(point_in_detector_reference_system(detector_param))
```

This is all kindof how we define the distinction between `motion_params` and `detector_params`. So, imo, as long as we keep motion_params, we should keep the rotation_matrix (and the det_ref_point, since these come in a pair).

> Anyway, this is probably better suited for a separate issue. The removal of the matrix was anyway motivated by the imminent PET geometry implementation, which took a turn towards SPECT in the end, so we can keep the matrix in this PR.

Agree, needs more thought.

niinimaki:
Much simplified version of the SPECT geometry. Since we don't have a back-end for variable radius, I removed the support, we can add it later on if we need it. The geometry now assumes circular acquisition and inherits from `Parallel3dAxisGeometry` due to large similarity.

odl/tomo/geometry/parallel.py
```diff
@@ -424,6 +423,12 @@ def __repr__(self):
     # Fix for bug in ABC thinking this is abstract
     rotation_matrix = AxisOrientedGeometry.rotation_matrix
 
+    # TODO: find a more generic place to put this property
```
adler-j: This is already in master since fa5a6f3

odl/tomo/geometry/parallel.py
```diff
@@ -57,8 +57,7 @@ def __init__(self, ndim, apart, detector, det_init_pos):
         detector : `Detector`
             The detector to use in this geometry
         det_init_pos : `array-like`
-            Initial position of the detector reference point. The zero
```
adler-j: Why remove this, is it not required?

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""SPECT geometry."""
```
adler-j: Write out spect:

"""Single-photon emission computed tomography (SPECT) geometry."""

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+from odl.tomo.geometry.parallel import Parallel3dAxisGeometry
+
+__all__ = ('ParallelHoleCollimatorGeometry', )
+
+
+class ParallelHoleCollimatorGeometry(Parallel3dAxisGeometry):
+
+    """Geometry for SPECT Parallel hole collimator."""
+
+    def __init__(self, apart, dpart, axis=[0, 0, 1], det_rad=0.0,
```
adler-j: Default 0 doesnt make much sense

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+            By default, the normalized cross product of ``axis`` and
+            ``det_init_pos`` is used as first axis and ``axis`` as second.
+        """
+        # Use det_init_pos to initialize the detector position, then rescale
+        orig_to_det_init = kwargs.pop('orig_to_det_init', None)
+        if orig_to_det_init is not None:
+            kwargs['det_init_pos'] = orig_to_det_init
+        super().__init__(apart, dpart, axis, **kwargs)
+        self._det_radius = float(det_rad)
+        if self.det_radius < 0:
```
adler-j: See above, why is 0 OK?

odl/tomo/geometry/parallel.py
kohr-h: Weird. We rebased on master and didn't get the property. @niinimaki can you check if it's duplicate? If yes, do the interactive rebase and drop the commit.

odl/tomo/geometry/spect.py
kohr-h: Better default? 1 (meter)? Another option would be to require it as input parameter and force it to be positive, of course.

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+            raise ValueError('expected a nonnegative radius, got {}.'
+                             ''.format(det_rad))
+
+        # Make sure that the initial position lies on the specified circle
+        init_pos_norm = np.linalg.norm(self._det_init_pos)
+        if init_pos_norm > 1e-10:
```
adler-j: This text is not needed since we would have thrown an error above if this was not the case due to the check here:

https://github.com/odlgroup/odl/blob/a40e5e5df88f7051c4b6b37505fab2a9bde62473/odl/tomo/geometry/parallel.py#L386

odl/tomo/geometry/parallel.py
kohr-h: It's not forbidden to put the detector into the origin. May not make too much sense physically but I see no reason to forbid it.

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+            self._det_init_pos *= self.det_radius / init_pos_norm
```
adler-j: We should in general not be modifying private variables of a parent class, this can cause all types of issues later on (and wierd dependencies), can we not do this by modifying the input to `super`?

odl/tomo/geometry/spect.py
kohr-h: That code only raises an error if the initial axes were not given, otherwise it will work fine. However, we should raise if the radius is positive and the initial position is zero, since those two parameters cannot be reconciled.

odl/tomo/geometry/spect.py
kohr-h: We can, but then we basically duplicate the checks from the parent class. Perhaps that's better though.

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+        Parameters
+        ----------
+        apart : 1-dim. `RectPartition`
+            Partition of the angle interval
+        dpart : 2-dim. `RectPartition`
+            Partition of the detector parameter rectangle
+        axis : `array-like`, shape ``(3,)``, optional
+            Fixed rotation axis.
+        det_rad : nonnegative `float`, optional
+            Radius of the circular acquisition curve
```
adler-j: Looks like copy paste, is this the standard naming in SPECT?

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+        orig_to_det_init : `array-like`, shape ``(3,)``, optional
+            Vector pointing towards the detector reference point in
+            the initial position. The zero vector is only allowed if
+            ``det_init_axes`` is explicitly given.
+            By default, a `perpendicular_vector` to ``axis`` is used.
```
adler-j: Default: ...

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+        det_init_axes : 2-tuple of `array-like` (shape ``(3,)``), optional
+            Initial axes defining the detector orientation.
+            By default, the normalized cross product of ``axis`` and
```
adler-j: Default: ...

Is our standard convention

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+        if init_pos_norm > 1e-10:
+            self._det_init_pos *= self.det_radius / init_pos_norm
+
+    @property
+    def det_radius(self):
+        """Radius of the detector curve parametrization."""
+        return self._det_radius
+
+    # Fix for bug in ABC thinking this is abstract
+    rotation_matrix = AxisOrientedGeometry.rotation_matrix
```
adler-j: rotation_matrix = Parallel3dAxisGeometry.rotation_matrix

test/tomo/geometry/spect_geometry_test.py
```diff
@@ -0,0 +1,59 @@
...
+    det_ny_pix = 64
+    det_nx_mm = 4
+    det_radius = 200
+    n_proj = 180
+    det_param = det_nx_mm * det_nx_pix
+    dpart = odl.uniform_partition([-det_param, -det_param],
+                                  [det_param, det_param],
+                                  [det_nx_pix, det_ny_pix])
+
+    apart = odl.uniform_partition(0, 2 * np.pi, n_proj, nodes_on_bdry=True)
```
adler-j: Why `nodes_on_bdry=True`?

odl/tomo/geometry/spect.py
kohr-h: That didn't work on first attempt, but I don't know the exact reason why, to be honest. Could be worth another try.

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+                             ''.format(det_rad))
+
+        # Make sure that the initial position lies on the specified circle
+        init_pos_norm = np.linalg.norm(self._det_init_pos)
+        if init_pos_norm > 1e-10:
+            self._det_init_pos *= self.det_radius / init_pos_norm
+
+    @property
+    def det_radius(self):
+        """Radius of the detector curve parametrization."""
```
adler-j: Has to be a more simple way to state this

kohr-h: "Radius of the circular acquisition curve"?

odl/tomo/geometry/spect.py
adler-j: If no sensible default exists we should simply not take it as a parameter. Also for consistiency with the cone beam guys, this should come before `axis`.

odl/tomo/geometry/spect.py
adler-j: Second thought, we are chaging the meaning of this guy here. In the parent class, this is a vector to the detector reference point. Here you change it to mean a unit vector _towards_ the detector reference point. This will cause bugs, try calling `det_refpoint` and you'll see.

adler-j: The proper way to do this would be to let `det_init_pos` keep its meaning, but perhaps scale this one by `det_rad` so you get the correct length. This could be done prior to calling super()

odl/tomo/geometry/spect.py
adler-j: "Distance from the axis to the detector", or something similar?

odl/tomo/geometry/spect.py
adler-j: ABC error or actual NameError?

odl/tomo/geometry/parallel.py
adler-j: Valid reasoning, as long as it doesnt cause us several divides by zero's later on.

odl/tomo/geometry/spect.py
kohr-h: > Second thought, we are chaging the meaning of this guy here. In the parent class, this is a vector to the detector reference point. Here you change it to mean a unit vector towards the detector reference point.

No, the meaning is the same. It is only ensured that radius and `orig_to_det_init` give consistent information. Probably the doc should say so, but the vector is interpreted (and rescaled) as a unit vector, multiplied by the radius, so the initial position is correct.



> The proper way to do this would be to let det_init_pos keep its meaning, but perhaps scale this one by det_rad so you get the correct length. This could be done prior to calling super()

In view of the confusion, I think it would be cleaner to do the rescaling first and feed that into the parent class in `super`.

odl/tomo/geometry/spect.py
kohr-h: ABC ("cannot instantiate ...")

odl/tomo/geometry/spect.py
adler-j: Reading through everything again realized and I'm indeed confused. Still, feeding this into super is the "proper" way to do it.

In the current way, the else statement should check that `det_radius` is actually given as 0.

odl/tomo/geometry/spect.py
adler-j: Damnit. This has to be a bug in ABC, we should report it. This way of doing it is quite "wrong" since we assume a certain inheritance chain.

odl/tomo/geometry/spect.py
niinimaki: "Radius of the detector orbit " seems to be the way to express this.

test/tomo/geometry/spect_geometry_test.py
niinimaki: Because I wanted that the first angle is exactly zero to be consistent with the SPECT data header.

adler-j: Well, since this test does not use the SPECT data header, we may was well go for the more simple approach of skipping `nodes_on_bdry`. For an actual example (in the examples folder) where we use this with spect, we would indeed need to do this.

niinimaki: Sure.

niinimaki:
According to the review

- improved the docs

- removed the default value from `det_radius` since no sensible default exists

- changed the order of `axis` and `det_rad`

- did the rescaling of `orig_to_det_init` prior calling `super`

- Removed `nodes_on_bdry` and related tests from the `spect_geometry_test`

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+            Fixed rotation axis.
+        orig_to_det_init : `array-like`, shape ``(3,)``, optional
+            Vector pointing towards the detector reference point in
+            the initial position. The zero vector is only allowed if
+            ``det_init_axes`` is explicitly given.
+            Default: a `perpendicular_vector` to ``axis``.
+        det_init_axes : 2-tuple of `array-like` (shape ``(3,)``), optional
+            Initial axes defining the detector orientation.
+            Default: the normalized cross product of ``axis`` and
+            ``det_init_pos`` is used as first axis and ``axis`` as second.
```
kohr-h: `det_init_pos` -> `orig_to_det_init`

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+            Partition of the detector parameter rectangle
+        det_rad : positive `float`
+            Radius of the circular detector orbit.
+        axis : `array-like`, shape ``(3,)``, optional
+            Fixed rotation axis.
+        orig_to_det_init : `array-like`, shape ``(3,)``, optional
+            Vector pointing towards the detector reference point in
+            the initial position. The zero vector is only allowed if
+            ``det_init_axes`` is explicitly given.
+            Default: a `perpendicular_vector` to ``axis``.
```
kohr-h: When I think about it again, we should disallow zero here. It's optional anyway, and zero initial position can be reached by choosing `det_rad = 0`. Simply delete the corresponding sentence here and raise an error (see below).

odl/tomo/geometry/spect.py
```diff
@@ -0,0 +1,82 @@
...
+        if self.det_radius <= 0:
+            raise ValueError('expected a positive radius, got {}.'
+                             ''.format(det_rad))
+
+        orig_to_det_init = kwargs.pop('orig_to_det_init',
+                                      perpendicular_vector(axis))
+
+        init_pos_norm = np.linalg.norm(orig_to_det_init)
+        if init_pos_norm > 1e-10:
+            orig_to_det_init *= self.det_radius / init_pos_norm
```
kohr-h:
```python
else:
    raise ValueError('...')
```
Take the same error message wording as in the parent class (w/o reference to the axes thing).

odl/tomo/geometry/spect.py
kohr-h: :+1: for orbit - makes me feel like a rocket scientist :-)

niinimaki:
Disallowed `orig_to_det_init` of being zero vector.

odl/tomo/geometry/spect.py
```diff
@@ -70,6 +69,10 @@ def __init__(self, apart, dpart, det_rad, axis=[0, 0, 1],
         init_pos_norm = np.linalg.norm(orig_to_det_init)
         if init_pos_norm > 1e-10:
             orig_to_det_init *= self.det_radius / init_pos_norm
+        else:
+            raise ValueError('initial detector position {} is close to '
+                             'zero. This is only allowed for explicit '
+                             'det_init_axes.'.format(orig_to_det_init))
```
kohr-h: The `det_init_axes` part slipped through, please remove. The first sentence can simply be `'initial detector position {} is too close to zero.'`

kohr-h:
For my part, except from one small thing in the error message, I'm ok with this.

niinimaki:
Corrected the "copy-paste" error from the error message. 
Ready to be merged?

odl/tomo/geometry/spect.py
kohr-h: Don't forget this.

niinimaki: I think I misunderstood you notation, so you mean than change `orig_to_det_init` to read `det_init_pos`?

kohr-h: No, I think you understood me correctly. I commented under the exact line I meant, not 3 lines up.

adler-j:
Some comments last, except for that the commit messages should be cleaned up. Many seem to have "state" e.g. refer to knowledge in this thread. In general commit messages should be readable on their own with no extra knowledge (except perhaps of the codebase). Example:

MAINT: improv<b>ed</b> error message

could be

MAINT: improv<b>e</b> error message in spect geometry

odl/tomo/geometry/spect.py
niinimaki: Thought I had done it already but apparently not, but now it is fixed.

niinimaki:
A bit of a review.
About the PSF: in SPECT PSF is used for the correction of the collimator-detector response (CDR). CDR includes both the geometric properties of the collimator and the properties of the crystal
and detector itself. For a camera with a parallel-hole collimator, the PSF can be represented by a Gaussian which width is given as the Full-Width at Half-Maximum (FWHM).  Like attenuation correction the PSF correction is rather a part of the forward operator and it will be included in the forthcoming `spect_trafo`.

About several radii: if scanner stores several radii, a radius is stored per each angular position, the curve will then be a “contoured orbit”, so not necessarily a reasonably regular curve. But circular acquisition is also used in clinic. After discussing this with @kohr-h,  and since we don't at the moment have a back-end that would support  variable radius, the support for that is removed. If needed, we can add it later on.

Making the spect geometry to be a subclass of `Parallel3dAxisGeometry` simplified and cleaned up the code a lot.

With respect to the clinical data and our present back-end for spect forward operator, imo this representation of the spect geometry is sufficient. 

I suppose that what remains now is, at least the cleaning up of commit messages as @adler-j  suggested.

niinimaki:
Can this be merged?

PR #309, opened 2016-03-17
ENH: add option to show by coordinate in DiscreteLp

### LINKED ISSUE #260: Alternative to indices in show
We strive quite hard to allow users to think continuous, the `indices=...` concept kindoff breaks this, and is often hard to think about (at least I have to do some mental gymnastics each time I call it for nontrivial cases).

How about adding some "continuous" syntax, that translates to indices. Some examples of how this could look is:

```python
# Make space
X = uniform_discr([-1, -1], [1, 1], 157)
x = X.element(...)

# Current
x.show(indices=np.s_[:, 78])  # slice through the middle

# Some alternatives
x.show(values=[None, 0.0])
x.show(values=np.s_[:, 0.0])
x.show(y=0.0)
```

adler-j:
This closes #260

### DIFF
--- a/examples/visualization/show_2d.py
+++ b/examples/visualization/show_2d.py
@@ -28,5 +28,11 @@ import odl
 spc = odl.uniform_discr([0, 0], [1, 1], [100, 100])
 vec = odl.util.shepp_logan(spc, modified=True)
 
-# Can also force "instant" plotting
+# Show all data
 vec.show(show=True)
+
+# We can show subsets by index:
+vec.show(indices=[slice(None), 50])
+
+# Or we can show by coordinate
+vec.show(coords=[None, 0.5])
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -628,8 +628,8 @@ class DiscreteLpVector(DiscretizationVector):
         """
         return DiscreteLpUFuncs(self)
 
-    def show(self, title=None, method='', indices=None, show=False, fig=None,
-             **kwargs):
+    def show(self, title=None, method='', coords=None, indices=None,
+             show=False, fig=None, **kwargs):
         """Display the function graphically.
 
         Parameters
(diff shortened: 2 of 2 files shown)


examples/visualization/show_2d.py
```diff
@@ -28,5 +28,11 @@
...
 
-# Can also force "instant" plotting
+# Show all data
 vec.show(show=True)
+
+# We can show subsets by index:
+vec.show(indices=[slice(None), 50])
+
+# Or we can show by coordinate
+vec.show(coords=[None, 0.5])
```
adler-j: I prefer this syntax so much to the indices guy, we should perhaps try to modify all examples where it occurs aswell. This is much more "odly".

odl/discr/lp_discr.py
```diff
@@ -688,6 +697,26 @@ def show(self, title=None, method='', indices=None, show=False, fig=None,
...
+            interv = self.space.uspace.domain
+            shape = self.shape
+            indices = []
+            for begin, end, n, coord in zip(interv.begin, interv.end,
+                                            shape, coords):
+                if coord is None:
+                    indices += [slice(None)]
+                else:
+                    if not begin <= coord <= end:
+                        raise ValueError('Coord {} not in range {}'
```
kohr-h: "coord"

odl/discr/lp_discr.py
```diff
@@ -688,6 +697,26 @@ def show(self, title=None, method='', indices=None, show=False, fig=None,
...
+                                         ''.format(coord, [begin, end]))
+
+                    # Todo: handle end = begin
```
kohr-h: Throw an error? I guess using `None` is the only way not to run into rounding errors.

adler-j:
Making a new for merge into master

adler-j:
Obviously closed the wrong one...

adler-j:
Fixed comments, merge?

examples/visualization/show_2d.py
kohr-h: odlarious maybe?

kohr-h:
Looking good to me

Issue #220, opened 2016-01-15
Add split bregman solver


adler-j:
We seem to need a split bregman solver, a proposed syntax would be:

```python
def SplitBregmanReconstruct(A, Phi, x, rhs, la, mu, inner_solver=None
                            niter=1, niter_inner=1, partial=None):
```

We could also need some special solver for the case of isotropic tv, which requires some modifications of the general algorithm.

adler-j:
Pull #311 was closes and we decided to implement ADMM if we do this later.

PR #311, opened 2016-03-18
Issue 220  split bregman

adler-j:
This adds Split Bregman according to #220.

The format is still kindoff early and can (should) be discussed. This is also an option for us to discuss the use of objective functionals with gradients, this function currently takes the gradient of the target functional (and never needs to evaluate functional values).

I think we should try to get this solver into ODL rather soon since it is quite useful in many circumstances, for one, it's interface is about 3 times smaller than that of chambolle pock:

Example with bregman:

```python
grad = odl.Gradient(A.domain, method='forward')
hgrad = A.adjoint * odl.ResidualOperator(A, rhs)

mu = 6000.0
la = 5.0

x = A.domain.zero()
odl.solvers.split_bregman_solver(mu * hgrad, grad, x, la,  niter=500)
```

VS CP:

```python
grad = odl.Gradient(A.domain, method='forward')

op = odl.ProductSpaceOperator([[A],
                               [gradient]])

op_norm = 1.1 * odl.operator.oputils.power_method_opnorm(op, 100)

proximal_primal = odl.solvers.proximal_zero(op.domain)

prox_convconj_l2 = odl.solvers.proximal_convexconjugate_l2(discr_space, g=data)
prox_convconj_l1 = odl.solvers.proximal_convexconjugate_l1(
    gradient.range, lam=0.01)
proximal_dual = odl.solvers.combine_proximals(
    [prox_convconj_l2, prox_convconj_l1])

tau = 1 / op_norm
sigma = 1 / op_norm

x = op.domain.one()
odl.solvers.chambolle_pock_solver(
    op, x, tau=tau, sigma=sigma, proximal_primal=proximal_primal,
    proximal_dual=proximal_dual, niter=500)
```

obviously the branch seems to need some work on git.


odl/util/numerics.py
```diff
@@ -298,6 +298,38 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+    Examples
+    --------
+    >>> import odl
+    >>> X = odl.ProductSpace(odl.Rn(3), 2)
+    >>> x = X.element([[1, 2, 3], [4, 5, 6]])
+    >>> pspace_squared_sum(x)
+    Rn(3).element([17.0, 29.0, 45.0])
+    """
+    if out is None:
+        out = x.space[0].element()
```
adler-j: Should be zero

odl/solvers/advanced/split_bregman.py
```diff
@@ -0,0 +1,154 @@
...
+            inner_op = H_grad + lam * (Phi.adjoint * Phi)
+            inner_rhs = lam * Phi.adjoint(d - b)
+            inner_solver(inner_op, x, inner_rhs)
+
+            # Solve for d using soft threshholding
+            s = Phi(x) + b
+            if isotropic:
+                sn = pspace_squared_sum(s)
+                sn.ufunc.add(1e-7, out=sn)  # avoid 0/0 issues
+                print(sn.ufunc.absolute().ufunc.max(), 1.0 / lam)
```
adler-j: obviously leftover

odl/solvers/advanced/split_bregman.py
```diff
@@ -0,0 +1,154 @@
...
+            # Solve tomography part using the given solver
+            inner_op = H_grad + lam * (Phi.adjoint * Phi)
+            inner_rhs = lam * Phi.adjoint(d - b)
+            inner_solver(inner_op, x, inner_rhs)
+
+            # Solve for d using soft threshholding
+            s = Phi(x) + b
+            if isotropic:
+                sn = pspace_squared_sum(s)
+                sn.ufunc.add(1e-7, out=sn)  # avoid 0/0 issues
```
adler-j: Not quite sure how to handle this

odl/solvers/advanced/split_bregman.py
```diff
@@ -0,0 +1,154 @@
...
+    lam = float(lam)
+
+    # If no solver is given, create a new.
+    if inner_solver is None:
+        def inner_solver(op, x, rhs):
+            conjugate_gradient(op, x, rhs, 1)
+    else:
+        assert callable(inner_solver)
+
+    b = Phi.range.zero()
```
adler-j: Warm start option?

adler-j:
Frankly. Will we want this solver as a "simple alternative", or should we go straight for a proper ADMM implementation?

kohr-h:
I don't know about this variant, and as usual in optimization, method X is usually just a special case of method Y where one chooses this and this. But I would also be surprised if "Split Bregman" was a strict subset of ADMM in the above sense, as your comment implies @adler-j . Unless that is really the case, ADMM would not make this branch (or the implicit issue of implementing a split Bregman method) obsolete.

chongchenmath:
As I know, the `Split Bregman` is a kind of `ADMM`.

kohr-h:
Okay, after reading a bit, I think the following is safe to say:

- Split Bregman only solves the TV problem (ROF model) with linear operator and L^2 distance in the data fitting term
- It is a special case of
  * forward-backward splitting
  * basis pursuit for parameter to infinity
  * augmented Lagrangian for some other problem
  * ADMM (?)

So then I guess it's okay to drop it and stick to more flexible and high-level methods.

PR #320, opened 2016-03-27
ENH: implement product space weighting properly

kohr-h:
See  919b527 for details on changes.

The major change is the factored out weighting into a new module which is re-usable by `FnBase` and `ProductSpace`. The space-specific classes simply implement `inner`, `norm` and `dist`, while much of the rest is used unchanged. Lots of saved code duplication.

~~In addition, I cleaned up some parts, mostly the import section and the doctest invocation which used to give tons of pylint errors.~~

I also moved `pspace.py` to the `space` submodule since I started to get circular deps with stuff from space (due to `FnWeightingBase` before moving that to a module on its own, but putting the latter into `set` would have been a bad fit). I think product spaces are a better fit there anyway. We could even consider moving `space.py`, but maybe that's unnecessary.

EDIT: doctest and cleanup stuff done in a separate PR #321


odl/discr/grid.py
```diff
@@ -1343,5 +1341,4 @@ def uniform_sampling(begin, end, num_nodes, nodes_on_bdry=True):
 
 
 if __name__ == '__main__':
-    from doctest import testmod, NORMALIZE_WHITESPACE
-    testmod(optionflags=NORMALIZE_WHITESPACE)
+    run_doctests()
```
adler-j: would really have prefered this on another branch, can we cherry pick it?

adler-j:
This is quite hard to read due to the very extensive changes, we really should condense it to what it is (aka cherry pick 266e352).

Some design considerations: How well does these weightings generalize to non cpu spaces? Rn^2^3 type spaces?

odl/__init__.py
```diff
@@ -25,7 +25,7 @@
 
 from __future__ import absolute_import
 
-__version__ = '0.9b1'
+__version__ = '0.2.2'
```
adler-j: Why is this here?

odl/discr/grid.py
kohr-h: Okay, okay

odl/__init__.py
kohr-h: Cherry pick.

kohr-h:
We first merge #321, then I'll rebase, and the pull will be more readable.

kohr-h:
> Some design considerations: How well does these weightings generalize to non cpu spaces? (Rn^2)^3 type spaces?

At least they handle the stuff we already had, including the CUDA spaces. Basically this is just a big refactor that moves all the boilerplate code out of the way and leaves the concrete weighting classes with just the job to implement the 3 essential functions. I'll double-check if there is anything in the setup that is "anti-CUDA".

kohr-h:
Pulled out the doctest stuff.

adler-j:
Great, will do short review, looks good overall.

odl/space/weighting.py
```diff
@@ -0,0 +1,1047 @@
...
+                             ''.format(vector, self._vector.ndim))
+
+    @property
+    def vector(self):
+        """Weighting vector of this inner product."""
+        return self._vector
+
+    def vector_is_valid(self):
+        """Test if the vector is a valid weight, i.e. positive."""
+        return np.all(np.greater(self.vector, 0))
```
adler-j: why not 

```python
np.all(self.vector > 0)
```

and where is this used externally?

odl/space/weighting.py
```diff
@@ -0,0 +1,1047 @@
...
+        ----------
+        impl : `str`
+            Specifier for the implementation backend
+        exponent : positive `float`, optional
+            Exponent of the norm. For values other than 2.0, the inner
+            product is not defined.
+        dist_using_inner : `bool`, optional
+            Calculate `dist` using the following formula::
+
+                ||x - y||^2 = ||x||^2 + ||y||^2 - 2 * Re <x, y>
```
adler-j: use \`\`?

odl/space/weighting.py
```diff
@@ -0,0 +1,1047 @@
...
+        ----------
+        x1 : `LinearSpaceVector`
+            Vector whose norm is calculated
+
+        Returns
+        -------
+        norm : `float`
+            The norm of the vector
+        """
+        return float(sqrt(self.inner(x, x).real))
```
adler-j: why not simply call np.sqrt?

odl/space/weighting.py
```diff
@@ -0,0 +1,1047 @@
...
+            dist_squared = (self.norm(x1) ** 2 + self.norm(x2) ** 2 -
+                            2 * self.inner(x1, x2).real)
+            if dist_squared < 0:  # Compensate for numerical error
+                dist_squared = 0.0
+            return float(sqrt(dist_squared))
+        else:
+            return self.norm(x1 - x2)
+
+
+class MatrixWeighting(WeightingBase):
```
adler-j: It took me quite a while to figure out that this is not a implementation of `WeightingBase` but an abstract subclass, perhaps rename to `MatrixWeightingBase`?

odl/space/weighting.py
kohr-h: Sorry?

odl/space/weighting.py
kohr-h: Yeah, why not?

odl/space/weighting.py
kohr-h: True

odl/space/weighting.py
kohr-h: Not remembering your own suggestions? :-)

This version works with `CudaFnVector`.

The method is not used externally at the moment, but I see these validation functions as a debugging interface. Maybe for vectors, it's simple, but for matrices, it gets more involved.

adler-j: Sure :)

Anyway, perhaps simply `is_valid` so you dont have to consider if it is a matrix or vector then?

odl/space/weighting.py
adler-j: Make this display as code, but perhaps the :: does that?

kohr-h: That's the double colon, right. I guess it's better not to make it look like code.

odl/space/weighting.py
kohr-h: Sure

adler-j:
Should be good!

kohr-h:
Okay, there were some failing tests due to too strict check of the result (check for 7 places where computation was only in single precision). Fixed now. I'll merge after that.

PR #321, opened 2016-03-28
MAINT: employ run_doctests, closes #256

### LINKED ISSUE #256: Utility to run doctests
Besides pylint warnings, the lines in each file responsible for doctest add some redundancy. This should be reduced by making a small `run_doctests()` utility.

kohr-h:


kohr-h:
Checks pass.

odl/discr/discr_mappings.py
```diff
@@ -23,20 +23,18 @@
...
-# External imports
 from itertools import product
 import numpy as np
 
-# ODL imports
 from odl.operator.operator import Operator
 from odl.discr.partition import RectPartition
 from odl.space.base_ntuples import NtuplesBase, FnBase
 from odl.space.fspace import FunctionSet, FunctionSpace
+from odl.util.testutils import run_doctests
```
adler-j: How about put this inside the "if main"?

adler-j:
As mentioned above I feel like this import should go into the if main thing, else I'm fine.

odl/discr/discr_mappings.py
kohr-h: That's where all the pylint complaints come from :-)

Now we don't have to obey a linter, and I agree that the additional import is somewhat clumsy. I'll move the import.

adler-j: Perhaps add an ignore to this line then for pylint?

kohr-h: Yeah, that one place is acceptable.

PR #324, opened 2016-03-29
ENH: add as_scipy_operator, closes #323

### LINKED ISSUE #323: Interfacing with scipy's Linear algebra
It seems scipy has a  [LinearOperator](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.sparse.linalg.LinearOperator.html) abstraction that allows access to their rather extensive [solvers](http://docs.scipy.org/doc/scipy/reference/sparse.linalg.html).

We could easily add support for this by adding a helper function

```python
def as_flat_array(arr):
    if hasattr(arr, 'order'):
        return arr.asarray().ravel(arr.order)
    else:
        return arr.asarray()

def as_scipy_operator(op):
    assert op.is_linear
    
    shape = (op.range.size, op.domain.size)
    
    def matvec(v):
        return as_flat_array(op(v))
        
    def rmatvec(v):
        return as_flat_array(op.adjoint(v))
        
    dtype = op.domain.dtype
    assert op.range.dtype == dtype
    
    return scipy.sparse.linalg.LinearOperator(shape=shape,
                                              matvec=matvec,
                                              rmatvec=rmatvec,
                                              dtype=dtype)
```

adler-j:
Very simple code that adds interface to scipys linear solvers, also has a simple example.

### DIFF
--- a/examples/solvers/scipy_solvers.py
+++ b/examples/solvers/scipy_solvers.py
@@ -1,0 +1,63 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Example of interfacing with scipy solvers
+
+This example solves the problem
+
+    -Laplacian(x) = b
+
+Where b is a dirac delta impulse
+"""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import scipy.sparse.linalg as sl
+import odl
+
+# Create discrete space, a square from [-1, 1] x [-1, 1] with (11 x 11) points
+space = odl.uniform_discr([-1, -1], [1, 1], [11, 11])
+
+# Create odl operator for negative laplacian
+laplacian = -odl.Laplacian(space)
+
+# Create right hand side, a gaussian around the point (0, 0)
+rhs = space.element(lambda x: np.exp(-(x[0]**2 + x[1]**2)/0.1**2))
+
+# Convert laplacian to scipy operator
+scipy_laplacian = odl.operator.oputils.as_scipy_operator(laplacian)
+
+# Convert to array and flatten
+rhs_arr = rhs.asarray().ravel(space.order)
+
+# Solve using scipy
+result, info = sl.cgs(scipy_laplacian, rhs_arr)
+
+# Other options include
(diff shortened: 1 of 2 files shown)


examples/solvers/scipy_solvers.py
```diff
@@ -0,0 +1,63 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Example of interfacing with scipy solvers
+
+This example solves the problem
+
+    -Laplacian(x) = b
```
kohr-h: Backticks?

examples/solvers/scipy_solvers.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Example of interfacing with scipy solvers
```
kohr-h: No space in the beginning, dot at the end.

examples/solvers/scipy_solvers.py
```diff
@@ -0,0 +1,63 @@
...
+
+This example solves the problem
+
+    -Laplacian(x) = b
+
+Where b is a dirac delta impulse
```
kohr-h: small "w", dot at the end

examples/solvers/scipy_solvers.py
```diff
@@ -0,0 +1,63 @@
...
+import odl
+
+# Create discrete space, a square from [-1, 1] x [-1, 1] with (11 x 11) points
+space = odl.uniform_discr([-1, -1], [1, 1], [11, 11])
+
+# Create odl operator for negative laplacian
+laplacian = -odl.Laplacian(space)
+
+# Create right hand side, a gaussian around the point (0, 0)
+rhs = space.element(lambda x: np.exp(-(x[0]**2 + x[1]**2)/0.1**2))
```
kohr-h: Can you PEP-8ify this?

examples/solvers/scipy_solvers.py
kohr-h: Not in sync with the code, as far as I see it.

odl/operator/oputils.py
```diff
@@ -178,6 +180,69 @@ def power_method_opnorm(op, niter, xstart=None):
...
+    This is intended to be used with the scipy sparse linear solvers.
+
+    Parameters
+    ----------
+    op : `Operator`
+        A linear operator that should be wrapped
+
+    Returns
+    -------
+    `scipy.sparse.linalg.LinearOperator` : linear_op
```
kohr-h: If this link works, you could use single backticks in the one-line summary, too.

kohr-h:
Cool stuff. I guess we don't have to bother with the simple iterative methods any longer. We should definitely have a documentation site on this, otherwise it's hard to find for users.

adler-j:
> Cool stuff. I guess we don't have to bother with the simple iterative methods any longer.

Slightly disagree here since these work very badly with cuda and distributed systems.

> We should definitely have a documentation site on this, otherwise it's hard to find for users.

We should have a whole page on "numpy / scipy" interactions on site, it should cover this and more.

kohr-h:
> Slightly disagree here since these work very badly with cuda and distributed systems.

Yes, sure. And product spaces won't work easily either.

kohr-h:
Looking good

PR #333, opened 2016-04-01
DOC: moved installation doc from README to doc, freshened up.

adler-j:
* Removed lots of rather complicated documentation from the README, instead point users to the documentation
* Improved installation doc


README.md
```diff
@@ -30,125 +29,15 @@ Features
...
-testing. See the next section.
-
-Running unit tests
-------------------
-Unit tests in ODL are based on [pytest](https://pypi.python.org/pypi/pytest), and coverage reports are created by the [coverage](https://pypi.python.org/pypi/coverage/) module. These packages are installed if you run
-
-    pip install --user -e .[testing]
-
-Now you can check that everything was installed properly by running
+You can check that everything was installed properly by running
```
kohr-h: I don't think we should mention this without telling that extra dependencies are required. People will run this, see an error and think WTF. This is anyway more for developers - if we instead tell them to run `python -c "import odl"`, that would be easier to understand.

doc/source/guide/introduction/installing.rst
```diff
@@ -2,36 +2,124 @@
...
+Installing odl is intended to be straightforward, and this guide is meant for new users. For a minimal installation you should do the following steps:
 
 1. `Install Python`_
-2. `Install Git`_
-3. `Install ODL`_
-4. `Run tests`_
+2. `Install ODL`_
+3. `Run tests`_
 
 If you have done any step before, you can ofc skip it.
```
kohr-h: How did "ofc" slip in? -> "you can skip it, of course."

doc/source/guide/introduction/installing.rst
```diff
@@ -2,36 +2,124 @@
...
 
 Install Python
 ==============
-To begin with, you need a python distribution. If you are an experienced user, you can use any distribution you'd like.
+To begin with, you need a Python distribution. Any modern Python distribution should do for the core library, but some extensions require CPython.
```
kohr-h: "CPython" could confuse users. We can write "[...] CPython, the ubiquitous Python implementation shipped by most OS vendors and scientific Python distributors." Something like that.

doc/source/guide/introduction/installing.rst
```diff
@@ -2,36 +2,124 @@
...
 
 Anaconda
 --------
-If you are a python novice using Windows, we recommend that you install a full package such as Anaconda. To install Anaconda
+If you are a python novice using Windows, we recommend that you install a full package such as Anaconda. To install Anaconda, download Anaconda from `Anaconda's webpage <https://www.continuum.io/downloads>`_. Once installed, run in a console
```
kohr-h: You capitalized "Python" above (good). Do so here, too. A few too many "Anaconda" occurrences here.

doc/source/guide/introduction/installing.rst
```diff
@@ -2,36 +2,124 @@
...
-Install Git
+to make sure you have the latest versions of all packages
+
+Python 2 vs 3
+-------------
+ODL fully supports both Python 2 and Python 3, so if you already have a python distribution (as is the case on most linux distributions) it may be a good idea to keep using it. If you are making a new installation we recommend using Python 3, since Python 2 support will be discontinued in 2020.
+
+Development enviroment
+----------------------
+Since ODL is object oriented, using a Integrated Development Environment (IDE) is recommended, but not required. The most popular ones are `Spyder
```
kohr-h: "an Integrated [...]"

README.md
adler-j: I dunno, if they are that new to Python then they will have to click the install documentation anyway. For me this is the steps you should do if you know what they mean, else go to the in depth doc. Omitting it here could cause advanced users (who will not feel like they need to read the install doc) to miss the tests.

This is how many do it. Numpy has their own test discovery, but do basically the same:

> After installation, tests can be run (if ``nose`` is installed) with:
> `python -c 'import numpy; numpy.test()'`

Theano [does the same](http://deeplearning.net/software/theano/install.html#basic-user-install-instructions)

> The easiest way to obtain the released version of Theano is from PyPI using pip (a replacement for easy_install provided by setuptools/distribute) by typing
> `pip install Theano`
> This should work under Python 2 or Python 3. To test, run
> `nosetests theano`

Perhaps we should add a note about "if you have `pytest` installed" like numpy does, to help if users get errors.

doc/source/guide/introduction/installing.rst
```diff
@@ -2,36 +2,124 @@
...
+<https://pythonhosted.org/spyder/>`_ which works on all major platforms and can be installed through ``pip``, and `PyCharm
+<https://www.jetbrains.com/pycharm/>`_ which can be integrated with any text editor of your choice, such as EMACS or Vim.
+
+Install ODL
 ===========
-You also need to install Git to be able to download odl.
+There are two ways to install ODL, either using ``pip`` or from source. Unless you will be contributing to the library we recommend using ``pip`` directly.
```
kohr-h: "Unless you plan to contribute [...]"

doc/source/guide/introduction/installing.rst
adler-j: Perhaps simply rename the `Python 2 vs 3` header to `What python version to use` and discuss that there? Feels like it would "dominate" this part by too much text.

kohr-h: I don't see the whole picture now, but I guess you're right. This can go into the discussion below.

kohr-h:
You're right, the instructions were a bit too detailed for the front page. Looks very good to me.

adler-j:
I also did this partly because it caused extra work for us by having duplication. Will fix comments and merge.

PR #338, opened 2016-04-05
Issue 334  derivative

### LINKED ISSUE #67: A better way of generating samples from a set
Currently we have the [odl/diagnostics/examples](https://github.com/odlgroup/odl/blob/master/odl/diagnostics/examples.py) package which helps us create elements in some sets, namely `RealNumbers`, `ComplexNumbers`, `FnBase` and `DiscreteLp`. Using this, we can generate some "standard" elements using the [`odl.diagnostics.samples`](https://github.com/odlgroup/odl/blob/0963cf4a03f42442221460fd7a34d3c892de1486/odl/diagnostics/examples.py#L189-#L215) function.

I propose that we generalize this somehow, specifically I have three goals:

* Extensibility: Users should be able to add their own set and still use our functions such as `OperatorTest` on them.
* Separation: We shouldnt have all code in one place.
* Ease of use: Users should be able to actually use this on sets when prototyping code.

The best solution to this would (imo) be some kind of iterator syntax over example elements with metadata, such as a name. Some ideas include

* add `__iter__` method to `Set`. This would allow users to do 

	```python
	for element in my_set: 
		pass
	```
	
	but comes at the cost of interfering with indexable sets, for example some code currently uses 	
	
	```python
	for subspace in my_productspace: 
		pass
	```
	
	and this would break.
	
* Add a free function syntax. For example we could have users somehow add a function (say `my_set_examples` and then register this with a central function

	```python
	def my_set_examples(space):
		pass
	
	register_examples(my_set_examples, My_Set_Class)
	``` 
	or perhaps using a decorator:
	
	```python
	@register_examples(My_Set_Class)
	def my_set_examples(space):
		pass
	``` 
	
	This could then be called either explicitly: 
	
	```python
	for ex in my_set_examples(my_set): 
		pass
	```
	
	or, with the registered polymorphic callback `samples`: 
	
	```python
	for ex in samples(my_set): 
		pass
	```
	
	this would also allow us to use this in functions such as `OperatorTest`. Imo this would work, but I kindoff feel it is somewhat complicated for new users.
	
* Finally we could make a member function `samples` optionally available on `Set` instances. Using this users could call

	```python
	for ex in my_set.samples(): 
		pass
	```
	
	This is my personal favourite since it is easy to think about as a user, and also easy to implement.

In addition to all of this, I suppose our users would like a way to get a single example. This could probably be done implicitly using any of the above implementations.

### LINKED ISSUE #334: Derivative test needs some rethinking
Our current derivative test computes

    ||A(x + c*p) - A(x) - c * A'(x)(p)|| / (|c| * ||p|| * ||A||)

and produces an error if this quantity is larger than 0.00001.

There are a couple of issues with that:

1. The test requires the specification of an estimate of the operator norm `||A||`, but that is hard to get for non-linear operators which one usually has in this test.
2. If the tested operator is differentiable and the implementation of its derivative is correct, then

        ||A(x + c*p) - A(x) - c * A'(x)(p)|| = o(c ||p||)

    i.e. the linearisation error drops asymptotically super-linearly. If the operator is twice continuously differentiable, the convergence is quadratic. However, the constant is basically unknown - it would be smaller than the norm of the second derivative `||A''||` times `||p||^2`, but who knows how to compute that first guy..
Anyway, the point is that normalizing by `1 / ||p||` and an operator norm and requiring the resulting relative error to be small is somewhat arbitrary. Also, if the convergence is only slightly faster than linear, reaching the required threshold would be hard even for correctly implemented derivatives.
3. Finally, at least for some types of operators, there is a sort of "saturation" effect. For example, with the deformation operator we're currently implementing, a very small disturbance of the deformation parameters (smaller than the sampling grid step) will have no effect, hence the first term

        A(x + c*p) - A(x)

   will be *exactly* zero for all `c` below a certain threshold (grid step if `p` is about unit length), hence the error term is *exactly*

        |c| * ||A'(x)(p)||

   i.e. one will observe exactly linear asymptotic behaviour, not super-linear. I suspect that all kinds of "domain manipulation" operators will behave like that.

#### What to do about that

I don't exactly know. Perhaps we can try to observe the asymptotic by testing a couple of `c` values of different scales and print the error together with its measured convergence rate.

adler-j:
* Made diagnostics in general less verbose.
* Improvements to derivative test, should close #334
* Moved sample generation to `set`, can now be arbitrarily customized. Closes #67
* Added `PowerOperator` (does what you'd expect), needed for example.


odl/set/sets.py
```diff
@@ -128,6 +128,16 @@ def element(self, inp=None):
...
         raise NotImplementedError("'element' method not implemented.")
 
+    @property
+    def examples(self):
+        """Returns a `generator` with elements in the set as name-value pairs.
+
+        Can return a finite set of examples or an infinite set.
+
+        Optional to implement, intended to be used for diagnostics.
+        Default yields ``('element()', self.element())``."""
```
kohr-h: `"""` on wrong line

odl/diagnostics/operator.py
```diff
@@ -41,7 +41,7 @@ class OperatorTest(object):
     adjoint definition.
     """
 
-    def __init__(self, operator, operator_norm=None):
+    def __init__(self, operator, operator_norm=None, verbose=True):
```
kohr-h: That's good!

odl/diagnostics/operator.py
```diff
@@ -213,27 +214,40 @@ def adjoint(self):
...
+                    err = (exact_step - expected_step).norm() / c
+
+                    if err < 1e-4:
+                        derivative_ok = True
+                        break
+                    else:
+                        minerror = min(minerror, err)
 
-    def derivative(self, step=0.0001):
+                    c /= 2.0
```
kohr-h: This will run 20 times in the worst case. Feels a bit much if we're doing that for all examples. How about dividing by 10?

odl/diagnostics/operator.py
```diff
@@ -213,27 +214,40 @@ def adjoint(self):
...
 
-    def _derivative_convergence(self, step):
-        print('\nTesting derivative is linear approximation')
+    def _derivative_convergence(self):
+        name = 'Testing derivative is linear approximation'
 
-        with FailCounter("error = ||A(x+c*dx)-A(x)-c*A'(x)(dx)|| / "
-                         "|c| ||dx|| ||A||") as counter:
+        with FailCounter(name, "error = "
+                         "inf_c ||A(x+cp)-A(x)-A'(x)(cp)|| / c") as counter:
```
kohr-h: This is definitely a better solution than the one before. Probably it's very hard to be more general than that.

odl/discr/lp_discr.py
```diff
@@ -258,6 +258,97 @@ def element(self, inp=None):
...
     @property
+    def examples(self):
+        """Return example functions in the space.
+
+        Example functions include:
+
+        Zero
+        One
+        Heaviside function
+        Hypercube
```
kohr-h: Hypercube characteristic function

odl/discr/lp_discr.py
```diff
@@ -258,6 +258,97 @@ def element(self, inp=None):
...
+        Hypersphere
```
kohr-h: same here

odl/discr/lp_discr.py
```diff
@@ -258,6 +258,97 @@ def element(self, inp=None):
...
+        Gaussian
+        Gradients
```
kohr-h: Perhaps linear gradients, could be confused with gradient operator

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+
+    def __init__(self, domain, n):
+        """Initialize a MultiplyOperator instance.
+
+        Parameters
+        ----------
+        n : Number
+            The power to take
+        domain : `LinearSpace` or `Field`, optional
+            The set to take values in.
```
kohr-h: No dot at the end

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+    ``MultiplyOperator(n)(x) <==> x ** n``
+
+    Here, ``x`` is a `LinearSpaceVector` or `Field` element and
+    ``y`` is a number.
+    Hence, this operator can be defined either on a `LinearSpace` or on
+    a `Field`.
+    """
+
+    def __init__(self, domain, n):
+        """Initialize a MultiplyOperator instance.
```
kohr-h: So we need some normalization here: do we use the class name in this summary or not? We should be stringent then, now it's mixed.

kohr-h: I don't think it's really needed since it will be not far below the class name in the documentation.

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+            out.assign(x)
+            out **= self.n
+
+    def derivative(self, point):
+        """The derivative operator.
+
+        Parameters
+        ----------
+        point : domain element
+            The point in which to take the derivative.
```
kohr-h: no dot

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+
+        Returns
+        -------
+        derivative : `MultiplyOperator`
+            The derivative in ``point``.
```
kohr-h: no dot

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+
+    ``MultiplyOperator(n)(x) <==> x ** n``
+
+    Here, ``x`` is a `LinearSpaceVector` or `Field` element and
+    ``y`` is a number.
+    Hence, this operator can be defined either on a `LinearSpace` or on
+    a `Field`.
+    """
+
+    def __init__(self, domain, n):
```
kohr-h: `n -> exponent`?

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+        derivative : `MultiplyOperator`
+            The derivative in ``point``.
+
+        Examples
+        --------
+        >>> from odl import Rn, RealNumbers
+
+        Use with vectors
+
+        >>> op = PowerOperator(Rn(3), 2)
```
kohr-h: use keyword arg

odl/operator/default_ops.py
```diff
@@ -366,7 +369,118 @@ def __repr__(self):
...
+        out : ``range`` `element`
+            Result of the multiplication. If ``out`` was provided, the
+            returned object is a reference to it.
+
+        Examples
+        --------
+        Use with vectors
+
+        >>> from odl import Rn, RealNumbers
+        >>> op = PowerOperator(Rn(3), 2)
```
kohr-h: Keyword argument makes it more clear

odl/set/sets.py
kohr-h: "By default, the generator yields ..."

odl/space/pspace.py
```diff
@@ -390,6 +391,14 @@ def element(self, inp=None, cast=True):
 
         return self.element_type(self, parts)
 
+    @property
+    def examples(self):
+        """Return examples from all sub-spaces."""
+        for examples in product(*[spc.examples for spc in self.spaces]):
+            name = ', '.join(name for name, _ in examples)
+            vector = self.element([vec for _, vec in examples])
+            yield (name, vector)
```
kohr-h: Perhaps draw a number of examples at random? Or the diagnostics method could take a "random" flag?

kohr-h:
Looking good in general. Just some minor remarks.

odl/diagnostics/operator.py
adler-j: Sadly havnt done more :D

odl/diagnostics/operator.py
adler-j: Sure why not

odl/discr/lp_discr.py
adler-j: Now that I think of it, shouldnt these be examples in the underlying uspace, and we simply discretize these here? That way we could move this to the general discretization class.

kohr-h: Totally correct!

odl/space/pspace.py
adler-j: Not quite sure how to skip x elements in iterators, also we dont know how many exist in the subspaces. This would be non-trivial to say the least. I agree with the sentiment though.

adler-j:
Fixed most comments, two bigger style changes

* `np.array([1.0])` is no longer a valid input array, it caused several bugs (it was cast to a scalar, which threw an error). It should likely be interpreted as a scalar.
* Moved discretization examples to fspace, now simply discretizes the uspace examples.

Also minor style, for example update the examples file.

odl/space/pspace.py
kohr-h: Yeah, it was just an idea. I don't know either how to do that.

kohr-h:
Good to go

Issue #344, opened 2016-04-08
Problem with backtracking step length calculations


aringh:
I think the backtracking step length calculations could need some caution.

1. If f(x) is finite, but f(x + a * direction) = infinite, and there is not enough iterations (before one hits the maximum number) in order to make f(x + a * direction) finite, this will cause problems for the user.

2. If f(x) = NaN, caution is also important. I'm not sure how NaN is interpreted, but for example
```
>>> np.nan >= 1
False
>>> np.nan <= 1
False
```
so if f(x + a*direction) = NaN, it is never going to enter the loop to make the step length smaller.

Edit: I have a fix to this, just need to find it. Then I will push it.

adler-j:
I wrote this in the pull request, but my opinion is that we should be vary to enforce fixes here. In general, feeding a NaN to a function expecting a float is regarded as an exception. Perhaps for some applications it is reasonable to interpret NaN as Inf (thus shortening the step until it is OK), but in other it is not. I.m.o, a function returning NaN should be interpreted as an error, and we should throw some reasonable exception. It is then up to the user to fix this inside the functional, i.e. add this before returning `result`

```python
result = float('Inf') if result == float('Nan') else result
```

aringh:
Your pull request in #348 looks fine to me. It gives clear answers to problems encountered.

aringh:
I think this was closed via #348

PR #345, opened 2016-04-08
ENH: Safeguarding of backtracking linesearch.

aringh:
See issue #344.

### DIFF
--- a/odl/solvers/scalar/steplen.py
+++ b/odl/solvers/scalar/steplen.py
@@ -148,6 +148,21 @@ class BacktrackingLineSearch(LineSearch):
             The computed step length
         """
         alpha = 1.0
+
+        # Since all comparisons with nan returns False, in order to not get a
+        # problem with nan-values we throw an error.
+        if self.function(x) == float('nan'):
+            raise NameError('ObjFunIsNanError')
+
+        # Next step is done in order to get a step length that does not give
+        # cost nan or infinite. Note that for convex functions, if the current
+        # point and the current point + alpha * direction does not give nan or
+        # inf this guarantees that any point along the line combinding these
+        # will not be inf either.
+        while (self.function(x + alpha * direction) == float('nan') or
+                self.function(x + alpha * direction) == float('inf')):
+            alpha *= self.tau
+
         fx = self.function(x)
         num_iter = 0
         while ((self.function(x + alpha * direction) >=


odl/solvers/scalar/steplen.py
```diff
@@ -148,6 +148,21 @@ def __call__(self, x, direction, dir_derivative):
             The computed step length
         """
         alpha = 1.0
+
+        # Since all comparisons with nan returns False, in order to not get a
+        # problem with nan-values we throw an error.
+        if self.function(x) == float('nan'):
```
kohr-h: Doing this check with `fx` below saves one function evaluation.

odl/solvers/scalar/steplen.py
```diff
@@ -148,6 +148,21 @@ def __call__(self, x, direction, dir_derivative):
...
+            raise NameError('ObjFunIsNanError')
```
kohr-h: `NameError` [is the wrong exception](https://docs.python.org/3.5/library/exceptions.html#NameError) for this case. It's raised when you refer to a name in the code which has never been bound to a value. Like
```python
>>> u  # in a fresh console
NameError: name 'u' is not defined
```
The correct exception is `ValueError` with a reasonable message like `'function value is NaN, comparison for line search impossible.'`.

The more interesting issue is how to handle this issue one level above. Should the method bail out or rather try some new point `x` to get out of the NaN regime?

In any case, the docstring of this line search class should clearly state that this exception is raised if the function value NaN is occurred. Add a section like this:
```
Raises
------
ValueError
    if the function value at ``x`` is NaN
```

odl/solvers/scalar/steplen.py
```diff
@@ -148,6 +148,21 @@ def __call__(self, x, direction, dir_derivative):
             The computed step length
         """
         alpha = 1.0
+
+        # Since all comparisons with nan returns False, in order to not get a
+        # problem with nan-values we throw an error.
```
kohr-h: Good to have a comment. A bit twisted sentence though. Perhaps (slightly shorter)
```python
# Cannot meaningfully compare to NaN -> raise an exception
```

odl/solvers/scalar/steplen.py
```diff
@@ -148,6 +148,21 @@ def __call__(self, x, direction, dir_derivative):
...
+        if self.function(x) == float('nan'):
+            raise NameError('ObjFunIsNanError')
+
+        # Next step is done in order to get a step length that does not give
+        # cost nan or infinite. Note that for convex functions, if the current
+        # point and the current point + alpha * direction does not give nan or
+        # inf this guarantees that any point along the line combinding these
+        # will not be inf either.
```
kohr-h: Also a bit long-ish. Could be shortened to
```python
# Get out of the NaN- or Inf-region first. Note that for convex functions,
# subsequent points along the line will be finite numbers, too.
```

kohr-h:
Minor stuff mostly.

How about a small test for this?

odl/solvers/scalar/steplen.py
adler-j: This is critical to do, as of now it would require two calls, often doubling execution time.

adler-j:
I disagree with fixing this in this way. i.m.o it is up to the writer of the functional to fix this. We can however check for correct input by checking any calls to `self.function` and see if the result is `NaN`, in that case we simply throw and ValueError:

Function returned NaN at (point, step).

If the user wants to interpret NaN as Inf, he should do so inside the function definition.

odl/solvers/scalar/steplen.py
```diff
@@ -148,6 +148,21 @@ def __call__(self, x, direction, dir_derivative):
...
+        while (self.function(x + alpha * direction) == float('nan') or
+                self.function(x + alpha * direction) == float('inf')):
```
kohr-h: Didn't see this the first time: This is also very sub-optimal. It will almost always evaluate twice, even if both conditions are false. And the value is not reused in the next loop.

The value `fxstep = self.function(x + alpha * direction)` should be stored, then used in the condition here, and finally used as first value "proper" line search:
```python
if fxstep < fx + alpha * dir_derivative * self.discount:
    return alpha
else:
    num_iter = 1
while ...
```

odl/solvers/scalar/steplen.py
```diff
@@ -148,6 +148,21 @@ def __call__(self, x, direction, dir_derivative):
...
+            alpha *= self.tau
+
         fx = self.function(x)
```
kohr-h: Another slight optimization would be to check if `fx == float('inf')` and return immediately if `True`, saving one evaluation by not entering the loop.

kohr-h:
> I disagree with fixing this in this way. i.m.o it is up to the writer of the functional to fix this.

If you mean the NaN issue by this, I agree - that's more a sign of wrong implementation than a well-defined situation we need to handle.
The situation of infinite value is well-defined, though. That's a typical way of marking an absolutely forbidden area. Well, now that I think about it, `inf` does not require extra handling at all - it will already be correctly handled by the `while` loop. One could argue that the extra loop for the infinite part of the cost function will run quicker since the expression `fx + alpha * ...` is not evaluated. (And no, the comparison is not short-cut with `float('inf')` on the left).

odl/solvers/scalar/steplen.py
aringh: For now I think it can raise a ```ValueError```. But it might be worth thinking about what can be done instead...

aringh:
Just an example on the issue that might occur: if maximum number of iterations is not specified by the user then right now it is set so that one can reach a step length of 0.0001. If this is not short enough to get out of the inf-region, this will most likely cause problems. It can be fixed by setting maximum number of iterations higher, however it can also be hard to detect where the error comes from.

We can either see the problem as wrongly posed from the user side and let them solve it. Or try to help them. (My way of expressing myself probably indicates what I think would be nice ;) However, if you think that this causes extra overhead and blurs the code, then we skip it.)

aringh:
This should maybe go in another forum, but how do you even handle the case of NaN? Am I doing something wrong here?
```
>>> np.nan == np.nan
False
>>> np.nan == float('nan')
False
>>> float('nan') == float('nan')
False
```
Or are you forced to use ```np.isnan``` in order to check?
```
>>> np.isnan(np.nan)
True
```

kohr-h:
I think basically `NaN` is interpreted as "coming from an invalid calculation", and (I guess) to make failure most likely, *any* comparison with `NaN` (even of `NaN`) will result in `False`.

In my view, we should find a different solution and say in the documentation that `NaN` is not handled. Generally, the principle here is KISS - keep it small and simple. Including too much "intelligence" will likely introduce bugs or screw up in other valid situations.

adler-j:
I agree that `NaN` should be interpreted as "coming from an invalid calculation", hence the nice way is to simply remind the user that his function has returned a bad value and exit. That would be enough intelligence for me.

`Inf` should however be handled properly.

adler-j:
Made an alternative Pull to show what I intend.

adler-j:
Merged #348 instead.

Issue #322, opened 2016-03-28
ProductSpaceOperator on weighted spaces


kohr-h:
Weighted product spaces are on the way, so we need to handle them correctly when we define operators on them. It's mostly the adjoints which need to be adapted. The same is true for our `MatVecOperator`, by the way.

PR #346, opened 2016-04-09
ENH: add point-wise norm and inner product for vector fields

kohr-h:
This is convenience functionality useful when working with vector fields. It adds the pointwise norm and inner product operators, the first with derivative, the second with adjoint.

Not sure about the place in the ecosystem. If we plan to support vector and tensor fields properly in the future, this module may want to go to the same place. Some things could be simplified then, of course, this is what we can do right now.

I left out variable exponent for now, perhaps I'll add that in short. Haven't decided yet.

The tests run fine, and I also checked that the adjoint and derivatives are correct (not with zero, of course). As a side remark, these operators handle weighting of the product spaces correctly. We should fix that also for the `ProductSpaceOperator` in general (see #322).

### DIFF
--- a/odl/discr/tensor_ops.py
+++ b/odl/discr/tensor_ops.py
@@ -1,0 +1,409 @@
+﻿# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Operators defined for tensor fields."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import super
+from future.utils import raise_from
+
+import numpy as np
+
+from odl.operator.operator import Operator
+from odl.set.sets import RealNumbers
+from odl.space.fspace import FunctionSpace
+from odl.space.pspace import ProductSpace
+
+
+__all__ = ('PointwiseOperator', 'PointwiseNorm')
+
+
+class PointwiseOperator(Operator):
+
+    """Abstract operator for point-wise tensor field manipulations.
+
+    A point-wise operator acts on a space of vector or tensor fields,
+    i.e. a power space ``X^d`` of a discretized function space ``X``.
+    Its range is the power space ``X^k`` with a possibly different
+    number ``k`` of components.
+
+    For example, if ``X`` is a `DiscreteLp` space, then
+    ``ProductSpace(X, d)`` is a valid domain for any positive integer
+    ``d``.
+
+    Currently, only vector fields, i.e. one-dimensional products of
+    ``X``, are supported.
+
+    See also
+    --------
+    ProductSpace
+    """
+
+    def __init__(self, domain, ran_size, linear=False):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain : `ProductSpace` of `Discretization`
+            Space of vector fields on which the operator acts.
+            It has to be a product space of identical discretized
+            function spaces, i.e. a power space.
+        ran_size : positive `int`
+            Number of components of the range
+        linear : `bool`, optional
+            If `True`, assume that the operator is linear.
+        """
+        if not isinstance(domain, ProductSpace):
+            raise TypeError('domain {!r} is not a ProductSpace instance.'
+                            ''.format(domain))
+        if not domain.is_power_space:
+            raise TypeError('domain {!r} is not a power space.'.format(domain))
+
+        if not isinstance(domain[0].uspace, FunctionSpace):
+            raise TypeError('domain component {!r} is not a function space '
+                            'discretization.'.format(domain[0]))
+
+        ran_size, ran_size_in = int(ran_size), ran_size
+        if ran_size <= 0:
+            raise ValueError('number of range components must be positive, '
+                             'got {}.'.format(ran_size_in))
+        self._ran_size = ran_size
+
+        # Range will be just 'X' if there is only one component
+        if self.ran_size == 1:
+            ran = domain[0]
+        else:
+            ran = ProductSpace(domain[0], self.ran_size)
+        super().__init__(domain=domain, range=ran, linear=linear)
+
+    @property
+    def discr_fspace(self):
+        """The base space ``X`` of this operator's domain and range."""
+        return self.domain[0]
+
+    @property
+    def ran_size(self):
+        """Number of component spaces in the range."""
+        return self._ran_size
+
+
+class PointwiseNorm(PointwiseOperator):
+
+    """Take the point-wise norm of a tensor field.
+
+    This operator takes the (weighted) ``p``-norm
+
+        ``||F(x)|| = [ sum_j( w_j * |F_j(x)|^p ) ]^(1/p)``
+
+    for ``p`` finite and
+
+        ``||F(x)|| = max_j( w_j * |F_j(x)| )``
+
+    for ``p = inf``, where ``F`` is a tensor field. This implies that
+    the `Operator.domain` is a power space of a discretized function
+    space. For example, if ``X`` is a `DiscreteLp` space, then
+    ``ProductSpace(X, d)`` is a valid domain for any positive integer
+    ``d``.
+
+    Currently only vector fields, i.e. product spaces with lenght-1
+    shape, are supported.
+    """
+
+    def __init__(self, domain, exponent=None, weight=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain : `ProductSpace`
+            Space of vector fields on which the operator acts.
+            It has to be a product space of identical discretized
+            function spaces, i.e. a power space. Its
+            `ProductSpace.shape` must have length 1.
+        exponent : non-zero `float`, optional
+            Exponent of the norm in each point. Values between
+            0 and 1 are currently not supported due to numerical
+            instability.
+            Default: ``domain.exponent``
+        weight : `array-like` or `float`, optional
+            Weighting array or constant for the norm. If an array is
+            given, its length must be equal to ``domain.size``, and
+            all entries must be positive. A provided constant must be
+            positive.
+            By default, the weights are is taken from
+            ``domain.weighting``. Note that this excludes unusual
+            weightings with custom inner product, norm or dist.
+        """
+        super().__init__(domain, ran_size=1, linear=False)
+        if len(self.domain.shape) != 1:
+            raise NotImplementedError
+
+        if exponent is None:
+            if domain.exponent is None:
+                raise ValueError('cannot determine exponent from {}.'
+                                 ''.format(self.domain))
+            self._exponent = self.domain.exponent
+        elif 0 <= exponent < 1:
+            raise ValueError('exponent between 0 and 1 not allowed.')
+        else:
+            self._exponent = float(exponent)
+
+        # Handle weighting, including sanity checks
+        if weight is None:
+            if hasattr(self.domain.weighting, 'vector'):
+                self._weights = self.domain.weighting.vector
+            elif hasattr(self.domain.weighting, 'const'):
+                self._weights = (self.domain.weighting.const *
+                                 np.ones(len(self.domain)))
+            else:
+                raise ValueError('weighting scheme {!r} of the domain does '
+                                 'not define a weighting vector or constant.'
+                                 ''.format(self.domain.weighting))
+        elif np.isscalar(weight):
+            if weight <= 0:
+                raise ValueError('weighting constant must be positive, got '
+                                 '{}.'.format(weight))
+            self._weights = float(weight) * np.ones(self.domain.size)
+        else:
+            self._weights = np.asarray(weight, dtype='float64')
+            if (not np.all(self.weights > 0) or
+                    not np.all(np.isfinite(self.weights))):
+                raise ValueError('weighting array {} contains invalid '
+                                 'entries.'.format(weight))
+        self._is_weighted = (not np.all(np.array_equiv(self.weights, 1.0)))
+
+    @property
+    def exponent(self):
+        """Exponent ``p`` of this norm."""
+        return self._exponent
+
+    @property
+    def weights(self):
+        """Weighting vector of this norm."""
+        return self._weights
+
+    @property
+    def is_weighted(self):
+        """`True` if weighting is not 1 or all ones."""
+        return self._is_weighted
+
+    def _call(self, f, out):
+        """Implement ``self(f, out)``."""
+        if len(self.domain.shape) > 1:
+            raise NotImplementedError
+
+        if self.exponent == 1.0:
+            self._call_vecfield_1(f, out)
+        elif self.exponent == float('inf'):
+            self._call_vecfield_inf(f, out)
+        else:
+            self._call_vecfield_p(f, out)
+
+    def _call_vecfield_1(self, vf, out):
+        """Implement ``self(vf, out)`` for exponent 1."""
+        vf[0].ufunc.absolute(out=out)
+        if self.is_weighted:
+            out *= self.weights[0]
+
+        if self.domain.size == 1:
+            return
+
+        tmp = self.range.element()
+        for fi, w in zip(vf[1:], self.weights[1:]):
+            fi.ufunc.absolute(out=tmp)
+            if self.is_weighted:
+                tmp *= w
+            out += tmp
+
+    def _call_vecfield_inf(self, vf, out):
+        """Implement ``self(vf, out)`` for exponent ``inf``."""
+        vf[0].ufunc.absolute(out=out)
+        if self.is_weighted:
+            out *= self.weights[0]
+
+        if self.domain.size == 1:
+            return
+
+        tmp = self.range.element()
+        for fi, w in zip(vf[1:], self.weights[1:]):
+            fi.ufunc.absolute(out=tmp)
+            if self.is_weighted:
+                tmp *= w
+            out.ufunc.maximum(tmp, out=out)
+
+    def _call_vecfield_p(self, vf, out):
+        """Implement ``self(vf, out)`` for exponent 1 < p < ``inf``."""
+        # Optimization for 1 component - just absolute value (maybe weighted)
+        if self.domain.size == 1:
+            vf[0].ufunc.absolute(out=out)
+            if self.is_weighted:
+                out *= self.weights[0] ** (1 / self.exponent)
+            return
+
+        # Initialize out, avoiding one copy
+        self._abs_pow_ufunc(vf[0], out=out)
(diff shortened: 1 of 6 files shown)


odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,556 @@
...
+
+        ``||F(x)|| = max_j( w_j * |F_j(x)| )``
+
+    for ``p = inf``, where ``F`` is a tensor field. This implies that
+    the `Operator.domain` is a power space of a discretized function
+    space. For example, if ``X`` is a `DiscreteLp` space, then
+    ``ProductSpace(X, d)`` is a valid domain for any positive integer
+    ``d``.
+
+    Currently only vector fields, i.e. product spaces with lenght-1
```
kohr-h: typo: lenght -> length

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,556 @@
...
+
+    @property
+    def base_space(self):
+        """The base space ``X`` of this operator's domain and range."""
+        return self._base_space
+
+
+class PointwiseNorm(PointwiseOperator):
+
+    """Take the point-wise norm of a tensor field.
```
kohr-h: Perhaps write "vector field" for now

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,556 @@
...
+
+    This operator takes the (weighted) ``p``-norm
+
+        ``||F(x)|| = [ sum_j( w_j * |F_j(x)|^p ) ]^(1/p)``
+
+    for ``p`` finite and
+
+        ``||F(x)|| = max_j( w_j * |F_j(x)| )``
+
+    for ``p = inf``, where ``F`` is a tensor field. This implies that
```
kohr-h: Definitions will change for higher order tensor fields, so probably sticking with vector fields it should be for now.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,556 @@
...
+        if self.domain.field == ComplexNumbers():
+            raise NotImplementedError('operator not Frechet-differentiable '
+                                      'on a complex space.')
+
+        if self.exponent == float('inf'):
+            raise NotImplementedError('operator not Frechet-differentiable '
+                                      'for exponent = inf.')
+
+        vf = self.domain.element(vf)
+        vf_pwnorm_fac = self(vf) ** (self.exponent - 1)
```
kohr-h: Two-step with exponentiation in place

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,556 @@
...
+        # Bail out if the space is complex but we cannot take the complex
+        # conjugate.
+        if (self.domain.field == ComplexNumbers() and
+                not hasattr(self.base_space.element_type, 'conj')):
+            raise NotImplementedError(
+                'base space element type {!r} does not implement conj() '
+                'method required for complex inner products.'
+                ''.format(self.base_space.element_type))
+
+        # Store vector field and complex conjugate if desired
```
kohr-h: Old comment

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,556 @@
...
+            if (not np.all(self.weights > 0) or
+                    not np.all(np.isfinite(self.weights))):
+                raise ValueError('weighting array {} contains invalid '
+                                 'entries.'.format(weight))
+        self._is_weighted = (not np.all(np.array_equiv(self.weights, 1.0)))
+
+    @property
+    def vecfield(self):
+        """Fixed vector field ``G`` of this inner product.
+        """
```
kohr-h: Wrong line break

kohr-h:
Bump. Main questions are placement and optimisation.

kohr-h:
I added some doctests.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+from odl.operator.operator import Operator
+from odl.set.sets import RealNumbers, ComplexNumbers
+from odl.space.fspace import FunctionSpace
+from odl.space.pspace import ProductSpace
+
+
+__all__ = ('PointwiseOperator', 'PointwiseNorm', 'PointwiseInner')
+
+
+class PointwiseOperator(Operator):
```
adler-j: Perhaps this name is somewhat too general. Some could interpret it as any "diagonal" operator X - > X. We may need to add something to the name to enforce that we mean for productspaces.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+    --------
+    ProductSpace
+    """
+
+    def __init__(self, domain, range, linear=False):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain, range : {`ProductSpace`, `Discretization`}
```
adler-j: We write discretization here, but seem to accept anything. Any reason to enforce discretization at all?

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+            ``X`` or the base space itself (only one of them).
+        linear : `bool`, optional
+            If `True`, assume that the operator is linear.
+        """
+        if isinstance(domain, ProductSpace):
+            if not domain.is_power_space:
+                raise TypeError('domain {!r} is not a power space.'
+                                ''.format(domain))
+
+            if not isinstance(domain[0].uspace, FunctionSpace):
```
adler-j: Here you assume that uspace exists, needs not be true at this point

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+            dom_base = domain[0]
+        else:
+            dom_base = domain
+
+        if isinstance(range, ProductSpace):
+            if not range.is_power_space:
+                raise TypeError('range {!r} is not a power space.'
+                                ''.format(range))
+
+            if not isinstance(range[0].uspace, FunctionSpace):
```
adler-j: uspace not known to exist

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+        if isinstance(domain, ProductSpace):
+            if not domain.is_power_space:
+                raise TypeError('domain {!r} is not a power space.'
+                                ''.format(domain))
+
+            if not isinstance(domain[0].uspace, FunctionSpace):
+                raise TypeError('domain base space {!r} is not a function '
+                                'space discretization.'.format(domain[0]))
+
+            dom_base = domain[0]
```
adler-j: We actually accept 0 dimensional productspaces, in which case this fails. Dunno if we want to check for that though, its mostly a hack to enable things like:
```python
# space is 3 dimensional
for x in space[3:]: print(x)
```

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+                                'space discretization.'.format(range[0]))
+
+            ran_base = range[0]
+        else:
+            ran_base = range
+
+        if dom_base != ran_base:
+            raise ValueError('domain and range have different base spaces '
+                             '({!r} != {!r}).'
+                             ''.format(domain[0], range[0]))
```
adler-j: This will fail if the space is not a productspace, should be `dom_base, ran_base`

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+        >>> print(pw_norm(x))
+        [[1.0, 2.0],
+         [2.0, 7.0]]
+        """
+        if not isinstance(vfspace, ProductSpace):
+            raise TypeError('vector field space {!r} is not a ProductSpace '
+                            'instance.'.format(vfspace))
+        super().__init__(domain=vfspace, range=vfspace[0], linear=False)
+
+        if len(self.domain.shape) != 1:
```
adler-j: What do we test for here?

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+            raise NotImplementedError
+
+        if exponent is None:
+            if self.domain.exponent is None:
+                raise ValueError('cannot determine exponent from {}.'
+                                 ''.format(self.domain))
+            self._exponent = self.domain.exponent
+        elif 0 <= exponent < 1:
+            raise ValueError('exponent between 0 and 1 not allowed.')
```
adler-j: But we do allow negative exponents?

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+                raise ValueError('weighting constant must be positive, got '
+                                 '{}.'.format(weight))
+            self._weights = float(weight) * np.ones(self.domain.size)
+        else:
+            self._weights = np.asarray(weight, dtype='float64')
+            if (not np.all(self.weights > 0) or
+                    not np.all(np.isfinite(self.weights))):
+                raise ValueError('weighting array {} contains invalid '
+                                 'entries.'.format(weight))
+        self._is_weighted = (not np.all(np.array_equiv(self.weights, 1.0)))
```
adler-j: The first parentesis is not needed, nor do i think the np.all is.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+        elif self.exponent == float('inf'):
+            self._call_vecfield_inf(f, out)
+        else:
+            self._call_vecfield_p(f, out)
+
+    def _call_vecfield_1(self, vf, out):
+        """Implement ``self(vf, out)`` for exponent 1."""
+        vf[0].ufunc.absolute(out=out)
+        if self.is_weighted:
+            out *= self.weights[0]
```
adler-j: Why weights[0]?

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+
+        if self.domain.size == 1:
+            return
+
+        tmp = self.range.element()
+        for fi, w in zip(vf[1:], self.weights[1:]):
+            fi.ufunc.absolute(out=tmp)
+            if self.is_weighted:
+                tmp *= w
```
adler-j: why do this at each step? It should be possible to wait.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,610 @@
...
+
+        # Initialize out, avoiding one copy
+        self._abs_pow_ufunc(vf[0], out=out)
+        if self.is_weighted:
+            out *= self.weights[0]
+
+        tmp = self.range.element()
+        for fi, w in zip(vf[1:], self.weights[1:]):
+            self._abs_pow_ufunc(fi, out=tmp)
+            if self.is_weighted:
```
adler-j: As above, apply weighting after loop

odl/discr/tensor_ops.py
kohr-h: Should check that it's a discretization first

odl/discr/tensor_ops.py
kohr-h: Need `uspace`.
Actually, it should be a space of vector fields, i.e. a power space of discretized function spaces. It's a bit involved since we haven't properly defined vector fields yet.

odl/discr/tensor_ops.py
kohr-h: We could write `PointwiseTensorFieldOperator` to be totally future-proof.

odl/discr/tensor_ops.py
kohr-h: I'll add a check that length is at least 1.

odl/discr/tensor_ops.py
kohr-h: Of course

odl/discr/tensor_ops.py
kohr-h: Currently nothing. Perhaps it should be a comment for later, when (perhaps) product spaces with more than 1-dimensional shapes are allowed.

odl/discr/tensor_ops.py
kohr-h: The `np.all` is a mistake, where I first wanted to write `self.weights == 1.0`

odl/discr/tensor_ops.py
kohr-h: Shouldn't. If it's not checked, I'll add it.

odl/discr/tensor_ops.py
kohr-h: How would you do that? The `w` is different for each product space component, so I guess we have to do one weighting component at a time.

kohr-h: I'm computing `sum_j(w[j] * x[j])` so I don't think I can do that outside.

odl/discr/tensor_ops.py
kohr-h: On the other hand, we could just allow any power space and duck-type

kohr-h:
Removed the overly angry checks, now any power space can be (mis-)used.

kohr-h:
Anything important left to do or is it mergable?

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+            if not range.is_power_space:
+                raise TypeError('range {!r} is not a power space.'
+                                ''.format(range))
+
+            if range.size == 0:
+                raise ValueError('range is a product space of size 0.')
+
+            ran_base = range[0]
+        else:
+            ran_base = range
```
adler-j: check LinearSpace requirement

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+            if not domain.is_power_space:
+                raise TypeError('domain {!r} is not a power space.'
+                                ''.format(domain))
+
+            if domain.size == 0:
+                raise ValueError('domain is a product space of size 0.')
+
+            dom_base = domain[0]
+        else:
+            dom_base = domain
```
adler-j: check LinearSpace requirement

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+from builtins import super
+
+import numpy as np
+
+from odl.operator.operator import Operator
+from odl.set.sets import RealNumbers, ComplexNumbers
+from odl.space.pspace import ProductSpace
+
+
+__all__ = ('PointwiseTensorFieldOperator', 'PointwiseNorm', 'PointwiseInner')
```
adler-j: Regarding #358, this should probably not expose `PointwiseTensorFieldOperator` all the way out to the top namespace

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+
+    def __init__(self, domain, range, linear=False):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain, range : {`ProductSpace`, `LinearSpace`}
+            Spaces of vector fields between which the operator maps.
+            They have to be either power spaces of the same base space
+            ``X`` or the base space itself (only one of them).
```
kohr-h: Say that length 0 is not allowed.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+    ``ProductSpace(X, d)`` is a valid domain for any positive integer
+    ``d``.
+    """
+
+    def __init__(self, vfspace, exponent=None, weight=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        vfspace : `ProductSpace`
```
adler-j: what does vfspace imply?

odl/discr/tensor_ops.py
kohr-h: Right, I'll remove it.

odl/discr/tensor_ops.py
kohr-h: Sure. Was too quick removing all checks.

odl/discr/tensor_ops.py
kohr-h: vector field space

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+        """
+        if self.domain.field == ComplexNumbers():
+            raise NotImplementedError('operator not Frechet-differentiable '
+                                      'on a complex space.')
+
+        if self.exponent == float('inf'):
+            raise NotImplementedError('operator not Frechet-differentiable '
+                                      'for exponent = inf.')
+
+        vf = self.domain.element(vf)
```
adler-j: i love this, but we don't do it in general. Perhaps we should.

kohr-h:
I think I'll detach 615da37 (removing the dist=0 tests) from this PR, has nothing to do with it.

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+        >>> vfspace = odl.ProductSpace(spc, 2)
+        >>> fixed_vf = np.array([[[0, 1], [-1, 1]],
+        ...                      [[1, -1], [1, -1]]])
+        >>> pw_inner = PointwiseInner(vfspace, fixed_vf)
+        >>> pw_inner.range == spc
+        True
+
+        Now we can calculate the inner product in each point:
+
+        >>> x = vfspace.element([[[1, 2], [0, 3]],
```
adler-j: nonstandard formatting for 3d arrays

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+            raise NotImplementedError(
+                'base space element type {!r} does not implement conj() '
+                'method required for complex inner products.'
+                ''.format(self.base_space.element_type))
+
+        self._vecfield = self.domain.element(vecfield)
+
+        # Handle weighting, including sanity checks
+        if weight is None:
+            if hasattr(self.domain.weighting, 'vector'):
```
adler-j: this should perhaps be standardized at some point. Maybe write a TODO

odl/discr/tensor_ops.py
```diff
@@ -0,0 +1,599 @@
...
+            self._ran_weights = (self.range.weighting.const *
+                                 np.ones(len(self.range)))
+        else:
+            raise ValueError('weighting scheme {!r} of the range does '
+                             'not define a weighting vector or constant.'
+                             ''.format(self.range.weighting))
+
+    def _call(self, f, out):
+        """Implement ``self(vf, out)``."""
+        for vfi, oi, vi, wi in zip(self.vecfield, out,
```
adler-j: naming not obvious

odl/space/pspace.py
```diff
@@ -519,6 +525,7 @@ def __eq__(self, other):
         else:
             return (isinstance(other, ProductSpace) and
                     len(self) == len(other) and
```
adler-j: change this to `self.shape == other.shape`, more future proof

adler-j:
Minor nitpicky stuff, looks nice.

odl/discr/tensor_ops.py
kohr-h: I know, but it doesn't look any better with one row per line. Hard to make clear that we're creating a product space vector with 2 components, each 2x2.

kohr-h: Perhaps I'll just write a comment

adler-j: Why not use a 1d space, or even a Rn type space instead?

odl/discr/tensor_ops.py
kohr-h: Here it's especially useful since you can give the vector field directly as a Numpy array of shape `(d, n1, ..., nd)` with `d = ndim`. I agree that this would be good to do in general.

odl/discr/tensor_ops.py
kohr-h: That's not a bad idea!

odl/discr/tensor_ops.py
kohr-h: I'll do that

odl/space/pspace.py
kohr-h: Right.

kohr-h:
I'll fix the stuff, rebase and merge.

PR #361, opened 2016-04-24
BUG: fix interpolation not writing to CUDA arrays

### LINKED ISSUE #347: Interpolation fails silently for cuda spaces
The following line fails to write to the output in case of cuda data.

```python
            func(mesh, out=out.asarray().reshape(self.grid.shape,
                                                 order=self.order))
```

https://github.com/odlgroup/odl/blob/master/odl/discr/discr_mappings.py#L273-L274

kohr-h:
Should fix #347

### DIFF
--- a/test/discr/discr_mappings_test.py
+++ b/test/discr/discr_mappings_test.py
@@ -32,7 +32,8 @@ from odl.discr.grid import sparse_meshgrid
 from odl.discr.discr_mappings import (
     PointCollocation, NearestInterpolation, LinearInterpolation,
     PerAxisInterpolation)
-from odl.util.testutils import all_almost_equal, all_equal, almost_equal
+from odl.util.testutils import (
+    all_almost_equal, all_equal, almost_equal, skip_if_no_cuda)
 
 
 def test_nearest_interpolation_1d_complex():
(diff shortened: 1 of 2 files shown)


adler-j:
Perhaps we should have a test as well?

kohr-h:
Here comes the test, works on my machine.

test/discr/discr_mappings_test.py
```diff
@@ -345,5 +346,19 @@ def test_collocation_interpolation_identity():
...
 
 
+@skip_if_no_cuda
+def test_collocation_cuda():
+    rect = odl.Rectangle([0, 0], [1, 1])
+    part = odl.uniform_partition_fromintv(rect, [4, 2])
+    space = odl.FunctionSpace(rect)
+    dspace = odl.CudaRn(part.size)
+
+    coll_op = PointCollocation(space, part, dspace, order='F')
```
adler-j: Why order='F'?

adler-j:
Minor note, otherwise looks fine.

test/discr/discr_mappings_test.py
kohr-h: Hm, coincidence I guess. Wasn't Fortran ordering the natural thing on CUDA spaces?

adler-j: Yes, but I think we should stay with the defaults until such a time that we change that.

kohr-h:
Found another CUDA related bug along the way, triggered by the new test. Merging after Travis is happy.

kohr-h:
And yet another bug. Now we ravel before assignment.

PR #362, opened 2016-04-24
BUG: fix error due to direct usage of tuple in 1d meshgrids

### LINKED ISSUE #343: Non-obvious error messages from vectorization/meshgrid
I have been lagging on following the updates on odl the last months, but when I tried to run my old code I got the strange error message
```ValueError: unable to create an element of uniform_discr([0.0, 0.0], [256.0, 256.0], [256, 256]) from FunctionSpace(Rectangle([0.0, 0.0], [256.0, 256.0]))Vector: float() argument must be a string or a number, not 'FunctionSpaceVector' ```
For me this is a counter-intuitive error message, but after some help from @adler-j it turned out to be due to that when the discretization calls the continuous function, it does so with a meshgrid (which I could not handle). Hence with a simple ```vectorized=False``` the problem was solved. However, it is not obvious from the error message that that is what is wrong.

I also had trouble with a 1D case. A minimal working example is
```
def my_cos(x):
    return np.cos(x * 2)

dom = odl.FunctionSpace(odl.Interval(-1, 1))
f = dom.element(my_cos)
disc_domain = odl.uniform_discr_fromspace(dom, 10, exponent=2.0)
f_disc = disc_domain.element(f)
```
Again, a ```vectorized = False``` in the continuous definition solves the problem. But in this case there should, in principle, be no problem with having vectorized input. Another fix is, for example, to redefine the function to ```return np.cos(x[0] * 2)```. But then the function cannot take scalars as input, which one would maybe expect.

So:
1. Should ```vectorized = True``` or ```False``` be default?
2. Is there a way to handle the 1D-case?

### PR DESCRIPTION
**Pull request by kohr-h**
Fixes #343

### DIFF
--- a/odl/space/fspace.py
+++ b/odl/space/fspace.py
@@ -340,6 +340,9 @@ class FunctionSetVector(Operator):
         if is_valid_input_meshgrid(x, ndim):
             out_shape = out_shape_from_meshgrid(x)
             scalar_out = False
+            # Avoid operations on tuples like x * 2 by casting to array
+            if ndim == 1:
+                x = np.asarray(x)
         elif is_valid_input_array(x, ndim):
             x = np.asarray(x)
             out_shape = out_shape_from_array(x)
--- a/odl/space/fspace.py
+++ b/odl/space/fspace.py
@@ -342,7 +342,7 @@ class FunctionSetVector(Operator):
             scalar_out = False
             # Avoid operations on tuples like x * 2 by casting to array
             if ndim == 1:
-                x = np.asarray(x)
+                x = x[0][None, ...]
         elif is_valid_input_array(x, ndim):
             x = np.asarray(x)
             out_shape = out_shape_from_array(x)

### CODE REVIEW

File: odl/space/fspace.py [hunk 1]
```diff
@@ -340,6 +340,9 @@ def __call__(self, x, out=None, **kwargs):
         if is_valid_input_meshgrid(x, ndim):
             out_shape = out_shape_from_meshgrid(x)
             scalar_out = False
+            # Avoid operations on tuples like x * 2 by casting to array
+            if ndim == 1:
+                x = np.asarray(x).ravel()
```
adler-j: Why would we need ravel in this case?

adler-j:
Does this make code like this behave differently in 1d and nd?

```python
def sum_of_coordinates(x):
   return sum(x) 
```

we may want to put some thought into it in that case.

kohr-h:
It wouldn't if we wouldn't ravel.

I thought initially that it would be good to get rid of the extra dimension, but probably it's good to keep until the end.

kohr-h:
I removed the ravel, now everything is fine. There was a failing doctest triggering something similar as in your example. Now both versions work (and the case in #343).

adler-j:
Current implementation creates a copy, we could do:

```python
x = x[0][None, ...]
```

for a minimal efficiency gain, else looks nice.

kohr-h:
You're right, I thought NumPy would be slightly smarter here. If the space is large, this may not even be minimal. I'll change that. Merge after that?

adler-j:
Merge!

PR #364, opened 2016-04-25
ENH: add a fbp example

adler-j:
A simple example of implementing FBP in ODL, shows both syntax and is probably requested by several users.

### DIFF
--- a/examples/tomo/filtered_backprojection_parallel_2d.py
+++ b/examples/tomo/filtered_backprojection_parallel_2d.py
@@ -15,52 +15,59 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Example using the ray transform with 2d parallel beam geometry."""
+"""
+Example using a filtered backprojection in 2d using the ray transform and a
+ramp filter. The ramp filter is implemented in fourier space.
 
-# Imports for common Python 2/3 codebase
-from __future__ import print_function, division, absolute_import
-from future import standard_library
-standard_library.install_aliases()
+See https://en.wikipedia.org/wiki/Radon_transform#Inversion_formulas for
+more information.
+"""
 
 import numpy as np
 import odl
 
 
+# --- Set-up of geometry of problem --- #
+
+
 # Discrete reconstruction space: discretized functions on the rectangle
 # [-20, 20]^2 with 300 samples per dimension.
 reco_space = odl.uniform_discr(
-    min_corner=[-20, -20], max_corner=[20, 20], nsamples=[200, 200],
+    min_corner=[-20, -20], max_corner=[20, 20], nsamples=[300, 300],
     dtype='float32')
 
+# Angles: uniformly spaced, n = 1000, min = 0, max = pi
+angle_partition = odl.uniform_partition(0, np.pi, 1000)
+
+# Detector: uniformly sampled, n = 500, min = -30, max = 30
+detector_partition = odl.uniform_partition(-30, 30, 500)
+
 # Make a parallel beam geometry with flat detector
-# Angles: uniformly spaced, n = 360, min = 0, max = 2 * pi
-angle_partition = odl.uniform_partition(0, 2 * np.pi, 360)
-# Detector: uniformly sampled, n = 558, min = -30, max = 30
-detector_partition = odl.uniform_partition(-30, 30, 2000)
 geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+
+# --- Create FilteredBackProjection (FBP) operator --- #
+
 
 # ray transform aka forward projection. We use ASTRA CUDA backend.
 ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+
 # fourier transform in detector direction
 fourier = odl.trafos.FourierTransform(ray_trafo.range, axes=[1])
 
-# Create filter
-theta_max = detector_partition.extent() / 2.0
+# Create ramp in the detector direction
+ramp_function = fourier.range.element(lambda x: 0*x[0] + np.abs(x[1]))
 
-
-def window(x):
-    # Select cutoff value
-    cutoff = 0.7
-    nonzero = np.abs(x) < theta_max * cutoff
-    return nonzero * (1 + np.cos(np.abs(x) / (theta_max / cutoff)))
-
-
-ramp_function = fourier.range.element(lambda x: 0*x[0] + np.abs(x[1]) * window(x[1]))
+# Create ramp filter via the convolution formula with fourier transforms
 ramp_filter = fourier.inverse * ramp_function * fourier
 
+# Create filtered backprojection by composing the backprojection (adjoint) with
+# the ramp filter. Also apply a scaling.
+fbp = ray_trafo.adjoint * ramp_filter / (2 * np.pi)
 
-# Create filtered backprojection
-fbp = ray_trafo.adjoint * ramp_filter
+
+# --- Show some examples --- #
+
 
 # Create a discrete Shepp-Logan phantom (modified version)
 phantom = odl.util.phantom.shepp_logan(reco_space, modified=True)
@@ -68,15 +75,11 @@ phantom = odl.util.phantom.shepp_logan(reco_space, modified=True)
 # Create projection data by calling the ray transform on the phantom
 proj_data = ray_trafo(phantom)
 
-# Back-projection can be done by simply calling the adjoint operator on the
-# projection data (or any element in the projection space).
-backproj = ray_trafo.adjoint(proj_data)
-
 # Calculate filtered backprojection of data
 fbp_reconstruction = fbp(proj_data)
 
 # Shows a slice of the phantom, projections, and reconstruction
 phantom.show(title='Phantom')
 proj_data.show(title='Projection data (sinogram)')
-backproj.show(title='Back-projected data')
 fbp_reconstruction.show(title='Filtered backprojection')
+(phantom - fbp_reconstruction).show(title='Error')
--- a/examples/tomo/filtered_backprojection_parallel_2d.py
+++ b/examples/tomo/filtered_backprojection_parallel_2d.py
@@ -27,7 +27,7 @@ import numpy as np
 import odl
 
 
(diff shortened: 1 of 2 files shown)


examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+# [-20, 20]^2 with 300 samples per dimension.
+reco_space = odl.uniform_discr(
+    min_corner=[-20, -20], max_corner=[20, 20], nsamples=[200, 200],
+    dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+# Angles: uniformly spaced, n = 360, min = 0, max = 2 * pi
+angle_partition = odl.uniform_partition(0, 2 * np.pi, 360)
+# Detector: uniformly sampled, n = 558, min = -30, max = 30
+detector_partition = odl.uniform_partition(-30, 30, 2000)
```
adler-j: changed size by misstake

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using the ray transform with 2d parallel beam geometry."""
```
adler-j: copy paste

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+    nonzero = np.abs(x) < theta_max * cutoff
+    return nonzero * (1 + np.cos(np.abs(x) / (theta_max / cutoff)))
+
+
+ramp_function = fourier.range.element(lambda x: 0*x[0] + np.abs(x[1]) * window(x[1]))
+ramp_filter = fourier.inverse * ramp_function * fourier
+
+
+# Create filtered backprojection
+fbp = ray_trafo.adjoint * ramp_filter
```
kohr-h: A bit of explanation of what this does would be cool

adler-j: Sure that would indeed be needed

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+# ray transform aka forward projection. We use ASTRA CUDA backend.
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+# fourier transform in detector direction
+fourier = odl.trafos.FourierTransform(ray_trafo.range, axes=[1])
+
+# Create filter
+theta_max = detector_partition.extent() / 2.0
+
+
+def window(x):
```
kohr-h: One line about this function? Is it a Fourier space filter? What do the parameters do? Does the filter have a name?

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+    # Select cutoff value
+    cutoff = 0.7
+    nonzero = np.abs(x) < theta_max * cutoff
+    return nonzero * (1 + np.cos(np.abs(x) / (theta_max / cutoff)))
```
kohr-h: Sure that the parentheses are right? Just wondering because of the division in the denominator.

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+# Make a parallel beam geometry with flat detector
+# Angles: uniformly spaced, n = 360, min = 0, max = 2 * pi
+angle_partition = odl.uniform_partition(0, 2 * np.pi, 360)
+# Detector: uniformly sampled, n = 558, min = -30, max = 30
+detector_partition = odl.uniform_partition(-30, 30, 2000)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# ray transform aka forward projection. We use ASTRA CUDA backend.
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+# fourier transform in detector direction
```
kohr-h: Capitalize "ray" and "fourier"

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+
+def window(x):
+    # Select cutoff value
+    cutoff = 0.7
+    nonzero = np.abs(x) < theta_max * cutoff
+    return nonzero * (1 + np.cos(np.abs(x) / (theta_max / cutoff)))
+
+
+ramp_function = fourier.range.element(lambda x: 0*x[0] + np.abs(x[1]) * window(x[1]))
+ramp_filter = fourier.inverse * ramp_function * fourier
```
adler-j: This could also use some explanation

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def window(x):
+    # Select cutoff value
+    cutoff = 0.7
+    nonzero = np.abs(x) < theta_max * cutoff
+    return nonzero * (1 + np.cos(np.abs(x) / (theta_max / cutoff)))
+
+
+ramp_function = fourier.range.element(lambda x: 0*x[0] + np.abs(x[1]) * window(x[1]))
```
kohr-h: Playing autopep8 now: `0 * x[0]`

adler-j: There has to be a way of not needing this BTW, perhaps automatic broadcasting is something we should implement?

kohr-h:
Maybe we should, in the future, add optical helps to separate example files into sections, e.g.

```python
# --- Set-up of reconstruction space --- #

# some code

# --- Definition of the forward operator --- #

# some more code
```
This helps digesting the contents and, as a nice side effect, also appears as markers in the Spyder file summary (more relevant to longer files, though).

Anyway, nice idea, and it's good to show that using FBP is that simple.

examples/tomo/filtered_backprojection_parallel_2d.py
kohr-h: I'm not quite sure how to do this. It does the automatic broadcasting if you "do enough" to reach the required shape. One way I could think of is to check the shape which comes out, compare it to the required shape, and if the output shape is a "subset", we could fill in the zero-addition broadcasting operations in the "gaps".

adler-j: Yeah,in this case we have a (n, m) space, and it gets a (1, m) input. Then this should "obviously" be broadcast to a (n, m) array.

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,82 @@
...
+# Angles: uniformly spaced, n = 360, min = 0, max = 2 * pi
+angle_partition = odl.uniform_partition(0, 2 * np.pi, 360)
+# Detector: uniformly sampled, n = 558, min = -30, max = 30
+detector_partition = odl.uniform_partition(-30, 30, 2000)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# ray transform aka forward projection. We use ASTRA CUDA backend.
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+# fourier transform in detector direction
+fourier = odl.trafos.FourierTransform(ray_trafo.range, axes=[1])
```
adler-j: using axes here is technically correct, but can be omitted. Perhaps skip it too make this easier to read?

kohr-h: I would keep it - perhaps with a more detailed explanation like "Taking the Fourier transform along the detector axis (1), not along the angles (0)", then it's perfectly clear and makes sense mathematically.

adler-j: I'll get that done.

adler-j:
Fixed all comments, merge?

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,85 @@
...
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+
+# --- Create FilteredBackProjection (FBP) operator --- #
+
+
+# ray transform aka forward projection. We use ASTRA CUDA backend.
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+
+# fourier transform in detector direction
```
kohr-h: Capitalize "ray"

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,85 @@
...
+detector_partition = odl.uniform_partition(-30, 30, 500)
+
+# Make a parallel beam geometry with flat detector
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+
+# --- Create FilteredBackProjection (FBP) operator --- #
+
+
+# ray transform aka forward projection. We use ASTRA CUDA backend.
```
kohr-h: Capitalize "fourier", write "We use the ASTRA CUDA backend."

examples/tomo/filtered_backprojection_parallel_2d.py
```diff
@@ -0,0 +1,85 @@
...
+
+See https://en.wikipedia.org/wiki/Radon_transform#Inversion_formulas for
+more information.
+"""
+
+import numpy as np
+import odl
+
+
+# --- Set-up of geometry of problem --- #
```
kohr-h: of the geometry for the problem

kohr-h:
Some minor things in the comments, after that good to go!

adler-j:
Fixed comments, merging after travis etc

PR #368, opened 2016-04-26
ENH: change 3d shepp-logan to match 2d

adler-j:
Very minor fix. Changes defintion of 3d shepp logan to look exactly like the 2d version when looking at the middle slice. This is more "as expected".


kohr-h:
Looking good

Issue #295, opened 2016-03-08
Unclear meaning of starting angle in geometries


kohr-h:
From the discussion on the `issue-87__conebeam` branch:

> We're not clear on how we interpret the initial positions / vectors / axes. Are they associated
>     with the _start value_ of the angles, i.e. `angles=motion_params.begin` or with `angles=0`?
>     The current implementation does the latter, which I think is debatable. What is more practical?
>     If one wants to use an angle interval of `[-pi/2, pi/2]`, is it better to have the user give the state
>     of the geometry at angle 0 or at `-pi/2`? This seems to depend on the application.

Not exactly sure what would be the best. We could default to `motion_params.begin` and allow users to specify a point-value pair if the value at a different point shall be fixed. Would add some complexity though, since we would need to store both.

adler-j:
I feel that the current system is the only logical solution. In any application, you will in general have a fixed "reference position" of your system which defines the coordinate system. Typically this can be straight up, or down. In this case, we can give the initial detector axes etc in this system (by defining it as angle=0). This way, you can typically let the detector axes be [1, 0, 0] and [0, 1, 0], or something equally trivial. If we were to change this to be given by `angles.begin`, users would need to recalculate the detector position for each new data set.

Since this is basically the only use case for these options (anyone working with artificial data might as well go for the defaults in 99% of cases), changing would add work to users at debatable gain.

kohr-h:
> I feel that the current system is the only logical solution. In any application, you will in general have a fixed "reference position" of your system which defines the coordinate system.

Correct, so then we should not use "init" in the names but rather something like "ref" since it's a (somewhat arbitrary) reference point or system of axes. And the docstrings don't say a word about this, they should be much clearer since it's a crucial distinction.

adler-j:
I agree, `ref` would be a nice name for these, and the doc needs some help.

kohr-h:
This is still relevant, and connected to #784.

adler-j:
The mother of these issues is #183, which really needs to be solved.

adler-j:
Was this fixed by #968?

kohr-h:
This is settled now for `angle=0` as the reference.

PR #369, opened 2016-04-26
Angles property

adler-j:


odl/tomo/geometry/conebeam.py
```diff
@@ -156,6 +156,11 @@ def pitch_offset(self):
         """Vertical offset at ``angle=0``"""
         return self._pitch_offset
 
+    @property
+    def angles(self):
+        """The discrete angles given in this geometry"""
```
kohr-h: end with a dot

test/tomo/geometry/geometry_test.py
```diff
@@ -154,6 +156,10 @@ def test_fanflat():
     assert all_almost_equal(rot_mat.dot([1, 0]), [0, 1])
     assert all_almost_equal(rot_mat.dot([0, 1]), [-1, 0])
 
+    # check str and repr works without crashing and return something
```
kohr-h: work

test/tomo/geometry/geometry_test.py
```diff
@@ -154,6 +156,10 @@ def test_fanflat():
...
+    assert str(geom)
+    assert repr(geom)
```
kohr-h: Good to check this

kohr-h:
Can you make an attempt at finding the locations where this can be used to simplify existing code? Perhaps a "find in files -> motion_grid\\.points" would reveal all places.

Nit-picky stuff, otherwise good to go.

odl/tomo/geometry/parallel.py
```diff
@@ -178,25 +178,36 @@ def __init__(self, apart, dpart, **kwargs):
...
-                         det_init_pos=det_init_pos)
+                         det_init_pos=self.det_init_pos)
 
         if self.motion_partition.ndim != 1:
             raise ValueError('angle partition has dimension {}, expected 1.'
                              ''.format(self.motion_partition.ndim))
 
+    @property
+    def det_init_pos(self):
+        """The position of the detector reference point at angle=0"""
```
kohr-h: don't forget the dot :-)

kohr-h: Side remark: we seem to tend towards the angle=0 solution to the problem raised in #295. Shall we definitely settle for that?

PR #371, opened 2016-04-27
Updates to test skipping and fixtures

adler-j:
Did some fixes to test skipping etc, the number of tests should now be the same across all platforms and with any combination of optional packages, any non-usable package will properly show as `skipped` instead of staying silent.

### DIFF
--- a/test/discr/lp_discr_test.py
+++ b/test/discr/lp_discr_test.py
@@ -29,7 +29,7 @@ import numpy as np
 import odl
 from odl.discr.lp_discr import DiscreteLp
 from odl.util.testutils import (almost_equal, all_equal, all_almost_equal,
-                                skip_if_no_cuda)
+                                never_skip, skip_if_no_cuda)
 
 
 def _array(fn):
@@ -66,11 +66,20 @@ def _vectors(fn, n=1):
 # Simply modify exp_params to modify the fixture
 exp_params = [2.0, 1.0, float('inf'), 0.5, 1.5]
 exp_ids = [' p = {} '.format(p) for p in exp_params]
-exp_fixture = pytest.fixture(scope="module", ids=exp_ids, params=exp_params)
 
 
-@exp_fixture
+@pytest.fixture(scope="module", ids=exp_ids, params=exp_params)
 def exponent(request):
+    return request.param
+
+
+impl_params = [never_skip('numpy'),
+               skip_if_no_cuda('cuda')]
+impl_ids = [' impl = {} '.format(p.args[1]) for p in impl_params]
+
+
+@pytest.fixture(scope="module", ids=impl_ids, params=impl_params)
+def impl(request):
     return request.param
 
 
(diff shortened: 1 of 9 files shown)


test/largescale/space/fn_base_slow_test.py
```diff
@@ -134,31 +150,45 @@ def test_setitem(fn):
         assert x[index] == -index
 
 
-@pytest.mark.xfail(reason='Expected to fail since inner scaling is not public')
```
adler-j: These xfails have been removed, now we test that the weighting is correct, but assume that the constant weighting is used, should be improved but works for all current test cases.

test/largescale/trafos/fourier_slow_test.py
```diff
@@ -69,24 +68,30 @@ def _vectors(discr, num=1):
...
-                                dtype='complex')]
-
-dom_ids = [' {!r} '.format(dom) for dom in dom_params]
-
-
 impl_params = [never_skip('numpy'), skip_if_no_pyfftw('pyfftw')]
 impl_ids = [" impl = '{}'".format(p.args[1]) for p in impl_params]
 
 
+# bug in pytest (ignores pytestmark) forces us to do this this
```
adler-j: Bloody irritating bug this, but can be sidestepped like this

adler-j:
This also fixes a bug on master where the skip if no largescale is missing, thus making the tests take a minute instead of 5 seconds.

test/discr/lp_discr_test.py
```diff
@@ -66,14 +66,23 @@ def _vectors(fn, n=1):
...
-exp_fixture = pytest.fixture(scope="module", ids=exp_ids, params=exp_params)
 
 
-@exp_fixture
+@pytest.fixture(scope="module", ids=exp_ids, params=exp_params)
 def exponent(request):
     return request.param
 
 
+impl_params = [never_skip('numpy'),
```
kohr-h: Nice and intuitive

test/discr/lp_discr_test.py
```diff
@@ -761,79 +756,18 @@ def test_norm_rectangle(exponent):
...
-    discr = DiscreteLp(odl.FunctionSpace(rect), part, dspace,
-                       exponent=exponent)
-
-    if exponent == float('inf'):
-        assert discr.one().norm() == 1
-    else:
-        assert almost_equal(discr.one().norm(),
-                            (rect.volume) ** (1 / exponent))
-
+        pytest.xfail('inf-norm not implemented in CUDA')
```
kohr-h: This is really much better than having to write a separate copy-and-paste test for CUDA.

test/largescale/space/fn_base_slow_test.py
kohr-h: Well yeah, it's the only one implemented :-). TODO, but not super high prio

test/largescale/trafos/fourier_slow_test.py
kohr-h: Bug report in pytest?

test/space/ntuples_test.py
```diff
@@ -1609,9 +1609,43 @@ def other_dist(x, y):
...
+                     'mod',
+                     'modf',
+                     'fmod',
+                     'logaddexp2',
+                     'logaddexp',
+                     'hypot',
+                     'arctan2',
+                     'floor',
+                     'ceil']):
+        # Skip real only methods for complex
```
kohr-h: Huh? Looks like you do the opposite.

kohr-h:
Found just one minor thing. Number of tests went up from ~1500 to almost 2000. Good for people who like numbers :-)

Looks very good otherwise.

kohr-h:
Oh, and :+1: for removing more code than adding

test/space/ntuples_test.py
adler-j: perhaps not a good formulation but i want to say: "skip methods that only work on real numbers if dtype is complex". Will update that.

test/largescale/trafos/fourier_slow_test.py
adler-j: https://github.com/pytest-dev/pytest/issues/1540

PR #373, opened 2016-04-28
ENH: add odl.test() call, update readme

adler-j:
The old readme made no sense, this new one works perfectly.


kohr-h:
Makes sense.

PR #376, opened 2016-05-03
ENH: improve show axes in case of nan or inf

adler-j:
Add handling of some corner cases where we got nonsense values, now tries to "make the best of it". Also unified complex and real handling in these cases.

### DIFF
--- a/odl/util/graphics.py
+++ b/odl/util/graphics.py
@@ -31,6 +31,31 @@ from odl.util.utility import is_real_dtype
 __all__ = ('show_discrete_data',)
 
 
+def _safe_minmax(values):
+    """Calculate min and max of array with guards for nan and inf."""
+
+    # Nan and inf guarded min and max
+    minval = np.min(values[np.isfinite(values)])
+    maxval = np.max(values[np.isfinite(values)])
+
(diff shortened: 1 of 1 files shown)


odl/util/graphics.py
```diff
@@ -31,6 +31,36 @@
...
+    """Calculate min and max of array with guards for nan and inf."""
+    # Nan guarded min and max
+    minval = np.nanmin(values)
+    maxval = np.nanmax(values)
+
+    # Get something sensible in case of infinite values
+    if minval == -np.inf:
+        minval = np.nanmin(values[values > -np.inf])
+    if maxval == np.inf:
+        maxval = np.nanmin(values[values < np.inf])
```
kohr-h: This can be compressed into
```python
minval = np.nanmin(values[np.isfinite(values)])
maxval = np.nanmax(values[np.isfinite(values)])
```

kohr-h:
One remark, otherwise good to go.

odl/util/graphics.py
adler-j: Isfinite also skips `NaN` values, so can change to `np.min` will do that instead.

PR #377, opened 2016-05-03
API: change Divergence init pattern

### LINKED ISSUE #375: Nonsensical init pattern for Divergence operator
Currently, we do the following:
```python
>>> space = odl.uniform_discr([0, 0], [1, 1], (10, 10))
>>> div = odl.Divergence(space)
>>> div.domain
ProductSpace(uniform_discr([0.0, 0.0], [1.0, 1.0], [10, 10]), 2)
>>> div.range
uniform_discr([0.0, 0.0], [1.0, 1.0], [10, 10])
```
I think this is very counter-intuitive and convenience at the wrong place. The obvious choice is to create an operator with its domain only if the range can be inferred, never the other way around.

kohr-h:
As discussed in #375, here are the proposed changes to the differential operators.

Changes in detail:
- move finite_diff, PartialDerivative, Gradient, Divergence, Laplacian
  to tensor_ops.py
- make PartialDerivative, Gradient, Divergence, Laplacian subclass of
  PointwiseTensorFieldOperator
- implement adjoint of PartialDerivative and add a test for it
- allow Gradient and Divergence to be initialized with domain or range
  (or both); add doctests for those init patterns
- use global tuple of supported difference schemes instead of a hard-coded tuple

Fixes #375


adler-j:
> - move finite_diff, PartialDerivative, Gradient, Divergence, Laplacian
  to tensor_ops.py

Disagree with this change. They should be moved to their own file instead. They are simply "too much" by now.

> - make PartialDerivative, Gradient, Divergence, Laplacian subclass of
  PointwiseTensorFieldOperator

:+1:

> - implement adjoint of PartialDerivative and add a test for it

:+1:

> - allow Gradient and Divergence to be initialized with domain or range
  (or both); add doctests for those init patterns

:+1:

> - use global tuple of supported difference schemes instead of a hard-coded tuple

:+1:

Fixes #375

odl/discr/tensor_ops.py
```diff
@@ -112,6 +117,479 @@ def base_space(self):
...
+                raise TypeError('range {!r} is not a ProductSpace instance.'
+                                ''.format(range))
+            domain = range[0]
+
+        if range is None:
+            if not isinstance(domain, DiscreteLp):
+                raise TypeError('domain {!r} is not a `DiscreteLp` '
+                                'instance.'.format(domain))
+            range = ProductSpace(domain, domain.ndim)
+
```
adler-j: What if both domain and rage is not none, then we have no checks?

odl/discr/tensor_ops.py
```diff
@@ -112,6 +117,479 @@ def base_space(self):
...
+        >>> grad_op3.range == ran
+        True
+        """
+        if domain is None and range is None:
+            raise ValueError('either domain or range must be specified.')
+
+        if domain is None:
+            if not isinstance(range, ProductSpace):
+                raise TypeError('range {!r} is not a ProductSpace instance.'
+                                ''.format(range))
```
adler-j: Also verify it is a product space of the domain, if given

odl/discr/tensor_ops.py
kohr-h: The actual checks happen in the parent class, except for the `DiscreteLp` thing. Otherwise I only test to make sure I can use the attributes I need.

odl/discr/tensor_ops.py
kohr-h: Again, that's checked in the parent.

kohr-h:
>> move finite_diff, PartialDerivative, Gradient, Divergence, Laplacian to tensor_ops.py
> 
> Disagree with this change. They should be moved to their own file instead. They are simply "too much" by now.

Regarding that more vector field stuff may be coming in, I agree. The original place was no good either, for that matter. How about `diff_ops.py` (like differential)? Depending on how much we implement, a subpackage has to be taken into consideration. But not now.

adler-j:
Agree with `diff_ops.py`.

kohr-h:
Changed the place. Merging if the tests run through.

PR #380, opened 2016-05-04
Issue 379  diagonal precon

### LINKED ISSUE #325: Move oputils to util
This would allow us to call the oputils things as

```
odl.util.MYUTIL
```

instead of the clumsy and less obvious

```
odl.operator.oputils
```

adler-j:
* Add preconditioning to chambolle pock
* Add example using diagonal preconditioner
* Changed `diagonal_operator` into `DiagonalOperator`
* Extended `&` syntax of partials to accept `Partial & callable`
* Expose `power_method_opnorm` and others to main namespace, closes #325 

Try the example, the difference is extremely big.


odl/solvers/advanced/chambolle_pock.py
```diff
@@ -201,10 +214,18 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
...
+        else:
+            dual_tmp = y + sigma * preconditioner_dual(op(x_relax))
+        proximal_dual(sigma)(dual_tmp, out=y)
 
         # Gradient descent in the primal variable x
-        proximal_primal(tau)(x + (- tau) * op_adjoint(y), out=x)
+        if preconditioner_primal is None:
+            primal_tmp = x + (- tau) * op_adjoint(y)
+        else:
+            x + (- tau) * preconditioner_primal(op_adjoint(y))
```
adler-j: obviously missing

examples/solvers/chambolle_pock_deconvolve.py
```diff
@@ -155,17 +155,17 @@ def adjkernel(x):
 
 
 # Estimated operator norm, add 10 percent to ensure ||K||_2^2 * sigma * tau < 1
-op_norm = 1.1 * odl.operator.oputils.power_method_opnorm(op, 5)
+op_norm = 1.1 * odl.power_method_opnorm(op, 5)
```
kohr-h: Ah, good.

examples/solvers/chambolle_pock_preconditioner_denoising.py
```diff
@@ -0,0 +1,180 @@
...
+identity mapping id.
+
+In order to use the Chambolle-Pock solver, we have to create the column
+operator K, choose a starting point x, create the proximal operator for G,
+create the proximal operator for the convex conjugate of F, choose the
+step sizes tau and sigma such that tau sigma ||K||_2^2 < 1, and set the
+total number of iterations.
+
+For details see :ref:`chambolle_pock`, :ref:`proximal_operators`, and
+references therein.
```
kohr-h: This is copy & paste, right? No word about pre-conditioning. Could be greatly shortened.

adler-j: Totally copy and paste. Shortening should probably be done in conjunction with moving this to the documentation.

examples/solvers/chambolle_pock_preconditioner_denoising.py
```diff
@@ -0,0 +1,180 @@
...
+from X to Y with induce norm ||K||. The (primal) minimization problem we
+want to solve is
+
+    min_{x in X} F(K x) + G(x)
+
+where the proper, convex, lower-semicontinuous functionals
+F : Y -> [0, +infinity] and G : X -> [0, +infinity] are given
+by an l2-data fitting term regularized by isotropic total variation
+
+    F(K x) = 1/2 ||x - g||_2^2 + lam || |grad(x)| ||_1
```
kohr-h: Here it should say `F(x)` not `F(K x)`

odl/operator/__init__.py
```diff
@@ -34,3 +34,5 @@
 __all__ += pspace_ops.__all__
 
 from . import oputils
```
kohr-h: This line can be skipped.

examples/solvers/chambolle_pock_preconditioner_denoising.py
```diff
@@ -0,0 +1,180 @@
...
+op_norm_identity = 1.0
+op_norm_gradient = 1.5 * odl.power_method_opnorm(gradient, 100, noisy)
+print('Operator norms, I: {}, gradient: {}'.format(op_norm_identity,
+                                                   op_norm_gradient))
+
+# --- Run algorithms without preconditioner
+
+niter = 50
+
+# Create a function to save the partial errors
```
adler-j: Not quite sure if there's a better way to do this. We could for example use a `StorePartial`, but that would store the whole result.

odl/operator/pspace_ops.py
```diff
@@ -883,28 +891,135 @@ def adjoint(self):
...
+
+        For example, if A and B are operators
+
+            [[A, 0],
+             [0, B]]
+
+        The adjoint is given by:
+
+            [[A^T, 0],
+             [0, B^T]]
```
kohr-h: `A^*, B^*`

odl/operator/pspace_ops.py
```diff
@@ -883,28 +891,135 @@ def adjoint(self):
...
-        Range of the operator. If not provided, it is tried to be
-        inferred from the operators. This requires each **row**
-        to contain at least one operator.
     """
 
-    indices = [range(len(operators)), range(len(operators))]
-    shape = (len(operators), len(operators))
-    op_matrix = sp.sparse.coo_matrix((operators, indices), shape)
+    def __init__(self, *operators, **kwargs):
+        """Initialize a DiagonalOperator
```
kohr-h: dot at the end

odl/operator/pspace_ops.py
```diff
@@ -883,28 +891,135 @@ def adjoint(self):
...
+    def operators(self):
+        """A tuple of sub-operators"""
+        return self._operators
+
+    def __getitem__(self, index):
+        """Return an operator by index"""
+        return self.operators[index]
+
+    def derivative(self, point):
+        """The derivative operator
```
kohr-h: dot

odl/operator/pspace_ops.py
```diff
@@ -883,28 +891,135 @@ def adjoint(self):
...
+        ProductSpaceOperator.derivative
+        """
+        point = self.domain.element(point)
+
+        derivs = [op.derivative(p) for op, p in zip(self.operators, point)]
+        return DiagonalOperator(derivs, self.domain, self.range)
+
+    @property
+    def adjoint(self):
+        """The adjoint operator
```
kohr-h: dot

odl/operator/__init__.py
adler-j: That is a rather general statement about these. In general they should all be removable, but we may need to change to relational paths `from . import XXX` instead then. Imo we should make an issue of this. This would also help making the global namespace smaller.

odl/solvers/advanced/chambolle_pock.py
```diff
@@ -100,6 +100,14 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
     partial : `Partial`, optional
         If not `None` the `Partial` instance(s) are executed in each
         iteration, e.g. plotting each iterate. Default: `None`
+    preconditioner_primal : `Operator`, optional
```
kohr-h: This name is too long IMO. "itioner" could be scrapped without loss.

odl/solvers/advanced/chambolle_pock.py
```diff
@@ -100,6 +100,14 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
...
+        Preconditioning operator for the primal variable, needs an
+        `Operator.inverse` method. The operator needs to be symmetric and
+        positive definite. Default: No preconditioning
+    preconditioner_dual : `Operator`, optional
```
kohr-h: same here

odl/solvers/advanced/chambolle_pock.py
```diff
@@ -168,10 +176,15 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
 
     # Partial object
     partial = kwargs.pop('partial', None)
-    if not (partial is None or isinstance(partial, Partial)):
+    if not (partial is None or callable(partial)):
```
kohr-h: Error message doesn't fit any longer.

odl/solvers/advanced/chambolle_pock.py
kohr-h: In the primal? That would be in `LinearSpace`?

examples/solvers/chambolle_pock_preconditioner_denoising.py
kohr-h: I think it's totally okay to allow functions as callbacks in an algorithm and just define a way it's called. Apart from that, a `ResidualNormPartial` would be an option since it's one of the most common cases.

odl/operator/__init__.py
kohr-h: My remark was actually more technical. If you do `from .oputils import *`, then `oputils` is also imported to the namespace, so the subsequent line makes sense.

examples/solvers/chambolle_pock_preconditioner_denoising.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def print_diff(x):
+    error = (x-orig).norm()
+    differences.append(error)
+
+# Optional: pass partial objects to solver
+partial = (odl.solvers.ShowPartial() &
+           odl.solvers.PrintIterationPartial() &
+           print_diff)
```
kohr-h: Running the example, it's a bit hard to see what's happening when. Would be good to have a possibility to provide a figure title to the show partial. Would be a new issue then.

examples/solvers/chambolle_pock_preconditioner_denoising.py
```diff
@@ -0,0 +1,180 @@
...
+x = op.domain.zero()  # Starting point
+odl.solvers.chambolle_pock_solver(
+    op, x, tau=tau, sigma=sigma, proximal_primal=proximal_primal,
+    proximal_dual=proximal_dual, niter=niter, partial=partial,
+    preconditioner_dual=preconditioner_dual)
+
+# results
+plt.figure()
+plt.loglog(np.arange(niter), differences)
+plt.loglog(np.arange(niter), differences_precondtioned)
```
kohr-h: Some legend and a title would be great.

kohr-h:
> Add preconditioning to chambolle pock
:+1: 
> Add example using diagonal preconditioner
:+1: 
> Changed diagonal_operator into DiagonalOperator
:+1: 
> Extended & syntax of partials to accept Partial & callable
Somewhat closer to the spirit of SciPy callbacks :+1: 
> Expose power_method_opnorm and others to main namespace, closes #325
:+1:

examples/solvers/chambolle_pock_preconditioner_denoising.py
adler-j: `odl.solvers.ShowPartial(title='mytitle')` is valid ODL.

The issue would be that `ShowPartial` currently only supports kwargs, so

`odl.solvers.ShowPartial('mytitle')` is ill formed

adler-j: Added in  e9ad140

odl/operator/__init__.py
adler-j: Not quite sure what you mean, i cant get that too work. Short example:

```python
>>> from scipy import *
>>> scipy
NameError: name 'scipy' is not defined
```

odl/solvers/advanced/chambolle_pock.py
adler-j: It is missing

```
primal_tmp =
```

examples/solvers/chambolle_pock_preconditioner_denoising.py
adler-j: This is not the residual norm, but the error norm, only available with artificial data like this. Should perhaps add it anyway.

examples/solvers/chambolle_pock_preconditioner_denoising.py
kohr-h: Cool. I looked into the `ShowPartial` doc but didn't read all the way, so I overlooked that the `kwargs` are passed on. Probably the additional positional args passing is most natural.

odl/operator/__init__.py
kohr-h: No, that's different. What I mean is that the *relative* import `from .local_pkg import stuff` also imports `local_pkg` into the namespace. That's not true for absolute imports.

Basically we can remove all of the `from . import subpackage` things.

odl/solvers/advanced/chambolle_pock.py
kohr-h: Ah, ok

adler-j:
Fixed all comments (except the general one). Merge?

kohr-h:
In the doc: search-and-replace "chambolle-pock" -> "Chambolle-Pock". Also "chambolle pock" exists in the text.
Also ODL should be capitalized when not used as Python package name explicitly.

In math formulas, you can use matrices and align environments by using the `:nowrap:` directive and specifying the LaTeX environment by hand. Without it, Sphinx will guess an adequate one. That would enable a nicer formatting of the definition of `G(x)` and the result of `K(x)`

Otherwise I think I'm ok with this.

Issue #381, opened 2016-05-04
RectPartition.__getitem__ missing


kohr-h:
Grids and interval products can be indexed, partitions cannot. This should be added.

PR #386, opened 2016-05-09
ENH: add __getitem__ to RectPartition

kohr-h:
Initial PR. Seems to work fine with regular partitions. Remaining TODOs:

- [x] Tests
- [x] Subpartitions of uniform partitions don't always print themselves as uniform, probably due to numerical error. ~~If possible, this should be fixed.~~ WONTFIX
- [x] Indexing with one integer index `idx` should be implemented (also for `TensorGrid` - probably the latter could be significantly simplified) as equivalent to `partition[idx, ...]`.
- [x] Relates to the above: `Ellipsis` should be handled.

EDIT: Added 2 points to the list.
EDIT2: Probably the self-printing problem is not easily solvable.

Fixes: #381

### DIFF
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -689,80 +689,80 @@ class TensorGrid(Set):
         """
         return sparse_meshgrid(*self.coord_vectors)
 
-    def __getitem__(self, slc):
-        """Return ``self[slc]``.
+    def __getitem__(self, indices):
+        """Return ``self[indices]``.
 
         Parameters
         ----------
-        slc : `int` or slice
-            Negative indices and `None` (new axis) are not supported.
+        indices : `int` or slice
+            `None` (new axis) is not supported
 
         Examples
         --------
+        Indexing with integers along all axes produces an array (a point):
+
         >>> g = TensorGrid([-1, 0, 3], [2, 4], [5], [2, 4, 7])
         >>> g[0, 0, 0, 0]
         array([-1.,  2.,  5.,  2.])
+
+        Otherwise, a new TensorGrid is returned:
+
         >>> g[:, 0, 0, 0]
         TensorGrid([-1.0, 0.0, 3.0], [2.0], [5.0], [2.0])
         >>> g[0, ..., 1:]
         TensorGrid([-1.0], [2.0, 4.0], [5.0], [4.0, 7.0])
         >>> g[::2, ..., ::2]
         TensorGrid([-1.0, 3.0], [2.0, 4.0], [5.0], [2.0, 7.0])
+
+        Too few indices are filled up with an ellipsis from the right:
+
+        >>> g[0]
+        TensorGrid([-1.0], [2.0, 4.0], [5.0], [2.0, 4.0, 7.0])
+        >>> g[0] == g[0, :, :, :] == g[0, ...]
+        True
         """
-        slc_list = list(np.atleast_1d(np.s_[slc]))
-        if None in slc_list:
-            raise IndexError('Creation of new axes not supported.')
+        # Support indexing with fewer indices as indexing along the first
+        # corresponding axes. In the other cases, normalize the input.
+        if np.isscalar(indices):
+            indices = [indices, Ellipsis]
+        elif isinstance(indices, slice) or indices == Ellipsis:
+            indices = [indices]
 
-        try:
-            # All single indices
-            idx = np.array(slc_list).astype('int64', casting='safe')
-            if len(idx) < self.ndim:
-                raise IndexError('too few indices ({} < {}).'
-                                 ''.format(len(idx), self.ndim))
-            elif len(idx) > self.ndim:
-                raise IndexError('too many indices ({} > {}).'
-                                 ''.format(len(idx), self.ndim))
+        indices = list(indices)
+        if len(indices) < self.ndim and Ellipsis not in indices:
+            indices.append(Ellipsis)
 
-            return np.array([v[i] for i, v in zip(idx, self.coord_vectors)])
+        # Turn Ellipsis into the correct number of slice(None)
+        if Ellipsis in indices:
+            if indices.count(Ellipsis) > 1:
+                raise ValueError('cannot use more than one Ellipsis.')
 
-        except TypeError:
-            pass
+            eidx = indices.index(Ellipsis)
+            extra_dims = self.ndim - len(indices) + 1
+            indices = (indices[:eidx] + [slice(None)] * extra_dims +
+                       indices[eidx + 1:])
 
-        if Ellipsis in slc_list:
-            if slc_list.count(Ellipsis) > 1:
-                raise IndexError("Cannot use more than one ellipsis.")
-            if len(slc_list) == self.ndim + 1:  # Ellipsis without effect
-                ellipsis_idx = self.ndim
-                num_after_ellipsis = 0
-            else:
-                ellipsis_idx = slc_list.index(Ellipsis)
-                num_after_ellipsis = len(slc_list) - ellipsis_idx - 1
-            slc_list.remove(Ellipsis)
+        # Catch most common errors
+        if any(s.start == s.stop and s.start is not None or
+               s.start == n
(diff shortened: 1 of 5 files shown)


kohr-h:
Indexing with fewer indices now works similarly as in Numpy, in the sense that `part[i] is the same as part[i, ...]`, and the same is true for more than one and less than `ndim` index expressions. The same is true for the grids, strongly simplified the code along the way.

kohr-h:
Good for a review, I think. The problem with the inconsistent self-printing is hard to handle without manipulating the begin and end vectors, which I'm hesitant to do. Perhaps we need to live with this slight ugliness.

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
         """
         return sparse_meshgrid(*self.coord_vectors)
 
-    def __getitem__(self, slc):
-        """Return ``self[slc]``.
+    def __getitem__(self, indices):
+        """Return ``self[indices]``.
```
adler-j: Can you check that this is done across the board. Like the change.

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
...
 
         Parameters
         ----------
-        slc : `int` or slice
-            Negative indices and `None` (new axis) are not supported.
+        indices : `int` or slice
+            `None` (new axis) is not supported
```
adler-j: Perhaps dumb, but i think a word on "The index (dimension) that should be extracted" or something.

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
         """
         return sparse_meshgrid(*self.coord_vectors)
 
-    def __getitem__(self, slc):
-        """Return ``self[slc]``.
+    def __getitem__(self, indices):
+        """Return ``self[indices]``.
```
adler-j: We could use a sentence here on what this does, it may be nonobvious.

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
...
         Parameters
         ----------
-        slc : `int` or slice
-            Negative indices and `None` (new axis) are not supported.
+        indices : `int` or slice
+            `None` (new axis) is not supported
 
         Examples
         --------
+        Indexing with integers along all axes produces an array (a point):
```
adler-j: Great additions

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
...
+        TensorGrid([-1.0], [2.0, 4.0], [5.0], [2.0, 4.0, 7.0])
+        >>> g[0] == g[0, :, :, :] == g[0, ...]
+        True
+        """
+        # Support indexing with fewer indices as indexing along the first
+        # corresponding axes. In the other cases, normalize the input.
+        if np.isscalar(indices):
+            indices = [indices, Ellipsis]
+        elif (isinstance(indices, slice) or
+              (not isinstance(indices, np.ndarray) and indices == Ellipsis)):
```
adler-j: indices is Ellipsis

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
...
+
+        # Turn Ellipsis into the correct number of slice(None)
+        if Ellipsis in indices:
+            if indices.count(Ellipsis) > 1:
+                raise ValueError('cannot use more than one Ellipsis.')
+
+            eidx = indices.index(Ellipsis)
+            extra_dims = self.ndim - len(indices) + 1
+            indices = (indices[:eidx] + [slice(None)] * extra_dims +
+                       indices[eidx + 1:])
```
adler-j: All code down to here seems quite general, shouldnt we make a utility?

adler-j:
Since this improves the index handling to be quite similar to numpy, I'll have one comment on what we're lacking. Numpy treats tuple as a special case, namely:

```python
>>> a = np.zeros([2, 2])
>>> a[(0, 1)]
0.0
>>> a[[0, 1]]
array([[ 0.,  0.],
       [ 0.,  0.]])
```

That is, tuple represents per axis, while all other sequences represent in the first axis.

Currently (at least grid) does this:

```python
>>> g = odl.TensorGrid([-1, 1], [2, 3])
>>> g[(0, 1)]
array([-1.,  3.])
>>> g[[0, 1]]
array([-1.,  3.])
```

To match numpy, we would expect this behaviour

```python
>>> g = odl.TensorGrid([-1, 1], [2, 3])
>>> g[(0, 1)]
array([-1.,  3.])
>>> g[[0, 1]]
TensorGrid([-1, 1], [2, 3])
```

I think it would be nice if we followed this.

odl/discr/grid.py
```diff
@@ -689,80 +689,81 @@ def meshgrid(self):
...
+
+        # Catch most common errors
+        if any(s.start == s.stop and s.start is not None or
+               s.start == n
+               for s, n in zip(indices, self.shape) if isinstance(s, slice)):
+            raise ValueError('Slices with empty axes not allowed.')
+        if None in indices:
+            raise ValueError('creating new axes is not supported.')
```
adler-j: '<b>C</b>reating new axes is not supported. <b>Use ``insert``.</b>'

Perhaps? Im not sure, we soldom do this (add tips) but this could be one of those places.

odl/discr/partition.py
```diff
@@ -359,6 +359,61 @@ def __eq__(self, other):
         # Optimized version for exact equality
         return self.set == other.set and self.grid == other.grid
 
+    def __getitem__(self, indices):
+        """Return ``self[indices]``."""
```
adler-j: Could use a few rows of doc. Especially on how we handle the corner cases.

odl/discr/grid.py
kohr-h: Totally right, the current sentence should come second.

odl/discr/grid.py
kohr-h: Will do.

odl/discr/grid.py
kohr-h: Sure, this was the uncreative way of avoiding `==` comparison with an array.

odl/discr/grid.py
kohr-h: I was having a similar thought when copying around the code. I'll factor it out.

odl/discr/grid.py
kohr-h: Yeah, why not? Regarding capitalization: You're right, but then we should change that across the code. Currently most error messages start with a small letter.

odl/discr/partition.py
kohr-h: Okay.

odl/discr/grid.py
adler-j: As long as we use punctuation we should use capitalization, these kindoff go hand in hand for me.

kohr-h: Okay, but not in this PR then. I'll make a new issue (one of those low-hanging fruits :-) )

adler-j:
Review done. Except for the numpy compliance and minor stuff we're good. I'd try to get numpy compliance in since we already got it started.

kohr-h:
Done with all comments.

adler-j:
Does this support

```python
>>> g = odl.TensorGrid([-1, 1], [2, 3])
>>> g[[0, 1], [0, 1]]
TensorGrid([-1, 1], [2, 3])
>>> g[[0, 1], :]
TensorGrid([-1, 1], [2, 3])
>>> g[[0, 1], ...]
TensorGrid([-1, 1], [2, 3])
>>> g[[0, 1], 0]
TensorGrid([-1, 1], [2])
```

etc?

adler-j:
The above is technically not needed and I know I'm adding some work to you, but there are some dark corners of numpy indexing :D

kohr-h:
Hm, that would add one level of conversion to the helper. Perhaps it's doable. We don't want to go any further, though, since we need ascending entries in the resulting slice of the coordinate vectors, so fancy indexing is rather restricted.

adler-j:
Ok, lets go for this then.

PR #388, opened 2016-05-11
ENH: add inverse where missing to some operators, closes #164

### LINKED ISSUE #164: Inverse missing on many operators
When writing some tomography code, I decided to use FBP as a "prior", and thus added an `inverse` property to my `Projector`. It worked well, but when I did some arithmetic on the operator (multiply by vector), it turned out that this operator did not have an inverse (which it should have had). This is something we should improve.

adler-j:
Ran into this again, got it fixed:

* Reduce clunky syntax when creating inverses etc of left/right mult type operators
* Add inverse to vector multiplication operators

### DIFF
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -1346,7 +1346,8 @@ class OperatorLeftScalarMult(Operator):
         """
         if self._scalar == 0.0:
             raise ZeroDivisionError('{} not invertible.'.format(self))
-        return OperatorLeftScalarMult(self._op.inverse, 1.0 / self._scalar)
+
+        return self._op.inverse * (1.0 / self._scalar)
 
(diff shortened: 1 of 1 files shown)


odl/operator/operator.py
```diff
@@ -1346,7 +1346,8 @@ def inverse(self):
         """
         if self._scalar == 0.0:
             raise ZeroDivisionError('{} not invertible.'.format(self))
-        return OperatorLeftScalarMult(self._op.inverse, 1.0 / self._scalar)
+
+        return self._op.inverse * (1.0 / self._scalar)
```
adler-j: This was actually a bug before

kohr-h:
OK, cool.

PR #390, opened 2016-05-12
Revert "Issue 379  diagonal precon"

adler-j:
Reverts odlgroup/odl#380

Well basically I missed a term in the preconditioning paper which makes this more complicated. Reverting for now.


adler-j:
Basically the issue is that the expression involves the preconditoner also in the proximal. I missed this.

The example ran just fine without it, but when trying preconditioned tomography it failed miserably. I'm trying to find a way to solve this without having to write new proximals for each operator, but im not quite sure it is doable. Worst case we need to add preconditioning support to the proximal operators as well.

PR #406, opened 2016-05-29
Pass kwargs to sampling in DiscretizedSet.element

kohr-h:
This change makes it possible to do this:

```python
>>> def f(x, **kwargs):
...     c = kwargs.pop('c', 0.0)
...     return np.maximum(x, c)
...
>>> space = odl.uniform_discr(-1, 1, 4)
>>> space.element(f, c=0.5)
uniform_discr(-1.0, 1.0, 4).element([0.5, 0.5, 0.5, 0.75])
```

We won't get simpler than that since positional arguments with defaults are forbidden due to the subclassing of `Operator`. Better than nothing.

Furthermore, `RawDiscretization` and `Discretization` were renamed to `DiscretizedSet` and `DiscretizedSpace`, resp.

### DIFF
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -310,10 +310,8 @@ class DiscretizedSetVector(NtuplesBaseVector):
             to the size of the slice (same size, shape (1,)
             or single value).
         """
-        if isinstance(values, DiscretizedSetVector):
-            self.ntuple.__setitem__(indices, values.ntuple)
-        else:
-            self.ntuple.__setitem__(indices, values)
+        input_data = getattr(values, 'ntuple', values)
+        self.ntuple.__setitem__(indices, input_data)
 
     def sampling(self, ufunc, **kwargs):
         """Restrict a continuous function and assign to this vector
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -36,13 +36,13 @@ from odl.util.utility import (
 CudaCn = type(None)
 
 
-__all__ = ('RawDiscretization', 'RawDiscretizationVector',
-           'Discretization', 'DiscretizationVector')
+__all__ = ('DiscretizedSet', 'DiscretizedSetVector',
+           'DiscretizedSpace', 'DiscretizedSpaceVector')
 
 
-class RawDiscretization(NtuplesBase):
+class DiscretizedSet(NtuplesBase):
 
-    """Abstract raw discretization class.
+    """Abstract discretization class for general sets.
 
     A discretization in ODL is a way to encode the transition from
     an arbitrary set to a set of n-tuples explicitly representable
@@ -172,7 +172,7 @@ class RawDiscretization(NtuplesBase):
 
         Returns
         -------
-        element : `RawDiscretizationVector`
+        element : `DiscretizedSetVector`
             The discretized element, calculated as
             ``dspace.element(inp)`` or
             ``sampling(uspace.element(inp))``, tried in this order.
@@ -191,9 +191,9 @@ class RawDiscretization(NtuplesBase):
         Returns
         -------
         equals : `bool`
-            `True` if ``other`` is a `RawDiscretization`
+            `True` if ``other`` is a `DiscretizedSet`
             instance and all attributes `uspace`, `dspace`,
-            `RawDiscretization.sampling` and `RawDiscretization.interpolation`
+            `DiscretizedSet.sampling` and `DiscretizedSet.interpolation`
             of ``other`` and this discretization are equal, `False` otherwise.
         """
         if other is self:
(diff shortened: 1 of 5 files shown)


odl/discr/discretization.py
```diff
@@ -160,27 +160,33 @@ def interpolation(self):
...
-        inp : `object`, optional
-            The input data to create an element from. Must be
-            recognizable by the `LinearSpace.element`
-            method of either `dspace` or `uspace`.
+        inp : optional
+            The input data to create an element from. It needs to be
+            understood by either the `sampling` operator of this
+            instance or by its ``dspace.element`` method.
+        kwargs :
+            Additional arguments passed on to `sampling`. This can be
```
adler-j: Some word on "when `inp` is used in a call to `sampling"

odl/discr/discretization.py
```diff
@@ -303,7 +309,7 @@ def __setitem__(self, indices, values):
             to the size of the slice (same size, shape (1,)
             or single value).
         """
-        if isinstance(values, RawDiscretizationVector):
+        if isinstance(values, DiscretizedSetVector):
```
adler-j: shouldnt we ducktype this?

test/discr/lp_discr_test.py
```diff
@@ -273,6 +273,60 @@ def test_element_from_array_2d_shape():
...
+        return x * c + np.maximum(x, 0)
+
+    elem_f_default = space.element(f)
+    true_elem = [x * 0 + max(x, 0) for x in points]
+    assert all_equal(elem_f_default, true_elem)
+
+    elem_f_2 = space.element(f, c=2)
+    true_elem = [x * 2 + max(x, 0) for x in points]
+    assert all_equal(elem_f_2, true_elem)
+
```
adler-j: should add with a lambda, easy to do while we're at it.

adler-j: Should also have something where we use the nd syntax:

```python
def f(x):
    return x[0] * 2 + np.maximum(x, 0)
```

test/discr/lp_discr_test.py
```diff
@@ -273,6 +273,60 @@ def test_element_from_array_2d_shape():
...
+    # Without parameter
+    def f(x):
+        return x * 2 + np.maximum(x, 0)
+
+    elem_f = space.element(f)
+    true_elem = [x * 2 + max(x, 0) for x in points]
+    assert all_equal(elem_f, true_elem)
+
+    # With parameter
+    def f(x, **kwargs):
```
adler-j: No need to take **kwargs here, works without. See for example:

```python
def f(x, c=0):
    return x + c
    
def g(x, **kwargs):
    return f(x, **kwargs)
    
print(g(3, c=2))  # prints 5
```

Also works without `c=0`


```python
def f(x, c):
    return x + c
    
def g(x, **kwargs):
    return f(x, **kwargs)
    
print(g(3, c=2))  # prints 5
```

odl/discr/lp_discr.py
```diff
@@ -212,23 +212,28 @@ def exponent(self):
...
         """Create an element from ``inp`` or from scratch.
 
         Parameters
         ----------
         inp : `object`, optional
-            The input data to create an element from. Must be
-            recognizable by the `LinearSpace.element` method
-            of either `RawDiscretization.dspace` or
-            `RawDiscretization.uspace`.
+            The input data to create an element from. It needs to be
```
adler-j: This really could use an example

adler-j:
Nice clean change, some minor comments that i found when reviewing. Mostly on old stuff.

test/discr/lp_discr_test.py
kohr-h: The problem here is that the second positional argument is interpreted as `out` and has to have the name `out` since we pass the function through `uspace.element`, which makes a `FunctionSpaceVector` from it, which in turn is based on `Operator`. That's where the restriction comes from.

test/discr/lp_discr_test.py
kohr-h: What do you mean by "the nd syntax"? It's not obvious from your example.

odl/discr/lp_discr.py
kohr-h: This never had an example to start with, but I agree. I'll take the one from this conversation plus a bit.

odl/discr/discretization.py
kohr-h: Yes, good idea. After all, `ntuple` is not that common for an attribute name.

odl/discr/discretization.py
kohr-h: OK

test/discr/lp_discr_test.py
adler-j: I have an extra `[0]` after `x`, indicating use first dimension. This basically checks that `x` is two dimensional

test/discr/lp_discr_test.py
adler-j: Valid point, but somewhat problematic I suppose. Keep this for now then.

kohr-h: In the long run, we should probably make `FunctionSpaceVector` independent from `Operator`, at the expense of code duplication. But exactly this issue would be worth solving.

test/discr/lp_discr_test.py
kohr-h: That won't work as long as I don't use `x[1]`. We should make a separate issue on this.

kohr-h:
Updated according to the comments. When (if) Travis is happy, I'll merge.

test/discr/lp_discr_test.py
adler-j: In 1d there is no `x[1]`, right?

test/discr/lp_discr_test.py
adler-j: I'll have to agree here. If we ever go for operator spaces we need to add them separately.

test/discr/lp_discr_test.py
kohr-h: Ah, now I get it. Didn't properly check the context, thought it was the 2d case. Of course, here we can add it. I'll do that.

PR #407, opened 2016-05-29
Cherrypick diagonal precon

adler-j:
This basically cherrypicks the content of the diagonal preconditioning branch without the actual preconditioning, mostly minor fixups.

Was already reviewed.


kohr-h:
"remove" in the commit message. Otherwise, I remember that I looked through that once.

PR #420, opened 2016-05-31
ENH: Add mayo clinic reader

adler-j:
Initial pull for mayo clinic data reading

### DIFF
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -546,6 +546,81 @@ class RectPartition(object):
         newset = self.set[self.grid.inondeg]
         return RectPartition(newset, self.grid.squeeze())
 
+    def index_of(self, value, floating=False):
+        """Return the index of a value in the domain.
+        
+        Parameters
+        ----------
+        value : ``self.set`` element
+            The point to find the index of.
+        floating : bool
+            True if the index should also give the position inside the voxel.
+            This is given by returning the integer valued index of the voxel
+            plus the partial distance from the left to the right boundary.
+            
+        Returns
+        -------
+        index : float
+            Index of the value, as counted from the left.
+            
+        Examples
+        --------
+        Get the indices of start and end:
+        
+        >>> p = uniform_partition(0, 2, 5)
+        >>> p.index_of(0)
+        (0,)
+        >>> p.index_of(2)
+        (4,)
+        
+        Also supports points inside voxels, in that case returning the index
+        of the voxel:
+        
+        >>> p.index_of(0.2)
+        (0,)
+        
+        By using the ``floating`` argument, partial positions inside the voxels
+        can instead be determined:
+        
+        >>> p.index_of(0.2, floating=True)
+        (0.5,)
+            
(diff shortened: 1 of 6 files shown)


odl/tomo/data/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+
+"""Helpers for reading in standardized projection data formats."""
```
kohr-h: ~~projection~~ could also be other stuff

adler-j:
Before I move much further with this implementation I could use a review on the design used. Basically goes like this when working with MAYO data:

* User creates a `recon_space` as usual.
* User calls `mayo_projector_from_folder` (subject to rename) with `recon_space`, folder name and optionally some description on what subset of the data to use (currently start:end, possibly an arbitrary index later on). This returns a tuple `(projector, projection_data)` given by the data.
* User works with the data and projector as usual.

A possible change would be to make step 2 return `(geometry, projection_data)`, which would likely make much more sense. The problem with that is that then interpolating the data would be much harder.

kohr-h:
> User calls mayo_projector_from_folder (subject to rename) with recon_space, folder name and optionally some description on what subset of the data to use (currently start:end, possibly an arbitrary index later on).

Nothing wrong with that, I guess this is the way DICOM data is typically given.

> This returns a tuple (projector, projection_data) given by the data.

> A possible change would be to make step 2 return (geometry, projection_data), which would likely make much more sense. The problem with that is that then interpolating the data would be much harder.

Of course, returning geometry and data would be ideal since it's exactly the information contained in the data. Choosing the projection operator is the user's task, although there is not so much left to choose, basically just the implementation.

How about factoring out the operator creation and data interpolation part to a separate function? I guess that's where most of the future changes will happen. Possibly you need to pass on some parameters from the data sets.

Generally, I'd say go for whatever works best *right now*, we'll have to break API anyway at some point, hopefully at a low level nobody sees.

odl/discr/partition.py
```diff
@@ -546,6 +546,81 @@ def squeeze(self):
         newset = self.set[self.grid.inondeg]
         return RectPartition(newset, self.grid.squeeze())
 
+    def index_of(self, value, floating=False):
```
kohr-h: Call it just `index`, like python lists do.

adler-j:
Proposed action plan for this:

* Move it to `contrib`
* Get it merged without the curved detectors for now, fix that once astra gets it implemented.

kohr-h:
> Proposed action plan for this:
> 
>     Move it to contrib
>     Get it merged without the curved detectors for now, fix that once astra gets it implemented.

Agreed.

adler-j:
Closing in favor of #1321.

Issue #400, opened 2016-05-19
Improved integration with PyWavelets


ozanoktem:
The PyWavelet package has an extensive set of wavelets with over 70 built in filters, see [webpage](http://wavelets.pybytes.com). Furthermore, there is already an ODL wavelet operator structure that couples to PyWavelet, thus making this extensive library available from within ODL. The integration can however make better use of the capabilities offered by PyWavelet. 
1. Computing adjoint wavelet operator 
   Currently, there is no easy way to compute the adjoint of the ODL wavelet operator that makes us of PyWavelet routines. It should be rather straight forward to derive the adjoint wavelet operator even when we consider the PyWavelet routines as black/grey box. The first step is to express the filter for the adjoint. Next, these adjoint filters need to be computed for the built in filters in PyWavelet. This will allow us to implement the adjoint.
2. Applying wavelet operators along specific coordinates
   PyWavelet package offers its users the possibility to apply the Wavelet transform along specific axis, e.g., one may take the 2D wavelet transform in the first and third coordinates of function defined on R^3. One can here take a look at how @kohr-h implemented the Fourier transform in ODL.
3. Integrating wavelet functions
   There are PyWavelet routines for [Integrating wavelet functions](http://pywavelets.readthedocs.io/en/latest/ref/other-functions.html#integrating-wavelet-functions). This can be used to compute the various p-norms of wavelet functions.

adler-j:
What's up with this one. Will it ever happen in this form?

kohr-h:
I don't  have the time, and it doesn't seem to be terribly high on the list of folks. If someone complains about a missing feature, we can open a more focused issue IMO.

PR #421, opened 2016-05-31
Improve PyWavelet integration (issue #400) 

niinimaki:
- Added  the option to apply wavelet transform along a specific axes
- Added tests according to the the enhancement (tests for the new helper functions and tests for applying wavelet transform along a specific axes)
- Cleaned the tests a bit in overall


odl/trafos/wavelet.py
```diff
@@ -41,6 +41,114 @@
...
+                                           mode=mode)
+                nz = shp_z
+        else:
+            shp_z = nz
+
+        size_list.append((shp_x, shp_y, shp_z))
+        size_list.append(size_list[-1])
+        size_list.reverse()
+
+    return size_list
```
kohr-h: The 2d and 3d cases can be handled in the same way:

```python
ndim = len(shape)
axes_counts = [axes.count(i) for i in range(ndim)]
for ax_len, ax_count in zip(shape, axes_counts):
    if ax_count != 0:
        n = ax_len
        for _ in range(ax_count):
            n = pywt.dwt_coeff_len(n, filter_len=wbasis.dec_len,
                                   mode=mode)
    else:
        n = ax_len
    size_list.append(n)
```

odl/trafos/wavelet.py
```diff
@@ -41,6 +41,114 @@
 _SUPPORTED_IMPL = ('pywt',)
 
 
+def coeff_size_list_axes(shape, wbasis, mode, axes):
```
kohr-h: This doesn't have to be a whole new function. You can add an optional `axes` argument to the existing function and do the default case if it's not provided. The easiest way to do it is recursively, something like this:
```python
def coeff_size_list_axes(shape, wbasis, mode, axes=None):
    if axes is None:
        return coeff_size_list_axes(shape, wbasis, mode, [1] * len(shape))

    # Below the case with axes is implemented
```

kohr-h: Addition: I see that this function has no `nscales` argument. You can easily add it and make `nscales` and `axes` mutually exclusive.

odl/trafos/wavelet.py
```diff
@@ -136,7 +244,112 @@ def coeff_size_list(shape, nscales, wbasis, mode):
     return size_list
 
 
-def pywt_coeff_to_array(coeff, size_list):
+def pywt_dict_to_array(coeffs, size_list):
+    """Convert a PyWavelet coefficiet dictionary into a flat array.
```
kohr-h: coefficient

odl/trafos/wavelet.py
```diff
@@ -136,7 +244,112 @@ def coeff_size_list(shape, nscales, wbasis, mode):
...
+        flat_coeff[start:stop] = a.ravel()
+        start, stop = stop, stop + flat_sizes[1]
+        flat_coeff[start:stop] = d.ravel()
+
+    elif ndim == 2:
+        aa = coeffs['aa']
+        ad = coeffs['ad']
+        da = coeffs['da']
+        dd = coeffs['dd']
+        details = (ad, da, dd)
```
kohr-h:
```python
keys = list(coeffs.keys())
keys.sort()
details = tuple(coeffs[key] for key in keys if 'd' in key)
```

kohr-h: Works for 2 and 3 dimensions.

odl/trafos/wavelet.py
niinimaki: This part does need cleaning up. 
But, the above does not quite work exactly as I would like it to (if I understood you correctly, that is):
As an example:
`shape = (16, 16, 16)`
`axes = [0, 0, 2]`
would result to 
`size_list = [(16, 16, 16), 4, 16, 8]`
which after appending the last one and reversing would results
`size_list = [8, 8, 16, 4, (16, 16, 16)]`

But it should be
`size_list = [(4, 16, 8), (4, 16, 8), (16, 16, 16)]`

that is the values
`4, 16, 8`
need to be in a tuple...

odl/trafos/wavelet.py
niinimaki: True, I'll do this change.

odl/trafos/wavelet.py
niinimaki: Thanks. I'll modify the code overall accordingly.

niinimaki: Further improved the integration of `wavelet.py` with `PyWavelet`:

1.  use new PyWavelet functions `waverecn` and `wavedecn` to compute the multiresolution wavelet transform in 3D

2. Related to 1. modified the helper functions and tests

Also some of the review comments are used in this version.

odl/trafos/wavelet.py
kohr-h: Well, in that case, make an intermediate list and add as a tuple in the end:
```python
ndim = len(shape)
axes_counts = [axes.count(i) for i in range(ndim)]
reduced_shape = []
for ax_len, ax_count in zip(shape, axes_counts):
    if ax_count != 0:
        n = ax_len
        for _ in range(ax_count):
            n = pywt.dwt_coeff_len(n, filter_len=wbasis.dec_len,
                                   mode=mode)
    else:
        n = ax_len
    reduced_shape.append(n)

size_list.append(tuple(reduced_shape))
```

odl/trafos/wavelet.py
```diff
@@ -242,7 +455,95 @@ def pywt_coeff_to_array(coeff, size_list):
     return flat_coeff
 
 
-def array_to_pywt_coeff(coeff, size_list):
+def array_to_pywt_dict(coeff, size_list):
```
kohr-h: The name "dict" is misleading since there is no dictionary involved.

odl/trafos/wavelet.py
```diff
@@ -571,6 +872,11 @@ def __init__(self, dom, nscales, wbasis, mode):
 
             'per': periodization -- like periodic-padding but gives the
             smallest possible number of decomposition coefficients.
+        axes : sequence of `int`, optional
+            Dimensions in which to calculate the wavelet transform.
+            The sequence's length has to be equal to dimension of the ``grid``
+            `None` means traditional transform along the axes in ``grid``.
```
kohr-h: traditional??

odl/trafos/wavelet.py
```diff
@@ -571,6 +872,11 @@ def __init__(self, dom, nscales, wbasis, mode):
...
+            If axes is given nscales is discarded.
```
kohr-h: To silently discard (that's what this text suggests) is not good. The code should raise an exception (`ValueError`) if both options are given, and the text in "Parameters" should say that "this option cannot be combined with xyz" or similar.

odl/trafos/wavelet.py
```diff
@@ -606,23 +912,47 @@ def __init__(self, dom, nscales, wbasis, mode):
...
-        if dom.ndim == 1:
-            ran_size += sum(np.prod(shape) for shape in
-                            self.size_list[1:-1])
-        elif dom.ndim == 2:
-            ran_size += sum(3 * np.prod(shape) for shape in
-                            self.size_list[1:-1])
-        elif dom.ndim == 3:
-            ran_size += sum(7 * np.prod(shape) for shape in
-                            self.size_list[1:-1])
+        self._axes = axes
```
kohr-h: The argument should be sanitized, i.e. something like
```python
self._axes = tuple(int(ax) for ax  in axes)
```
This ensures that the attribute can be *guaranteed* to be of a certain type rather than user input dependent.
On top of that, `axes` is an important attribute of the transform, make it public (add a property).

odl/trafos/wavelet.py
```diff
@@ -606,23 +912,47 @@ def __init__(self, dom, nscales, wbasis, mode):
...
+            ran_size = np.prod(self.size_list[0])
+            if dom.ndim == 1:
+                ran_size += sum(np.prod(shape) for shape in
+                                self.size_list[1:-1])
+            elif dom.ndim == 2:
+                ran_size += sum(3 * np.prod(shape) for shape in
+                                self.size_list[1:-1])
+            elif dom.ndim == 3:
+                ran_size += sum(7 * np.prod(shape) for shape in
+                                self.size_list[1:-1])
```
kohr-h: This is a good candidate for a helper function, like `flat_size_from_size_list`.

odl/trafos/wavelet.py
```diff
@@ -652,21 +982,28 @@ def _call(self, x):
...
+                return self.range.element(coeff_arr)
+
+            if x.space.ndim == 3:
+                coeff_list = wavelet_decomposition3d(x, self.wbasis, self.mode,
+                                                     self.nscales)
+                coeff_arr = pywt_list_to_array(coeff_list, self.size_list)
+
+                return self.range.element(coeff_arr)
+        else:
+            coeff_dict = pywt.dwtn(x, self.wbasis, self.mode, self._axes)
```
kohr-h: Just curious: What happens if you use this function for standard axes, i.e. `[1] * self.domain.ndim`? Should give the same result as the three guys above. If that is the case, use this function in all cases, unless it is significantly slower (which is unlikely).

odl/trafos/wavelet.py
```diff
@@ -652,21 +982,28 @@ def _call(self, x):
...
-            coeff_arr = pywt_coeff_to_array(coeff_list, self.size_list)
-            return self.range.element(coeff_arr)
-
-        if x.space.ndim == 3:
-            coeff_dict = wavelet_decomposition3d(x, self.wbasis, self.mode,
-                                                 self.nscales)
-            coeff_arr = pywt_coeff_to_array(coeff_dict, self.size_list)
-
+        if self._axes is None:
+            if x.space.ndim == 1:
```
kohr-h: Slight remark: better use `self.domain.ndim` to determine dimension. Of course, we know that `x in self.domain` 99% sure (reserve 1% for bugs), but for `self.domain.ndim` we know 100% that it's right.

odl/trafos/wavelet.py
```diff
@@ -683,14 +1020,14 @@ def inverse(self):
         """The inverse wavelet transform."""
         return WaveletTransformInverse(
             ran=self.domain, nscales=self.nscales, wbasis=self.wbasis,
-            mode=self.mode)
+            mode=self.mode, axes=self._axes)
 
 
 class WaveletTransformInverse(Operator):
 
     """Discrete inverse wavelet trafo between discrete L2 spaces."""
```
kohr-h: trafo -> transform 

Also, there is no need to disallow other space exponents. Just need to filter that in the adjoint.

odl/trafos/wavelet.py
```diff
@@ -571,6 +872,11 @@ def __init__(self, dom, nscales, wbasis, mode):
 
             'per': periodization -- like periodic-padding but gives the
             smallest possible number of decomposition coefficients.
+        axes : sequence of `int`, optional
+            Dimensions in which to calculate the wavelet transform.
+            The sequence's length has to be equal to dimension of the ``grid``
```
kohr-h: This is not true. You can have lists of almost arbitrary lengths, as long as your input is large enough. The interpretation needs a bit more explanation here, i.e. what does it mean if an axis appears more than once?

odl/trafos/wavelet.py
kohr-h: In that function, create a dictionary like
```python
multiplicity = {1: 1, 2: 3, 3: 7}
```
Then you can simply write
```python
ran_size = (np.prod(self.size_list[0]) +
              sum(multiplicity[self.ndim] * np.prod(shape)
                  for shape in self.size_list[1:-1]))
```

odl/trafos/wavelet.py
```diff
@@ -813,19 +1180,26 @@ def _call(self, coeff):
         arr : `DiscreteLpVector`
```
kohr-h: We shouldn't hard-code `DiscreteLpVector`, better to write `domain element` and `range element`.

odl/trafos/wavelet.py
```diff
@@ -813,19 +1180,26 @@ def _call(self, coeff):
...
+                return self.range.element(x)
+            elif len(self.range.shape) == 3:
+                coeff_list = array_to_pywt_list(coeff, self.size_list)
+                x = wavelet_reconstruction3d(coeff_list, self.wbasis,
+                                             self.mode, self.nscales)
+                return x
+                #return self.range.element(x)
+        else:
+            coeff_dict = array_to_pywt_dict(coeff, self.size_list)
+            x = pywt.idwtn(coeff_dict, self.wbasis, self.mode, self._axes)
```
kohr-h: See comment above

kohr-h:
Mostly down to maintenance stuff now, looking good otherwise. Do we have tests?

odl/trafos/wavelet.py
kohr-h: Addition: I saw that you did not add full support for this feature yet. In that case, the documentation is correct. Would be good to have it fully supported anyway.

niinimaki: Yes, as we actually discussed (at least with @ozanoktem ) I decided first to implement this with the above "restriction" and add the full support later on.

odl/trafos/wavelet.py
adler-j: Could also be simplified further to

```python
ndim = len(shape)
axes_counts = [axes.count(i) for i in range(ndim)]
reduced_shape = []
for ax_len, ax_count in zip(shape, axes_counts):
    n = ax_len
    for _ in range(ax_count):
        n = pywt.dwt_coeff_len(n, filter_len=wbasis.dec_len,
                               mode=mode)
    reduced_shape.append(n)

size_list.append(tuple(reduced_shape))
```

odl/trafos/wavelet.py
adler-j: Also write parameter names with ticks.

If \`\`axes\`\` is given \`\`nscales\`\` is discarded.

odl/trafos/wavelet.py
niinimaki: If you use `pywt.dwtn` with standard axes would give you wavelet transform on **one** scaling level. For multilevel decomposition [`PyWavelet` API](http://pywavelets.readthedocs.io/en/latest/ref/nd-dwt-and-idwt.html)  recommends to use `pywt.wavedecn` instead. This `pywt.wavedecn` and its pair `pywt.waverecn` are new features of `PyWavelet` (on version 0.4.0). 

In the previous versions multilevel decomposition for higher than 2D had to be computed using `pywt.dwtn` which then did not have the `axes` option and if you wanted to perform the multilevel decomposition for 3D you had to write helper for that yourself. This was the case in the previous version of `wavelet.py` i.e. helper functions `wavelet_decomposition3d` and `wavelet_reconstruction3d`. Now these have been removed and `pywt.wavedecn` and `pywt.waverecn` are used instead.

kohr-h: Okay, good. Thanks for the link. A follow-up question, just to be sure: Is (in 2 dimensions) "wavelet decomposition using 2 levels" the same as "wavelet transform with axes [0, 0, 1, 1]"? In other words, is a wavelet decomposition to level n the same as n repeated wavelet transforms? It would be important to understand this, should be easy to test numerically.

odl/trafos/wavelet.py
kohr-h: Of course, I didn't see the [the gorilla in the scene](https://www.youtube.com/watch?v=vJG698U2Mvo) :-)

odl/trafos/wavelet.py
niinimaki: No they are not (though at first one would think that they are). But `wavedec2` using 2 levels returns the coefficients at each level, as for `pywt.dwtn` using `axes = [0, 0, 1, 1]` returns the final resulting coefficients at the "end" level. For an example a 2D case with Daubechies 2 wavelets:

`x = np.ones((32,  32))`
`wbasis = pywt.Wavelet('db2')`
`coeffs_using_wavedec2 = pywt.wavedec2(x, wbasis, mode='sym', level=2)`

returns a list

`[aa0, (ad0, da0, dd0), (ad1, da1, dd1)]`

where  the coefficient in the coarsest level (`aa0, ad0, da0,dd0`) are of size `10 x 10` and the coefficient at the finest (at this case 2nd level, i.e coeffs `ad1, da1, dd1`), are of size `17 x 17`

As for

`coeffs_using_dwtn_with_axes = pywt.dwtn(x, wbasis, mode='sym', axes=[0, 0, 1, 1])`

returns a dictionary with keywords

`['aaaa',  'aaad',  'aada',  'aadd',  'adaa',  'adad',  'adda',  'addd',  'daaa',  'daad',  'dada',  'dadd',  'ddaa',  'ddad',   ddda',  'dddd']`

which all are of (the final size) `10 x 10 `.

kohr-h: Then these two functionalities are quite different. The first one is (as I understand it) the wavelet decomposition that stores the minimum amount of coefficients necessary to uniquely reconstruct the original signal. The second one seems to take inner products with either the scaling function or the wavelet in the provided axes, including the usual size reduction. This seems to be closer to the notion of "Continuous wavelet transform".

Comparing sizes, one gets
```python
>>> 32 ** 2
1024
>>> sum(np.size(val) for val in using_wavedec2)
1267
>>> sum(np.size(val) for val in using_dwtn.values())
1600
```

Do the PyWavelets guys have a document lying around that explains what exactly is computed? Something like [this](http://www.fftw.org/doc/The-1d-Discrete-Fourier-Transform-_0028DFT_0029.html#The-1d-Discrete-Fourier-Transform-_0028DFT_0029)?

niinimaki: Of course,  the sizes are dependent on the chosen boundary behaviour and the chosen wavelet basis...

> Do the PyWavelets guys have a document lying around that explains what exactly is computed? Something like this?

Can't say off-hand, but I'll try to look for a one.

niinimaki:
Improved and cleaned up `wavelet_trafo.py` according to the suggestions.
Major modifs:

- now there is only one `coeff_size_list` helper function, with optional inputs `nscales` or `axes`

- `nscales` and `axes` are now both optional but one of them has to be defined when creating the wavelet transform operator

TODO: full support for `axes` option (work in progress for this).

odl/trafos/wavelet.py
```diff
@@ -70,20 +61,34 @@ def coeff_size_list(shape, nscales, wbasis, mode):
...
 
-        'ppd': periodic padding -- signal is trated as a periodic one
+        'periodic': periodic padding -- signal is trated as a periodic one
 
-        'sp1': smooth padding -- signal is extended according to the
+        'smooth': smooth padding -- signal is extended according to the
         first derivatives calculated on the edges (straight line)
 
-        'per': periodization -- like periodic-padding but gives the
+        'periodization': periodization -- like periodic-padding but gives the
```
kohr-h: Great change, much more readable!

A small comment: since this changes the API (and the function is public), such a change should have "API:" in the commit message. Maybe you did that. If not: if you happen to have this isolated in one commit, change the commit message with `git rebase -i`, otherwise leave it and do it next time :-)

odl/trafos/wavelet.py
```diff
@@ -106,37 +111,163 @@ def coeff_size_list(shape, nscales, wbasis, mode):
...
-
-    # Add a duplicate of the last entry for the approximation coefficients
-    size_list.append(size_list[-1])
-
-    # We created the list in reversed order compared to what pywt expects
-    size_list.reverse()
+    if nscales is None and axes is None:
+        raise ValueError('Either nscales or axes has to be defined')
+
+    if axes is None:
```
kohr-h: It's more readable if you structure is like this
```python
if axes is not None:
    if nscales is not None:
        raise ValueError(...)
    # Do something with axes
elif nscales is not None:
    # Do something with nscales
else:
    raise ValueError('either `nscales` or `axes` has to be given')
```
You are using opposite logic, i.e. check if something is *not* given, and then do the other thing, which makes the code more non-obvious.

niinimaki:
Full support for the `PyWavelets` `axes` option has been added to the `wavelet.py`. This resulted in some necessary changes in the helper functions `pywt_dict_to_array` and `array_to_pywt_dict`.
Tests i.e. `wavelet_test.py` has also been modified accordingly (quite small modif there).

niinimaki:
Added adjoint for biorthogonal wavelets ( biorthogonal wavelets from `PyWavelet` wavelet families). 
Added adjoint tests into `wavelet_test.py`.
The adjoint is correct for all wavelet that have filter length 2 and for what ever filter lengths when boundary handling is either `zero` or `periodization`.
WIP for other combination of boundaries and filters.

niinimaki:
WIP: Added helpers to `wavelet.py` related to the adjoint correction.

niinimaki:
Added an operator for the adjoint that uses the new helper functions. 
The adjoint call of `WaveletTransform` then check whether to call `WaveletTransformInverse` or `WaveletTransformAdjoint` depending on the filters and on the boundary handling.

niinimaki:
According to discussion with @kohr-h, I added a helper function that constructs also the full (i.e. the NOT down sampled) adjoint correction matrices. Both down sampled and not down sampled correction matrices are written for signal extension modes `constant`, `periodic` and `symmetric`.

niinimaki:
Smooth padding is added into codes that construct both the down sampled and full correction matrices.
TODO: add proper handling of these matrices into the helper function `correct_adjoint_of_wavelet_transform`.

niinimaki:
Correction matrices and corrections for adjoints of PyWavelet's wavelet transform have been added into `wavelet.py` for all PyWavelet's signal extension modes.

Cases with filter length 2 or with signal extension modes `'zero'` or `'periodization'` (both with orthogonal and biorthogonal wavelets): the adjoint is given by the inverse wavelet transform, no further corrections are needed.
For any other filter lengths or when using signal extension modes `'constant'`, `'periodic'`, `'symmetric'` or `'smooth'`, helper functions have been added that construct the adjoint correction matrices (these are small size matrices). Also a helper function has been written that corrects for the "false" adjoints with above mentioned paddings in 1D, 2D and in 3D with one or several scaling levels.

Tests according to the improvement of PyWavelet's ODL integration have equally been added.

adler-j:
What happens here now?

kohr-h:
This is work in progress. The basic functionality has been added by @niinimaki , but the errors are larger than they should be. Needs a thorough check of the math and the implementation. Probably also clean-up. It's on my table in some way.

adler-j:
I tried `git rebase master` on this, but it seems the branch has some decent divergence. It would probably be wise to rebase it quite soon to lessen the burden for later.

kohr-h:
That's a bit surprising since there are only 3 files touched, 2 of which are wavelet related. Those shouldn't have seen too much change on master. Apart from changes across all files like style and similar, of course.

adler-j:
It seems the large changes with this one clashes with the naming/style fixes we have done recently, so many lines have been changed by both.

adler-j:
Is this dead after #547?

kohr-h:
No, it implements the adjoint with boundary correction. But it's not completely correct yet and will have tons of conflicts with #547. I'll pick the relevant code from this branch and close the PR when #547 is in.

Issue #419, opened 2016-05-31
Rules for proximal operators


aringh:

adler-j:
See the ones that already exist for inspiration:

* Separable sum: [combine_proximals](https://github.com/odlgroup/odl/blob/deec86db698cbf3def7984704a9e9185b63893f5/odl/solvers/advanced/proximal_operators.py#L49)
* Complex conjugate (Moreau identity): [proximal_cconj](https://github.com/odlgroup/odl/blob/deec86db698cbf3def7984704a9e9185b63893f5/odl/solvers/advanced/proximal_operators.py#L88)
* Composition with (certain) linear operators [proximal_composition](https://github.com/odlgroup/odl/blob/deec86db698cbf3def7984704a9e9185b63893f5/odl/solvers/advanced/proximal_operators.py#L135)

Which ones are we planning to add here?

aringh:
Versions of i), ii) and iv) from https://www.ljll.math.upmc.fr/~plc/prox.pdf

Issue #279, opened 2016-02-17
Style of citation


kohr-h:
### Examples

- Journal article, 1 author
```
.. [Eul1741] Euler, L. *Solutio problematis ad geometriam situs pertinentis*.
   Commentarii academiae scientiarum Petropolitanae, 8 (1741), pp 128--140.
```
- Book, 2 authors
```
.. [SW1971] Stein, E and Weiss, G. *Introduction to Fourier Analysis on
   Euclidean Spaces*. Princeton University Press, 1971.
```
- Book, 3+ authors
```
.. [Pre+2007] Press, W H, Teukolsky, S A, Vetterling, W T, and Flannery, B P.
   *Numerical Recipes in C - The Art of Scientific Computing* (Volume 3).
   Cambridge University Press, 2007.
```
- Proceedings article, 3 authors
```
.. [BLZ2008] Bertero M, Lanteri H, and Zanni, L. *Iterative image
   reconstruction: a point of view*. In: Censor, Y, Jiang M, Louis, A K.
   Proceedings of the Interdisciplinary Workshop on Mathematical
   Methods in Biomedical Imaging and Intensity-Modulated Radiation
   (IMRT), Pisa, Italy, 2008, pp 37--63.
```

### Tags

Since [these references are global](http://www.sphinx-doc.org/en/stable/rest.html#citations) throughout the doc, we need unique citation tags. I suggest the following, close to what @moosmann has been using in the Chambolle-Pock branch, which is a slight variant of the "alpha" BibTeX style (see examples):

- 1 author: **First three letters** of the family name plus **full year**. Append "a", "b" etc. if there are multiples.
- 2-3 authors: Capitalized **first letters** of the family names plus **full year**. Disambiguation as above.
- More than 3 authors: **First three letters of first author** plus **"+"** plus **full year**. Disambiguation as above.

### Authors
- **Family name** (no special characters) - comma - **first letters of given names** without punctuation, separated by space
- 2 authors: "and" between author names without preceding comma
- 3 or more authors: "and" between second last and last author names, including preceding comma

Full stop at the end.

### Title
**Full title** (no special characters) as **emphasized** type (between `*`). For books, volume info is appended after the closing `*` as "(Volume n)".

Full stop at the end.

### For books
**Full name of the publisher** (without city) - comma - **year**.

Full stop at the end.

### For journal articles
**Full name of the journal** - comma - **issue** - **year in parentheses** - comma - **"pp" plus page range**.

Full stop at the end.

[...]

PR #422, opened 2016-06-01
Issue 419  proximal rules

aringh:
This adds three identities for proximal operators; when the function is
1) translated
2) the argument is scaled
3) when the function is perturbed by a quadratic term.
(see i), ii) and iv) in table 10.1 in https://www.ljll.math.upmc.fr/~plc/prox.pdf, i.e. chapter 10 in http://link.springer.com/book/10.1007/978-1-4419-9569-8).

It also adds three tests to check these implementations.

(This relates to issue #419 )

### DIFF
--- a/doc/source/refs.rst
+++ b/doc/source/refs.rst
@@ -5,11 +5,11 @@ References
    gradient methods*. IMA Journal of Numerical Analysis, 8 (1988),
    pp 141--148.
 
-.. [BC2015] Bot, R.I, and Csetnek, E.R. *On the convergence rate of
+.. [BC2015] Bot, R I, and Csetnek, E R. *On the convergence rate of
    a forward-backward type primal-dual splitting algorithm for convex
    optimization problems*. Optimization, 64.1 (2015), pp 5--23.
 
-.. [BH2013] Bot, R.I, and Hendrich, C. *A Douglas-Rachford type
+.. [BH2013] Bot, R I, and Hendrich, C. *A Douglas-Rachford type
    primal-dual method for solving inclusions with mixtures of
    composite and parallel-sum type monotone operators*. SIAM Journal
    on Optimization, 23.4 (2013), pp 2541--2565.
@@ -21,15 +21,21 @@ References
 .. [BV2004] Boyd, S, and Vandenberghe, L. *Convex optimization*.
    Cambridge university press, 2004.
 
-.. [CP2011a] Chambolle, Antonin and Pock, Thomas. *A First-Order
+.. [CP2011a] Chambolle, A and Pock, T. *A First-Order
    Primal-Dual Algorithm for Convex Problems with Applications to
    Imaging*. Journal of Mathematical Imaging and Vision, 40 (2011),
    pp 120-145.
 
-.. [CP2011b] Chambolle, Antonin and Pock, Thomas. *Diagonal
+.. [CP2011b] Chambolle, A and Pock, T. *Diagonal
    preconditioning for first order primal-dual algorithms in convex
    optimization*. 2011 IEEE International Conference on Computer Vision
    (ICCV), 2011, pp 1762-1769.
+
+.. [CP2011c] Combettes, P L, and Pesquet, J-C. *Proximal splitting
+   methods in signal processing.* In:  Bauschke, H H, Burachik, R S,
+   Combettes, P L, Elser, V, Luke, D R, and Wolkowicz, H. Fixed-point
+   algorithms for inverse problems in science and engineering, Springer,
+   2011.
 
 .. [GNS2009] Griva, I, Nash, S G, and Sofer, A. *Linear and nonlinear
    optimization*. Siam, 2009.
@@ -41,7 +47,7 @@ References
    Scherzer, O. Handbook of Mathematical Methods in Imaging.
    Springer, 2015, pp 937--1031.
 
-.. [PB2014] Parikh, Neal and Boyd, Stephen. *Proximal Algorithms*.
+.. [PB2014] Parikh, N, and Boyd, S. *Proximal Algorithms*.
    Foundations and Trends in Optimization, 1 (2014), pp 127-239.
 
 .. [Pre+2007] Press, W H, Teukolsky, S A, Vetterling, W T, and Flannery, B P.
@@ -55,12 +61,12 @@ References
 .. [Roc1970] Rockafellar, R. Tyrrell. *Convex analysis*. Princeton
    University Press, 1970.
 
-.. [Sid+2012] Sidky, Emil Y, Jorgensen, Jakob H, and Pan, Xiaochuan.
+.. [Sid+2012] Sidky, E Y, Jorgensen, J H, and Pan, X.
    *Convex optimization problem prototyping for image reconstruction in
    computed tomography with the Chambolle-Pock algorithm*. Physics in
    Medicine and Biology, 57 (2012), pp 3065-3091.
 
-.. [SW1971] Stein, E and Weiss, G.
+.. [SW1971] Stein, E, and Weiss, G.
    *Introduction to Fourier Analysis on Euclidean Spaces*.
    Princeton University Press, 1971.
 
--- a/odl/solvers/advanced/proximal_operators.py
+++ b/odl/solvers/advanced/proximal_operators.py
@@ -30,13 +30,18 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import super
 
+import numpy as np
+
 from odl.operator.operator import Operator
-from odl.operator.default_ops import IdentityOperator
+from odl.operator.default_ops import (IdentityOperator, ScalingOperator,
+                                      ConstantOperator, ResidualOperator)
 from odl.operator.pspace_ops import DiagonalOperator
 from odl.space.pspace import ProductSpace
+from odl.set.space import LinearSpaceVector
 
 
-__all__ = ('combine_proximals', 'proximal_cconj',
+__all__ = ('combine_proximals', 'proximal_cconj', 'proximal_translation',
+           'proximal_arg_scaling', 'proximal_quadratic_perturbation',
            'proximal_composition', 'proximal_zero',
            'proximal_box_constraint', 'proximal_nonnegativity',
            'proximal_l1', 'proximal_cconj_l1',
@@ -85,18 +90,18 @@ def combine_proximals(*factory_list):
     return make_diag
 
 
-def proximal_cconj(proximal):
+def proximal_cconj(prox_factory):
     """Calculate the proximal of the dual using Moreau decomposition.
 
     The Moreau identity states that for any convex function ``F`` with
     convex conjugate ``F^*``, the proximals satisfy
 
-        prox[a * F^*](x) + a * prox[F / a](x / a) = x
+        prox[s * F^*](x) + s * prox[F / s](x / s) = x
 
-    where ``a`` is a scalar step size. Using this, we find the proximal of the
+    where ``s`` is a scalar step size. Using this, we find the proximal of the
     convex conjugate
 
-        prox[a * F^*](x) = x - a * prox[F / a](x / a)
+        prox[s * F^*](x) = x - s * prox[F / s](x / s)
 
     Note that since ``(F^*)^* = F``, this can be used to get the proximal of
     the original function from the proximal of the convex conjugate.
@@ -111,6 +116,10 @@ def proximal_cconj(proximal):
     -------
     prox : `callable`
         Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the Moreau identity, see [ComPes2011]_.
     """
 
     def cconj_prox_factory(step_size):
@@ -124,12 +133,182 @@ def proximal_cconj(proximal):
         Returns
         -------
         proximal : `Operator`
-            The proximal operator of ``prox[a * f^*](x)``
+            The proximal operator of ``s * F^*`` where ``s`` is the step size
         """
-        prox_other = step_size * proximal(1.0 / step_size) * (1.0 / step_size)
+        prox_other = (step_size * prox_factory(1.0 / step_size) *
+                      (1.0 / step_size))
         return IdentityOperator(prox_other.domain) - prox_other
 
     return cconj_prox_factory
+
+
+def proximal_translation(prox_factory, y):
+    """Calculate the proximal of the translated function F(x - y).
+
+    This is calculated according to the rule
+
+        prox[s * F( . - y)](x) = y + prox[s * F](x - y)
+
+    where ``y`` is the translation, and ``s`` is the step size.
+
+    Parameters
+    ----------
+    prox_factory : `callable`
+        A factory function that, when called with a step size, returns the
+        proximal operator of ``F``.
+    y : Element in domain of ``F``.
+
+    Returns
+    -------
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [CP2011c]_.
+    """
+
+    def translation_prox_factory(step_size):
+        """Create proximal for the translation with a given step_size.
+
+        Parameters
+        ----------
+        step_size : positive `float`
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * F( . - y)`` where ``s`` is the
+            step size
+        """
+
+        return (ConstantOperator(y) + prox_factory(step_size) *
+                (IdentityOperator(y.space) - ConstantOperator(y)))
+
+    return translation_prox_factory
+
+
+def proximal_arg_scaling(prox_factory, scaling):
+    """Calculate the proximal of function F(x * scaling).
+
+    This is calculated according to the rule
+
+        prox[s * F( . * scaling)](x) =
+        1/scaling * prox[s * scaling^2 * F ](x * scaling)
+
+    where ``scaling`` is the scaling parameter, and ``s`` is the step size.
+
+    Parameters
+    ----------
+    prox_factory : `callable`
+        A factory function that, when called with a step size, returns the
+        proximal operator of ``F``
+    scaling : `float`
+        Scaling parameter
+
+    Returns
+    -------
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [CP2011c]_.
+    """
+
+    scaling = float(scaling)
+    if scaling == 0:
+        return proximal_zero(prox_factory(1.0).domain)
+
+    def arg_scaling_prox_factory(step_size):
+        """Create proximal for the translation with a given step_size.
+
+        Parameters
+        ----------
+        step_size : positive `float`
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * F( . * a)`` where ``s`` is the
+            step size
+        """
+        prox = prox_factory(step_size * scaling ** 2)
+        return (1 / scaling) * prox * scaling
+
+    return arg_scaling_prox_factory
+
+
+def proximal_quadratic_perturbation(prox_factory, a, u=None):
+    """Calculate the proximal of function F(x) + a * ||x||^2 + <u,x>.
+
+    This is calculated according to the rule
+
+        prox[s * (F( . ) + a * || . ||^2 + <u, . >)](x) =
+        c prox[s*f( . * c)]((x - s*u)*c)
+
+    where ``c`` is the constant c = 1/sqrt(s*2*a + 1), ``a`` is the scaling
+    parameter belonging to the quadratic term, ``u`` is the vector defining the
+    linear functional, and ``s`` is the step size.
+
+    Parameters
+    ----------
+    prox_factory : `callable`
+        A factory function that, when called with a step size, returns the
+        proximal operator of ``F``
+    a : non-negative `float`
+        Scaling of the quadratic term
+    u : Element in domain of F, optional
+        Defines the linear functional
+        Default: Treated as zero vector
+
+    Returns
+    -------
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [CP2011c]_. Note that this identity
+    is not the exact one given in the reference, but was recalculated for
+    arbitrary step lengths.
+    """
+
+    a = float(a)
+    if a < 0:
+        raise ValueError('scaling parameter {} not non-negative'.format(a))
+
+    if u is not None and not isinstance(u, LinearSpaceVector):
+        raise TypeError('vector {!r} not None or a LinearSpaceVector instance.'
+                        ''.format(u))
+
+    def quadratic_perturbation_prox_factory(step_size):
+        """Create proximal for the quadratic perturbation with a given
+        step_size.
+
+        Parameters
+        ----------
+        step_size : positive `float`
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * (F(x) + a * ||x||^2 + <u,x>)``,
+            where ``s`` is the step size
+        """
+        const = 1.0 / np.sqrt(step_size * 2.0 * a + 1)
+        prox = proximal_arg_scaling(prox_factory, const)(step_size)
+        if u is not None:
+            return (const * prox *
+                    ResidualOperator(ScalingOperator(u.space, const),
+                                     step_size * const * u))
+        else:
+            return const * prox * const
+
+    return quadratic_perturbation_prox_factory
 
 
 def proximal_composition(proximal, operator, mu):
@@ -218,7 +397,7 @@ def proximal_zero(space):
         ----------
         tau : positive `float`
             Unused step size parameter. Introduced to provide a unified
-            interface
+            interface.
 
         Returns
         -------
@@ -292,7 +471,7 @@ def proximal_box_constraint(space, lower=None, upper=None):
             Parameters
             ----------
             tau : positive `float`
-                Step size parameter. Not used.
+                Step size parameter, not used.
             """
             super().__init__(domain=space, range=space, linear=False)
 
@@ -572,7 +751,7 @@ def proximal_l2_squared(space, lam=1, g=None):
     For a step size ``tau``, the proximal operator of ``tau * F`` is
     given by
 
-        prox[tau * F](x) = (x + 2 * tau * lam * g) / (2 * tau * lam)
+        prox[tau * F](x) = (x + 2 * tau * lam * g) / (1 + 2 * tau * lam)
 
     Parameters
     ----------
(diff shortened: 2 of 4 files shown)


adler-j:
Merge the last doc/style commits into a single commit to clean up the commit history abit, will review.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -29,14 +29,19 @@
 from future import standard_library
 standard_library.install_aliases()
 from builtins import super
+from numpy import sqrt
```
adler-j: Use import numpy as np instead

odl/solvers/advanced/proximal_operators.py
```diff
@@ -29,14 +29,19 @@
...
 
 from odl.operator.operator import Operator
 from odl.operator.default_ops import IdentityOperator
+from odl.operator.default_ops import ScalingOperator
```
adler-j: from odl.operator import IdentityOperator, ScalingOperator, ConstantOperator, DiagonalOperator

odl/solvers/advanced/proximal_operators.py
```diff
@@ -111,6 +116,10 @@ def proximal_cconj(proximal):
     -------
     prox : `callable`
         Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [ComPes2011]_.
```
adler-j: Be explicit:

For reference on the Moreau identity, see [ComPes2011]_.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
 
 
+def proximal_translation(prox_factory, y):
+    """Calculate the proximal of the translated function F(x - y).
+
+    This is calculated according to the rule
+
+        prox[s * F( . - y)](x) = y + prox[s * F](x - y)
+
+    where ``a`` is the translation, and ``s`` is the step size.
```
adler-j: There is no `a` above

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+        """
+
+        return (ConstantOperator(y) + prox_factory(step_size) *
+                (IdentityOperator(y.space) - ConstantOperator(y)))
+
+    return translation_prox_factory
+
+
+def proximal_arg_scaling(prox_factory, a):
+    """Calculate the proximal of function F(x * 1/a).
```
adler-j: Wouldnt defining it by ``F(a * x)`` be more intuitive for the reader? That would also match our "right multiplication" type operators.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+
+        prox[s * F( . * 1/a)](x) = a * prox[s/a^2 * F ](x * 1/a)
+
+    where ``a`` scaling parameter, and ``s`` is the step size.
+
+    Parameters
+    ----------
+    prox_factory : `callable`
+        A factory function that, when called with a step size, returns the
+        proximal operator of ``F``
```
adler-j: dot

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * F( . * 1/a)`` where ``s`` is the
+            step size
+        """
+        the_prox = prox_factory(step_size / a**2)
+        return a * the_prox * ScalingOperator(the_prox.domain, 1 / a)
```
adler-j: We support right multiplication, so this could be rewritten

```python
return a * the_prox * (1 / a)
```

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+        step_size : positive `float`
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * F( . * 1/a)`` where ``s`` is the
+            step size
+        """
+        the_prox = prox_factory(step_size / a**2)
```
adler-j: wierd name, perhaps simply `prox`?

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+    Returns
+    -------
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [ComPes2011]_.
+    """
+
```
adler-j: a = float(a)

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+    if a == 0:
```
adler-j: If we chang the definition to F(a * x) this check wouldnt be needed.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+            step size
+        """
+        the_prox = prox_factory(step_size / a**2)
+        return a * the_prox * ScalingOperator(the_prox.domain, 1 / a)
+
+    return arg_scaling_prox_factory
+
+
+def proximal_quadratic_perturbation(prox_factory, a, u):
+    """Calculate the proximal of function F(x) + a/2 * ||x||^2 + <u,x>.
```
adler-j: a / 2 feels like a computational thing, shouldnt we be able to do without it?

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+
+    This is calculated according to the rule
+
+        prox[s * (F( . ) + a/2 * || . ||^2 + <u, . >)](x) =
+        1/sqrt(s*a + 1) prox[s*f( . * 1/sqrt(s*a + 1))]((x - s*u)/sqrt(s*a+1))
```
adler-j: Define `1 / sqrt(s * a + 1)` on the next line to make this easier to read

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+
+    where ``a`` is the scaling parameter belonging to the quadratic term, and
+    ``u`` is the vector defining the linear functional.
```
adler-j: And s is a step size

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * (F(x) + a/2 * ||x||^2 + <u,x>)``,
+            where ``s`` is the step size
+        """
+        const = 1.0 / sqrt(step_size * a + 1)
+        the_prox = proximal_arg_scaling(prox_factory, 1.0 / const)(step_size)
+        return (const * the_prox *
```
adler-j: Slight optimization and readability improvement:

```python
const * the_prox * (ScalingOperator(u.space, const) + ConstantOperator((- step_size * const) * u))
```

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,180 @@ def cconj_prox_factory(step_size):
...
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * (F(x) + a/2 * ||x||^2 + <u,x>)``,
+            where ``s`` is the step size
+        """
+        const = 1.0 / sqrt(step_size * a + 1)
+        the_prox = proximal_arg_scaling(prox_factory, 1.0 / const)(step_size)
```
adler-j: This does not seem to agree with your text, since the defintion of `proximal_arg_scaling` is `F(x / a)` this would be equivalent to `F(x * sqrt( ... ))`, but in the doc you say `F(x / sqrt( ... ))`. This would be fixed if the definition of `proximal_arg_scaling` was changed.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -595,6 +770,9 @@ def proximal_l2_squared(space, lam=1, g=None):
     proximal_cconj_l2_squared : proximal for convex conjugate
     """
 
+    # TODO: Is doc wrong? Should it be
```
adler-j: A review of that would certainly be nice, doc can be wrong.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -230,7 +230,7 @@ def test_proximal_convconj_l2_sq_wo_data():
     x = space.element(x_arr)
 
     # Create data
-    g = space.element(-2 * x_arr)
+    g = space.zero()
```
adler-j: Why?

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,117 @@
...
+import odl.solvers as odls
+
+from odl.util.testutils import all_almost_equal
+
+# Places for the accepted error when comparing results
+PLACES = 8
+
+
+def test_proximal_translation():
+    """Test for the proximal of a translation: prox[f(. - g)]"""
```
adler-j: Not quite sure I liks this test design, makes the tests circular. What if  `proximal_l2` was implemented using `proximal_translation`? Then it would be true by definition.

odl/solvers/advanced/proximal_operators.py
aringh: If you change definition, wouldn't you get 1/a in front of the prox in the formula? Because if so then you still need to check ```if a == 0```.

However, with this new definition then a = 0 would return the Identity operator (right?). So you could do the check in the beginning and simply return this it is true. Is this better?

adler-j: I would say that it is better, but I'm not 100% certain.

odl/solvers/advanced/proximal_operators.py
aringh: There is not suppose to be an ```a``` above, so I guess this is not necessary.

However: should we have ```step_size = float(step_size)``` in the beginning of all proximal operators? Is that what you mean?

odl/solvers/advanced/proximal_operators.py
aringh: In this case ```a = 0``` gives you the identity operator. Is this expected or not? Also see my answer to your comment below.

odl/solvers/advanced/proximal_operators.py
adler-j: You take `a` as an input parameter, right? Then you can do this make sure it is a float.

odl/solvers/advanced/proximal_operators.py
adler-j: The more I think of it, the more I think that yes, that is what should happen.

odl/solvers/advanced/proximal_operators.py
aringh: ```const = 1 / sqrt(...)``` so ```1 / const``` is ```sqrt(...)``` and by sending ```1/const = sqrt(...)``` to scaling defined for ```F(x * 1/a)``` where ```a``` is the input parameter we get the scaling ```F(x *1/sqrt(...))``` as desired. But yes, if scaling is changed to ```F(x*a)```, this needs to be changed.

adler-j: Darn, you are correct. Well goes to show we should probably do the change since stuff like this becomes less complicated.

test/solvers/advanced/proximal_operator_test.py
aringh: Because the naming of the test is ```l2_sq_wo_data```, and I assume that wo is suppose to mean "with out". And yes, it is still doing something. It is testing the ```l2_sq``` when there is "no off-set".

The test just below it in the code is identical if you don't do this change.

adler-j: If it is w.o. data, you are indeed correct, but then this should be removed. The reason we have this differently set up is because we know the function does optimization depending on if it is given `g` or not. By setting it to 0 instead of `None` we don't get to test these optimizations.

odl/solvers/advanced/proximal_operators.py
aringh: The input parameter the we take is ```y``` (it was a typo in the Doc where it says ```a```). Why is a translation, so it lives in the domain of the function. The "problem" is that the proximal factory cannot tell me what that domain is, so if you want to check that you have to create a proximal operator and then ask that one for the domain. Should this be done?

The only float value that we take in this class is the step length. Should I add this to all proximal factories? (basically all of them take take a step length as input)

adler-j: The row bellow this is literally `if a == 0`, are you sure you are thinking about the correct code?

aringh: I see your comment on line 170, which belongs to ```proximal_translation```... anyway now I know where the fix needs to be done; will do :)

adler-j: Wierd, says 218 to me :s

With that said, your issue with the proximal not knowing about its domain is something we face across the board, I'm not sure how we would fix this without making the proximal factories classes, something i would prefer to avoid.

odl/solvers/advanced/proximal_operators.py
kohr-h: no

doc/source/refs.rst
```diff
@@ -21,6 +21,10 @@ References
 .. [BV2004] Boyd, S, and Vandenberghe, L. *Convex optimization*.
    Cambridge university press, 2004.
 
+.. [ComPes2011] Combettes, P.L, and Pesquet, J.-C. *Proximal splitting
```
kohr-h: CP instead of ComPes. Space after P.
Check out #279

aringh: I know this is how it should be, but if you look at the references just below it is also CP, but none of the authors are the same. Are we going to do something about it, or just call this CP2011c?

doc/source/refs.rst
```diff
@@ -21,6 +21,10 @@ References
...
+   methods in signal processing.* In: Fixed-point algorithms for
+   inverse problems in science and engineering, Springer, 2011.
+
 .. [CP2011a] Chambolle, Antonin and Pock, Thomas. *A First-Order
```
kohr-h: While we're at it: Antonin -> A, Thomas -> T
Forgot it back then

adler-j:
Short comment regarding commit messages:

Commits like `Maint: Update according to review` are essentially no different at all to `Maint:`. Commit messages should be stateless, that is, not implicitly refer other commits or parts of the development process (such as discussions or pull requests). It is better to write something like `Maint: style fixes to proximals`, then at least a user knows what parts of the code were touched.

If you are only making temporary commits that you later plan to rebase before a merge, call them something like `TEMPROARY: updates according to review`, that way we wont forget to rebase them.

odl/solvers/advanced/proximal_operators.py
aringh: I apparently forgot to change this one. dot or no dot after ``F``?

adler-j: This has been discussed N times, but I am personally in favor of using a dot every time we have full sentences. Numpy does the same in their documentation, if that's a guideline.

doc/source/refs.rst
kohr-h: We have to live with this ambiguity, I guess. Next time, you have two authors which start with "Com" and "Pes" which are not the ones we have here. Or, even worse, think of Chinese authors -> unambiguous abbreviation totally impossible. :-)

Go for CP2011c.

odl/solvers/advanced/proximal_operators.py
kohr-h: No dot since it's not a full sentence (has no predicate)

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,183 @@ def cconj_prox_factory(step_size):
...
+        Scaling parameter
+
+    Returns
+    -------
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [ComPes2011]_.
```
kohr-h: Make a `References` section and just add `[CP2011c]`, no extra text.

kohr-h: If you want, you can add a citation in the text above.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,183 @@ def cconj_prox_factory(step_size):
...
+        prox[s * F( . * a)](x) = 1/a * prox[s * a^2 * F ](x * a)
+
+    where ``a`` scaling parameter, and ``s`` is the step size.
+
+    Parameters
+    ----------
+    prox_factory : `callable`
+        A factory function that, when called with a step size, returns the
+        proximal operator of ``F``
+    a : non-zero `float`
```
kohr-h: Call it `scaling` or something more telling

odl/solvers/advanced/proximal_operators.py
adler-j: Darn seems I need to take a grammar class to be able to write doc, you are indeed correct.

aringh:
Note taken on commit messages. I have tried to update them somewhat.

aringh:
Forgot to change the doc when renaming the input, but I have fixed that now as well. Ready to merge?

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # Element in the image space where the proximal operator is evaluated
+    translation = space.element(np.arange(10, 20))
+
+    # Parameters
+    lam = 1 / 3
+    step_size = 2.4
+
+    # Factory function returning the proximal operators
+    prox_factory = odls.proximal_l2_squared(space, lam=lam, g=None)
```
adler-j: `g=None` not needed

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # Create an element in the space, in which to evaluate the proximals
+    x = space.element(np.arange(-5, 5))
+
+    # Explicit computation:
+    # (x + 2 * step_size * lam * translation) / (1 + 2 * ste_size * lam)
+    x_verify = ((x + 2 * step_size * lam * translation) /
+                (1 + 2 * step_size * lam))
+
+    print(x_verify)
```
adler-j: remove print

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # Create an element in the space, in which to evaluate the proximals
+    x = space.element(np.arange(-5, 5))
+
+    # Uses the fact that 1/2||x 1/a||^2 = 1/2a^2 ||x||^2
+    assert all_almost_equal(prox_1(x), prox_2(x), places=PLACES)
+
+
+def test_proximal_arg_scaling_zero():
+    """Test for the proximal of scaling: prox[f(. * 1/a)]"""
```
adler-j: `prox[F(. * a)]`

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+    step_size = 2.0
+    scaling_param = 0
+
+    # Initialize proximal operators
+    prox = odls.proximal_arg_scaling(prox_factory, scaling_param)(step_size)
+
+    # Create an element in the space, in which to evaluate the proximals
+    x = space.element(np.arange(-5, 5))
+
+    # Uses the fact that 1/2||x 1/a||^2 = 1/2a^2 ||x||^2
```
adler-j: This doesnt realy test very much, perhaps use another proximal as an example here?

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def test_proximal_quadratic_perturbation():
+    """Test for the proximal of quadratic perturbation"""
+
+    # Image space
+    space = odl.uniform_discr(0, 1, 10)
+
+    # The parameter for the quadratic perturbation
+    a = 4.6
```
adler-j: I like that you have "random" numbers for these, makes the chance of it working because of luck much smaller, with that said, you could probably use the same one across all of these, currently it feels like these were selected to something specific.

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+                (1 + 2 * step_size * lam))
+
+    print(x_verify)
+    print(prox(x))
+
+    assert all_almost_equal(prox(x), x_verify, places=PLACES)
+
+
+def test_proximal_arg_scaling():
+    """Test for the proximal of scaling: prox[f(. * 1/a)]"""
```
adler-j: prox[F(. * a)]

test/solvers/advanced/proximal_utils_test.py
```diff
@@ -0,0 +1,148 @@
...
+    scaling_param = 3.4
+
+    # Initialize proximal operators
+    prox_1 = prox_factory(step_size * scaling_param**2)
+    prox_2 = odls.proximal_arg_scaling(prox_factory, scaling_param)(step_size)
+
+    # Create an element in the space, in which to evaluate the proximals
+    x = space.element(np.arange(-5, 5))
+
+    # Uses the fact that 1/2||x 1/a||^2 = 1/2a^2 ||x||^2
```
adler-j: Needs to be more complicated, this works for any `a` so doesn't really test that this works.

adler-j: Also, the `1/2` isnt part of the defintion.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -229,12 +229,9 @@ def test_proximal_convconj_l2_sq_wo_data():
     x_arr = np.arange(-5, 5)
     x = space.element(x_arr)
 
-    # Create data
-    g = space.element(-2 * x_arr)
-
     # Factory function returning the proximal operator
     lam = 2
-    prox_factory = proximal_cconj_l2_squared(space, lam=lam, g=g)
+    prox_factory = proximal_cconj_l2_squared(space, lam=lam, g=None)
```
adler-j: remove `g=None`

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,184 @@ def cconj_prox_factory(step_size):
...
+    """
+
+    if a < 0:
+        raise ValueError('scaling parameter {} not non-negative'.format(a))
+
+    if not isinstance(u, LinearSpaceVector):
+        raise TypeError('vector {!r} not a LinearSpaceVector instance.'
+                        ''.format(u))
+
+    if not u.space == prox_factory(1).domain:
```
adler-j: We dont do this in the others and I'd skip it here. If we add it we'll add to all of them. This could possibly incur a cost, or even a crash (perhaps 1 is not a valid value?)

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,184 @@ def cconj_prox_factory(step_size):
...
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [CP2011c]_.
+    """
+
+    scaling = float(scaling)
+
```
adler-j: remove blank row

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,184 @@ def cconj_prox_factory(step_size):
...
+    if scaling == 0:
+        return proximal_zero(prox_factory(1.0).domain)
```
adler-j: dislike this way of fishing out the space, but perhaps we'll have to live with it. The option is to move this into the `arg_scaling_prox_factory`.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,184 @@ def cconj_prox_factory(step_size):
...
+
+        Parameters
+        ----------
+        step_size : positive `float`
+            Step size parameter
+
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * F( . * 1/a)`` where ``s`` is the
```
adler-j: `F( . * a)`

test/solvers/advanced/proximal_utils_test.py
adler-j: Also reads `1/a`

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,184 @@ def cconj_prox_factory(step_size):
...
+        Returns
+        -------
+        proximal : `Operator`
+            The proximal operator of ``s * (F(x) + a * ||x||^2 + <u,x>)``,
+            where ``s`` is the step size
+        """
+        const = 1.0 / np.sqrt(step_size * 2.0 * a + 1)
+        prox = proximal_arg_scaling(prox_factory, const)(step_size)
+        return (const * prox *
+                (ScalingOperator(u.space, const) +
```
adler-j: More thought put into this says that this is probably the best

```python
ResidualOperator(ScalingOperator(u.space, const), step_size * const * u)
```

odl/solvers/advanced/proximal_operators.py
```diff
@@ -124,14 +133,184 @@ def cconj_prox_factory(step_size):
...
+    prox : `callable`
+        Factory for the proximal operator to be initialized
+
+    Notes
+    -----
+    For reference on the identity used, see [CP2011c]_. Note that this identity
+    is not the exact one given in the reference, but was recalculated for
+    arbitrary step lengths.
+    """
+
```
adler-j: a = float(a)

odl/solvers/advanced/proximal_operators.py
adler-j: Also, if we want to keep this, we would write it as

```python
f u not in prox_factory(1).domain:
```

adler-j:
Did some further review, mostly style, but some of the tests seem excessively simple (tests cases that are independent of the input parameters is a bug waiting to happen).

odl/solvers/advanced/proximal_operators.py
aringh: You can move it in to the factory. Then you have to check each time the factory is called if ```scaling == 0 ``` and return the identity operator. This was just a way to only have to check it one time.

I'll leave it here for now.

aringh:
Unfortunately I will be gone for about a week, so if there is anything else before I'll have to take it when I come back.

adler-j:
Minor changes and rebased. Will merge after travis.

PR #424, opened 2016-06-01
ENH: Add __len__ to RectPartition

### LINKED ISSUE #423: RectPartition has no length
Sometimes we need to know the number of partitions. Trying to execute the `len` command on a partition gives an error message;
```
TypeError: object of type 'RectPartition' has no len()
```

ozanoktem:
Closes #423

### DIFF
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -178,6 +178,10 @@ class RectPartition(object):
         """Total number of cells, equal to ``self.grid.size``."""
         return self.grid.size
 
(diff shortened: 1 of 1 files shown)


adler-j:
Disagree with this fix, this does not agree with numpy:

```python
>>> arr = np.array([[1, 2], 
                    [3, 4]])
>>> len(arr)
2
>>> arr[0]
array([1, 2])
```

Our behaviour should mirror this. I made another pull request that does this change systematically, see #425.

PR #425, opened 2016-06-01
ENH: add len to several places

adler-j:
See #424, but this adds it to several more places and follows numpy convention.

### DIFF
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -242,16 +242,6 @@ class DiscretizedSetVector(NtuplesBaseVector):
         """Structure for data storage."""
         return self._ntuple
 
-    @property
-    def dtype(self):
-        """type of data storage."""
-        return self.ntuple.dtype
-
-    @property
-    def size(self):
-        """size of data storage."""
-        return self.ntuple.size
-
     def copy(self):
         """Create an identical (deep) copy of this vector."""
         return self.space.element(self.ntuple.copy())
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -305,6 +305,10 @@ class ProductSpace(LinearSpace):
         """The number of factors."""
         return self._size
 
+    def __len__(self):
+        """Return ``len(self)``."""
+        return self.size
+
     @property
     def shape(self):
         """Number of spaces per axis."""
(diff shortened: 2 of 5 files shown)


odl/discr/lp_discr.py
```diff
@@ -188,6 +188,13 @@ def size(self):
         """Total number of underlying partition cells."""
         return self.partition.size
 
+    def __len__(self):
+        """Return ``len(self)``.
+
+        Total number of underlying partition cells.
+        """
+        return self.size
```
kohr-h: This should return `len(self.partition)`, no? `DiscreteLp` "looks like" the partition it's defined on.

odl/space/pspace.py
```diff
@@ -305,6 +305,10 @@ def size(self):
         """The number of factors."""
         return self._size
 
+    def __len__(self):
+        """Return ``len(self)``."""
+        return self._size
```
kohr-h: w/o underscore

kohr-h:
Is this interpretation of the "length" of a discretization obvious? IMO, it is only "kind of" obvious. The alternative, which we should consider an option, is not to implement it at all for `Discretization` classes. For an element, on the other hand, the interpretation is very clear.

odl/discr/lp_discr.py
adler-j: Actually, it does not. The `[ ]` operator of `DiscreteLp` is in fact linear.

adler-j:
It is the only valid interpretation however, otherwise stuff like `list(vector)` will fail.

The only obvious fix (and something we need to do at some point I suppose) is to start supporting multi-indexes for `DiscreteLpVector`, then we could change that.

adler-j:
Additional note:

I just did some checking and `__len__` already exists for `DiscreteLp`, (by inheritance from `BaseNtuple`)

kohr-h:
> It is the only valid interpretation however, otherwise stuff like list(vector) will fail.

Vectors are perfectly fine, I was more concerned about the space itself.



> The only obvious fix (and something we need to do at some point I suppose) is to start supporting multi-indexes for DiscreteLpVector, then we could change that.

That would be the best outcome, but that depends on a supporting CUDA back-end I guess. But actually, you're right, `DiscreteLp` should be considered flat for now since multiple indices are not supported.

> I just did some checking and __len__ already exists for DiscreteLp, (by inheritance from BaseNtuple). It returns the same as size.

Then it's probably best to do nothing right now and inherit, and change behavior as soon as multi-indexing is available.

odl/discr/discretization.py
```diff
@@ -242,16 +242,6 @@ def ntuple(self):
         """Structure for data storage."""
         return self._ntuple
 
-    @property
```
adler-j: These follow from inheritance

adler-j:
Updated according to comments, merge?

kohr-h:
LGTM

Issue #408, opened 2016-05-29
Documentation review


kohr-h:
Some parts of the documentation (especially the older stuff) are a bit rough and not on the same level of clarity and completeness as the newer additions. So a check and rewrite of those parts wouldn't hurt. While we're at it, there are a few points we should address as well. From the top of my head, probably more to come while going through the code:

- **Magic methods:** Documentation should be uniform. Since `_call` in `Operator` has a similar functionality, it should follow the same pattern.
- **`__init__`:** I think it suffices to write `Initialize a new instance.` instead of naming the operator explicitly. Also more friendly towards subclassing.
- **Math stuff:** In the math-heavy operators and functions, we should not try to write "pseudo-formulas". Instead, we should reduce everything before the `Parameters` section to bare minimum and leave the mathematical background to a separate `Notes` section, where it's allowed to use LaTeX markup. This is the same approach as NumPy/SciPy uses, and I think it's good. But even those math parts should be kept very brief. If more explanation is necessary, a section in the `Mathematical Background` part of the main documentation should be created instead.
- **Doctests:** I think our examples are good. But they should also follow the same style rules as the rest of the code. This includes PEP8 and variable name conventions (e.g. not `X` for a space). If lines get too long, they can be broken with a `...` in the beginning of the next line. The doctests are the code pieces that users are most exposed to, so they should be good in style.
- **Application of back-ticks:** Currently we overuse back-ticks IMO. There is no need to use them for each float or list occurrence. I think we should restrict them to internal links. Otherwise, the online documentation looks cluttered.
- **Remove leading "The":** Instead of writing e.g. "The object to test for equality", we should simply write "Object to test for equality". Less noise.
- **Consistency with trailing dots:** ~~The rule is to set a dot after a proper sentence (with predicate), or if more sentences follow.~~ We go for the solution to always put a dot at the end, see #432.
- **Properties:** Since properties look like attributes, the docstring should not say "Return ...", but rather what it *is*.

Feel free to edit and add stuff to the list.

adler-j:
Does this issue have a definition of done or a reasonable timeframe in which something of this will be worked on? Otherwise I say we classify it as ongoing work. Doing an actual readthrough of the doc with fixes is way beyond our current scope.

kohr-h:
There's actually a branch for this with updates for a couple of modules. I used to spend some time on it once in a while, but not so much lately. I'd still say we keep this issue open until that branch has gone in. After that it's ongoing.

[...]

PR #427, opened 2016-06-01
DOC: fixes to capitalization

adler-j:
Minor doc fixes to unify doc, basically going through the code using regexpr, work in progress.

### DIFF
--- a/odl/discr/diff_ops.py
+++ b/odl/discr/diff_ops.py
@@ -87,7 +87,7 @@ class PartialDerivative(PointwiseTensorFieldOperator):
             accuracy in the interior.
         """
         if not isinstance(space, DiscreteLp):
-            raise TypeError('space {!r} is not a DiscreteLp instance.'
+            raise TypeError('`space` {!r} is not a DiscreteLp instance'
                             ''.format(space))
 
         # Method is affine if nonzero padding is given.
@@ -166,8 +166,8 @@ class PartialDerivative(PointwiseTensorFieldOperator):
     def adjoint(self):
         """Return the adjoint operator."""
         if not self.is_linear:
-            raise ValueError('Operator with nonzero padding_value ({}) is not'
-                             ' linear and has no adjoint.'
+            raise ValueError('operator with nonzero padding_value ({}) is not'
+                             ' linear and has no adjoint'
                              ''.format(self.padding_value))
 
         return -PartialDerivative(self.domain, self.axis,
@@ -235,18 +235,18 @@ class Gradient(PointwiseTensorFieldOperator):
         True
         """
         if domain is None and range is None:
-            raise ValueError('either domain or range must be specified.')
+            raise ValueError('either `domain` or `range` must be specified')
 
         if domain is None:
             if not isinstance(range, ProductSpace):
-                raise TypeError('range {!r} is not a ProductSpace instance.'
+                raise TypeError('`range` {!r} is not a ProductSpace instance'
                                 ''.format(range))
             domain = range[0]
 
         if range is None:
             if not isinstance(domain, DiscreteLp):
-                raise TypeError('domain {!r} is not a `DiscreteLp` '
-                                'instance.'.format(domain))
+                raise TypeError('`domain` {!r} is not a `DiscreteLp` '
+                                'instance'.format(domain))
             range = ProductSpace(domain, domain.ndim)
 
         linear = not (padding_method == 'constant' and padding_value != 0)
@@ -348,8 +348,8 @@ class Gradient(PointwiseTensorFieldOperator):
         this operator.
         """
         if not self.is_linear:
-            raise ValueError('Operator with nonzero padding_value ({}) is not'
-                             ' linear and has no adjoint.'
+            raise ValueError('operator with nonzero padding_value ({}) is not'
+                             ' linear and has no adjoint'
                              ''.format(self.padding_value))
 
         return - Divergence(domain=self.range, range=self.domain,
@@ -416,17 +416,17 @@ class Divergence(PointwiseTensorFieldOperator):
         True
         """
         if domain is None and range is None:
-            raise ValueError('either domain or range must be specified.')
+            raise ValueError('either `domain` or `range` must be specified')
 
         if domain is None:
             if not isinstance(range, DiscreteLp):
-                raise TypeError('range {!r} is not a DiscreteLp instance.'
+                raise TypeError('`range` {!r} is not a DiscreteLp instance'
                                 ''.format(range))
             domain = ProductSpace(range, range.ndim)
 
         if range is None:
             if not isinstance(domain, ProductSpace):
-                raise TypeError('domain {!r} is not a ProductSpace instance.'
+                raise TypeError('`domain` {!r} is not a ProductSpace instance'
                                 ''.format(domain))
             range = domain[0]
 
@@ -525,8 +525,8 @@ class Divergence(PointwiseTensorFieldOperator):
         the method and padding.
         """
         if not self.is_linear:
-            raise ValueError('Operator with nonzero padding_value ({}) is not'
-                             ' linear and has no adjoint.'
+            raise ValueError('operator with nonzero padding_value ({}) is not'
+                             ' linear and has no adjoint'
                              ''.format(self.padding_value))
 
         return - Gradient(domain=self.range, range=self.domain,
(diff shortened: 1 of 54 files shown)


adler-j:
Probabily the most booring review anyone has ever done, but needs a review. Basically fixes some style issues in the doc.

odl/discr/diff_ops.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, space, axis=0, method='forward', padding_method=None,
             accuracy in the interior.
         """
         if not isinstance(space, DiscreteLp):
-            raise TypeError('space {!r} is not a DiscreteLp instance.'
+            raise TypeError('`space` {!r} is not a DiscreteLp instance.'
```
kohr-h: General question: do we want back-ticks in error messages or not? Consistency needed.

kohr-h: I'm tending towards not using them since they require to exactly give the name of the parameter, which is unfortunate for subclassing. For example, in geometries, when we give a partition `motion_part` for the motion parameters, and in a subclass, the parameter is named `angle_part`, then consider these two error messages (made up):
```
ValueError: motion parameter partition must be two-dimensional.
```
against this:
```
ValueError: `motion_part` must be two-dimensional.
```
I'd say the first one can be more easily reconciled with a different parameter name since natural language is used. The second message is in full conflict with a different parameter name.

adler-j: I think it is good (obviously, since i did all of this) because it highlights what is a word and what is a parameter. Reduces confusion quite a bit.

kohr-h:
An argument against capitalization: If the sentence starts with a parameter name, it makes no sense to capitalize it. That results in a perceived inconsistency in the error messages. Starting non-capitalized is always consistent.

odl/discr/diff_ops.py
kohr-h: That implies that we *have to* rewrite the error messages for subclasses which use a different parameter name than their parents. We'll get some duplicate checks there. We can decide to do it that way.

adler-j: I dont see why we need to rewrite them in subclasses, users get the full callstack. This is not intended to be the easiest to read error message, it's to allow users to fix their problem in the shortest possible time.

If we use "common language", users may need to reason about which parameter is intended. For example in Chambolle-Pock "step length for the primal variable" or w/e is i.m.o much less clear than `tau` (or `sigma`, even I dont know which one it is).

Another point for this is that it is trivial to refactor if we decide to change the wording or the paramter, just find replace in the file. I found several of this type of error when doing this.

Final comment, if we decide to not capitalize the first letter, we should remove the dots at the end as well, it makes no sense to have dots but not capitalization.

adler-j:
This obviously needs a-bit of discussion. Lets try to think it through properly before we do this merge, fixing it is not that hard. All I know is that we should be consistent.

odl/discr/diff_ops.py
kohr-h: Okay, I buy that argument.

kohr-h:
Some examples for inspiration:
```python
>>> list(1)
...
TypeError: 'int' object is not iterable

>>> [1, 2, 3][4]
...
IndexError: list index out of range

>>> np.ones(1)[2]
...
IndexError: index 2 is out of bounds for axis 0 with size 1

>>> np.ones(1)[lambda x: 1]
...
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

kohr-h:
Not sure how consistent it is, but the pattern seems to be

- no capitalization
- back-ticks are ok
- no dot at the end (even after full sentences)

adler-j:
Agree with the above. I'd make a simple addition

* Multi-sentence error messages should use both capitalization and punctuation.

This shouldn't happen particularly often since error messages should be short.

kohr-h:
Sounds good.

adler-j:
Updated according to comments, needs a review now

adler-j:
Could really use a review, merging this into master becomes more and more tiresome since this touches basically all files.

odl/discr/diff_ops.py
```diff
@@ -774,26 +774,27 @@ def finite_diff(f, axis=0, dx=1.0, method='forward', out=None, **kwargs):
...
     if f_arr.shape[axis] < 2:
-        raise ValueError('In axis {}: at least two elements required, got {}.'
+        raise ValueError('in axis {}: at least two elements required, got {}'
                          ''.format(axis, f_arr.shape[axis]))
 
     if axis < 0:
         axis += ndim
     if not (0 <= axis < ndim):
-        raise IndexError('axis {} outside the valid range 0 ... {}'
+        raise IndexError('`axis` {} outside the valid range 0 ... {}'
```
kohr-h: I write this here, but actually it's a separate issue.

This error message is problematic (I think I wrote it) since it contradicts the code and the doc which allows negative indices. We should adhere to a style similar to NumPy's:
```python
>>> np.ones(1)[2]
IndexError: index 2 is out of bounds for axis 0 with size 1
>>> np.ones(1)[-2]
IndexError: index -2 is out of bounds for axis 0 with size 1
>>> np.sum(np.ones(1), axis=2)
ValueError: 'axis' entry is out of bounds
```

We could simply say
```
`axis` {} is out of bounds
```

odl/discr/diff_ops.py
```diff
@@ -774,26 +774,27 @@ def finite_diff(f, axis=0, dx=1.0, method='forward', out=None, **kwargs):
...
                          ''.format(axis, ndim - 1))
 
     dx, dx_in = float(dx), dx
     if dx <= 0 or not np.isfinite(dx):
-        raise ValueError("step length {} not positive.".format(dx_in))
+        raise ValueError("`dx` {} not positive".format(dx_in))
```
kohr-h: we should write "is not positive". Actually better is
```
ValueError("`dx` must be positive, got {}".format(dx_in))
```

odl/discr/diff_ops.py
```diff
@@ -774,26 +774,27 @@ def finite_diff(f, axis=0, dx=1.0, method='forward', out=None, **kwargs):
...
 
     method, method_in = str(method).lower(), method
     if method not in _SUPPORTED_DIFF_METHODS:
-        raise ValueError('method {} is not understood'.format(method_in))
+        raise ValueError('`method` {} is not understood'.format(method_in))
```
kohr-h: Either ~~is~~ or "was not understood". I prefer "blabla not understood"

odl/discr/diff_ops.py
```diff
@@ -774,26 +774,27 @@ def finite_diff(f, axis=0, dx=1.0, method='forward', out=None, **kwargs):
...
 
     padding_method = kwargs.pop('padding_method', None)
     if padding_method not in _SUPPORTED_PADDING_METHODS:
-        raise ValueError('padding value {} not valid'.format(padding_method))
+        raise ValueError('`padding_method` {} not valid'
```
kohr-h: not understood

odl/discr/diff_ops.py
```diff
@@ -804,23 +805,22 @@ def finite_diff(f, axis=0, dx=1.0, method='forward', out=None, **kwargs):
         else:
             edge_order = 1
     elif edge_order not in (1, 2):
-            raise ValueError('edge order {} not valid'.format(edge_order))
+            raise ValueError('`edge_order` {} not valid'.format(edge_order))
```
kohr-h:
```
ValueError('`edge_order` must be 1 or 2, got {}'.format(edge_order))
```

odl/discr/discr_mappings.py
```diff
@@ -78,27 +78,27 @@ def __init__(self, map_type, fset, partition, dspace, linear=False,
         """
         map_type_ = str(map_type).lower()
         if map_type_ not in ('sampling', 'interpolation'):
-            raise ValueError('mapping type {!r} not understood.'
+            raise ValueError('`map_type` {!r} not understood'
```
kohr-h: ~~!r~~

odl/discr/discr_mappings.py
```diff
@@ -423,7 +426,7 @@ def _call(self, x, out=None):
         """
         # TODO: pass reasonable options on to the interpolator
         def nearest(arg, out=None):
-            """Interpolating function with vectorization."""
+            """Interpolating function with vectorization"""
```
kohr-h: With dot

kohr-h:
I'm sure the changes make sense. Run the tests, build the doc and check for errors. I'll anyway go through large parts again in fixing #408, so go ahead.

adler-j:
Merged after quite some work. Big branches like this need to go in faster :*(

PR #429, opened 2016-06-01
ENH: optimzize repeated scalar times operator

adler-j:
Been wanting to do this for a while, minor optimization in case of several multiplications.


kohr-h:
Looking good, just fix the typo in the commit message. Add "multiplication" at the end of the message?

PR #430, opened 2016-06-01
API: change Partial to Callback

adler-j:
Changes any mention of `Partial` to the more standardized `Callback`

This agrees precisely with how scipy does it, down to naming, for example [scipy.optimize.minimize](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.optimize.minimize.html):

> callback : callable, optional
> Called after each iteration, as callback(xk), where xk is the current parameter vector.

The interface of all solvers has also been changed so only `callable` is required, the extra features provided by `SolverCallback` (such as the `&` operator) are thus only optional and not needed by the solvers.

### DIFF
--- a/odl/solvers/advanced/chambolle_pock.py
+++ b/odl/solvers/advanced/chambolle_pock.py
@@ -30,7 +30,6 @@ standard_library.install_aliases()
 import numpy as np
 
 from odl.operator.operator import Operator
-from odl.solvers.util import Partial
 
 
 __all__ = ('chambolle_pock_solver',)
@@ -97,8 +96,8 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
         variable relaxation parameter and step sizes with ``tau`` and
         ``sigma`` as initial values. Requires G or F^* to be uniformly
         convex. Default: `None`
-    partial : `Partial`, optional
-        If not `None` the `Partial` instance(s) are executed in each
+    callback : `callable`, optional
+        If not `None` the ``callback`` instance(s) are executed in each
         iteration, e.g. plotting each iterate. Default: `None`
     x_relax : element in the domain of ``op``, optional
         Required to resume iteration. If `None` it is a copy of the primal
@@ -166,11 +165,11 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
         else:
             gamma = float(gamma)
 
-    # Partial object
-    partial = kwargs.pop('partial', None)
-    if partial is not None and not callable(partial):
-        raise TypeError('partial {} is not an instance of {}'
-                        ''.format(op, Partial))
+    # Callback object
+    callback = kwargs.pop('callback', None)
+    if callback is not None and not callable(callback):
+        raise TypeError('`callback` {} is not ``callable``'
+                        ''.format(callback))
 
     # Initialize the relaxation variable
     x_relax = kwargs.pop('x_relax', None)
@@ -217,8 +216,8 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
         # Over-relaxation in the primal variable x
         x_relax.lincomb(1 + theta, x, -theta, x_old)
 
-        if partial is not None:
-            partial(x)
+        if callback is not None:
+            callback(x)
 
 
 if __name__ == '__main__':
--- a/odl/solvers/advanced/chambolle_pock.py
+++ b/odl/solvers/advanced/chambolle_pock.py
@@ -168,7 +168,7 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
     # Callback object
     callback = kwargs.pop('callback', None)
     if callback is not None and not callable(callback):
-        raise TypeError('`callback` {} is not ``callable``'
+        raise TypeError('`callback` {} is not `callable`'
                         ''.format(callback))
 
     # Initialize the relaxation variable
--- a/odl/solvers/findroot/newton.py
+++ b/odl/solvers/findroot/newton.py
@@ -31,7 +31,7 @@ __all__ = ('bfgs_method', 'broydens_first_method', 'broydens_second_method')
 # TODO: update all docs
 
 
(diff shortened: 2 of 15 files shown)


odl/solvers/advanced/chambolle_pock.py
```diff
@@ -166,11 +165,11 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
...
 
-    # Partial object
-    partial = kwargs.pop('partial', None)
-    if partial is not None and not callable(partial):
-        raise TypeError('partial {} is not an instance of {}'
-                        ''.format(op, Partial))
+    # Callback object
+    callback = kwargs.pop('callback', None)
+    if callback is not None and not callable(callback):
+        raise TypeError('`callback` {} is not ``callable``'
```
kohr-h: Remove double backticks

odl/solvers/findroot/newton.py
```diff
@@ -67,8 +67,8 @@ def bfgs_method(grad, x, line_search, niter=1, partial=None):
         Strategy to choose the step length
     niter : `int`, optional
         Number of iterations
-    partial : `Partial`, optional
-        Object executing code per iteration, e.g. plotting each iterate
+    callback : `callable`, optional
+        Object executing code per iteration, e.g. plotting each iterate.
```
kohr-h: No dot

odl/solvers/findroot/newton.py
```diff
@@ -136,8 +136,8 @@ def broydens_first_method(grad, x, line_search, niter=1, partial=None):
...
```
kohr-h: No dot

odl/solvers/util/callback.py
```diff
@@ -15,30 +15,27 @@
...
+           'CallbackPrintNorm', 'CallbackShow')
 
-__all__ = ('Partial', 'StorePartial', 'ForEachPartial', 'PrintTimingPartial',
-           'PrintIterationPartial', 'PrintNormPartial', 'ShowPartial')
 
-
-class Partial(with_metaclass(ABCMeta, object)):
+class SolverCallback(object):
 
     """Abstract base class for sending partial results of iterations."""
```
kohr-h: Could also use an update

odl/solvers/util/callback.py
```diff
@@ -53,62 +50,62 @@ def __call__(self, result):
     def __and__(self, other):
         """Return ``self & other``
 
-        Compose partials, calls both in sequence.
+        Compose callbacks, calls both in sequence.
 
         Parameters
         ----------
-        other : `Partial`
+        other : `Partial` or `callable`
```
kohr-h: Remaining Partial

odl/solvers/util/callback.py
```diff
@@ -53,62 +50,62 @@ def __call__(self, result):
...
     def __repr__(self):
         """Return ``repr(self)``"""
         return '{}()'.format(self.__class__.__name__)
 
 
-class AndPartial(Partial):
+class _CallbackAnd(SolverCallback):
 
-    """Partial used for combining several partials"""
+    """Callback used for combining several callbacks"""
```
kohr-h: Dot at the end (only applies to one-line summaries, not in parameter lists)
(Sorry, not my rule: see [PEP257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings))

odl/solvers/util/callback.py
```diff
@@ -53,62 +50,62 @@ def __call__(self, result):
...
 
     def __init__(self, *callbacks):
         """Initialize an instance.
 
         Parameters
         ----------
-        *callbacks : `callable` or `Partial`'s
-            Partials to be called in sequence as listed.
+        *callbacks : `callable` or `SolverCallback`'s
```
kohr-h: This is written as `callback1,...,callbackN` (not a hard rule, but the Numpy way; I think it makes more clear that the arguments are given individually, not packed)

kohr-h: And remove the `'s` at the end, for comparison:
```
x1, x2: float
```

odl/solvers/util/callback.py
```diff
@@ -53,62 +50,62 @@ def __call__(self, result):
...
-        partials = [c if isinstance(c, Partial) else ForEachPartial(c)
-                    for c in callbacks]
+        callbacks = [c if isinstance(c, SolverCallback) else CallbackApply(c)
+                     for c in callbacks]
 
-        self.partials = partials
+        self.callbacks = callbacks
 
     def __call__(self, result):
         """Apply all partials to result."""
```
kohr-h: partials

odl/solvers/util/callback.py
```diff
@@ -53,62 +50,62 @@ def __call__(self, result):
...
     def __repr__(self):
         """Return ``repr(self)``"""
-        return ' & '.join('{}'.format(p) for p in self.partials)
+        return ' & '.join('{}'.format(p) for p in self.callbacks)
 
 
-class StorePartial(Partial):
+class CallbackStore(SolverCallback):
 
     """Simple object for storing all partial results of the solvers.
```
kohr-h: partial results -> iterates

odl/solvers/util/callback.py
```diff
@@ -266,7 +263,7 @@ def __call__(self, result):
         self.iter += 1
 
 
-class ShowPartial(Partial):
+class CallbackShow(SolverCallback):
 
     """Show the partial result."""
```
kohr-h: partial result -> iterate

kohr-h:
Do a "Find in files: partial" and replace the remnants from "old times". Looking good otherwise. Nice change.

odl/solvers/util/callback.py
adler-j: Np, I just spend an hour getting this fixed throughout the codebase, searching by regexpr for `[a-zA-Z]\"\"\"` found like 100 of these :/

odl/solvers/findroot/newton.py
adler-j: Made a separate issue on this #432 will leave this and make a general go at this once that is settled.

PR #433, opened 2016-06-02
API: change `dom`/`ran` to `domain`/`range`.

### LINKED ISSUE #416: dom vs domain as operator argument
Several operator take their domain as an argument. This is often done done by taking a `domain` OR `dom` parameter. The same applies to the `range` OR `ran`.

We need to decide on one way of doing this:

* `domain` and `range`

OR

* `dom` and `ran`

adler-j:
Closes #416 

Basically a style update.


kohr-h:
I'm not going to go through it in detail. The tests run, so go for it.

PR #434, opened 2016-06-02
Indexing of geometries

ozanoktem:
This PR implements indexing of geometries via `__getitem__`. Should work fine, but some tests would not hurt.

### DIFF
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -541,6 +541,40 @@ class TensorGrid(Set):
                     self.coord_vectors[idx:])
         return TensorGrid(*new_vecs)
 
+    def remove(self, axes):
+        """Remove ``axes`` and return the reduced grid.
+
+        Note that no changes are made in-place.
+
+        Parameters
+        ----------
+        axes : index expression
+            Object determining which parts of the grid to remove
+
+        Returns
+        -------
+        newgrid : `TensorGrid`
+            The reduced grid
+
+        Examples
+        --------
+        >>> g1 = TensorGrid([0, 1], [-1, 2, 5])
+        >>> g1.remove(1)
+        TensorGrid([0.0, 1.0])
+        >>> g1.remove(0)
+        TensorGrid([-1.0, 2.0, 5.0])
+        See Also
+        --------
+        append
+        """
+        index_arr = np.arange(self.ndim)
+        mask = np.ones_like(index_arr, dtype=bool)
+        mask[axes] = False
+        remaining_axes = list(index_arr[mask])
+
+        new_vecs = [self.coord_vectors[i] for i in remaining_axes]
+        return TensorGrid(*new_vecs)
+
     def append(self, other):
         """Insert grid ``other`` at the end.
 
@@ -1045,6 +1079,43 @@ class RegularGrid(TensorGrid):
 
         else:
             raise TypeError('{!r} is not a TensorGrid instance'.format(other))
+
+    def remove(self, axes):
+        """Remove ``axes`` and return the reduced grid.
+
+        Note that no changes are made in-place.
+
+        Parameters
+        ----------
+        axes : index expression
+            Object determining which parts of the grid to remove
+
+        Returns
+        -------
+        newgrid : `RegularGrid`
+            The reduced grid
+
+        Examples
+        --------
+        >>> g = RegularGrid([0, 0, 0], [1, 0, 1], (5, 1, 5))
+        >>> g.remove(1)
+        RegularGrid([0.0, 0.0], [1.0, 1.0], (5, 5))
+        >>> g.remove(0)
+        RegularGrid([0.0, 0.0], [0.0, 1.0], (1, 5))
+        See Also
+        --------
+        append
+        """
+        index_arr = np.arange(self.ndim)
+        mask = np.ones_like(index_arr, dtype=bool)
+        mask[axes] = False
+        remaining_axes = index_arr[mask]
+
+        new_minpt = self.min_pt[remaining_axes]
+        new_maxpt = self.max_pt[remaining_axes]
+        new_shape = tuple(self.shape[i] for i in remaining_axes)
+
+        return RegularGrid(new_minpt, new_maxpt, new_shape)
 
     def squeeze(self):
         """Return the grid with removed degenerate dimensions.
--- a/odl/tomo/geometry/geometry.py
+++ b/odl/tomo/geometry/geometry.py
@@ -237,6 +237,17 @@ class Geometry(with_metaclass(ABCMeta, object)):
         """
         return self._implementation_cache
 
+    def _sliced_partitions(self, indices):
+        """Return motion and detector partition for a given slice."""
+        sliced_part = self.partition[indices]
+        motion_intv = sliced_part.set[0]
+        motion_grid = sliced_part.grid.remove(slice(1, None, None))
+        motion_part = RectPartition(motion_intv, motion_grid)
+        det_intv = sliced_part.set[1:]
+        det_grid = sliced_part.grid.remove(0)
+        det_part = RectPartition(det_intv, det_grid)
+        return motion_part, det_part
+
 
 class DivergentBeamGeometry(Geometry):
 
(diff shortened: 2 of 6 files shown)


odl/discr/grid.py
```diff
@@ -515,6 +515,40 @@ def insert(self, index, other):
                     self.coord_vectors[idx:])
         return TensorGrid(*new_vecs)
 
+    def remove(self, axes):
```
adler-j: Should be done using `__del__`, `remove` methods removes *by value*.

See
http://stackoverflow.com/questions/11520492/difference-between-del-remove-and-pop-on-lists

odl/discr/grid.py
```diff
@@ -1013,6 +1047,43 @@ def insert(self, index, other):
...
+            The reduced grid
+
+        Examples
+        --------
+        >>> g = RegularGrid([0, 0, 0], [1, 0, 1], (5, 1, 5))
+        >>> g.remove(1)
+        RegularGrid([0.0, 0.0], [1.0, 1.0], (5, 5))
+        >>> g.remove(0)
+        RegularGrid([0.0, 0.0], [0.0, 1.0], (1, 5))
+        See Also
```
adler-j: missing blank row

odl/tomo/geometry/geometry.py
```diff
@@ -237,6 +237,17 @@ def implementation_cache(self):
...
         return self._implementation_cache
 
+    def _sliced_partitions(self, indices):
+        """Return motion and detector partition for a given slice."""
+        sliced_part = self.partition[indices]
+        motion_intv = sliced_part.set[0]
+        motion_grid = sliced_part.grid.remove(slice(1, None, None))
+        motion_part = RectPartition(motion_intv, motion_grid)
+        det_intv = sliced_part.set[1:]
+        det_grid = sliced_part.grid.remove(0)
```
adler-j: Why do we change the detector at all?

odl/discr/grid.py
adler-j: __delitem__ is the correct one.

odl/tomo/geometry/geometry.py
```diff
@@ -237,6 +237,17 @@ def implementation_cache(self):
...
+        det_part = RectPartition(det_intv, det_grid)
+        return motion_part, det_part
```
adler-j: Cant we do this in one line?

```
return self.motion_part[1:], self.det_part
```

odl/tomo/geometry/geometry.py
```diff
@@ -237,6 +237,17 @@ def implementation_cache(self):
         """
         return self._implementation_cache
 
+    def _sliced_partitions(self, indices):
+        """Return motion and detector partition for a given slice."""
```
adler-j: Could use more info, what can indices be?

odl/tomo/geometry/spect.py
```diff
@@ -80,5 +81,39 @@ def det_radius(self):
         """Radius of the detector orbit."""
         return self._det_radius
 
+    @property
+    def orig_to_det_init(self):
+        """"Direction vector towards the initial detector reference point."""
+        return self._orig_to_det_init
+
+    def __getitem__(self, indices):
+        """Return ``self[indices]``."""
```
adler-j: The interpretation of this is nonobvious, what does it do?

odl/tomo/geometry/spect.py
```diff
@@ -80,5 +81,39 @@ def det_radius(self):
         """Radius of the detector orbit."""
         return self._det_radius
 
+    @property
+    def orig_to_det_init(self):
+        """"Direction vector towards the initial detector reference point."""
```
adler-j: Unit vector?

odl/tomo/geometry/geometry.py
```diff
@@ -237,6 +237,17 @@ def implementation_cache(self):
...
+        sliced_part = self.partition[indices]
+        motion_intv = sliced_part.set[0]
+        motion_grid = sliced_part.grid.remove(slice(1, None, None))
```
adler-j: Why not?

`motion_grid = sliced_part.grid[0]`

odl/discr/grid.py
```diff
@@ -515,6 +515,40 @@ def insert(self, index, other):
...
+            The reduced grid
+
+        Examples
+        --------
+        >>> g1 = TensorGrid([0, 1], [-1, 2, 5])
+        >>> g1.remove(1)
+        TensorGrid([0.0, 1.0])
+        >>> g1.remove(0)
+        TensorGrid([-1.0, 2.0, 5.0])
+        See Also
```
adler-j: Missing newline

odl/discr/grid.py
```diff
@@ -515,6 +515,40 @@ def insert(self, index, other):
...
         return TensorGrid(*new_vecs)
 
+    def remove(self, axes):
+        """Remove ``axes`` and return the reduced grid.
+
+        Note that no changes are made in-place.
+
+        Parameters
+        ----------
+        axes : index expression
```
adler-j: This indicates that an arbitrary index expression works, but I'm quite sure this only works with integers or lists of integers.

odl/tomo/geometry/geometry.py
kohr-h: Because we need to remove an *axis* from the grid, which is not what happens if you slice into it.

odl/discr/grid.py
kohr-h: I don't want to make grids mutable, so `__delitem__` is not an option for me. The list analogy is a valid point, so `remove_axes` would be an alternative.

odl/tomo/geometry/geometry.py
kohr-h: Not quite. The indices could slice into the detector part as well. What we actually need is a method to split the partition into motion partition and detector partition again.

adler-j: There has to be a nicer way of getting that done than this, we should try to do something general here. This code doesnt really convey what it wants to do at all.

odl/discr/grid.py
adler-j: Then at least call it `removeaxis` or something?

odl/tomo/geometry/geometry.py
adler-j: Another option would be to let `None` mean that nothing should be taken along that axis, then we could write

```python
return sliced_part[(0, [None]*(sliced_part.ndim-1))]), sliced_part[None, :]
```

kohr-h: Another suggestion: splitting at a certain index, i.e.
```python
part1, part2 = sliced_part.split_axes(1)
```

kohr-h: > Another option would be to let None mean that nothing should be taken along that axis, then we could write
> 
> return sliced_part[:, None], sliced_part[None, :]

That would be a nice solution if `None` would not mean the opposite in Numpy.

odl/tomo/geometry/spect.py
kohr-h: Yeah

odl/tomo/geometry/geometry.py
adler-j: Agree, it is a big negative.

adler-j: The more I think about it, the more I come to think that the `sliced_part.byaxis[0]` syntax would be the best. This is much more general than this specific problem and can be used for other things. It is also quite obvious what it means.

adler-j:
Any progress on this?

ozanoktem:
The geometry_getitem branch was motivated as a general framework to handle ordered sub-sets type of sub-division of data and operators. Currently this is put on hold since work on convolution is more important.

adler-j:
This functionality is actually rather nice, any chance for a quick push and get it in?

kohr-h:
We have a `byaxis` property in `RectPartition` that does the job, right? Perhaps I can make a quick hack, but as a new PR then.

adler-j:
Fine by me, but a shape to lose this.

kohr-h:
It's not gone, I have it in my fork.

kohr-h:
> a shape to lose this

You have typed the word "shape" too much recently :-D
I also find it hard to type "old" recently^^

PR #436, opened 2016-06-02
BUG: fix rotation by 90 degrees for 2d parallel

### LINKED ISSUE #435: Parallel 2d backprojections are rotated by 90 degrees
We [explicitly rotate](https://github.com/odlgroup/odl/blob/88fb4739cf7b79409507097ec9dc0265860a4d97/odl/tomo/backends/astra_setup.py#L359) by `pi/2` when translating from our geometry to ASTRA's convention. Either they changed something on their side so our fix becomes a bug, or our code was a bug in the first place.

kohr-h:
Closes #435 

Can you check if you get the correct result?

### DIFF
--- a/odl/tomo/backends/astra_setup.py
+++ b/odl/tomo/backends/astra_setup.py
@@ -355,8 +355,7 @@ def astra_projection_geometry(geometry):
         # TODO: change to parallel_vec when available
         det_width = geometry.det_partition.cell_sides[0]
         det_count = geometry.detector.size
-        # convention in 'astra_conebeam_2d_geom_to_vec' differs from ASTRA's
-        angles = geometry.angles - np.pi / 2
+        angles = geometry.angles
         proj_geom = astra.create_proj_geom('parallel', det_width, det_count,
                                            angles)
 
--- a/odl/tomo/backends/scikit_radon.py
+++ b/odl/tomo/backends/scikit_radon.py
@@ -31,7 +31,7 @@ __all__ = ('scikit_radon_forward', 'scikit_radon_back_projector',
 
 def scikit_theta(geometry):
     """Calculate angles in degrees with ODL scikit conventions."""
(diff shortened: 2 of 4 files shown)


adler-j:
Can you also check that we get the correct result in the 3d case? Perhaps it is wrong as well.

adler-j:
Checklist:

- [x] 2d parallel
- [x] 3d parallel
- [ ] 2d fanflat. ERROR: rotates the wrong direction
- [x] 3d cone

Do you fix this failing case? Perhaps we should add a test case for this?

kohr-h:
The failing case (fan flat) uses ASTRA's vec geometries, so the issue is on their side (inconsistency between 2 and 3 dimensions). I'll make a workaround for it.

adler-j:
I have a feeling a test would be nice here, both to check that our impl's work as expected, and also to give us a warning if astra decides to change their conventions.

adler-j:
This could use getting pushed in, should I fix the last part and add a test?

kohr-h:
Would be nice. I got stuck fixing the fan beam case since I don't exactly know the whereabouts of the conversion functions from our geometry to ASTRA's vec geometry.

adler-j:
Merged from master but for some reason this seems to have fed up the git tree. Any ideas?

adler-j:
Added a test for the angles, seems the fan beam has been fixed at some point since it works correctly on this computer. Merge?

kohr-h:
> Merged from master but for some reason this seems to have fed up the git tree. Any ideas?

Better to rebase. I pushed [a new rebased branch](https://github.com/odlgroup/odl/tree/issue-435__parallel2d_rotation_rebase). As far as I can see it has the same commits as this one, but can you also check @adler-j? If everything is in it, then I'll push it over this one and we can merge.

adler-j:
Looking good for me!

PR #438, opened 2016-06-03
ENH: allow broadcasting in discr element

adler-j:
Convenience fix, result should be obvious from tests. Also had to fix an erroneous test in chambolle pock (that was "saved" before this change).

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -20,7 +20,6 @@
 # Imports for common Python 2/3 codebase
 from __future__ import print_function, division, absolute_import
 from future import standard_library
-from future.utils import raise_from
 standard_library.install_aliases()
 from builtins import super, str
 
@@ -286,12 +285,24 @@ class DiscreteLp(DiscretizedSpace):
         else:
             # Sequence-type input
             arr = np.asarray(inp, dtype=self.dtype, order=self.order)
-            if arr.ndim > 1 and arr.shape != self.shape:
+            input_shape = arr.shape
+
+            # Attempt to handle some cases with exact match
+            if arr.ndim > 1 and self.ndim == 1:
                 arr = np.squeeze(arr)  # Squeeze could solve the problem
                 if arr.shape != self.shape:
                     raise ValueError(
-                        'input shape {} does not match grid shape {}.'
-                        ''.format(arr.shape, self.shape))
+                        'input shape {} does not match grid shape {}'
+                        ''.format(input_shape, self.shape))
+            elif arr.shape != self.shape and arr.ndim == self.ndim:
+                # Try to broadcast the array if possible
+                try:
+                    arr = np.broadcast_to(arr, self.shape)
+                except AttributeError:
+                    # The above requires numpy 1.10, fallback impl else
+                    shape = [m if n == 1 and m != 1 else 1
+                             for n, m in zip(arr.shape, self.shape)]
+                    arr = arr + np.zeros(shape, dtype=arr.dtype)
             arr = arr.ravel(order=self.order)
             return self.element_type(self, self.dspace.element(arr))
 
--- a/odl/space/fspace.py
+++ b/odl/space/fspace.py
@@ -405,9 +405,15 @@ class FunctionSetVector(Operator):
                 out = out.astype(self.out_dtype)
 
             if out_shape != (1,) and out.shape != out_shape:
-                raise ValueError('output shape {} not equal to shape '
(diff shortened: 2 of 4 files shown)


odl/discr/lp_discr.py
```diff
@@ -286,12 +286,24 @@ def element(self, inp=None, **kwargs):
...
                     raise ValueError(
-                        'input shape {} does not match grid shape {}.'
-                        ''.format(arr.shape, self.shape))
+                        'input shape {} does not match grid shape {}'
+                        ''.format(input_shape, self.shape))
+            elif arr.shape != self.shape and arr.ndim == self.ndim:
+                # Try to broadcast the array if possible
+                for i in range(self.ndim):
+                    if arr.shape[i] == 1 and self.shape[i] != 1:
+                        arr = arr.repeat(self.shape[i], axis=i)
```
kohr-h: Did you benchmark this? I think this solution is quite slow compared to broadcasting with factor zero. I suggest that you test this and try it against the alternative of building a zero array for the missing dimensions, i.e.
```python
shape = [m if n == 1 and m != 1 else n
         for n, m in zip(arr.shape, self.shape)]
arr = arr + np.zeros(shape)
```

odl/space/fspace.py
```diff
@@ -405,9 +405,14 @@ def __call__(self, x, out=None, **kwargs):
                 out = out.astype(self.out_dtype)
 
             if out_shape != (1,) and out.shape != out_shape:
-                raise ValueError('output shape {} not equal to shape '
-                                 '{} expected from input.'
-                                 ''.format(out.shape, out_shape))
+                # Try to broadcast the returned element if possible
+                for i in range(len(out_shape)):
+                    if out.shape[i] == 1 and out_shape[i] != 1:
+                        out = out.repeat(out_shape[i], axis=i)
```
kohr-h: same here

kohr-h:
Nice small change, improves usability significantly. Just check which version is faster.

odl/discr/lp_discr.py
kohr-h: CORRECTION:
```python
shape = [m if n == 1 and m != 1 else 1
         for n, m in zip(arr.shape, self.shape)]
arr = arr + np.zeros(shape)
```

adler-j:
Results from performance tests:

code 1
```python
shape = [m if n == 1 and m != 1 else 1
         for n, m in zip(out.shape, out_shape)]
out = out + np.zeros(shape, dtype=out.dtype)
```

code 2
```python
shape = [m if n == 1 and m != 1 else 1
         for n, m in zip(out.shape, out_shape)]
out = np.tile(out, shape)
```

code 3
```python
for i in range(len(out_shape)):
    if out.shape[i] == 1 and out_shape[i] != 1:
        out = out.repeat(out_shape[i], axis=i)
```

code 4
```python
out = np.broadcast_to(out, out_shape)
```


Results with broadcast (1, 1, 500) -> (500, 500, 500)

| Case |  Timing |
|------|---------|
|  1   |  0.084s |
|  2   |  0.404s  |
|  3   |  0.410s  |
|  4   |  0.084s |

Results with broadcast (1, 500, 500) -> (500, 500, 500)

| Case |  Timing |
|------|---------|
|  1   |  0.151s  |
|  2   |  0.396s  |
|  3   |  0.420s  |
|  4    |  0.086s |

The winner is clearly `broadcast_to` followed by add zeros. The issue with `broadcast_to` is that it needs numpy > 1.10, we currently require numpy 1.9. I'll make some kind of if statement to check that and fallback to add zeros.

adler-j:
Updated according to the above, please review again.

odl/discr/lp_discr.py
```diff
@@ -286,12 +285,24 @@ def element(self, inp=None, **kwargs):
...
-                        'input shape {} does not match grid shape {}.'
-                        ''.format(arr.shape, self.shape))
+                        'input shape {} does not match grid shape {}'
+                        ''.format(input_shape, self.shape))
+            elif arr.shape != self.shape and arr.ndim == self.ndim:
+                # Try to broadcast the array if possible
+                try:
+                    arr = np.broadcast_to(arr, self.shape)
+                except AttributeError:
+                    # The above requires numpy 1.9, fallback impl else
```
kohr-h: 1.9 or 1.10?

adler-j: 1.10

kohr-h:
Otherwise good!

PR #439, opened 2016-06-03
API: enforce element(vec) noop

adler-j:
Enforces that the call

    x = x.space.element(x)

does nothing. Implements this in the spaces and adds a test.

### DIFF
--- a/odl/space/ntuples.py
+++ b/odl/space/ntuples.py
@@ -119,12 +119,15 @@ class Ntuples(NtuplesBase):
                 return self.element_type(self, arr)
         else:
             if data_ptr is None:
-                arr = np.array(inp, copy=False, dtype=self.dtype, ndmin=1)
-                if arr.shape != (self.size,):
-                    raise ValueError('expected input shape {}, got {}.'
-                                     ''.format((self.size,), arr.shape))
+                if inp in self:
+                    return inp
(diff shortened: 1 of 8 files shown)


odl/space/ntuples.py
```diff
@@ -119,12 +119,15 @@ def element(self, inp=None, data_ptr=None):
...
         else:
             if data_ptr is None:
-                arr = np.array(inp, copy=False, dtype=self.dtype, ndmin=1)
-                if arr.shape != (self.size,):
-                    raise ValueError('expected input shape {}, got {}.'
-                                     ''.format((self.size,), arr.shape))
+                if inp in self:
+                    return inp
+                else:
+                    arr = np.array(inp, copy=False, dtype=self.dtype, ndmin=1)
```
kohr-h: This does not guarantee a no-op right? It will just try hard not to copy, but will do it if there is no other way.

kohr-h:
Should be alright.

odl/space/ntuples.py
adler-j: I think it actually does, but it is dependent on many other things. I decided to add an if statement above in order to return the input as is instead of returning a view. This should preserve `a is b` type statements and perhaps help in some cases.

PR #441, opened 2016-06-03
Issue 336  move odlpp

adler-j:
* Makes odlpp a plugin
* Changes all tight cuplings to odlpp to be loose cuplings, thus forcing interface changes in many places.
* `FnBase` and `NtuplesBase` type spaces can now be added to ODL by external packages through the `odl.space` hook. This allows users to add new implementations to the odl spaces.
* Changed `Rn` etc syntax to be `rn`, added `impl` parameter. Now any `rn` type space can be created using `odl.rn(3, impl)`
* In line with the above, all numpy spaces are now named `NumpyFn` etc.


adler-j:
Should now be ready for a review. Everything seems to be working as expected.

examples/space/simple_rn.py
```diff
@@ -88,23 +88,23 @@ def asarray(self, *args):
...
 simple_spc = SimpleRn(n)
 
 x, y, z = np.random.rand(n), np.random.rand(n), np.random.rand(n)
 ox, oy, oz = (opt_spc.element(x.copy()), opt_spc.element(y.copy()),
               opt_spc.element(z.copy()))
 sx, sy, sz = (simple_spc.element(x.copy()), simple_spc.element(y.copy()),
               simple_spc.element(z.copy()))
-if odl.CUDA_AVAILABLE:
-    cu_spc = odl.CudaRn(n)
+if 'cuda' in odl.FN_IMPLS:
```
adler-j: This is obviously more general, but perhaps not the prefered method of doing this. I'm not sure how to do it otherwise.

odl/discr/discr_mappings.py
```diff
@@ -242,11 +242,11 @@ def _call(self, func, out=None, **kwargs):
 
         Partition the rectangle by a tensor grid:
 
-        >>> from odl import TensorGrid, Rectangle, RectPartition, Rn
-        >>> rect = Rectangle([1, 3], [2, 5])
-        >>> grid = TensorGrid([1, 2], [3, 4, 5])
-        >>> partition = RectPartition(rect, grid)
-        >>> rn = Rn(grid.size)
+        >>> import odl
```
adler-j: This should be applied across the board.

examples/space/simple_rn.py
kohr-h: Well, this is an example, so a bit of hard-coding may not be that troublesome.

odl/space/numpy_ntuples.py
```diff
@@ -145,27 +145,38 @@ def one(self):
...
         >>> x = c3.one()
         >>> x
-        Cn(3).element([(1+0j), (1+0j), (1+0j)])
+        ntuples(3, 'complex').element([(1+0j), (1+0j), (1+0j)])
         """
         return self.element(np.ones(self.size, dtype=self.dtype))
 
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        class_name = 'ntuples'
```
kohr-h: That name is a bit misleading (always was). `constructor_name` would be more precise.

odl/space/numpy_ntuples.py
```diff
@@ -145,27 +145,38 @@ def one(self):
...
+        inner_str = '{}'.format(self.size)
+        inner_str += ', {}'.format(dtype_repr(self.dtype))
+
+        return '{}({})'.format(class_name, inner_str)
+
     @property
     def element_type(self):
-        """ `NtuplesVector` """
-        return NtuplesVector
+        """ `NumpyNtuplesVector` """
```
kohr-h: Fix spacing

odl/space/numpy_ntuples.py
```diff
@@ -403,37 +414,50 @@ def __setitem__(self, indices, values):
...
+        if isinstance(values, NumpyNtuplesVector):
             self.data[indices] = values.data
         else:
             self.data[indices] = values
 
+    @staticmethod
+    def available_dtypes():
+        """Return the available `Fn` data types.
+
+        This is all dtypes available to numpy.
```
kohr-h: Reformulate

odl/space/numpy_ntuples.py
```diff
@@ -403,37 +414,50 @@ def __setitem__(self, indices, values):
...
+
+
+        """
+        all_types = []
+        for types in np.sctypes:
+            all_types += types
```
kohr-h: This produces the following list for `all_types`:
```python
['u', 'i', 'n', 't', 'c', 'o', 'm', 'p', 'l', 'e', 'x', 'f', 'l', 'o', 'a', 't', 'i', 'n', 't', 'o', 't', 'h', 'e', 'r', 's']
```
Did you mean this:
```python
types = []
for val in np.sctypes.values():
    types.extend(val)
```
?

odl/space/numpy_ntuples.py
```diff
@@ -563,7 +587,7 @@ class Fn(FnBase, Ntuples):
     Its elements are represented as instances of the `FnVector` class.
     """
 
-    def __init__(self, size, dtype, **kwargs):
+    def __init__(self, size, dtype='float64', **kwargs):
```
kohr-h: Why this change? `Fn` is generic and to me there is no canonical choice of default `dtype`.

odl/space/numpy_ntuples.py
```diff
@@ -840,16 +837,23 @@ def _dist(self, x1, x2):
         Examples
         --------
         >>> from numpy.linalg import norm
-        >>> c2_2 = Cn(2, dist=lambda x, y: norm(x - y, ord=2))
+
+        Euclidean distance
```
kohr-h: Could be a bit more verbose, complete sentences. This minimum text reads a bit ugly in the formatted doc.

odl/space/numpy_ntuples.py
```diff
@@ -992,50 +1001,50 @@ def __eq__(self, other):
...
-        return (Ntuples.__eq__(self, other) and
+        return (NumpyNtuples.__eq__(self, other) and
                 self.weighting == other.weighting)
 
     def __repr__(self):
-        """s.__repr__() <==> repr(s)."""
+        """Return ``repr(self)``."""
         if self.is_rn:
-            class_name = 'Rn'
+            class_name = 'rn'
```
kohr-h: Same as above

odl/space/numpy_ntuples.py
```diff
@@ -1047,53 +1056,89 @@ def __repr__(self):
         return '{}({})'.format(class_name, inner_str)
 
     # Copy these to handle bug in ABCmeta
```
kohr-h: How about we scrap ABCMeta for spaces? I don't think many people will ever implement a space, and those who do can probably deal with the "soft" abstract method implementation via `NotImplementedError`.

odl/space/numpy_ntuples.py
```diff
@@ -1047,53 +1056,89 @@ def __repr__(self):
...
+        """
+        return np.sctypes['int'] + np.sctypes['float'] + np.sctypes['complex']
+
+    @staticmethod
+    def default_dtype(field=None):
+        """Return the default of `Fn` data type for a given field.
+
+        Parameters
+        ----------
+        field : `Field`
```
kohr-h: optional

odl/space/numpy_ntuples.py
```diff
@@ -1047,53 +1056,89 @@ def __repr__(self):
...
 
     @property
     def real(self):
         """The real part of this vector.
 
         Returns
         -------
-        real : `FnVector` view with dtype
-            The real part this vector as a vector in `Rn`
+        real : `NumpyFnVector` view with dtype
```
kohr-h: with real dtype?

odl/space/numpy_ntuples.py
adler-j: Yes!

odl/space/space_utils.py
```diff
@@ -22,14 +22,27 @@
...
+
+NTUPLES_IMPLS = {'numpy': NumpyNtuples}
+FN_IMPLS = {'numpy': NumpyFn}
+for entry_point in iter_entry_points(group='odl.space', name=None):
+    try:
+        module = entry_point.load()
+        NTUPLES_IMPLS.update(module.ntuples_impls())
+        FN_IMPLS.update(module.fn_impls())
+    except ImportError:
+        pass
```
kohr-h: Not sure `space_utils` is the right place for this. It's an internal thing. I'd put it into something like an `_entry_points.py` file under `odl/space`.

odl/space/numpy_ntuples.py
adler-j: Well yes, but this is the same as the value of `default_type()`, which seemed reasonable?

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+        space_type = fn
     else:
-        raise ValueError("implementation '{}' not understood.".format(impl_in))
+        space_type = ntuples
+
+    return space_type(len(arr), dtype=space_dtype, impl=impl).element(arr)
+
+
+def ntuples(size, dtype, impl='numpy', **kwargs):
+
```
kohr-h: no blank

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+    See also
+    --------
+    Fn : n-tuples over a field :math:`\mathbb{F}` with arbitrary scalar
+        data type
+    """
+
+    ntuple = NTUPLES_IMPLS[impl](size, dtype, **kwargs)
+
+    return ntuple
```
kohr-h: `return NTUPLES_IMPLS[impl](size, dtype, **kwargs)`?

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+    dtype : `object`
+        The data type of the storage array. Can be provided in any
+        way the `numpy.dtype` function understands, most notably
+        as built-in type, as one of NumPy's internal datatype
+        objects or as string.
+
+        Only complex floating-point data types are allowed.
+    impl : `str`
+        The backend to use.
+        TODO: words
```
kohr-h: do

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+        way the `numpy.dtype` function understands, most notably
+        as built-in type, as one of NumPy's internal datatype
+        objects or as string.
+
+        Only complex floating-point data types are allowed.
+        Default: default of the implementation given by
+        ``FnImpl.default_dtype()``.
+    impl : `str`
+        The backend to use.
+        TODO: words
```
kohr-h: do

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+        The data type of the storage array. Can be provided in any
+        way the `numpy.dtype` function understands, most notably
+        as built-in type, as one of NumPy's internal datatype
+        objects or as string.
+
+        Only complex floating-point data types are allowed.
+        ``FnImpl.default_dtype(ComplexNumbers())``.
+    impl : `str`
+        The backend to use.
+        TODO: words
```
kohr-h: do it!

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+    kwargs : {'weight', 'dist', 'norm', 'inner', 'dist_using_inner'}
+        See `Fn`
```
kohr-h: fix link

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+        The number of dimensions of the space
+    dtype : `object`
+        The data type of the storage array. Can be provided in any
+        way the `numpy.dtype` function understands, most notably
+        as built-in type, as one of NumPy's internal datatype
+        objects or as string.
+
+        Only complex floating-point data types are allowed.
+        Default: default of the implementation given by
+        ``FnImpl.default_dtype()``.
```
kohr-h: What is `FnImpl`?

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+    impl : `str`
+        The backend to use.
+        TODO: words
+    kwargs : {'weight', 'dist', 'norm', 'inner', 'dist_using_inner'}
+        See `Fn`
```
kohr-h: link

odl/space/space_utils.py
```diff
@@ -96,33 +108,165 @@ def vector(array, dtype=None, impl='numpy'):
...
+
+    See also
+    --------
+    Fn : n-tuples over a field :math:`\mathbb{F}` with arbitrary scalar
+        data type
+    """
+
```
kohr-h: remove blank line

odl/tomo/backends/astra_cpu.py
```diff
@@ -64,18 +64,19 @@ def astra_cpu_forward_projector(vol_data, geometry, proj_space, out=None):
     if not isinstance(vol_data, DiscreteLpVector):
         raise TypeError('volume data {!r} is not a DiscreteLpVector '
                         'instance.'.format(vol_data))
-    if not isinstance(vol_data.space.dspace, Ntuples):
+    if not isinstance(vol_data.space.dspace, NumpyNtuples):
         raise TypeError('data type {!r} of the volume space is not an '
```
kohr-h: This error message makes no sense.

odl/trafos/fourier.py
```diff
@@ -1569,7 +1565,7 @@ def __init__(self, domain, range=None, impl='numpy', **kwargs):
         if not isinstance(domain, DiscreteLp):
             raise TypeError('domain {!r} is not a `DiscreteLp` instance.'
                             ''.format(domain))
-        if not isinstance(domain.dspace, Ntuples):
+        if not isinstance(domain.dspace, NumpyNtuples):
```
kohr-h: `if not domain.dspace.impl == 'numpy'`

test/discr/lp_discr_test.py
```diff
@@ -83,104 +75,66 @@ def test_init(exponent):
...
-
-    for dtype in complex_float_dtypes:
-        discr = odl.uniform_discr(0, 1, 10, impl='numpy', dtype=dtype)
-        assert isinstance(discr.dspace, odl.Fn)
+        assert isinstance(discr.dspace, FnBase)
+        assert discr.dspace.impl == fn_impl
         assert discr.is_cn
-        assert discr.dspace.element().space.dtype == dtype
+        assert discr.dspace.exponent == exponent
+    except TypeError:
```
kohr-h: I don't think a wrong data type argument should raise a `TypeError`. It's technically wrong since the provided argument is not of the wrong type, but rather has the wrong value. If `ValueError` is too broad for this, maybe a custom error class would be preferable.

odl/space/space_utils.py
```diff
@@ -22,14 +22,27 @@
...
+
+NTUPLES_IMPLS = {'numpy': NumpyNtuples}
+FN_IMPLS = {'numpy': NumpyFn}
+for entry_point in iter_entry_points(group='odl.space', name=None):
+    try:
+        module = entry_point.load()
+        NTUPLES_IMPLS.update(module.ntuples_impls())
+        FN_IMPLS.update(module.fn_impls())
+    except ImportError:
```
kohr-h: One question: does this try to import the "plugin module" that shall dock at the entry point (from the exception that's what I guess)? If yes, will it fail if the module is misconfigured, e.g. lacks a library? 

Another point that came to my mind is that sometimes you successfully import the CUDA stuff, but at runtime it crashes, e.g. due to a `MemoryError` or simply because no GPU can be found. I think it would be good if the module would be excluded in that case, since many tests will fail in that scenario, and it happens basically every time you suspend & resume on Linux (some Nvidia driver or Linux kernel bug). Technically one could implement a very simple import hook which does a minimal test and bails out on failure.

adler-j: Yes the library is imported somehow inside the "load" call. And yes, it will fail if the submodule lacks a library or similarily (basically triggering an importerror). I added this to make sure that a faulty pluggin does not render a odl installation unusable.

Regarding doing a small test Ill have to disagree. We used to do this a long time ago, but it came with a notable performance penalty (import odl became about 3 times slower, and were already slower than numpy) and honestly only helped hiding bugs.

If a plugin is installed, we should i.m.o run the tests for it. That they fail is in this case correct and indicates an actual bug.

What we could do is add some exclude flag: "--exclude-plugins=cuda" to the pytest config. Then users can opt to skip them, if the user wants to.

odl/space/numpy_ntuples.py
kohr-h: Not exactly. The `default_dtype()` knew about the field (before you made it optional :-)) so it knew if a real or complex space was desired, and there were the canonical defaults. Before that, there is none IMO.

odl/space/numpy_ntuples.py
```diff
@@ -40,13 +40,13 @@
     WeightingBase, MatrixWeightingBase, VectorWeightingBase,
```
kohr-h: For the file name, we use in the CUDA case `cu_ntuples.py`. I suggest `npy_ntuples` here as abbreviation, just for consistency. The rename itself is good.

odl/space/space_utils.py
kohr-h: > If a plugin is installed, we should i.m.o run the tests for it. That they fail is in this case correct and indicates an actual bug.

Not necessarily. In my case it made the tests print 500 `MemoryError` test failures, which hid the actual failing tests.



> What we could do is add some exclude flag: "--exclude-plugins=cuda" to the pytest config. Then users can opt to skip them, if the user wants to.

That would be a viable solution.

kohr-h:
We should also adapt the names of the `weighted_blabla` factories to match the abbreviation used in the module names. We have `cu_weighed_...` for CUDA, then here it would be `npy_weighted_...`.

odl/space/numpy_ntuples.py
adler-j: We should make a new issue in that case, I tend to agree here (could also remove for geometries?)

odl/space/space_utils.py
adler-j: Wouldnt make it private, we need to expose `FN_IMPLS` outside of ODL.

odl/space/space_utils.py
adler-j: > Not necessarily. In my case it made the tests print 500 MemoryError test failures, which hid the actual failing tests.

Well, `MemoryError` is also an error that needs to be fixed.

> That would be a viable solution.

I'll do that.

test/discr/lp_discr_test.py
kohr-h: Yes I know. It's also mentioned in #446 so the change goes into a branch for that issue.

adler-j:
Updates according to comments

adler-j:
Everything should be fixed up now. Ready for final review, this should be merged asap to get it out of the way since it touches everything.

odl/discr/discretization.py
```diff
@@ -214,6 +212,11 @@ def __eq__(self, other):
                 other.interpolation == self.interpolation)
 
     @property
+    def impl(self):
+        """The underlying implmentation type for the dspace."""
```
kohr-h: Drop "The", and replace "dspace" by something more telling, e.g. "data storage"?

odl/discr/lp_discr.py
```diff
@@ -34,9 +34,7 @@
 from odl.discr.partition import RectPartition, uniform_partition_fromintv
 from odl.set.sets import RealNumbers, ComplexNumbers
 from odl.set.domain import IntervalProd
-from odl.space.cu_ntuples import CUDA_AVAILABLE, CudaFn
-from odl.space.fspace import FunctionSpace
-from odl.space.ntuples import Fn
+from odl.space import FunctionSpace, FN_IMPLS
```
kohr-h: We should have some sort of naming convention for such global sequences, I'd be good with `<XYZ>_IMPLS`. Should be applied across the board then.

odl/discr/lp_discr.py
```diff
@@ -387,14 +385,9 @@ def __repr__(self):
...
-            if isinstance(self.dspace, Fn):
-                impl = 'numpy'
-                default_dtype = np.float64
-            elif isinstance(self.dspace, CudaFn):
-                impl = 'cuda'
-                default_dtype = np.float32
-            else:  # This should never happen
-                raise RuntimeError('unable to determine data space impl')
+            impl = self.dspace.impl
+            default_dtype = self.dspace.default_dtype(RealNumbers())
```
kohr-h: How about default ordering? You wrote earlier that `'F'` should be default for CUDA. Maybe that information should be brought in by the plug-in?

odl/discr/lp_discr.py
```diff
@@ -1022,10 +1007,8 @@ def uniform_discr_fromintv(interval, nsamples, exponent=2.0, interp='nearest',
 
     Parameters
     ----------
-    domain : `IntervalProd`
+    interval : `IntervalProd`
```
kohr-h: Gives 1d impression. Name should be dimension-agnostic.

odl/space/entry_points.py
```diff
@@ -0,0 +1,55 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Entry points for adding more spaces to ODL using external packages.
+
+External packages can add implementations of `NtuplesBase` and `FnBase` by
+hooking into the setuptools entry point 'odl.space' and exposing the methods
+``ntuples_impls`` and ``fn_impls``.
```
kohr-h: Should be capitalized?

odl/space/entry_points.py
```diff
@@ -0,0 +1,55 @@
...
+
+Attributes
+----------
+NTUPLES_IMPLS : `dict`
+    A dictionary that maps a string to a `NtuplesBase` implementation.
+FN_IMPLS : `dict`
+    A dictionary that maps a string to a `FnBase` implementation.
```
kohr-h: to an

odl/space/entry_points.py
```diff
@@ -0,0 +1,55 @@
...
+"""Entry points for adding more spaces to ODL using external packages.
+
+External packages can add implementations of `NtuplesBase` and `FnBase` by
+hooking into the setuptools entry point 'odl.space' and exposing the methods
+``ntuples_impls`` and ``fn_impls``.
+
+Attributes
+----------
+NTUPLES_IMPLS : `dict`
+    A dictionary that maps a string to a `NtuplesBase` implementation.
```
kohr-h: to an

odl/space/entry_points.py
```diff
@@ -0,0 +1,55 @@
...
+
+__all__ = ('NTUPLES_IMPLS', 'FN_IMPLS')
+
+NTUPLES_IMPLS = {'numpy': NumpyNtuples}
+FN_IMPLS = {'numpy': NumpyFn}
+for entry_point in iter_entry_points(group='odl.space', name=None):
+    try:
+        module = entry_point.load()
+        NTUPLES_IMPLS.update(module.ntuples_impls())
+        FN_IMPLS.update(module.fn_impls())
```
kohr-h: Should we require capitalization here, for consistency with the internal handling?

odl/discr/discretization.py
adler-j: Id probably simply add ticks to `dspace`, that would create a link to the `dspace` property which explains further?

odl/space/npy_ntuples.py
```diff
@@ -406,37 +418,53 @@ def __setitem__(self, indices, values):
...
+        ntuples(2, 'int8').element([0, 0])
         """
-        if isinstance(values, NtuplesVector):
+        if isinstance(values, NumpyNtuplesVector):
             self.data[indices] = values.data
         else:
             self.data[indices] = values
 
+    @staticmethod
+    def available_dtypes():
```
kohr-h: Shouldn't this be a space method?

odl/discr/lp_discr.py
adler-j: If we at some point add that then sure, for now we dont have that problem. The more I think about that one the more sceptical I am as-well. It makes the code more indirect and harder to follow.

odl/discr/lp_discr.py
adler-j: Possible, but I wont touch that in this branch. This was simply harmonizing with the input parameter.

odl/space/entry_points.py
adler-j: nope, they are methods so they should not.

odl/space/entry_points.py
adler-j: Which part should be capitalized?

odl/space/npy_ntuples.py
```diff
@@ -842,17 +843,32 @@ def _dist(self, x1, x2):
...
+
+        >>> c2_2 = NumpyFn(2, dtype=complex)
         >>> x = c2_2.element([3+1j, 4])
         >>> y = c2_2.element([1j, 4-4j])
         >>> c2_2.dist(x, y)
         5.0
 
-        >>> c2_2 = Cn(2, dist=lambda x, y: norm(x - y, ord=1))
+        If the user has given another dist function, that one is used instead.
+        For example, the 2-norm can be given explicitly
```
kohr-h: Dot or colon at the end

odl/discr/lp_discr.py
adler-j: I also feel this convention is OK. In what other places would you want it? `FFT_IMPLS`?

odl/space/npy_ntuples.py
```diff
@@ -995,50 +1016,50 @@ def __eq__(self, other):
...
-        >>> c3_lambda1 = Cn(3, dist=lambda x, y: norm(x-y, ord=1))
-        >>> c3_lambda2 = Cn(3, dist=lambda x, y: norm(x-y, ord=1))
-        >>> c3_lambda1 == c3_lambda2
+        >>> r3_lambda1 = NumpyFn(3, dist=lambda x, y: norm(x-y, ord=1))
+        >>> r3_lambda2 = NumpyFn(3, dist=lambda x, y: norm(x-y, ord=1))
+        >>> r3_lambda1 == r3_lambda2
         False
 
-        An `Fn` space with the same data type is considered
+        An `NumpyFn` space with the same data type is considered
```
kohr-h: . or :

odl/space/npy_ntuples.py
adler-j: You are 100% correct.

odl/space/space_utils.py
```diff
@@ -89,40 +89,176 @@ def vector(array, dtype=None, impl='numpy'):
...
+    if dtype is None:
+        dtype = fn_impl.default_dtype()
+
+    fn = fn_impl(size, dtype, **kwargs)
+
+    return fn
+
+
+def cn(size, dtype=None, impl='numpy', **kwargs):
+
```
kohr-h: Remove blank line

odl/space/space_utils.py
```diff
@@ -89,40 +89,176 @@ def vector(array, dtype=None, impl='numpy'):
...
+    See also
+    --------
+    fn : n-tuples over a field :math:`\mathbb{F}` with arbitrary scalar data
+         type.
+    """
+    return NTUPLES_IMPLS[impl](size, dtype, **kwargs)
+
+
+def fn(size, dtype=None, impl='numpy', **kwargs):
+
```
kohr-h: Remove blank

odl/space/space_utils.py
```diff
@@ -89,40 +89,176 @@ def vector(array, dtype=None, impl='numpy'):
...
+    """The vector space :math:`F^n` with vector multiplication.
```
kohr-h: Could be more informative, perhaps
```
Return the space `F^n` for arbitrary field `F`.
```
or "for real or complex numbers `F`"

odl/space/space_utils.py
```diff
@@ -89,40 +89,176 @@ def vector(array, dtype=None, impl='numpy'):
...
+        as built-in type, as one of NumPy's internal datatype
+        objects or as string.
+
+        Only complex floating-point data types are allowed.
+
+        Default: default of the implementation given by calling
+        ``default_dtype(ComplexNumbers())`` on the `FnBase` implementation.
+    impl : `str`
+        The backend to use. See `FN_IMPLS` for available options.
+    **kwargs :
```
kohr-h: no stars

odl/space/space_utils.py
```diff
@@ -89,40 +89,176 @@ def vector(array, dtype=None, impl='numpy'):
...
+        The data type of the storage array. Can be provided in any
+        way the `numpy.dtype` function understands, most notably
+        as built-in type, as one of NumPy's internal datatype
+        objects or as string.
+
+        Default: default of the implementation given by calling
+        ``default_dtype()`` on the `FnBase` implementation.
+    impl : `str`
+        The backend to use. See `FN_IMPLS` for available options.
+    **kwargs :
```
kohr-h: no stars

odl/space/space_utils.py
```diff
@@ -89,40 +89,176 @@ def vector(array, dtype=None, impl='numpy'):
...
+    cn = cn_impl(size, dtype, **kwargs)
+
+    if not cn.is_cn:
+        raise TypeError('data type {!r} not a complex floating-point type.'
+                        ''.format(dtype))
+    return cn
+
+
+def rn(size, dtype=None, impl='numpy', **kwargs):
+
```
kohr-h: See above

odl/space/entry_points.py
```diff
@@ -0,0 +1,55 @@
...
+----------
+NTUPLES_IMPLS : `dict`
+    A dictionary that maps a string to a `NtuplesBase` implementation.
+FN_IMPLS : `dict`
+    A dictionary that maps a string to a `FnBase` implementation.
+
+Notes
+-----
+This is used with functions such as `rn`, `fn` and `uniform_discr` in order
+to allow arbitrary implementations.
```
kohr-h: Is there any place to which we can link for guidance regarding the attributes that are expected? Cheapest would be to have a `See also` linking to `NumpyNtuples` and `NumpyFn` (if that works properly for a module docstring), but that could be overwhelming for a "power user" who wants to make a new space. Maybe a link to `LinearSpace` and a list of some additional "desirable" attributes would be a good addition.

odl/discr/lp_discr.py
kohr-h: Probably if you search for "SUPPORTED" you'll get some hits. We also have ASTRA, STIR, ...

odl/space/entry_points.py
adler-j: I honestly dont know myself what is required, but perhaps that would be a good exercise to do :)

odl/discr/lp_discr.py
kohr-h: We can add `default_order` to an `NtuplesBase` type space if we want, so that info would come in "automatically" by the plug-in. Seems a bit meaningless for one-dimensional storage, though, it becomes meaningful only in combination with `DiscretizedSet`. One we have multi-dimensional storage (*knock knock) this one becomes more interesting.

odl/discr/lp_discr.py
kohr-h: ok

odl/space/entry_points.py
kohr-h: `module.FN_IMPLS`

If those guys are module-level constants they should be capitalized. If they are something else, e.g. module-level functions, then it's okay.
The fact that I don't know from the documentation what `ntuples_impls` and `fn_impls` are indicates that the documentation should be more precise about the expected interface.

kohr-h:
Looking good, just some minor things regarding documentation, especially regarding the (admittedly quite advanced) entry points interface.
I'm going to run some builds now.

adler-j:
Ready for merge?

kohr-h:
I haven't built it yet. I need to do something else right now and my time is really limited today and tomorrow. I'll try to find some time in between.

adler-j:
I just built it here at elekta on Ubuntu and it worked perfectly. Are we ok with merge then?

kohr-h:
Well then go for it. The rest we can fix on the conda build branch.

PR #444, opened 2016-06-09
Issue 443  uniform partition

### LINKED ISSUE #443: Make uniform_partition more flexible
We have 4 possible parameters in the creation of a uniform partition:

- `begin`
- `end`
- `num_nodes`
- `cell_sides`

They are not independent, of course -- given 3, one can compute the missing one. But currently, we only support creation of partitions with the first three. It would be much more convenient and flexible to support any combination (there are 4) of 3 parameters and calculate the fourth.

If this clutters the interface of `uniform_partition` too much, we can have it as a new function.

kohr-h:
Implements `uniform_partition_options` as mentioned in #443. The name is sub-optimal but quite clearly, the function should be separate from `uniform_partition`.

Closes #443

### DIFF
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -591,71 +591,6 @@ def uniform_partition_fromintv(intv_prod, num_nodes, nodes_on_bdry=False):
     return RectPartition(intv_prod, grid)
 
 
-def uniform_partition(begin, end, num_nodes, nodes_on_bdry=False):
-    """Return a partition of [begin, end] with equally sized cells.
-
-    Parameters
-    ----------
-    begin, end : `array-like`
-        Vectors defining the begin end end points of an `IntervalProd`
-        (a rectangular box)
-    num_nodes : `int` or `sequence` of `int`
-        Number of nodes per axis. For 1d intervals, a single integer
-        can be specified.
-    nodes_on_bdry : `bool` or `sequence`, optional
-        If a sequence is provided, it determines per axis whether to
-        place the last grid point on the boundary (True) or shift it
-        by half a cell size into the interior (False). In each axis,
-        an entry may consist in a single `bool` or a 2-tuple of
-        `bool`. In the latter case, the first tuple entry decides for
-        the left, the second for the right boundary. The length of the
-        sequence must be ``array.ndim``.
-
-        A single boolean is interpreted as a global choice for all
-        boundaries.
-
-    See also
-    --------
-    uniform_partition_fromintv : partition an existing set
-    uniform_partition_fromgrid : use an existing grid as basis
-
-    Examples
-    --------
-    By default, no grid points are placed on the boundary:
-
-    >>> part = uniform_partition(0, 1, 4)
-    >>> part.cell_boundary_vecs
-    (array([ 0.  ,  0.25,  0.5 ,  0.75,  1.  ]),)
-    >>> part.grid.coord_vectors
-    (array([ 0.125,  0.375,  0.625,  0.875]),)
-
-    This can be changed with the nodes_on_bdry parameter:
-
-    >>> part = uniform_partition(0, 1, 3, nodes_on_bdry=True)
-    >>> part.cell_boundary_vecs
-    (array([ 0.  ,  0.25,  0.75,  1.  ]),)
-    >>> part.grid.coord_vectors
-    (array([ 0. ,  0.5,  1. ]),)
-
-    We can specify this per axis, too. In this case we choose both
-    in the first axis and only the rightmost in the second:
-
-    >>> part = uniform_partition([0, 0], [1, 1], (3, 3),
-    ...                          nodes_on_bdry=(True, (False, True)))
-    ...
-    >>> part.cell_boundary_vecs[0]  # first axis, as above
-    array([ 0.  ,  0.25,  0.75,  1.  ])
-    >>> part.grid.coord_vectors[0]
-    array([ 0. ,  0.5,  1. ])
-    >>> part.cell_boundary_vecs[1]  # second, asymmetric axis
-    array([ 0. ,  0.4,  0.8,  1. ])
-    >>> part.grid.coord_vectors[1]
-    array([ 0.2,  0.6,  1. ])
-    """
-    return uniform_partition_fromintv(
-        IntervalProd(begin, end), num_nodes, nodes_on_bdry)
-
-
 def uniform_partition_fromgrid(grid, begin=None, end=None):
     """Return a partition of an interval product based on a given grid.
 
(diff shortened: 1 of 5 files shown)


odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+    ...                                  cell_sides=[0.5, None])
+    >>> part.cell_boundary_vecs
+    (array([ 0. ,  0.5,  1. ]), array([ 0.,  1.,  2.]))
+    >>> part = uniform_partition_options(begin=[0, None], end=[None, 2],
+    ...                                  num_nodes=[2, 2],
+    ...                                  cell_sides=[0.5, 0.25])
+    >>> part.cell_boundary_vecs
+    (array([ 0. ,  0.5,  1. ]), array([ 1.5 ,  1.75,  2.  ]))
+    """
+    # np.size(None) = 1
```
kohr-h: nitpicking: `# np.size(None) == 1`

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
     return RectPartition(IntervalProd(begin_vec, end_vec), grid)
 
 
+def uniform_partition_options(begin=None, end=None, num_nodes=None,
+                              cell_sides=None):
+    """Return a partition with equally sized cells.
+
+    This function offers more flexibility over `uniform_discr` in that
```
adler-j: `uniform_partition`

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+    ...                                  num_nodes=[None, 2],
+    ...                                  cell_sides=[0.5, None])
+    >>> part.cell_boundary_vecs
+    (array([ 0. ,  0.5,  1. ]), array([ 0.,  1.,  2.]))
+    >>> part = uniform_partition_options(begin=[0, None], end=[None, 2],
+    ...                                  num_nodes=[2, 2],
+    ...                                  cell_sides=[0.5, 0.25])
+    >>> part.cell_boundary_vecs
+    (array([ 0. ,  0.5,  1. ]), array([ 1.5 ,  1.75,  2.  ]))
+    """
```
adler-j: See Also

...

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+    # np.size(None) = 1
+    sizes = [np.size(p) for p in (begin, end, num_nodes, cell_sides)]
+    max_size = int(np.max(sizes))
```
adler-j: what if the sizes are different?

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+        (a rectangular box). For one-dimensional partitions, single
+        floats can be provided. ``None`` entries mean "compute the value".
+    num_nodes : int or sequence of int, optional
+        Number of nodes per axis. For 1d intervals, a single integer
+        can be specified. ``None`` entries mean "compute the value".
+    cell_sides : float or array-like, optional
+        Side length of the partition cells per axis. For 1d intervals,
+        a single integer can be specified. ``None`` entries mean
+        "compute the value".
+
```
adler-j: Why no `nodes_on_bdry`?

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+                             'parameters `begin`, `end`, `num_nodes`, '
+                             '`cell_sides`, got {}'
+                             ''.format(i, num_params))
+        if b is None:
+            begin[i] = e - n * s
+        elif e is None:
+            end[i] = b + n * s
+        elif n is None:
+            n_calc = (e - b) / s
+            n_round = int(round(n_calc))
```
adler-j: Are we sure this rounding goes the way we want it?

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+            if not np.isclose(n_calc, n_round, atol=1e-5):
```
adler-j: Isnt this a very complicated way of writing

```python
if abs(n_calc - n_round) > 1e-5:
    ...
```

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+            cell_sides = [None] * max_size
+
+        begin = list(begin)
+        end = list(end)
+        num_nodes = list(num_nodes)
+        cell_sides = list(cell_sides)
+
+    for i, (b, e, n, s) in enumerate(zip(begin, end, num_nodes, cell_sides)):
+        num_params = sum(p is not None for p in (b, e, n, s))
+        if num_params < 3:
```
adler-j: Do we really want to accept anything else than 3? 4 parameters will only lead to confusion, no?

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+
+    Examples
+    --------
+    Any combination of three of the four parameters can be used for
+    creation of a partition:
+
+    >>> part = uniform_partition_options(begin=0, end=1, num_nodes=4)
+    >>> part.cell_boundary_vecs
+    (array([ 0.  ,  0.25,  0.5 ,  0.75,  1.  ]),)
+    >>> part = uniform_partition_options(begin=0, num_nodes=4, cell_sides=0.5)
```
adler-j: I would much prefer if you used the same example in all of these, that way users see how to make the same thing.

A fix would be to change `cell_sides` to `0.25`.

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+    In higher dimensions, the parameters can be given differently in
+    each axis. Where None is given, the value will be computed:
+
+    >>> part = uniform_partition_options(begin=[0, 0], end=[1, 2],
+    ...                                  num_nodes=[4, 2])
+    >>> part.cell_boundary_vecs
+    (array([ 0.  ,  0.25,  0.5 ,  0.75,  1.  ]), array([ 0.,  1.,  2.]))
+    >>> part = uniform_partition_options(begin=[0, 0], end=[1, 2],
+    ...                                  num_nodes=[None, 2],
+    ...                                  cell_sides=[0.5, None])
```
adler-j: See above comment on creating the same partition in all cases. Here, `cell_sides=[0.25, None]`

odl/discr/partition.py
kohr-h: Hm, I guess I could check after turning `None` into lists.

odl/discr/partition.py
```diff
@@ -766,6 +767,132 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+    ...                                  cell_sides=[0.5, 0.25])
+    >>> part.cell_boundary_vecs
+    (array([ 0. ,  0.5,  1. ]), array([ 1.5 ,  1.75,  2.  ]))
+    """
+    # np.size(None) = 1
+    sizes = [np.size(p) for p in (begin, end, num_nodes, cell_sides)]
+    max_size = int(np.max(sizes))
+
+    if max_size == 1:
+        begin = [begin]
```
adler-j: This could perhaps be done on a by parameter basis:

`uniform_partition_options(begin=0, end=1, num_nodes=[1, 2, 3, 4])

adler-j:
Functionality i great, but I'm not sure that this should be separate from `uniform_partition`, there is a degree of clutter to this.

Regarding `tol` there is a leftover in `contains_set`, and `contains_all` seems to be lacking the parameter fully. Might as well fix while we're at it.

odl/discr/partition.py
kohr-h: I thought it would become too complicated. But I'll go through the different situations and check what we got. We have the following 4 cases for provided parameters:

- `begin`, `end`, `num_nodes`: equivalent to `uniform_sampling`, easy to introduce `nodes_on_bdry`
- `begin`, `end`, `cell_sides`: Here, `nodes_on_bdry` could be used to determine if we have half cells only at the boundaries. That would mean to calculate `num_nodes` differently.
- `begin`, `num_nodes`, `cell_sides`: Same as the second case, but here the computation of `end` has to be adapted.
- `end`, `num_nodes`, `cell_sides`: Backwards version of the third case

Conclusion: ok, why not?

odl/discr/partition.py
kohr-h: It goes towards the closest integer. Corner cases in the middle raise, so no need to think too much about it.

odl/discr/partition.py
kohr-h: Yeah, this version is "historically grown". First I had your version, then I thought "Ah, I'll just use `np.isclose` and whatever it defaults to", then I thought the default is too strict and changed `atol`, an voilà, here's the bloated version of the above code.

odl/discr/partition.py
kohr-h: Valid alternative to require exactly 3 arguments. Checks are quite sensitive to numerical error anyway, so perhaps it's better not to allow 4 parameters.

odl/discr/partition.py
kohr-h: Sure

odl/discr/partition.py
kohr-h: Addition: With `nodes_on_bdry` added, the argument for keeping this function separate from `uniform_partition` us much weaker. I'll reconsider.

odl/discr/partition.py
kohr-h: > This could perhaps be done on a by parameter basis:
> 
> `uniform_partition_options(begin=0, end=1, num_nodes=[1, 2, 3, 4])

What exactly do you mean by this?

adler-j: That the current implementation requires all input parameters to be either "scalars" or arrays. I propose that we add "broadcasting" so users can mix these.

kohr-h:
Done with all the comments and rebased. Waiting for Travis to be happy.

Now the functionality is merged into `uniform_partition`.

odl/discr/partition.py
```diff
@@ -766,6 +701,202 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
     return RectPartition(IntervalProd(begin_vec, end_vec), grid)
 
 
+def uniform_partition(begin=None, end=None, num_nodes=None,
+                      cell_sides=None, nodes_on_bdry=False):
+    """Return a partition with equally sized cells.
+
+    This function offers more flexibility over `uniform_partition` in
```
adler-j: Old comment

odl/discr/partition.py
```diff
@@ -766,6 +701,202 @@ def uniform_partition_fromgrid(grid, begin=None, end=None):
...
+        `bool`. In the latter case, the first tuple entry decides for
+        the left, the second for the right boundary. The length of the
+        sequence must be ``array.ndim``.
+
+        A single boolean is interpreted as a global choice for all
+        boundaries.
+
+    Notes
+    -----
+    In each axis, 3 of the 4 possible parameters must be given. If all
```
adler-j: Somewhat redundant given the introduction. Also, this now takes 5 parameters so we need to specify which 4.

test/set/domain_test.py
```diff
@@ -409,7 +409,7 @@ def test_div():
     quotient = IntervalProd([1 / 2., 2 / 3., 3 / 4., 4 / 5., 5 / 6.],
                             [2 / 1., 3 / 2., 4 / 3., 5 / 4., 6 / 5.])
 
-    assert (interv1 / interv2).approx_equals(quotient, tol=10 ** -10)
+    assert (interv1 / interv2).approx_equals(quotient, atol=10 ** -10)
```
adler-j: `atol=1e-10`

adler-j:
Great changes, some minor comments then we should be fine.

adler-j:
Missclick on close...

odl/discr/partition.py
kohr-h: Oops

odl/discr/partition.py
kohr-h: Right

kohr-h:
Fixed the old stuff, will merge after tests have run.

PR #453, opened 2016-06-15
Issue 450  clintec spect data

niinimaki:
Added functionality to read in CLINTEC SPECT/CT data
`clintec_spect.py` contains 3 functions

- `read_clintec_CT_reconstruction` reads in CT reconstructions in Dicom form and returns an array of  Houndsfield units

- `linear_attenuation_map_from_HU_data` calls `read_clintec_CT_reconstruction` and converts HU to linear attenuation map and returns the corresponding array

- `read_clintec_raw_spect_data` reads the Dicom SPECT data and returns to and array

### DIFF
--- a/doc/source/refs.rst
+++ b/doc/source/refs.rst
@@ -20,6 +20,11 @@ References
    simultaneous equations*. Mathematics of computation, 33 (1965),
    pp 577--593.
 
+.. [Bro+2008] Brown, S, and Bailey, D L, and Willowson, K, and Baldock, C
+   *Investigation of the relationship between linear attenuation
+   coefficients and CT Houndsfield units using radionuclides for SPECT*.
+   Applied Radiation and Isotopes, 66 (2008), pp 1206--1212
+
 .. [BV2004] Boyd, S, and Vandenberghe, L. *Convex optimization*.
    Cambridge university press, 2004.
 
--- a/doc/source/refs.rst
+++ b/doc/source/refs.rst
@@ -26,13 +26,13 @@ References
 .. [Bro+2008] Brown, S, and Bailey, D L, and Willowson, K, and Baldock, C
    *Investigation of the relationship between linear attenuation
    coefficients and CT Houndsfield units using radionuclides for SPECT*.
-   Applied Radiation and Isotopes, 66 (2008), pp 1206--1212
+   Applied Radiation and Isotopes, 66 (2008), pp 1206--1212.
 
-..  Burger, C, and Goerres, G, and Schoenes, S, and Buck, A,
+.. [Bur+2002] Burger, C, and Goerres, G, and Schoenes, S, and Buck, A,
    and Lonn, A H R, and von Schulthess, G K *PET attenuation coefficient
    from CT images: experimental evaluation of the transformation of CT into
    PET 511-keV attenuation coefficients*.
-   European Journal of Nuclear Medicine, 29 (2002), pp 922--927
+   European Journal of Nuclear Medicine, 29 (2002), pp 922--927.
 
 .. [CP2011a] Chambolle, A and Pock, T. *A First-Order
    Primal-Dual Algorithm for Convex Problems with Applications to
--- a/doc/source/refs.rst
+++ b/doc/source/refs.rst
@@ -16,6 +16,9 @@ References
    composite and parallel-sum type monotone operators*. SIAM Journal
    on Optimization, 23.4 (2013), pp 2541--2565.
 
+.. [BV2004] Boyd, S, and Vandenberghe, L. *Convex optimization*.
+   Cambridge university press, 2004.
+
 .. [Bro1965] Broyden, C G. *A class of methods for solving nonlinear
    simultaneous equations*. Mathematics of computation, 33 (1965),
    pp 577--593.
@@ -25,8 +28,11 @@ References
    coefficients and CT Houndsfield units using radionuclides for SPECT*.
    Applied Radiation and Isotopes, 66 (2008), pp 1206--1212
 
-.. [BV2004] Boyd, S, and Vandenberghe, L. *Convex optimization*.
-   Cambridge university press, 2004.
+..  Burger, C, and Goerres, G, and Schoenes, S, and Buck, A,
+   and Lonn, A H R, and von Schulthess, G K *PET attenuation coefficient
+   from CT images: experimental evaluation of the transformation of CT into
+   PET 511-keV attenuation coefficients*.
+   European Journal of Nuclear Medicine, 29 (2002), pp 922--927
 
 .. [CP2011a] Chambolle, A and Pock, T. *A First-Order
    Primal-Dual Algorithm for Convex Problems with Applications to
--- a/odl/tomo/data/clintec_spect.py
+++ b/odl/tomo/data/clintec_spect.py
@@ -1,0 +1,120 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers for reading CLINTEC SPECT data formats.
+
+For the bilinear relationship of linear attenuation coefficients and
+Houndsfield units see, for example
+
+Investigation of the relationship between linear attenuation
+coefficients and CT Houndsfield units using radionuclides for SPECT
+S. Brown et al, Applied Radiation and Isotopes 66 (2008) 1206-1212
+
+and references therein.
+"""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import dicom
+import os
+from natsort import natsorted
+
+
+def read_clintec_CT_reconstruction(path):
+    """Reads the data in Dicom format and returns a HU CT volume.
+
+    Parameters
+    ----------
+    path : a `path` to the folder where Dicom data is
+
+    """
+
+    lstFilesDCM = []  # create an empty list
+    for dirName, subdirList, fileList in os.walk(path):
+        for filename in natsorted(fileList):
+            if ".dcm" in filename.lower():
+                # check whether the file ends by .dcm
+                lstFilesDCM.append(os.path.join(dirName, filename))
+            else:
+                lstFilesDCM.append(os.path.join(dirName, filename))
+
+    dataset = dicom.read_file(lstFilesDCM[0])
+
+    # Load dimensions based on the number of rows, columns, and slices
+    shape = (int(dataset.Rows), int(dataset.Columns), len(lstFilesDCM))
+    data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
+    # loop through all the DICOM files
+    for filenameDCM in lstFilesDCM:
+        # read the file
+        ds = dicom.read_file(filenameDCM)
+        # store the raw image data
+        data_array[:, :, lstFilesDCM.index(filenameDCM)] = ds.pixel_array
+
+    rescale_intercept = int(dataset.RescaleIntercept)
+    rescale_slope = int(dataset.RescaleSlope)
+
+    ct_volume = rescale_slope*np.float32(data_array) + rescale_intercept
+
+    return ct_volume, dataset
+
+
+def linear_attenuation_map_from_HU_data(path, a, b):
+    """Converts HU data to linear attenuation map.
+
+    Parameters
+    ----------
+    path : a `path` to the data folder
+    a, b : `tuple`
+        a = (a1, a2) where a1 `float` and a2 `float` are the intercepts
+            of the bilinear curve
+        b = (b1, b2) where b1 `float` and b2 `float` are the slopes
+            of the bilinear curve
+        a and b are parameters of a bilinear curve for converting
+        HU-values to linear attenuation map, such that
+        mu = a1 + b1 * ct_volume for HU < 0
+        mu = a2 + b2 * ct_volume for HU >= 0
+
+    """
+    a1, a2 = a
+    b1, b2 = b
+
+    ct_volume, dataset = read_clintec_CT_reconstruction(path)
+    (i1, j1, k1) = np.where(ct_volume < 0)
+    (i2, j2, k2) = np.where(ct_volume >= 0)
+
+    attenuation_map = np.zeros_like(ct_volume)
+    attenuation_map = np.zeros_like(ct_volume)
+    attenuation_map[i1, j1, k1] = a1 + b1 * ct_volume[i1, j1, k1]
+    attenuation_map[i2, j2, k2] = a2 + b2 * ct_volume[i2, j2, k2]
+    attenuation_map = np.absolute(attenuation_map)
+
+    return attenuation_map
+
+
+def read_clintec_raw_spect_data(path):
+    """Reads the raw spect Dicom data and retuns an array of the data."""
+
+    filenames = [f for f in os.listdir(path)]
+    data_dicom = os.path.join(path, filenames[0])
+    dataset = dicom.read_file(data_dicom)
+    spect_data = np.float32(dataset.pixel_array)
+
+    return spect_data
--- a/odl/tomo/data/clintec_spect.py
+++ b/odl/tomo/data/clintec_spect.py
@@ -71,7 +71,7 @@ def read_clintec_CT_reconstruction(path):
     rescale_intercept = int(dataset.RescaleIntercept)
     rescale_slope = int(dataset.RescaleSlope)
 
-    ct_volume = rescale_slope*np.float32(data_array) + rescale_intercept
+    ct_volume = rescale_slope * np.float32(data_array) + rescale_intercept
 
     return ct_volume, dataset
 
--- a/odl/tomo/data/clintec_spect.py
+++ b/odl/tomo/data/clintec_spect.py
@@ -18,13 +18,8 @@
 """Helpers for reading CLINTEC SPECT data formats.
 
 For the bilinear relationship of linear attenuation coefficients and
-Houndsfield units see, for example
+Houndsfield units see [Bro+2008]_ and references therein, for example.
 
-Investigation of the relationship between linear attenuation
-coefficients and CT Houndsfield units using radionuclides for SPECT
-S. Brown et al, Applied Radiation and Isotopes 66 (2008) 1206-1212
-
-and references therein.
 """
 
 # Imports for common Python 2/3 codebase
@@ -35,38 +30,37 @@ standard_library.install_aliases()
 import numpy as np
 import dicom
 import os
-from natsort import natsorted
 
 
 def read_clintec_CT_reconstruction(path):
-    """Reads the data in Dicom format and returns a HU CT volume.
+    """Reads the data in DICOM format and returns a HU CT volume.
 
     Parameters
     ----------
-    path : a `path` to the folder where Dicom data is
+    path : `str`
+        a path to the folder where DICOM data is
+
+    Returns
+    -------
+    ct_volume : `numpy.ndarray`
+        an array of CT reconstruction in Houndsfield units.
+    dataset : `DICOM` File Meta Information
 
     """
 
-    lstFilesDCM = []  # create an empty list
+    dcm_file_list = []
     for dirName, subdirList, fileList in os.walk(path):
-        for filename in natsorted(fileList):
-            if ".dcm" in filename.lower():
-                # check whether the file ends by .dcm
-                lstFilesDCM.append(os.path.join(dirName, filename))
-            else:
-                lstFilesDCM.append(os.path.join(dirName, filename))
+        for filename in sorted(fileList):
+            dcm_file_list.append(os.path.join(dirName, filename))
 
-    dataset = dicom.read_file(lstFilesDCM[0])
+    dataset = dicom.read_file(dcm_file_list[0])
 
-    # Load dimensions based on the number of rows, columns, and slices
-    shape = (int(dataset.Rows), int(dataset.Columns), len(lstFilesDCM))
+    shape = (int(dataset.Rows), int(dataset.Columns), len(dcm_file_list))
     data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
-    # loop through all the DICOM files
-    for filenameDCM in lstFilesDCM:
-        # read the file
-        ds = dicom.read_file(filenameDCM)
-        # store the raw image data
-        data_array[:, :, lstFilesDCM.index(filenameDCM)] = ds.pixel_array
+
+    for i, fname in enumerate(dcm_file_list):
+        ds = dicom.read_file(fname)
+        data_array[:, :, i] = ds.pixel_array
 
     rescale_intercept = int(dataset.RescaleIntercept)
     rescale_slope = int(dataset.RescaleSlope)
@@ -76,45 +70,60 @@ def read_clintec_CT_reconstruction(path):
     return ct_volume, dataset
 
 
-def linear_attenuation_map_from_HU_data(path, a, b):
-    """Converts HU data to linear attenuation map.
+def linear_attenuation_map_from_HU(volume, a, b):
+    """Converts Houndsfield units to a linear attenuation map.
 
     Parameters
     ----------
-    path : a `path` to the data folder
-    a, b : `tuple`
+    volume : `numpy.ndarray`
+        an array of Houndsfield units to be converted
+    a, b : `sequence`
         a = (a1, a2) where a1 `float` and a2 `float` are the intercepts
             of the bilinear curve
         b = (b1, b2) where b1 `float` and b2 `float` are the slopes
             of the bilinear curve
+        see `Notes` for furher information
+
+    Notes
+    -----
         a and b are parameters of a bilinear curve for converting
-        HU-values to linear attenuation map, such that
+        Houndsfield units to linear attenuation map, such that
         mu = a1 + b1 * ct_volume for HU < 0
         mu = a2 + b2 * ct_volume for HU >= 0
 
+    Returns
+    -------
+    attenuation_map : `numpy.ndarray`
+        An array of linear attenuation values.
     """
     a1, a2 = a
     b1, b2 = b
-
-    ct_volume, dataset = read_clintec_CT_reconstruction(path)
-    (i1, j1, k1) = np.where(ct_volume < 0)
-    (i2, j2, k2) = np.where(ct_volume >= 0)
-
-    attenuation_map = np.zeros_like(ct_volume)
-    attenuation_map = np.zeros_like(ct_volume)
-    attenuation_map[i1, j1, k1] = a1 + b1 * ct_volume[i1, j1, k1]
-    attenuation_map[i2, j2, k2] = a2 + b2 * ct_volume[i2, j2, k2]
+    mask = np.less(volume, 0)
+    attenuation_map = np.empty_like(volume)
+    attenuation_map[mask] = a1 + b1 * volume[mask]
+    attenuation_map[~mask] = a2 + b2 * volume[~mask]
     attenuation_map = np.absolute(attenuation_map)
-
     return attenuation_map
 
 
 def read_clintec_raw_spect_data(path):
-    """Reads the raw spect Dicom data and retuns an array of the data."""
+    """Reads the raw spect DICOM data and retuns an array of the data.
 
-    filenames = [f for f in os.listdir(path)]
+    Parameters
+    ----------
+    path : `str`
+        a path to the data folder
+
+    Returns
+    -------
+    spect_data : `numpy.ndarray`
+        an array of SPECT projection data
+    dataset : `DICOM` File Meta Information
+    """
+
+    filenames = list(os.listdir(path))
     data_dicom = os.path.join(path, filenames[0])
     dataset = dicom.read_file(data_dicom)
-    spect_data = np.float32(dataset.pixel_array)
+    spect_data = dataset.pixel_array
 
-    return spect_data
+    return spect_data, dataset
--- a/odl/tomo/data/clintec_spect.py
+++ b/odl/tomo/data/clintec_spect.py
@@ -19,7 +19,6 @@
 
 For the bilinear relationship of linear attenuation coefficients and
 Houndsfield units see [Bro+2008]_ and references therein, for example.
-
 """
 
 # Imports for common Python 2/3 codebase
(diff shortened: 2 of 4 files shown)


odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers for reading CLINTEC SPECT data formats.
+
+For the bilinear relationship of linear attenuation coefficients and
+Houndsfield units see, for example
+
+Investigation of the relationship between linear attenuation
+coefficients and CT Houndsfield units using radionuclides for SPECT
+S. Brown et al, Applied Radiation and Isotopes 66 (2008) 1206-1212
```
kohr-h: Put this in the [references file](https://github.com/odlgroup/odl/blob/5b5d5fd04347d175593acd4e8eef936aec03d289/doc/source/refs.rst) using our [citation style](https://github.com/odlgroup/odl/issues/279) (compare with the stuff in the file).

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+standard_library.install_aliases()
+
+import numpy as np
+import dicom
+import os
+from natsort import natsorted
+
+
+def read_clintec_CT_reconstruction(path):
+    """Reads the data in Dicom format and returns a HU CT volume.
```
kohr-h: Read
DICOM

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Parameters
+    ----------
+    path : a `path` to the folder where Dicom data is
```
kohr-h:
```
path : string
    blabla
```

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import dicom
+import os
+from natsort import natsorted
```
kohr-h: Why do we need this?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    Parameters
+    ----------
+    path : a `path` to the folder where Dicom data is
+
+    """
+
+    lstFilesDCM = []  # create an empty list
+    for dirName, subdirList, fileList in os.walk(path):
+        for filename in natsorted(fileList):
+            if ".dcm" in filename.lower():
```
kohr-h: `filename.lower().endswith('.dcm')`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+                # check whether the file ends by .dcm
+                lstFilesDCM.append(os.path.join(dirName, filename))
+            else:
+                lstFilesDCM.append(os.path.join(dirName, filename))
```
kohr-h: Exactly the same in the `else` case?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+
+    dataset = dicom.read_file(lstFilesDCM[0])
+
+    # Load dimensions based on the number of rows, columns, and slices
+    shape = (int(dataset.Rows), int(dataset.Columns), len(lstFilesDCM))
+    data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
```
kohr-h: `np.zeros_like(dataset.pixel_array)` should work?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    # loop through all the DICOM files
+    for filenameDCM in lstFilesDCM:
+        # read the file
+        ds = dicom.read_file(filenameDCM)
+        # store the raw image data
+        data_array[:, :, lstFilesDCM.index(filenameDCM)] = ds.pixel_array
```
kohr-h: Use
```python
for i, fname in enumerate(file_list):
    ds = dicom.read_file(fname)
    data_array[:, :, i] = ds.pixel_array
```

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+from natsort import natsorted
+
+
+def read_clintec_CT_reconstruction(path):
+    """Reads the data in Dicom format and returns a HU CT volume.
+
+    Parameters
+    ----------
+    path : a `path` to the folder where Dicom data is
+
```
kohr-h: Returns?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    rescale_intercept = int(dataset.RescaleIntercept)
+    rescale_slope = int(dataset.RescaleSlope)
+
+    ct_volume = rescale_slope * np.float32(data_array) + rescale_intercept
+
+    return ct_volume, dataset
+
+
+def linear_attenuation_map_from_HU_data(path, a, b):
```
kohr-h: Better not call it data, it has the connotation of projection data. `linear_attenuation_map_from_HU` should suffice.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    """Converts HU data to linear attenuation map.
+
+    Parameters
+    ----------
+    path : a `path` to the data folder
+    a, b : `tuple`
+        a = (a1, a2) where a1 `float` and a2 `float` are the intercepts
```
kohr-h: sequence

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+            of the bilinear curve
+        b = (b1, b2) where b1 `float` and b2 `float` are the slopes
+            of the bilinear curve
+        a and b are parameters of a bilinear curve for converting
+        HU-values to linear attenuation map, such that
+        mu = a1 + b1 * ct_volume for HU < 0
+        mu = a2 + b2 * ct_volume for HU >= 0
```
kohr-h: Put this more detailed explanation into a `Notes` section.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    """Converts HU data to linear attenuation map.
+
+    Parameters
+    ----------
+    path : a `path` to the data folder
+    a, b : `tuple`
+        a = (a1, a2) where a1 `float` and a2 `float` are the intercepts
+            of the bilinear curve
+        b = (b1, b2) where b1 `float` and b2 `float` are the slopes
+            of the bilinear curve
```
kohr-h: Point to the (to be created) Notes section, otherwise this is just jargon for users.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+        a and b are parameters of a bilinear curve for converting
+        HU-values to linear attenuation map, such that
+        mu = a1 + b1 * ct_volume for HU < 0
+        mu = a2 + b2 * ct_volume for HU >= 0
+
+    """
+    a1, a2 = a
+    b1, b2 = b
+
+    ct_volume, dataset = read_clintec_CT_reconstruction(path)
```
kohr-h: Don't intertwine this function with data reading. Just let it convert an existing array(-like).

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+
+def read_clintec_raw_spect_data(path):
+    """Reads the raw spect Dicom data and retuns an array of the data."""
+
+    filenames = [f for f in os.listdir(path)]
+    data_dicom = os.path.join(path, filenames[0])
+    dataset = dicom.read_file(data_dicom)
+    spect_data = np.float32(dataset.pixel_array)
+
+    return spect_data
```
kohr-h: This cannot be all. There must be geometry information stored in the files that we need to read as well.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    return attenuation_map
+
+
+def read_clintec_raw_spect_data(path):
+    """Reads the raw spect Dicom data and retuns an array of the data."""
+
+    filenames = [f for f in os.listdir(path)]
+    data_dicom = os.path.join(path, filenames[0])
+    dataset = dicom.read_file(data_dicom)
+    spect_data = np.float32(dataset.pixel_array)
```
kohr-h: Why do we convert to `float32`?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+        HU-values to linear attenuation map, such that
+        mu = a1 + b1 * ct_volume for HU < 0
+        mu = a2 + b2 * ct_volume for HU >= 0
+
+    """
+    a1, a2 = a
+    b1, b2 = b
+
+    ct_volume, dataset = read_clintec_CT_reconstruction(path)
+    (i1, j1, k1) = np.where(ct_volume < 0)
```
kohr-h: We don't need to get the index arrays. Just use
```python
negative = np.negative(ct_volume)  # More robust, allows array-like
attenuation_map[negative] = a1 + b1 * ct_volume[negative]
nonnegative = ~negative
attenuation_map[nonnegative] = a2 + b2 * ct_volume[nonnegative]
```

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    (i2, j2, k2) = np.where(ct_volume >= 0)
+
+    attenuation_map = np.zeros_like(ct_volume)
+    attenuation_map = np.zeros_like(ct_volume)
+    attenuation_map[i1, j1, k1] = a1 + b1 * ct_volume[i1, j1, k1]
```
kohr-h: We should consider avoiding temporary arrays, the volumes can be large. To use in-place arithmetic, this can be split into (see above comment)
```python
attenuation_map[negative] = ct_volume[negative]
attenuation_map[negative] *= b1
attenuation_map[negative] += a1
```
This code runs in fact a bit slower than the original one due to the multiple slicing. However, it's more memory friendly, so I would prefer it.

odl/tomo/data/clintec_spect.py
kohr-h: Here comes the even more memory-saving version 2:
```python
mask = np.negative(ct_volume)  # More robust, allows array-like
attenuation_map[mask] = a1 + b1 * ct_volume[mask]
np.logical_not(mask, out=mask)
attenuation_map[mask] = a2 + b2 * ct_volume[mask]
```

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    attenuation_map[i2, j2, k2] = a2 + b2 * ct_volume[i2, j2, k2]
+    attenuation_map = np.absolute(attenuation_map)
+
+    return attenuation_map
+
+
+def read_clintec_raw_spect_data(path):
+    """Reads the raw spect Dicom data and retuns an array of the data."""
+
+    filenames = [f for f in os.listdir(path)]
```
kohr-h: `list(os.listdir(path))` should do.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    data_dicom = os.path.join(path, filenames[0])
```
kohr-h: What is special about the first entry?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+and references therein.
+"""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import dicom
```
kohr-h: This will fail on default installations since the `dicom` dependency is missing. Add this to the `setup.py` `extra_dependencies` under a name like `dicom`.

odl/tomo/data/clintec_spect.py
niinimaki: To identify numbers anywhere in a string and to sort them naturally, as [in here](https://github.com/odlgroup/odl/blob/issue-450__clintec-spect-data/odl/tomo/data/clintec_spect.py#L52).

But if there is another way to do this without `natsort`, that could be used instead.

odl/tomo/data/clintec_spect.py
niinimaki: True, I'll remove this.

odl/tomo/data/clintec_spect.py
niinimaki: That is true and I have modified this such that it also returns the `dataset`.

odl/tomo/data/clintec_spect.py
niinimaki: I'll remove this.

odl/tomo/data/clintec_spect.py
niinimaki: Contains the DICOM data, in the case of SPECT this list actually consists of one entry.

odl/tomo/data/clintec_spect.py
adler-j: There has to be another way, adding dependencies like this is not very good. Worst case we implement it ourselves. See [Stackoverflow](http://stackoverflow.com/questions/4836710/does-python-have-a-built-in-function-for-string-natural-sort) for a discussion.

odl/tomo/data/clintec_spect.py
adler-j:
```
path : str
    blabla
```

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+def read_clintec_CT_reconstruction(path):
+    """Reads the data in Dicom format and returns a HU CT volume.
+
+    Parameters
+    ----------
+    path : a `path` to the folder where Dicom data is
+
+    """
+
+    lstFilesDCM = []  # create an empty list
```
adler-j: Also `# create an empty list` is overdocumentation, that is quite obvious from the code. If you want to write doc here, it would be to explain *why* the list is created.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    a1, a2 = a
+    b1, b2 = b
+
+    ct_volume, dataset = read_clintec_CT_reconstruction(path)
+    (i1, j1, k1) = np.where(ct_volume < 0)
+    (i2, j2, k2) = np.where(ct_volume >= 0)
+
+    attenuation_map = np.zeros_like(ct_volume)
+    attenuation_map = np.zeros_like(ct_volume)
```
adler-j: duplicate

odl/tomo/data/clintec_spect.py
adler-j: Have you tried this code? Negative is not "is negative" but "the negative of". I suppose you want to write `mask = np.less_equal(a, 0)`?

odl/tomo/data/clintec_spect.py
niinimaki: This does not work
`negative = np.negative(ct_volume)`
`attenuation_map[negative] = ct_volume[negative]`

Doesn't `np.negative` take the element-wise numerical negatives of the values, but I would like to get the indices in the array for the negative (&positive) values...

adler-j: This creates copys anyway since `negative` is not a slice, but an indicator array. To be "very efficient" here, I'd go for one copy:

```python
mask = np.less(ct_volume, 0)

attenuation_map = np.empty_like(ct_volume)

negative_part = ct_volume[mask]

negative_part *= b1
negative_part += a1

attenuation_map[mask] = negative_part

mask = np.logical_not(mask, out=mask)
positive_part = ct_volume[mask]

positive_part *= b2
positive_part += a2

attenuation_map[mask] = positive_part
```

With that said, I think we shouldnt overoptimize this at this point. The straightforward solution

```python
mask = np.less(ct_volume, 0)
attenuation_map = np.empty_like(ct_volume)
attenuation_map[mask] = a1 + b1 * ct_volume[mask]
attenuation_map[mask] = a2 + b2 * ct_volume[~mask]
```

should do fine.

odl/tomo/data/clintec_spect.py
niinimaki: This has been replaced by

`for filename in sorted(fileList):`

Seems to work. So `natsort` is no longer needed.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    a, b : `sequence`
+        a = (a1, a2) where a1 `float` and a2 `float` are the intercepts
+            of the bilinear curve
+        b = (b1, b2) where b1 `float` and b2 `float` are the slopes
+            of the bilinear curve
+        see `Notes` for furher information
+
+    Notes
+    -----
+        a and b are parameters of a bilinear curve for converting
```
adler-j: bilinear is perhaps not the correct word here

niinimaki:
Revision of `clintec_spect`:

- moved the reference from `clintec_spect` into `refs.rst`

- added dicom to `setup` as `extra_require`

- improved the docstrings in `clintec_spect`

- improved the code of `clintec_spect` according to suggestions

odl/tomo/data/clintec_spect.py
adler-j: Even after adding it to `extra_dependencies` this will cause a fail. We need to have an include guard here, the classical

```python
try:
    import dicom
    DICOM_AVAILABLE = True
except ImportError:
    DICOM_AVAILABLE = False
```

odl/tomo/data/clintec_spect.py
niinimaki: bilinear is used for example in `[Bro+2008]` to which I refer.
There are two linear curves a one for HU < 0 and another for HU >= 0.
So piecewise linear would work also but bilinear seems to be used in the emission tomography literature.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    -------
+    ct_volume : `numpy.ndarray`
+        an array of CT reconstruction in Houndsfield units.
+    dataset : `DICOM` File Meta Information
+
+    """
+
+    dcm_file_list = []
+    for dirName, subdirList, fileList in os.walk(path):
+        for filename in sorted(fileList):
```
adler-j: This could be in the "wrong" order:

```python
>>> sorted(['a1', 'a2', 'a10', 'a11', 'a20'])
['a1', 'a10', 'a11', 'a2', 'a20']
```

I suppose that was why you used `natsort`?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    dataset = dicom.read_file(dcm_file_list[0])
+
+    shape = (int(dataset.Rows), int(dataset.Columns), len(dcm_file_list))
+    data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
+
+    for i, fname in enumerate(dcm_file_list):
+        ds = dicom.read_file(fname)
+        data_array[:, :, i] = ds.pixel_array
+
+    rescale_intercept = int(dataset.RescaleIntercept)
```
adler-j: why `int`?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers for reading CLINTEC SPECT data formats.
+
+For the bilinear relationship of linear attenuation coefficients and
+Houndsfield units see [Bro+2008]_ and references therein, for example.
+
```
adler-j: Remove blank line

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+import os
+
+
+def read_clintec_CT_reconstruction(path):
+    """Reads the data in DICOM format and returns a HU CT volume.
+
+    Parameters
+    ----------
+    path : `str`
+        a path to the folder where DICOM data is
```
adler-j: `Path of the folder where DICOM data is stored`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    dcm_file_list = []
+    for dirName, subdirList, fileList in os.walk(path):
+        for filename in sorted(fileList):
+            dcm_file_list.append(os.path.join(dirName, filename))
+
+    dataset = dicom.read_file(dcm_file_list[0])
+
+    shape = (int(dataset.Rows), int(dataset.Columns), len(dcm_file_list))
+    data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
```
adler-j: `np.empty`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    dcm_file_list = []
+    for dirName, subdirList, fileList in os.walk(path):
+        for filename in sorted(fileList):
+            dcm_file_list.append(os.path.join(dirName, filename))
+
+    dataset = dicom.read_file(dcm_file_list[0])
+
+    shape = (int(dataset.Rows), int(dataset.Columns), len(dcm_file_list))
```
adler-j: We usually store the angles in the first index, why use the last here?

odl/tomo/data/clintec_spect.py
niinimaki: Yes. Is there another options for this not using `natsort`?

odl/tomo/data/clintec_spect.py
niinimaki: should be `float` instead I think.

odl/tomo/data/clintec_spect.py
niinimaki: No strong reason, I'll change such that angles are first.

odl/tomo/data/clintec_spect.py
adler-j: Yes several, e.g. [this](http://stackoverflow.com/questions/2545532/python-analog-of-natsort-function-sort-a-list-using-a-natural-order-algorithm).

setup.py
```diff
@@ -137,8 +137,9 @@ def run_tests(self):
         'show': 'matplotlib',
         'fftw': 'pyfftw',
         'pywavelets': 'Pywavelets',
-        'scikit' : 'scikit-image',
-        'all': test_requires + ['matplotlib', 'pyfftw', 'Pywavelets', 'scikit-image']
+        'scikit': 'scikit-image',
+        'dicom': 'dicom',
```
adler-j: Are we sure this is not `pydicom`?

```bash
$ pip install dicom
...
No matching distribution found for dicom
$ pip install pydicom
...
Installing collected packages: pydicom
```

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Returns
+    -------
+    spect_data : `numpy.ndarray`
+        an array of SPECT projection data
+    dataset : `DICOM` File Meta Information
+    """
+
+    filenames = list(os.listdir(path))
+    data_dicom = os.path.join(path, filenames[0])
```
adler-j: How do we know it is the first one? What is the ordering convention of `os.listdir`?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    Parameters
+    ----------
+    path : `str`
+        a path to the data folder
+
+    Returns
+    -------
+    spect_data : `numpy.ndarray`
+        an array of SPECT projection data
+    dataset : `DICOM` File Meta Information
```
adler-j: is `DICOM` a dead link?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    return ct_volume, dataset
+
+
+def linear_attenuation_map_from_HU(volume, a, b):
+    """Converts Houndsfield units to a linear attenuation map.
+
+    Parameters
+    ----------
+    volume : `numpy.ndarray`
```
adler-j: `array-like`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+            dcm_file_list.append(os.path.join(dirName, filename))
+
+    dataset = dicom.read_file(dcm_file_list[0])
+
+    shape = (int(dataset.Rows), int(dataset.Columns), len(dcm_file_list))
+    data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
+
+    for i, fname in enumerate(dcm_file_list):
+        ds = dicom.read_file(fname)
+        data_array[:, :, i] = ds.pixel_array
```
adler-j: My old comment got removed here, but why do we store the "angles" in the last index and not the first like we do in `RayTransform`?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns
+    -------
+    ct_volume : `numpy.ndarray`
+        an array of CT reconstruction in Houndsfield units.
+    dataset : `DICOM` File Meta Information
+
+    """
+
+    dcm_file_list = []
+    for dirName, subdirList, fileList in os.walk(path):
```
adler-j: `subdirList` unused, replace with `_`:

```
for dirName, _, fileList in os.walk(path):
```

odl/tomo/data/clintec_spect.py
adler-j: Short edit here: if this reads in a volume, we use the x, y, z convention, so then it should be last. Perhaps I missed that.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    path : `str`
+        a path to the folder where DICOM data is
+
+    Returns
+    -------
+    ct_volume : `numpy.ndarray`
+        an array of CT reconstruction in Houndsfield units.
```
adler-j: `Array of CT reconstruction in Houndsfield units.`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import dicom
+import os
+
+
+def read_clintec_CT_reconstruction(path):
```
adler-j: In what way is this specific to `clintec`? It looks exactly like the MAYO data to me.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    data_array = np.zeros(shape, dtype=dataset.pixel_array.dtype)
+
+    for i, fname in enumerate(dcm_file_list):
+        ds = dicom.read_file(fname)
+        data_array[:, :, i] = ds.pixel_array
+
+    rescale_intercept = int(dataset.RescaleIntercept)
+    rescale_slope = int(dataset.RescaleSlope)
+
+    ct_volume = rescale_slope * np.float32(data_array) + rescale_intercept
```
adler-j: `np.float32(data_array)` should be redundant?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    Parameters
+    ----------
+    volume : `numpy.ndarray`
+        an array of Houndsfield units to be converted
+    a, b : `sequence`
+        a = (a1, a2) where a1 `float` and a2 `float` are the intercepts
+            of the bilinear curve
+        b = (b1, b2) where b1 `float` and b2 `float` are the slopes
+            of the bilinear curve
+        see `Notes` for furher information
```
adler-j: Is `Notes` a link here?

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    -----
+        a and b are parameters of a bilinear curve for converting
+        Houndsfield units to linear attenuation map, such that
+        mu = a1 + b1 * ct_volume for HU < 0
+        mu = a2 + b2 * ct_volume for HU >= 0
+
+    Returns
+    -------
+    attenuation_map : `numpy.ndarray`
+        An array of linear attenuation values.
```
adler-j: ~~An~~

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    mask = np.less(volume, 0)
+    attenuation_map = np.empty_like(volume)
+    attenuation_map[mask] = a1 + b1 * volume[mask]
+    attenuation_map[~mask] = a2 + b2 * volume[~mask]
+    attenuation_map = np.absolute(attenuation_map)
+    return attenuation_map
+
+
+def read_clintec_raw_spect_data(path):
+    """Reads the raw spect DICOM data and retuns an array of the data.
```
adler-j: `Read raw SPECT DICOM data to array.`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    path : `str`
+        a path to the data folder
+
+    Returns
+    -------
+    spect_data : `numpy.ndarray`
+        an array of SPECT projection data
```
adler-j: `Array of SPECT projection data.`

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+    rescale_intercept = int(dataset.RescaleIntercept)
+    rescale_slope = int(dataset.RescaleSlope)
+
+    ct_volume = rescale_slope * np.float32(data_array) + rescale_intercept
+
+    return ct_volume, dataset
+
+
+def linear_attenuation_map_from_HU(volume, a, b):
+    """Converts Houndsfield units to a linear attenuation map.
```
adler-j: Convert~~s~~ Houndsfield units to linear attenuation coefficent.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    volume : `numpy.ndarray`
+        an array of Houndsfield units to be converted
```
adler-j: ~~an~~

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,129 @@
...
+
+    rescale_intercept = int(dataset.RescaleIntercept)
+    rescale_slope = int(dataset.RescaleSlope)
+
+    ct_volume = rescale_slope * np.float32(data_array) + rescale_intercept
+
+    return ct_volume, dataset
+
+
+def linear_attenuation_map_from_HU(volume, a, b):
```
adler-j: Why do we use "linear attenuation **map**" but not ""HU **map**", I feel like we should be consistent here.

adler-j:
Review done, mostly minor stuff.

Do we have any code that can use the output of this, or is it simply dangling code?

setup.py
niinimaki: One installs `pydicom` but imports it as [`import dicom`](http://pydicom.readthedocs.io/en/stable/getting_started.html).

So maybe this should be 
`'pydicom': 'dicom'`
or?

odl/tomo/data/clintec_spect.py
niinimaki: As far as I know spect data in DICOM format is just one file, hence it first and only entry in the list...

odl/tomo/data/clintec_spect.py
niinimaki: I'll remove the ticks.

odl/tomo/data/clintec_spect.py
niinimaki: > We usually store the angles in the first index, why use the last here?

> Short` edit here: if this reads in a volume, we use the x, y, z convention, so then it should be last. Perhaps I missed that.

Is this not the same comment as above?? under line 58??

odl/tomo/data/clintec_spect.py
niinimaki: Well if they really are same, maybe we then should have one functions that does this, the name of which not referring to any specific clinic..

setup.py
adler-j: Other way around. To the left of the colon is our name for it, to the right is the pip package name.

odl/tomo/data/clintec_spect.py
adler-j: Shouldnt we do an assert then that `filenames` has exactly one element? This feels like a bug waiting to happen.

odl/tomo/data/clintec_spect.py
adler-j: Yes, just me being confused.

odl/tomo/data/clintec_spect.py
adler-j: Perhaps keep this for now and I can merge this when we fix the MAYO data.

odl/tomo/data/clintec_spect.py
adler-j: Suppose we stay with that then!

setup.py
```diff
@@ -137,8 +137,9 @@ def run_tests(self):
...
+        'all': test_requires + ['matplotlib', 'pyfftw', 'Pywavelets', 'scikit-image', 'dicom']
```
adler-j: `dicom` -> `pydicom`

niinimaki:
> Do we have any code that can use the output of this, or is it simply dangling code?

Not yet, except that `linear_attenuation_map_from_HU` can be used to get the `attenuation` for `spect_trafo`, provided that one has a volume in HU...

odl/tomo/data/clintec_spect.py
kohr-h: You're right. I tried it with a trivial example which happened to work :-)

odl/tomo/data/clintec_spect.py
kohr-h: I was guessing something like that. Can you write a comment?

odl/tomo/data/clintec_spect.py
kohr-h: I was about to write the same thing, but if I understood correctly, this is for reading a reconstruction by axial slices, so the length should come last.

odl/tomo/data/clintec_spect.py
niinimaki: With respect to this I think that what @adler-j  wrote 

> Shouldnt we do an assert then that filenames has exactly one element? This feels like a bug waiting to happen.

might be a good idea...

odl/tomo/data/clintec_spect.py
niinimaki: The options of not using `natsort` need `re`, wouldn't this be just another dependency?

adler-j: `re` is a core package so that is fine, kindoff like using `os` or `sys`.

odl/tomo/data/clintec_spect.py
niinimaki: If I understand what you mean, aren't these data/scanner dependent??

niinimaki:
Added one more reference related to `linear_attenuation_from_HU`.
Improved further the docstrings and code of `clintec_spect` according to suggestions.

odl/tomo/data/clintec_spect.py
adler-j: Well, in general you want these types of methods to be quite specific in what they output, but ofc it depends on the call patterns. 

Thats why I feel like we need to get some code that actually calls this into ODL before we merge. An example or similar.

doc/source/refs.rst
```diff
@@ -16,12 +16,23 @@ References
...
 .. [Bro1965] Broyden, C G. *A class of methods for solving nonlinear
    simultaneous equations*. Mathematics of computation, 33 (1965),
    pp 577--593.
 
-.. [BV2004] Boyd, S, and Vandenberghe, L. *Convex optimization*.
-   Cambridge university press, 2004.
+.. [Bro+2008] Brown, S, and Bailey, D L, and Willowson, K, and Baldock, C
+   *Investigation of the relationship between linear attenuation
+   coefficients and CT Houndsfield units using radionuclides for SPECT*.
+   Applied Radiation and Isotopes, 66 (2008), pp 1206--1212
```
kohr-h: Almost :-) Final dot missing.

doc/source/refs.rst
```diff
@@ -16,12 +16,23 @@ References
...
+
+..  Burger, C, and Goerres, G, and Schoenes, S, and Buck, A,
+   and Lonn, A H R, and von Schulthess, G K *PET attenuation coefficient
+   from CT images: experimental evaluation of the transformation of CT into
+   PET 511-keV attenuation coefficients*.
+   European Journal of Nuclear Medicine, 29 (2002), pp 922--927
```
kohr-h: `[Bur+2002]` and final dot missing.

niinimaki:
Fixed the missing dot etc in the `refs.rst` related to `clintec_spect`.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,152 @@
...
+    dataset : DICOM File Meta Information
+    """
+
+    filenames = list(os.listdir(path))
+    if len(filenames) == 1:
+        data_dicom = os.path.join(path, filenames[0])
+        dataset = dicom.read_file(data_dicom)
+        spect_data = dataset.pixel_array
+    else:
+        raise NotImplementedError('folder contais more than one item')
```
adler-j: I'd go for `ValueError` here.

odl/tomo/data/clintec_spect.py
```diff
@@ -0,0 +1,152 @@
...
+        See [Bro+2008]_ and [Bur+2002]_ for example.
+
+    """
+    a1, a2 = a
+    b1, b2 = b
+    mask = np.less(volume, 0)
+    attenuation = np.empty_like(volume)
+    attenuation[mask] = a1 + b1 * volume[mask]
+    attenuation[~mask] = a2 + b2 * volume[~mask]
+    attenuation = np.absolute(attenuation)
```
adler-j: This step is not mentioned in the documentation above and could come as a surprise to users.

odl/tomo/data/clintec_spect.py
kohr-h: If it's always one file, why don't we simply give the filename instead of a folder?

adler-j:
Starting to look very good by now. My only major concern is that this is dangling code with no examples nor tests and no obvious way of using it.

What use-case do we intend it for, can we somehow merge this into that branch instead of master?

odl/tomo/data/clintec_spect.py
niinimaki: That could actually be a good idea.

odl/tomo/data/clintec_spect.py
niinimaki: This step is to ensure that there are no negative values in the attenuation coefficients (as there should not be). Negative values can occur due to the fact that the parameters `a` and `b` are approximations (evaluated for example for different scanners) and contain some errors them selves.  
But I see your point.

niinimaki:
Making a new PR into the SPECT trafo branch.

PR #454, opened 2016-06-15
ENH: Add optimal information transport

adler-j:


odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,167 @@
...
+        # solve for v
+        v = grad.range.element()
+        for i in range(u.size):
+            v[i] = poisson_solver(u[i])
+            v[i] -= v[i][0]
+
+        new_points = op.domain.points().T
+        for i in range(tmp.size):
+            new_points[i] -= eps * v[i].ntuple.asarray()
+
```
chongchenmath: The phi inverse in step 10 is not updated?

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,167 @@
...
+        # print(np.min(new_points), np.max(new_points))
+
+        I.assign(op.domain.element(I.interpolation(new_points, bounds_check=False)))
```
chongchenmath: As the last comment, it seems that the I should compose with the new phi inverse, not (Id - eps v).

odl/deform/optimal_information_transport.py
adler-j: This currently does not provide a way to compute phi inverse, mainly because we have no way of representing it in a simple manner. The way holger discussed should work however, so we could implement that.

odl/deform/optimal_information_transport.py
adler-j: This is not exactly the algorithm as written in the paper, instead I update I and the jacobian in each iteration. The result should be the same (but this is perhaps slightly less numerically stable).

adler-j:
Another comment that can be made about this implementation and that you notice while using it is that the current implementation is not mass, nor range preserving. 

This is because we interpolate in each step, and the interpolated value will lie "in between" values in the range of the function. Thus if the maximum at T=0 is 1, the maximum at T=dt will be slightly smaller than 1, this is iterated for each iteration and causes a quite noticable effect over time.

The only way of improving this with the current method is to use a higher resolution in the interpolation, but that isnt really feasible.

Another way would be to implement the method that @kohr-h and @ozanoktem suggested, wherein we actually compute and update the displacement field in each iteration. That way we dont touch the values, and hence we dont need to worry about the range of the function changing. That implementation could be slower than this one, so perhaps we should have both?

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,234 @@
...
+            DPhiJacobian.interpolation(new_points, bounds_check=False))
+
+        if callback is not None:
+            callback(I)
+
+
+# Code for the example starts here
+
+
+class GradSOperator(Operator):
```
adler-j: This can be done with

```
gradient_op = op.adjoint * odl.ResidualOperator(op, rhs)
```

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,234 @@
...
+        else:
+            snr = 10.0 * np.log10(s_power/n_power)
+
+        return snr
+
+    else:
+        return float('inf')
+
+
+if __name__ == '__main__':
```
adler-j: This needs to be moved to odl/examples

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,234 @@
...
+        ----------
+        template : `DiscreteLpVector`
+
+        g: 'DiscreteLpVector'
+        """
+        out = self.op.adjoint(self.op(template) - g)
+        return out
+
+
+def SNR(signal, noise, impl='general'):
```
adler-j: As discussed in the other pull, this goes in util

odl/deform/optimal_information_transport.py
chongchenmath: Yes, that is more self-contained. Done.

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,205 @@
...
+    # Add white Gaussion noise onto the noiseless data
+    noise = odl.util.white_noise(ray_trafo.range) * 3
+
+    # Create the noisy projection data
+    noise_proj_data = proj_data + noise
+
+    # Compute the signal-to-noise ratio in dB
+    snr = SNR(proj_data, noise, impl='dB')
+
+    # Output the signal-to-noise ratio
```
adler-j: This is overdocumenting, the code is quite easy to read without these comments. In general comments should tell the reader **why** something is done, especially when it is not obvious. **What** is done should in general be obvious from the code (if it is not, then documenting is needed).

I would write

```python
# Create projection data
proj_data = ray_trafo(ground_truth)
noise_proj_data = proj_data + odl.util.white_noise(ray_trafo.range) * 3
print('snr = {!r}'.format(SNR(proj_data, noise, impl='dB')))
```

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,205 @@
...
+    print('snr = {!r}'.format(snr))
+
+    # Maximum iteration number
+    niter = 2000
+
+    callback = odl.solvers.CallbackShow(
```
adler-j:
```python
callback = (odl.solvers.CallbackShow('iterates', display_step=50) & 
            odl.solvers.CallbackPrintIteration())
```

odl/deform/optimal_information_transport.py
chongchenmath: This has moved into util.

odl/util/utility.py
```diff
@@ -417,6 +417,36 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
     return tuple(indices)
 
 
+def snr(signal, noise, impl='general'):
+    """Compute the signal-to-noise ratio.
+    This compute::
```
adler-j: This needs to be fleshed out with more information. See [the documentation on how to document](http://odl.readthedocs.io/dev/document.html) for examples and style.

odl/util/utility.py
```diff
@@ -417,6 +417,36 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+        impl='dB'
+            SNR = 10 * log10 (
+                s_power / n_power)
+    Parameters
+    ----------
+    signal : projection
+    noise : white noise
+    impl : implementation method
+    """
+    if np.abs(np.asarray(noise)).sum() != 0:
```
adler-j: This is equivalent to

```
if np.any(np.asarray(noise) != 0):
```

which conveys the goal better.

odl/util/utility.py
```diff
@@ -417,6 +417,36 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+        ave2 = np.sum(noise)/noise.size
+        s_power = np.sqrt(np.sum((signal - ave1) * (signal - ave1)))
+        n_power = np.sqrt(np.sum((noise - ave2) * (noise - ave2)))
+        if impl == 'general':
+            snr = s_power/n_power
+        else:
+            snr = 10.0 * np.log10(s_power/n_power)
+
+        return snr
+
```
adler-j: Remove blank row

odl/util/utility.py
```diff
@@ -417,6 +417,36 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    impl : implementation method
+    """
+    if np.abs(np.asarray(noise)).sum() != 0:
+        ave1 = np.sum(signal)/signal.size
+        ave2 = np.sum(noise)/noise.size
+        s_power = np.sqrt(np.sum((signal - ave1) * (signal - ave1)))
+        n_power = np.sqrt(np.sum((noise - ave2) * (noise - ave2)))
+        if impl == 'general':
+            snr = s_power/n_power
+        else:
```
adler-j:
```python
elif impl == 'dB':
    ...
else:
    raise ValueError('unknown `impl` {}'.format(impl))
```

odl/util/utility.py
```diff
@@ -417,6 +417,36 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+            SNR = 10 * log10 (
+                s_power / n_power)
+    Parameters
+    ----------
+    signal : projection
+    noise : white noise
+    impl : implementation method
+    """
+    if np.abs(np.asarray(noise)).sum() != 0:
+        ave1 = np.sum(signal)/signal.size
```
adler-j: Use built in functions where possible. Here [np.std](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.std.html):

```python
s_power = np.std(signal)
n_power = np.std(noise)
```

odl/util/utility.py
```diff
@@ -417,6 +417,36 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+        ave2 = np.sum(noise)/noise.size
+        s_power = np.sqrt(np.sum((signal - ave1) * (signal - ave1)))
+        n_power = np.sqrt(np.sum((noise - ave2) * (noise - ave2)))
+        if impl == 'general':
+            snr = s_power/n_power
```
adler-j: You should run all tests before you commit, this will fail the PEP8 checker. To run all tests do:

```bash
$ py.test --pep8 --largescale
```

This should be written

```python
snr = s_power / n_power
```

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,176 @@
...
+    I0 = np.rot90(plt.imread(I0name).astype('float'), -1)[::2, ::2]
+    I1 = np.rot90(plt.imread(I1name).astype('float'), -1)[::2, ::2]
+
+    # Discrete reconstruction space: discretized functions on the rectangle
+    space = odl.uniform_discr(
+        min_corner=[-16, -16], max_corner=[16, 16], nsamples=[128, 128],
+        dtype='float32', interp='linear')
+
+    # Make a parallel beam geometry with flat detector
+    # Angles: uniformly spaced, n = 360, min = 0, max = 2 * pi
```
adler-j: This documentation is wrong

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,176 @@
...
+            callback(I)
+
+
+# Code for the example starts here
+
+
+if __name__ == '__main__':
+    import odl
+
+    I0name = '../../ddmatch/Example3 letters/c_highres.png'
```
adler-j: We cannot have references to external data in ODL, this needs to be changed to something we support internally.

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,176 @@
...
+    DPhiJacobian = gradS.domain.one()
+
+    grad = Gradient(gradS.domain, method='central')
+
+    # We solve poisson using the fourier transform
+    # ft = FourierTransform(op.domain)
+    ft = FourierTransform(gradS.domain)
+    k2_values = sum((ft.range.points() ** 2).T)
+    k2 = ft.range.element(np.maximum(np.abs(k2_values), 0.01))
+    poisson_solver = ft.inverse * (1 / k2) * ft
```
adler-j: This is currently quite a hack. What should be done is that we should set the fourier transform to 0 in the 0, that is, no DC component.

```python
k2_values = sum((ft.range.points() ** 2).T)
k2_values[k2_values == 0] = np.inf
k2 = ft.range.element(np.abs(k2_values))
poisson_solver = ft.inverse * (1 / k2) * ft
```

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,176 @@
...
+        tmp = tmp.space.element([tp * PhiStarX for tp in tmp])
+
+        u = sigma * grad(1 - np.sqrt(DPhiJacobian)) - 2 * tmp
+
+        # solve for v
+        v = grad.range.element()
+
+        for i in range(u.size):
+            v[i] = poisson_solver(u[i])
+            v[i] -= v[i][0]
```
adler-j: This is a very arbitrary way of me to enforce the boundary conditions (v=0 on the boundary), we should have a better way.

odl/deform/optimal_information_transport.py
```diff
@@ -15,21 +15,13 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""
```
adler-j: This file still needs a docstring that says what it does

odl/deform/optimal_information_transport.py
chongchenmath: Separate the solver and its test. The test is in odl/examples now.

odl/deform/optimal_information_transport.py
chongchenmath: The test pictures are moved into odl/examples/pictures.

adler-j: Not 100% convinced we need even more data in ODL. Best would be to have analytic test cases.  The disk <-> submarine should do quite well, no?

odl/deform/optimal_information_transport.py
chongchenmath: Yes. We may try the periodic boundary conditions.

odl/util/utility.py
chongchenmath: May be np.var? [https://en.wikipedia.org/wiki/Signal-to-noise_ratio](url)

odl/util/utility.py
chongchenmath: Done.

odl/util/utility.py
adler-j: Well, your old implementation used the square root, which implies the standard deviation. Wikipedia does however seem to refer to the variance, so we should probably use that instead.

chongchenmath: All done -:)

odl/deform/optimal_information_transport.py
adler-j: Then we need to handle that in the interpolation somehow.

odl/deform/optimal_information_transport.py
chongchenmath: For disc to submarine example, a disc phantom has been added into the branch of linearized_deformations. Need to add it in this branch?

adler-j: You can `cherry-pick` [see some info here](http://stackoverflow.com/questions/9339429/what-does-cherry-picking-a-commit-with-git-mean) that specific commit to this branch.

odl/deform/optimal_information_transport.py
chongchenmath: A simple docstring is added.

odl/deform/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+"""Deformation related solvers."""
+
+from __future__ import absolute_import
+
+__all__ = ()
+
+from . import optimal_information_transport
+from .optimal_information_transport import *
+__all__ += optimal_information_transport.__all__
```
adler-j: Missing newline

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,87 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Solver for the shape-based reconstruction using
+optimal information transportation.
+
+The model is:
```
adler-j: This should be moved to the documentation of the function `optimal_information_transport_solver`

odl/util/utility.py
```diff
@@ -417,6 +417,33 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
     return tuple(indices)
 
 
+def snr(signal, noise, impl):
+    """Compute the signal-to-noise ratio.
+    This compute::
```
adler-j: This doc needs to be updated to conform to the ODL standard.

http://odl.readthedocs.io/dev/document.html

chongchenmath: Updated.

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
     return tuple(indices)
 
 
+def snr(signal, noise, impl):
+    """Compute the signal-to-noise ratio.
+
+    Parameters
+    ----------
+    signal :
```
adler-j:
```
signal : array-like
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+        Noiseless data.
+    noise :
```
adler-j:
```
signal : `array-like`
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    -------
+    The value of signal-to-noise ratio.
+        If the power of noise is zero, then the return is 'inf',
+        otherwise, the return is the computed value.
+    """
+    if np.any(noise):
+        s_power = np.var(signal)
+        n_power = np.var(noise)
+        if impl == 'general':
+            snr = s_power / n_power
```
adler-j:
```
return s_power / n_power
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+        elif impl == 'dB':
+            snr = 10.0 * np.log10(s_power / n_power)
```
adler-j:
```
10.0 * np.log10(s_power / n_power)
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+        else:
+            raise ValueError('unknown `impl` {}'.format(impl))
+        return snr
```
adler-j: Can be removed after the above changes

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+def snr(signal, noise, impl):
+    """Compute the signal-to-noise ratio.
+
+    Parameters
+    ----------
+    signal :
+        Noiseless data.
+    noise :
+        White noise.
+    impl : implementation method
```
adler-j:
```
impl : {'general', 'dB'}
    Implementation method.
    'general' means SNR = s_power / n_power,
    'dB' means SNR = 10 * log10 (s_power / n_power).
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
 
+def snr(signal, noise, impl):
+    """Compute the signal-to-noise ratio.
+
+    Parameters
+    ----------
+    signal :
+        Noiseless data.
+    noise :
+        White noise.
```
adler-j: Does it need to be white noise?

chongchenmath: No necessary. The white noise is the general case.

odl/util/utility.py
chongchenmath: Done.

odl/util/utility.py
chongchenmath: Done.

examples/optimal_information_transport_test.py
```diff
@@ -0,0 +1,117 @@
...
+Example of shape-based image reconstruction
+using optimal information transformation.
+"""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+import numpy as np
+import matplotlib.pyplot as plt
+from odl.util import snr
```
adler-j: Just do a `import odl` and then use these as `odl.util.snr` instead.

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,87 @@
...
+from odl.discr import Gradient
+from odl.trafos import FourierTransform
+standard_library.install_aliases()
+
+__all__ = ('optimal_information_transport_solver',)
+
+
+def optimal_information_transport_solver(gradS, I, niter, eps,
+                                         sigma, callback=None):
+
```
adler-j: We should now add documentation here, similar to the doc in [chambolle pock](https://github.com/odlgroup/odl/blob/3543720d91841c602bfebf94bdbe26986a6f3718/odl/solvers/advanced/chambolle_pock.py#L43-L123)

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,87 @@
...
+    DPhiJacobian = gradS.domain.one()
+
+    grad = Gradient(gradS.domain, method='central')
+
+    # We solve poisson using the fourier transform
+    # ft = FourierTransform(op.domain)
```
adler-j: remove

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,87 @@
...
+    k2_values = sum((ft.range.points() ** 2).T)
+    k2 = ft.range.element(np.maximum(np.abs(k2_values), 0.01))
+    poisson_solver = ft.inverse * (1 / k2) * ft
+
+    for _ in range(niter):
+        PhiStarX = DPhiJacobian * I
+
+        grads = gradS(PhiStarX)
+
+        tmp = grad(grads)
```
adler-j: This is much more efficient to write:

```python
tmp = grad(grads)
for i in range(tmp.size):
    tmp[i] *= PhiStarX
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    noise : `array-like`
+        Noise.
+    impl : {'general', 'dB'}
+        Implementation method.
+        'general' means SNR = s_power / n_power,
+        'dB' means SNR = 10 * log10 (s_power / n_power).
+
+    Returns
+    -------
+    The value of signal-to-noise ratio.
```
adler-j:
```python
snr : `float`
    value of signal-to-noise ratio. 
    If the power of noise is zero, then the return is 'inf',
    otherwise, the computed value.
```

odl/util/utility.py
```diff
@@ -417,6 +417,39 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+
+    Parameters
+    ----------
+    signal : `array-like`
+        Noiseless data.
+    noise : `array-like`
+        Noise.
+    impl : {'general', 'dB'}
+        Implementation method.
+        'general' means SNR = s_power / n_power,
```
adler-j: s_power and n_power have not been defined so far, you need to say what they mean.

chongchenmath: These two guys are not input or output arguments. Do we need to do that?

odl/util/utility.py
chongchenmath: Done.

odl/util/utility.py
adler-j: Well if a user reads this documentation, they will not know what `s_power` means, so the statement

```
'general' means SNR = s_power / n_power
```

is nonsense. You need to be explicit in what you mean, for example by writing it out explicitly:

```
'general' means SNR = variance(signal) / variance(noise),
```

odl/util/utility.py
```diff
@@ -433,9 +433,10 @@ def snr(signal, noise, impl):
 
     Returns
     -------
-    The value of signal-to-noise ratio.
-        If the power of noise is zero, then the return is 'inf',
-        otherwise, the return is the computed value.
+    snr : `float`
+    Value of signal-to-noise ratio.
```
adler-j: This should be indented

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # ft = FourierTransform(op.domain)
+    ft = FourierTransform(gradS.domain)
+    k2_values = sum((ft.range.points() ** 2).T)
+    k2 = ft.range.element(np.maximum(np.abs(k2_values), 0.01))
+    poisson_solver = ft.inverse * (1 / k2) * ft
+
+    for _ in range(niter):
+
+#        PhiStarX = DPhiJacobian * I
```
adler-j: Lots of leftover code here

odl/deform/optimal_information_transport.py
chongchenmath: Done.

odl/deform/oit_reconstruction_in_operator.py
```diff
@@ -0,0 +1,762 @@
+# Copyright 2014-2016 The ODL development group
```
adler-j: I'm not quite sure what this file is meant to do, but it feels severely malplaced. The `Functional` stuff should not be here (it has its own branch) and all the manual tests should be moved somewhere else. Also, the dependence on `ddmatch` should not be in ODL.

examples/optimal_information_transport_test.py
```diff
@@ -0,0 +1,117 @@
...
+from odl.deform import optimal_information_transport_solver
+import odl
+standard_library.install_aliases()
+
+
+I0name = './pictures/c_highres.png'
```
adler-j: We shouldnt need to add pictures for simple things as this, there has to be some other phantom that we can use.

examples/oit_reconstruction_new.py
```diff
@@ -0,0 +1,981 @@
+# Copyright 2014-2016 The ODL development group
```
adler-j: This file is way to long for an example, examples need to be at most 100 lines or so.

If we want to have examples this complicated, they should be moved to some external place (some `odl-examples` repo, for example).

odl/util/utility.py
chongchenmath: done.

odl/util/utility.py
chongchenmath: Indented.

odl/deform/optimal_information_transport.py
chongchenmath: Updated. Two different methods are implemented here.

```
        # implementation for mass-preserving case
        PhiStarX = DPhiJacobian * I
        grads = gradS(PhiStarX)
        tmp = grad(grads)
        for i in range(tmp.size):
            tmp[i] *= PhiStarX
        # tmp = tmp.space.element([tp * PhiStarX for tp in tmp])
        u = sigma * grad(1 - np.sqrt(DPhiJacobian)) - 2 * tmp

#        # implementation for non-mass-preserving case
#        PhiStarX = I
#        grads = gradS(PhiStarX)
#        tmp = grad(PhiStarX)
#        for i in range(tmp.size):
#            tmp[i] *= grads
#        # tmp = tmp.space.element([tp * grads for tp in tmp])
#        u = sigma * grad(1 - np.sqrt(DPhiJacobian)) + 2 * tmp
```

examples/optimal_information_transport_test.py
ozanoktem: We already have the submarine phantom in the utilities part. Why not using that one? If we want to add further phantoms, like the 2D letters C, I, V, ... and the 2D hand and brain images, that is where they should be added.

adler-j: We should be rather careful in adding data to the core package since it makes ODL larger and harder to install.

But as you say. *if* we add more of these, they should be properly added to `odl.util` so that they are accessible by everyone.

examples/oit_reconstruction_new.py
chongchenmath: This is an unfinished work.

ozanoktem: If it is unfinished, then it should not be part of a pull request.

examples/optimal_information_transport_test.py
chongchenmath: Maybe we need to find a place for data. 

The  submarine phantom is added.
```
# Create the ground truth as the submarine phantom
ground_truth = odl.util.submarine_phantom(space, smooth=True, taper=50.0)
       ......
# Create the template as the disc phantom
template = odl.util.disc_phantom(space, smooth=True, taper=50.0)
```

ozanoktem: Data should really **not** be part of ODL. In examples we create data by simulations. Also, as @adler-j wrote, we should be conservative in what phantoms to add.

examples/oit_reconstruction_new.py
adler-j: To elaborate further on what @ozanoktem said:

The content of a pull request should be things that you expect others to read and give feedback on. Either fully finished code that you feel is ready for a review and merge, or partially finished code that you want someone else to have a look at and provide input on. For example, you may want input on if a design is good before you make further work on it.

Another guideline is that all code in a pull request should be runnable *as is* so that others can try your code on their own system and verify that it works. Thus you should not have any dependencies to external code that is not listed in the official dependency list (basically `numpy`, `scipy` and `matplotlib`).

chongchenmath: Great.

odl/deform/optimal_information_transport.py
```diff
@@ -0,0 +1,121 @@
...
+from odl.trafos import FourierTransform
+import odl
+standard_library.install_aliases()
+
+__all__ = ('optimal_information_transport_solver',)
+
+
+def optimal_information_transport_solver(gradS, I, niter, eps,
+                                         sigma, callback=None):
+    """
```
adler-j: Please consult the [How to document](http://odl.readthedocs.io/dev/document.html) guide for information on how to write function docstrings.

odl/util/phantom.py
```diff
@@ -540,6 +540,87 @@ def rect(x):
     return out.ufunc.minimum(1, out=out)
 
 
+def disc_phantom(discr, smooth=True, taper=20.0):
+    """Return a 'disc' phantom consisting in an disc.
+    This phantom is used in [Okt2015]_ for shape-based reconstruction.
```
adler-j: Missing newline

odl/util/phantom.py
```diff
@@ -540,6 +540,87 @@ def rect(x):
...
+    ----------
+    discr : `DiscreteLp`
+        Discretized space in which the phantom is supposed to be created
+    smooth : `bool`, optional
+        If `True`, the boundaries are smoothed out. Otherwise, the
+        function steps from 0 to 1 at the boundaries.
+    taper : `float`, optional
+        Tapering parameter for the boundary smoothing. Larger values
+        mean faster taper, i.e. sharper boundaries.
+    Returns
```
adler-j: Missing newline

odl/util/phantom.py
```diff
@@ -540,6 +540,87 @@ def rect(x):
...
+    Parameters
+    ----------
+    discr : `DiscreteLp`
+        Discretized space in which the phantom is supposed to be created
+    smooth : `bool`, optional
+        If `True`, the boundaries are smoothed out. Otherwise, the
+        function steps from 0 to 1 at the boundaries.
+    taper : `float`, optional
+        Tapering parameter for the boundary smoothing. Larger values
+        mean faster taper, i.e. sharper boundaries.
```
adler-j: A comment on that this only works if `smooth=True`

odl/util/phantom.py
kohr-h: Additional remark: given that the paper never got published (yet), we should remove the citation, also for the submarine.

odl/deform/optimal_information_transport.py
chongchenmath: Done.

odl/util/phantom.py
chongchenmath: Modified the comments.

odl/util/phantom.py
chongchenmath: changed.

odl/util/phantom.py
chongchenmath: changed.

examples/optimal_information_transport_test.py
chongchenmath: Done.

adler-j:
This is still waiting on issue #505, mass preserving deformations before it can go on.

adler-j:
Rebased master on top of this, but am now failing with rebasing these 28 commits into like, 1 commit. Any input on how to do that?

Anyway this could use some more attention soon, after #505.

kohr-h:
> Rebased master on top of this, but am now failing with rebasing these 28 commits into like, 1 commit. Any input on how to do that?

Like, you rebased this branch on current master and try to do `rebase -i`? Shouldn't be a problem at all. What exactly is failing?

adler-j:
> Like, you rebased this branch on current master and try to do rebase -i? Shouldn't be a problem at all. What exactly is failing?

I get the commits intermixed with all other commits on master, and would have to sit and cherry pick with around 120 commits.

kohr-h:
Hm, then if you haven't pushed yet you could make a copy of the remote branch and squash the commits first, then rebase on master. Otherwise, `git reflog` is your friend to undo things :-)

adler-j:
Really dont feel like undoing this merge, it took like 30 minutes :angry:, is there no reasonable way to squash the commits in this branch?

kohr-h:
Well, you could also try reordering the commits you want to squash as one big block in `rebase -i` and `fixup` all but one, which you `reword`. Not sure if that doesn't break in different ways, though.
On the other hand, I'm quite sure that just trying to squash first on a copy of the remote branch and then rebasing on master would be much simpler. Something like
```
git branch oit_rebase origin/issue-451__optimal_mass_transport
git checkout oit_rebase
git rebase -i <commit>
# squash squash
git rebase origin/master
# no conflicts?
```

adler-j:
Too the best of my knowledge, this branch is heavily out of sync with master and needs some serious improvements. I would probably venture to say it should be scrapped and rewritten using the new tools we have (deformation operators). Opinions?

chongchenmath:
Totally agree with you @adler-j. To convenient test, I have included its newest version into deformation operators branch.

kohr-h:
No opinion. Do whatever is most useful.

adler-j:
No, this does not get merged into master. It hasn't been review and the tests are failing.

PR #462, opened 2016-06-17
Issue 460  param helper

kohr-h:
Rather simple addition. I hope I got all corner cases right. ~~I'll also test~~ I also tested the version which does not go via `np.broadcast_to`. 
One problem with that fallback implementation is that parameter types which are themselves iterable (e.g. strings) can be misinterpreted if the length happens to be right, e.g. `'100'` will be converted to `['1', '0', '0']` for `length=3`.


odl/util/normalize.py
```diff
@@ -0,0 +1,193 @@
...
+        Target shape for error checking of out-of-bounds indices.
+        Also needed to determine the number of axes.
+    int_to_slice : `bool`, optional
+        If `True`, turn integers into corresponding slice objects.
+
+    Returns
+    -------
+    normalized : `tuple` of `slice`
+        Normalized index expression
+    """
```
adler-j: Could use an example here

odl/util/normalize.py
```diff
@@ -0,0 +1,193 @@
...
+    param_conv : callable, optional
+        Conversion applied to each list element. None means no conversion.
+    keep_none : bool, optional
+        If True, None is not converted.
+
+    Returns
+    -------
+    plist : list
+        Input parameter turned into a list of length ``length``.
+    """
```
adler-j: An example would be great

odl/util/normalize.py
```diff
@@ -0,0 +1,193 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utilities for internal use."""
```
adler-j: This looks like copy paste

odl/util/normalize.py
```diff
@@ -0,0 +1,193 @@
...
+                         ''.format(len(out_list), length))
+
+    if param_conv is not None:
+        try:
+            for i, p in enumerate(out_list):
+                if p is None and keep_none:
+                    pass
+                else:
+                    out_list[i] = param_conv(p)
+        except TypeError as error:
```
adler-j: I would prefer simply letting this error trickle through. In python 2 raise_from kills the call stack making error searching hard.

adler-j:
Looks good, just make sure to rebase before merge.

kohr-h:
> One problem with that fallback implementation is that parameter types which are themselves iterable (e.g. strings) can be misinterpreted if the length happens to be right, e.g. '100' will be converted to ['1', '0', '0'] for length=3.

Solved that one with checking `np.isscalar(param)` first.

odl/util/normalize.py
kohr-h: Done

odl/util/normalize.py
kohr-h: Done

odl/util/normalize.py
kohr-h: Okay, done.

kohr-h:
Done with all comments and rebased a bit. Will resolve conflicts with master and merge after tests are done.

kohr-h:
> Will resolve conflicts with master and merge after tests are done.

Waiting for the tests to run through, will merge after that.

PR #465, opened 2016-06-22
ENH: add deformation operators

adler-j:
Initial work on adding deformation operators to ODL, design probably not 100% yet. Documentation far from done. The design should however be "reviewable" for comments and ideas.

Note that this is a merge into the OIT branch.

### DIFF
--- a/odl/deform/linear_deformation.py
+++ b/odl/deform/linear_deformation.py
@@ -1,0 +1,192 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Shape-based reconstruction using optimal information transportation.
+
+The Fisher-Rao metric is used in regularization term. And L2 data matching
+term is used in fitting term.
+"""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+
+from odl.operator import Operator
+from odl.discr import DiscreteLp, Divergence
+from odl.space import ProductSpace
+
+__all__ = ('LinearDeformation', 'MassPreservingLinearDeformation')
+
+
+class LinearDeformation(Operator):
+    """Linearized deformation."""
+    def __init__(self, offsets=None):
+        """Initialize a linear deformation
+
+        This has the form
+
+        Id + offsets
+
+        Parameters
+        ----------
+        vector_field : `ProductSpace` of `DiscreteLp` spaces
+            The points to calculate the interpolation in.
+
+        Examples
+        --------
+        Identity deformation:
+
+        >>> import odl
+        >>> space = odl.uniform_discr(0, 1, 5)
+        >>> pspace = ProductSpace(space, space.ndim)
+        >>> vector_field = pspace.element(space.points().T)
+        >>> deformation = LinearDeformation(vector_field)
+        """
+
+        assert isinstance(offsets.space, ProductSpace)
+        assert offsets.space.is_power_space
+        assert isinstance(offsets.space[0], DiscreteLp)
+
+        ndim = offsets.space.size
+        space = offsets.space[0]
+
+        interp_points = space.points().T
+        for i in range(ndim):
+            interp_points[i] += offsets[i].ntuple.asarray()
+
+        self.interp_points = interp_points
+        self.offsets = offsets
+
+        Operator.__init__(self,
+                          domain=space,
+                          range=space,
+                          linear=True)
+
+    def _call(self, x):
+        return x.interpolation(self.interp_points, bounds_check=False)
+
+    @staticmethod
+    def identity(space):
+        """Create the identity transformation on a space."""
+        pspace = ProductSpace(space, space.ndim)
+        vector_field = pspace.zero()
+        deformation = LinearDeformation(vector_field)
+        return deformation
+
+
+class MassPreservingLinearDeformation(Operator):
+    def __init__(self, offsets, vector_field_jacobian, divergence=None):
+        """Initialize a linear deformation
+
+        Parameters
+        ----------
+        vector_field : `ProductSpace` of `DiscreteLp` spaces
+            The (small) deformation field.
+        vector_field_jacobian : `DiscreteLp`
+            The determinant of the jacobian of the deformation.
+        divergence : `Operator`
+            The divergence operator used in composition.
+            Default:
+            Divergence(range=vector_field_jacobian.space, method='central')
+
+        Examples
+        --------
+        Identity deformation:
+
+        >>> import odl
+        >>> space = odl.uniform_discr(0, 1, 5)
+        >>> pspace = ProductSpace(space, space.ndim)
+        >>> vector_field = pspace.element(space.points().T)
+        >>> jacobian = space.zero()
+        >>> deformation = MassPreservingLinearDeformation(vector_field,
+        ...                                               jacobian)
+        """
+
+        assert isinstance(offsets.space, ProductSpace)
+        assert offsets.space.is_power_space
+        assert isinstance(offsets.space[0], DiscreteLp)
+        assert vector_field_jacobian in offsets.space[0]
+
+        ndim = offsets.space.size
+        space = offsets.space[0]
+
+        interp_points = space.points().T
+        for i in range(ndim):
+            interp_points[i] += offsets[i].ntuple.asarray()
+
+        self.interp_points = interp_points
+        self.offsets = offsets
+        self.vector_field_jacobian = vector_field_jacobian
+
+        #offsets.show('offsets')
+
+        if divergence is None:
+            self.divergence = Divergence(range=space, method='central')
+
+        Operator.__init__(self,
+                          domain=space,
+                          range=space,
+                          linear=True)
+
+    def _call(self, x):
+        result = x.interpolation(self.interp_points, bounds_check=False)
+        result = self.range.element(result)
+        result *= self.vector_field_jacobian
+        return result
+
+    def compose(self, linear_deformation):
+        assert isinstance(linear_deformation, LinearDeformation)
+
+        # Find the deformations for the output field
+        out_offsets = self.offsets.space.element()
+
+        for i in range(self.interp_points.shape[0]):
+            as_element = self.domain.element(self.interp_points[i])
+            interpolated = as_element.interpolation(
+                linear_deformation.interp_points, bounds_check=False)
+            out_offsets[i][:] = interpolated - self.domain.points().T[i]
+
+        # Find the jacobian
+        out_jacobian = self.domain.element(
+            self.vector_field_jacobian.interpolation(
+                linear_deformation.interp_points, bounds_check=False))
+
+        # Weighting
+        div = self.divergence(linear_deformation.offsets)
+        out_jacobian *= np.exp(div)
+
+        # out_jacobian = self.domain.one()
+        return MassPreservingLinearDeformation(out_offsets, out_jacobian)
+
+    @staticmethod
+    def identity(space):
+        """Create the identity transformation on a space."""
+        pspace = ProductSpace(space, space.ndim)
+        vector_field = pspace.zero()
+        jacobian = space.one()
+        deformation = MassPreservingLinearDeformation(vector_field,
+                                                      jacobian)
+        return deformation
+
+if __name__ == '__main__':
+    # pylint: disable=wrong-import-position
+    from odl.util.testutils import run_doctests
+    run_doctests()
--- a/odl/deform/linear_deformation.py
+++ b/odl/deform/linear_deformation.py
@@ -31,9 +31,192 @@ import numpy as np
 
 from odl.operator import Operator
 from odl.discr import DiscreteLp, Divergence
-from odl.space import ProductSpace
+import odl
 
 __all__ = ('LinearDeformation', 'MassPreservingLinearDeformation')
+
+
+class LinearizedDeformationOperator(Operator):
+    """
+    Linearized deformation operator mapping parameters into a
+    deformed template. It is a left action given by pushforward.
+
+    This operator computes the deformed template for a fixed
+    template ``I``:
+
+        v --> I(Id + v)
+
+    where ``Id`` is the identity mapping:
+
+        y --> y
+
+    and the vector field ``v`` is a displacement mapping:
+
+        y --> v(y)
+
+    Here, the ``y`` is an element in the domain of the target.
+    In diffeomorphic sense,  the domain of ``I`` is the same
+    as the target.
+    """
+
+    def __init__(self, template):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        template : `DiscreteLpVector`
+            Fixed template deformed by the vector field.
+        """
+        self.template = template
+        self.domain_space = odl.ProductSpace(self.template.space,
+                                             self.template.space.ndim)
+
+        super().__init__(self.domain_space, self.template.space, linear=False)
+
+    def _call(self, displacement):
+        """Implementation of ``self(displacement)``.
+
+        Parameters
+        ----------
+        displacement : `ProductSpaceVector`
+            Linearized deformation parameters for image grid points.
+        """
+        image_pts = self.template.space.grid.points()
+        image_pts += np.asarray(displacement).T
+
+        return self.template.interpolation(image_pts.T, bounds_check=False)
+
+    def linear_deform(self, template, displacement):
+        """Implementation of ``self(template, displacement)``.
+
+        This function computes the deformed template for a given
+        template ``I`` and displacement ``v``:
+
+            (I, v) --> I(Id + v)
+
+        Parameters
+        ----------
+        template : `DiscreteLpVector`
+            Fixed template deformed by the vector field.
+        displacement : `ProductSpaceVector`
+            Linearized deformation parameters for image grid points.
+        """
+        image_pts = template.space.grid.points()
+        image_pts += np.asarray(displacement).T
+
+        return template.interpolation(image_pts.T, bounds_check=False)
+
+
+class MassPreservingDeformationOperator(Operator):
+    """
+    Mass-preserving deformation operator mapping parameters to a
+    deformed template. It is a left action given by pushforward.
+
+    This operator computes the deformed template
+    for a fixed template ``I``:
+
+        (jacdetinvdef, invdef) --> jacdetinvdef * I(invdef)
+
+    where ``invdef`` is the inverse of deformation mapping:
+
+        y --> invdef(y)
+
+    and ``jacdetinvdef`` is the Jacobian determinant of ``invdef``:
+
+        y --> jacdetinvdef(y)
+
+    Here, ``y`` is an element in the domain of target.
+    In diffeomorphic sense,  the domain of ``I`` is the same
+    as the target.
+    """
+
+    def __init__(self, template):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        template : `DiscreteLpVector`
+            Fixed template deformed by the vector field.
+        """
+        self.template = template
+        self.domain_space = odl.ProductSpace(self.template.space,
+                                             self.template.space.ndim + 1)
+        self.invdef_pts = np.empty([self.template.space.ndim,
+                                    self.template.size])
+
+        super().__init__(self.domain_space, self.template.space, linear=False)
+
+    def _call(self, jacdetinvdef_invdef):
+        """Implementation of ``self(jacdetinvdef_invdef)``.
+
+        Parameters
+        ----------
+        jacdetinvdef_invdef : 'ProductSpaceVector'
+
+                R^{n} --> R^{1} times R^{n}
+
+            The ``jacdetinvdef_invdef`` is defined on a product
+            space ``R^{1} times R^{n}`` for ``jacdetinvdef`` and
+            ``invdef``, respectively, where
+
+            jacdetinvdef : 'DiscreteLpVector'
+
+                    R^{n} --> R^{1}
+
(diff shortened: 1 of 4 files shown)


odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Shape-based reconstruction using optimal information transportation.
```
adler-j: Old comment

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+from odl.operator import Operator
+from odl.discr import DiscreteLp, Divergence
+from odl.space import ProductSpace
+
+__all__ = ('LinearDeformation', 'MassPreservingLinearDeformation')
+
+
+class LinearDeformation(Operator):
+    """Linearized deformation."""
```
adler-j: Needs spacing and more words

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+
+from odl.operator import Operator
+from odl.discr import DiscreteLp, Divergence
+from odl.space import ProductSpace
+
+__all__ = ('LinearDeformation', 'MassPreservingLinearDeformation')
```
adler-j: Should I try to merge these two into the same kind of operator?

kohr-h: Depends. Is it going to be only that one flag? Who knows? 

For now it is, though, so we can decide to go for it now and change it later once we see the need for more versions.

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class LinearDeformation(Operator):
```
kohr-h: Don't agree with the name "Linear". It isn't.

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+    """Linearized deformation."""
+    def __init__(self, offsets=None):
```
kohr-h: It is not at all clear to me that this is what we always want to do. In my view, we actually have 3 versions, namely

- Fixed deformation, variable template (this version)
- Fixed template, variable deformation
- Both variable

The middle one seems to be at least as important in applications as the first.

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+        """Initialize a linear deformation
+
+        This has the form
+
+        Id + offsets
```
kohr-h: `displacement`?

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+            out_offsets[i][:] = interpolated - self.domain.points().T[i]
+
+        # Find the jacobian
+        out_jacobian = self.domain.element(
+            self.vector_field_jacobian.interpolation(
+                linear_deformation.interp_points, bounds_check=False))
+
+        # Weighting
+        div = self.divergence(linear_deformation.offsets)
+        out_jacobian *= np.exp(div)
```
kohr-h: Is this always true? I guess it's an asymptotic formula for small `linear_deformation.offsets`.

adler-j: Its an asymptotic formula I'm quite sure.

kohr-h: We should be very clear about that then. May break things if we have large deformations.

odl/deform/linear_deformation.py
adler-j: Well the middle one requires some knowledge of a reasonable "basis" for the deformation, i.e. some vector space structure or similar. RKHS would be fine, but that should (i.m.o.) be added in another branch by someone more competent in that than I am.

odl/deform/linear_deformation.py
adler-j: Any suggestion on naming? Found this one freaking hard.

Perhaps @chongchenmath or @ozanoktem have some suggestions from the math literature? What would this be called?

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+        self.interp_points = interp_points
+        self.offsets = offsets
+
+        Operator.__init__(self,
+                          domain=space,
+                          range=space,
+                          linear=True)
+
+    def _call(self, x):
+        return x.interpolation(self.interp_points, bounds_check=False)
```
adler-j: Perhaps allow using other interpolations than the space interpolation here. I'm not 100% sure.

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @staticmethod
+    def identity(space):
```
adler-j: Found this method quite useful. Do we all agree?

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,192 @@
...
+        Identity deformation:
+
+        >>> import odl
+        >>> space = odl.uniform_discr(0, 1, 5)
+        >>> pspace = ProductSpace(space, space.ndim)
+        >>> vector_field = pspace.element(space.points().T)
+        >>> deformation = LinearDeformation(vector_field)
+        """
+
+        assert isinstance(offsets.space, ProductSpace)
```
adler-j: Obviously needs to be freshened up

odl/deform/linear_deformation.py
adler-j: Well this all breaks with large deformations, the divergence can go to infinity (this happens if your step is to large in the OIT code). Sadly the theory is all about "epsilon sized deformations".

odl/deform/linear_deformation.py
kohr-h: A candidate for common code would be
```python
class LinearizedDeformationOperatorBase:
    def deform(self, template, deformation):
        # deform template with deformation (given in a certain way)
```

What else could there be? Derivatives and adjoints require too much internal knowledge of the concrete operator to be abstracted out completely, as far as I see it. However, the product space guy could easily use the derivatives and their adjoints from the "partial" operators.

odl/deform/linear_deformation.py
kohr-h: Not easy, admittedly. We had `LinearizedDeformation` before, but how do we distinguish the 3 versions? Suggestions:

- `LinearizedFixedTemplateDeformation`
- `LinearizedFixedDiffeoDeformation` (ok, it's not always a diffeo but what the heck)
- `LinearizedDeformation`

odl/deform/linear_deformation.py
kohr-h: For me the biggest TODO here is to enforce proper boundary behavior, but that actually has to go into `interpolation`. Other than that, I'd always use the space one. If higher order is required, the space needs to have it IMO.

odl/deform/linear_deformation.py
adler-j: Im not even sure that should be a class, perhaps it could be a free method. But I do agree, it would be nice.

My suggestion here is that we write this class with a nice interface and an acceptable implementation and merge this.

Then once we add more deformations, we have more hands on information on what we'll need so we can start factoring out things like that.

odl/deform/linear_deformation.py
adler-j: Abit too long all of them except the last I'd say. Regarding `LinearizedFixedDiffeoDeformation`, if we made such a thing, perhaps we could add a `is_diffeo` method that somehow verifies that it actually is a diffeomorphism somehow.

odl/deform/linear_deformation.py
adler-j: Well if we go by that approach, this implementation is about done (once we get boundary behaviour in we can drop the `bounds_check=False`).

odl/deform/linear_deformation.py
kohr-h: Not sure it belongs here. We are getting a bit into the "implement deformations as group elements" line. Adding this method to the deformation operator class feels a lot like mixing group action (the operator) with group element (the deformation itself) things. I think we need a clear separation.

odl/deform/linear_deformation.py
kohr-h: > Well the middle one requires some knowledge of a reasonable "basis" for the deformation, i.e. some vector space structure or similar. RKHS would be fine, but that should (i.m.o.) be added in another branch by someone more competent in that than I am.

I think the middle one can be implemented without any knowledge of RKHS whatsoever. Just take a vector field (could lie in an RKHS) and apply it to the given template via the action of linearized deformation.

Agree in principle, but since we know that we are going to need at least the first two cases (actually so far we *always* considered the middle one), we should create structures for them already now. Going forward with an arbitrary choice and breaking it later seems a bit pointless, we already have that bit of information.

> My suggestion here is that we write this class with a nice interface and an acceptable implementation and merge this.
> 
> Then once we add more deformations, we have more hands on information on what we'll need so we can start factoring out things like that.

odl/deform/linear_deformation.py
kohr-h: That "somehow" would be interesting to learn more about :-)
No, honestly, for a fixed deformation, one could actually have something diagnostics-like that does a standardized test (if there is such a thing for diffeos). For variable deformation, it's harder.
On the other hand, I guess once we don't have a diffeo any longer, things just break down in the worst case. We can decide to just happily go on until the code crashes or produces NaNs.

Regarding the name, I didn't want to write `LinearizedFixedDeformationDeformation`, `Diffeo` was just nicer and shorter. Shouldn't imply anything.

As far as the total length is concerned, I'm not so worried. We probably have longer stuff, and this is still not quite STIR level. We could try to crank down on the `Linearized` part, though. Alternatives?

odl/deform/linear_deformation.py
chongchenmath: > >>
The Fixed deformation, variable template (this version)
Fixed template, variable deformation
Both variable
>>>

Since the "fixed template, variable deformation"  is much more useful, it is better to implement it in _call(). We may implement the "variable template, variable deformation"  as a function *_deform() in the same class. This case contained all the three situation. It is flexible. Please check the preliminary implementation in the latest pushed file:
[https://github.com/odlgroup/odl/blob/issue-464__deformation_operators/odl/deform/linear_deformation.py](url)

adler-j:
With the newest changes from @chongchenmath, I feel like there is some redundancy here. We need to sort that out shortly.

kohr-h:
So @ozanoktem and I had some discussion about this today, and @chongchenmath's additions go into the same direction. Therefore I think we should reach a design decision here. Let me cast some light on the issues that we are facing.

The mathematical standpoint
=======================
Scenario
-----------
For any deformation operator we will implement (read: that I can think of now), we are going to have 3 versions, possibly but not likely more. This is due to the fact that we have 2 "ingredients" to a deformation, namely the object defining the deformation itself (called "deformation" in the following) and the object to deform (the "template"), and hence there are 3 ways to combine these.

Let's say we have a set/space/whatever `V` defining deformations and a space `X` of deformable templates. Then, abstractly, we can write the action of the deformation operator as `d.f`, where `d` is a deformation and `f` a template.

Now the corresponding deformation *operator* can be defined in the following ways:

- **Fixed template, variable deformation.** For now, this is easily the most relevant use case. 
  ```
  W_f: V --> X,  W(d) := d.f
  ```

- **Fixed deformation, variable template.** Typically the simplest case, but less relevant.
  ```
  W_d: X --> X,  W(f) := d.f
  ```

- **Both template, and deformation variable.** Important in applications.
  ```
  W: V x X --> X,  W((d, f)) := d.f
  ```

Properties and connections
----------------------------------
The operators are, of course, not completely independent, but have a couple of connections. The most obvious one is that they all are evaluated in exactly the same way, once both template and deformation are known.
Another connection is through derivatives and adjoints:
```
W'((d, f))((e, g)) = W_f'(d)(e) + W_d'(f)(g)
W'((d, f))*(h) = ( W_f'(d)*(h), W_d'(f)*(h) )
```
Here, `f, g, h in X` and `d, e in V`.

So, for what currently matters, the third guy can easily be defined through the other two, and evaluation, derivative and adjoint follow nicely.

The obvious question is: *Why don't we simply implement it this way?* (ping @ozanoktem)

The software architectural standpoint
=============================
Interface of `Operator`
-----------------------------
We have designed the `Operator` class as *the* single internal interface that enables us to write e.g. a completely generic conjugate gradient method. The only things that the method needs to know about the operator are defined in its *public interface*, which is 

- `domain` and `range`,
- a method to call it, realized with the syntax `operator(argument)`,
- a method to get its `derivative` at a `point`, returning a new operator (*optional*), and
- the `adjoint` property, returning also a new operator (*optional*).

That's it. There is no flag `is_differentiable` -- if it isn't, don't implement `derivative`. There is no `norm` property -- how should one know that for a given operator? Estimate? Return nonsense for "don't know"? `NotImplementedError`? How about non-linear operators?
Likewise, there is nothing like "if defined on a product space, differentiate only in the first argument". Simply because this would put a huge burden on people who write operators, and for technical reasons. For example, why not "derivative in arguments 1 to 3" out of 4 possible?

So its universality is the only (and sufficient) reason why we really want to stick with this interface and rather shape new types of operators around it. All code in ODL that uses an operator is written in the assumption that the above named four things are available.

The problem with adding random methods
-------------------------------------------------------
And that's why adding random *public* functionality to operators is bad. Since only a small subset of operators will have that functionality, it is *worthless for any code that expects a generic operator*. 

For example, consider `*_deform()` methods for the generic deformation with template and deformation given, as proposed by @chongchenmath. As a method for internal use, it's perfectly fine, but as public method for evaluation, it's worthless. Just consider the case when you compose another operator with your new operator introducing, say, the additional `linear_deform` method. What happens to that method? Gone! Buried in the depths of the new `OperatorComp` instance which, of course, has never heard about anything like `linear_deform`. No code that uses this composition will ever be able to reliably call the `linear_deform` method.

However, there are good use cases for additional public methods. But these are almost exclusively *user-facing*, not extending or replacing an interface. For example, in the `FourierTransform`, we have a method [`init_fftw_plan`](https://github.com/odlgroup/odl/blob/611c4d6ecfd6ae065d8f691f650c871aa6bc1bdf/odl/trafos/fourier.py#L838-L881) that offers a user the possibility to manually initialize some stuff in the transform. That's okay since it's meant entirely for the user and not for generic code that only expects an `Operator`.

The problem with parallel interfaces
-------------------------------------------------
We face yet another set of problems when we write new classes that will have a new interface and live parallel to the existing `Operator` ecosystem. So one may think, why not doing the following:
```python
class DeformationOperator(Operator):
    def __init__(self, deform_space, templ_space, linear=False):
        domain = ProductSpace(deform_space, templ_space)
        super().__init__(domain=domain, range=templ_space, linear=linear)

    def _call(self, x):
        deform, templ = x
        # Evaluate deformation on template
        # Return deformed template

    def derivative_deform(self, x):
        deform, templ = x
        # Return operator V --> X that is the derivative in
        # x with respect to the deformation only

    def derivative_templ(self, x):
        deform, templ = x
        # Return operator X --> X that is the derivative in
        # x with respect to the template only

    # Why stop there? We want partial evaluations, too!
    def eval_deform(self, deform):
        # Return operator that has fixed deformation and
        # maps a template to the deformed template.
        # BUT how do we tell it to use the evaluation method
        # from above? Wasn't that the point of this exercise,
        # to keep things nicely together?

    def eval_templ(self, templ):
        # ...
```
Besides having defined a much more complex interface than the standard `Operator` one, we are doomed to make this monster play nicely together with the existing infrastructure. That means **a lot** of new code. Just think of the composition again. Compose this guy with another `ProductSpaceOperator` that maps `V x X` to itself. We better implement this case or say goodbye to all those new methods.
Then there is this hard-coded size 2 of the product space. We better be sure that we *never ever* are going to need a third argument. If we do one day, we're screwed. If there's no way to guarantee that, on the other hand, we are forced to implement it for *arbitrary size of the domain*. I don't even want to think about that.

Conclusions
=========
1. To me it's totally clear that each of the 3 possibilities for the operator domain are going to be 3 different operators. For sure not several evaluation methods in the same class.
2. *Maybe* certain code parts can be factored out, for example the evaluation method could be implemented in a "mixin" class that all subclasses can use.
3. One could go as far as using derivative of the two "one-variable" guys in the product space version, by (lazy version)
   ```python
   class MyDeformation(Operator):
       def __init__(self, def_space, templ_space):
           domain = ProductSpace(deform_space, templ_space)
           super().__init__(domain=domain, range=templ_space, linear=False)

       def derivative(self, x):
           deform, templ = x
           op_d = MyFixedTemplateDeformation(self.domain[0], template=templ)
           op_d_deriv = op_d.derivative(deform)
           op_t = MyFixedDiffeoDeformation(self.domain[1], diffeo=deform)
           op_t_deriv = op_t.derivative(templ)
           return ReductionOperator(op_d_deriv, op_t_deriv)
   ```
4. The "extended interface" idea doesn't fly IMO.
5. This post has become too long. Sorry for that.

chongchenmath:
@kohr-h gave a detailed analysis for the deformation development. 

A public interface of deformation operator is needed. The concrete deformation form is indicated by user with ``impl``.  The judgment statement needn't to do in each iteration. 

The *_deform() is useful and flexible to do arbitrary deformation. This is at least helpful to implement derivative and adjoint operator. Since we can give the arbitrary template as the fixed self.template.
......

adler-j:
> To me it's totally clear that each of the 3 possibilities for the operator domain are going to be 3 different operators. For sure not several evaluation methods in the same class.

Agree. If they share code, that can be kept in a separate free function. We'll see.

> One could go as far as using derivative of the two "one-variable" guys in the product space version, by (lazy version)

I dont see why not unless it introduces some small errors (different interpolations?).

> The "extended interface" idea doesn't fly IMO.

Agree. Anything that is supposed to go "into the ecosystem" should follow the basic interface. Anything beyond that is fine to add, but users should be aware that they are very specific and cant be expected to exist in general.

> This post has become too long. Sorry for that.

I found it a very nice read, perhaps even movable to some doc later on.

kohr-h:
> A public interface of deformation operator is needed. 

Yes. My suggestion was that it's the same as for any other `Operator` when it comes to evaluation, derivative and adjoint.

> The concrete deformation form is indicated by user with impl. 

Well, `impl` is more like "computational back-end", this one would rather be something like `variant`. Anyway, do I understand you correctly that you want something like this:
```python
class DeformationOperator(Operator):
    def __init__(self, deform_space, templ_space, variant):
        self.variant = variant
        if self.variant == 'fixed_template':
            domain = deform_space
            linear = False
        elif self.variant == 'fixed_deformation':
            domain = templ_space
            linear = True
        else:
            domain = ProductSpace(deform_space, templ_space)
            linear = False

        super().__init__(domain=domain, range=templ_space, linear=linear)

    def _call(self, x):
        if self.variant == 'fixed_template':
            return self._call_fixed_template(x)
        elif self.variant == 'fixed_deformation':
            return self._call_fixed_deformation(x)
        else:
            return self._call_both_variable(x)
```
While not being totally unfeasible, this approach is not desirable, for two main reasons. 

1. These `if...else` constructions are going to be *all over the class*, and to me, that is a strong indication that they should be separate classes since they share less code than they have differences.
2. This construction is quite bug-prone. Since the `domain` of the operator depends on the choice of `variant`, all code in the class has to get this mapping `variant -> domain` right, which means more opportunities to make mistakes.

> The *_deform() is useful and flexible to do arbitrary deformation. This is at least helpful to implement derivative and adjoint operator. Since we can give the arbitrary template as the fixed self.template.

I agree that such a method is **useful for implementation** -- it can be called by the `_call` method of all 3 operator variants. However, it can only serve that purpose, not as part of the public interface.

chongchenmath:
Another suggestion:

We may implement the fixed template operator ``W_I`` and deformation operator ``Def`` respectively.

We can consider ``W_I o Def``. This composition can be in different forms (e.g., linearized, mass-preserving, ......).

1. The fixed template operator ``W_I`` is a mapping: 

    W_I : deformation --> deformed template. 

Here the deformation is an arbitrary deformation we consider. This stuff is as I wrote in ``LinearizedDeformationOperator`` and ``MassPreservingDeformationOperator``. The derivatives and adjoints are implemented correspondingly. As mentioned before, we need to seek a desirable public interface.

2. The deformation operator is a pure mapping: 

    Def : deformation --> deformed deformation.

It seems like @adler-j wrote. But we needn't to consider different deformations. Here we just have the quite general deformation. This is useful for time-dependent reconstruction. Because we need to compose different time-step deformations (``phi_0``, ``phi_1``, ......, ``phi_{n-1}``) to obtain the deformed templates (``I_0``, ``I_1``, ......, ``I_{n-1}``.).

adler-j:
Well, if we get the ranges and domains appropriate then handling such a composition should be trivial @chongchenmath. The hard part would as I see it be to properly represent the space of deformations, we also need some way to interpret what a `deformed deformation` actually is.

kohr-h:
> The hard part would as I see it be to properly represent the space of deformations, we also need some way to interpret what a deformed deformation actually is.

Right. One additional slight difficulty here is that the deformations don't form a space at all, so we would need the notion of `Group`, which we currently haven't.

adler-j:
We really dont need to add that notion, we could just make it a `Set` with some group properties, no?

kohr-h:
Well, I guess the most interesting parts are composition of deformations (group "multiplication") and their action on functions. The latter is exactly what a deformation operator does, but the former part is not easily realizable with the "stupid" operator composition. How about putting some intelligence in `__mul__` and `__rmul__`?

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,377 @@
...
+    def _call(self, displacement):
+        """Implementation of ``self(displacement)``.
+
+        Parameters
+        ----------
+        displacement : `ProductSpaceVector`
+            Linearized deformation parameters for image grid points.
+        """
+        image_pts = self.template.space.grid.points()
+        image_pts += np.asarray(displacement).T
```
adler-j: This is extremely slow since it will access the displacement vector element by element.

odl/deform/linear_deformation.py
```diff
@@ -0,0 +1,377 @@
...
+        assert isinstance(linear_deformation, LinearDeformation)
+
+        # Find the deformations for the output field
+        out_offsets = self.offsets.space.element()
+
+        for i in range(self.interp_points.shape[0]):
+            as_element = self.domain.element(self.interp_points[i])
+            interpolated = as_element.interpolation(
+                linear_deformation.interp_points, bounds_check=False)
+            out_offsets[i][:] = interpolated - self.domain.points().T[i]
```
adler-j: precompute and store points

adler-j:
I guess this is closed in favor of #488.

Issue #463, opened 2016-06-21
Placement of phantoms in the library


adler-j:
Currently all phantoms are accessible under `odl.util`. As the number of phantom grows with relatively high speed right now I am considering if we should formalize this a bit more.

One option would be to put them under `odl.phantom`, that way we could also separate them a bit (like having CT phantoms SPECT phantoms in different files). It would also be easier to find what phantoms are available.

Another option is to start naming them something that starts with `phantom_`, this does however feel a bit arbitrary: `odl.util.phantom_shepp_logan` etc.

A final issue is that it seems we may want to add images to the library as phantoms. We need to discuss in what cases we should do this, and some limitations on the images

* Should we have images in at all?
* Shall we accept 3d volumes? 
* What format should the images be in? text? Png?

kohr-h:
I think `odl.phantom` or `odl.phantoms` would be the natural place for those guys.

Regarding images, I am very reluctant to include any at all. If an image is to be added, it better have a *real* extra value that none of the existing phantoms can provide. Regarding 3d volumes, even stronger objection. I don't want `git clone` to take minutes.

If we really want to have examples using digital 2d and 3d data (not programmatically created), we should create a separate repository `odl_test_data` or whatever, and refer to that in example code that potentially uses such data.

Issue #452, opened 2016-06-15
Different notions for domain corners


kohr-h:
We have a mismatch in names for the minimum and maximum corners of rectangular domains between partitions and discretizations, especially in `uniform_partition` and `uniform_discr`. One uses `begin, end` the other `min_corner, end_corner`. We should find a solution to harmonize these notions, and it should propagate all the way down to `IntervalProd`.

Things to harmonize:
- [x] `min_corner, max_corner`, `begin, end` --> `min_pt, max_pt`
- [x] `nsamples`, `num_nodes` --> `shape`
- [x] `center`, `midpoint` --> `mid_pt`

Edit 2: TODO list added
Edit 3: TODO list modified according to result of discussion & Edit 1 deleted.

kohr-h:
Since `uniform_discr` is probably the most heavily used function, we should consider using its notation. However, I think `nsamples` is not so great, `shape` is more "standard" at least for NumPy users.

adler-j:
I think `shape` is the obvious name instead of `nsamples` (especially since we expose it under the name `samples` already.)

Regarding `begin` vs `min_corner`, some inspiration is:

* [arange](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.arange.html): `start`/`stop`
* [linspace](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.linspace.html#numpy.linspace): `start`/`stop`
* [Dolfin, RectangleMesh](https://github.com/FEniCS/dolfin/blob/fcf70d934d63168e5ed037678e22ac66fb2b3474/dolfin/generation/RectangleMesh.h)     
```
    ///     p0 (_Point_)
    ///         First point.
    ///     p1 (_Point_)
    ///         Second point.
    ///     nx (double)
    ///         Number of cells in :math:`x`-direction.
    ///     ny (double)
    ///         Number of cells in :math:`y`-direction.
```

Couldnt really find more. My opinion is that `begin`/`end` is perhaps the easiest to write and quite simple to understand. `min_corner` is perhaps a bit more clean, but too verbose for my taste.

adler-j:
Another note: We have the exact same issue when comparing `center` with `midpoint`, that guy should also be standardized.

kohr-h:
I realized the same thing these days when skimming through the code of `grid.py`.

adler-j:
[...]

PR #467, opened 2016-06-23
Issue 385  subdiscr

### LINKED ISSUE #385: Utility to create subspaces of DiscreteLp
Not for the first time am I in the situation that a utility to create a subspace of a given (uniformly discretized) space would be very useful. And with subset I mean that the partitions should be subsets of each other, i.e. no resampling. For those cases we have the `Resampling` operator.

So basically the only thing needed is a way of handling a new function domain and a new shape, and both at the same time. Probably all the dirty work will be nicely handled `RectPartition`. A nice convenience option would be to combine a new shape together with an option regarding where in the large domain to place the small domain, something like

- `left` -- leftmost placement (same `begin` in old and new)
- `right` -- rightmost (same `end`)
- `center` -- centred (irrespective cell layout)
- `left_cell` -- leftmost but only counting complete cells
- `right_cell` -- obvioius
- `center_cell` -- centred, preferring the left (?) variant if ambiguous

This option could be given per axis.

kohr-h:
Closes #385 

Could perhaps use a test for the `nodes_on_bdry` helper. Okay with the decisions on the behavior in all the different cases?


odl/discr/lp_discr.py
```diff
@@ -158,6 +162,16 @@ def __init__(self, fspace, partition, dspace, exponent=2.0,
             self._complex_space = None
 
     @property
+    def min_corner(self):
```
adler-j: Lets keep this but we need to think about #452

odl/discr/lp_discr.py
```diff
@@ -316,6 +330,11 @@ def _astype(self, dtype):
                           exponent=self.exponent, interp=self.interp,
                           order=self.order)
 
+    def __copy__(self):
```
adler-j: Also add `copy` if you do this.

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1267,275 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+
+    Examples
+    --------
+    >>> discr = uniform_discr([0, 0], [1, 2], (10, 5))
+    >>> discr.cell_sides
+    array([ 0.1,  0.4])
+
+    No additional arguments -> copy:
+
+    >>> uniform_discr_fromdiscr(discr) == discr
```
adler-j:
```
>>> uniform_discr_fromdiscr(discr) is discr
???
```

odl/discr/lp_discr.py
kohr-h: Actually, this slipped in by accident since I don't longer need it. Do we want it?

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1267,275 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+
+    **3+ arguments:**
+        The underlying partition is uniquely determined by the new
+        parameters. See `uniform_partition`.
+
+    See also
+    --------
+    uniform_partition : underlying domain partitioning scheme
+
+    Examples
```
adler-j: First time i may say this, but I think this may be too much examples

odl/discr/lp_discr.py
adler-j: Honestly not unless we start adding it in general.

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1267,275 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+    >>> uniform_discr_fromdiscr(discr, min_corner=[1, 1],
+    ...                                max_corner=[3, 3]).cell_sides
+    >>> uniform_discr_fromdiscr(discr, min_corner=[1, 1],
+    ...                                nsamples=[5, 5])
+    uniform_discr([1.0, 1.0], [1.5, 3.0], [5, 5])
+    >>> uniform_discr_fromdiscr(discr, min_corner=[1, 1],
+    ...                                cell_sides=[1, 1])
+    uniform_discr([1.0, 1.0], [11.0, 6.0], [10, 5])
+
+    Specifying 3 or more new parameters will ignore ``discr``:
```
adler-j: I personally dislike ignored arguments. Imo this should be an error.

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1267,275 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+    >>> uniform_discr_fromdiscr(discr, min_corner=[None, 1],
+    ...                                max_corner=[3, None],
+                                       cell_sides=[None, 0.25])
+    uniform_discr([2.0, 1.0], [3.0, 2.25], [10, 5])
+    """
+    if not isinstance(discr, DiscreteLp):
+        raise TypeError('`discr` {!r} is not a DiscreteLp instance'
+                        ''.format(discr))
+    if not discr.partition.is_regular:
+        raise ValueError('`discr` {} is not uniformly discretized'
```
adler-j: This is not mentioned in the parameter list

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1267,275 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+    new_end = []
+    new_shape = []
+    new_csides = []
+    for i, (b, e, n, s, old_b, old_e, old_n, old_s) in enumerate(
+            zip(min_corner, max_corner, nsamples, cell_sides,
+                discr.min_corner, discr.max_corner, discr.shape,
+                discr.cell_sides)):
+        num_params = sum(p is not None for p in (b, e, n, s))
+
+        if num_params == 0:
```
adler-j: This implementation is way too verbose. There has to be a shorter implementation. Very easy to make a misstake here.

odl/util/normalize.py
```diff
@@ -262,6 +261,76 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
     return tuple(indices)
 
 
+def normalized_nodes_on_bdry(nodes_on_bdry, length):
```
adler-j: underscore?

odl/util/normalize.py
```diff
@@ -262,6 +261,76 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    Global per axis:
+
+    >>> normalized_nodes_on_bdry([True, False], length=2)
+    [(True, True), (False, False)]
+
+    Mixing global and explicit per axis:
+    >>> normalized_nodes_on_bdry([[True, False], False], length=2)
+    [(True, False), (False, False)]
+    """
+    shape = np.shape(nodes_on_bdry)
```
adler-j: What does `np.shape([[True, False], False])` even return?

odl/util/normalize.py
```diff
@@ -262,6 +261,76 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+                                 'expected (2,)'
+                                 .format(i, shape_i))
+    else:
+        raise ValueError('`nodes_on_bdry` has shape {}, expected ({},)'
+                         ''.format(shape, length))
+
+    return out_list
+
+
+def safe_int_conv(number):
```
adler-j: Is that really something we want here?

odl/discr/lp_discr.py
kohr-h: oops

odl/discr/lp_discr.py
kohr-h: But then only in the extreme case that this happens in all axes. If it's only in certain axes, I don't want to raise an error since consider it a valid use case.

odl/discr/lp_discr.py
kohr-h: `False`

odl/discr/lp_discr.py
kohr-h: Ha, so now I know where the border goes (roughly), can start bisecting :-)

Now that I see it again, I probably don't need to repeat all the cases between the example blocks. I'll reduce to one "simple" block and one block that does a bit of everything in different axes.

odl/discr/lp_discr.py
kohr-h: I'll check that. However, I explicitly flattened it at some point because I got confused with the cases.

odl/util/normalize.py
kohr-h: `(2,)`, since it would make an `object` array from it:
```python
>>> np.array([[True, False], True])
array([[True, False], True], dtype=object)
```

odl/util/normalize.py
kohr-h: I can underscore it. But you objected sometime when I turned an "int or sequence of int" input parameter blindly to an integer array with unsafe casting. So if I want some safe integer casting together with the `normalized_list` thing, you tell me the alternative.

odl/util/normalize.py
```diff
@@ -262,6 +261,76 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    >>> normalized_nodes_on_bdry([True, False], length=2)
+    [(True, True), (False, False)]
+
+    Mixing global and explicit per axis:
+    >>> normalized_nodes_on_bdry([[True, False], False], length=2)
+    [(True, False), (False, False)]
+    """
+    shape = np.shape(nodes_on_bdry)
+    if shape == ():
+        out_list = ([(bool(nodes_on_bdry), bool(nodes_on_bdry))] * length)
```
adler-j: unnecessary extra parenthesis

odl/util/normalize.py
```diff
@@ -262,6 +261,76 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    length : positive int
+        Desired length of the returned list.
+
+    Returns
+    -------
+    normalized : list of 2-tuples of bool
+        Normalized list with ``length`` entries, each of which is a
+        2-tuple of boolean values.
+
+    Examples
```
adler-j: No example with `length != 2`

odl/util/normalize.py
```diff
@@ -262,6 +261,76 @@ def normalized_index_expression(indices, shape, int_to_slice=False):
...
+    >>> normalized_nodes_on_bdry(True, length=2)
+    [(True, True), (True, True)]
+
+    Global per axis:
+
+    >>> normalized_nodes_on_bdry([True, False], length=2)
+    [(True, True), (False, False)]
+
+    Mixing global and explicit per axis:
+    >>> normalized_nodes_on_bdry([[True, False], False], length=2)
```
adler-j: Missing newline above this

odl/discr/lp_discr.py
kohr-h: Removed

odl/discr/lp_discr.py
kohr-h: Reduced the examples somewhat

odl/discr/lp_discr.py
kohr-h: So how about this? I think if you e.g. give 3 parameters in axis 0 and 2 in axis 1, it's valid. Do we want to raise if we don't use `discr` at all?

odl/discr/lp_discr.py
kohr-h: I have a hard time shortening this. Basically all cases are unique in some sense, so I don't see a way to apply some general rule. It is possible to shorten the code by collecting all parameters in a list and assign at the end, i.e.
```python
if num_params == 0:
    new_params = [old_b, old_e, old_n, old_s]
...
```
That would at least turn 4 lines into one in each case.

odl/util/normalize.py
kohr-h: It's in `odl.util` anyway, so it shouldn't be considered completely public. From my point of view, it's okay to have it non-underscored.

odl/util/normalize.py
kohr-h: What's our conclusion? Keep it with underscore or drop it and with it the safeguard against unsafe cast to integer?

odl/discr/lp_discr.py
kohr-h: I don't see a good way of testing if we are using information of `discr`. Perhaps we can just go with it.

odl/discr/lp_discr.py
kohr-h: It is

odl/util/normalize.py
kohr-h: Now there is.

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1262,250 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+        A sequence is interpreted as interpolation scheme per axis.
+
+            'nearest' : use nearest-neighbor interpolation
+
+            'linear' : use linear interpolation
+
+    impl : {'numpy', 'cuda'}, optional
+        Implementation of the data storage arrays
+    nodes_on_bdry : `bool` or `sequence`, optional
+        If a sequence is provided, it determines per axis whether to
```
adler-j: I feel like long docstrings like this should start with what it does, then go on with special cases. Here I have to read pretty much all of it before I learn if i should use this or not.

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1262,250 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+
+    Parameters
+    ----------
+    discr : `DiscreteLp`
+        Uniformly discretized space used as a template.
+    min_corner : `float` or `tuple` of `float`
+        Minimum corner of the result.
+    max_corner : `float` or `tuple` of `float`
+        Maximum corner of the result.
+    nsamples : `int` or `tuple` of `int`
```
adler-j: sequence of int

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1262,250 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+        Number of samples per axis. For dimension >= 2, a tuple is
```
adler-j: Second sentence really isnt needed

odl/discr/lp_discr.py
adler-j: Feels cleaner now!

odl/discr/lp_discr.py
adler-j: Id actually add this under this one

odl/discr/lp_discr.py
adler-j: Ok valid point here since you may use partial info in each of them. Lets go with this.

adler-j: Another comment though: Then the line `Specifying 3 or more new parameters will ignore ``discr``` is not correct.

odl/discr/lp_discr.py
adler-j: Well then we honestly may want to do a rename of the `is_regular` thing to `is_uniform`? For clarity.

odl/discr/lp_discr.py
```diff
@@ -1248,6 +1262,250 @@ def discr_sequence_space(shape, exponent=2.0, impl='numpy', **kwargs):
...
+    new_end = []
+    new_shape = []
+    new_csides = []
+    for i, (b, e, n, s, old_b, old_e, old_n, old_s) in enumerate(
+            zip(min_corner, max_corner, nsamples, cell_sides,
+                discr.min_corner, discr.max_corner, discr.shape,
+                discr.cell_sides)):
+        num_params = sum(p is not None for p in (b, e, n, s))
+
+        new_params = []
```
adler-j: This line should never be needed?

odl/util/normalize.py
adler-j: Ok fine. The more I think of this odl.util things, we really need to get #463 done so users dont need to use `odl.util` in any extent. That would keep them away from it.

odl/util/normalize.py
adler-j: Keep it in, I think I missunderstood where these were added. Thought it was in the uniform_discr area. The implementation should be fine.

adler-j:
Looking fine except some minor stuff.

odl/discr/lp_discr.py
kohr-h: I'll make a new PR into this branch so it shows up in the log. It's an API change.

kohr-h:
Okay, did some more cleanup. I'll wait for the tests and merge after it's done.

odl/discr/lp_discr.py
adler-j: Great.

PR #469, opened 2016-06-28
API: move phantoms to new odl.phantoms dir

adler-j:
Rather major API change, changes in lots of places, therefore updated ODL version to 0.2.4.


odl/phantom/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+from . import phantom_utils
+
+from .emission_phantoms import *
+__all__ += emission_phantoms.__all__
+
+from .geometric_phantoms import *
+__all__ += geometric_phantoms.__all__
+
+from .misc_phantoms import *
+__all__ += misc_phantoms.__all__
```
kohr-h: Don't think we need `_phantoms` for each module. It's pretty clear already from `odl.phantom.emission` what you get.

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+    Parameters
+    ----------
+    discr_space : `DiscretizedSpace`
+        Discretized space in which the phantom is supposed to be created.
+    begin : array-like or `float` in [0, 1]
+        The lower left corner of the cuboid within the space grid relative
+        to the extend of the grid.
+    end : array-like or `float` in [0, 1]
+        The upper right corner of the cuboid within the space grid relative
+        to the extend of the grid.
```
kohr-h: While we're at it: I don't quite like this handling of begin and end. It feels weird. IMO, we should take absolute coordinates like everywhere else.

adler-j:
We get naming conflicts in some cases (cant have `shepp_logan` inside a file called `shepp_logan`, or at least I got some errors trying it). But I'll have a look again.

odl/phantom/geometric_phantoms.py
adler-j: I agree, but that would need a notable rewrite of this and all uses. I guess it should be done now that we're at it.

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+     [1.0, 1.0, 1.0, 0.0, 0.0, 0.0],
+     [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]]
+    """
+    ndim = discr_space.ndim
+    shape = discr_space.shape
+
+    if np.isscalar(begin):
+        begin = (begin,) * ndim
+    if np.isscalar(end):
+        end = (end,) * ndim
```
kohr-h: This is in contrast to all the other places where similar input parameters are handled. We should expect sequences of correct size in nD and 1-sequence or float in 1D.

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+    # Create phantom
+    phan = np.zeros(shape)
+
+    slice1 = [slice(None)] * ndim
+
+    for nn in range(ndim):
+        start = np.floor(begin[nn] * shape[nn]).astype(int)
+        stop = np.ceil(end[nn] * shape[nn]).astype(int)
+
+        slice1[nn] = slice(start, stop)
```
kohr-h: Improve this loop in obvious ways please :-)

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+
+    return discr_space.element(phan)
+
+
+def indicate_proj_axis(discr_space, scale_structures=0.5):
+    """Phantom indicating along which axis it is projected.
+
+    The number (n) of rectangles in a parallel-beam projection along a main
+    axis (0, 1, or 2) indicates the projection to be along the (n-1)the
+    dimension.
```
kohr-h: From the documentation I have no idea what this function does.

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Parameters
+    ----------
+    discr_space : `DiscretizedSpace`
+        Discretized space in which the phantom is supposed to be created
+    scale_structures : positive `float` in (0, 1]
+        Scales objects (cube, cuboids)
```
kohr-h: More precise

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+     [ 0.  0.  0.  0.  0.  0.  0.  0.]]
+    >>> print(np.sum(phan, 2))
+    [[ 0.  0.  0.  0.  0.  0.  0.  0.]
+     [ 0.  0.  0.  2.  2.  0.  0.  0.]
+     [ 0.  0.  0.  2.  2.  0.  0.  0.]
+     [ 0.  0.  0.  0.  0.  0.  0.  0.]
+     [ 0.  0.  0.  0.  0.  0.  0.  0.]
+     [ 0.  0.  0.  0.  2.  0.  0.  0.]
+     [ 0.  0.  0.  2.  0.  0.  0.  0.]
+     [ 0.  0.  0.  0.  0.  0.  0.  0.]]
```
kohr-h: I still don't know what this function does.

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # 2nd cuboid of (dx[0], dx[1], 2 * dx[2]) touching the first diagonally
+    # at a long edge
+    x0 = (cen + 2 * dx)[1]
+    x1 = (cen + 3 * dx)[1]
+    y1 = cen[1]
+    z = (cen - dx)[2]
+    phan[x0:x1, y:y1, z:-z] = 1
+
+    return discr_space.element(phan)
```
kohr-h: @adler-j I know it's not your code, but I think it deserves some improvement anyway.

odl/phantom/geometric_phantoms.py
```diff
@@ -0,0 +1,189 @@
...
+
+        slice1[nn] = slice(start, stop)
+
+    phan[slice1] = 1
+
+    return discr_space.element(phan)
+
+
+def indicate_proj_axis(discr_space, scale_structures=0.5):
+    """Phantom indicating along which axis it is projected.
```
kohr-h: The name is also not great. Where does "projection" come from? No explanation.

odl/phantom/geometric_phantoms.py
adler-j: Neither do I, these were both added by Julian at some point. The examples give it away quite nicely though so I'd keep it. Hard to write it much better.

odl/phantom/misc_phantoms.py
```diff
@@ -0,0 +1,165 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Miscellaneous phantoms that do not fit in other cathegories."""
```
kohr-h: categories

odl/phantom/misc_phantoms.py
```diff
@@ -0,0 +1,165 @@
...
+import numpy as np
+
+
+__all__ = ('submarine',)
+
+
+def submarine(discr, smooth=True, taper=20.0):
+    """Return a 'submarine' phantom consisting in an ellipsoid and a box.
+
+    This phantom is used in [Okt2015]_ for shape-based reconstruction.
```
kohr-h: Remove citation, unpublished work

odl/phantom/geometric_phantoms.py
adler-j: Damn it, this is why you should never touch someone elses code : /

odl/phantom/noise.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Noise for phantoms."""
```
kohr-h: What does "for phantoms" mean? How about "Functions to create noise samples of different distributions."?

odl/phantom/phantom_utils.py
```diff
@@ -0,0 +1,254 @@
...
+            max_radius = np.sqrt([a_squared, b_squared, c_squared])
+            idx, shapes = _getshapes(center, max_radius, space.shape)
+
+            subgrid = [g[idi] for g, idi in zip(grid, shapes)]
+            squared_dist = [ai * (xi - x0i) ** 2
+                            for xi, ai, x0i in zip(subgrid,
+                                                   scales,
+                                                   [x0, y0, z0])]
+
+            # Parentisis to get best order for  broadcasting
```
kohr-h: Parentheses

odl/phantom/__init__.py
adler-j: Done for all except shepp logan where it cant be done nicely due to naming conflict (file and function has same name).

odl/phantom/shepp_logan_phantoms.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""The shepp-logan phantom, often used in transmission tomography."""
```
kohr-h: Shepp-Logan

odl/phantom/shepp_logan_phantoms.py
```diff
@@ -0,0 +1,117 @@
...
+from odl.phantom.phantom_utils import ellipse_phantom
+
+
+__all__ = ('shepp_logan',)
+
+
+def _shepp_logan_ellipse_2d():
+    """Return ellipse parameters for a 2d Shepp-Logan phantom.
+
+    This is the standard phantom in 2d medical imaging.
```
kohr-h: Sentence can be deleted, it's only a private helper function.

odl/phantom/shepp_logan_phantoms.py
```diff
@@ -0,0 +1,117 @@
...
+            [0.01, .0460, .0460, .050, 0.0000, -.1000, 0.00, 0.0, 0, 0],
+            [0.01, .0460, .0230, .050, -.0800, -.6050, 0.00, 0.0, 0, 0],
+            [0.01, .0230, .0230, .020, 0.0000, -.6060, 0.00, 0.0, 0, 0],
+            [0.01, .0230, .0460, .020, 0.0600, -.6050, 0.00, 0.0, 0, 0]]
+
+
+def _modified_shepp_logan_ellipses(ellipses):
+    """Modify ellipses to give the modified shepp-logan phantom.
+
+    Works for both 1d and 2d
```
kohr-h: 1d?

odl/phantom/shepp_logan_phantoms.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def shepp_logan(space, modified=False):
+    """Create a Shepp-Logan phantom.
+
+    The standard Shepp-Logan phantom in 2 or 3 dimensions.
+
+    References
+    ----------
+    Wikipedia : https://en.wikipedia.org/wiki/Shepp%E2%80%93Logan_phantom
```
kohr-h: Use reference in text, like 
```
The standard `Shepp-Logan phantom`_ in 2 or 3 dimensions.

References
----------
.. Shepp-Logan phantom: https://en.wikipedia.org/wiki/Shepp%E2%80%93Logan_phantom
```

odl/phantom/phantom_utils.py
```diff
@@ -0,0 +1,254 @@
...
+        scales = [1 / a_squared, 1 / b_squared]
+        mat = [[cos_p, sin_p],
+               [-sin_p, cos_p]]
+        radius = np.dot(scales, np.dot(mat, offset_points.T) ** 2)
+        inside = radius <= 1
+
+        # Add the ellipse intensity to those pixels
+        p[inside] += intensity
+
+    return space.element(p)
```
kohr-h: Any chance to get the same acceleration as in the 3d case? The 2d Shepp-Logan for example is a lot slower than the 3d one for large spaces.

odl/phantom/__init__.py
kohr-h: How about renaming the file to `transmission.py` if we want to make that distinction (analogous to `emission`)? Users are going to use `odl.phantom.shepp_logan` anyway.

setup.py
```diff
@@ -89,7 +89,7 @@ def run_tests(self):
 setup(
     name='odl',
 
-    version='0.2.3',
+    version='0.2.4',
```
kohr-h: This is not the only place. Make sure you get all of them (3 if I remember right).

odl/phantom/__init__.py
adler-j: Excellent idea

odl/phantom/geometric_phantoms.py
adler-j: I quite honestly think we'd need to rewrite this fully if we want to improve it.

kohr-h:
Nothing big to complain about, but some old code and doc could really use a refreshment. Bumping the version also requires to write release notes and making a Git tag. Depending on the status of the CUDA plugin branch, you may want to wait till that's done.

odl/phantom/phantom_utils.py
adler-j: If I was willing to spend half a day on it, yes. Dont really feel like it to save 50 ms right now. For the 3d case, the savings were about 10 seconds, so I found it worth the while.

We'll see.

setup.py
adler-j: Where are the others?

odl/phantom/phantom_utils.py
kohr-h: I agree that we shouldn't be hunting milliseconds, but 
```python
>>> discr = odl.uniform_discr([0, 0], [2, 2], [4000, 4000])
>>> %timeit odl.util.phantom.shepp_logan(discr)
1 loop, best of 3: 4.07 s per loop
```
Should keep it in mind in any case.

setup.py
kohr-h: One more is `__int__.py` I think. Need to check.

odl/phantom/phantom_utils.py
adler-j: Ok, so it seems the code was pretty trivial to port. With my computer I got 3.25s with the old code, and 0.17s with the new code on your example.

kohr-h: Great!

adler-j:
Fixed all comments except the version, need update there.

adler-j:
Also fixed a bug in `submarine`, should be mergable now.

kohr-h:
> Fixed all comments except the version, need update there.

I get the following hits when searching, disregarding binary files, bytecode and distribution stuff:
```
(odl-py35)hkohr@cormack:~/git/odl$ grep -r "0\.2\.3" .
./odl/__init__.py:__version__ = '0.2.3'
./conda/meta.yaml:    version: "0.2.3"
./conda/meta.yaml:    fn: odl-0.2.3.tar.gz
./conda/meta.yaml:    url: https://pypi.python.org/packages/c2/23/b208c56d0c25ec647cab9f164f050205c4792f5129deaf6b7a9cd3b99b1c/odl-0.2.3.tar.gz
./conda/meta.yaml:    # git_rev: v0.2.3
./setup.py:    version='0.2.3',
./doc/source/release_notes.rst:ODL 0.2.3 Release Notes (2016-06-12)
```

kohr-h:
When the version stuff is fixed, go ahead. It's probably better to have 0.2.4 in master anyway and make a release at a time we feel would be convenient.

kohr-h:
> When the version stuff is fixed, go ahead.

To be precise, this means only `setup.py` and `__init__.py`.

adler-j:
At what point was conda merged? Anyway I'll go ahead and fix those.

adler-j:
Updated version and made a release. Will merge after tests.

kohr-h:
Good. What I meant was that perhaps we should have 0.2.4 in master even if we haven't released yet.
If you decide to release, will you also write the release notes? Shouldn't be too much since 0.2.3. You can get an overview over the changes since 0.2.3 [here](https://github.com/odlgroup/odl/compare/v0.2.3...master). Then we also need an upload to PyPI and Anaconda.

kohr-h:
> Thats how numpy does it at least.

No, they push every minor release to PyPI. Check [here](https://pypi.python.org/pypi?%3Aaction=search&term=numpy&submit=search). I'm also against having Git tags and other releases out of sync.

kohr-h:
However, they only publish a minor version roughly every 3 months. Perhaps one should have 0.2.3.1 instead next time.

adler-j:
I guess, but we're in beta as well so I guess we'll publish more often. To me the last index isnt very serious, its mostly to make releasing the big ones (0.3) easier.

adler-j:
Also, they do not seem to do a PyPI release per tag, they are missing at least 1.10.0, 1.10.3 and 1.11.0. So i wouldnt worry to much about that.

This is breaking changes, and if anything should actually warrant a new minor release (0.3), but since we're in beta I guess we can call it a patch.

kohr-h:
Yeah, it's probably not that bad at the end of the day. Anyway, I'll put the package up once I have time. There's even a branch with a release script, but incomplete unfortunately :-)

kohr-h:
One last related note: We should have a running `upcoming_release.rst` file in the doc where each time a PR is merged, the corresponding changes are added (that's how Numpy handles it :-) ). Makes it much less work at the time of release to compile the release notes.

adler-j:
> One last related note: We should have a running upcoming_release.rst file in the doc where each time a PR is merged, the corresponding changes are added (that's how Numpy handles it :-) ). Makes it much less work at the time of release to compile the release notes.

Certainly agree.

PR #471, opened 2016-06-28
Issue 466  private variables

adler-j:
Initial go at changing protected variables to actual private ones in the correct places.

Are there any more we should change in this initial go at this? This felt like the most "obvious" ones.


adler-j:
I'm doing this across the code atm. Already found some "bugs" i.e. code that shouldn't really work but somehow does anyway.

odl/discr/discr_mappings.py
```diff
@@ -450,8 +455,8 @@ def __repr__(self):
         if self.order != 'C':
             inner_str += sep + "order='{}'".format(self.order)
             sep = ', '
-        if self._variant != 'left':
-            inner_str += sep + "variant='{}'".format(self._variant)
+        if self.__variant != 'left':
```
kohr-h: `self.variant`

odl/discr/grid.py
```diff
@@ -256,6 +256,16 @@ def max_pt(self):
...
+    @property
+    def ideg(self):
+        """Indices of the degenerate axes of this grid."""
+        return self.__ideg
+
+    @property
+    def inondeg(self):
+        """Indices of the non-degenerate axes of this grid."""
+        return self.__inondeg
+
```
kohr-h: Not sure we need these public. If we don't want to use the protected ones directly, I guess we can make underscored properties also.

odl/discr/grid.py
```diff
@@ -1044,9 +1054,9 @@ def squeeze(self):
         >>> g.squeeze()
         RegularGrid([0.0, 0.0], [1.0, 1.0], (5, 5))
         """
-        sq_minpt = [self.min_pt[axis] for axis in self._inondeg]
-        sq_maxpt = [self.max_pt[axis] for axis in self._inondeg]
-        sq_shape = [self.shape[axis] for axis in self._inondeg]
+        sq_minpt = [self.min_pt[axis] for axis in self.inondeg]
+        sq_maxpt = [self.max_pt[axis] for axis in self.inondeg]
+        sq_shape = [self.shape[axis] for axis in self.inondeg]
```
kohr-h: Here it would be okay to use `self.__inondeg`.

odl/discr/tensor_ops.py
```diff
@@ -369,7 +369,104 @@ def derivative(self, vf):
         return PointwiseInner(self.domain, inner_vf, weight=self.weights)
 
 
-class PointwiseInner(PointwiseTensorFieldOperator):
+class PointwiseInnerBase(PointwiseTensorFieldOperator):
+    """ Base class for `PointwiseInner` and `PointwiseInnerAdjoint`.
```
kohr-h: No space

odl/discr/tensor_ops.py
```diff
@@ -429,45 +526,8 @@ def __init__(self, vfspace, vecfield, weight=None):
...
-            self._weights = float(weight) * np.ones(self.domain.size)
-        else:
-            self._weights = np.asarray(weight, dtype='float64')
-            if (not np.all(self.weights > 0) or
-                    not np.all(np.isfinite(self.weights))):
-                raise ValueError('weighting array {} contains invalid '
-                                 'entries'.format(weight))
-        self._is_weighted = not np.array_equiv(self.weights, 1.0)
+        super().__init__(adjoint=False, vfspace=vfspace, vecfield=vecfield,
+                         weight=weight)
```
kohr-h: Yeah, this code duplication was horrible. Good move.

odl/discr/tensor_ops.py
```diff
@@ -571,10 +631,8 @@ def __init__(self, sspace, vecfield, vfspace=None, weight=None):
                 raise ValueError('base space of the range is different from '
                                  'the given scalar space ({!r} != {!r})'
                                  ''.format(vfspace[0], sspace))
-        super().__init__(vfspace, vecfield, weight=weight)
-
-        # Switch domain and range
-        self._domain, self._range = self._range, self._domain
+        super().__init__(adjoint=True, vfspace=vfspace, vecfield=vecfield,
+                         weight=weight)
```
kohr-h: Ah the good old super-hacky pattern :-)

odl/operator/operator.py
```diff
@@ -470,10 +470,10 @@ def __init__(self, domain, range, linear=False):
...
                             ''.format(range))
 
-        self._domain = domain
-        self._range = range
-        self._is_linear = bool(linear)
-        self._is_functional = isinstance(range, Field)
+        self.__domain = domain
+        self.__range = range
+        self.__is_linear = bool(linear)
+        self.__is_functional = isinstance(range, Field)
```
kohr-h: Perhaps a comment here that we cache this for efficiency, otherwise people (including me) will ask this over and again.

odl/space/weighting.py
```diff
@@ -117,9 +122,9 @@ def __eq__(self, other):
         `equiv` method.
         """
         return (isinstance(other, WeightingBase) and
+                self.impl == other.impl and
                 self.exponent == other.exponent and
-                self._dist_using_inner == other._dist_using_inner and
-                self.impl == other.impl)
+                self.dist_using_inner == other.dist_using_inner)
```
kohr-h: This one needs to be public, agreed.

odl/trafos/fourier.py
```diff
@@ -721,17 +726,7 @@ def _call_numpy(self, x):
...
-        if self.halfcomplex:
-            return np.fft.rfftn(x, axes=self.axes)
-        else:
-            if self.sign == '-':
-                return np.fft.fftn(x, axes=self.axes)
-            else:
-                # Need to undo Numpy IFFT scaling
-                return (np.prod(np.take(self.domain.shape, self.axes)) *
-                        np.fft.ifftn(x, axes=self.axes))
+        raise NotImplemented
```
kohr-h: This cannot be raised.

odl/trafos/fourier.py
```diff
@@ -1742,28 +1779,7 @@ def _call_numpy(self, x):
...
-            else:
-                out = np.fft.ifftn(preproc, axes=self.axes)
-                # Numpy's FFT normalizes by 1 / prod(shape[axes]), we
-                # need to undo that
-                out *= np.prod(np.take(self.domain.shape, self.axes))
-
-        # Post-processing accounting for shift, scaling and interpolation
-        self._postprocess(out, out=out)
-        return out
+        raise NotImplementedError('abstract method')
```
kohr-h: Should we adhere to this pattern?

kohr-h:
Quick skim, looking good so far. I didn't check for additional candidates for protected members, though.

odl/discr/discr_mappings.py
```diff
@@ -771,10 +776,10 @@ def __call__(self, x, out=None):
                 raise ValueError('output shape {} not equal to expected '
                                  'shape {}'.format(out.shape, out_shape))
 
-        indices, norm_distances = self._find_indices(x)
+        indices, norm_distances = self.__find_indices(x)
```
kohr-h: This method could actually be worth having non-protected to allow subclasses to implement more efficient variants.

odl/discr/grid.py
```diff
@@ -878,10 +888,10 @@ def __init__(self, min_pt, max_pt, shape):
         coord_vecs = [np.linspace(mi, ma, num, endpoint=True, dtype=np.float64)
                       for mi, ma, num in zip(min_pt, max_pt, shape)]
         TensorGrid.__init__(self, *coord_vecs)
-        self._center = (self.max_pt + self.min_pt) / 2
-        self._stride = np.zeros(len(shape), dtype='float64')
+        self.__center = (self.max_pt + self.min_pt) / 2
```
kohr-h: `midpoint` would be consistent with `RectPartition`

odl/discr/tensor_ops.py
```diff
@@ -369,7 +369,104 @@ def derivative(self, vf):
...
+        # Bail out if the space is complex but we cannot take the complex
+        # conjugate.
+        if (vfspace.field == ComplexNumbers() and
+                not hasattr(self.base_space.element_type, 'conj')):
+            raise NotImplementedError(
+                'base space element type {!r} does not implement conj() '
+                'method required for complex inner products'
+                ''.format(self.base_space.element_type))
+
+        self._vecfield = vfspace.element(vecfield)
```
kohr-h: How about dunder here?

odl/discr/tensor_ops.py
```diff
@@ -369,7 +369,104 @@ def derivative(self, vf):
...
+
+        # Handle weighting, including sanity checks
+        if weight is None:
+            if hasattr(vfspace.weighting, 'vector'):
+                self._weights = vfspace.weighting.vector
```
kohr-h: `__`?

odl/discr/tensor_ops.py
```diff
@@ -369,7 +369,104 @@ def derivative(self, vf):
...
+                raise ValueError('weighting constant must be positive, got '
+                                 '{}'.format(weight))
+            self._weights = float(weight) * np.ones(vfspace.size)
+        else:
+            self._weights = np.asarray(weight, dtype='float64')
+            if (not np.all(self.weights > 0) or
+                    not np.all(np.isfinite(self.weights))):
+                raise ValueError('weighting array {} contains invalid '
+                                 'entries'.format(weight))
+        self._is_weighted = not np.array_equiv(self.weights, 1.0)
```
kohr-h: `__`?

odl/discr/tensor_ops.py
```diff
@@ -369,7 +369,104 @@ def derivative(self, vf):
...
+
+    @property
+    def adjoint(self):
+        """Adjoint operator."""
+        raise NotImplementedError('abstract method')
+
+
+class PointwiseInner(PointwiseInnerBase):
 
     """Take the point-wise norm of a vector field.
```
kohr-h: Wrong text

odl/discr/tensor_ops.py
```diff
@@ -571,10 +631,8 @@ def __init__(self, sspace, vecfield, vfspace=None, weight=None):
...
                                  'the given scalar space ({!r} != {!r})'
                                  ''.format(vfspace[0], sspace))
-        super().__init__(vfspace, vecfield, weight=weight)
-
-        # Switch domain and range
-        self._domain, self._range = self._range, self._domain
+        super().__init__(adjoint=True, vfspace=vfspace, vecfield=vecfield,
+                         weight=weight)
 
         # Get weighting from range
```
kohr-h: Some dunder candidates below

odl/operator/operator.py
```diff
@@ -470,10 +470,10 @@ def __init__(self, domain, range, linear=False):
             raise TypeError('`range` {!r} is not a `Set` instance'
```
kohr-h: How about the guys in `__new__`? Or do we consider them part of the subclass API?

odl/space/base_ntuples.py
```diff
@@ -406,34 +406,54 @@ def __init__(self, size, dtype):
...
+        return self.astype(self.real_dtype)
+
+    @property
+    def complex_space(self):
+        """The space corresponding to this space's `complex_dtype`."""
+        return self.astype(self.complex_dtype)
 
     def _astype(self, dtype):
-        """Internal helper for ``astype``."""
+        """Internal helper for ``astype``. Can be overwritten by subclasses."""
```
kohr-h: overridden

kohr-h:
Some more places where attributes could be protected, but not too many. Otherwise I guess it's fine.

kohr-h:
But make sure to fix the commit messages, though.

odl/discr/grid.py
adler-j: It seems they are used in subclasses too. I'd keep them public?

odl/trafos/fourier.py
```diff
@@ -708,6 +670,49 @@ def _call(self, x, out, **kwargs):
...
+                'no adjoint defined for exponents ({}, {}) != (2, 2)'
+                ''.format(self.domain.exponent, self.range.exponent))
+
+    @property
+    def inverse(self):
+        """Inverse Fourier transform.
+
+        Abstract method.
+        """
+        return NotImplemented
```
adler-j: Should be raise

odl/trafos/fourier.py
adler-j: I find it quite useful aswell and it basically conveys what is wrong.

odl/discr/discr_mappings.py
adler-j: good idea.

odl/discr/tensor_ops.py
adler-j: Tried keeping the changes to places where we may expect someone to use a private variable by misstake, there are about 500 cases where we could do this change so I'm currently abit lazy in these specific classes.

odl/operator/operator.py
adler-j: Apparently we overwrite these in some other class (the functionspacevector) so they seem to be part of the subclass API.

This is kindoff the exact reason I'm doing this change, so things like this is obvious (and it should force us to document these things).

adler-j:
A quick search for the regexp `self.\_[^\_]` gives about 100 hits. Some are valid subclass things (like `_norm`) but many still missing. I'll keep going.

This also forces us to actually expose some things (like `op1` in `OperatorCompositon`)

odl/discr/grid.py
kohr-h: In subclasses you would also write `self.__ideg`, no? As long as we don't use `other.__ideg`.

adler-j: Now that these have found a public use [here](https://github.com/odlgroup/odl/blob/8c82c0b47db1c2b46048982e403e917c5ce10d27/odl/discr/partition.py#L536) I'll keep these public.

kohr-h: ok

odl/discr/grid.py
adler-j: Not an issue that will be fixed here.

adler-j:
How about a quick review and get this in. We can do more work later but this is the most urgent ones.

odl/discr/grid.py
kohr-h: It's already mentioned in #452, added a TODO list there.

kohr-h:
> How about a quick review and get this in. We can do more work later but this is the most urgent ones.

I'm gonna look at it. Get 529 hits with the regex search though. May be a lot in `diagnostics`.

kohr-h:
Some questions, starting with things that have no double underscore (yet):

- In `tensor_ops.py`: `_exponent`, `_weights`, `_is_weighted` in a number of places. Do we consider that part of the subclassing API?
- `default_ops.py`:
  * `ScalingOperator` has `_space` as an alias for `domain`, that could be removed completely. 
  * `MultiplyOperator` and others have `_domain_is_field` and `_range_is_field` - same kind of flag as `__is_linear`?
- `operator.py`:
  * `OperatorSum`: `_op1` and `_op2` could be public? (and renamed `left` and `right`)
  * `OperatorComp` - `left` and `right` public?
  * `OperatorPointwiseProduct`: see `OperatorSum`
  * `OperatorLeftScalarMult`: `_op` public?
  * ... goes on like that, principal decision needed
- `weighting.py`:
  * `MatrixWeightingBase`: The `_eigval` related stuff should probably be private, it's a kind of state
- Lots of stuff in the `tomo` part.

I'm a bit unsure where to draw the line here. Probably anything that can be set explicitly via `__init__` can easily be "private + public property", right? Since there is a canonical way to set those effectively even for the subclass. More unclear seem to be the protected attributes that are defined independently from or loosely derived from input parameters. Those could in principle be private, too. For me, the only good use case for protected attributes are attributes (mostly methods) that can or should explicitly be overridden by subclasses. Am I missing something?

kohr-h:
So, to be more clear: Do we want this?

- `__init__` argument `arg` stored as `__arg` and accessed publicly through `arg` property
- Attributes directly derived from `__init__` args, e.g. `isinstance(range, Field)` stored as protected (single `_`) to allow subclasses to change them.
- Attributes loosely derived from `__init__` args - ? Is there such a category? If yes, what to do with it?
- State attributes for class-internal stuff private (`__`).
- How do we handle the case of state variables that take much memory? We probably don't want each subclass to add another state variable in that case. Just make them private so we don't get that duplication?

adler-j:
> In `tensor_ops.py`: `_exponent`, `_weights`, `_is_weighted` in a number of places. Do we consider that part of the subclassing API?

Currently, most if not all things I've made private are variables that can be set through the `__init__` method. This is basically a way to enforce "There should be one-- and preferably only one --obvious way to do it.". According to that logic, those should be made private, and sublclasses should access them through the `PointwiseNorm.__init__` etc methods.

>  `ScalingOperator` has `_space` as an alias for `domain`, that could be removed completely. 

:+1:

> `MultiplyOperator` and others have `_domain_is_field` and `_range_is_field` - same kind of flag as `__is_linear`?

Yes

>   * `OperatorSum`: `_op1` and `_op2` could be public? (and renamed `left` and `right`)
  * `OperatorComp` - `left` and `right` public?
  * `OperatorPointwiseProduct`: see `OperatorSum`
  * `OperatorLeftScalarMult`: `_op` public?
  * ... goes on like that, principal decision needed

#486

> `MatrixWeightingBase`: The `_eigval` related stuff should probably be private, it's a kind of state

Yes

> Lots of stuff in the `tomo` part.

Sure, but there arn't many who will subclass them anytime soon, so we could leave that for now.

>  `__init__` argument `arg` stored as `__arg` and accessed publicly through `arg` property.

In what classes? If you mean in general, certainly no. I can think of a hundred memory leaks and inefficiencies this would cause.

> Attributes directly derived from `__init__` args, e.g. `isinstance(range, Field)` stored as protected (single `_`) to allow subclasses to change them.

No. Anything that follows from the `__init__` args should be left as is.

> Attributes loosely derived from `__init__` args - ? Is there such a category? If yes, what to do with it?

Don't exactly follow what you mean, could use an example.

> State attributes for class-internal stuff private (`__`)

In general yes

kohr-h:
>> In `tensor_ops.py`: `_exponent`, `_weights`, `_is_weighted` in a number of places. Do we consider that part of the subclassing API?

> Currently, most if not all things I've made private are variables that can be set through the `__init__` method. This is basically a way to enforce "There should be one-- and preferably only one --obvious way to do it.". According to that logic, those should be made private, and sublclasses should access them through the `PointwiseNorm.__init__` etc methods.

Same thought here.

>>   * `OperatorSum`: `_op1` and `_op2` could be public? (and renamed `left` and `right`)
  * `OperatorComp` - `left` and `right` public?
  * `OperatorPointwiseProduct`: see `OperatorSum`
  * `OperatorLeftScalarMult`: `_op` public?
  * ... goes on like that, principal decision needed

> #486

:+1: 

>> Lots of stuff in the `tomo` part.

> Sure, but there arn't many who will subclass them anytime soon, so we could leave that for now.

Okay.

>>  `__init__` argument `arg` stored as `__arg` and accessed publicly through `arg` property.

> In what classes? If you mean in general, certainly no. I can think of a hundred memory leaks and inefficiencies this would cause.

Usually I would expect things provided in initialization to be accessible afterwards. Public property and `__arg` seems like the natural pattern. If we have concerns about memory (e.g. in `PointwiseInner` where a vector field is stored), we should perhaps be more clear in the documentation that users should delete the class when it's no longer used. Or use arguments in a way that does not require copying.

>> Attributes directly derived from `__init__` args, e.g. `isinstance(range, Field)` stored as protected (single `_`) to allow subclasses to change them.

> No. Anything that follows from the `__init__` args should be left as is.

Okay.

>> Attributes loosely derived from `__init__` args - ? Is there such a category? If yes, what to do with it?

> Don't exactly follow what you mean, could use an example.

Probably I didn't know what I meant myself..

>> How do we handle the case of state variables that take much memory? We probably don't want each subclass to add another state variable in that case. Just make them private so we don't get that duplication?

> I don't have an example where the current design (in this branch) has issues with this, do you? If the superclass stores too much things that are not changeable according to its interface, we have a bad design in that place.

I don't think we have something like that currently, but classes with internal large stuff, like `MatrixWeighting` which stores a matrix decomposition. A subclass would then add things on top, for example if a decomposition could be given as arguments to the subclass. This is not a real use case, though, since the caching can be disabled.

adler-j:
> Usually I would expect things provided in initialization to be accessible afterwards. Public property and __arg seems like the natural pattern. If we have concerns about memory (e.g. in PointwiseInner where a vector field is stored), we should perhaps be more clear in the documentation that users should delete the class when it's no longer used. Or use arguments in a way that does not require copying.

My main issue with this is that it is a hugely sweeping change that we would need to apply to all classes across ODL (we're talking hundreds). This would cause ODL objects to be somehow different from other objects in python. Currently we have no `odlobject` style class, and frankly I'd like it to stay that way if at all possible.

> I don't think we have something like that currently, but classes with internal large stuff, like MatrixWeighting which stores a matrix decomposition. A subclass would then add things on top, for example if a decomposition could be given as arguments to the subclass. This is not a real use case, though, since the caching can be disabled.

I say we leave it as is for now and see if something comes up.

kohr-h:
Maybe just to clarify: with "accessible" I mostly meant "read-accessible", i.e. if you provide `arg` to an initializer, you would expect to be able to get it back somehow, especially when you want to create a new instance with a subset of the old parameters. Correct me if I'm wrong, but I think that's what 95 % of the current classes do, no big sweeping change.

Otherwise, I don't think there is more to do here apart from the few things I saw above.

adler-j:
> Maybe just to clarify: with "accessible" I mostly meant "read-accessible", i.e. if you provide arg to an initializer, you would expect to be able to get it back somehow, especially when you want to create a new instance with a subset of the old parameters. Correct me if I'm wrong, but I think that's what 95 % of the current classes do, no big sweeping change.

Obviously a lapse of communication, I thought you suggested that we store all arguments as `__args`, allowing `type(self)(self.args)`. Now I see what you mean and I agree.

kohr-h:
> Obviously a lapse of communication, I thought you suggested that we store all arguments as __args, allowing type(self)(self.args). Now I see what you mean and I agree.

Okay cool. How you understood me is actually something for [`pickle`](https://docs.python.org/3/library/pickle.html). Here one can configure a class in a way that makes it possible to serialize and re-create, using [a standardized protocol](http://rafekettler.com/magicmethods.html#pickling). But that seems to be a maintenance nightmare and definitely out of scope here. Still interesting :-)

adler-j:
A short comment:

> MatrixWeightingBase: The _eigval related stuff should probably be private, it's a kind of state

Apparently it is used as a subclass API in `NumpyFnMatrixWeighting`. Debatable design, but I'll leave it for now.

Except for that I'll get this ready for a merge.

kohr-h:
Yeah well. I've had some thoughts about removing that one altogether while writing tensor stuff. I don't think we ever used it, and the idea really doesn't scale at all to higher-dimensional objects. Even for 1D it will be bloody slow in most cases.
Probably a better and more useful weighting would be an "outer product" type of thing, where a series of 1D manipulations is performed, one per axis. That's at least feasible for tensors, too.
Not the topic of this issue, though.

adler-j:
> Except for that I'll get this ready for a merge.

Done. Merge?

odl/operator/default_ops.py
```diff
@@ -64,20 +64,17 @@ def scalar(self):
 
     def _call(self, x, out=None):
         """Scale input and write to output.
-
```
kohr-h: What happened here?

odl/operator/default_ops.py
```diff
@@ -64,20 +64,17 @@ def scalar(self):
...
         Parameters
         ----------
         x : ``domain`` `element`
             input vector to be scaled
         out : ``range`` `element`, optional
             Output vector to which the result is written
-
```
kohr-h: ?

odl/operator/default_ops.py
```diff
@@ -64,20 +64,17 @@ def scalar(self):
...
         Returns
         -------
         out : ``range`` `element`
             Result of the scaling. If ``out`` was provided, the
             returned object is a reference to it.
-
```
kohr-h: ?

odl/operator/default_ops.py
```diff
@@ -101,7 +98,6 @@ def _call(self, x, out=None):
     @property
     def inverse(self):
         """Return the inverse operator.
-
```
kohr-h: ?

kohr-h:
Apart from those blank line errors and some of the commit messages good to go.

PR #472, opened 2016-06-28
WIP: Implement convolution operator(s)

kohr-h:
The real-space convolution is fully working now and could use a design review. Here's a quick summary of its behavior with respect to the different possibilities a kernel can be provided:
1. **As a `DiscreteLpVector`.** The kernel has information about its own function domain, which results in the convolution operator to map to a translated space, where the shift is the midpoint of the kernel domain. The reason for this is that mathematically, the region shifted in this way contains the main "mass" of the convolved function.
2. **As a `callable`.** In this case, a "standard" kernel space is created, which is a zero-centered version of the operator domain (same cell sizes and shape). The operator range is the same as the domain.
3. **As an array-like.** Similarly to case 2, a standard zero-centered kernel space is created, but this time with the same shape as the given array. This is actually the most useful (=efficient) case when one wants to calculate a convolution with a very small kernel.

The Fourier-based convolution is currently not functioning, ignore it for now.

### DIFF
--- a/examples/trafos/convolution.py
+++ b/examples/trafos/convolution.py
@@ -1,0 +1,90 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example for the usage of the convolution operator."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import odl
+
+
+# Parameters for the gaussian kernel
+gamma = 0.1
+norm_const = 2 * np.pi * gamma ** 2
+
+
+def gaussian(x):
+    sum_sq = sum(xi ** 2 for xi in x)
+    return np.exp(-sum_sq / (2 * gamma ** 2)) / norm_const
+
+
+# Test function
+def square(x):
+    onedim_arrs = [np.where((xi >= -1) & (xi <= 1), 1.0, 0.0) for xi in x]
+    out = onedim_arrs[0]
+    for arr in onedim_arrs[1:]:
+        out = out * arr
+    return out
+
+
+space = odl.uniform_discr([-2, -2], [2, 2], (2048, 2048))
+
+# Showing just for fun
+real_ker = space.element(gaussian)
+real_ker.show(title='Gaussian kernel')
+func = space.element(square)
+func.show(title='Test function, a square')
+
+conv = odl.Convolution(space, kernel=gaussian, kernel_mode='real',
+                       impl='pyfftw_ft')
+
+out = space.element()
+conv.transform.create_temporaries()
+conv.transform.init_fftw_plan()
+
+with odl.util.testutils.Timer('first run, mode real'):
+    func_conv = conv(func, out=out)
+
+with odl.util.testutils.Timer('second run, mode real'):
+    func_conv = conv(func, out=out)
+
+conv.transform.create_temporaries()
+with odl.util.testutils.Timer('third run, mode real, with tmp'):
+    func_conv = conv(func, out=out)
+
+func_conv.show(title='Convolved function')
+
+# Giving the kernel via FT
+ft = odl.trafos.FourierTransform(space, impl='pyfftw')
+ker_ft = ft(gaussian)
+ker_ft.show(title='FT of Gaussian kernel, half-complex')
+
+conv_with_ker_ft = odl.Convolution(space, kernel=ker_ft, kernel_mode='ft_hc',
+                                   impl='pyfftw_ft')
+
+with odl.util.testutils.Timer('first run, mode ft_hc'):
+    func_conv = conv_with_ker_ft(func, out=out)
+
+conv_with_ker_ft.transform.create_temporaries()
+with odl.util.testutils.Timer('second run, mode ft_hc, using tmp'):
+    func_conv = conv_with_ker_ft(func, out=out)
+
+func_conv.show(title='Convolved function, using kernel ft')
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -192,6 +192,11 @@ class DiscreteLp(DiscretizedSpace):
         return self.partition.max_pt
 
     @property
+    def mid_pt(self):
+        """Midpoint of the function domain."""
+        return self.partition.mid_pt
+
+    @property
     def order(self):
         """Axis ordering for array flattening."""
         return self.__order
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -480,14 +480,13 @@ class DiscreteLp(DiscretizedSpace):
             arg_fstr = '''
     {!r},
     {!r},
-    {!r}
-    '''
+    {!r}'''
             if self.exponent != 2.0:
-                arg_fstr += ', exponent={ex}'
-            if self.interp != 'nearest':
-                arg_fstr += ', interp={interp!r}'
+                arg_fstr += ',\n    exponent={ex}'
+            if self.interp != ['nearest'] * self.ndim:
+                arg_fstr += ',\n    interp={interp!r}'
             if self.order != 'C':
-                arg_fstr += ', order={order!r}'
+                arg_fstr += ',\n    order={order!r}'
 
             arg_str = arg_fstr.format(
                 self.uspace, self.partition, self.dspace, interp=self.interp,
@@ -1539,6 +1538,8 @@ def uniform_discr_fromdiscr(discr, min_pt=None, max_pt=None,
 
     return uniform_discr_frompartition(new_part, exponent=discr.exponent,
                                        interp=discr.interp, impl=discr.impl,
+                                       order=discr.order, dtype=discr.dtype,
+                                       weighting=discr.weighting,
                                        **kwargs)
 
 
--- a/odl/trafos/__init__.py
+++ b/odl/trafos/__init__.py
@@ -33,3 +33,6 @@ __all__ += fourier.__all__
 
 from .wavelet import *
 __all__ += wavelet.__all__
+
+from .convolution import *
+__all__ += convolution.__all__
--- a/odl/trafos/convolution.py
+++ b/odl/trafos/convolution.py
@@ -408,7 +408,8 @@ class RealSpaceConvolution(Convolution):
 
     """Convolution implemented in real space."""
 
-    def __init__(self, domain, kernel, range=None, **kwargs):
+    def __init__(self, domain, kernel, range=None, impl='scipy_convolve',
+                 **kwargs):
         """Initialize a new instance.
 
         Parameters
@@ -505,8 +506,8 @@ class RealSpaceConvolution(Convolution):
         if not isinstance(domain, DiscreteLp):
             raise TypeError('domain {!r} is not a DiscreteLp instance'
                             ''.format(domain))
-        if not domain.partition.is_regular:
-            raise NotImplementedError('irregular sampling not supported')
+        if not domain.is_uniform:
+            raise ValueError('irregular sampling not supported')
 
         if range is not None:
             # TODO
@@ -518,27 +519,23 @@ class RealSpaceConvolution(Convolution):
 
         super().__init__(domain, range, linear=True)
 
-        # Handle kernel mode and impl
-        impl = kwargs.pop('impl', 'scipy_convolve')
+        # Handle impl
         impl, impl_in = str(impl).lower(), impl
         if impl not in _REAL_CONV_SUPPORTED_IMPL:
             raise ValueError("implementation '{}' not understood"
                              ''.format(impl_in))
-
         self._impl = impl
 
         # Handle the `resample` input parameter
         resample = normalized_scalar_param_list(kwargs.pop('resample', 'down'),
                                                 length=self.domain.ndim,
                                                 param_conv=str)
-
         for i, r in enumerate(resample):
             if r.lower() not in ('up', 'down'):
                 raise ValueError("in axis {}: `resample` '{}' not understood"
                                  ''.format(i, r))
             else:
                 resample[i] = r.lower()
-
         self._resample = resample
 
         # Initialize resampling operators
@@ -589,35 +586,30 @@ class RealSpaceConvolution(Convolution):
         range : `DiscreteLp`
             Range of the convolution operator.
         """
-        if isinstance(kernel, DiscreteLpVector) and kernel not in domain:
+        if isinstance(kernel, domain.element_type):
             if kernel.ndim != domain.ndim:
                 raise ValueError('`kernel` has dimension {}, expected {}'
                                  ''.format(kernel.ndim, domain.ndim))
-            # The kernel lies in a different space. We first zero-center
-            # its domain and set the range to the domain shifted by
-            # the the midpoint of the kernel domain.
-
-            # TODO: DiscreteLp.element should probably bail out if
-            # kernel is a DiscreteLpVector but not in the same space
+            # Set the range equal to the domain shifted by the midpoint of
+            # the kernel space.
             kernel_shift = kernel.space.partition.midpoint
             range = uniform_discr_fromdiscr(
                 domain, min_corner=domain.min_corner + kernel_shift)
+            return kernel, range
 
-        elif callable(kernel):
-            # The kernel is a Python function or some other callable object.
-            kernel = domain.element(kernel, **kernel_kwargs)
-            range = domain
         else:
-            try:
-                # kernel is an array-like of the correct shape or an
-                # element of `domain`
-                kernel = domain.element(kernel)
-            except (TypeError, ValueError):
-                # Array-like of wrong shape -> find adequate space for the
-                # kernel. Assume 0-centering and same cell size.
-
-                # TODO: handle the case when the outer cells are of
-                # different size
+            if callable(kernel):
+                # Use "natural" kernel space, which is the zero-centered
+                # version of the domain.
+                extent = domain.partition.extent()
+                std_kernel_space = uniform_discr_fromdiscr(
+                    domain, min_corner=-extent / 2)
+                kernel = std_kernel_space.element(kernel, **kernel_kwargs)
+                range = domain
+                return kernel, range
+            else:
+                # Make a zero-centered space with the same cell sides as
+                # domain, but shape according to the kernel.
                 kernel = np.asarray(kernel, dtype=domain.dtype)
                 nsamples = kernel.shape
                 extent = domain.cell_sides * nsamples
@@ -625,10 +617,8 @@ class RealSpaceConvolution(Convolution):
                     domain, min_corner=-extent / 2, max_corner=extent / 2,
                     nsamples=nsamples)
                 kernel = kernel_space.element(kernel)
-
-            range = domain
-
-        return kernel, range
+                range = domain
+                return kernel, range
 
     @staticmethod
     def _get_resamp_csides(resample, ker_csides, dom_csides):
@@ -667,8 +657,8 @@ class RealSpaceConvolution(Convolution):
         """Return the real-space kernel of this convolution operator.
 
         Note that internally, the scaled kernel is stored if
-        ``scale=True`` was chosen during initialization, i.e. calculating
-        the original kernel requires to reverse the scaling.
+        ``scale=True`` (default) was chosen during initialization, i.e.
+        calculating the original kernel requires to reverse the scaling.
         Otherwise, the original unscaled kernel is returned.
         """
         if self._scale:
@@ -680,27 +670,41 @@ class RealSpaceConvolution(Convolution):
         """Implement ``self(x)``."""
         if np.allclose(self._kernel.space.cell_sides, self.domain.cell_sides):
             # No resampling needed
-            print('no resampling')
-
-            # According to the documentation of scipy.signal.convolve, we
-            # would need to switch order of x and kernel since `convolve`
-            # expects the larger array first. However, this does not seem
-            # to be needed.
             return signal.convolve(x, self._kernel, mode='same')
         else:
             if self.domain != self._domain_resampling_op.range:
                 # Only resample if necessary
-                print('resample function')
                 x = self._domain_resampling_op(x)
 
             if self._kernel.space != self._kernel_resampling_op.range:
-                print('resample kernel')
                 kernel = self._kernel_resampling_op(self._kernel)
             else:
                 kernel = self._kernel
 
             conv = signal.convolve(x, kernel, mode='same')
             return self._domain_resampling_op.inverse(conv)
+
+    @property
+    def adjoint(self):
+        """Adjoint operator defined by the reversed kernel."""
+        # Make space for the adjoint kernel, -S if S is the kernel space
+        adj_min_corner = -self._kernel.space.max_corner
+        adj_max_corner = -self._kernel.space.min_corner
+        adj_ker_space = uniform_discr_fromdiscr(
+            self._kernel.space,
+            min_corner=adj_min_corner, max_corner=adj_max_corner)
+
+        # Reverse the kernel
+        reverse_slice = (slice(None, None, -1),) * self.domain.ndim
+        adj_kernel = adj_ker_space.element(
+            self._kernel.asarray()[reverse_slice])
+
+        adj_conv = RealSpaceConvolution(self.range, adj_kernel, impl=self.impl,
+                                        scale=False, resample=self.resample)
+        # Kernel is already scaled, but we need to make sure that the
+        # kernel() method returns the rescaled kernel.
+        adj_conv._scale = True
+        return adj_conv
 
 
 if __name__ == '__main__':
--- a/odl/trafos/convolution.py
+++ b/odl/trafos/convolution.py
@@ -26,23 +26,23 @@ from builtins import super
 import numpy as np
 import scipy.signal as signal
 
-from odl.discr.discr_ops import Resampling
-from odl.discr.lp_discr import (
-    DiscreteLp, DiscreteLpElement, uniform_discr, uniform_discr_fromdiscr)
-from odl.operator.operator import Operator
-from odl.set.sets import ComplexNumbers
-from odl.trafos.fourier import FourierTransform
+from odl.discr import (
+    DiscreteLp, uniform_discr, uniform_discr_fromdiscr)
+from odl.operator import Operator, OpDomainError, OpRangeError
+from odl.set import ComplexNumbers
+from odl.trafos import FourierTransform, DiscreteFourierTransform
 from odl.util.normalize import normalized_scalar_param_list, safe_int_conv
 
 
-__all__ = ('Convolution', 'FourierSpaceConvolution', 'RealSpaceConvolution')
+__all__ = ('FourierSpaceConvolution', 'RealSpaceConvolution')
 
 
 _REAL_CONV_SUPPORTED_IMPL = ('scipy_convolve',)
-_FOURIER_CONV_SUPPORTED_IMPL = ('default', 'pyfftw')
+# TODO: get from fourier.py
+_FOURIER_CONV_SUPPORTED_IMPL = ('numpy', 'pyfftw')
 
 
-class Convolution(Operator):
+class ConvolutionBase(Operator):
 
     """Discretization of the convolution integral as an operator.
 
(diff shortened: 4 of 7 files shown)


examples/trafos/convolution.py
```diff
@@ -0,0 +1,90 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example for the usage of the convolution operator."""
```
kohr-h: Ignore for now, too.

odl/trafos/__init__.py
```diff
@@ -33,3 +33,6 @@
 from . import wavelet
 from .wavelet import *
 __all__ += wavelet.__all__
+
+from .convolution import *
+__all__ += convolution.__all__
```
kohr-h: Not sure about the location. Needs to be at least as "high" as `trafos`, though, since the Fourier implementation will depend on the FT.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Operators based on (integral) transformations."""
```
kohr-h: Obviously old text

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+__all__ = ('Convolution', 'FourierSpaceConvolution', 'RealSpaceConvolution')
+
+
+_REAL_CONV_SUPPORTED_IMPL = ('scipy_convolve',)
+_FOURIER_CONV_SUPPORTED_IMPL = ('default_ft', 'pyfftw_ft')
+_FOURIER_CONV_SUPPORTED_KER_MODES = ('real', 'ft', 'ft_hc')
+
+
+class Convolution(Operator):
+    pass
```
kohr-h: Common attributes will go here if we want to use this pattern.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+            else:
+                kernel_mode = 'ft'
+            return Convolution(dom=self.domain,
+                               kernel=adj_kernel_ft, kernel_mode=kernel_mode,
+                               impl=self.impl, axes=self.axes)
+        else:
+            raise RuntimeError('both kernel and kernel_transform are None.')
+
+
+class RealSpaceConvolution(Convolution):
```
kohr-h: Reviewable code begins here.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+          kernel :math:`k`, i.e. the linear operator :math:`\mathcal{C}_k`
+          given by
+
+              :math:`\mathcal{C}_k(f): x \mapsto [k \\ast f](x)`.
+
+        - If the convolution kernel :math:`k` is defined on a rectangular
+          domain :math:`\Omega_0` with midpoint :math:`m_0`, then the
+          convolution :math:`k \\ast f` of a function
+          :math:`f:\Omega \\to \mathbb{R}` with :math:`k` has a support
+          that is a superset of :math:`\Omega + m_0`. Therefore, we
```
kohr-h: subset!!

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+          `Minkowski sum
+          <https://en.wikipedia.org/wiki/Minkowski_addition>`_
+          :math:`\Omega + \Omega_0`.
+
+          For example, if :math:`\Omega = [0, 5]` and
+          :math:`\Omega_0 = [1, 3]`, i.e. :math:`m_0 = 2`, then
+          :math:`\Omega + \Omega_0 = [1, 8]`. However, we choose the
+          shifted interval :math:`\Omega + m_0 = [2, 7]` for the range
+          of the convolution since it contains the "main mass" of the
+          result.
```
kohr-h: Perhaps a bit long-ish, could as well go into a chapter in the online doc.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+        """
+        # Basic checks
+        if not isinstance(domain, DiscreteLp):
```
kohr-h: Question here: How about convolutions on `FnBase`? Would make sense mathematically, but feels like lots of `if...else` stuff that would make the code unreadable.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+        resample = normalized_scalar_param_list(kwargs.pop('resample', 'down'),
+                                                length=self.domain.ndim,
+                                                param_conv=str)
+        for i, r in enumerate(resample):
+            if r.lower() not in ('up', 'down'):
+                raise ValueError("in axis {}: `resample` '{}' not understood"
+                                 ''.format(i, r))
+            else:
+                resample[i] = r.lower()
+        self._resample = resample
```
kohr-h: The `str(foo).lower()` thing should be a converter in `odl.util.normalize`.

kohr-h: Or `lambda s: str(s).lower()` as `param_conv`.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,713 @@
...
+        # Reverse the kernel
+        reverse_slice = (slice(None, None, -1),) * self.domain.ndim
+        adj_kernel = adj_ker_space.element(
+            self._kernel.asarray()[reverse_slice])
+
+        adj_conv = RealSpaceConvolution(self.range, adj_kernel, impl=self.impl,
+                                        scale=False, resample=self.resample)
+        # Kernel is already scaled, but we need to make sure that the
+        # kernel() method returns the rescaled kernel.
+        adj_conv._scale = True
```
kohr-h: This is a slight hack, but I don't want to scale back and forth in a property, could be expensive.

examples/trafos/convolution.py
```diff
@@ -0,0 +1,90 @@
...
+    return np.exp(-sum_sq / (2 * gamma ** 2)) / norm_const
+
+
+# Test function
+def square(x):
+    onedim_arrs = [np.where((xi >= -1) & (xi <= 1), 1.0, 0.0) for xi in x]
+    out = onedim_arrs[0]
+    for arr in onedim_arrs[1:]:
+        out = out * arr
+    return out
```
adler-j: why not

```python
return (x[0] > -1) & (x[0] < 1) & (x[1] > -1) & (x[1] < 1)
```

or use the phantom in ODL

```python
func = odl.phantom.cuboid([-1, -1], [1, 1])
```

examples/trafos/convolution.py
```diff
@@ -0,0 +1,90 @@
...
+func.show(title='Test function, a square')
+
+conv = odl.Convolution(space, kernel=gaussian, kernel_mode='real',
+                       impl='pyfftw_ft')
+
+out = space.element()
+conv.transform.create_temporaries()
+conv.transform.init_fftw_plan()
+
+with odl.util.testutils.Timer('first run, mode real'):
```
adler-j: It seems `Timer` should be made accessible in `odl.util`

examples/trafos/convolution.py
```diff
@@ -0,0 +1,90 @@
...
+conv.transform.create_temporaries()
+with odl.util.testutils.Timer('third run, mode real, with tmp'):
+    func_conv = conv(func, out=out)
+
+func_conv.show(title='Convolved function')
+
+# Giving the kernel via FT
+ft = odl.trafos.FourierTransform(space, impl='pyfftw')
+ker_ft = ft(gaussian)
+ker_ft.show(title='FT of Gaussian kernel, half-complex')
```
adler-j: `title=` is not needed anymore, the string goes first.

examples/trafos/convolution.py
adler-j: This file could need more documenting, many steps are non-obvious.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+from builtins import super
+
+import numpy as np
+import scipy.signal as signal
+
+from odl.discr.discr_ops import Resampling
+from odl.discr.lp_discr import (
+    DiscreteLp, DiscreteLpVector, uniform_discr, uniform_discr_fromdiscr)
+from odl.operator.operator import Operator
+from odl.space.ntuples import Ntuples
```
adler-j: We need to improve these in general. We should do this like:

```python
from odl.space import Ntuples, CudaNtuples
```

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+
+__all__ = ('Convolution', 'FourierSpaceConvolution', 'RealSpaceConvolution')
+
+
+_REAL_CONV_SUPPORTED_IMPL = ('scipy_convolve',)
+_FOURIER_CONV_SUPPORTED_IMPL = ('default', 'pyfftw')
+_FOURIER_CONV_SUPPORTED_KER_MODES = ('real', 'fourier')
+
+
+class Convolution(Operator):
```
adler-j: There seems to be a missmatch with the example. This code doesnt do anything, but the example uses it?

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+    """
+
+    def kernel(self):
+        """Return the kernel of this convolution operator."""
+        raise NotImplementedError('abstract method')
+
+
+class FourierSpaceConvolution(Convolution):
+
+    """Convolution with a kernel, implemented in Fourier space.
```
adler-j: `with a kernel` is rather obvious, no? What else could you do?

Id go for `Convolution implemented in Fourier space.`

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+
+    This operator implements a discrete approximation to the continuous
+    convolution with a fixed kernel. It supports real-space and
+    Fourier based back-ends.
+    """
+
+    def __init__(self, domain, kernel, **kwargs):
```
adler-j: Why should I need to give both `domain` and `kernel` in case `kernel` is a `domain element` this should be clear, no?

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+            # Sparse kernel (less dimensions), blow up
+            slc = [None] * self.domain.ndim
+            for ax in axes:
+                slc[ax] = slice(None)
+
+            kernel = kernel[slc]
+            # Assuming uniform discretization
+            min_corner = -self.domain.cell_sides * self.domain.shape / 2
+            max_corner = self.domain.cell_sides * self.domain.shape / 2
+            if isinstance(self.domain.dspace, Ntuples):
```
adler-j: This shall clearly be changed to `impl = self.domain.impl` once the odlcuda stuff is merged.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+            slc = [None] * self.domain.ndim
+            for ax in axes:
+                slc[ax] = slice(None, None, -1)
+            return kernel[slc]
+
+    @property
+    def adjoint(self):
+        """Adjoint operator."""
+        if self.kernel is not None:
+            # TODO: this could be expensive. Move to init?
```
adler-j: Leave for now. We can profile later. Moving it to init also carries a space cost.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+            return Convolution(dom=self.domain,
+                               kernel=adj_kernel, kernel_mode='real',
+                               impl=self.impl, axes=self.axes)
+
+        elif self.kernel_transform is not None:
+            # TODO: this could be expensive. Move to init?
+            adj_kernel_ft = self.kernel_transform.conj()
+
+            if self.transform.halfcomplex:
+                kernel_mode = 'ft_hc'
```
adler-j: You seem to have two conventions for how you name this.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+            else:
+                kernel_mode = 'ft'
+            return Convolution(dom=self.domain,
+                               kernel=adj_kernel_ft, kernel_mode=kernel_mode,
+                               impl=self.impl, axes=self.axes)
+        else:
+            raise RuntimeError('both kernel and kernel_transform are None.')
```
adler-j: remove dot

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+
+
+class RealSpaceConvolution(Convolution):
+
+    """Convolution implemented in real space."""
```
adler-j: This could clearly use some work

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+
+    def __init__(self, domain, kernel, range=None, impl='scipy_convolve',
+                 **kwargs):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain : `DiscreteLp`
+            Domain of the operator
```
adler-j: This could perhaps be more clear

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+            sized than the convolved function. It is assumed to be an
+            element of a space with same `DiscreteLp.cell_sides` as
+            ``domain``. Note that the ``scipy_convolve`` back-end does
+            not allow the kernel to be larger.
+            Valid for ``impl``: ``'scipy_convolve'``
+
+        range : `DiscreteLp`, optional
+            Range of the operator. By default, the range is calculated
+            from the kernel.
+            Note: custom range is currently not supported.
```
adler-j: Dont add it this, imo.

odl/trafos/convolution.py
adler-j: I say no for now. That would need to be a new class, or at least a later work.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+          shifted interval :math:`\Omega + m_0 = [2, 7]` for the range
+          of the convolution since it contains the "main mass" of the
+          result.
+        """
+        # Basic checks
+        if not isinstance(domain, DiscreteLp):
+            raise TypeError('domain {!r} is not a DiscreteLp instance'
+                            ''.format(domain))
+        if not domain.is_uniform:
+            raise ValueError('irregular sampling not supported')
```
adler-j: non-uniform

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+                                                kernel_resamp_space)
+
+        # Scale the kernel if desired
+        self._kernel_scaling = np.prod(new_kernel_csides)
+        self._scale = bool(kwargs.pop('scale', True))
+        if self._scale:
+            # Don't modify the input
+            self._kernel = self._kernel * self._kernel_scaling
+
+    @staticmethod
```
adler-j: Static methods should in general be free. Only reason to have static methods is if users need to ask the class something, in general before initialization, e.g.

```python
for impl im Convolution.valid_impls():
    conv = Convolution(impl=impl)
```

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+        domain : `DiscreteLp`
+            Domain of the convolution operator.
+        kernel_kwargs : dict
+            Used as ``space.element(kernel, **kernel_kwargs)`` if
+            ``kernel`` is callable.
+
+        Returns
+        -------
+        kernel : `DiscreteLpVector`
+            Element in an adequate space.
```
adler-j: This really adds nothing

odl/trafos/convolution.py
```diff
@@ -0,0 +1,804 @@
...
+
+        return new_ker_csides, new_dom_csides
+
+    @property
+    def impl(self):
+        """Implementation of this operator."""
+        return self._impl
+
+    @property
+    def resample(self):
```
adler-j: `resampling` perhaps? `resample` sounds like a function that will do something, i.e. this feels natural

```python
conv.resample(something)
```

odl/trafos/convolution.py
adler-j: Too hacky for me, no way to solve it?

adler-j:
Some initial review done. I feel like the design (as shown in the example) looks good.

odl/trafos/convolution.py
kohr-h: It's a dependency issue in this case. Apparently we need some place for operators that can depend on *anything* in the rest of the package.

odl/trafos/convolution.py
kohr-h: Didn't put any effort into it since it's an internal helper.

odl/trafos/convolution.py
kohr-h: And the docstring says "resampling" :-)

odl/trafos/convolution.py
kohr-h: One way of doing this would be to take `domain = kernel.space` if possible. However, if `kernel` is given as a function, there is no way to deduce a reasonable domain. 
Note also that the convolution of a function with support in `[a, b]` with a kernel supported in `[-c, c]` is supported in `[a-c, b+c]`, so for small kernels roughly in the same domain. That's not true if the kernel support is not zero-centered, say `[c, d]`. Then one gets `[a+c, b+d]` with midpoint `m = (a+b)/2 + (c+d)/2`, i.e. the "main mass" of the convolved function is in `[a, b] + (c + d)/2`, the old interval shifted by the midpoint of the kernel support.

To make a long story short: I don't think it's natural to assume that input function and kernel are supported in the same spatial domain. Therefore I would keep the requirement to give both.

odl/trafos/convolution.py
kohr-h: done

odl/trafos/convolution.py
kohr-h: found a nicer solution

examples/trafos/convolution.py
adler-j: It already is

odl/trafos/convolution.py
kohr-h: Old code anyway, was just there for the `impl` thing.

odl/trafos/convolution.py
kohr-h: done

odl/trafos/convolution.py
kohr-h: Added a doctest. Otherwise, I don't see extra work here.

adler-j:
This could really use some attention since it's been lagging. What is needed to get this done?

kohr-h:
I need the resizing functionality first. Working on that.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+
+        This can be expressed as multiplication with the complex
+        conjugate kernel in Fourier space, which is used in this
+        implementation.
+        """
+        return self.__adjoint
+
+    @property
+    def inverse(self):
+        """Poor man's deconvolution using the reciprocal kernel FT."""
```
adler-j: I thought we had an agreement to not do poor mans inverses? You explicitly argued against this for the FBP as inverse, while I was in favor.

Just stating this for thoughts, personally I'm in favor of this.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+        Returns
+        -------
+        kernel_ft : element of ``kernel_ft_op.range``
+            FT of the convolution kernel. If ``out`` was given, this
+            object is a reference to it.
+        """
+        if out is None:
+            out = self.kernel_ft_op.range.element()
+
+        out.lincomb(1 / self.__kernel_scaling, self.__scaled_kernel_ft)
```
adler-j: This two step implementation is both slower and perhaps harder to follow than

```
if out is None:
    return (1 / self.__kernel_scaling) * self.__scaled_kernel_ft
else:
    out.lincomb(1 / self.__kernel_scaling, self.__scaled_kernel_ft)
```

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+        self.__kernel_ft_op = ker_ft_op
+
+        # FT range temporary
+        ft_range_tmp = kwargs.pop('ft_range_tmp', None)
+        if ft_range_tmp is not None and ft_range_tmp not in self.ft_op.range:
+            raise OpRangeError('`ft_range_tmp` not in `ft_op.range`')
+        self.__ft_range_tmp = ft_range_tmp
+
+        # Cache adjoint operator
+        self.__adjoint = FourierSpaceConvolution(
```
adler-j: Doesnt this cause a cascade of intializations where operators initialize operators who initialize operators infinitely?

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+
+        out.lincomb(1 / self.__kernel_scaling, self.__scaled_kernel_ft)
+        return out
+
+    def kernel(self, out=None):
+        """Return the real-space kernel of this convolution operator.
+
+        Parameters
+        ----------
+        out : element of ``kernel_ft_op.domain``, optional
```
adler-j: ``kernel_ft_op.domain`` is the same as `domain`, no? And also we typeset this
```
`domain` element
```

odl/trafos/convolution.py
adler-j: Where does the "reciprocal" come in the code?

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+            Keyword arguments passed to the call of the kernel function
+            if given as a callable.
+
+        See also
+        --------
+        scipy.signal.convolve : real-space convolution
+
+        Examples
+        --------
+        >>> import odl
```
adler-j: not needed anymore

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+    def adjoint(self):
+        """Adjoint operator defined by the reversed kernel."""
+        # Make space for the adjoint kernel, -S if S is the kernel space
+        adj_min_pt = -self._kernel.space.max_pt
+        adj_max_pt = -self._kernel.space.min_pt
+        adj_ker_space = uniform_discr_fromdiscr(
+            self._kernel.space,
+            min_pt=adj_min_pt, max_pt=adj_max_pt)
+
+        # Adjoint kernel is k_adj(x) = k(-x), i.e. the kernel array is
```
adler-j: This computation method only holds if the domain is centered at 0, right? Otherwise you need a more precise method.

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+    subdiscr = odl.uniform_discr_fromdiscr(discr, min_corner=[0, 0],
+                                           nsamples=[3, 3])
+    std_ker_discr = odl.uniform_discr([-1, -0.5], [1, 0.5], [5, 10])
+
+    full_ker = discr.one()
+    full_ker_arr = full_ker.asarray()
+    small_ker = subdiscr.one()
+    small_ker_arr = small_ker.asarray()
+
+    def func_ker(x):
```
adler-j: why not simply `func_ker = sum`

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+        return sum(x)
+
+    conv_full = odl.trafos.RealSpaceConvolution(discr, full_ker)
+    assert conv_full.kernel() == full_ker
+
+    conv_full_arr = odl.trafos.RealSpaceConvolution(discr, full_ker_arr)
+    assert np.array_equal(conv_full_arr.kernel(), full_ker_arr)
+    assert np.allclose(conv_full_arr.kernel().space.partition.midpoint, 0)
```
adler-j: Use the new `approx` in pytest:

```
assert conv_full_arr.kernel().space.partition.midpoint == pytest.approx([0, 0])
```

odl/discr/lp_discr.py
```diff
@@ -175,6 +175,11 @@ def max_pt(self):
         return self.partition.max_pt
 
     @property
+    def mid_pt(self):
```
adler-j: We have `partition.midpoint`, we need to standardize this one, i prefer this notation here.

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+    cdiscr = odl.uniform_discr([-1, 0], [1, 1], [5, 10], dtype=complex)
+    odl.trafos.RealSpaceConvolution(cdiscr, small_ker_arr)
+
+    with pytest.raises(TypeError):
+        odl.trafos.RealSpaceConvolution(odl.rn(5), [1, 1, 1])
+
+    with pytest.raises(ValueError):
+        fspace = odl.FunctionSpace(odl.Interval(0, 1))
+        grid = odl.TensorGrid([0, 0.4, 1])
+        part = odl.RectPartition(fspace.domain, grid)
```
adler-j: `odl.nonuniform_partition` should do the job here

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+
+    discr = odl.uniform_discr([-1, 0], [1, 1], [5, 10])
+    ker_discr = odl.uniform_discr([-1, -0.25], [1, 0.25], [20, 2])
+    kernel = ker_discr.one()
+
+    cs_up = [0.1, 0.1]
+    discr_up = odl.uniform_discr_fromdiscr(discr, cell_sides=cs_up)
+    ker_discr_up = odl.uniform_discr_fromdiscr(ker_discr, cell_sides=cs_up)
+    conv_up = odl.trafos.RealSpaceConvolution(discr, kernel, resample='up')
+    assert conv_up._domain_resampling_op.domain == discr
```
adler-j: do we test private variables like this?

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+    assert conv_up._domain_resampling_op.range == discr_up
+    assert conv_up._kernel_resampling_op.domain == ker_discr
+    assert conv_up._kernel_resampling_op.range == ker_discr_up
+
+    cs_down = [0.4, 0.25]
```
adler-j: Frankly I have no idea if this is correct or not. Could you try adding a few lines that tell us why?

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def test_real_conv_call_1d():
+
+    discr = odl.uniform_discr(0, 1, 5)
+    ker_array = np.ones(3)
+    conv = odl.trafos.RealSpaceConvolution(discr, ker_array)
+
+    def char(x):
+        return (x[0] >= 0) & (x[0] <= 0.5)
```
adler-j: no need for `x[0]`

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+def test_real_conv_call_2d():
+
+    discr = odl.uniform_discr([-1, 0], [1, 1], [5, 10])
+    ker_array = np.ones((3, 3))
+    conv = odl.trafos.RealSpaceConvolution(discr, ker_array)
+
+    def rectangle(x):
+        return (x[0] >= 0) & (x[0] <= 0.5) & (x[1] >= 0.5) & (x[1] <= 1)
+
+    func = discr.element(rectangle)
```
adler-j: perhaps simply use the cuboid phantom instead? Easier to read if nothing else.

adler-j: This applies to all of theses tests.

test/trafos/convolution_test.py
```diff
@@ -0,0 +1,292 @@
...
+
+    # Switched roles
+    discr = odl.uniform_discr(0, 1, 3)
+    scaling = discr.cell_volume
+    conv = odl.trafos.RealSpaceConvolution(discr, func_arr)
+    assert np.allclose(conv(ker_array),
+                       scaling * convolve(ker_array, func_arr, mode='same'))
+
+
+def test_real_conv_call_2d():
```
adler-j: Can't we convert this to a n-d test and make a fixture for the dimension?

kohr-h:
> Overall very good and well thought through, as usual.
> 
> The major comments I have would be on the tests. Here we only test 1d and 2d, and only with the default dspace impl and dtype (basically numpy, float64).
> 
> I would very much like a largescale test that tried this in higher dimensions (notably, 3) and that also runs it for various dtypes (float32, complex?) and backends.

Thanks for the review. I see that as an incentive to "kill it" :-)
I'll look at your comments, but regarding tests, they are ancient and I'll probably rewrite them from scratch using fixtures as you say, and put them into the right place for that matter.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+
+__all__ = ('FourierSpaceConvolution', 'RealSpaceConvolution')
+
+
+_REAL_CONV_SUPPORTED_IMPL = ('scipy_convolve',)
+# TODO: get from fourier.py
+_FOURIER_CONV_SUPPORTED_IMPL = ('numpy', 'pyfftw')
+
+
+class ConvolutionBase(Operator):
```
kohr-h: I thought in the beginning it would be more useful, but it turns out to be rather pointless. I'll remove it.

odl/trafos/convolution.py
kohr-h: Probably yes, I never tested it :-)
But the issue of overhead for the adjoint is real, so I guess I'll settle for some kind of caching instead, or the possibility to give the adjoint (or the scaled kernel FT) in the `kwargs` of `__init__`. What would you prefer?

odl/trafos/convolution.py
kohr-h: I was realizing that and just forgot to add a TODO.

odl/trafos/convolution.py
kohr-h: No, the kernel space is usually different from `domain` -- the kernel is most of the time zero-centered, while `domain` can be anything.

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+        self.kernel_ft_op.inverse(self.__scaled_kernel_ft, out=out)
+        out /= self.__kernel_scaling
+        return out
+
+    def _call(self, x, out, **kwargs):
+        """Implement ``self(x, out[, **kwargs])``.
+
+        Keyword arguments are passed on to the transform.
+        """
+        x_trafo = self.ft_op(x, out=self.__ft_range_tmp, **kwargs)
```
kohr-h: Hm, didn't I do this? It's `None` by default and the user can specify it.

kohr-h: Perhaps there should be a public property then.

odl/trafos/convolution.py
kohr-h: Hard questions...
Perhaps we should bounce this question with more people - @ozanoktem, @aringh, @hilding79. I actually don't remember being consistently against it back then, for sure in the beginning, but now after some implementations of obvious "kind-of-inverses" I can see the benefit, in particular when there is an obvious choice.

test/trafos/convolution_test.py
kohr-h: This doesn't even exist anymore.

test/trafos/convolution_test.py
kohr-h: Obsolete

odl/trafos/convolution.py
adler-j: I prefer passing the adjoint as an argument in the kwargs.

examples/trafos/convolution.py
```diff
@@ -0,0 +1,90 @@
...
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import odl
+
+
+# Parameters for the gaussian kernel
+gamma = 0.1
+norm_const = 2 * np.pi * gamma ** 2
```
chongchenmath: Maybe using ``sigma`` is more common?

odl/trafos/convolution.py
```diff
@@ -0,0 +1,849 @@
...
+
+class FourierSpaceConvolution(ConvolutionBase):
+
+    """Convolution implemented in Fourier space.
+
+    This operator implements a discrete approximation to the continuous
+    convolution with a fixed kernel. It is based on the Fourier
+    transform and usually performs better than `RealSpaceConvolution`,
+    except for small kernels.
+    """
```
chongchenmath: Could you please provide some simple formula to explain this?

odl/trafos/convolution.py
kohr-h: Hm, good point. I need to think it through. My guess is that it's still correct, but the kernel domain needs to be flipped, too (which is a no-op if it's symmetric around 0).

adler-j:
Well. No too be that guy, but how are we doing?

ozanoktem:
Well, the road to Golgota is long and slow ... This is definitely Holger's cross to carry. I wonder what the odds are that it will be done before the end of the year, I hope they are low 👍 
BTW: The beast landed today, i.e. the GPU-server has arrived. This proves that all processes have a termination :-)

adler-j:
>  I hope they are low 

High?

> This proves that all processes have a termination :-)

[https://en.wikipedia.org/wiki/Halting_problem](https://en.wikipedia.org/wiki/Halting_problem) 😉

ozanoktem:
low odds = high likelihood of success, aka there is a generic convolution operator in place before end of 2016
high odds = low likelihood of success, aka there is no generic convolution operator in place before end of 2016

kohr-h:
It's NP-hard to decide when to do this..

adler-j:
Bump for justice. This really has to get in at some point rather soon.

adler-j:
Bumping this again. Is there anything I can do to push this through?

kohr-h:
> Bumping this again. Is there anything I can do to push this through?

Bad story this. I'm currently really inclined towards just pushing in the real space convolution since that's probably ready to go, and perhaps just some restricted version of the Fourier space convolution. But on that side there's still some work to do to get all the things done that I expect from this (different resolutions, non-centered kernels etc.). Unfortunately this also depends on #735 to some degree.

kohr-h:
Closing this in favor of a new attempt.

PR #475, opened 2016-07-01
REL: add release script

kohr-h:
Preliminary PR for a release script for simplification. Never tested it :-)

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,91 @@
+#!/bin/sh
+
+function print_usage_and_quit {
+    echo "Usage: $0 <version> [--dry-run]"
+    exit 1
+}
+
+# Process command line args
+if [ "$#" -eq 0 ]; then
+    print_usage_and_quit
+fi
+
+set +x
+
+PACKAGE_NAME="odl"
+
+# Clumsy parsing of command line arguments
+if [ "$#" == 1 ]; then
+    if [ "$1" == "--dry-run" ]; then
+        print_usage_and_quit
+    fi
+    DRY_RUN=0
+    VERSION=$1
+elif [ "$#" == 2 ]; then
+    if [ "$1" == "--dry-run" ]; then
+        VERSION=$2
+    elif [ "$2" == "--dry-run" ]; then
+        VERSION=$1
+    else
+        print_usage_and_quit
+    fi
+
+    DRY_RUN=1
+else
+    print_usage_and_quit
+fi
+
+
+### PyPI upload
+
+PYPI_USER=odlgroup
+TWINE=$(which twine)
+if [ -z $TWINE ]; then
+    echo 'Error: twine was not found. Please install it (e.g. via "pip install twine)".'
+    exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


adler-j:
Did some changes so it gets abit further, but usage is still un-clear. What is the `dist` folder? How is it created?

kohr-h:
> Did some changes so it gets abit further, but usage is still un-clear. What is the dist folder? How is it created?

`dist` is created when you run `python setup.py sdist` or `python setup.py bdist_wheel`. It's the default `TARGET` of the distribution creation command. For wheels, you can also run `pip wheel`, but [according to its documentation](https://pip.pypa.io/en/stable/reference/pip_wheel/#build-system-interface), it doesn't do anything more than the above command.

We could make the above commands part of the script, but we may want to test installing the distribution locally before uploading, so I'd rather not include it. Perhaps a comment in the beginning could clarify. Or one could check
```bash
if [ ! -d "dist" ]; then
    echo "The `dist` directory does not exist."
    echo "You need to run `python setup.py sdist` and `python setup.py bdist_wheel` to build the distribution."
    exit 1
fi
```

adler-j:
Adding a proper check with a recommendation would be nice and sufficient.

kohr-h:
Made some amendments, should be more informative now.

adler-j:
Well we're getting further, but:

```bash
$:~/Programming/ODL$ sh tools/release.sh 0.3.0
tools/release.sh: 64: [: odl-0.3.0-py2.py3-none-any.whl: unexpected operator
ValueError: Cannot find file (or expand pattern): 'odl-0.3.0-py2.py3-none-any.whl'
```

adler-j:
If you do work on this, could you rebase onto master as well?

adler-j:
rebased master on this

kohr-h:
Yeah, missed that during vacation :-)
Did you try to run it since the last commit @adler-j ?

adler-j:
I did some fixes and this now runs, as evidenced by the new version available on pip. Users now need to set the environment variable "PYPI_PASSWORD" before running, but I guess that's fine.

kohr-h:
> Users now need to set the environment variable "PYPI_PASSWORD" before running, but I guess that's fine.

That shouldn't be necessary, and I'm reluctant to say it's fine. Having passwords in the environment is not a good idea IMO. `twine` should ask for the password.

kohr-h:
Did some further improvements. The script will ask for the passwords now, which I think is the far better solution (copy-paste from password manager). It will also just exit if there are no anaconda files or the anaconda executable is missing.
I just used it to upload the 0.3.1 release, seems to work fine, so I suggest we merge and make minor adjustments along the way.

adler-j:
Merged. Well done.

kohr-h:
Thanks :-). I've added a [running section in the release notes](https://odl.readthedocs.io/release_notes.html#next-release) - would you add a line now that you merged? I forgot after the last release, but it's quite some work to compile all PRs for a new release, so if everybody who merges a PR adds a line to the release notes, that will greatly speed up the release process (and the stuff is still fresh, no guessing "WTF  was this about?").

PR #476, opened 2016-07-01
Improvements to DiscreteLp.element

adler-j:
* Allow non-vectorized input by exposing the `vectorized` parameter.
* Allow scalar functions (e.g. `lambda x: 1.0`) as input to `DiscreteLp.element`.
* Also removed the old check in `DiscreteLp.element` that disallowed specifically the above (but only in 1d).

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -234,13 +234,13 @@ class DiscreteLp(DiscretizedSpace):
         Parameters
         ----------
         inp : optional
-            The input data to create an element from.
+            Input data to create an element from.
 
             If `callable`, it needs to be understood by the ``uspace.element``
             method.
 
-            Otherwise, it has to be understandable by the
-            ``dspace.element`` method.
+            Otherwise, it has to be understood by the ``dspace.element``
+            method.
         kwargs :
             Additional arguments passed on to `sampling` when called
             on ``inp``, in the form ``sampling(inp, **kwargs)``.
@@ -249,8 +249,10 @@ class DiscreteLp(DiscretizedSpace):
         Other Parameters
         ----------------
         vectorized : `bool`
-            Can only be used if ``inp`` is `callable`, in that case, indicates
-            if ``inp`` is vectorized.
+            Can only be used if ``inp`` is `callable`, in which case it
+            indicates if ``inp`` is vectorized. If not, it will be wrapped
+            with a vectorizer.
+            Default: True
 
         Returns
         -------
@@ -274,10 +276,7 @@ class DiscreteLp(DiscretizedSpace):
         On the other hand, non-discretized objects like Python functions
         can be discretized "on the fly":
 
-        >>> def f(x):
-        ...     return x * 2
-        ...
-        >>> space.element(f)
+        >>> space.element(lambda x: x * 2)
         uniform_discr(-1.0, 1.0, 4).element([-1.5, -0.5, 0.5, 1.5])
 
         This works also with parameterized functions, however only
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -234,13 +234,23 @@ class DiscreteLp(DiscretizedSpace):
         Parameters
         ----------
         inp : optional
-            The input data to create an element from. It needs to be
-            understood by either the `sampling` operator of this
-            instance or by its ``dspace.element`` method.
+            The input data to create an element from.
+
+            If `callable`, it needs to be understood by the ``uspace.element``
+            method.
+
+            Otherwise, it has to be understandable by the
+            ``dspace.element`` method.
         kwargs :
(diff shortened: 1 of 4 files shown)


odl/space/fspace.py
```diff
@@ -379,14 +395,6 @@ def __call__(self, x, out=None, **kwargs):
             if ndim == 1:
                 try:
                     out = self._call(x, **kwargs)
-                    if np.ndim(out) == 0 and not scalar_out:
-                        # Don't accept scalar result. A typical situation where
-                        # this occurs is with comparison operators, e.g.
-                        # "return x > 0" which simply gives 'True' for a
-                        # non-empty tuple (in Python 2). We raise TypeError
-                        # to trigger the call with x[0].
```
kohr-h: Is the `x > 0` issue unaffected?

adler-j: It wasn't even tested for, but i.m.o. we shouldnt interfere with user code. If the user wants to do that test, they should.

Silently failing like this will only be problematic (and stops scalar output from beeing valid, as shown when @ozanoktem tried making a function like `lambda x : 1.0`).

adler-j:
bump for review

odl/discr/lp_discr.py
```diff
@@ -234,14 +234,24 @@ def element(self, inp=None, **kwargs):
         Parameters
         ----------
         inp : optional
-            The input data to create an element from. It needs to be
-            understood by either the `sampling` operator of this
-            instance or by its ``dspace.element`` method.
+            The input data to create an element from.
```
kohr-h: ~~The~~

odl/discr/lp_discr.py
```diff
@@ -234,14 +234,24 @@ def element(self, inp=None, **kwargs):
...
+
+            If `callable`, it needs to be understood by the ``uspace.element``
+            method.
+
+            Otherwise, it has to be understandable by the
```
kohr-h: understood

odl/discr/lp_discr.py
```diff
@@ -234,14 +234,24 @@ def element(self, inp=None, **kwargs):
...
+            ``dspace.element`` method.
         kwargs :
             Additional arguments passed on to `sampling` when called
             on ``inp``, in the form ``sampling(inp, **kwargs)``.
             This can be used e.g. for functions with parameters.
 
+        Other Parameters
+        ----------------
+        vectorized : `bool`
+            Can only be used if ``inp`` is `callable`, in that case, indicates
```
kohr-h: , in which case it indicates

odl/discr/lp_discr.py
```diff
@@ -293,8 +303,9 @@ def element(self, inp=None, **kwargs):
         elif inp in self.dspace:
             return self.element_type(self, inp)
         elif callable(inp):
+            vectorized = kwargs.pop('vectorized', True)
             # uspace element -> discretize
-            inp_elem = self.uspace.element(inp)
+            inp_elem = self.uspace.element(inp, vectorized=vectorized)
```
kohr-h: At some point, we need to rethink the practice of just "handing down" `kwargs`, especially in cases as this one where there are multiple possible places where to put the arguments.
A viable solution to that issue is to use "named" (`args` and) `kwargs`, e.g. `interp_kwargs` in the interpolation (just an example).

Here it's still somewhat okay but on the boundary to bug-prone. And increasingly tricky to explain in the docs. If there are more things like this we should consider being more specific.

odl/space/fspace.py
kohr-h: Then it would be good to have an FAQ entry that describes the issue with the 1D function `lambda x: x > 0` returning all ones in Python 2.

kohr-h:
Done. Some stuff to think about, else good.

odl/space/fspace.py
adler-j: It doesnt

```python
>>> X = odl.uniform_discr(-1, 1, 5)
>>> X.element(lambda x: x>0)
uniform_discr(-1.0, 1.0, 5).element([0.0, 0.0, 0.0, 1.0, 1.0])
```

adler-j: Update:

It seems we return a 2d array for the 1d case (instead of a tuple as in the higher dimensional cases), this cheeses this problem since the above evaluates as expected.

kohr-h: Return from where? The `meshgrid`?

adler-j: Not quite sure to be frank, this piece of code is a rather deep rabbit hole.

Apparently it is [here](https://github.com/odlgroup/odl/blob/60e2af8b790877dfe7fc04d2565d24e48eb88655/odl/space/fspace.py#L363-L365).

kohr-h: That looks indeed like the culprit. Maybe one needs to squeeze before returning. But if it works, better not touch it :-)

PR #479, opened 2016-07-04
MAINT: minor performance improvements to hot paths

adler-j:
Improves runtime for `py.test` from 7.1 seconds to 6.5 seconds on my machine (with CUDA).

No functional changes


kohr-h:
ok

PR #480, opened 2016-07-05
Forward backward primal dual

aringh:
Adds the forward-backward primal-dual solver. Reference article is found [here](http://www.mat.univie.ac.at/~rabot/publications/jour15-01.pdf) (algorithm is found in Theorem 2).

### DIFF
--- a/examples/solvers/forward_backward_primal_dual_denoising.py
+++ b/examples/solvers/forward_backward_primal_dual_denoising.py
@@ -1,0 +1,69 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Total variation denoising using the Forward-backward solver.
+
+Solves the optimization problem
+
+    min_{0 <= x <= 255}  1/2 ||x - g||_2^2 + lam || |grad(x)| ||_1
+
+where ``grad`` is the spatial gradient and ``g`` the given noisy data.
+"""
+
+import numpy as np
+import scipy.misc
+import odl
+import odl.solvers as odlsol
+
+# Parameters
+n = 256
+
+# Create a space
+space = odl.uniform_discr([0, 0], [n, n], [n, n])
+
+# Load image and noise
+data = space.element(np.rot90(scipy.misc.ascent()[::2, ::2], 3))
+noise = odl.phantom.white_noise(space) * 10.0
+
+# Create noisy data
+noisy_data = data + noise
+data.show('Original data')
+noisy_data.show('Noisy convolved data')
+
+
+# Gradient for TV regularization
+gradient = odl.Gradient(space)
+
+# Assemble all operators
+lin_ops = [gradient]
+
+# Create proximals as needed
+prox_cc_g = [odlsol.proximal_cconj_l1(gradient.range, lam=1e1,
+                                      isotropic=False)]
+prox_f = odlsol.proximal_box_constraint(space, 0, 255)
+
+# Create gradient needed. This is the derivative of the 2-norm squared
+grad_h = odl.ResidualOperator(odl.IdentityOperator(space), noisy_data)
+
+# Solve
+x = noisy_data.copy()
+
+callback = (odl.solvers.CallbackShow(display_step=20, clim=[0, 255]) &
+            odl.solvers.CallbackPrintIteration())
+
+odlsol.forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
+                           sigma=[0.01], niter=1000, callback=callback)
--- a/examples/solvers/forward_backward_primal_dual_denoising.py
+++ b/examples/solvers/forward_backward_primal_dual_denoising.py
@@ -27,41 +27,41 @@ where ``grad`` is the spatial gradient and ``g`` the given noisy data.
 import numpy as np
 import scipy.misc
 import odl
-import odl.solvers as odlsol
 
-# Parameters
-n = 256
+# Load image
+image = np.rot90(scipy.misc.ascent()[::2, ::2], 3)
+
+# Reading the size
+n = image.shape[0]
+m = image.shape[1]
 
 # Create a space
-space = odl.uniform_discr([0, 0], [n, n], [n, n])
+space = odl.uniform_discr([0, 0], [n, m], [n, m])
 
-# Load image and noise
-data = space.element(np.rot90(scipy.misc.ascent()[::2, ::2], 3))
+# Create data, noise and noisy data
+data = space.element(image)
 noise = odl.phantom.white_noise(space) * 10.0
-
-# Create noisy data
 noisy_data = data + noise
 data.show('Original data')
-noisy_data.show('Noisy convolved data')
-
+noisy_data.show('Noisy data')
 
 # Gradient for TV regularization
 gradient = odl.Gradient(space)
 
-# Assemble the linear operators. Here the TV-term is mapped into a composition
-# of the 1-norm and the gradient. See the documentation of the solver
-# `forward_backward_pd` for the general form of the problem.
+# Assemble the linear operators. Here the TV-term is represented as a
+# composition of the 1-norm and the gradient. See the documentation of the
+# solver `forward_backward_pd` for the general form of the problem.
 lin_ops = [gradient]
 
-# Create proximals for the convex conjugate of the 1-norm and the bound
+# Create proximals for the convex conjugates of the 1-norm and the bound
 # constrains.
-prox_cc_g = [odlsol.proximal_cconj_l1(gradient.range, lam=1e1,
-                                      isotropic=False)]
-prox_f = odlsol.proximal_box_constraint(space, 0, 255)
+prox_cc_g = [odl.solvers.proximal_cconj_l1(gradient.range, lam=1e1,
+                                           isotropic=False)]
+prox_f = odl.solvers.proximal_box_constraint(space, 0, 255)
 
 # This gradient encodes the differentiable term(s) of the goal functional,
 # which corresponds to the "forward" part of the method. In this example the
-# differentiable part is the 2-norm squared.
+# differentiable part is the squared 2-norm.
 grad_h = odl.ResidualOperator(odl.IdentityOperator(space), noisy_data)
 
 # Create initial guess for the solver.
@@ -71,6 +71,6 @@ x = noisy_data.copy()
 callback = (odl.solvers.CallbackShow(display_step=20, clim=[0, 255]) &
             odl.solvers.CallbackPrintIteration())
 
-# Call the solver
-odlsol.forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
-                           sigma=[0.01], niter=1000, callback=callback)
+# Call the solver. x is updated in-place with the consecutive iterates.
+odl.solvers.forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
+                                sigma=[0.01], niter=1000, callback=callback)
--- a/examples/solvers/forward_backward_primal_dual_denoising.py
+++ b/examples/solvers/forward_backward_primal_dual_denoising.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Total variation denoising using the Forward-backward solver.
+"""Total variation denoising using the forward-backward primal dual solver.
 
 Solves the optimization problem
 
@@ -48,22 +48,29 @@ noisy_data.show('Noisy convolved data')
 # Gradient for TV regularization
 gradient = odl.Gradient(space)
 
-# Assemble all operators
+# Assemble the linear operators. Here the TV-term is mapped into a composition
+# of the 1-norm and the gradient. See the documentation of the solver
+# `forward_backward_pd` for the general form of the problem.
 lin_ops = [gradient]
 
-# Create proximals as needed
+# Create proximals for the convex conjugate of the 1-norm and the bound
+# constrains.
 prox_cc_g = [odlsol.proximal_cconj_l1(gradient.range, lam=1e1,
                                       isotropic=False)]
 prox_f = odlsol.proximal_box_constraint(space, 0, 255)
 
-# Create gradient needed. This is the derivative of the 2-norm squared
+# This gradient encodes the differentiable term(s) of the goal functional,
+# which corresponds to the "forward" part of the method. In this example the
+# differentiable part is the 2-norm squared.
 grad_h = odl.ResidualOperator(odl.IdentityOperator(space), noisy_data)
 
-# Solve
+# Create initial guess for the solver.
 x = noisy_data.copy()
 
+# Used to display intermediate results and print iteration number.
 callback = (odl.solvers.CallbackShow(display_step=20, clim=[0, 255]) &
             odl.solvers.CallbackPrintIteration())
 
+# Call the solver
 odlsol.forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
                            sigma=[0.01], niter=1000, callback=callback)
--- a/examples/solvers/forward_backward_primal_dual_denoising.py
+++ b/examples/solvers/forward_backward_primal_dual_denoising.py
@@ -32,8 +32,7 @@ import odl
 image = np.rot90(scipy.misc.ascent()[::2, ::2], 3)
 
 # Reading the size
-n = image.shape[0]
-m = image.shape[1]
+n, m = image.shape
 
 # Create a space
 space = odl.uniform_discr([0, 0], [n, m], [n, m])
--- a/odl/solvers/advanced/forward_backward.py
+++ b/odl/solvers/advanced/forward_backward.py
@@ -1,0 +1,165 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Implementation of the Forward-Backward splitting algorithm """
+
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+from odl.operator import Operator
+
+
+__all__ = ('forward_backward_pd',)
+
+
+def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
+                        callback=None, **kwargs):
+    """ The forward-backward primal-dual splitting algorithm
+
+    Minimizes the sum of several convex functions composed with linear
+    operators
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
+
+    Where f, g_i, h are convex functions, L_i are linear `Operator`'s, and h is
+    also differentiable.
+
+    Can also be used to solve the more general problem
+
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``
+
+    where l_i are convex functions and @ is the infimal convolution:
+
+        ``(f @ g)(x) = inf_y { f(x-y) + g(y) }``
+
+    Parameters
+    ----------
+    x : `Vector`
+        Initial point
+    prox_f : `callable`
+        Funciton returning an `Operator` when called with `tau`.
+        The Operator should be the proximal of `tau * f`.
+    prox_cc_g : `sequence` of `callable`'s
+        Sequence of functions returning an operator when called with stepsize
+        `sigma`.
+        The Operator should be the proximal of `sigma_i * g_i^*`.
+    grad_h : `Operator`
+        Operators representing the gradient of  `h`.
+    L : `sequence` of `Operator`'s
+        A sequence with as many elements as ``prox__cc_gs`` of operators
+        ``L_i``
+    tau : `float`
+        Stepsize of ``f``
+    sigma : `sequence` of  `float`
+        Stepsize of the ``g_i``'s
+    niter : `int`
+        Number of iterations
+    callback : `Callback`, optional
+        Show partial results
+
+    Other Parameters
+    ----------------
+    grad_cc_l : `sequence` of `Operator`'s, optional
+        Sequence of operators representing the gradient of  `l_i^*`.
+        If omitted, the simpler problem will be considered, which corresponds
+        to the convex conjugate functionals of the l_i:s being zero functionals
+        and hence the gradient being the zero-operator.
+
+    Notes
+    -----
+    Strictly, we have the following conditions on the functions involved: f and
+    g_i are proper, convex and lower semicontinuous, and h is convex and
+    differentialbe with :math:`\\eta`-Lipschitz continuous gradient.
+
+    For the optional input l_i we need it to be proper, convex, lower
+    semicontinuous, and :math:`\\nu_i^{-1}`-strongly convex. The fact that l_i
+    is :math:`\\nu_i^{-1}`-strongly convex implies that the convex conjugate
+    functional of l_i is differentiable with :math:`\\nu_i`-Lipschitz
+    continuous gradient.
+
+    To guarantee convergence, the parameters ``tau``, ``sigma`` and
+    ``L`` need to satisfy
+
+    .. math::
+
+       2 * \min \{ \\frac{1}{\\tau}, \\frac{1}{\sigma_1}, \\ldots,
+       \\frac{1}{\sigma_m} \} * \min\{ \\eta, \\nu_1, \\ldots, \\nu_m  \} *
+       \\sqrt{1 - \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2} > 1
+
+    where, if the simpler problem is considered, all :math:`\\nu_i` can be
+    considered to be :math:`\\infty`.
+
+    For references on the Forward-Backward algorithm, see [BC2015]_.
+
+    For more on convex analysis including convex conjugates and
+    resolvent operators see [Roc1970]_.
+
+    For more on proximal operators and algorithms see [PB2014]_.
+    """
+
+    # Problem size
+    m = len(L)
+
+    # Validate input
+    if not all(isinstance(op, Operator) for op in L):
+        raise ValueError('`L` not a sequence of operators')
+    if not all(op.is_linear for op in L):
+        raise ValueError('not all operators in `L` are linear')
+    if not all(x in op.domain for op in L):
+        raise ValueError('`x` not in the domain of all operators')
+    if len(sigma) != m:
+        raise ValueError('len(sigma) != len(L)')
+    if len(prox_cc_g) != m:
+        raise ValueError('len(prox_cc_g) != len(L)')
+
+    # Get parameters from kwargs
+    grad_cc_l = kwargs.pop('prox_cc_l', None)
+    if grad_cc_l is not None and len(grad_cc_l) != m:
+        raise ValueError('`grad_cc_l` not same length as `L`')
+
+    # Check for unused parameters
+    if kwargs:
+        raise TypeError('unexpected keyword argument: {}'.format(kwargs))
+
+    # Pre-allocate values
+    v = [Li.range.zero() for Li in L]
+    y = x.space.zero()
+
+    # Iteratively solve
+    for k in range(niter):
+        x_old = x
+
+        tmp_1 = grad_h(x) + sum(Li.adjoint(vi) for Li, vi in zip(L, v))
+        prox_f(tau)(x - tau * tmp_1, out=x)
+        y.lincomb(2.0, x, -1, x_old)
+
+        for i in range(m):
+            # In this case gradients were given
+            if grad_cc_l is not None:
+                tmp_2 = sigma[i] * (L[i](y) + grad_cc_l[i](v[i]))
+
+            # In this case there were not. "Applying" zero-operator
+            else:
+                tmp_2 = sigma[i] * L[i](y)
+
+            v[i] = prox_cc_g[i](sigma[i])(v[i] + tmp_2)
+
+        if callback is not None:
+            callback(x)
--- a/odl/solvers/advanced/forward_backward.py
+++ b/odl/solvers/advanced/forward_backward.py
@@ -50,7 +50,7 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
 
         ``(g @ l)(x) = inf_y { g(y) + l(x-y) }``.
 
-    Note that the strong convetiy of ``l_i`` makes the convex conjugate
+    Note that the strong convexity of ``l_i`` makes the convex conjugate
     ``l_i^*`` differentialbe; see the Notes section for more information on
     this.
 
@@ -61,13 +61,13 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
     prox_f : `callable`
         `Proximal factory` for the functional ``f``.
     prox_cc_g : `sequence` of `callable`'s
-        Sequence of `proximal factories` for the convex conjuates of the
+        Sequence of `proximal factorie`'s for the convex conjuates of the
         functionals ``g_i``.
     grad_h : `Operator`
         Operator representing the gradient of  ``h``.
     L : `sequence` of `Operator`
-        Sequence with as many elements as ``prox_cc_gs`` of linear operators
-        ``L_i``.
+        Sequence of linear operators ``L_i``, with as many elements as
+        ``prox_cc_gs``.
     tau : `float`
         Step size-like parameter for ``prox_f``.
     sigma : `sequence` of  `float`
@@ -100,7 +100,7 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
 
     The exact conditions on the involved functionals are as follows: :math:`f`
     and :math:`g_i` are proper, convex and lower semicontinuous, and :math:`h`
-    is convex and differentialbe with :math:`\\eta^{-1}`-Lipschitz continuous
+    is convex and differentiable with :math:`\\eta^{-1}`-Lipschitz continuous
     gradient, :math:`\\eta > 0`.
 
     The optional operators :math:`\\nabla l_i^*` need to be
--- a/odl/solvers/advanced/forward_backward.py
+++ b/odl/solvers/advanced/forward_backward.py
@@ -15,7 +15,9 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-""" Implementation of the Forward-Backward splitting algorithm """
+""" Implementation of the forward-backward primal-dual splitting algorithm for
+optimization.
+"""
 
 
 # Imports for common Python 2/3 codebase
@@ -31,44 +33,42 @@ __all__ = ('forward_backward_pd',)
 
 def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
                         callback=None, **kwargs):
-    """ The forward-backward primal-dual splitting algorithm
+    """ The forward-backward primal-dual splitting algorithm.
 
-    Minimizes the sum of several convex functions composed with linear
-    operators
+    The algorithm minimizes the sum of several convex functionals composed with
+    linear operators,
 
-        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``,
 
-    Where f, g_i, h are convex functions, L_i are linear `Operator`'s, and h is
-    also differentiable.
+    where ``f``, ``g_i`` are convex functionals, ``L_i`` are linear
+    `Operator`'s, and ``h`` is a convex and differentiable functional.
 
-    Can also be used to solve the more general problem
+    The method can also be used to solve the more general problem
 
-        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``,
 
-    where l_i are convex functions and @ is the infimal convolution:
+    where ``l_i`` are convex functionals and @ is the infimal convolution:
 
         ``(f @ g)(x) = inf_y { f(x-y) + g(y) }``
 
     Parameters
     ----------
-    x : `Vector`
-        Initial point
+    x : `LinearSpaceVector`
+        Initial point, updated in place.
     prox_f : `callable`
-        Funciton returning an `Operator` when called with `tau`.
-        The Operator should be the proximal of `tau * f`.
+        `Proximal factory` for the functional ``f``.
     prox_cc_g : `sequence` of `callable`'s
-        Sequence of functions returning an operator when called with stepsize
-        `sigma`.
-        The Operator should be the proximal of `sigma_i * g_i^*`.
+        Sequence of `proximal factories` for the convex conjuates of the
+        functionals ``g_i``.
     grad_h : `Operator`
         Operators representing the gradient of  `h`.
     L : `sequence` of `Operator`'s
         A sequence with as many elements as ``prox__cc_gs`` of operators
         ``L_i``
     tau : `float`
-        Stepsize of ``f``
+        Step size-like parameter for ``prox_f``
     sigma : `sequence` of  `float`
-        Stepsize of the ``g_i``'s
+        Sequence of step size-like parameter for the sequence ``prox_cc_g``
     niter : `int`
         Number of iterations
     callback : `Callback`, optional
@@ -78,35 +78,40 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
     ----------------
     grad_cc_l : `sequence` of `Operator`'s, optional
         Sequence of operators representing the gradient of  `l_i^*`.
-        If omitted, the simpler problem will be considered, which corresponds
-        to the convex conjugate functionals of the l_i:s being zero functionals
-        and hence the gradient being the zero-operator.
+        If omitted, the simpler problem will be considered.
 
     Notes
     -----
-    Strictly, we have the following conditions on the functions involved: f and
-    g_i are proper, convex and lower semicontinuous, and h is convex and
-    differentialbe with :math:`\\eta`-Lipschitz continuous gradient.
+    The exact conditions on the involved functionals are as follows: :math:`f`
+    and :math:`g_i` are proper, convex and lower semicontinuous, and :math:`h`
+    is convex and differentialbe with :math:`\\eta`-Lipschitz continuous
+    gradient.
 
-    For the optional input l_i we need it to be proper, convex, lower
-    semicontinuous, and :math:`\\nu_i^{-1}`-strongly convex. The fact that l_i
-    is :math:`\\nu_i^{-1}`-strongly convex implies that the convex conjugate
-    functional of l_i is differentiable with :math:`\\nu_i`-Lipschitz
-    continuous gradient.
+    The optional input :math:`\\nabla l_i^*` need to be :math:`\\nu_i`
+    -Lipschitz continuous. Note that in the reference [BC2015]_, the condition
+    is formulated as :math:`l_i` being proper, lower semicontinuous, and
+    :math:`\\nu_i^{-1}`-strongly convex, which implies that :math:`l_i^*` have
+    :math:`\\nu_i`-Lipschitz continuous gradient.
+
+    If the optional input :math:`\\nabla l_i^*` is omitted, the simpler problem
+    will be considered. Mathematically, this is done by taking :math:`l_i` to
+    be the functional that is zero only in the zero element and :math:`\\infty`
+    otherwise. This gives that :math:`l_i^*` is the zero functional, and hence
+    the corresponding gradient is the zero operator.
 
     To guarantee convergence, the parameters ``tau``, ``sigma`` and
     ``L`` need to satisfy
 
     .. math::
 
-       2 * \min \{ \\frac{1}{\\tau}, \\frac{1}{\sigma_1}, \\ldots,
-       \\frac{1}{\sigma_m} \} * \min\{ \\eta, \\nu_1, \\ldots, \\nu_m  \} *
-       \\sqrt{1 - \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2} > 1
+       2 \min \{ \\frac{1}{\\tau}, \\frac{1}{\sigma_1}, \\ldots,
+       \\frac{1}{\sigma_m} \} \cdot \min\{ \\eta, \\nu_1, \\ldots, \\nu_m  \}
+       \cdot \\sqrt{1 - \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2} > 1,
 
     where, if the simpler problem is considered, all :math:`\\nu_i` can be
     considered to be :math:`\\infty`.
 
-    For references on the Forward-Backward algorithm, see [BC2015]_.
+    For reference on the forward-backward primal-dual algorithm, see [BC2015]_.
 
     For more on convex analysis including convex conjugates and
     resolvent operators see [Roc1970]_.
@@ -129,12 +134,10 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
     if len(prox_cc_g) != m:
         raise ValueError('len(prox_cc_g) != len(L)')
 
-    # Get parameters from kwargs
-    grad_cc_l = kwargs.pop('prox_cc_l', None)
+    grad_cc_l = kwargs.pop('grad_cc_l', None)
     if grad_cc_l is not None and len(grad_cc_l) != m:
         raise ValueError('`grad_cc_l` not same length as `L`')
 
-    # Check for unused parameters
     if kwargs:
         raise TypeError('unexpected keyword argument: {}'.format(kwargs))
 
@@ -151,15 +154,15 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
         y.lincomb(2.0, x, -1, x_old)
 
         for i in range(m):
-            # In this case gradients were given
             if grad_cc_l is not None:
+                # In this case gradients were given.
                 tmp_2 = sigma[i] * (L[i](y) + grad_cc_l[i](v[i]))
-
-            # In this case there were not. "Applying" zero-operator
             else:
+                # In this case gradients were not given. Therefore the gradient
+                # step is omitted. For more details, see the documentation.
                 tmp_2 = sigma[i] * L[i](y)
 
-            v[i] = prox_cc_g[i](sigma[i])(v[i] + tmp_2)
+            prox_cc_g[i](sigma[i])(v[i] + tmp_2, out=v[i])
 
         if callback is not None:
             callback(x)
--- a/odl/solvers/advanced/forward_backward.py
+++ b/odl/solvers/advanced/forward_backward.py
@@ -43,11 +43,16 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
 
     The method can also be used to solve the more general problem
 
-        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``,
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x) + h(x)``,
 
-    where ``l_i`` are convex functionals and @ is the infimal convolution:
+    where ``l_i`` are strongly convex functionals and @ is the infimal
+    convolution:
 
-        ``(f @ g)(x) = inf_y { f(y) + g(x-y) }``.
+        ``(g @ l)(x) = inf_y { g(y) + l(x-y) }``.
+
+    Note that the strong convetiy of ``l_i`` makes the convex conjugate
+    ``l_i^*`` differentialbe; see the Notes section for more information on
+    this.
 
     Parameters
     ----------
@@ -61,7 +66,7 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
     grad_h : `Operator`
         Operator representing the gradient of  ``h``.
     L : `sequence` of `Operator`
-        A sequence with as many elements as ``prox__cc_gs`` of linear operators
+        Sequence with as many elements as ``prox_cc_gs`` of linear operators
         ``L_i``.
     tau : `float`
         Step size-like parameter for ``prox_f``.
@@ -69,8 +74,8 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
         Sequence of step size-like parameters for the sequence ``prox_cc_g``.
     niter : `int`
         Number of iterations.
-    callback : `Callback`, optional
-        Show partial results.
+    callback : `callable`, optional
+        Function called with the current iterate after each iteration.
 
     Other Parameters
     ----------------
@@ -80,23 +85,36 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
 
     Notes
     -----
+    The mathematical problem to solve is
+
+     .. math::
+
+        \min_x f(x) + \sum_{i=0}^n (g_i \Box l_i)(L_i x) + h(x),
+
+     where :math:`L_i` are linear functionals and the infimal convolution
+     :math:`g \Box l` is defined by
+
+     .. math::
+
+        (g \Box l)(x) = \inf_y g(y) + l(x - y).
+
     The exact conditions on the involved functionals are as follows: :math:`f`
     and :math:`g_i` are proper, convex and lower semicontinuous, and :math:`h`
     is convex and differentialbe with :math:`\\eta^{-1}`-Lipschitz continuous
     gradient, :math:`\\eta > 0`.
 
     The optional operators :math:`\\nabla l_i^*` need to be
-    :math:`\\nu_i^{-1}`-Lipschitz continuous. Note that in the reference
+    :math:`\\nu_i`-Lipschitz continuous. Note that in the reference
     [BC2015]_, the condition is formulated as :math:`l_i` being proper, lower
     semicontinuous, and :math:`\\nu_i^{-1}`-strongly convex, which implies that
     :math:`l_i^*` have :math:`\\nu_i`-Lipschitz continuous gradients.
 
     If the optional operators :math:`\\nabla l_i^*` are omitted, the simpler
-    problem without :mat:`l_i` will be considered. Mathematically, this is done
-    by taking :math:`l_i` to be the functionals that are zero only in the zero
-    element and :math:`\\infty` otherwise. This gives that :math:`l_i^*` are
-    the zero functionals, and hence the corresponding gradients are the zero
-    operators.
+    problem without :math:`l_i` will be considered. Mathematically, this is
+    done by taking :math:`l_i` to be the functionals that are zero only in the
+    zero element and :math:`\\infty` otherwise. This gives that :math:`l_i^*`
+    are the zero functionals, and hence the corresponding gradients are the
+    zero operators.
 
     To guarantee convergence, the parameters ``tau``, ``sigma`` and
     ``L`` need to satisfy
@@ -109,13 +127,6 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
 
     where, if the simpler problem is considered, all :math:`\\nu_i` can be
     considered to be :math:`\\infty`.
-
-
-
-    Note here that if the functional
-    :math:`h` is taken as the zero functional, its gradient will be the zero
-    operator and thus :math:`\\eta = 0`. Hence the algorithm CANNOT be
-    guaranteed to converge in this case.
 
     For reference on the forward-backward primal-dual algorithm, see [BC2015]_.
 
--- a/odl/solvers/advanced/forward_backward.py
+++ b/odl/solvers/advanced/forward_backward.py
@@ -15,9 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-""" Implementation of the forward-backward primal-dual splitting algorithm for
-optimization.
-"""
+"""Optimization methods based on a forward-backward splitting scheme."""
 
 
 # Imports for common Python 2/3 codebase
@@ -33,7 +31,7 @@ __all__ = ('forward_backward_pd',)
 
 def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
                         callback=None, **kwargs):
-    """ The forward-backward primal-dual splitting algorithm.
+    """The forward-backward primal-dual splitting algorithm.
 
     The algorithm minimizes the sum of several convex functionals composed with
     linear operators,
@@ -41,7 +39,7 @@ def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
         ``min_x f(x) + sum_i g_i(L_i x) + h(x)``,
 
     where ``f``, ``g_i`` are convex functionals, ``L_i`` are linear
-    `Operator`'s, and ``h`` is a convex and differentiable functional.
+    operator's, and ``h`` is a convex and differentiable functional.
 
     The method can also be used to solve the more general problem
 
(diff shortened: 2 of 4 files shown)


examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,69 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Total variation denoising using the Forward-backward solver.
```
adler-j: If we're only going to have one example, I'd prefer it to be deconvolution or tomography.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+from future import standard_library
+standard_library.install_aliases()
+
+from odl.operator import Operator
+
+
+__all__ = ('forward_backward_pd',)
+
+
+def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
```
adler-j: Not quite sure I like this name. It feels too general. Isn't there several algorithms that could go under this name? Perhaps we need to be more specific.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    # Iteratively solve
+    for k in range(niter):
+        x_old = x
+
+        tmp_1 = grad_h(x) + sum(Li.adjoint(vi) for Li, vi in zip(L, v))
+        prox_f(tau)(x - tau * tmp_1, out=x)
+        y.lincomb(2.0, x, -1, x_old)
+
+        for i in range(m):
+            # In this case gradients were given
```
adler-j: I would move these comments inside the statements:

```
if something:
     # comment on something
else:
    # comment on what happens now
```

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+            if grad_cc_l is not None:
+                tmp_2 = sigma[i] * (L[i](y) + grad_cc_l[i](v[i]))
+
+            # In this case there were not. "Applying" zero-operator
+            else:
+                tmp_2 = sigma[i] * L[i](y)
+
+            v[i] = prox_cc_g[i](sigma[i])(v[i] + tmp_2)
```
adler-j: out parameter

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``
+
+    where l_i are convex functions and @ is the infimal convolution:
+
+        ``(f @ g)(x) = inf_y { f(x-y) + g(y) }``
+
+    Parameters
+    ----------
+    x : `Vector`
```
adler-j: `LinearSpaceVector`

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+# Internal
+import odl
+from odl.solvers import forward_backward_pd
+from odl.util.testutils import all_almost_equal, example_element
+
+# Places for the accepted error when comparing results
+PLACES = 8
+
+
+def test_forward_backward_inpit_handeling():
```
adler-j: input

adler-j: handling

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+    """Test to see that input is handled correctly.
+    """
```
adler-j: Oneline comments should have the citation marks on the same row

"""Some text."""

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+    prox_cc_g = [odl.solvers.proximal_zero(space1),  # Identity operator
+                 odl.solvers.proximal_zero(space1)]  # Identity operator
+    prox_f = odl.solvers.proximal_zero(space1)  # Identity operator
+    grad_h = odl.ZeroOperator(space1)
+
+    # Check that the algorithm runs. With the above operators the algorithm
+    # is the identity operator in each iteration
+    x0 = example_element(space1)
+    x = x0.copy()
+    niter = np.random.randint(low=3, high=20)
```
adler-j: no need to randomize this

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+                            sigma=[1.0, 1.0], niter=niter)
+
+
+def test_forward_backward():
+    """Test for the forward-backward solver by minimizing ||x||_2^2. The
+    general problem is of the form
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
+
+    and here we take f(x) = g(x) = 0, h(x) = ||x||_2^2 and L is the
```
adler-j: I would prefer a test that actually uses the other ones, even if trivial.

adler-j:
Nice addition. My main worry is that the naming is too general, but someone more knowledgable than me in these methods should know this better than I do.

At some point we also need to benchmark all of our solvers for some typical problem (TV regularized CT recon, I guess).

aringh:
Will update according to comments :)

Regarding the naming I just took the name as given in the paper (more or less at least). I do agree that it is a very general name; will see if I can find something more "specific". But this is definitely an open problem/discussion.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,69 @@
...
+# Create noisy data
+noisy_data = data + noise
+data.show('Original data')
+noisy_data.show('Noisy convolved data')
+
+
+# Gradient for TV regularization
+gradient = odl.Gradient(space)
+
+# Assemble all operators
```
kohr-h: Very general comment, could be more tailored to this situation.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,69 @@
...
+lin_ops = [gradient]
+
+# Create proximals as needed
```
kohr-h: Same here, the comment does not really add anything. Should be specific to this example and the method.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,69 @@
...
+prox_cc_g = [odlsol.proximal_cconj_l1(gradient.range, lam=1e1,
+                                      isotropic=False)]
+prox_f = odlsol.proximal_box_constraint(space, 0, 255)
+
+# Create gradient needed. This is the derivative of the 2-norm squared
```
kohr-h: Goes into the right direction (though the first sentence is incomplete). I'd appreciate a comment like
```
# This gradient encodes the differentiable term(s) of the goal functional,
# which corresponds to the "forward" part of the method.
```

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,69 @@
...
+grad_h = odl.ResidualOperator(odl.IdentityOperator(space), noisy_data)
+
+# Solve
```
kohr-h: More details

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,69 @@
...
+x = noisy_data.copy()
+
+callback = (odl.solvers.CallbackShow(display_step=20, clim=[0, 255]) &
+            odl.solvers.CallbackPrintIteration())
```
kohr-h: Comment on this?

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Implementation of the Forward-Backward splitting algorithm """
```
kohr-h: There are actually more algorithms than this one which fall into the "forward-backward" category. To be more general, consider writing `Optimization methods based on a forward-backward splitting scheme.` or similar.

odl/solvers/advanced/forward_backward.py
kohr-h: Depends. The "pd" suffix indicates that it's the primal-dual one and not the "vanilla" version without operator. Since this is somewhat the forefront of algorithm prototyping, I'd be less worried here to change names as needed later. Currently, this seems like an okay name to me.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+                        callback=None, **kwargs):
+    """ The forward-backward primal-dual splitting algorithm
+
+    Minimizes the sum of several convex functions composed with linear
+    operators
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
+
+    Where f, g_i, h are convex functions, L_i are linear `Operator`'s, and h is
+    also differentiable.
```
kohr-h: Write f, g_i etc. in double backticks.
"also" is ambiguous, I'd prefer an explicit "convex and differentiable".

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Can also be used to solve the more general problem
```
kohr-h: Make this a full sentence.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+__all__ = ('forward_backward_pd',)
+
+
+def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
+                        callback=None, **kwargs):
+    """ The forward-backward primal-dual splitting algorithm
+
+    Minimizes the sum of several convex functions composed with linear
```
kohr-h: This should be a proper sentence.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    operators
```
kohr-h: operators,

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
+
+    Where f, g_i, h are convex functions, L_i are linear `Operator`'s, and h is
```
kohr-h: "where" not capitalized

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
+                        callback=None, **kwargs):
+    """ The forward-backward primal-dual splitting algorithm
+
+    Minimizes the sum of several convex functions composed with linear
+    operators
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
```
kohr-h: comma at the end

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Where f, g_i, h are convex functions, L_i are linear `Operator`'s, and h is
+    also differentiable.
+
+    Can also be used to solve the more general problem
+
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``
```
kohr-h: comma at the end

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    where l_i are convex functions and @ is the infimal convolution:
+
+        ``(f @ g)(x) = inf_y { f(x-y) + g(y) }``
+
+    Parameters
+    ----------
+    x : `Vector`
+        Initial point
```
kohr-h: Not sufficient as description. For example, it should me mentioned that the values are mutated (iterates are stored here).

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    prox_f : `callable`
+        Funciton returning an `Operator` when called with `tau`.
```
kohr-h: Function

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+        A sequence with as many elements as ``prox__cc_gs`` of operators
+        ``L_i``
+    tau : `float`
+        Stepsize of ``f``
+    sigma : `sequence` of  `float`
+        Stepsize of the ``g_i``'s
+    niter : `int`
+        Number of iterations
+    callback : `Callback`, optional
+        Show partial results
```
kohr-h: This stuff seems to be largely copy & paste from the DR method, same comments apply here.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Other Parameters
+    ----------------
+    grad_cc_l : `sequence` of `Operator`'s, optional
+        Sequence of operators representing the gradient of  `l_i^*`.
+        If omitted, the simpler problem will be considered, which corresponds
+        to the convex conjugate functionals of the l_i:s being zero functionals
+        and hence the gradient being the zero-operator.
```
kohr-h: In principle good, but a bit too detailed. It's sufficient to write that the simpler problem without the `l_i`'s is considered if `None` is given. The rest could possibly be an item in the `Notes` section.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+        Number of iterations
+    callback : `Callback`, optional
+        Show partial results
+
+    Other Parameters
+    ----------------
+    grad_cc_l : `sequence` of `Operator`'s, optional
+        Sequence of operators representing the gradient of  `l_i^*`.
+        If omitted, the simpler problem will be considered, which corresponds
+        to the convex conjugate functionals of the l_i:s being zero functionals
```
kohr-h: I'm not sure if l_i:s is understood outside Sweden :-)

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+        and hence the gradient being the zero-operator.
+
+    Notes
+    -----
+    Strictly, we have the following conditions on the functions involved: f and
+    g_i are proper, convex and lower semicontinuous, and h is convex and
```
kohr-h: f, g_i, h in math notation

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    differentialbe with :math:`\\eta`-Lipschitz continuous gradient.
+
+    For the optional input l_i we need it to be proper, convex, lower
```
kohr-h: l_i in math

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    ----------------
+    grad_cc_l : `sequence` of `Operator`'s, optional
+        Sequence of operators representing the gradient of  `l_i^*`.
+        If omitted, the simpler problem will be considered, which corresponds
+        to the convex conjugate functionals of the l_i:s being zero functionals
+        and hence the gradient being the zero-operator.
+
+    Notes
+    -----
+    Strictly, we have the following conditions on the functions involved: f and
```
kohr-h: What does "strictly" mean? Is the following expressing the intentions correctly? "The exact conditions on the involved functionals are as follows:"

I'd also prefer the term "functional" over "function".

odl/solvers/advanced/forward_backward.py
kohr-h: The "we" style is not good for documentation (one can debate if it's good for papers or not). It can always be formulated more elegantly. My suggestion is 
```
The optional functionals :math:`l_i` need to be ...
```

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    g_i are proper, convex and lower semicontinuous, and h is convex and
+    differentialbe with :math:`\\eta`-Lipschitz continuous gradient.
+
+    For the optional input l_i we need it to be proper, convex, lower
+    semicontinuous, and :math:`\\nu_i^{-1}`-strongly convex. The fact that l_i
```
kohr-h: redundant "convex" and "strongly convex"

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    is :math:`\\nu_i^{-1}`-strongly convex implies that the convex conjugate
+    functional of l_i is differentiable with :math:`\\nu_i`-Lipschitz
+    continuous gradient.
+
+    To guarantee convergence, the parameters ``tau``, ``sigma`` and
+    ``L`` need to satisfy
+
+    .. math::
+
+       2 * \min \{ \\frac{1}{\\tau}, \\frac{1}{\sigma_1}, \\ldots,
```
kohr-h: "*" should be "\cdot" in math notation if a multiplication dot is desired. I'd omit it here.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+       \\frac{1}{\sigma_m} \} * \min\{ \\eta, \\nu_1, \\ldots, \\nu_m  \} *
```
kohr-h: no need for "*"

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    Notes
+    -----
+    Strictly, we have the following conditions on the functions involved: f and
+    g_i are proper, convex and lower semicontinuous, and h is convex and
+    differentialbe with :math:`\\eta`-Lipschitz continuous gradient.
+
+    For the optional input l_i we need it to be proper, convex, lower
+    semicontinuous, and :math:`\\nu_i^{-1}`-strongly convex. The fact that l_i
+    is :math:`\\nu_i^{-1}`-strongly convex implies that the convex conjugate
+    functional of l_i is differentiable with :math:`\\nu_i`-Lipschitz
```
kohr-h: math l_i

kohr-h:
```
the convex conjugate :math:`l_i^*`
```

odl/solvers/advanced/forward_backward.py
kohr-h: You pick one `l_i` here, so the text should say "The fact that a functional l_i is ..."
Note the "a".

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    continuous gradient.
+
+    To guarantee convergence, the parameters ``tau``, ``sigma`` and
+    ``L`` need to satisfy
+
+    .. math::
+
+       2 * \min \{ \\frac{1}{\\tau}, \\frac{1}{\sigma_1}, \\ldots,
+       \\frac{1}{\sigma_m} \} * \min\{ \\eta, \\nu_1, \\ldots, \\nu_m  \} *
+       \\sqrt{1 - \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2} > 1
```
kohr-h: comma at the end

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    where, if the simpler problem is considered, all :math:`\\nu_i` can be
+    considered to be :math:`\\infty`.
```
kohr-h: Some guideline of parameter choice would be great, i.e. is there some easy way of getting the condition right? Something like "choose one parameter such and such, then the other one can be picked such and such".

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    For references on the Forward-Backward algorithm, see [BC2015]_.
+
+    For more on convex analysis including convex conjugates and
+    resolvent operators see [Roc1970]_.
+
+    For more on proximal operators and algorithms see [PB2014]_.
```
kohr-h: Slight monotonicity in the text here.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+    if not all(op.is_linear for op in L):
+        raise ValueError('not all operators in `L` are linear')
+    if not all(x in op.domain for op in L):
+        raise ValueError('`x` not in the domain of all operators')
+    if len(sigma) != m:
+        raise ValueError('len(sigma) != len(L)')
+    if len(prox_cc_g) != m:
+        raise ValueError('len(prox_cc_g) != len(L)')
+
+    # Get parameters from kwargs
```
kohr-h: Obvious

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Check for unused parameters
+    if kwargs:
+        raise TypeError('unexpected keyword argument: {}'.format(kwargs))
+
+    # Pre-allocate values
+    v = [Li.range.zero() for Li in L]
+    y = x.space.zero()
+
+    # Iteratively solve
```
kohr-h: obvious

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+        raise ValueError('len(sigma) != len(L)')
+    if len(prox_cc_g) != m:
+        raise ValueError('len(prox_cc_g) != len(L)')
+
+    # Get parameters from kwargs
+    grad_cc_l = kwargs.pop('prox_cc_l', None)
+    if grad_cc_l is not None and len(grad_cc_l) != m:
+        raise ValueError('`grad_cc_l` not same length as `L`')
+
+    # Check for unused parameters
```
kohr-h: obvious

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,165 @@
...
+        tmp_1 = grad_h(x) + sum(Li.adjoint(vi) for Li, vi in zip(L, v))
+        prox_f(tau)(x - tau * tmp_1, out=x)
+        y.lincomb(2.0, x, -1, x_old)
+
+        for i in range(m):
+            # In this case gradients were given
+            if grad_cc_l is not None:
+                tmp_2 = sigma[i] * (L[i](y) + grad_cc_l[i](v[i]))
+
+            # In this case there were not. "Applying" zero-operator
```
kohr-h: This comment is a bit misleading. Applying the zero operator would correspond to setting something to zero, which is not what happens. I'd write a comment before the `if grad_cc_l ...` saying that the `grad_cc_l[i]` part is omitted if the `l_i` functionals were not given.

test/solvers/advanced/forward_backward_test.py
kohr-h: Also: we don't use docstrings in tests, the function name should be telling enough. If necessary, a comment at the beginning can be added.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Create "correct" set of operators
+    lin_ops = [odl.ZeroOperator(space1), odl.ZeroOperator(space1)]
+    prox_cc_g = [odl.solvers.proximal_zero(space1),  # Identity operator
+                 odl.solvers.proximal_zero(space1)]  # Identity operator
+    prox_f = odl.solvers.proximal_zero(space1)  # Identity operator
+    grad_h = odl.ZeroOperator(space1)
+
+    # Check that the algorithm runs. With the above operators the algorithm
+    # is the identity operator in each iteration
```
kohr-h: "the algorithm is the identity operator" is not really understandable.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def test_forward_backward_inpit_handeling():
+    """Test to see that input is handled correctly.
+    """
+
+    # Create spaces
+    space1 = odl.uniform_discr(0, 1, 10)
+
+    # Create "correct" set of operators
```
kohr-h: Comments in tests should be used very sparely. The only two use cases are (1) to structure tests (e.g. one section for real, one for complex) and (2) explain complex tests or non-obvious things that happen.

In this case, the comment below on the expected behavior is good, the others can be deleted without loss of information.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+    # Testing that sizes needs to agree:
+    # To few sigma_i:s
+    with pytest.raises(ValueError):
+        forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
+                            sigma=[1.0], niter=niter)
+
+    # To many sigma_i:s
+    with pytest.raises(ValueError):
+        forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
+                            sigma=[1.0, 1.0, 1.0], niter=niter)
```
kohr-h: These two tests do the same. Since the solver checks for lengths, there is no difference between a too short list and a too long list. Remove one.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,125 @@
...
+        forward_backward_pd(x, prox_f, prox_cc_g_to_few, lin_ops, grad_h,
+                            tau=1.0, sigma=[1.0, 1.0], niter=niter)
+
+    # To many operators
+    prox_cc_g_to_many = [odl.solvers.proximal_zero(space1),
+                         odl.solvers.proximal_zero(space1),
+                         odl.solvers.proximal_zero(space1)]
+    with pytest.raises(ValueError):
+        forward_backward_pd(x, prox_f, prox_cc_g_to_many, lin_ops, grad_h,
+                            tau=1.0, sigma=[1.0, 1.0], niter=niter)
```
kohr-h: One or two length mismatch tests should be enough.

test/solvers/advanced/forward_backward_test.py
kohr-h: Agreed. The tests can test different aspects separately, one "basic" like this, one "with_lin_ops" and one "with_li". If you find a simple problem that can be solved explicitly, even better. Could be something like a least squares problem with Tikhonov-type regularization that can be solved using the normal equation.

odl/solvers/advanced/forward_backward.py
aringh: The honest answer here is that I don't know I good way of doing it. The example is done by trial and error... I have tried to glance through the article but didn't find any such guidance of how to select them.

odl/solvers/advanced/forward_backward.py
aringh: Your comment is probably more clear, but mathematically I think you can see it as `grad_cc_l[i]` being the zero operator (however, instead of creating a list of zero operators and using the same line `tmp_2 = sigma[i] * (L[i](y) + grad_cc_l[i](v[i]))` in all cases, I do this if-else check).

odl/solvers/advanced/forward_backward.py
kohr-h: That's okay. But since the condition looks quite checkable with given parameters, a small helper function like `fb_pd_param_choice_is_valid(sigma, tau, ...)` would take some burden off the user.

test/solvers/advanced/forward_backward_test.py
aringh: Sure. I just tried to see the algorithm as kind of a black box when I wrote the test. "Not knowing" exactly what checks it does on input data, I want it to catch both to long and to short lists.

odl/solvers/advanced/forward_backward.py
kohr-h: Well, if you insist on your zero operators, you need to explain also where they come from (why is `None` leading to zero operators?). While you're right mathematically, I think this zero operator connection just adds noise to the logic -- to me the implication `None -> omit stuff` is much more obvious than `None -> zero operators`.

examples/solvers/forward_backward_primal_dual_denoising.py
aringh: This example uses the partition that `f` are the constraints, `g` is the TV term, and `h` contains the smooth l2-squared term. With deconvolution you need to have the l2-squared term as one of the `g_i` terms, since you need a linear operator acting on `x`.

I can add a deconvolution example as well. More or less the same example as for the Douglas-Rachford?

test/solvers/advanced/forward_backward_test.py
aringh: I do agree that the naming should say what the test does, and in this case the docstring might be redundant. But in the second test I think the docstring saves a lot of effort for the person how wants to read and see what the test does (which function is what). Should I remove that one too?

kohr-h: At least in this case, the function name already conveys the same information as the subsequent docstring. If the test is too complex to be described in a function name, a comment can clarify stuff. One can safely assume that tests are exclusively read by developers, and only when they're failing. So the test name should state what it tests for such that it is easier to pin down the code that is causing the failure. And one can assume a certain level of sophistication of the one looking at the test code -- a minimal amount of comments of non-obvious code is sufficient.

test/solvers/advanced/forward_backward_test.py
kohr-h: But this level of "completeness" is a bit over the top for my taste. To be consistent with this granularity, you would have to add even more tests, for example give different kinds of sequences or put garbage into a sequence. For a check that is as simple as `len(this) == len(that)` we don't need 4 tests.

aringh:
Good to go?

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+space = odl.uniform_discr([0, 0], [n, n], [n, n])
+
+# Load image and noise
+data = space.element(np.rot90(scipy.misc.ascent()[::2, ::2], 3))
+noise = odl.phantom.white_noise(space) * 10.0
+
+# Create noisy data
+noisy_data = data + noise
+data.show('Original data')
+noisy_data.show('Noisy convolved data')
```
kohr-h: Ain't no convolution nowhere to be seen.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Create initial guess for the solver.
+x = noisy_data.copy()
+
+# Used to display intermediate results and print iteration number.
+callback = (odl.solvers.CallbackShow(display_step=20, clim=[0, 255]) &
+            odl.solvers.CallbackPrintIteration())
+
+# Call the solver
+odlsol.forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
```
kohr-h: Only place where the `odlsol` name is used. Doesn't seem like a major pain to use `odl.solvers` even here.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+import odl.solvers as odlsol
+
+# Parameters
+n = 256
+
+# Create a space
+space = odl.uniform_discr([0, 0], [n, n], [n, n])
+
+# Load image and noise
+data = space.element(np.rot90(scipy.misc.ascent()[::2, ::2], 3))
```
kohr-h: I'd prefer reading the image first and rotating it, and then use its `shape` as `max_corner` and `shape` for the space. More robust.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Gradient for TV regularization
+gradient = odl.Gradient(space)
+
+# Assemble the linear operators. Here the TV-term is mapped into a composition
+# of the 1-norm and the gradient. See the documentation of the solver
+# `forward_backward_pd` for the general form of the problem.
+lin_ops = [gradient]
+
+# Create proximals for the convex conjugate of the 1-norm and the bound
```
kohr-h: conjugates

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+data.show('Original data')
+noisy_data.show('Noisy convolved data')
+
+
+# Gradient for TV regularization
+gradient = odl.Gradient(space)
+
+# Assemble the linear operators. Here the TV-term is mapped into a composition
+# of the 1-norm and the gradient. See the documentation of the solver
+# `forward_backward_pd` for the general form of the problem.
```
kohr-h: Good. Only "mapped into" could be "represented as", easier to understand (for me at least).

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Create proximals for the convex conjugate of the 1-norm and the bound
+# constrains.
+prox_cc_g = [odlsol.proximal_cconj_l1(gradient.range, lam=1e1,
+                                      isotropic=False)]
+prox_f = odlsol.proximal_box_constraint(space, 0, 255)
+
+# This gradient encodes the differentiable term(s) of the goal functional,
+# which corresponds to the "forward" part of the method. In this example the
+# differentiable part is the 2-norm squared.
```
kohr-h: "2-norm squared" -> "squared 2-norm"

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+grad_h = odl.ResidualOperator(odl.IdentityOperator(space), noisy_data)
+
+# Create initial guess for the solver.
+x = noisy_data.copy()
+
+# Used to display intermediate results and print iteration number.
+callback = (odl.solvers.CallbackShow(display_step=20, clim=[0, 255]) &
+            odl.solvers.CallbackPrintIteration())
+
+# Call the solver
```
kohr-h: Say that `x` is updated in-place.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Implementation of the forward-backward primal-dual splitting algorithm for
+optimization.
```
kohr-h: One line. No space in the beginning.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Implementation of the forward-backward primal-dual splitting algorithm for
```
kohr-h: This text should be general, the primal-dual variant is only one possible implementation. Things in here can be anything *based* on the forward-backward splitting scheme.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+
+from odl.operator import Operator
+
+
+__all__ = ('forward_backward_pd',)
+
+
+def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
+                        callback=None, **kwargs):
+    """ The forward-backward primal-dual splitting algorithm.
```
kohr-h: No space in the beginning.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+
+    The algorithm minimizes the sum of several convex functionals composed with
+    linear operators,
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``,
+
+    where ``f``, ``g_i`` are convex functionals, ``L_i`` are linear
+    `Operator`'s, and ``h`` is a convex and differentiable functional.
```
kohr-h: I wouldn't make a link in all places. Here it seems to look better to just use the word "operators".

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    ----------
+    x : `LinearSpaceVector`
+        Initial point, updated in place.
+    prox_f : `callable`
+        `Proximal factory` for the functional ``f``.
+    prox_cc_g : `sequence` of `callable`'s
+        Sequence of `proximal factories` for the convex conjuates of the
+        functionals ``g_i``.
+    grad_h : `Operator`
+        Operators representing the gradient of  `h`.
```
kohr-h: Why plural?

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    L : `sequence` of `Operator`'s
+        A sequence with as many elements as ``prox__cc_gs`` of operators
+        ``L_i``
```
kohr-h: dot

kohr-h: Also: mention "linear"

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    tau : `float`
+        Step size-like parameter for ``prox_f``
```
kohr-h: dot

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    sigma : `sequence` of  `float`
+        Sequence of step size-like parameter for the sequence ``prox_cc_g``
```
kohr-h: parameters
dot at the end

kohr-h: Good formulation btw, very clear

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    niter : `int`
+        Number of iterations
```
kohr-h: dot

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    callback : `Callback`, optional
+        Show partial results
```
kohr-h: dot

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    x : `LinearSpaceVector`
+        Initial point, updated in place.
+    prox_f : `callable`
+        `Proximal factory` for the functional ``f``.
+    prox_cc_g : `sequence` of `callable`'s
+        Sequence of `proximal factories` for the convex conjuates of the
+        functionals ``g_i``.
+    grad_h : `Operator`
+        Operators representing the gradient of  `h`.
+    L : `sequence` of `Operator`'s
```
kohr-h: no `'s`

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    sigma : `sequence` of  `float`
+        Sequence of step size-like parameter for the sequence ``prox_cc_g``
+    niter : `int`
+        Number of iterations
+    callback : `Callback`, optional
+        Show partial results
+
+    Other Parameters
+    ----------------
+    grad_cc_l : `sequence` of `Operator`'s, optional
```
kohr-h: no `'s`

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+        Sequence of operators representing the gradient of  `l_i^*`.
```
kohr-h: gradients

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+        If omitted, the simpler problem will be considered.
```
kohr-h: simpler problem without `l_i`?

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Notes
+    -----
+    The exact conditions on the involved functionals are as follows: :math:`f`
+    and :math:`g_i` are proper, convex and lower semicontinuous, and :math:`h`
+    is convex and differentialbe with :math:`\\eta`-Lipschitz continuous
+    gradient.
+
+    The optional input :math:`\\nabla l_i^*` need to be :math:`\\nu_i`
```
kohr-h: "input" -> "functionals"

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    -Lipschitz continuous. Note that in the reference [BC2015]_, the condition
```
kohr-h: I guess "-Lipschitz" will appear detached from the nu_i thing. Need to be on the same row.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    is formulated as :math:`l_i` being proper, lower semicontinuous, and
+    :math:`\\nu_i^{-1}`-strongly convex, which implies that :math:`l_i^*` have
+    :math:`\\nu_i`-Lipschitz continuous gradient.
+
+    If the optional input :math:`\\nabla l_i^*` is omitted, the simpler problem
+    will be considered. Mathematically, this is done by taking :math:`l_i` to
+    be the functional that is zero only in the zero element and :math:`\\infty`
+    otherwise. This gives that :math:`l_i^*` is the zero functional, and hence
+    the corresponding gradient is the zero operator.
```
kohr-h: Good text.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,168 @@
...
+    # Problem size
+    m = len(L)
+
+    # Validate input
+    if not all(isinstance(op, Operator) for op in L):
+        raise ValueError('`L` not a sequence of operators')
+    if not all(op.is_linear for op in L):
+        raise ValueError('not all operators in `L` are linear')
+    if not all(x in op.domain for op in L):
+        raise ValueError('`x` not in the domain of all operators')
```
kohr-h:
```
all operators in `L`
```

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,181 @@
...
+    assert x == x0
+
+    # Testing that sizes needs to agree:
+    # To few sigma_i:s
+    with pytest.raises(ValueError):
+        forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
+                            sigma=[1.0], niter=niter)
+
+    # To many operators
+    prox_cc_g_to_many = [odl.solvers.proximal_zero(space1),
```
kohr-h: "to" -> "too" (some more places)

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,181 @@
...
+    space2 = odl.uniform_discr(1, 2, 10)
+    x = example_element(space2)
+    with pytest.raises(ValueError):
+        forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=1.0,
+                            sigma=[1.0, 1.0], niter=niter)
+
+
+def test_forward_backward_basic():
+    """Test for the forward-backward solver by minimizing ||x||_2^2. The
+    general problem is of the form
```
kohr-h: If you insist on having a docstring instead of comments, you need to follow the docstring convention :-). One-liner first, then the rest.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_forward_backward_with_li():
+    """Test for the forward-backward solver by minimizing the functional
+    ``(g @ l)(x) + h(x)``, where
+
+        ``(g @ l)(x) = inf_y { g(x-y) + l(y) }``,
+
+    g is the indicator function on [-1, 1], and l(x) = h(x) = 1/2||x - 5||_2^2.
+    The optimal solution to this problem is given by x in [4,6].
```
kohr-h: Any `x in [4, 6]`? Is this not a convex problem?

kohr-h:
Some minor, mostly stylistic things. Otherwise this looks good to me.

aringh:
I will be away for a conference from tomorrow, and all of next week. If something more needs to be done will fix it afterwards. If everything is ok, please merge it :)

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+    linear operators,
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``,
+
+    where ``f``, ``g_i`` are convex functionals, ``L_i`` are linear
+    operator's, and ``h`` is a convex and differentiable functional.
+
+    The method can also be used to solve the more general problem
+
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``,
```
adler-j: This is not a generalization of the above, since it lacks ``h(x)``.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+
+    where ``l_i`` are convex functionals and @ is the infimal convolution:
+
+        ``(f @ g)(x) = inf_y { f(y) + g(x-y) }``.
```
adler-j: Perhaps write with the same letters as above

```
(g @ l)(x) = inf_y { g(y) + l(x-y) }
```

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+    gradient, :math:`\\eta > 0`.
+
+    The optional operators :math:`\\nabla l_i^*` need to be
+    :math:`\\nu_i^{-1}`-Lipschitz continuous. Note that in the reference
+    [BC2015]_, the condition is formulated as :math:`l_i` being proper, lower
+    semicontinuous, and :math:`\\nu_i^{-1}`-strongly convex, which implies that
+    :math:`l_i^*` have :math:`\\nu_i`-Lipschitz continuous gradients.
+
+    If the optional operators :math:`\\nabla l_i^*` are omitted, the simpler
+    problem without :mat:`l_i` will be considered. Mathematically, this is done
```
adler-j: `:mat` -> `:math:`

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``,
+
+    where ``f``, ``g_i`` are convex functionals, ``L_i`` are linear
+    operator's, and ``h`` is a convex and differentiable functional.
+
+    The method can also be used to solve the more general problem
+
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x)``,
+
+    where ``l_i`` are convex functionals and @ is the infimal convolution:
```
adler-j: In the math section you note that `l_i` should be convex and differentiable

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+        Initial point, updated in place.
+    prox_f : `callable`
+        `Proximal factory` for the functional ``f``.
+    prox_cc_g : `sequence` of `callable`'s
+        Sequence of `proximal factories` for the convex conjuates of the
+        functionals ``g_i``.
+    grad_h : `Operator`
+        Operator representing the gradient of  ``h``.
+    L : `sequence` of `Operator`
+        A sequence with as many elements as ``prox__cc_gs`` of linear operators
```
adler-j: `prox__cc_gs` -> `prox_cc_gs`

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+        Show partial results.
+
+    Other Parameters
+    ----------------
+    grad_cc_l : `sequence` of `Operator`, optional
+        Sequence of operators representing the gradients of  ``l_i^*``.
+        If omitted, the simpler problem without ``l_i``  will be considered.
+
+    Notes
+    -----
```
adler-j: After having read this as a third party, I feel that we man want to restate the problem here with a-bit more depth and math syntax. Remembering all parameters from up there gets a bit hard.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+        A sequence with as many elements as ``prox__cc_gs`` of linear operators
+        ``L_i``.
+    tau : `float`
+        Step size-like parameter for ``prox_f``.
+    sigma : `sequence` of  `float`
+        Sequence of step size-like parameters for the sequence ``prox_cc_g``.
+    niter : `int`
+        Number of iterations.
+    callback : `Callback`, optional
+        Show partial results.
```
adler-j: Perhaps we should go for "Handle partial results". Or even write what actually happens: "Function that gets called with the current iterate after each iteration."

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+    L : `sequence` of `Operator`
+        A sequence with as many elements as ``prox__cc_gs`` of linear operators
+        ``L_i``.
+    tau : `float`
+        Step size-like parameter for ``prox_f``.
+    sigma : `sequence` of  `float`
+        Sequence of step size-like parameters for the sequence ``prox_cc_g``.
+    niter : `int`
+        Number of iterations.
+    callback : `Callback`, optional
```
adler-j: `Callback` -> `callable`

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+                                                     lower=lower_lim,
+                                                     upper=upper_lim))]
+    prox_f = odl.solvers.proximal_zero(space)
+    grad_h = odl.ZeroOperator(space)
+
+    grad_cc_ls = [odl.IdentityOperator(space)]
+
+    # Centering around a point further away from [-3,-1].
+    x = example_element(space) + space.element(3)
+    print(x)
```
adler-j: leftover print statements

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+
+    forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=0.5,
+                        sigma=[1.0], niter=20, grad_cc_l=grad_cc_ls)
+
+    solution_interv = odl.Interval(lower_lim - LOW_ACCURACY,
+                                   upper_lim + LOW_ACCURACY)
+    print(x)
```
adler-j: leftover print statements

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+    grad_cc_ls = [odl.IdentityOperator(space)]
+
+    # Centering around a point further away from [-3,-1].
+    x = example_element(space) + space.element(3)
+    print(x)
+
+    forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=0.5,
+                        sigma=[1.0], niter=20, grad_cc_l=grad_cc_ls)
+
+    solution_interv = odl.Interval(lower_lim - LOW_ACCURACY,
```
adler-j: A bit to complicated, simply do

```
assert lower_lim - LOW_ACCURACY < x
assert x < upper_lim + LOW_ACCURACY
```

this will also give better error messages on failure

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+                 odl.solvers.proximal_box_constraint(space,
+                                                     lower=lower_lim,
+                                                     upper=upper_lim))]
+    prox_f = odl.solvers.proximal_zero(space)
+    grad_h = odl.ZeroOperator(space)
+
+    grad_cc_ls = [odl.IdentityOperator(space)]
+
+    # Centering around a point further away from [-3,-1].
+    x = example_element(space) + space.element(3)
```
adler-j: No need to go this far, might aswell just set `x = space.element(10)` or something

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+    general problem is of the form
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``
+
+    and here we take f = 0, g = ||.||_2^2, L = alpha * IndentityOperator,
+    and h = ||. - b||_2^2.
+    """
+
+    space = odl.rn(10)
+    scaling = 0.1
```
adler-j: should be named alpha according to doc above

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+    x_global_min = space.zero()
+
+    forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=0.5,
+                        sigma=[1.0], niter=10)
+
+    assert all_almost_equal(x, x_global_min, places=HIGH_ACCURACY)
+
+
+def test_forward_backward_with_lin_ops():
+    """Test for the forward-backward solver with linear operatros.
```
adler-j: Just a general statement: I really like these introductions, well written, concise and sufficient.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+    lin_op = scaling * odl.IdentityOperator(space)
+    lin_ops = [lin_op]
+    prox_cc_g = [odl.solvers.proximal_cconj_l2_squared(space)]
+    prox_f = odl.solvers.proximal_zero(space)
+    # Gradient of two-norm square
+    grad_h = 2.0 * (odl.IdentityOperator(space) - odl.ConstantOperator(b))
+
+    x = example_element(space)
+
+    # Explicit solution: x_hat = (I^T * I + (alpha*I)^T * (alpha*I))^-1 * (I*b)
```
adler-j: The exact solution is given by

```python
x_global_min = b / (1 + alpha**2)
```

No need to use CG.

adler-j: Further: depending on another solver (in the case CG) makes reading the test quite hard, and can also (at least theoretically) give us circular tests, i.e. both fail but we fail to find it because the tests depend on each other.

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+                                                     lower=lower_lim,
+                                                     upper=upper_lim))]
+    prox_f = odl.solvers.proximal_zero(space)
+    grad_h = odl.IdentityOperator(space)
+
+    grad_cc_ls = [odl.IdentityOperator(space)]
+
+    # Centering the initial guess around the optimal solution, to reduce the
+    # number of iterations needed.
+    x = example_element(space)
```
adler-j: x = 10 is fine

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+
+def test_forward_backward_with_li_and_h():
+    """Test for the forward-backward solver with infimal convolution.
+
+    The test is done by minimizing the functional ``(g @ l)(x) + h(x)``, where
+
+        ``(g @ l)(x) = inf_y { g(y) + l(x-y) }``,
+
+    g is the indicator function on [-3, -1], and l(x) = h(x) = 1/2||x||_2^2.
+    The optimal solution to this problem is given by x = 0.5.
```
adler-j: x = -0.5

odl/solvers/advanced/forward_backward.py
adler-j: I guess we'll live with it for now. Naming of research things with yet to be settled names is a damn mess.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def forward_backward_pd(x, prox_f, prox_cc_g, L, grad_h, tau, sigma, niter,
+                        callback=None, **kwargs):
+    """The forward-backward primal-dual splitting algorithm.
+
+    The algorithm minimizes the sum of several convex functionals composed with
+    linear operators,
+
+        ``min_x f(x) + sum_i g_i(L_i x) + h(x)``,
```
adler-j: I know we follow the article here, but is it only me who feels that this order is arbitrary?

Why not simply go for:

```python
min_x f(x) + h(x) + sum_i g_i(L_i x)
```

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+    where ``l_i`` are convex functionals and @ is the infimal convolution:
+
+        ``(f @ g)(x) = inf_y { f(y) + g(x-y) }``.
+
+    Parameters
+    ----------
+    x : `LinearSpaceVector`
+        Initial point, updated in place.
+    prox_f : `callable`
+        `Proximal factory` for the functional ``f``.
```
adler-j: `Proximal factory` is still a dead link, I guess its my work to get it working however.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+    .. math::
+
+       2 \min \{ \\frac{1}{\\tau}, \\frac{1}{\sigma_1}, \\ldots,
+       \\frac{1}{\sigma_m} \} \cdot \min\{ \\eta, \\nu_1, \\ldots, \\nu_m  \}
+       \cdot \\sqrt{1 - \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2} > 1,
+
+    where, if the simpler problem is considered, all :math:`\\nu_i` can be
+    considered to be :math:`\\infty`.
+
+
```
adler-j: Too many newlines

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Note here that if the functional
```
adler-j: weird line-break

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,227 @@
...
+
+    space = odl.rn(10)
+    scaling = 0.1
+    b = example_element(space)
+
+    lin_op = scaling * odl.IdentityOperator(space)
+    lin_ops = [lin_op]
+    prox_cc_g = [odl.solvers.proximal_cconj_l2_squared(space)]
+    prox_f = odl.solvers.proximal_zero(space)
+    # Gradient of two-norm square
```
adler-j: newline above here

odl/solvers/advanced/forward_backward.py
adler-j: ~~A~~

test/solvers/advanced/forward_backward_test.py
adler-j: Also, since `LOW_ACCURACY` is 4 here, this tests that

`-5 < x < 1`

I'm guessing that is too generous?

aringh: Yes, my intent was of course 10^-4

odl/solvers/advanced/forward_backward.py
adler-j: Fixed now.

test/solvers/advanced/forward_backward_test.py
aringh: I don't get `lower_lim - 10**(-LOW_ACCURACY) <= x` to work... I get the error message

```
TypeError: unorderable types: float() <= NumpyFnVector()
```

Is this a bug? Or am I doing something wrong? (A quick-fix for this test is simply to use `x.data` in the comparison)

odl/solvers/advanced/forward_backward.py
adler-j: It confused me since you later asked for the gradient, so some more clarity would help.

odl/solvers/advanced/forward_backward.py
aringh: I think the order is a bit arbitrary as well... we might as well also take the order
```
min_x h(x) + f(x) + sum_i g_i(L_i x)
```
because, in some sense, the functionals are then in the order of "niceness" (h is differentiable, so nicest while g is composed with the linear operator and also involved in the infimal convolution if `l_i` is set).

I think we should stick to the same conventions as the articles; at least it makes comparing the algorithm with the article more easy.

test/solvers/advanced/forward_backward_test.py
adler-j: Use x[0] it seems. we may want to add a `__float__` to 1d vectors, that would fix this. What do you think @kohr-h?

odl/solvers/advanced/forward_backward.py
aringh: Too many lines and weird line breaks because this was supposed to be removed... I missed to do that. (This is wrong, and when I harmonized the notation with the notation in the article I realized that).

aringh:
The comments are now addressed :) Merge?

aringh:
Also: I'm not sure of how to fix the conflicts in a "nice" way. Can @adler-j  or @kohr-h fix that? :)

test/solvers/advanced/forward_backward_test.py
kohr-h: That doesn't help out of the box, you still need to call `float` on it. Or implement comparison, which I'm highly doubtful towards.
`__float__` would be in line with what Numpy does, we could add `__complex__` while we're at it and just pass to the underlying array. It's been added in version 1.11 I think. Could be done for CUDA, too.

examples/solvers/forward_backward_primal_dual_denoising.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+
+import numpy as np
+import scipy.misc
+import odl
+
+# Load image
+image = np.rot90(scipy.misc.ascent()[::2, ::2], 3)
+
+# Reading the size
```
adler-j: Redundant. Why not simply use `image.shape` below?

odl/solvers/advanced/forward_backward.py
adler-j: Agree with what you wrote, but sadly

> I think we should stick to the same conventions as the articles; at least it makes comparing the algorithm with the article more easy.

Is a very strong argument and we should stay with it.

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,184 @@
...
+    The method can also be used to solve the more general problem
+
+        ``min_x f(x) + sum_i (g_i @ l_i)(L_i x) + h(x)``,
+
+    where ``l_i`` are strongly convex functionals and @ is the infimal
+    convolution:
+
+        ``(g @ l)(x) = inf_y { g(y) + l(x-y) }``.
+
+    Note that the strong convetiy of ``l_i`` makes the convex conjugate
```
adler-j: convetiy -> convexity

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,184 @@
...
+    this.
+
+    Parameters
+    ----------
+    x : `LinearSpaceVector`
+        Initial point, updated in place.
+    prox_f : `callable`
+        `Proximal factory` for the functional ``f``.
+    prox_cc_g : `sequence` of `callable`'s
+        Sequence of `proximal factories` for the convex conjuates of the
```
adler-j: `proximal factories` does not exist as a link, need to write `proximal factory`'s

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,184 @@
...
+        functionals ``g_i``.
+    grad_h : `Operator`
+        Operator representing the gradient of  ``h``.
+    L : `sequence` of `Operator`
+        Sequence with as many elements as ``prox_cc_gs`` of linear operators
```
adler-j: Abit clunky sentence

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,184 @@
...
+     .. math::
+
+        \min_x f(x) + \sum_{i=0}^n (g_i \Box l_i)(L_i x) + h(x),
+
+     where :math:`L_i` are linear functionals and the infimal convolution
+     :math:`g \Box l` is defined by
+
+     .. math::
+
+        (g \Box l)(x) = \inf_y g(y) + l(x - y).
```
adler-j: `\Box` -> `\box`?

odl/solvers/advanced/forward_backward.py
```diff
@@ -0,0 +1,184 @@
...
+
+    The exact conditions on the involved functionals are as follows: :math:`f`
+    and :math:`g_i` are proper, convex and lower semicontinuous, and :math:`h`
+    is convex and differentialbe with :math:`\\eta^{-1}`-Lipschitz continuous
```
adler-j: differentiable

odl/solvers/advanced/forward_backward.py
adler-j: Looking great now!

test/solvers/advanced/forward_backward_test.py
```diff
@@ -0,0 +1,222 @@
...
+    grad_cc_ls = [odl.IdentityOperator(space)]
+
+    # Creating an element not to far away from -0.5, in order to converge in
+    # a few number of iterations.
+    x = space.element(10)
+
+    forward_backward_pd(x, prox_f, prox_cc_g, lin_ops, grad_h, tau=0.5,
+                        sigma=[1.0], niter=20, grad_cc_l=grad_cc_ls)
+
+    expected_result = space.element(-0.5)
```
adler-j:
```
expected_result = -0.5
assert almost_equal(x[0], expected_result, places=LOW_ACCURACY)
```

`data` is implementation specific for the numpy backend, indexing is not.

test/solvers/advanced/forward_backward_test.py
adler-j: Nicely fixed, but the string above should be simplifyable?

adler-j:
A few style comments then I can fix the merge for you, or you try yourself. Its a learning experiance and this should be a relatively simple one :)

kohr-h:
And don't merge as the GitHub instructions say, we don't want those "merge from master" messages in the final commit history. Go @aringh :-)

adler-j:
Basic conflict resolving strategy:

First update master to the latest

```
git checkout master
git pull
```

Then start the rebase

```
git checkout forward_backward_primal_dual
git rebase master
```

You will now get some message of the form CONFLICT: .... file.py. Go into that file(s) and fix the conflicts (they are denoted by `<<<<` etc and should be highlighted by your editor). Once all conflicts are fixed, go ahead and

```
git add file.py
git rebase --continue
```

you may then be asked to solve more conflicts, depending on how many commits your branch is behind, anyway simply repeat the process untill there are no more conflicts. Then you can push the fixed branch

```
git push --force
```

After which you should be able to here!

aringh:
Ok, I'll fix it as soon as I have time. Then I'll merge the branch :)

PR #483, opened 2016-07-08
ENH: simplified linearized deformations implementation and test

adler-j:
* Removed code that duplicates functionality from other parts of ODL
* Removed tests for the above (since it is already tested)
* Changed test function form hat to Gaussian, reduced test error bars by factor 10.


odl/deform/linearized.py
```diff
@@ -132,167 +130,26 @@ def derivative(self, displacement):
...
-        -------
-        adjoint : `LinDeformFixedTemplDerivAdj`
-            The adjoint of the operator.
-        """
-        return LinDeformFixedTemplDerivAdj(self.template, self.displacement,
-                                           self.def_grad)
-
-
-class LinDeformFixedTemplDerivAdj(Operator):
-
```
chongchenmath: The adjoint is removed?

adler-j: Adjoint is provided by the `PointwiseInner` class.

adler-j:
This has been rebased to master, which the upstream branch has not. I suggest we get this merged to that branch and get a final review done.

kohr-h:
Yes, do that.

PR #485, opened 2016-07-10
ENH: simplify _call assignment in Operator.__new__

kohr-h:
Small improvement I realized during review of PR #471. The old problem that the `self` argument was not filled in had to do with internal Python things. Basically it boils down to this: If a function is assigned as an attribute of a *class*, `self` is filled in magically. If it is assigned as an *instance* attribute, no `self`.

### DIFF
--- a/odl/space/fspace.py
+++ b/odl/space/fspace.py
@@ -216,7 +216,7 @@ class FunctionSetVector(Operator):
 
     """Representation of a `FunctionSet` element."""
 
-    def __init__(self, fset, fcall, out_dtype=None):
+    def __init__(self, fset, fcall):
         """Initialize a new instance.
 
         Parameters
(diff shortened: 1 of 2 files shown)


odl/space/fspace.py
```diff
@@ -216,7 +216,7 @@ class FunctionSetVector(Operator):
 
     """Representation of a `FunctionSet` element."""
 
-    def __init__(self, fset, fcall, out_dtype=None):
```
adler-j: why remove this?

adler-j:
looks ok!

odl/space/fspace.py
kohr-h: It's not used anywhere, and it only makes sense as a parameter for the space (this is the vector class). Also, the almost-deleted documentation was another hint :-)

PR #487, opened 2016-07-11
Add NormOperator and DistOperator

adler-j:
Minor pull, noticed that these guys were missing.

### DIFF
--- a/odl/operator/default_ops.py
+++ b/odl/operator/default_ops.py
@@ -492,6 +492,11 @@ class InnerProductOperator(Operator):
     ``InnerProductOperator(vec)(x) <==> x.inner(vec)``
 
     This is only applicable in inner product spaces.
+
+    See Also
+    --------
+    DistOperator : Distance to fixed vector.
+    NormOperator : Norm of a vector.
     """
 
     def __init__(self, vector):
@@ -597,7 +602,7 @@ class NormOperator(Operator):
 
         Parameters
         ----------
-        vector : `LinearSpace`
+        space : `LinearSpace`
             Space to take the norm in.
         """
         super().__init__(space, RealNumbers(), linear=False)
@@ -668,6 +673,102 @@ class NormOperator(Operator):
     def __str__(self):
         """Return ``str(self)``."""
         return 'NormOperator({})'.format(self.domain)
+
+
+class DistOperator(Operator):
+    """Operator taking the distance to a fixed vector.
+
+    ``NormOperator(x)(y) <==> x.dist(y)``
+
+    This is only applicable in metric spaces.
+
+    See Also
+    --------
+    InnerProductOperator : Inner product with fixed vector.
+    NormOperator : Norm of a vector.
+    """
+
+    def __init__(self, vector):
+        """Initialize a DistOperator instance.
+
+        Parameters
+        ----------
+        vector : `LinearSpaceVector`
+            Point to calculate the distance to.
+        """
+        self.vector = vector
+        super().__init__(vector.space, RealNumbers(), linear=False)
+
+    def _call(self, x):
+        """Return the distance to x.
+
+        Parameters
+        ----------
+        x : `domain` `element`
+            A vector in the domain.
+
+        Returns
+        -------
+        out : ``field`` `element`
+            Result of the inner product calculation
+
+        Examples
+        --------
+        >>> import odl
+        >>> r2 = odl.rn(2)
+        >>> x = r2.element([1, 1])
+        >>> op = DistOperator(x)
+        >>> op([4, 5])
+        5.0
+        """
+        return self.vector.dist(x)
+
+    def derivative(self, point):
+        """The derivative operator.
+
+        This is only applicable in inner product spaces.
+
+        Parameters
+        ----------
+        x : `domain` `element-like`
+            Point in which to take the derivative.
+
+        Returns
+        -------
+        derivative : `InnerProductOperator`
+
+        Raises
+        ------
+        ValueError
+            If ``point == self.vector``, in which case the derivative is not
+            well defined in the Frechet sense.
+
+        Examples
+        --------
+        >>> import odl
+        >>> r2 = odl.rn(2)
+        >>> x = r2.element([1, 1])
+        >>> op = DistOperator(x)
+        >>> derivative = op.derivative([2, 1])
+        >>> derivative([1, 0])
+        1.0
+        """
+        point = self.domain.element(point)
+        diff = point - self.vector
+        norm = diff.norm()
+        if norm == 0:
+            raise ValueError('Not differentiable at the reference vector {!r}.'
+                             ''.format(self.vector))
+
+        return InnerProductOperator(diff / norm)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return 'DistOperator({!r})'.format(self.vector)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return 'DistOperator({})'.format(self.vector)
 
 
 class ConstantOperator(Operator):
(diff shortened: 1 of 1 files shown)


odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        >>> derivative = op.derivative([1, 0, 0])
+        >>> derivative([1, 0, 0])
+        1.0
+        """
+        point = self.domain.element(point)
+        norm = point.norm()
+        if norm == 0:
+            raise ValueError('not differentiable in 0')
+
+        return InnerProductOperator(point / norm)
```
kohr-h: This assumes quite a bit on the norm of the space. The inner product operator needs to use the same weighting as the norm - I guess that's okay. But a custom norm implementation cannot be handled consistently, that needs to be filtered out (and documented that it won't work).

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+            defined in the Frechet sense.
+
+        Notes
+        -----
+        The derivative cannot be written in a general sense except in Hilbert
+        spaces, in that case, it is given by
+
+        .. Math::
+
+            (D ||.||)(x)(y) = < x / ||x||, y >
```
kohr-h: Notation for the differential is `D`?

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        ValueError
+            If ``point.norm() == 0``, in which case the derivative is not well
+            defined in the Frechet sense.
+
+        Notes
+        -----
+        The derivative cannot be written in a general sense except in Hilbert
+        spaces, in that case, it is given by
+
+        .. Math::
```
kohr-h: Will that work capitalized?

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        >>> import odl
+        >>> r2 = odl.rn(2)
+        >>> op = NormOperator(r2)
+        >>> op([3, 4])
+        5.0
+        """
+        return x.norm()
+
+    def derivative(self, point):
+        """The derivative operator.
```
kohr-h: Derivative of this operator in ``point``.

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        Raises
+        ------
+        ValueError
+            If ``point.norm() == 0``, in which case the derivative is not well
+            defined in the Frechet sense.
+
+        Notes
+        -----
+        The derivative cannot be written in a general sense except in Hilbert
+        spaces, in that case, it is given by
```
kohr-h: , in which case it is given by

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
         return "{}.T".format(self.vector)
 
 
+class NormOperator(Operator):
```
kohr-h: blank line

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+    This is only applicable in normed spaces.
+
+    See Also
+    --------
+    InnerProductOperator : Inner product with fixed vector.
+    DistOperator : Distance to fixed vector.
+    """
+
+    def __init__(self, space):
+        """Initialize a NormOperator instance.
```
kohr-h: I wouldn't hard-code the class name here, no need for that.

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        """
+        super().__init__(space, RealNumbers(), linear=False)
+
+    def _call(self, x):
+        """Return the norm of x.
+
+        Parameters
+        ----------
+        x : `domain` `element`
+            A vector in the domain.
```
kohr-h: The vector thing again :-)

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+
+        Returns
+        -------
+        out : ``field`` `element`
+            Result of the inner product calculation
```
kohr-h: inner product?
dot

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        point = self.domain.element(point)
+        norm = point.norm()
+        if norm == 0:
+            raise ValueError('not differentiable in 0')
+
+        return InnerProductOperator(point / norm)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return 'NormOperator({!r})'.format(self.domain)
```
kohr-h: Use `self.__class__.__name__`

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return 'NormOperator({})'.format(self.domain)
```
kohr-h: Use `self.__class__.__name__`

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+
+
+class DistOperator(Operator):
```
kohr-h: blank row

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+    This is only applicable in metric spaces.
+
+    See Also
+    --------
+    InnerProductOperator : Inner product with fixed vector.
+    NormOperator : Norm of a vector.
+    """
+
+    def __init__(self, vector):
+        """Initialize a DistOperator instance.
```
kohr-h: see above

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+
+        Parameters
+        ----------
+        space : `LinearSpace`
+            Space to take the norm in.
+        """
+        super().__init__(space, RealNumbers(), linear=False)
+
+    def _call(self, x):
+        """Return the norm of x.
```
kohr-h:
```
``x``
```
We have a bit mixed style for the `_call` docstring. Sometimes we say
```
Return ``self(x)``.
```
or implement in the case of `out`, but sometimes we have some other text. Would be good to have some common style.

odl/operator/default_ops.py
```diff
@@ -579,6 +585,211 @@ def __str__(self):
...
+        1.0
+        """
+        point = self.domain.element(point)
+        diff = point - self.vector
+        dist = self.vector.dist(point)
+        if dist == 0:
+            raise ValueError('not differentiable at the reference vector {!r}'
+                             ''.format(self.vector))
+
+        return InnerProductOperator(diff / dist)
```
kohr-h: Same here, need to exclude custom stuff.

kohr-h:
Minor things to be fixed, otherwise alright.

odl/operator/default_ops.py
adler-j: > This assumes quite a bit on the norm of the space. The inner product operator needs to use the same weighting as the norm - I guess that's okay.

Sure, but this is a requirement on the space that we have written in the doc.

>  But a custom norm implementation cannot be handled consistently, that needs to be filtered out (and documented that it won't work).

Depends on the implementer I guess, here we can only assume the norm and inner product match.

odl/operator/default_ops.py
adler-j: It did in my online editor, but I'll change it

odl/operator/default_ops.py
adler-j: It felt good here, what else do we go for in the doc?

kohr-h: We have `\partial`, e.g. in `landweber`. Feels a bit more robust against other stuff that may also want to use the letter `D`.

PR #488, opened 2016-07-11
Linearized deformations second take

adler-j:
Second take at linearized deformations. I've cleaned up most of the implementation and updated the doc, should be ready for final review @chongchenmath, @kohr-h, @ozanoktem .

### DIFF
--- a/examples/deform/linearized_fixed_displacement.py
+++ b/examples/deform/linearized_fixed_displacement.py
@@ -15,12 +15,23 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Example using the operator of fixed-displacement linearized deformation."""
+"""Example using the operator of fixed-displacement linearized deformation.
 
-# Imports for common Python 2/3 codebase
-from __future__ import print_function, division, absolute_import
-from future import standard_library
-standard_library.install_aliases()
+The linearized deformation operator with fixed displacement ``v`` maps
+a given template ``I`` to the function ``x --> I(x + v(x))``.
+This operator is linear.
+
+Here, we consider a 2D example, where the displacement field ``v``
+is a Gaussian in each component, with positive sign in the first and
+negative sign in the second component. Note that in the deformed image,
+the value at ``x`` is **taken from** the original image at ``x + v(x)``,
+hence the values are moved by ``-v(x)`` when comparing deformed and
+original templates.
+
+The adjoint is based on an approximation given by the transformation
+``x --> I(x - v(x))``, multiplied by an integral transformation factor
+``exp(-div(v))``.
+"""
 
 import numpy as np
 import odl
@@ -29,25 +40,24 @@ import odl
 # --- Create template and displacement field --- #
 
 
-# Discrete reconstruction space: discretized functions on the rectangle
-# [-1, 1]^2 with 100 samples per dimension.
-discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+# Template space: discretized functions on the rectangle [-1, 1]^2 with
+# 100 samples per dimension.
+templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# We use a rectangle as template
+template = odl.phantom.cuboid(templ_space, [-0.5, -0.25], [0.5, 0.25])
 
 # Create a product space for displacement field
-disp_field_space = discr_space.vector_field_space
+disp_field_space = templ_space.vector_field_space
 
-# Create a template, and set the template as an offset rectangle
-template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
+# Define a displacement field that bends the template a bit towards the
+# upper left. We use a list of 2 functions and discretize it using the
+# disp_field_space.element() method.
+sigma = 0.5
+disp_func = [
+    lambda x: 0.4 * np.exp(-(x[0] ** 2 + x[1] ** 2) / (2 * sigma ** 2)),
+    lambda x: -0.3 * np.exp(-(x[0] ** 2 + x[1] ** 2) / (2 * sigma ** 2))]
 
-# Define a displacement field that rotates an image clockwise by theta. Then it
-# shifts the image by hx to the right and by hy upwards.
-hx = -0.5
-hy = -0.5
-theta = np.pi / 4
-disp_func = [lambda x: (np.cos(theta) - 1) * x[0] - np.sin(theta) * x[1] + hx,
-             lambda x: np.sin(theta) * x[0] + (np.cos(theta) - 1) * x[1] + hy]
-
-# Create a displacement field based on ``disp_func``
 disp_field = disp_field_space.element(disp_func)
 
 # Show template and displacement field
@@ -55,18 +65,18 @@ template.show('Template')
 disp_field.show('Displacement field')
 
 
-# --- Example of LinDeformFixedDisp and its adjoint --- #
+# --- Apply LinDeformFixedDisp and its adjoint --- #
 
 
-# Define the deformation operator where the displacement field is fixed
-fixed_disp_op = odl.deform.LinDeformFixedDisp(disp_field)
+# Initialize the deformation operator with fixed displacement
+deform_op = odl.deform.LinDeformFixedDisp(disp_field)
 
-# Apply the deformation operator to get the deformed template
-deform_templ_fixed_disp = fixed_disp_op(template)
+# Apply the deformation operator to get the deformed template.
+deformed_template = deform_op(template)
 
-# Compute the adjoint of the operator applied to the template
-fixed_disp_adj = fixed_disp_op.adjoint(template)
+# Evaluate the adjoint at the same template.
+adj_result = deform_op.adjoint(template)
 
 # Show results
-deform_templ_fixed_disp.show('Deformed template')
-fixed_disp_adj.show('Adjoint applied to template')
+deformed_template.show('Deformed template')
+adj_result.show('Adjoint applied to the template')
--- a/examples/deform/linearized_fixed_displacement.py
+++ b/examples/deform/linearized_fixed_displacement.py
@@ -1,0 +1,72 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using the operator of fixed-displacement linearized deformation."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import odl
+
+
+# --- Create template and displacement field --- #
+
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-1, 1]^2 with 100 samples per dimension.
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# Create a product space for displacement field
+disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
+
+# Create a template
+template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
+
+# Define a displacement field that shifts an image by hx
+# to the right and by hy upwards. Then it rotates clockwise by theta.
+hx = -0.5
+hy = -0.5
+theta = np.pi / 4
+disp_func = [lambda x: (np.cos(theta) - 1) * x[0] - np.sin(theta) * x[1] + hx,
+             lambda x: np.sin(theta) * x[0] + (np.cos(theta) - 1) * x[1] + hy]
+
+# Create a displacement field based on ``disp_func``
+disp_field = disp_field_space.element(disp_func)
+
+# Show template and displacement field
+template.show('Template')
+disp_field.show('Displacement field')
+
+
+# --- Example of LinDeformFixedDisp and its adjoint --- #
+
+
+# Define the deformation operator where the displacement field is fixed
+fixed_disp_op = odl.deform.LinDeformFixedDisp(disp_field)
+
+# Apply the deformation operator to get the deformed template
+deform_templ_fixed_disp = fixed_disp_op(template)
+
+# Compute the adjoint of the operator
+fixed_disp_adj = fixed_disp_op.adjoint(template)
+
+# Show results
+deform_templ_fixed_disp.show('Deformed template')
+fixed_disp_adj.show('Adjoint of operator applied to template')
--- a/examples/deform/linearized_fixed_displacement.py
+++ b/examples/deform/linearized_fixed_displacement.py
@@ -34,13 +34,13 @@ import odl
 discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
 
 # Create a product space for displacement field
-disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
+disp_field_space = discr_space.vector_field_space
 
-# Create a template
+# Create a template, and set the template as an offset rectangle
 template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
 
-# Define a displacement field that shifts an image by hx
-# to the right and by hy upwards. Then it rotates clockwise by theta.
+# Define a displacement field that rotates an image clockwise by theta. Then it
+# shifts the image by hx to the right and by hy upwards.
 hx = -0.5
 hy = -0.5
 theta = np.pi / 4
@@ -64,9 +64,9 @@ fixed_disp_op = odl.deform.LinDeformFixedDisp(disp_field)
 # Apply the deformation operator to get the deformed template
 deform_templ_fixed_disp = fixed_disp_op(template)
 
-# Compute the adjoint of the operator
+# Compute the adjoint of the operator applied to the template
 fixed_disp_adj = fixed_disp_op.adjoint(template)
 
 # Show results
 deform_templ_fixed_disp.show('Deformed template')
-fixed_disp_adj.show('Adjoint of operator applied to template')
+fixed_disp_adj.show('Adjoint applied to template')
--- a/examples/deform/linearized_fixed_displacement.py
+++ b/examples/deform/linearized_fixed_displacement.py
@@ -19,9 +19,9 @@
 
 The linearized deformation operator with fixed displacement ``v`` maps
 a given template ``I`` to the function ``x --> I(x + v(x))``.
-This operator is linear.
+This operator is linear (in ``I``).
 
-Here, we consider a 2D example, where the displacement field ``v``
+This example considers a 2D case, where the displacement field ``v``
 is a Gaussian in each component, with positive sign in the first and
 negative sign in the second component. Note that in the deformed image,
 the value at ``x`` is **taken from** the original image at ``x + v(x)``,
@@ -42,9 +42,9 @@ import odl
 
 # Template space: discretized functions on the rectangle [-1, 1]^2 with
 # 100 samples per dimension.
-templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100))
 
-# We use a rectangle as template
+# The template is a rectangle of size 1.0 x 0.5
 template = odl.phantom.cuboid(templ_space, [-0.5, -0.25], [0.5, 0.25])
 
 # Create a product space for displacement field
--- a/examples/deform/linearized_fixed_template.py
+++ b/examples/deform/linearized_fixed_template.py
@@ -34,13 +34,13 @@ import odl
 discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
 
 # Create a product space for displacement field
-disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
+disp_field_space = discr_space.vector_field_space
 
-# Create a template
+# Create a template, and set the template as an offset rectangle
 template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
 
-# Define a displacement field that shifts an image by hx
-# to the right and by hy upwards. Then it rotates clockwise by theta.
+# Define a displacement field that rotates an image clockwise by theta. Then it
+# shifts the image by hx to the right and by hy upwards.
 hx = -0.5
 hy = -0.5
 theta = np.pi / 4
--- a/examples/deform/linearized_fixed_template.py
+++ b/examples/deform/linearized_fixed_template.py
@@ -20,7 +20,7 @@
 The linearized deformation operator with fixed template (image) ``I`` maps
 a given displacement field ``v`` to the function ``x --> I(x + v(x))``.
 
-Here, we consider a 2D example, where the displacement field ``v``
+This example consider a 2D case, where the displacement field ``v``
 is a Gaussian in each component, with positive sign in the first and
 negative sign in the second component. Note that in the deformed image,
 the value at ``x`` is **taken from** the original image at ``x + v(x)``,
@@ -28,8 +28,8 @@ hence the values are moved by ``-v(x)`` when comparing deformed and
 original templates.
 
 The derivative and its adjoint are based on the deformation of the
-gradient of the template, hence we expect to see some kind of edge image
-and an "edge vector field".
+gradient of the template, hence the result is expected to be some kind of
+edge image or "edge vector field", respectively.
 """
 
 import numpy as np
@@ -40,10 +40,11 @@ import odl
 
 
 # Template space: discretized functions on the rectangle [-1, 1]^2 with
-# 100 samples per dimension.
+# 100 samples per dimension. Usage of 'linear' interpolation ensures that
+# the template gradient is well-defined.
 templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
 
-# We use a rectangle as template
+# The template is a rectangle of size 1.0 x 0.5
 template = odl.phantom.cuboid(templ_space, [-0.5, -0.25], [0.5, 0.25])
 
 # Create a product space for displacement field
--- a/examples/deform/linearized_fixed_template.py
+++ b/examples/deform/linearized_fixed_template.py
@@ -1,0 +1,84 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using the operator of fixed-template linearized deformation."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+import odl
+
+
+# --- Create template and displacement field --- #
+
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-1, 1]^2 with 100 samples per dimension.
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# Create a product space for displacement field
+disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
+
+# Create a template
+template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
+
+# Define a displacement field that shifts an image by hx
+# to the right and by hy upwards. Then it rotates clockwise by theta.
+hx = -0.5
+hy = -0.5
+theta = np.pi / 4
+disp_func = [lambda x: (np.cos(theta) - 1) * x[0] - np.sin(theta) * x[1] + hx,
+             lambda x: np.sin(theta) * x[0] + (np.cos(theta) - 1) * x[1] + hy]
+
+# Create a displacement field based on ``disp_func``
+disp_field = disp_field_space.element(disp_func)
+
+# Show template and displacement field
+template.show('Template')
+disp_field.show('Displacement field')
+
+
+# --- Example of LinDeformFixedTempl and its derivative,
+# and the adjoint of the derivative --- #
+
+
+# Define the deformation operator where template is fixed
+fixed_templ_op = odl.deform.LinDeformFixedTempl(template)
+
+# Apply the deformation operator to get the deformed template
+deform_templ_fixed_templ = fixed_templ_op(disp_field)
+
+# Define the derivative of the deformation operator
+fixed_templ_deriv_op = fixed_templ_op.derivative(disp_field)
+
+# Evaluate the derivative at the vector field that is only 1. This should be
+# the same as the pointwise inner product between linearly deformed
+# gradient and the said vector field
+vector_field = disp_field_space.one()
+fixed_templ_deriv = fixed_templ_deriv_op(vector_field)
+
+# Evaluate the adjoint of derivative at the element that is only 1
+func = discr_space.one()
+fixed_templ_adj = fixed_templ_deriv_op.adjoint(func)
+
+# Show results
+deform_templ_fixed_templ.show('Deformed template')
+fixed_templ_deriv.show('Operator derivative applied to one()')
+fixed_templ_adj.show('Adjoint of operator derivative applied to one()')
--- a/examples/deform/linearized_fixed_template.py
+++ b/examples/deform/linearized_fixed_template.py
@@ -15,12 +15,22 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Example using the operator of fixed-template linearized deformation."""
+"""Example using the operator of fixed-template linearized deformation.
 
-# Imports for common Python 2/3 codebase
-from __future__ import print_function, division, absolute_import
-from future import standard_library
-standard_library.install_aliases()
+The linearized deformation operator with fixed template (image) ``I`` maps
+a given displacement field ``v`` to the function ``x --> I(x + v(x))``.
+
+Here, we consider a 2D example, where the displacement field ``v``
+is a Gaussian in each component, with positive sign in the first and
+negative sign in the second component. Note that in the deformed image,
+the value at ``x`` is **taken from** the original image at ``x + v(x)``,
+hence the values are moved by ``-v(x)`` when comparing deformed and
+original templates.
+
+The derivative and its adjoint are based on the deformation of the
+gradient of the template, hence we expect to see some kind of edge image
+and an "edge vector field".
+"""
 
 import numpy as np
 import odl
@@ -29,25 +39,24 @@ import odl
 # --- Create template and displacement field --- #
 
 
-# Discrete reconstruction space: discretized functions on the rectangle
-# [-1, 1]^2 with 100 samples per dimension.
-discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+# Template space: discretized functions on the rectangle [-1, 1]^2 with
+# 100 samples per dimension.
+templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# We use a rectangle as template
+template = odl.phantom.cuboid(templ_space, [-0.5, -0.25], [0.5, 0.25])
 
 # Create a product space for displacement field
-disp_field_space = discr_space.vector_field_space
+disp_field_space = templ_space.vector_field_space
 
-# Create a template, and set the template as an offset rectangle
-template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
+# Define a displacement field that bends the template a bit towards the
+# upper left. We use a list of 2 functions and discretize it using the
+# disp_field_space.element() method.
+sigma = 0.5
+disp_func = [
+    lambda x: 0.4 * np.exp(-(x[0] ** 2 + x[1] ** 2) / (2 * sigma ** 2)),
+    lambda x: -0.3 * np.exp(-(x[0] ** 2 + x[1] ** 2) / (2 * sigma ** 2))]
 
-# Define a displacement field that rotates an image clockwise by theta. Then it
-# shifts the image by hx to the right and by hy upwards.
-hx = -0.5
-hy = -0.5
-theta = np.pi / 4
-disp_func = [lambda x: (np.cos(theta) - 1) * x[0] - np.sin(theta) * x[1] + hx,
-             lambda x: np.sin(theta) * x[0] + (np.cos(theta) - 1) * x[1] + hy]
-
-# Create a displacement field based on ``disp_func``
 disp_field = disp_field_space.element(disp_func)
 
 # Show template and displacement field
@@ -55,30 +64,27 @@ template.show('Template')
 disp_field.show('Displacement field')
 
 
-# --- Example of LinDeformFixedTempl and its derivative,
-# and the adjoint of the derivative --- #
+# --- Apply LinDeformFixedTempl, derivative and its adjoint --- #
 
 
-# Define the deformation operator where template is fixed
-fixed_templ_op = odl.deform.LinDeformFixedTempl(template)
+# Initialize the deformation operator with fixed template
+deform_op = odl.deform.LinDeformFixedTempl(template)
 
-# Apply the deformation operator to get the deformed template
-deform_templ_fixed_templ = fixed_templ_op(disp_field)
+# Apply the deformation operator to get the deformed template.
+deformed_template = deform_op(disp_field)
 
-# Define the derivative of the deformation operator
-fixed_templ_deriv_op = fixed_templ_op.derivative(disp_field)
+# Initialize the derivative of the deformation operator at the
+# given displacement field. The result is again an operator.
+deform_op_deriv = deform_op.derivative(disp_field)
 
-# Evaluate the derivative at the vector field that is only 1. This should be
-# the same as the pointwise inner product between linearly deformed
-# gradient and the said vector field
-vector_field = disp_field_space.one()
-fixed_templ_deriv = fixed_templ_deriv_op(vector_field)
+# Evaluate the derivative at the vector field that has value 1 everywhere,
+# i.e. the global shift by (-1, -1).
+deriv_result = deform_op_deriv(disp_field_space.one())
 
-# Evaluate the adjoint of derivative at the element that is only 1
-func = discr_space.one()
-fixed_templ_adj = fixed_templ_deriv_op.adjoint(func)
+# Evaluate the adjoint of derivative at the image that is 1 everywhere.
+deriv_adj_result = deform_op_deriv.adjoint(templ_space.one())
 
 # Show results
-deform_templ_fixed_templ.show('Deformed template')
-fixed_templ_deriv.show('Operator derivative applied to one()')
-fixed_templ_adj.show('Adjoint of operator derivative applied to one()')
+deformed_template.show('Deformed template')
+deriv_result.show('Operator derivative applied to one()')
+deriv_adj_result.show('Adjoint of the derivative applied to one()')
--- a/odl/deform/linearized.py
+++ b/odl/deform/linearized.py
@@ -23,8 +23,9 @@ from future import standard_library
 standard_library.install_aliases()
 
 from odl.operator.operator import Operator
-from odl import ProductSpace, DiscreteLp, Gradient, Divergence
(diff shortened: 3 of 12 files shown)


adler-j:
Cleaned up commit history and implementation, also rebased to master. Should be reviewable.

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+        # TODO allow users to select what method to use here.
+        grad = Gradient(domain=self.range, method='central',
+                        padding_method='symmetric')
+        grad_templ = grad(self._template)
+        def_grad = self.domain.element(
+            [_linear_deform(gf, displacement) for gf in grad_templ])
+
+        return PointwiseInner(self.domain, def_grad)
+
+    def __repr__(self):
```
chongchenmath: What is this mean?

adler-j: `repr` is short for representation and is intended to give a representation of the object. It is called when you write things like

```python
>>> space = odl.rn(3)
>>> space
rn(3)  # Here repr is called
```

and can also be explicitly invoked through the `repr` function

```python
>>> space = odl.rn(3)
>>> print(repr(space))
rn(3)
```

You can read some more in the [official documentation](https://docs.python.org/3/library/functions.html#repr).

adler-j: Generally, you should be able to create the object from the repr string, but this is not always possible.

```python
>>> np.array([1, 2, 3, 4, 5])
array([1, 2, 3, 4, 5])  # can recerate the array from this
>>> np.eye(5)
array([[ 1.,  0.,  0.,  0.,  0.],
       [ 0.,  1.,  0.,  0.,  0.],
       [ 0.,  0.,  1.,  0.,  0.],
       [ 0.,  0.,  0.,  1.,  0.],
       [ 0.,  0.,  0.,  0.,  1.]])  # can still recreate the array
>>> np.eye(50)
array([[ 1.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  1.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  1., ...,  0.,  0.,  0.],
       ..., 
       [ 0.,  0.,  0., ...,  1.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  1.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  1.]])  # array too big, cannot recreate it
```

ODL generally works the same:

```python
>>> odl.vector([1., 2, 3, 4, 5])
```

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+        displacement = self.domain.element(displacement)
+
+        # TODO allow users to select what method to use here.
+        grad = Gradient(domain=self.range, method='central',
+                        padding_method='symmetric')
+        grad_templ = grad(self._template)
+        def_grad = self.domain.element(
+            [_linear_deform(gf, displacement) for gf in grad_templ])
+
+        return PointwiseInner(self.domain, def_grad)
```
chongchenmath: What a simplified and clever implementation! Question: the call of adjoint `PointwiseInnerAdjoint` here is more efficient than before?

adler-j: I think that the call should take exactly the same time as before, down to minor things like input checking. It may actually be a bit slower now since we have more rigorous checks in place.

odl/deform/linearized.py
chongchenmath:
```
discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
fixed_templ_op = odl.deform.LinDeformFixedTempl(template)
```

Without explicit `__repr__()`
```
>>> fixed_templ_op
LinDeformFixedTempl: ProductSpace(uniform_discr([-1.0, -1.0], [1.0, 1.0], [100, 100],
interp=u'linear'), 2) -> uniform_discr([-1.0, -1.0], [1.0, 1.0], [100, 100], interp=u'linear')
```

With explicit `__repr__()`

```
>>> fixed_templ_op
LinDeformFixedTempl(uniform_discr([-1.0, -1.0], [1.0, 1.0], [100, 100], interp=u'linear')
.element([[0.0, 0.0, 0.0, ..., 0.0, 0.0, 0.0],
 [0.0, 0.0, 0.0, ..., 0.0, 0.0, 0.0],
 [0.0, 0.0, 0.0, ..., 0.0, 0.0, 0.0],
 ...,
 [0.0, 0.0, 0.0, ..., 0.0, 0.0, 0.0],
 [0.0, 0.0, 0.0, ..., 0.0, 0.0, 0.0],
 [0.0, 0.0, 0.0, ..., 0.0, 0.0, 0.0]]))
```

It seems the former is more clear. Maybe I am wrong.

adler-j: While you are correct in some sense, it depends on your definition of clear. In python there are two "print" statements, `__str__` and `__repr__`. `__str__` is intended to give a short representation of the object so users know what is meant, while `__repr__` should be used for debugging and be more precise.

Here I only implemented `__repr__`, mostly because I was lazy. We could want to implement `__str__` as well however.

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+        --------
+        LinDeformFixedDisp : Deformation with a fixed displacement.
+        """
+        if domain is None:
+            if not isinstance(template, DiscreteLpVector):
+                raise TypeError('`template` must be a `DiscreteLpVector`'
+                                'instance if `domain` is None, got {!r}'
+                                ''.format(template))
+
+            domain = template.space.tangent_space
```
chongchenmath: In differential geometry, one can attach to `every point x` of a differentiable manifold `a tangent space`. Here we attach `a tangent_space` to `a space`. It seems using `space.tangent_space` is not exact. Probably ` tangent bundle`?

adler-j: Since the space is flat they should be equivalent, but sure, the naming here is not perfect. We may need some more thought on it.

chongchenmath: Yes. It's quite a geometric concept here. How about `vector filed space`, short for `vf_space` or more explicit name?

ozanoktem: The source for the discussion regarding the notion of tangent space is that ODL lacks a notion of manifold (curved space). In ODL one has either a set or a vector space, the notion of a. manifold does not exist. We have already encountered the need to handle manifolds since data space in tomography is a non-Euclidian manifold. The operations required for points (=tomographic data) on such manifolds could however be rephrased as operations on the underlying coordinate space (products of intervals), ignoring the fact that a single coordinate space is insufficient as an atlas. Now, dealing with deformations requires us yet again to handle points (=vector fields) in curved spaces, i.e. manifolds. One option is to try to recast the operations to the underlying coordinate spaces, e.g. assuming a RKHS. Another is to introduce notions manifold in ODL with a tangent bundle. The latter seems to be quite time consuming though.

kohr-h: An option, and I think that's what @adler-j was trying to achieve, is to represent the space for the vector fields directly, without having the notion of manifolds. That's of course a bit limiting, as for the manifolds in acquisition geometries, but from my perspective it's a sufficient solution and can be handled quite smoothly with the means we have currently. I don't think we necessarily need RHKS to do that.

ozanoktem: I don't get this discussion about manifolds, tangent spaces etc! The deformations, e.g. diffeomorphisms, are the ones that reside in a manifold. The only operations we will do with  deformations is to let them act on an image and to compose deformations to form new ones. That's it. All vector space operations are done on the set of vector fields that generate the deformations. So, why on earth do we need to book keep the relation in ODL between the set (semi-group) of deformations and the associated vector space of vector fields?

adler-j: Well for the fixed template style deformations we need a proper domain for the deformations., and in the fixed deformation case we need to be able to give it as input. Basically we need a name for the space

```python
ProductSpace(space.astype(space.real_dtype), space.ndim)
```

I picked `tangent_space` since it felt natural, but as @chongchenmath noted this may not be future proof since it is only correct for flat spaces.

@kohr-h was really on spot on what we try to achieve here. RKHS has nothing to do with this one and probably shouldn't need to be involved for now.

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+    field ``v`` to the new function ``x --> I(x + v(x))``.
+    """
+
+    def __init__(self, template, domain=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        template : `DiscreteLpVector` or element-like
+            Fixed template that is to be deformed.
```
adler-j: The should be adjusted so it fits inside the 80 chars, currently it is too narrow

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+            If ``domain`` is not given, ``template`` must
+            be a `DiscreteLpVector`, and the domain of this operator
+            is inferred from ``template.space``. If ``domain`` is
+            given, ``template`` can be anything that is understood
+            by the ``domain[0].element()`` method.
+        domain : product space of `DiscreteLp`, optional
```
adler-j: `ProductSpace` of ...

odl/deform/linearized.py
ozanoktem: Are we talking about the set of deformations or the set of vector/velocity fields generating deformations? The set of deformations is not a vector space, it is a semi group and a Lie group at best. The set of vector/velocity fields generating deformations is however a vector space, which in some theoretical frameworks can be interpreted as the tangent space to the Lie group of deformations. RKHS was mentioned as an example of how this Lie group could be equipped coordinates. So, what are we trying to name here?

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+            Space of displacement fields on which this operator
+            acts, i.e. the operator domain. If not given, it is
```
adler-j: "If not given, it is..." should be replaced by a "Default: "

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+            is inferred from ``displacement[0].space``. If ``domain`` is
+            given, ``displacement`` can be anything that is understood
+            by the ``domain.tangent_space.element()`` method.
+        domain : `DiscreteLp`, optional
+            Space of templates on which this operator acts, i.e. the operator
+            domain. If not given, ``displacement[0].space`` is used as domain.
+
+        Examples
+        --------
+        Create a given deformation and use it to deform a function.
```
adler-j: function -> template

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+        points, 0.1, we expect to get the mean of the values.
+
+        >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
+        >>> disp_field = space.tangent_space.element([[0, 0, 0, -0.1, 0]])
+        >>> op = LinDeformFixedDisp(disp_field)
+        >>> template = [0, 0, 1, 0, 0]
+        >>> print(op(template))
+        [0.0, 0.0, 1.0, 0.5, 0.0]
+
+        See Also
```
adler-j: This See Also should be in the class documentation, not in the init doc

odl/deform/linearized.py
```diff
@@ -0,0 +1,325 @@
...
+        >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
+        >>> template = space.element([0, 0, 1, 0, 0])
+        >>> op = LinDeformFixedTempl(template)
+        >>> disp_field = [[0, 0, 0, -0.1, 0]]
+        >>> print(op(disp_field))
+        [0.0, 0.0, 1.0, 0.5, 0.0]
+
+        See Also
+        --------
+        LinDeformFixedDisp : Deformation with a fixed displacement.
```
adler-j: This See Also should be in the class documentation, not in the init doc

odl/deform/linearized.py
adler-j: > Are we talking about the set of deformations or the set of vector/velocity fields generating deformations?

We're talking about "the set of vector/velocity fields generating deformations".

chongchenmath: As I suggested, one abbreviation of  `vector field space `  is probably a proper choice here.

adler-j: Agree, something along those lines would be good.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
```diff
@@ -0,0 +1,323 @@
...
+    """Linearized deformation of a template with a displacement field.
+
+    The function maps a gived template ``I`` and a given displacement
+    field ``v`` to the new function ``x --> I(x + v(x))``.
+
+    Parameters
+    ----------
+    template : `DiscreteLpVector`
+        Template to be deformed by a displacement field.
+    displacement : `ProductSpace` element
```
adler-j: ProductSpace of DiscreteLp element

odl/deform/linearized.py
```diff
@@ -0,0 +1,323 @@
...
+    --------
+    LinDeformFixedTempl : Deformation with a fixed template.
+    """
+
+    def __init__(self, displacement, domain=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        displacement : `ProductSpace` element-like
```
adler-j: ProductSpace of DiscreteLp

odl/deform/linearized.py
```diff
@@ -0,0 +1,323 @@
...
+        else:
+            if not isinstance(domain, DiscreteLp):
+                raise TypeError('`displacement[0]` {!r} not an `DiscreteLp`'
+                                ''.format(displacement[0]))
+
+            displacement = domain.vector_field_space.element(displacement)
+
+        Operator.__init__(self, domain, domain, linear=True)
+
+        self._displacement = displacement
```
adler-j: We should add an accessor for this.

adler-j: That is, renqme to `__displacement` and add a property `displacement`

odl/deform/linearized.py
```diff
@@ -0,0 +1,323 @@
...
+        """Implementation of ``self(template)``.
+
+        Parameters
+        ----------
+        template : `domain` element
+            Given template that is to be deformed by the fixed
+            displacement field.
+        """
+        return _linear_deform(template, self._displacement, out)
+
```
adler-j: Short comment here, don't we have a formula for the inverse operator as well? Perhaps it should be added.

odl/discr/lp_discr.py
```diff
@@ -228,6 +234,12 @@ def exponent(self):
         """The exponent ``p`` in ``L^p``."""
         return self._exponent
 
+    @property
+    def vector_field_space(self):
+        """vector field space."""
+        real_space = self.astype(self._real_dtype)
+        return ProductSpace(real_space, self.ndim)
```
adler-j: Apparently we have a `real_space` property so this could read `ProductSpace(self.real_space, self.ndim)`

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-1, 1]^2 with 100 samples per dimension.
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# Create a product space for displacement field
+disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
+
+# Create a template
```
adler-j: A word on what the template is:

"Set template as an offset rectangle"

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
+
+# Define a displacement field that shifts an image by hx
+# to the right and by hy upwards. Then it rotates clockwise by theta.
```
adler-j: Is this correct? I was to the understanding that it first rotates by theta and then shifts.

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Apply the deformation operator to get the deformed template
+deform_templ_fixed_disp = fixed_disp_op(template)
+
+# Compute the adjoint of the operator
+fixed_disp_adj = fixed_disp_op.adjoint(template)
+
+# Show results
+deform_templ_fixed_disp.show('Deformed template')
+fixed_disp_adj.show('Adjoint of operator applied to template')
```
adler-j: "Adjoint applied to template" should suffice

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+# --- Example of LinDeformFixedDisp and its adjoint --- #
+
+
+# Define the deformation operator where the displacement field is fixed
+fixed_disp_op = odl.deform.LinDeformFixedDisp(disp_field)
+
+# Apply the deformation operator to get the deformed template
+deform_templ_fixed_disp = fixed_disp_op(template)
+
+# Compute the adjoint of the operator
```
adler-j: "... applied at the template"

perhaps also a word on what the adjoint is expected to do

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+
+# --- Create template and displacement field --- #
+
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-1, 1]^2 with 100 samples per dimension.
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# Create a product space for displacement field
+disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
```
adler-j: Remove this and simply use `discr_space.tangent_space` instead.

examples/deform/linearized_fixed_template.py
```diff
@@ -0,0 +1,84 @@
...
+
+# --- Create template and displacement field --- #
+
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-1, 1]^2 with 100 samples per dimension.
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# Create a product space for displacement field
+disp_field_space = odl.ProductSpace(discr_space, discr_space.ndim)
```
adler-j: Remove this and simply use `discr_space.tangent_space` instead.

odl/deform/linearized.py
```diff
@@ -0,0 +1,323 @@
...
+    displacement : `ProductSpace` element
+        The vector field (displacement field) used in the linearized
+        deformation.
+
+    Returns
+    -------
+    deformed_template : `numpy.ndarray`
+
+    Examples
+    --------
```
adler-j: This example seems to have lost the text explaining it.

adler-j:
Did a further short review of this. We should focus on getting this in asap since it is needed for other things to progress.

odl/deform/linearized.py
chongchenmath: done.

odl/deform/linearized.py
chongchenmath: `ProductSpace` of `DiscreteLp` element-like.

odl/deform/linearized.py
chongchenmath: `self.__displacement = displacement`

odl/deform/linearized.py
chongchenmath: Nice comment. For the linearized case, the inverse operator can be approximated by `_linear_deform(template, -self.__displacement, out)`. Perhaps we needn't to give it explicitly.

odl/discr/lp_discr.py
chongchenmath: Changed into `self.real_space` in lines 240 and 241.

examples/deform/linearized_fixed_displacement.py
chongchenmath: Added.

examples/deform/linearized_fixed_displacement.py
chongchenmath: Simplified.

examples/deform/linearized_fixed_displacement.py
chongchenmath: `... applied to the template`

odl/deform/linearized.py
```diff
@@ -0,0 +1,323 @@
...
+    >>> space = odl.uniform_discr(0, 1, 5)
+    >>> disp_field_space = odl.ProductSpace(space, space.ndim)
+    >>> template = space.element([0, 0, 1, 0, 0])
+    >>> displacement_field = disp_field_space.element([[0, 0, 0, -0.2, 0]])
+    >>> _linear_deform(template, displacement_field)
+    array([ 0.,  0.,  1.,  1.,  0.])
+    """
+    image_pts = template.space.points()
+    for i, vi in enumerate(displacement):
+        image_pts[:, i] += vi.asarray().ravel()
```
chongchenmath: Changed.

odl/deform/linearized.py
chongchenmath: Added.

chongchenmath:
Merge?

kohr-h:
> Merge?

Not yet. Tests are failing, and the code is not completely reviewed yet. I'll check it asap.

chongchenmath:
The problem is `test/solvers/advanced/douglas_rachford_test.py:109: AssertionError`.

kohr-h:
> The problem is test/solvers/advanced/douglas_rachford_test.py:109: AssertionError.

Okay, that's not your fault :-)
Anyway, I'll go through the code again.

odl/discr/lp_discr.py
```diff
@@ -228,6 +234,12 @@ def exponent(self):
         """The exponent ``p`` in ``L^p``."""
         return self._exponent
 
+    @property
+    def vector_field_space(self):
+        """vector field space."""
+        self.real_space = self.astype(self._real_dtype)
+        return ProductSpace(self.real_space, self.ndim)
```
kohr-h: I think we should generalize this (and simplify the code). There is nothing special about vector fields, which are just rank-1 tensors. We should allow arbitrary rank. In practice, this means that we make this a method with a `rank=1` optional parameter and raise `NotImplementedError` for anything else than 0 or 1 right now. That's not great, but at least the change to general tensors won't break API then.

odl/deform/linearized.py
```diff
@@ -0,0 +1,342 @@
...
+            [_linear_deform(gf, displacement) for gf in grad_templ])
+
+        return PointwiseInner(self.domain, def_grad)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        if self.domain == self._template.space.vector_field_space:
+            domain_repr = ''
+        else:
+            domain_repr = ', domain={!r}'.format(self.domain)
```
kohr-h: Will this case ever occur? If `domain` is not given during init, we are always in the first case. If it is given, we know that `template in domain[0]` by construction. So when would the condition in the `if` part not hold?

adler-j: I suppose it technically could be that the spaces have different `dspaces`, thus representing the same thing but in slightly different ways.

With that said, the proper way of removing this (if we want to, and I'd be fine with it) would be to fully remove the `domain` parameter to the `__init__` method.

odl/discr/lp_discr.py
adler-j: Well that makes the naming even more of a nightmare, what would you call such a method?

Also are we sure things generalize nicely to tensors?

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using the operator of fixed-displacement linearized deformation."""
```
adler-j: Could use some more text. How is it used? I feel like the text length for the solvers examples is about right, for reference.

examples/deform/linearized_fixed_template.py
```diff
@@ -0,0 +1,84 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using the operator of fixed-template linearized deformation."""
```
adler-j: This text could also be somewhat longer.

examples/deform/linearized_fixed_template.py
```diff
@@ -0,0 +1,84 @@
...
+
+import numpy as np
+import odl
+
+
+# --- Create template and displacement field --- #
+
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-1, 1]^2 with 100 samples per dimension.
```
adler-j: Also a word on why we use `interp='linear'`

examples/deform/linearized_fixed_template.py
```diff
@@ -0,0 +1,84 @@
...
+# Define the deformation operator where template is fixed
+fixed_templ_op = odl.deform.LinDeformFixedTempl(template)
+
+# Apply the deformation operator to get the deformed template
+deform_templ_fixed_templ = fixed_templ_op(disp_field)
+
+# Define the derivative of the deformation operator
+fixed_templ_deriv_op = fixed_templ_op.derivative(disp_field)
+
+# Evaluate the derivative at the vector field that is only 1. This should be
```
adler-j: This string is unclear and could use improvement.

* `the vector field that is only 1` could perhaps be written `at the vector field which shifts all points by 1 in both directions`, or something like that.
* `should be the same as the pointwise inner product ...` is (i.m.o.) way to complicated for an example. It would be better to focus on the intuition of what it actually means.

odl/deform/linearized.py
```diff
@@ -0,0 +1,342 @@
...
+"""Operators and functions for linearized deformation."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+from odl.operator.operator import Operator
+from odl import ProductSpace, DiscreteLp, Gradient, Divergence
+from odl import PointwiseInner, DiscreteLpVector
```
adler-j: `from odl import` repeated, should use parenthesis instead.

odl/discr/lp_discr.py
kohr-h: I think it's rather straightforward - the gradient is an element of `X^d`, the Hessian (second derivative) in `X^(d x d)`, ..., the k-th derivative in `X^(d x ... x d)` with `k` factors.
Of course, another option `order` would be required...

Naming wouldn't be that horrible, like `tensor_field_space`.

A compromise could be to keep the property above and change the implementation to `return self.tensor_field_space(rank=1)` later.

odl/deform/linearized.py
```diff
@@ -0,0 +1,342 @@
...
+            inferred from ``displacement[0].space``. If ``domain`` is given,
+            ``displacement`` can be anything that is understood by the
+            ``domain.vector_field_space.element()`` method.
+        domain : `DiscreteLp`, optional
+            Space of templates on which this operator acts, i.e. the operator
+            domain. Default: ``displacement[0].space`` is used as domain.
+
+        Examples
+        --------
+        Create a given displacement field and use it to deform a template.
```
adler-j: ~~given~~

that word doesn't add much.

odl/deform/linearized.py
```diff
@@ -0,0 +1,342 @@
...
+        ----------
+        template : `domain` element
+            Given template that is to be deformed by the fixed
+            displacement field.
+        """
+        return _linear_deform(template, self.__displacement, out)
+
+    @property
+    def adjoint(self):
+        """Adjoint of the operator.
```
adler-j: Could be nice to write a few words here on what the adjoint is, like 2 rows or so.

odl/deform/linearized.py
kohr-h: > I suppose it technically could be that the spaces have different dspaces, thus representing the same thing but in slightly different ways.

I don't quite follow. If `template in domain[0]`, it means that `template.space == domain[0]`, so how can possibly `domain` be different from `template.space.vector_field_space`?



> With that said, the proper way of removing this (if we want to, and I'd be fine with it) would be to fully remove the domain parameter to the __init__ method.

That actually makes sense. If the template is not a `DiscreteLpVector`, you still need to give the `domain`, which is probably more work than simply making the template an element of the appropriate space.

odl/deform/linearized.py
chongchenmath: Adjoint of the linear operator that maps a template ``I`` to the new function ``x --> exp(-div(v(x))) * I(x - v(x))``.

odl/deform/linearized.py
chongchenmath: Done.

odl/deform/linearized.py
chongchenmath: Changed.

odl/deform/linearized.py
```diff
@@ -0,0 +1,342 @@
...
+        -------
+        adjoint: `Operator`
+            The adjoint of the operator.
+        """
+
+        # TODO allow users to select what method to use here.
+        div_op = Divergence(domain=self.__displacement.space, method='forward',
+                            padding_method='symmetric')
+        jacobian_det = self.domain.element(
+            np.exp(-div_op(self.__displacement)))
```
kohr-h: Would be nice to have a reference for this approximation.

chongchenmath: The Algorithm 2 in: M. Bauer, S. Joshi and K. Modin, Diffeomorphic density matching by optimal information transport, SIAM J. Imaging Sci., 8(3), 1718–1751.

adler-j: That cannot be the first nor the most accessible reference, no?

chongchenmath: That is the first order approximation of |D((Id + v)^{-1})|*I((Id + v)^{-1}), which is enough precise if v is close to 0. It is almost the same as the preceding reference.

kohr-h: That reference uses such an approximation, but the exponential term does not (at least not obviously) come from the approximation of `D(id + v)^(-1)`.
In fact, as I read it, the inverse deformation itself is simply approximated as `x --> x - v(x)` in the argument of the template. The exponential only plays a role for the Jacobian, but I don't see how. They write something about a "Lie-Trotter splitting", but to my understanding, this requires that the diffeo `psi^(-1)` already is of the form `exp(-eps*v)`, and only then it follows that the Jacobian can be approximated as `exp(-eps*div(v))`, not in general.

I'm a bit doubtful that we're doing something reasonable at all here, a clear reference or a verification would be very useful.

kohr-h: Of course, one can argue that `id + v` is about the same as `exp(v)` for small `v`, so we can assume that the same step as in the algorithm in the mentioned paper is valid here, too. It's just a bit too shaky for my taste.

chongchenmath: Anyway, what we need to compute here is  ``|D((Id + v)^{-1})|*I((Id + v)^{-1})``. The ``Id - v`` is an approximation of ``Id + v``. Then the ``1 -div(v)`` is the Jacobian determinant of ``Id - v``, i.e., an approximation of ``|D((Id + v)^{-1})|``. I ever used that here. But inspired by the above reference, ``exp(-div(v))`` is also an approximation. And the ``1 -div(v)`` is just the first order Taylor expansion of ``exp(-div(v))``.

chongchenmath: I ever discussed with @adler-j about which one we should use. The final conclusion is that using ``exp(-div(v))`` is better.

kohr-h: I remember that, yes, the exponential ensures positivity so things at least don't crash. For the more elementary part one can argue as follows:

Let `A(x) = x + v(x)`. Then `DA(x) = id + Dv(x)` and (sorry for the illegible notation)
```
det( id + Dv(x) ) = 1   +   d/dt det( id + t * Dv(x) ) | t=0   +   O( |D^2 v(x)|^2 )
                  ~ 1 + tr( adj(id) * Dv(x) ) = 1 + tr( Dv(x) ) = 1 + div v(x)
```
according to [Jacobi's formula for the derivative of the determinant](https://en.wikipedia.org/wiki/Jacobi%27s_formula). So in that part we're safe, and the switch between `1 + something` and `exp(something)` is the only hand-waving part.

adler-j: The exponential also guarantees the existence of inverses. Without it, if you first translate by `v` and then by `-v` you end up with a jacobian of `(1 + div v) * (1 - div v) = 1 - (div v)^2 \neq 1`. With the jacobian you get `exp(div v) * exp(- div(v)) = 1`.

Except for that, the swap from `1 + something` to `exp(something`, is totally hand-waving and perhaps we should be clear about this in a comment.

chongchenmath: So the ``exp(something)`` is more robust for general case.

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,72 @@
...
+discr_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# Create a product space for displacement field
+disp_field_space = discr_space.vector_field_space
+
+# Create a template, and set the template as an offset rectangle
+template = odl.phantom.cuboid(discr_space, [-0.5, -0.5], [-0.25, 0])
+
+# Define a displacement field that rotates an image clockwise by theta. Then it
+# shifts the image by hx to the right and by hy upwards.
```
kohr-h: This example is good for educational purposes and to show that we can do rigid body motions, but for users expecting some action, it's rather boring. Why not use some simple Gaussian that tapers off fast enough?

kohr-h:
So I went through the code and changed some stuff, see the second last commit. I also cleaned up the commit history a bit and rebased on master. Should be ready for final review and then it's time to merge.

odl/deform/linearized.py
```diff
@@ -73,7 +73,7 @@ def _linear_deform(template, displacement, out=None):
     >>> _linear_deform(template, displacement_field)
     array([ 0.,  0.,  1.,  1.,  0.])
 
-    The result depends on the chosen interpolation. If we chose 'linear'
+    The result depends on the chosen interpolation. If choose 'linear'
```
kohr-h: was correct before

odl/deform/linearized.py
```diff
@@ -121,6 +121,8 @@ class LinDeformFixedTempl(Operator):
         W_I'(v)(u) = \\big< \\nabla I(\cdot + v(\cdot)),
         u \\big>_{\mathbb{R}^d},
 
+    i.e., :math:`grad(I)(x + v(x)).T u(x)`,
```
kohr-h: Use `\\nabla I` in math mode, also `^T` instead of `.T`.

odl/deform/linearized.py
```diff
@@ -243,7 +245,9 @@ class LinDeformFixedDisp(Operator):
 
     .. math::
 
-        W_v^*(I) \\approx \exp(-\mathrm{div}\, v) \, I(\cdot + v(\cdot)).
+        W_v^*(I) \\approx \exp(-\mathrm{div}\, v) \, I(\cdot - v(\cdot)),
+
+    i.e., :math:`exp(-div(v(x))) * I(x - v(x))`.
```
kohr-h: `\exp`
`\mathrm{div}`
No `*` for multiplication

odl/deform/linearized.py
```diff
@@ -149,7 +151,7 @@ def __init__(self, template):
         >>> print(op(disp_field))
         [0.0, 0.0, 1.0, 1.0, 0.0]
 
-        The result depends on the chosen interpolation. If we chose 'linear'
+        The result depends on the chosen interpolation. If choose 'linear'
```
kohr-h: see above

odl/deform/linearized.py
```diff
@@ -276,7 +280,7 @@ def __init__(self, displacement, templ_space=None):
         >>> print(op([0, 0, 1, 0, 0]))
         [0.0, 0.0, 1.0, 1.0, 0.0]
 
-        The result depends on the chosen interpolation. If we chose 'linear'
+        The result depends on the chosen interpolation. If choose 'linear'
```
kohr-h: see above

odl/deform/linearized.py
```diff
@@ -114,17 +114,26 @@ class LinDeformFixedTempl(Operator):
...
     i.e., :math:`W_I'(v)(u)(x) = \\nabla I(x + v(x))^T u(x)`,
 
     which is to be understood as a point-wise inner product, resulting
-    in a function in :math:`X`.
+    in a function in :math:`X`. And the adjoint of the preceding derivative
+    is also an operator that maps :math:`X` into :math:`V`:
+
+    .. math::
+
+        W_I^*(v) : X \\to V, \quad W_I^*(v)(J) =
```
kohr-h: This should be `W_I'(v)^*`, i.e. adjoint of the derivative.

odl/deform/linearized.py
```diff
@@ -151,7 +160,7 @@ def __init__(self, template):
         >>> print(op(disp_field))
         [0.0, 0.0, 1.0, 1.0, 0.0]
 
-        The result depends on the chosen interpolation. If choose 'linear'
+        The result depends on the chosen interpolation. If we chose 'linear'
```
kohr-h: choose

odl/deform/linearized.py
```diff
@@ -280,7 +289,7 @@ def __init__(self, displacement, templ_space=None):
         >>> print(op([0, 0, 1, 0, 0]))
         [0.0, 0.0, 1.0, 1.0, 0.0]
 
-        The result depends on the chosen interpolation. If choose 'linear'
+        The result depends on the chosen interpolation. If we chose 'linear'
```
kohr-h: choose

odl/deform/linearized.py
chongchenmath: Is `we choose`?

kohr-h: Ah, sorry, it was also wrong before. Correct is `If we choose ...`.

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,82 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using the operator of fixed-displacement linearized deformation.
+
+The linearized deformation operator with fixed displacement ``v`` maps
+a given template ``I`` to the function ``x --> I(x + v(x))``.
+This operator is linear.
```
adler-j: linear in `I`

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,82 @@
...
+
+import numpy as np
+import odl
+
+
+# --- Create template and displacement field --- #
+
+
+# Template space: discretized functions on the rectangle [-1, 1]^2 with
+# 100 samples per dimension.
```
adler-j: Comment on interp='linear'

examples/deform/linearized_fixed_displacement.py
```diff
@@ -0,0 +1,82 @@
...
+templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# We use a rectangle as template
```
adler-j: We dont use the `We ...` style in comments. Here we could write `Use a rectangle as template` is more concise.

examples/deform/linearized_fixed_template.py
```diff
@@ -0,0 +1,90 @@
...
+
+import numpy as np
+import odl
+
+
+# --- Create template and displacement field --- #
+
+
+# Template space: discretized functions on the rectangle [-1, 1]^2 with
+# 100 samples per dimension.
```
adler-j: A word on interp='linear'

examples/deform/linearized_fixed_template.py
```diff
@@ -0,0 +1,90 @@
...
+templ_space = odl.uniform_discr([-1, -1], [1, 1], (100, 100), interp='linear')
+
+# We use a rectangle as template
```
adler-j: See above comment on We ...

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+
+    Returns
+    -------
+    deformed_template : `numpy.ndarray`
+        Function values of the deformed template. If ``out`` was given,
+        the returned object is a reference to it.
+
+    Examples
+    --------
+    We create a simple 1D template and a corresponding displacement
```
adler-j: ~~We~~

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+    field. Where the displacement is zero, we expect that the output
```
adler-j: ~~,~~

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+                            'instance, got {!r}'.format(displacement))
+        if not displacement.space.is_power_space:
+            raise TypeError('`displacement.space` must be a power space, '
+                            'got {!r}'.format(displacement.space))
+        if not isinstance(displacement[0].space, DiscreteLp):
+            raise TypeError('`displacement[0].space` must be a `DiscreteLp` '
+                            'instance, got {!r}'.format(displacement[0]))
+
+        if templ_space is None:
+            templ_space = displacement[0].space
```
adler-j: While equivalent, I feel it would be better to write `displacement.space[0]`

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+        >>> print(op(template))
+        [0.0, 0.0, 1.0, 0.5, 0.0]
+        """
+        if not isinstance(displacement, ProductSpaceVector):
+            raise TypeError('`displacement must be a `ProductSpaceVector` '
+                            'instance, got {!r}'.format(displacement))
+        if not displacement.space.is_power_space:
+            raise TypeError('`displacement.space` must be a power space, '
+                            'got {!r}'.format(displacement.space))
+        if not isinstance(displacement[0].space, DiscreteLp):
```
adler-j: displacement.space[0]

adler-j:
This is starting to look very good, I only found some minor style issues. We should be able to merge soon.

examples/deform/linearized_fixed_displacement.py
chongchenmath: Agree. Using `We` seems not so good.

odl/deform/linearized.py
chongchenmath: ?

adler-j: The comma is redundant and makes the flow of the sentence worse

examples/deform/linearized_fixed_displacement.py
kohr-h: Of course in `I`, there's no other dependency.

examples/deform/linearized_fixed_displacement.py
kohr-h: ok

examples/deform/linearized_fixed_displacement.py
kohr-h: I'll change it.

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+    """Linearized deformation of a template with a displacement field.
+
+    The function maps a given template ``I`` and a given displacement
+    field ``v`` to the new function ``x --> I(x + v(x))``.
+
+    Parameters
+    ----------
+    template : `DiscreteLpVector`
+        Template to be deformed by a displacement field.
+    displacement : `ProductSpace` of `DiscreteLp` element
```
kohr-h: Should be same as below

odl/deform/linearized.py
kohr-h: I don't agree. To me the sentence is more well-structured and readable with comma. It marks the end of a subordinate clause. Otherwise, readers have to fill this separation in their heads -- something I feel takes up energy and attention to other, more important things.

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+        """
+        if not isinstance(displacement, ProductSpaceVector):
+            raise TypeError('`displacement must be a `ProductSpaceVector` '
+                            'instance, got {!r}'.format(displacement))
+        if not displacement.space.is_power_space:
+            raise TypeError('`displacement.space` must be a power space, '
+                            'got {!r}'.format(displacement.space))
+        if not isinstance(displacement[0].space, DiscreteLp):
+            raise TypeError('`displacement[0].space` must be a `DiscreteLp` '
+                            'instance, got {!r}'.format(displacement[0]))
```
kohr-h: wrong argument to `format`

odl/deform/linearized.py
```diff
@@ -0,0 +1,367 @@
...
+        are written. It must have the same shape as ``template`` and
+        a data type compatible with ``template.dtype``.
+
+    Returns
+    -------
+    deformed_template : `numpy.ndarray`
+        Function values of the deformed template. If ``out`` was given,
+        the returned object is a reference to it.
+
+    Examples
```
kohr-h: IMO, since we use the exact same example in both operators, we don't need an example here. It's a private function anyway.

examples/deform/linearized_fixed_displacement.py
adler-j: You are indeed correct.

odl/deform/linearized.py
adler-j: This partly works as a doctest, hence I'd keep it.

It is mostly intended for later maintainers, not for users.

examples/deform/linearized_fixed_displacement.py
kohr-h: Well, I added it anyway for the sake of clarity :-)

odl/deform/linearized.py
kohr-h: ok

kohr-h:
I added an (approximate) inverse to the fixed-displacement operator since we're using it in the adjoint anyway. Fixed the comments.

kohr-h:
Cleaned up the commit history a bit, will merge after tests have passed.

PR #492, opened 2016-07-14
Too much imports

adler-j:
Cleanup of imports in the package

### DIFF
--- a/odl/operator/pspace_ops.py
+++ b/odl/operator/pspace_ops.py
@@ -26,7 +26,7 @@ from builtins import super
 import numpy as np
 import scipy as sp
 
-from odl.operator import Operator
+from odl.operator.operator import Operator
 from odl.space import ProductSpace
 
 
--- a/odl/operator/pspace_ops.py
+++ b/odl/operator/pspace_ops.py
@@ -26,8 +26,8 @@ from builtins import super
 import numpy as np
 import scipy as sp
 
-from odl.operator.operator import Operator
-from odl.space.pspace import ProductSpace
+from odl.operator import Operator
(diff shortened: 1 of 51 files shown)


odl/operator/pspace_ops.py
```diff
@@ -26,8 +26,8 @@
 import numpy as np
 import scipy as sp
 
-from odl.operator.operator import Operator
-from odl.space.pspace import ProductSpace
+from odl.operator import Operator
```
kohr-h: This seems a bit fragile to me. Doesn't it suffice that the imports in `operator/__init__.py` get out of order to get a circular dependency here? I haven't really experimented with this, but it seems a bit more robust to go all the way down *in the same subpackage*. In `default_ops.py` for example, you did not change the import path of `Operator`.

I'm not thinking so much of the core packages here, but rather about somebody who wants to add a module or subpackage. It is perhaps easier to explain that imports from the same subpackage should be with full path, while imports from a different subpackage can rely on this "import to top" feature.

Speaking of users: I think you get this spurious `could not import diagnostics` import error when there's a circular dependency, right? In that case, a FAQ entry on that topic would be great. And a short section in the developer docs about "How to import" would probably also be helpful.

adler-j: This one was a misstake, i tried only changing them in other libraries where there is no such risk, as you mention.

Regarding "could not import diagnostics", yes that's it.

kohr-h: Okay, cool.

kohr-h:
Apart from that one thing (and the "wish list" regarding FAQ :-) ) it's looking good.

PR #495, opened 2016-07-14
BUG: fix corner case bugs in power_method_opnorm

adler-j:
Fixes some inconsistencies as reported by @niinimaki.

### DIFF
--- a/odl/operator/oputils.py
+++ b/odl/operator/oputils.py
@@ -23,7 +23,6 @@ from future import standard_library
 from future.utils import raise_from
 standard_library.install_aliases()
 
-from math import ceil, sqrt
 import numpy as np
 
 from odl.space.base_ntuples import FnBase
@@ -121,6 +120,13 @@ def matrix_representation(op):
 def power_method_opnorm(op, niter, xstart=None):
     """Estimate the operator norm with the power method.
 
+    The operator norm ``||op(x)||`` is defined by as the smallest number such
+    that::
+
+        ||op(x)|| <= ||op|| ||x||
+
+    for all ``x``.
+
     Parameters
     ----------
     op : `Operator`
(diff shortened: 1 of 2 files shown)


odl/operator/oputils.py
```diff
@@ -152,33 +159,49 @@ def power_method_opnorm(op, niter, xstart=None):
...
             op(x, out=tmp)
             op.adjoint(tmp, out=x)
-            x /= x.norm()
 
-        op(x, out=tmp)
-        op.adjoint(tmp, out=x)
-        return sqrt(x.norm())
+            x_norm = x.norm()
+            if x_norm == 0:
+                raise ValueError('reached ``x=0``')
```
kohr-h: Out of curiosity: when does this occur?

adler-j: There is an test for exactly this further down. Basically when you have an operator like

```
[[0, 1],
 [0, 0]]
```

and if you start with `[1, 1]`, you then get `[1, 0]`, and then `[0, 0]`.

odl/operator/oputils.py
```diff
@@ -152,33 +159,49 @@ def power_method_opnorm(op, niter, xstart=None):
...
 
-    tmp = op.range.element()
+    x_norm = x.norm()
+    if x_norm == 0:
+        raise ValueError('input ``x=0``')
+    x /= x_norm
+
+    niter = int(niter)
+    if niter <= 0:
+        raise ValueError('niter not non-negative')
```
kohr-h: Better style: `'`foo` must be bar, got {}'.format(baz)`. More informative, and nice to have when the input is the result of another computation.

odl/operator/oputils.py
adler-j: Some reading material [here](https://en.wikipedia.org/wiki/Nilpotent_operator)

PR #499, opened 2016-07-19
Issue 411  padding

kohr-h:
Implements the resizing of a multi-dimensional array as an operator wrapping a helper function `resize_array` which will be very helpful in itself.

Basically, resizing means keeping the cell sizes and changing the number of cells per axis, plus a shift by a multiple of the cell size.

The extension (padding) is supported in 5 different versions: constant (i.e. fill with a constant), periodic, symmetric, order 0 (continue constantly) and order 1 (continue with constant slope). Not sure we need all of those, but was not too hard to add once the infrastructure was there.

~~What's missing currently is the adjoint. It's not super simple and needs some thought to get it right. I'd suggest we deal with it in a separate issue.~~

TODO:
- [x] Refactor code so forward and adjoint use as much common code as possible (basically the slice calculations)
- [x] More tests, especially corner cases and adjoint
- [x] Doc improvements
- [x] Decision on how to specify start position / distribution of cropped stuff between the left and right sides of the array. I went for `offset` ~~Currently implemented:~~
  * ~~`num_left` -- how many cells added/removed on the left side~~
  * ~~`frac_left` -- same as above, but as fraction. Less reliable due to rounding. Will be removed.~~
  
   ~~Questions: Do we need a `num_right` counterpart? Do we need another option to specify the amount?~~

Edit: last point updated.

### DIFF
--- a/doc/source/math/resizing_ops.rst
+++ b/doc/source/math/resizing_ops.rst
@@ -8,14 +8,13 @@ Resizing Operators
 Introduction
 ============
 In ODL, resizing of a discretized function is understood as the operation of shrinking or enlarging its domain in such a way that the size of the partition cells do not change.
-This "constant cell size" restriction is intentional since it ensures that the underlying operation can be implemented as array resizing without resampling, thus keeping those two functionalities separate.
+This "constant cell size" restriction is intentional since it ensures that the underlying operation can be implemented as array resizing without resampling, thus keeping those two functionalities separate (see `Resampling`).
 
 
 Basic setting
 =============
-Let now :math:`\mathbb{R}^n` with :math:`n \in \mathbb{N}` be the space of one-dimensional real vectors encoding values of a function defined on an interval :math:`[a, b] \subset \mathbb{R}`.
+Let now :math:`\mathbb{R}^n` with :math:`n \in \mathbb{N}` be the space of one-dimensional real vectors encoding values of a function defined on an interval :math:`[a, b] \subset \mathbb{R}` (see :ref:`discretizations` for details).
 Since values are not manipulated, the generalization to complex-valued functions is straightforward.
-Further, since all operations completely separate with respect to different dimensions, the results can easily applied to the multi-dimensional case.
 
 
 Restriction operator
@@ -317,3 +316,26 @@ Hence, the order-1 specific operator has the adjoint
 
 with the convention of summing values for overlapping cases, i.e. if :math:`i \in \{1, 2\}`.
 In practice, the adjoint for the order 1 padding case is applied by computing the zero'th and first order moments of :math:`y` and adding them to the two outmost entries of :math:`x` according to the above rule.
+
+
+Generalization to arbitrary dimension
+=====================================
+Fortunately, all operations are completely separable with respect to (coordinate) axes, i.e. resizing in higher-dimensional spaces can be written as a series of one-dimensional resizing operations.
+One particular issue should be mentioned with the extension operators and their adjoints, though.
+When extending a small, e.g., two-dimensional array to a larger size, there is an ambiguity in how the corner blocks should be handled.
+One possibility would be use the small array size for the extension in both axes, which would leave the corner blocks untouched (initialized to 0 usually):
+
+.. image:: images/resize_small.svg
+   :width: 100%
+
+However, this is not the behavior one would often want in practice.
+Instead, it is much more reasonable to also fill the corners in the same way the "inner" parts have been extended:
+
+.. image:: images/resize_large.svg
+   :width: 100%
+
+This latter behavior is implemented in the resizing operators in ODL.
+
+The adjoint operators of these "corner-filling" resizing operator are given by reversing the unfolding pattern, i.e. by "folding in" the large array axis by axis according to the adjoint formula for the given padding mode.
+This way, the corners also contribute to the final result, which leads to the correct adjoint of the 2D resizing operator.
+Of course, the same principle can easily be generalized to arbitrary dimension.
--- a/doc/source/math/resizing_ops.rst
+++ b/doc/source/math/resizing_ops.rst
@@ -7,34 +7,27 @@ Resizing Operators
 
 Introduction
 ============
-Resizing can be described in a continuous as well as in a fully discrete setting. In the
-continuous case, resizing corresponds to changing the domain of a function space, and the restriction
-and extension operations are well-defined.
+In ODL, resizing of a discretized function is understood as the operation of shrinking or enlarging its domain in such a way that the size of the partition cells do not change.
+This "constant cell size" restriction is intentional since it ensures that the underlying operation can be implemented as array resizing without resampling, thus keeping those two functionalities separate.
 
-However, due to practical reasons, it makes sense to investigate resizing in a discretized setting,
-mainly to make it reflect the resizing of an array directly. This means, for example, that in
-uniformly discretized spaces, the side lengths of the unit cell are not changed, which implies that
-no resampling is performed. In this way, resampling and resizing can be kept separate, with the
-possibility to chain them whenever necessary.
 
-Let now :math:`\mathbb{R}^n` with :math:`n \in \mathbb{N}` be the space of one-dimensional real
-vectors encoding values of a function defined on an interval :math:`[a, b] \subset \mathbb{R}`.
+Basic setting
+=============
+Let now :math:`\mathbb{R}^n` with :math:`n \in \mathbb{N}` be the space of one-dimensional real vectors encoding values of a function defined on an interval :math:`[a, b] \subset \mathbb{R}`.
 Since values are not manipulated, the generalization to complex-valued functions is straightforward.
-Further, since all operations completely separate with respect to different dimensions, the results
-can easily applied to the multi-dimensional case.
+Further, since all operations completely separate with respect to different dimensions, the results can easily applied to the multi-dimensional case.
 
 
 Restriction operator
 ====================
-We consider the space :math:`\mathbb{R}^m` for an :math:`m < n \in \mathbb{N}` and define the
-restriction operator
+We consider the space :math:`\mathbb{R}^m` for an :math:`m < n \in \mathbb{N}` and define the restriction operator
 
 .. math::
     R : \mathbb{R}^n \to \mathbb{R}^m, \quad R(x) := (x_p, \dots, x_{p+m-1})
     :label: def_restrict_op
 
-with a given index :math:`0 \leq p \leq n - m - 1`. Its adjoint with respect to the standard inner
-product is easy to determine:
+with a given index :math:`0 \leq p \leq n - m - 1`.
+Its adjoint with respect to the standard inner product is easy to determine:
 
 .. math::
     \langle R(x), y \rangle_{\mathbb{R}^m}
@@ -53,37 +46,29 @@ with the zero-padding operator
     \end{cases}
     :label: zero_pad_as_restr_adj
 
-In practice, this means that a new zero vector of size :math:`n` is created, and the values
-:math:`y` are filled in from index :math:`p` onwards. It is also clear that the operator :math:`R`
-is right-invertible by :math:`R^*`, i.e. :math:`R R^* = \mathrm{Id}_{\mathbb{R}^m}`. In fact, any
-operator of the form :math:`R^* + P`, where :math:`P` is linear and :math:`P(x)_i = 0` for
-:math:`i \not \in \{p, \dots, p+m-1\}` acts as a right inverse for :math:`R`. On the other hand,
-:math:`R` has no left inverse since it has a non-trivial kernel
-:math:`\mathrm{ker} R = \{x \in \mathbb{R}^n\,|\,x_i = 0 \text{ for } i = p, \dots, p+m-1\}`.
+In practice, this means that a new zero vector of size :math:`n` is created, and the values :math:`y` are filled in from index :math:`p` onwards.
+It is also clear that the operator :math:`R` is right-invertible by :math:`R^*`, i.e. :math:`R R^* = \mathrm{Id}_{\mathbb{R}^m}`.
+In fact, any operator of the form :math:`R^* + P`, where :math:`P` is linear and :math:`P(x)_i = 0` for :math:`i \not \in \{p, \dots, p+m-1\}` acts as a right inverse for :math:`R`.
+On the other hand, :math:`R` has no left inverse since it has a non-trivial kernel (null space) :math:`\mathrm{ker} R = \{x \in \mathbb{R}^n\,|\,x_i = 0 \text{ for } i = p, \dots, p+m-1\}`.
 
 
 Extension operators
 ===================
-Now we study the opposite case of resizing, namely extending a vector. We thus choose :math:`m > n`
-and consider different cases of enlarging a given vector :math:`x \in \mathbb{R}^n` to a vector in
-:math:`\mathbb{R}^m`. The start index is again denoted by :math:`p` and needs to fulfill
-:math:`0 \leq p \leq m - n - 1`, such that a vector of length :math:`n` "fits into" a vector of
-length :math:`m` when starting at index :math:`p`.
+Now we study the opposite case of resizing, namely extending a vector.
+We thus choose :math:`m > n` and consider different cases of enlarging a given vector :math:`x \in \mathbb{R}^n` to a vector in :math:`\mathbb{R}^m`.
+The start index is again denoted by :math:`p` and needs to fulfill :math:`0 \leq p \leq m - n - 1`, such that a vector of length :math:`n` "fits into" a vector of length :math:`m` when starting at index :math:`p`.
 
-It should be noted that all extension operators mentioned here are of the above form
-:math:`R^* + P` with :math:`P` acting on the "outer" indices only. Hence they all act as a right
-inverses for the restriction operator. This property can also be read as the fact that all extension
-operators are left-inverted by the restriction operator :math:`R`.
+It should be noted that all extension operators mentioned here are of the above form :math:`R^* + P` with :math:`P` acting on the "outer" indices only.
+Hence they all act as a right inverses for the restriction operator.
+This property can also be read as the fact that all extension operators are left-inverted by the restriction operator :math:`R`.
 
-Moreover, the "mixed" case, i.e. the combination of restriction and extension which would occur e.g.
-for a constant index shift :math:`x \mapsto (0, \dots, 0, x_0, \dots, x_{n-p-1})`, is not considered
-here. It can be represented by a combination of the two "pure" operations.
+Moreover, the "mixed" case, i.e. the combination of restriction and extension which would occur e.g. for a constant index shift :math:`x \mapsto (0, \dots, 0, x_0, \dots, x_{n-p-1})`, is not considered here.
+It can be represented by a combination of the two "pure" operations.
 
 
 Zero padding
 ------------
-In this most basic padding variant, one fills the missing values in the target vector with zeros,
-yielding the operator
+In this most basic padding variant, one fills the missing values in the target vector with zeros, yielding the operator
 
 .. math::
     E_{\mathrm{z}} : \mathbb{R}^n \to \mathbb{R}^m, \quad E_{\mathrm{z}}(x)_j :=
@@ -99,9 +84,8 @@ Hence, its adjoint is given by the restriction, :math:`E_{\mathrm{z}}^* = R`.
 
 Constant padding
 ----------------
-In constant padding with constant :math:`c`, the extra zeros in :eq:`def_zero_pad_op` are replaced
-with :math:`c`. Hence, the operator performing constant padding can be written as
-:math:`E_{\mathrm{c}} = E_{\mathrm{z}} + P_{\mathrm{c}}`, where the second summand is given by
+In constant padding with constant :math:`c`, the extra zeros in :eq:`def_zero_pad_op` are replaced with :math:`c`.
+Hence, the operator performing constant padding can be written as :math:`E_{\mathrm{c}} = E_{\mathrm{z}} + P_{\mathrm{c}}`, where the second summand is given by
 
 .. math::
     P_{\mathrm{c}}(x) =
@@ -110,17 +94,14 @@ with :math:`c`. Hence, the operator performing constant padding can be written a
     c      , & \text{else}.
     \end{cases}
 
-Note that this operator is not linear, and its derivative is the zero operator, hence the derivative
-of the constant padding operator is :math:`E_{\mathrm{c}}' = E_{\mathrm{z}}`.
+Note that this operator is not linear, and its derivative is the zero operator, hence the derivative of the constant padding operator is :math:`E_{\mathrm{c}}' = E_{\mathrm{z}}`.
 
 
 Periodic padding
 ----------------
-This padding mode continues the original vector :math:`x` periodically in both directions. For
-reasons of practicability, at most one whole copy is allowed on both sides, which means that the
-numbers :math:`n`, :math:`m` and :math:`p` need to fulfill :math:`p \leq n` ("left" padding amount)
-and :math:`m - (p + n) \leq n` ("right" padding amount). The periodic padding operator is then
-defined as
+This padding mode continues the original vector :math:`x` periodically in both directions.
+For reasons of practicability, at most one whole copy is allowed on both sides, which means that the numbers :math:`n`, :math:`m` and :math:`p` need to fulfill :math:`p \leq n` ("left" padding amount) and :math:`m - (p + n) \leq n` ("right" padding amount).
+The periodic padding operator is then defined as
 
 .. math::
     E_{\mathrm{p}}(x)_j :=
@@ -131,8 +112,8 @@ defined as
     \end{cases}
     :label: def_per_pad_op
 
-Hence, one can at most get 3 full periods with :math:`m = 3n` and :math:`p = n`. Again, this operator
-can be written as :math:`E_{\mathrm{p}} = E_{\mathrm{z}} + P_{\mathrm{p}}` with an operator
+Hence, one can at most get 3 full periods with :math:`m = 3n` and :math:`p = n`.
+Again, this operator can be written as :math:`E_{\mathrm{p}} = E_{\mathrm{z}} + P_{\mathrm{p}}` with an operator
 
 .. math::
     P_{\mathrm{p}}(x)_j :=
@@ -168,24 +149,18 @@ and
     0,         & \text{else}.
     \end{cases}
 
-In practice, this means that that besides copying the values from the indices :math:`p, \dots, p+n-1`
-of a vector :math:`y \in \mathbb{R}^m` to a new vector :math:`x \in \mathbb{R}^n`, the values
-corresponding to the other indices are added to the vector :math:`x` as follows. The *first*
-:math:`m - n - p - 1` entries of :math:`y` (negative means 0) are added to the *last*
-:math:`m - n - p - 1` entries of :math:`x`, in the same ascending order. The *last* :math:`p` entries
-of :math:`y` are added to the *first* :math:`p` entries of :math:`x`, again keeping the order. This
-procedure can be interpreted as "folding back" the periodized structure of :math:`y` into a single
-period :math:`x` by adding the values from the two side periods.
+In practice, this means that that besides copying the values from the indices :math:`p, \dots, p+n-1` of a vector :math:`y \in \mathbb{R}^m` to a new vector :math:`x \in \mathbb{R}^n`, the values corresponding to the other indices are added to the vector :math:`x` as follows.
+The *first* :math:`m - n - p - 1` entries of :math:`y` (negative means 0) are added to the *last* :math:`m - n - p - 1` entries of :math:`x`, in the same ascending order.
+The *last* :math:`p` entries of :math:`y` are added to the *first* :math:`p` entries of :math:`x`, again keeping the order.
+This procedure can be interpreted as "folding back" the periodized structure of :math:`y` into a single period :math:`x` by adding the values from the two side periods.
 
 
 Symmetric padding
 -----------------
-In symmetric padding mode, a given vector is extended by mirroring at the outmost nodes to the
-desired extent. By convention, the outmost values are not repeated, and as in periodic mode, the
-input vector is re-used at most once on both sides. Since the outmost values are not doubled, the
-numbers :math:`n`, :math:`m` and :math:`p` need to fulfill the relations
-:math:`p \leq n - 1` ("left" padding amount) and :math:`m - (p + n) \leq n - 1` ("right" padding
-amount). Now the symmetric padding operator is defined as
+In symmetric padding mode, a given vector is extended by mirroring at the outmost nodes to the desired extent.
+By convention, the outmost values are not repeated, and as in periodic mode, the input vector is re-used at most once on both sides.
+Since the outmost values are not doubled, the numbers :math:`n`, :math:`m` and :math:`p` need to fulfill the relations :math:`p \leq n - 1` ("left" padding amount) and :math:`m - (p + n) \leq n - 1` ("right" padding amount).
+Now the symmetric padding operator is defined as
 
 .. math::
     E_{\mathrm{s}}(x)_j :=
@@ -232,18 +207,12 @@ and
     0,            & \text{else}.
     \end{cases}
 
-Note that the index condition :math:`m - (p + n) \leq n - 1` is equivalent to
-:math:`2n - 1 + p - m \geq 0`, hence the index range in the definition of
-:math:`P_{\mathrm{s},2}^*` is well-defined.
+Note that the index condition :math:`m - (p + n) \leq n - 1` is equivalent to :math:`2n - 1 + p - m \geq 0`, hence the index range in the definition of :math:`P_{\mathrm{s},2}^*` is well-defined.
 
-Practically, the evaluation of :math:`E_{\mathrm{s}}^*` consists in copying the "main" part of
-:math:`y \in \mathbb{R}^m` corresponding to the indices :math:`p, \dots, p + n - 1` to
-:math:`x \in \mathbb{R}^n` and updating the vector additively as follows. The values at indices 1 to
-:math:`p` are updated with the values of :math:`y` mirrored at the index position :math:`p`, i.e. in
-reversed order. The values at the indices :math:`2n - 1 + p - m` to :math:`n - 2` are updated with
-the values of :math:`y` mirrored at the position :math:`2n + 2 - p`, again in reversed order. This
-procedure can be interpreted as "mirroring back" the outer two parts of the vector :math:`y` at the
-indices :math:`p` and :math:`2n + 2 - p`, adding those parts to the "main" vector.
+Practically, the evaluation of :math:`E_{\mathrm{s}}^*` consists in copying the "main" part of :math:`y \in \mathbb{R}^m` corresponding to the indices :math:`p, \dots, p + n - 1` to :math:`x \in \mathbb{R}^n` and updating the vector additively as follows.
+The values at indices 1 to :math:`p` are updated with the values of :math:`y` mirrored at the index position :math:`p`, i.e. in reversed order.
+The values at the indices :math:`2n - 1 + p - m` to :math:`n - 2` are updated with the values of :math:`y` mirrored at the position :math:`2n + 2 - p`, again in reversed order.
+This procedure can be interpreted as "mirroring back" the outer two parts of the vector :math:`y` at the indices :math:`p` and :math:`2n + 2 - p`, adding those parts to the "main" vector.
 
 
 Order 0 padding
@@ -272,15 +241,15 @@ This operator is the sum of the zero-padding operator and
 We calculate the adjoint of :math:`P_{\mathrm{o0}}`:
 
 .. math::
- \langle P_{\mathrm{o0}}(x), y \rangle_{\mathbb{R}^m}
- &= \sum_{j=0}^{p-1} x_0\, y_j + \sum_{j=p+n}^{m-1} x_{n-1}\, y_j \\
- &= x_0 \sum_{j=0}^{p-1} y_j + x_{n-1} \sum_{j=p+n}^{m-1} y_j \\
- &= x_0 M_{\mathrm{l},0}(y) + x_{n-1} M_{\mathrm{r},0}(y)
+    \langle P_{\mathrm{o0}}(x), y \rangle_{\mathbb{R}^m}
+    &= \sum_{j=0}^{p-1} x_0\, y_j + \sum_{j=p+n}^{m-1} x_{n-1}\, y_j \\
+    &= x_0 \sum_{j=0}^{p-1} y_j + x_{n-1} \sum_{j=p+n}^{m-1} y_j \\
+    &= x_0 M_{\mathrm{l},0}(y) + x_{n-1} M_{\mathrm{r},0}(y)
 
 with the zero'th order moments
 
 .. math::
- M_{\mathrm{l},0}(y) := \sum_{j=0}^{p-1} y_j, \quad M_{\mathrm{r},0}(y) := \sum_{j=p+n}^{m-1} y_j.
+    M_{\mathrm{l},0}(y) := \sum_{j=0}^{p-1} y_j, \quad M_{\mathrm{r},0}(y) := \sum_{j=p+n}^{m-1} y_j.
 
 Hence, we get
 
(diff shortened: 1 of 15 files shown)


odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
...
+
+        For the particular case of constant padding with non-zero
+        constant, the derivative is the corresponding zero-padding
+        variant. In all other cases, this operator is linear, i.e.
+        the derivative is equal to ``self``.
+        """
+        if self.pad_mode == 'constant' and self.pad_const != 0:
+            return ResizingOperator(
+                domain=self.domain, range=self.range, pad_mode='constant',
+                pad_const=0.0)
```
kohr-h: `else` missing

odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
...
+        In axes where ``self`` extends, the adjoint is given by the
+        corresponding restriction. In restriction axes, the adjoint
+        performs zero-padding.
+        """
+        if not self.is_linear:
+            raise NotImplementedError('this operator is not linear and '
+                                      'thus has no adjoint')
+        # TODO: this is not correct, fix it
+        return ResizingOperator(domain=self.range, range=self.domain,
+                                pad_mode='constant', pad_const=0.0)
```
kohr-h: Change to `NotImplementedError` (text also wrong)

odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
...
+
+    @staticmethod
+    def _num_left_from_spaces(dom, ran):
+        """Return num_left corresponding to given spaces."""
+        diff_l = np.abs(ran.grid.min() - dom.grid.min())
+        num_left_float = diff_l / dom.cell_sides
+        num_left = np.around(num_left_float).astype(int)
+        if not np.allclose(num_left, num_left_float):
+            raise ValueError('range is shifted relative to domain by a '
+                             'non-multiple of cell_sides')
```
kohr-h: perhaps print shift

odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
...
+    the left (per axis). In axes where ``num_left`` is ``None``, the
+    points are distributed evenly.
+    """
+    new_begin, new_end = [], []
+    for b_orig, e_orig, n_orig, cs, n_new, num_l in zip(
+            discr.min_corner, discr.max_corner, discr.shape, discr.cell_sides,
+            newshp, num_left):
+
+        n_diff = n_new - n_orig
+#        print(n_diff)
```
kohr-h: Remove commented-out print statements

odl/util/numerics.py
```diff
@@ -297,6 +303,521 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+    newshp : sequence of int
+        Desired shape of the output array.
+    frac_left : float or sequence of float, optional
+        Number between 0 and 1 (inclusive) that determines which fraction
+        of the addition/removal is performed on the "left" side. If
+        ``frac_left`` times the shape difference is not integer, the
+        number of entries added to the left is gained by rounding to the
+        nearest integer. The default ``None`` is equivalent to 0.5.
+        This option can be specified per axis with a sequence.
+        Cannot be combined with ``num_left``.
```
kohr-h: Actually, if I think about it, this option should be removed. It's not super reliable due to rounding, and the same effect can easily be reached by a simple calculation by the user. Opinions?

odl/util/numerics.py
```diff
@@ -297,6 +303,521 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+        ``'order0'``: Extend constantly with the outmost values
+        (ensures continuity).
+
+        ``'order1'``: Extend with constant slope (ensures continuity of
+        the first derivative). This requires at least 2 values along
+        each axis where padding is applied.
+
+        Note that for ``'symmetric'`` and ``'periodic'`` padding, the
+        number of added values on each side of the array cannot exceed
+        the original size.
```
kohr-h: Remove this paragraph

odl/util/numerics.py
```diff
@@ -297,6 +303,521 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+    Returns
+    -------
+    resized : `numpy.ndarray`
+        Resized array created according to the above rules. If ``out``
+        was given, the returned object is a reference to it.
+
+    Examples
+    --------
+    The input can be shrunk by simply providing a smaller size.
+    The left/right distribution of removed entries is ananogous to
```
kohr-h: analogous

odl/util/numerics.py
```diff
@@ -297,6 +303,521 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+        arange_indcs = slice_tuple(ndim, axis, slice(None), None)
+
+        arr[arr_indcs_l] = (arr[bdry_indcs_l] +
+                            slope_arr_l * arange_l[arange_indcs])
+        arr[arr_indcs_r] = (arr[bdry_indcs_r] +
+                            slope_arr_r * arange_r[arange_indcs])
+
+
+def _resize_array_adj(arr, out, num_left, pad_mode='constant', pad_const=0):
+    """Adjoint operation to `resize_array`.
```
kohr-h: Shouldn't be part of this PR, remove the function.

odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
...
+        """Number of cells added to/removed from the left."""
+        return self.__num_left
+
+    @property
+    def pad_mode(self):
+        """Padding mode used by this operator."""
+        return self.__pad_mode
+
+    @property
+    def pad_const(self):
```
adler-j: I dont think this is the default name. Looking at [`scipy`](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.filters.gaussian_filter.html) they seem to go for `cval`. We may or may not want to go fro this.

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,356 @@
+.. _resizing_ops:
+
+##################
+Resizing Operators
+##################
+
+
+Introduction
+============
+Resizing can be described in a continuous as well as in a fully discrete setting. In the
```
adler-j: This paragraph does not (at least to me) fully describe what you mean by resizing. "Changing the domain" is abit vague. Change in what way? It would be good to be abit more clear here since this is the most important paragraph.

odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
         """
         return self.inverse
 
+
+class ResizingOperator(Operator):
+
+    """Operator mapping between discretized spaces of different shapes.
```
adler-j: Too vague

odl/discr/discr_ops.py
```diff
@@ -151,6 +159,226 @@ def adjoint(self):
...
+
+    This operator is intended as a mapping between uniformly
```
adler-j: "is intended as" is abit unnecessary, unless this is an abstract operator intended for inheritance.

kohr-h:
Some of the texts are not ideal yet, I'll get to that.

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,356 @@
+.. _resizing_ops:
+
+##################
+Resizing Operators
```
adler-j: Much of the text here is about boundary behaviors. We should probably write a separate doc on those (since this should match) and refer it from here. That can be done later.

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
     return out
 
 
+def _compute_slices(arr1, arr2, num_left, pad_mode, direction):
```
adler-j: could use a docstring

odl/util/numerics.py
```diff
@@ -24,8 +24,14 @@
...
 import numpy as np
 
+from odl.util.normalize import normalized_scalar_param_list, safe_int_conv
 
-__all__ = ('apply_on_boundary', 'fast_1d_tensor_mult')
+
+__all__ = ('apply_on_boundary', 'fast_1d_tensor_mult', 'resize_array')
+
+
+_SUPPORTED_PAD_MODES = ('constant', 'symmetric', 'periodic',
```
adler-j: Not future proof. Change to something like `_SUPPORTED_PAD_MODES_RESIZE`

odl/util/numerics.py
adler-j: Well the sheer length of the doc that is needed to explain it is a strong argument against keeping it.

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+        Desired shape of the output array.
+    frac_left : float or sequence of float, optional
+        Number between 0 and 1 (inclusive) that determines which fraction
+        of the addition/removal is performed on the "left" side. If
+        ``frac_left`` times the shape difference is not integer, the
+        number of entries added to the left is gained by rounding to the
+        nearest integer. The default ``None`` is equivalent to 0.5.
+        This option can be specified per axis with a sequence.
+        Cannot be combined with ``num_left``.
+    num_left : sequence of int, optional
```
adler-j: What does left even mean? Also no matching "right"?

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+        'forward' : in axes where ``out`` is larger than ``arr``,
+        apply padding. Otherwise, restrict to the smaller size.
+
+        'adjoint' : in axes where ``out`` is larger than ``arr``,
+        apply zero-padding. Otherwise, restrict to the smaller size
+        and add the outside contributions according to ``pad_mode``.
+
+    out : `numpy.ndarray`, optional
+        Array to write the result to. Must have shape ``newshp`` and
+        be able to hold the data type of the input array.
```
adler-j: Data order restrictions?

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+    for i, (slc_l, slc_r) in enumerate(zip(pad_l_slc, pad_r_slc)):
+        if direction == 'forward':
+            _apply_padding(out, slc_l, slc_r, i, pad_mode, pad_const)
+        else:  # direction == 'adjoint'
+            _apply_adjoint_padding(arr, out, slc_l, slc_r, i, pad_mode)
+
+    return out
+
+
+def _slice_dict(axis, ax_len, ndim, slc_l, slc_r):
```
adler-j: Very weird way of making a function. Why not simply make it take the padding variant as an argument and return the actual slice?

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+            raise ValueError('`out` must have shape {}, got {}'
+                             ''.format(newshp, out.shape))
+
+        arr = np.asarray(arr, dtype=out.dtype)
+        if arr.ndim != out.ndim:
+            raise ValueError('number of axes of `arr` and `out` do not match '
+                             '({} != {})'.format(arr.ndim, out.ndim))
+    else:
+        arr = np.asarray(arr)
+        out = np.empty(newshp, dtype=arr.dtype)
```
adler-j: how about order?

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+            pad_r_slc.append(slice(0))
+
+        elif n_new > n_orig:
+            # Padding case: calculate start and stop indices for the
+            # bigger new array, together with the padding slices
+            n_add = n_new - n_orig
+            istart = num_l
+            istop = n_new - (n_add - istart)
+
+            n_pad_l = len(range(istart))
```
adler-j: Wut? Why not simply `n_pad_l = istart`? Or if pad can be negative, `n_pad_l = max(0, istart)`

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+            n_pad_r = len(range(istop, n_new))
+
+            # Handle some error scenarios with illegal lengths
+            if pad_mode == 'order0' and n_orig == 0:
+                raise ValueError('in axis {}: need at least 1 value for '
+                                 'order 0 padding, got 0'
+                                 ''.format(i))
+
+            if pad_mode == 'order1' and n_orig == 1:
```
adler-j: `and n_orig < 2` is probably more consistent with the bellow, and also handles `n_orig = 0`.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,181 @@
...
+    true_res = np.zeros((8, 2))
+    true_res[:4, :] = 1
+    assert np.array_equal(out, true_res)
+
+    out = res_space.element()
+    res_op(space.one(), out=out)
+    assert np.array_equal(out, true_res)
+
+
+def test_resizing_op_deriv():
```
adler-j: This should have a fixture, and I.m.o should test much more corner cases (axis size 0 and 1, different dtypes, data orders, what happens if space is over integers? etc)

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,181 @@
...
+    res_op = odl.ResizingOperator(space, res_space, pad_const=1)
+    assert res_op.pad_const == 1.0
+    assert not res_op.is_linear
+
+    res_op = odl.ResizingOperator(space, res_space, pad_mode='symmetric')
+    assert res_op.pad_mode == 'symmetric'
+    assert res_op.is_linear
+
+
+def test_resizing_op_call(fn_impl):
```
adler-j: This should have a fixture, and I.m.o should test much more corner cases (axis size 0 and 1, different dtypes, data orders, what happens if space is over integers? etc)

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,181 @@
...
+    odl.ResizingOperator(space, res_space)
+    odl.ResizingOperator(space, ran_shp=(20, 15))
+    odl.ResizingOperator(space, ran_shp=(20, 15), num_left=(0, 5))
+    odl.ResizingOperator(space, ran_shp=(20, 15), pad_mode='symmetric')
+    odl.ResizingOperator(space, ran_shp=(20, 15), pad_const=1.0)
+    odl.ResizingOperator(space, ran_shp=(20, 15), pad_const=1.0,
+                         discr_kwargs={'nodes_on_bdry': True})
+
+
+def test_resizing_op_raise():
```
adler-j: What about different dimensions of domain and range?

odl/discr/discr_ops.py
kohr-h: I chose `pad_const` since it makes the connection to `'constant'` padding obvious.
We use `padding_value` in [other places](https://github.com/odlgroup/odl/blob/281d1d8763cc33e62a0e67268a5668fa033bdd95/odl/discr/diff_ops.py#L55-L56), which I think is inferior. `cval` is only shorter, not better. Not sure if one can speak of a "default" here. Even the names for the padding modes are not really standardized. For example, the mode `'constant'` in PyWavelets is the "replicate boundary values" mode, not "fill constant" as in SciPy. So I'd say we're free to choose whichever name we think is most appropriate.

doc/source/math/resizing_ops.rst
kohr-h: Will improve.

odl/discr/discr_ops.py
kohr-h: Hard to be more precise in one line, but will try..

odl/util/numerics.py
kohr-h: ok

odl/util/numerics.py
kohr-h: `num_right` can be caluclated from the shapes and `num_left`, but sure, I can add it for convenience. `left` is something I chose since it's intuitively clear in 1D, but it's less so in higher dimensions. Alternatives?
- left/right
- before/after 
- top/bottom
- ?

odl/util/numerics.py
kohr-h: None needed? We assign stuff by slicing, that will work irrespective of the data order. Performance will depend on it, sure.

odl/util/numerics.py
kohr-h: Will be refactored.

odl/util/numerics.py
kohr-h: I'll add that here, yes.

odl/util/numerics.py
```diff
@@ -297,6 +303,593 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
+        num_left = normalized_scalar_param_list(
+            num_left, out.ndim, param_conv=safe_int_conv, keep_none=False)
+
+    # Handle padding
+    pad_mode, pad_mode_in = str(pad_mode).lower(), pad_mode
+    if pad_mode not in _SUPPORTED_PAD_MODES:
+        raise ValueError("`pad_mode` '{}' not understood".format(pad_mode_in))
+
+    if (pad_mode == 'constant' and
+        not np.can_cast(pad_const, out.dtype) and
```
kohr-h: Should be `or`, right.

odl/util/numerics.py
kohr-h: How about `offset`? That should be perfectly clear and does not need a "right" counterpart. However, in that case, the default should probably be 0 in each axis. More reliable, too.

chongchenmath:
Hi, I want to use the temporary `ZeroPaddingOperator` to implement the LDDMM solver. I ever used that guy in `Linearized_deformations` branch. But I cannot find it now. So need help. Thanks a lot.

odl/discr/discr_ops.py
kohr-h: added

odl/util/numerics.py
kohr-h: now part of the PR anyway

odl/discr/discr_ops.py
kohr-h: ok?

odl/util/numerics.py
kohr-h: done

adler-j:
`Needs rebase`? :)

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,319 @@
...
+##################
+
+
+Introduction
+============
+In ODL, resizing of a discretized function is understood as the operation of shrinking or enlarging its domain in such a way that the size of the partition cells do not change.
+This "constant cell size" restriction is intentional since it ensures that the underlying operation can be implemented as array resizing without resampling, thus keeping those two functionalities separate.
```
adler-j: Link to resampling operators/doc?

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,319 @@
...
+
+
+Basic setting
+=============
+Let now :math:`\mathbb{R}^n` with :math:`n \in \mathbb{N}` be the space of one-dimensional real vectors encoding values of a function defined on an interval :math:`[a, b] \subset \mathbb{R}`.
```
adler-j: We have a long text on discretizations, how about linking to it as well?

kohr-h:
So now I refactored the code to a great deal so it has maximum code reuse between forward and adjoint of the padding (actually resizing) operation. Tests are parametrized and check more corner cases, adjoint operator is implemented and has a test. So apart from perhaps a missing doctest for the operator, this is ready for review (and rebased :-) ).

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,319 @@
...
+Since values are not manipulated, the generalization to complex-valued functions is straightforward.
+Further, since all operations completely separate with respect to different dimensions, the results can easily applied to the multi-dimensional case.
+
+
+Restriction operator
+====================
+We consider the space :math:`\mathbb{R}^m` for an :math:`m < n \in \mathbb{N}` and define the restriction operator
```
adler-j: This could perhaps be done while allowing `m = n`, this would allow some useful cases such as the trivial restriction.

kohr-h: It's allowed in the implementation, sure. Here, for the purpose of illustration, it would not be so interesting.

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,319 @@
...
+Hence, the operator performing constant padding can be written as :math:`E_{\mathrm{c}} = E_{\mathrm{z}} + P_{\mathrm{c}}`, where the second summand is given by
+
+.. math::
+    P_{\mathrm{c}}(x) =
+    \begin{cases}
+    0      , & \text{if } p \leq j \leq p + n - 1, \\
+    c      , & \text{else}.
+    \end{cases}
+
+Note that this operator is not linear, and its derivative is the zero operator, hence the derivative of the constant padding operator is :math:`E_{\mathrm{c}}' = E_{\mathrm{z}}`.
```
adler-j: the zero *padding* operator

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,319 @@
...
+
+
+Periodic padding
+----------------
+This padding mode continues the original vector :math:`x` periodically in both directions.
+For reasons of practicability, at most one whole copy is allowed on both sides, which means that the numbers :math:`n`, :math:`m` and :math:`p` need to fulfill :math:`p \leq n` ("left" padding amount) and :math:`m - (p + n) \leq n` ("right" padding amount).
```
adler-j: Feels rather arbitrary. Why is this done?

doc/source/math/resizing_ops.rst
```diff
@@ -0,0 +1,319 @@
...
+    \begin{cases}
+    -M_{\mathrm{l},1}(y), & \text{if } i = 0,     \\
+    M_{\mathrm{l},1}(y),  & \text{if } i = 1,     \\
+    -M_{\mathrm{r},1}(y), & \text{if } i = n - 2, \\
+    M_{\mathrm{r},1}(y),  & \text{if } i = n - 1, \\
+    0,                  & \text{else},
+    \end{cases}
+
+with the convention of summing values for overlapping cases, i.e. if :math:`i \in \{1, 2\}`.
+In practice, the adjoint for the order 1 padding case is applied by computing the zero'th and first order moments of :math:`y` and adding them to the two outmost entries of :math:`x` according to the above rule.
```
adler-j: It is not clear for me how n-d data is handled with this interpolation.

odl/discr/discr_ops.py
adler-j: Fine with this, but we should make an issue to standardize these across ODL.

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+    def pad_mode(self):
+        """Padding mode used by this operator."""
+        return self.__pad_mode
+
+    @property
+    def pad_const(self):
+        """Constant used by this operator in case of constant padding."""
+        return self.__pad_const
+
+    @staticmethod
```
adler-j: Why make this static instead of a free method?

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+        offset = np.around(offset_float).astype(int)
+        if not np.allclose(offset, offset_float):
+            raise ValueError('range is shifted relative to domain by a '
+                             'non-multiple of cell_sides')
+        return tuple(offset)
+
+
+class ResizingOperator(ResizingOperatorBase):
+
+    """Operator mapping a discrete function to a new domain.
```
adler-j: discretized perhaps?

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+
+    This operator is a mapping between uniformly discretized spaces with
```
adler-j: Should mention `\`DiscreteLp\`` somewhere here, so users get a link to it.

odl/discr/discr_ops.py
adler-j: Seem to miss several arguments here, like `discr_kwargs` and, `offset` and `ran_shape`.

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+        else:  # operator is linear
+            return self
+
+    @property
+    def adjoint(self):
+        """Adjoint of this operator."""
+        if not self.is_linear:
+            raise NotImplementedError('this operator is not linear and '
+                                      'thus has no adjoint')
+        return ResizingOperatorAdjoint(domain=self.range, range=self.domain,
```
adler-j: Missing arguments like `offset` etc

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+
+    @property
+    def inverse(self):
+        """(Pseudo-)Inverse of this operator.
+
+        Note that in axes where ``self`` extends, the returned operator
+        acts as left inverse, while in restriction axes, it is a
+        right inverse.
+        """
+        return ResizingOperator(domain=self.range, range=self.domain,
```
adler-j: Missing arguments

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+        # TODO: simplify once context manager is available
+        out[:] = resize_array(x.asarray(), self.range.shape,
+                              offset=self.offset, pad_mode=self.pad_mode,
+                              pad_const=0, direction='adjoint',
+                              out=out.asarray())
+
+    @property
+    def adjoint(self):
+        """Adjoint of this operator."""
+        return ResizingOperator(domain=self.range, range=self.domain,
```
adler-j: `if not self.is_linear`

and missing arguments

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+                                pad_mode=self.pad_mode, pad_const=0.0)
+
+
+def _resize_discr(discr, newshp, offset, discr_kwargs):
+    """Return ``discr`` resized to ``newshp``.
```
adler-j: I disagree with this docstring, it barely retents `discr`'s properties at all, basically only taking the underlying domain and size. Things like `dspace`, `dtype` etc seem to be taken from `discr_kwargs`.

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+                    '(difference {})'
+                    ''.format(range.cell_sides - domain.cell_sides))
+
+            self.__offset = self._offset_from_spaces(domain, range)
+
+        else:
+            raise ValueError('cannot combine `range` with `ran_shape`')
+
+        pad_mode = kwargs.pop('pad_mode', 'constant')
+        self.__pad_mode = str(pad_mode).lower()
```
adler-j: do the usual `pad_mode_in` thingie

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+    In axes where the domain is enlarged, the new entries are filled
+    ("padded") according to a provided parameter ``pad_mode``.
+
+    All resizing operator variants are linear, except constant padding
+    with constant != 0.
+
+    See `the online documentation
+    <https://odl.readthedocs.io/math/resizing_ops.html>`_
+    on resizing operators for mathematical details.
+    """
```
adler-j: This class is totally missing any examples, would be great with some simple examples.

doc/source/math/resizing_ops.rst
kohr-h: Relaxing this is certainly possible, but it would be a significant extra complexity. Instead of one slice left and right per axis, one would have to use lists, which I don't really feel like thinking about now. This is already complex enough, and if we really need multiple periods at some point, we can add it. Already now, you can get a poor man's version of that by simply extending in several steps.

doc/source/math/resizing_ops.rst
kohr-h: Yes.

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,353 @@ def _call(self, x, out=None):
...
+        Note that in axes where ``self`` extends, the returned operator
+        acts as left inverse, while in restriction axes, it is a
+        right inverse.
+        """
+        return ResizingOperator(domain=self.range, range=self.domain,
+                                pad_mode=self.pad_mode,
+                                pad_const=self.pad_const)
+
+
+class ResizingOperatorAdjoint(ResizingOperatorBase):
```
adler-j: Do we really need a separate operator for the `Adjoint`? It basically seems to be doing very little. Couldnt all three classes here be merged into a single class, or am I missing something?

doc/source/math/resizing_ops.rst
kohr-h: Well it's not really long and rather incomplete, not sure it helps much. But it can be linked anyway.

doc/source/math/resizing_ops.rst
kohr-h: It's like unfolding (extension, forward) and folding (restriction, adjoint) a piece of paper. The contributions are summed. I don't know how that plays along with interpolation schemes, but they're not really involved here.

test/util/numerics_test.py
```diff
@@ -216,5 +222,299 @@ def test_fast_1d_tensor_mult_error():
...
+    else:
+        other_arr = (np.random.uniform(-10, 10, size=newshp) +
+                     1j * np.random.uniform(-10, 10, size=newshp))
+
+    resized = resize_array(array, newshp, offset, pad_mode, pad_const,
+                           direction='forward')
+    resized_adj = resize_array(other_arr, array.shape, offset, pad_mode,
+                               pad_const, direction='adjoint')
+
+    print(resized)
```
adler-j: leftover?

adler-j:
Quite good looking pull request, had some minor points that could use fixing.

doc/source/math/resizing_ops.rst
adler-j: Could perhaps use some more text on this then, this was not obvious to me.

odl/discr/discr_ops.py
kohr-h: See #511

odl/discr/discr_ops.py
kohr-h: If it's useful on its own right, it can be a free function, sure. Here it is a bit more safe to make assumptions on the parameters.

odl/discr/discr_ops.py
kohr-h: I'm aware that there are no doctests yet.

odl/discr/discr_ops.py
kohr-h: Forgot to update, thx.

odl/discr/discr_ops.py
kohr-h: It doesn't do that much, no, it catches `pad_const != 0` and has a different implementation of `_call` and `adjoint`. And no inverse, which it perhaps also should have.

test/util/numerics_test.py
kohr-h: obviously

odl/discr/discr_ops.py
kohr-h: Actually no, these parameters are all encoded in `resize_op.range` so we don't need them.

kohr-h:
All done except doctests.

kohr-h:
Added a doctest for `ResizingOp` and a use case for padding in the Fourier transform example. Also spotted some errors, fixed now. Should be ready for merge.

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,407 @@ def _call(self, x, out=None):
...
+        domain : uniform `DiscreteLp`
+            Uniformly discretized space, the operator can be applied
+            to its elements.
+        range : uniform `DiscreteLp`, optional
+            Uniformly discretized space in which the result of the
+            application of this operator lies.
+            For the default ``None``, a space with the same attributes
+            as ``domain`` is used, except for its shape, which is set
+            to ``ran_shp``.
+        ran_shp : sequence of int
```
adler-j: , optional

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,407 @@ def _call(self, x, out=None):
...
+        discr_kwargs: dict, optional
+            Keyword arguments passed to the `uniform_discr` constructor.
+
+        Examples
+        --------
+        The simplest way of initializing a resizing operator is by
+        providing ``ran_shp`` and, optionally, parameters for the padding
+        variant to be used. The range is inferred from ``domain`` and
+        the supplied parameters. If no ``offset`` is given, the difference
+        in size is evenly distributed to both sides:
```
adler-j: What happens if the difference is a odd number?

odl/discr/discr_ops.py
```diff
@@ -101,56 +111,407 @@ def _call(self, x, out=None):
...
+    diff_l = np.abs(ran.grid.min() - dom.grid.min())
+    offset_float = diff_l / dom.cell_sides
+    offset = np.around(offset_float).astype(int)
+    if not np.allclose(offset, offset_float):
+        raise ValueError('range is shifted relative to domain by a '
+                         'non-multiple of cell_sides')
+    return tuple(offset)
+
+
+def _resize_discr(discr, newshp, offset, discr_kwargs):
```
adler-j: This does not preserve `impl` and others.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+import odl
+from odl.discr.discr_ops import _SUPPORTED_RESIZE_PAD_MODES
+from odl.util.testutils import almost_equal, example_element
+
+
+# --- ResizingOperator --- #
+
+
+def test_resizing_op_init():
```
adler-j: Add the `fn_impl` fixture and add `impl=fn_impl` to the spaces.

You may also want to test going from one impl to another, it is not clear if this is supported or not.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+    with pytest.raises(TypeError):
+        odl.ResizingOperator(odl.rn(5), ran_shp=(10,))
+
+    grid = odl.TensorGrid([0, 2, 3])
+    part = odl.RectPartition(odl.IntervalProd(0, 3), grid)
+    fspace = odl.FunctionSpace(odl.IntervalProd(0, 3))
+    dspace = odl.rn(3)
+    space = odl.DiscreteLp(fspace, part, dspace)
+    with pytest.raises(ValueError):
+        odl.ResizingOperator(space, ran_shp=(10,))
```
adler-j: why should this raise?

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+dtype_params = ['float64', 'complex64']
+dtype_ids = [' dtype = {} '.format(p) for p in dtype_params]
+
+
+@pytest.fixture(scope="module", ids=dtype_ids, params=dtype_params)
+def dtype(request):
+    return request.param
+
+
+def test_resizing_op_properties(dtype):
```
adler-j: Should also use `fn_impl` to test various fn, and use `odl.FN_IMPLS[fn_impl].available_dtypes()` to find dtypes.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+    res_op = odl.ResizingOperator(space, ran_shp=(20, 15), offset=[0, 5])
+    assert np.allclose(res_op.range.min_corner, res_space.min_corner)
+    assert np.allclose(res_op.range.max_corner, res_space.max_corner)
+    assert np.allclose(res_op.range.cell_sides, res_space.cell_sides)
+    assert res_op.range.dtype == res_space.dtype
+    assert res_op.offset == (0, 5)
+
+    # Different padding mode
+    res_op = odl.ResizingOperator(space, res_space, pad_const=1)
+    assert res_op.pad_const == 1.0
```
adler-j: Also `assert res_op.pad_mode = 'constant'`

test/discr/discr_ops_test.py
adler-j: Why not use the `pad_mode` fixture that you define below with this test and use it for an exhaustive test.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+    assert np.array_equal(out, true_res)
+
+    out = res_space.element()
+    res_op(space.one(), out=out)
+    assert np.array_equal(out, true_res)
+
+
+def test_resizing_op_deriv():
+
+    # Only non-trivial case is constant padding with const != 0
```
adler-j: Could also do a test that `op.derivative(x) is op` for other cases?

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+    space = odl.uniform_discr([0, -1], [1, 1], (4, 5), dtype=dtype,
+                              impl=fn_impl)
+    res_space = odl.uniform_discr([0, -1.4], [1.5, 1.4], (6, 7), dtype=dtype,
+                                  impl=fn_impl)
+    res_op = odl.ResizingOperator(space, res_space, pad_mode=pad_mode,
+                                  pad_const=pad_const)
+
+    # Only left inverse if the operator extentds in all axes
+    x = space.element(np.arange(space.size).reshape(space.shape))
```
adler-j: Should simply use `odl.util.testutils.example_element(space)`, it would be more obvious what you mean.

test/discr/discr_ops_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+@pytest.fixture(scope="module", params=pad_modes, ids=pad_mode_ids)
+def pad_mode(request):
+    return request.param
+
+
+def test_resizing_op_inverse(pad_mode, dtype, fn_impl):
+
+    if isinstance(pad_mode, tuple):
```
adler-j: Instead of repeating this, why not simply add this to the fixture definition?

odl/discr/discr_ops.py
kohr-h: Ah f**k, forgot that. Will add a test for that, too.

odl/discr/discr_ops.py
kohr-h: One more to the left, if that's an okay default.

odl/discr/discr_ops.py
kohr-h: right

test/discr/discr_ops_test.py
kohr-h: Not uniformly discretized.

test/util/numerics_test.py
```diff
@@ -216,5 +222,291 @@ def test_fast_1d_tensor_mult_error():
...
+pad_mode_ids = [" pad_mode = '{}' ".format(p) for p in pad_mode_params]
+
+
+@pytest.fixture(scope="module", ids=pad_mode_ids, params=pad_mode_params)
+def pad_mode(request):
+    return request.param
+
+
+variants = ['extend', 'restrict', 'mixed']
+setup_params = list(product(pad_mode_params, variants))
```
adler-j: It would be more clear what you meant if you instead passed the two fixtures as arguments to `resize_setup`. pytest supports "recursive" fixtures.

test/discr/discr_ops_test.py
adler-j: Also may want to consider storage order.

adler-j:
Some remarks on the tests, otherwise fine!

odl/discr/discr_ops.py
adler-j: Ok by me, but this should be mentioned.

kohr-h: [It is](https://github.com/odlgroup/odl/pull/499/files/b67607feef6f368f8ccac83f2b1eb6335e5137e8#diff-783324f48a9505cd5363ed78952719f9R189)

test/util/numerics_test.py
```diff
@@ -234,65 +234,73 @@ def dtype(request):
...
+
+@pytest.fixture(scope="module")
+def resize_setup(pad_mode, variant):
     array_in = [[1, 2, 3, 4],
                 [5, 6, 7, 8],
                 [9, 10, 11, 12]]
 
-    pad_mode, variant = request.param
-    pad_const = 0
+    mode, const = pad_mode
```
adler-j: probably should rename these to fit the parameters in the function

`pad_mode, pad_const = padding`

or something like that.

kohr-h:
All test comments processed, should be mergeable.

test/util/numerics_test.py
```diff
@@ -21,13 +21,16 @@
 from future import standard_library
 standard_library.install_aliases()
 
-# External
-import pytest
+from itertools import product
```
adler-j: unused

adler-j:
Should be cool, add release notes and go.

Issue #52, opened 2015-11-19
Decide on a (class) structure for solvers


adler-j:
Moved from https://github.com/odlgroup/odl-solvers/issues/3

I initialized a structure, but it is obviously far from optimal. We need to find sensible categories for solvers, e.g.

- methods for finding a zero of a goal function(al), e.g. quasi-Newton schemes
- simple methods for minimizing goal functions
- "more complex" methods involving e.g. Fenchel/Lagrange duals and methods combining primal and dual problems. These probably need a more structured approach.

We need to reflect these categories in classes in a sensible manner.

kohr-h:
To follow up on this, there is also the question of how to classify the methods that find a critical point of a function by simply finding a root of its gradient. An example is Newton's method. Other schemes explicitly go into the direction of smaller function value, so there it makes sense to speak of a minimizing method in the first place. So summarizing:

- Which methods are pure root finders? These should go to `findroot`
- Which methods are minimizers, possibly finding a gradient root, but with a strategy to ensure that it's a minimum?

kohr-h:
@ozanoktem @aringh  what's your take on this?

aringh:
I will have a look at it... part of this separation feels like it is highly 'personal' and the sense that for different persons it will be natural to look for different solvers under different categories.

adler-j:
We shouldn't have to many categories, the current structure is to fine imo. Categories should be mutually incompatible, for example a scalar optimizer vs a equation system solver.

aringh:
I agree that the number of categories should be kept small, in order to make it easier to find things. However the current categories are not optimal in my opinion; what is the difference between `vector` and `scalar`? Aren't most schemes `iterative`?

Also: the difference between root finders and optimization schemes is not always so sharp. Newton's method can be seen as solving `g(x) = 0`, but if `f(x)` is a (convex) function and one let `g(x) = grad_f(x)` then we are in fact solving the optimization problem with a second order method.

kohr-h:
There seems to be no really clear-cut distinction mathematically, so I guess we should choose something which makes sense from an implementation point of view. Not too many categories, I agree, rather a few "axes" along which we distinguish. Let me throw in another suggestion:

1. Solvers which directly attack the operator equation, be it `A(x) = y` or `A(x) = 0`. This typically includes methods which calculate an *approximate* solution to the normal equation in the linear case or do the same for the linearized problem in each step of a nonlinear problem. This would include
   * Landweber
   * CG and CGN
[...]

PR #502, opened 2016-08-11
ENH: add test for x in op.domain in all solvers, closes #291

### LINKED ISSUE #291: X in space checks for solvers
The solvers currently have very lightweight checks, it would occationally be nice to have these improved so we properly check input, that way we dont get cryptic messages and need to spend time figuring out what went wrong.

adler-j:
Added input sanitation to the solvers.

### DIFF
--- a/odl/solvers/advanced/chambolle_pock.py
+++ b/odl/solvers/advanced/chambolle_pock.py
@@ -148,13 +148,13 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
     """
     # Forward operator
     if not isinstance(op, Operator):
-        raise TypeError('`op` {} is not an instance of {}'
-                        ''.format(op, Operator))
+        raise TypeError('`op` {!r} is not an `Operator` instance'
+                        ''.format(op))
 
     # Starting point
     if x not in op.domain:
-        raise TypeError('`x` {} is not in the domain of `op` {}'
-                        ''.format(x.space, op.domain))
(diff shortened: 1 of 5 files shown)


kohr-h:
Good initiative. A few comments:

- When we raise `TypeError`, we should print the `repr` of at least the object that caused the error, so the user sees the provided type against the expected one in the error message. We could also just print the type, which would be a valid alternative.
- The Newton-type methods should not expect something named `grad` by default, that's a specific case. Now there was this inconsistency that he quasi-Newton ones take `grad` and Newton takes `op`, but IMO this inconsistency should be resolved in the other direction. Actually, I would prefer to keep the names untouched in this PR and do the restructuring separately, see #52.

adler-j:
Fixed according to comments. Merge?

odl/solvers/advanced/chambolle_pock.py
```diff
@@ -148,13 +148,13 @@ def chambolle_pock_solver(op, x, tau, sigma, proximal_primal, proximal_dual,
     """
     # Forward operator
     if not isinstance(op, Operator):
-        raise TypeError('`op` {} is not an instance of {}'
-                        ''.format(op, Operator))
+        raise TypeError('`op` {!r} is not an `Operator`'
```
kohr-h: We usually write `<type> instance`. Debatable for sure, but consistency is almost always good :-)

kohr-h:
See above, otherwise mergable.

PR #503, opened 2016-08-16
Add option to run examples with pytest

adler-j:
Adds the flag `--examples` to pytest, invoked as:

```
py.test --examples
```

this causes all examples in ODL to be run as part of the test suite, thus allowing easy tests that the examples at least do not crash.

Using this I also fixed several failing examples.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -38,6 +38,9 @@ def pytest_addoption(parser):
     parser.addoption('--benchmark', action='store_true',
                      help='Run benchmarks')
 
+    parser.addoption('--examples', action='store_true',
+                     help='Run examples')
+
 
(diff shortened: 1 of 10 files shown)


kohr-h:
Looks good. If possible, the option should be enabled in Travis, too.

adler-j:
This will not happen on Travis. The examples simply have way to many dependencies that travis will not support (ASTRA, STIR, etc) and adding checks for these would make the interface more complicated, something I'd like to avoid.

kohr-h:
Oh they need *all* dependencies? Okay, then no Travis.

kohr-h:
Release notes? :-)

PR #511, opened 2016-08-19
MAINT: harmonize names for padding parameters

kohr-h:
Simple pull to harmonize the names of the padding-related parameters as used in differential operators and the wavelet transform.

### DIFF
--- a/doc/source/release_notes.rst
+++ b/doc/source/release_notes.rst
@@ -12,6 +12,8 @@ Next release
 Changes
 --------
 - Changed definition of ``LinearSpaceVector.multiply`` to match the definition used by numpy (:pull:`509`)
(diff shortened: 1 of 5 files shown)


adler-j:
Release notes? :smile: 

Also have you made sure there are no changes in examples needed? Or in the doc?

kohr-h:
> Release notes? :smile:

There. Satisfied now?



> Also have you made sure there are no changes in examples needed? Or in the doc?

Apparently, there is neither an example nor documentation on differential operators and the wavelet transform.

adler-j:
> There. Satisfied now?

I was happy that I was given the chance to mention it :)

> Apparently, there is neither an example nor documentation on differential operators and the wavelet transform.

Added #512 and #513 for that.

Issue #410, opened 2016-05-29
Doctests should not need to import odl


kohr-h:
EDIT: hit the Enter button too early, here comes the text.

When writing running the doctests, often times one needs some functionality from another part of the library. Then, an explicit `import odl` is necessary, otherwise pytest complains. I noticed that in Numpy, they don't have to do that. Probably one can tell the `doctest` module thing what to import before running. I'll investigate that possibility.

adler-j:
Any idea on how to do this?

kohr-h:
In Numpy, they [modify the test context](https://github.com/numpy/numpy/blob/1429c606643d1ad305e710c4a31cb6f398d04c53/numpy/testing/noseclasses.py#L210-L221) by injecting `numpy` under the name `np` into all doctests. I'm not quite sure how to do that in `pytest`, but it must be possible.

adler-j:
Seems that we would need to do something similar to that then.

kohr-h:
Actually there is a functionality in `pytest` as shown in [this pull request](https://github.com/pytest-dev/pytest/pull/1428). It's been merged into the `features` branch but not yet to `master`. So at best, it becomes available in version 2.9.2. We'll have to wait.

adler-j:
That would require us to wait quite noticable time, at least half a few months after release, to allow people to catch up.

kohr-h:
Sure. We can still keep this issue open until then, just to remember it. Requiring a very new version of `pytest` is anyway only possible as soon as the major distributions provide it. Or we guarantee testing support only for installations using `pip` or `conda` where the newest packages are available.

adler-j:
This is done.

Issue #515, opened 2016-08-23
pytest 3.0, new features and breaking changes


adler-j:
[Pytest 3.0 has been released](http://blog.pytest.org/2016/whats-new-in-pytest-30/), notable changes that affect us include:
- [ ] entry point `py.test` changed to `pytest`, we need to update doc wherever needed
- [ ] `approx` function added, this should supersede many of our usages of `approximately_equal` etc. Basically all of them. `all_approximately_equal` may stay for a while since it is quite heavily optimized.
- [x] `doctest_namespace` added. We can use it to close #410 basically instantly.
- [x] `yield` tests are deprecated and need to be removed.

I assume we should update our pytest dependency to 3.0 rather soon. Perhaps in time for the 0.5.0 patch since this is breaking changes?

kohr-h:
These are really good news, a bunch of things we definitely want to have. On Anaconda, 3.0 is not yet available, but I guess that's only a matter of days. It's on PyPI already.
Not sure if we need a 0.5.0 release though. There are no breaking changes from the users' perspective. On the other hand, we need to update the dependency list, so that's a clear reason for a minor release. Would be okay with me, anyway.

adler-j:
This is done

PR #516, opened 2016-08-23
Issue 515  pytest 30

adler-j:
Updates to pytest 3.0 according to issue #515, also fixes issue #410

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -23,12 +23,29 @@ standard_library.install_aliases()
 
 import pytest
 import odl
+import numpy as np
 from odl.trafos.wavelet import PYWAVELETS_AVAILABLE
+
+
+# --- Add numpy and ODL to all doctests ---
+
+
+@pytest.fixture(autouse=True)
+def add_np(doctest_namespace):
+    doctest_namespace['np'] = np
+    doctest_namespace['odl'] = odl
+
+
+# --- Files to be ignored by the tests ---
+
 
 collect_ignore = ['setup.py', 'run_tests.py']
 
 if not PYWAVELETS_AVAILABLE:
     collect_ignore.append('odl/trafos/wavelet.py')
+
+
+# --- Options flags for the tests ---
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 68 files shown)


conftest.py
```diff
@@ -42,21 +59,25 @@ def pytest_addoption(parser):
...
 
 
-@pytest.fixture(scope="module", ids=ntuples_impl_ids, params=ntuples_impl_params)
+@pytest.fixture(scope="module", ids=ntuples_impl_ids,
+                params=ntuples_impl_params)
 def ntuples_impl(request):
+    """String with a available `NtuplesBase` implementation name."""
     return request.param
 
 ufunc_params = [ufunc for ufunc in odl.util.ufuncs.UFUNCS]
```
kohr-h: ??? `ufunc_params = odl.util.ufuncs.UFUNCS`?

odl/space/pspace.py
```diff
@@ -154,20 +154,21 @@ class instance can be retrieved from the space by the
...
-
         See also
         --------
         ProductSpaceVectorWeighting
         ProductSpaceConstWeighting
 
         Examples
         --------
-        >>> import odl
+        Product of r2 and r3
```
kohr-h: colon at the end

odl/space/pspace.py
```diff
@@ -154,20 +154,21 @@ class instance can be retrieved from the space by the
...
+
         >>> r2x3 = ProductSpace(odl.rn(2), odl.rn(3))
 
+        Can also create power spaces
```
kohr-h: Who can?
colon at the end

odl/space/pspace.py
kohr-h: Better: "Create the product of R^2 and R^3:"
Writing of the spaces is debatable, of course.

setup.py
```diff
@@ -131,7 +131,7 @@ def run_tests(self):
     package_dir={'odl': 'odl'},
 
     install_requires=[requires],
-    tests_require=['pytest'],
+    tests_require=['pytest>3.0'],
```
kohr-h: `>=`

kohr-h:
Update also the `conda/meta.yaml` file. Speaking of which, I'd like to wait with the merge until version 3.0 of `pytest` hits the anaconda repos.

Apart from that, all cool.

conftest.py
adler-j: *whistles*

adler-j:
Short update: pytest still has not been updated to version 3.0 in conda.

I am considering if we should break this into two parts, one that gets rid of the 3.0 deprecation warnings, and one with the new functionality.

kohr-h:
I know it's a pain having this PR lying around and assembling conflicts with master, but is getting rid of deprecation *warnings* really worth the trouble of splitting it? The pytest-3.0 specific part will have conflicts just in the same way. Honestly I think we can just wait.

adler-j:
Conda has now been updated to the new version of pytest, I say we go ahead here. I'll fix this.

kohr-h:
Ah, very good. I was already inclined to write that we go ahead anyway and recommend `pip install pytest` instead, but now that it's resolved, good.

adler-j:
This merge would be unbearable, faster to redo this. Will fix.

adler-j:
Closed in favor of new pull request

PR #518, opened 2016-08-24
Issue 464  deformation operators

chongchenmath:
Operators and functions for general and mass-preserving deformation are developed.

### DIFF
--- a/odl/deform/mass_preserving.py
+++ b/odl/deform/mass_preserving.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Operators and functions for general and mass-preserving deformation."""
+"""Operators and functions for geometric and mass-preserving deformations."""
 
 # Imports for common Python 2/3 codebase
 from __future__ import print_function, division, absolute_import
@@ -28,11 +28,11 @@ from odl.operator.operator import Operator
 from odl.space import ProductSpaceVector, ProductSpace
 
 
-__all__ = ('GeneralDeformFixedTempl', 'GeneralDeformFixedDeform',
+__all__ = ('GeometricDeformFixedTempl', 'GeometricDeformFixedDeform',
            'MassPresvDeformFixedTempl', 'MassPresvDeformFixedDeform')
 
 
-def _general_deform(template, deformation, out=None):
+def _geometric_deform(template, deformation, out=None):
     """Deform a template with a general deformation field.
 
     The function maps a given template ``I`` and a given deformation
@@ -67,7 +67,7 @@ def _general_deform(template, deformation, out=None):
     >>> deform_field_space = space.vector_field_space
     >>> template = space.element([0, 0, 1, 0, 0])
     >>> deform_field = deform_field_space.element([[0.1, 0.3, 0.5, 0.5, 0.9]])
-    >>> _general_deform(template, deform_field)
+    >>> _geometric_deform(template, deform_field)
     array([ 0.,  0.,  1.,  1.,  0.])
 
     The result depends on the chosen interpolation. With 'linear'
@@ -78,7 +78,7 @@ def _general_deform(template, deformation, out=None):
     >>> deform_field_space = space.vector_field_space
     >>> template = space.element([0, 0, 1, 0, 0])
     >>> deform_field = deform_field_space.element([[0.1, 0.3, 0.5, 0.6, 0.9]])
-    >>> _general_deform(template, deform_field)
+    >>> _geometric_deform(template, deform_field)
     array([ 0. ,  0. ,  1. ,  0.5,  0. ])
     """
     image_pts = template.space.points()
@@ -92,17 +92,18 @@ def _mass_presv_deform(template, mass_presv_dfield, out=None):
 
     The function maps a given template ``I``, a mass-preserving deformation
     field, i.e., a given deformation field ``phi`` and its Jacobian
-    determinant to the new function ``x --> |Dphi(x)|I(phi(x))``.
+    determinant ``|Dphi(x)|`` to the new function ``x --> |Dphi(x)|I(phi(x))``.
 
     Parameters
     ----------
     template : `DiscreteLpVector`
         Template to be deformed by a deformation field.
-    mass_presv_dfield : element of power space of power space and `DiscreteLp`
+    mass_presv_dfield : element of product space of vector field space
+        and `DiscreteLp`
         The mass-preserving deformation field used to deform the template.
-        It contains the deformation, i.e., an element of power space
-        of `DiscreteLp`, and the Jacobian determinant of the deformation,
-        i.e., a `DiscreteLpVector`.
+        It contains the deformation, i.e., an element of vector field space
+        or power space of `DiscreteLp`, and the Jacobian determinant of
+        the deformation, i.e., a `DiscreteLpVector`.
     out : `numpy.ndarray`, optional
         Array to which the function values of the deformed template
         are written. It must have the same shape as ``template`` and
@@ -118,32 +119,36 @@ def _mass_presv_deform(template, mass_presv_dfield, out=None):
     --------
     Create a simple 1D template ``I`` over ``[0, 1]``, a deffeomorphic
     deformation field and its Jacobian determinant.
-    If the deffeomorphism ``phi`` is ``sin(x*pi/2)``, its Jacobian determinant
+    If the diffeomorphism ``phi`` is ``sin(x*pi/2)``, its Jacobian determinant
     should be ``cos(x*pi/2)*pi/2`` and the output value is
     ``cos(x*pi/2)*pi/2*I(sin(x*pi/2))``. A mass-preserving deformation
     is obtained, namely, the integration of ``I(x)`` over ``[0, 1]``
     equals to the integration of ``|Dphi(x)|I(phi(x))`` over ``[0, 1]``.
 
-
+    Create relevant spaces
     >>> import odl
     >>> import numpy as np
     >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
     >>> vspace = space.vector_field_space
     >>> pspace = odl.ProductSpace(vspace, space)
+
+    Create template to be deformed
     >>> template = space.element([0, 0, 1, 0, 0])
-    >>> deform_field = vspace.element([np.sin(space.points() * np.pi / 2.0)])
-    >>> det = space.element(np.cos(space.points() * np.pi / 2.0) * np.pi / 2.0)
-    >>> mass_presv_field = pspace.element()
-    >>> mass_presv_field[0] = deform_field
-    >>> mass_presv_field[1] = det
-    >>> print(_mass_presv_deform(template, mass_presv_field))
+
+    Create mass-preserving deformation field
+    >>> deform_field = vspace.element([lambda x: np.sin(x * np.pi / 2.0)])
+    >>> det = space.element(lambda x: np.cos(x * np.pi / 2.0) * np.pi / 2.0)
+    >>> mass_presv_dfield = pspace.element([deform_field, det])
+
+    Output result
+    >>> print(_mass_presv_deform(template, mass_presv_dfield))
     [0.0, 1.07761764468, 0.0, 0.0, 0.0]
     """
-    general_deform_template = _general_deform(template, mass_presv_dfield[0])
+    general_deform_template = _geometric_deform(template, mass_presv_dfield[0])
     return mass_presv_dfield[1] * general_deform_template
 
 
-class GeneralDeformFixedTempl(Operator):
+class GeometricDeformFixedTempl(Operator):
 
     """Deformation operator with fixed template acting on deformation fields.
 
@@ -152,7 +157,7 @@ class GeneralDeformFixedTempl(Operator):
 
     See Also
     --------
-    GeneralDeformFixedDeform : Deformation with a fixed deformation.
+    GeometricDeformFixedDeform : Geometric deformation with fixed deformation.
 
     Notes
     -----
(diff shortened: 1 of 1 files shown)


adler-j:
Tests seem to be failing, try fixing them before we start reviewing.

chongchenmath:
Tests passed.

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+from odl.discr import DiscreteLp, DiscreteLpVector
+from odl.operator.operator import Operator
+from odl.space import ProductSpaceVector, ProductSpace
+
+
+__all__ = ('GeneralDeformFixedTempl', 'GeneralDeformFixedDeform',
+           'MassPresvDeformFixedTempl', 'MassPresvDeformFixedDeform')
+
+
+def _general_deform(template, deformation, out=None):
```
adler-j: This looks *very* similar to the "same" method in deform/linear. What is different? Could they perhaps be merged to a single function?

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+    >>> mass_presv_field[0] = deform_field
+    >>> mass_presv_field[1] = det
+    >>> print(_mass_presv_deform(template, mass_presv_field))
+    [0.0, 1.07761764468, 0.0, 0.0, 0.0]
+    """
+    general_deform_template = _general_deform(template, mass_presv_dfield[0])
+    return mass_presv_dfield[1] * general_deform_template
+
+
+class GeneralDeformFixedTempl(Operator):
```
adler-j: I don't quite follow how this is different from the linear case, what changes?

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+        image_pts[:, i] = vi.ntuple.asarray()
+    return template.interpolation(image_pts.T, out=out, bounds_check=False)
+
+
+def _mass_presv_deform(template, mass_presv_dfield, out=None):
+    """Deform a template with a mass-preserving deformation field.
+
+    The function maps a given template ``I``, a mass-preserving deformation
+    field, i.e., a given deformation field ``phi`` and its Jacobian
+    determinant to the new function ``x --> |Dphi(x)|I(phi(x))``.
```
adler-j: What is `Dphi`? What about `phi`?

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+
+    Parameters
+    ----------
+    template : `DiscreteLpVector`
+        Template to be deformed by a deformation field.
+    mass_presv_dfield : element of power space of power space and `DiscreteLp`
```
adler-j: "element of power space of power space and `DiscreteLp`" 

is complicated and feels wrong, should be something like

"element of **product** space of power space of (WHAT?) and `DiscreteLp`"

adler-j: Second thought:

this should definitely be split into two different parameters

odl/deform/mass_preserving.py
chongchenmath: In linear case, just the displacement is provided, but the deformation we use is `phi = Id + displacement`. Here the general deformation `phi` is given.

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+        ----------
+        template : `DiscreteLpVector`
+            Fixed template that is to be deformed.
+
+        Examples
+        --------
+        Create a simple 1D template ``I`` over ``[0, 1]`` to initialize the
+        operator and apply it to a mass-preserving deformation field, i.e.,
+        a deffeomorphic deformation field and its Jacobian determinant.
+        If the deffeomorphism ``phi`` is ``sin(x*pi/2)``, its Jacobian
```
adler-j: diffeomorphism

odl/deform/mass_preserving.py
chongchenmath: Please see the above.

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+        >>> import numpy as np
+        >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
+        >>> vspace = space.vector_field_space
+        >>> pspace = ProductSpace(vspace, space)
+        >>> template = space.element([0, 0, 1, 0, 0])
+        >>> op = MassPresvDeformFixedTempl(template)
+        >>> img_pts = space.points()
+        >>> deform_field = vspace.element([np.sin(img_pts * np.pi / 2.0)])
+        >>> det = space.element(np.cos(img_pts * np.pi / 2.0) * np.pi / 2.0)
+        >>> mass_presv_deform = pspace.element()
```
adler-j:
```python
>>> mass_presv_deform = pspace.element(deform_field, det)
```

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+
+        >>> import odl
+        >>> import numpy as np
+        >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
+        >>> vspace = space.vector_field_space
+        >>> pspace = ProductSpace(vspace, space)
+        >>> template = space.element([0, 0, 1, 0, 0])
+        >>> op = MassPresvDeformFixedTempl(template)
+        >>> img_pts = space.points()
+        >>> deform_field = vspace.element([np.sin(img_pts * np.pi / 2.0)])
```
adler-j: I prefer the syntax

```python
deform_field = vspace.element([lambda x: np.sin(x * np.pi / 2.0)])
```

Since it is more "continuous".

odl/deform/mass_preserving.py
```diff
@@ -0,0 +1,548 @@
...
+        >>> det = space.element(np.cos(img_pts * np.pi / 2.0) * np.pi / 2.0)
+        >>> mass_presv_deform = pspace.element()
+        >>> mass_presv_deform[0] = deform_field
+        >>> mass_presv_deform[1] = det
+        >>> print(op(mass_presv_deform))
+        [0.0, 1.07761764468, 0.0, 0.0, 0.0]
```
adler-j: This result is not obvious from the text and code, possibly because the example is quite length. Some further explanation could be needed.

odl/deform/mass_preserving.py
chongchenmath:
```python
>>> mass_presv_deform = pspace.element([deform_field, det])
```
works.

odl/deform/mass_preserving.py
adler-j: Not sure at all if we want to have a different method for this, the result is basically the same. Any opinions @kohr-h ?

odl/deform/mass_preserving.py
chongchenmath: Changed.

odl/deform/mass_preserving.py
chongchenmath: Modified.

odl/deform/mass_preserving.py
chongchenmath: I think this is supposed to be have. It will be used in LDDMM solver.

odl/deform/mass_preserving.py
chongchenmath: `phi`: deformation field
`|Dphi|`: Jacobian determinant of `phi`

odl/deform/mass_preserving.py
chongchenmath:
```python
mass_presv_dfield : element of product space of vector field space
        and `DiscreteLp`
```

odl/deform/mass_preserving.py
chongchenmath: More explanation given.

odl/deform/mass_preserving.py
kohr-h: Looks totally the same as `_linear_deform` to me, agree @adler-j . What's the point of this function?

odl/deform/mass_preserving.py
kohr-h: I'm sceptical to this design. The name suggests that this operator is supposed to perform "deformations in general", for "any deformation model that we choose". 

However, we never **explicitly represent** deformations, or, to put it more clearly, we don't have a type (class) that represents the **elements of some deformation (semi-) group**. 
In contrast, we directly represent **the group action** by a deformation operator and use a **vector field** instead to *implicitly* represent a deformation. 
As we learned from discussions with Klas, basically all deformation models are in some way based on vector fields, so that choice should not limit us in any way. Perhaps we need to revise the decision later, but now I don't see a reason for it.

So, in conclusion, I don't see a good reason to have a "general deformation operator", since our operators encompass both the steps "compute deformation from vector field" and "apply deformation", and the combination of those is *unique for each deformation model*.

odl/deform/mass_preserving.py
kohr-h: I don't remember the notion "deformation field" ever being used. To me, it should either be "deformation" or "displacement field".

odl/deform/mass_preserving.py
kohr-h: Definitely two separate parameters, lumping them into a single product space element just causes confusion.

kohr-h:
Before any more work is put into this PR:

What exactly is the point of the code in this branch? What does it do? Which issue does it solve? It's totally unclear to me, it seems like a mix of two things at least.

chongchenmath:
After click `view it on GitHub`, I cannot find several comments you @kohr-h did.

odl/deform/mass_preserving.py
chongchenmath: It is different. In linear case, the `displacement` is an argument, but the deformation `phi` we use is `phi = Id + displacement`. For this case, the general deformation `phi` is the argument.

odl/deform/mass_preserving.py
chongchenmath: The `deformation` is generated from `vector field` based on ODE. Then we need to use this `deformation` to deform the `template`. So the `GeometricDeformFixedTempl (GeneralDeformFixedTempl)` is presented here, which is just based on `deformation`. I think using combination is too complicated or not so flexible.

odl/deform/mass_preserving.py
chongchenmath: The `deformation` is Okay.

odl/deform/mass_preserving.py
chongchenmath: Any better method to deal with this?

kohr-h: Yes, make the `jacobian` an extra parameter here.

adler-j:
So, the functionality here seems ok (albeit lacking an example and test), but I think the design is not what we agreed on earlier. I'll try writing down my take-away of it here so we can discuss it.

Requirements
------------
There are a few things we need from the `deform` package in ODL which has become clear in discussions. Among these are:

* Fixed template deformations
* Fixed deformation deformations
* Mass preserving versions of the above
* Efficient composition of deformations of the above types

We have notably decided against adding a "general deformation" that takes both a deformation field *and* a template as input.

Design
------
We have a few options in the design, I'll try to break them down here.

### Deformation classes
Regarding the classes that should be added, we have (in my opinion) two main contenders:

* A two class design:
    * Fixed template (with option for pass preserving) 
    * Fixed deformation (with option for pass preserving)
* A four class design: 
    * Fixed template
    * Fixed deformation
    * Mass preserving fixed template
    * Mass preserving fixed deformation
    
In the discussion, we came to the conclusion that a two class design is favourable since it makes it easier to handle things like `(mass preserving deformation) composed with (non-mass preserving deformation)` while also reducing the amount of code required.

### Notation for deformations
In addition to this, we had a (probably too brief) discussion on how to represent the deformations. Here we have two options:

* The `linear_deform` style where the deformation field is a "offset": `x --> I(x + v(x))`
* The `general_deform` style where the deformation field is the point where the value should be taken from: `x --> I(x + v(x))`

Both of these are clearly interchangeable, simply add or subtract `x` from `v`. Hence i.m.o. we need to decide on one of these and stick with it. For me it is not obvious which one is best, and this may possibly be handled by adding a flag to the constructor so users can use the one they like.

Notes on this branch
--------------------

The current design seems to have a total of 6 classes, which I find to be redundant (for example, the `GeometricDeformFixedTempl` class has the exact same scope as `LinearDeformFixedTempl` but with a slightly different argument style). I think we should merge the classes like this:

* `GeometricDeformFixedTempl` + `LinearDeformFixedTempl` + `MassPresvDeformFixedTempl`  --> `LinearDeformFixedTempl`
* `GeometricDeformFixedDeform` + `LinearDeformFixedDeform` + `MassPresvDeformFixedDeform`  --> `LinearDeformFixedDeform`

where any options should be handled by flags in the `__init__` method.

We should also merge 

* `linear_deform` -> `general_deform` --> `deform_impl` (naming debatable)

Where we either decide for one notation (I'm not sure which one myself), or allow both through a flag.

adler-j:
What is the plan for this branch atm?

adler-j:
So my takeaway from our in person discussion yesterday is this:

* This branch has the correct functionality, but we should reduce the number of classes.
* We should remove the old `LinearDeform` classes and change all (if any) uses of them to use `GeometricDeform`, which does the same thing.
* Classes should be grouped according to their domain/range. I'll elaborate a bit more on why this is hard.

It is trivial for the non-mass preserving case since there we have the two cases: Fixed deformation and fixed template. However, it leads to a fundamental issue in how we handle the mass preserving cases. The core problem here is how we handle the `|Dv|` term, where the current implementation simply takes it as an extra parameter. Thus, the mass preserving cases are currently equivalent to simply left multiplying with `|Dv|`, which has to be provided by the user.

Since the user has to provide it, and since we provide no way at all for users to calculate it, the question of if this can truly be called a mass preserving deformation can be risen. Perhaps it is more of a "scaled deformation" or something? The natural next question is then ofc if we should keep this class at all, since the functionality already exists through left multiplication. I would probably answer yes to this, not because of the functionality currently available in this branch, but because it will be of interest once we add an efficient way to compose these, and/or once we add a factory that creates such a deformation given only a vector field (letting the deformation be `exp(-div(v))`.

So anyway, my takeaway is that it is hard to make judgements on if this is the correct design, mostly because we lack some core functionality. So the natural step for me would be to try adding composition of these operators and see where we end up.

Issue #520, opened 2016-08-24
power_method_opnorm should be intelligent


adler-j:
Currently it requires the parameter `niter` to be passed, I suggest we instead add a stop condition so that it can run "untill convergence".

PR #525, opened 2016-08-25
BUG: fix missing opnorm calc in diagnostics

### LINKED ISSUE #522: OperatorTest adjoint test bails out if no operator norm is given
This happens:
```python
>>> optest = odl.diagnostics.OperatorTest(my_op)
>>> optest.adjoint()
...

  File "/home/hkohr/git/odl/odl/diagnostics/operator.py", line 140, in _adjoint_definition
    denom = self.operator_norm * x_norm * y_norm

TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
```
Either a norm estimate should be computed in the init code, or an error should be raised in the above case. However, since it's a public method, it is "intended use" and should work.

kohr-h:
Closes #522

### DIFF
--- a/odl/diagnostics/operator.py
+++ b/odl/diagnostics/operator.py
@@ -95,15 +95,15 @@ class OperatorTest(object):
         return operator_norm
 
     def self_adjoint(self):
-        """Verify (Ax, y) = (x, Ay)."""
-        name = 'Verifying the identity (Ax, y) = (x, Ay)'
-
+        """Verify ``<Ax, y> == <x, Ay>``."""
         left_inner_vals = []
         right_inner_vals = []
 
-        quiet = not self.verbose
-        with FailCounter(name, 'error = |(Ax, y) - (x, Ay)| / '
-                         '||A|| ||x|| ||y||', quiet=quiet) as counter:
+        with FailCounter(
+                test_name='Verifying the identity <Ax, y> = <x, Ay>',
+                err_msg='error = |<Ax, y> - <x, Ay>| / ||A|| ||x|| ||y||',
+                verbose=self.verbose) as counter:
+
             for [name_x, x], [name_y, y] in samples(self.operator.domain,
                                                     self.operator.range):
                 x_norm = x.norm()
@@ -115,6 +115,7 @@ class OperatorTest(object):
                 denom = self.operator_norm * x_norm * y_norm
                 error = 0 if denom == 0 else abs(l_inner - r_inner) / denom
 
+                # TODO: allow specification of tolerance
                 if error > 0.00001:
                     counter.fail('x={:25s} y={:25s} : error={:6.5f}'
                                  ''.format(name_x, name_y, error))
@@ -128,14 +129,15 @@ class OperatorTest(object):
             print('(x, Ay) / (Ax, y) = {}. Should be 1.0'.format(scale))
 
     def _adjoint_definition(self):
-        """Verify (Ax, y) = (x, A^T y)."""
-        name = 'Verifying the identity (Ax, y) = (x, A^T y)'
-
+        """Verify ``<Ax, y> == <x, A^* y>``."""
         left_inner_vals = []
         right_inner_vals = []
 
-        with FailCounter(name, 'error = ||(Ax, y) - (x, A^T y)|| / '
-                         '||A|| ||x|| ||y||') as counter:
+        with FailCounter(
+                test_name='Verifying the identity <Ax, y> = <x, A^T y>',
+                err_msg='error = |<Ax, y< - <x, A^* y>| / ||A|| ||x|| ||y||',
+                verbose=self.verbose) as counter:
+
             for [name_x, x], [name_y, y] in samples(self.operator.domain,
                                                     self.operator.range):
                 x_norm = x.norm()
@@ -147,6 +149,7 @@ class OperatorTest(object):
                 denom = self.operator_norm * x_norm * y_norm
                 error = 0 if denom == 0 else abs(l_inner - r_inner) / denom
 
+                # TODO: allow specification of tolerance
                 if error > 0.00001:
                     counter.fail('x={:25s} y={:25s} : error={:6.5f}'
                                  ''.format(name_x, name_y, error))
@@ -160,7 +163,7 @@ class OperatorTest(object):
             print('(x, A^T y) / (Ax, y) = {}. Should be 1.0'.format(scale))
 
     def _adjoint_of_adjoint(self):
-        """Verify (A^*)^* = A"""
+        """Verify ``(A^*)^* == A``"""
         try:
             self.operator.adjoint.adjoint
         except AttributeError:
@@ -168,14 +171,14 @@ class OperatorTest(object):
             return
 
         if self.operator.adjoint.adjoint is self.operator:
-            print('(A^*)^* == A')
+            if self.verbose:
+                print('(A^*)^* == A')
             return
 
-        name = '\nVerifying the identity Ax = (A^T)^T x'
-
-        quiet = not self.verbose
-        with FailCounter(name, 'error = ||Ax - (A^T)^T x|| /'
-                         '||A|| ||x||', quiet=quiet) as counter:
+        with FailCounter(
+                test_name='\nVerifying the identity Ax = (A^*)^* x',
+                err_msg='error = ||Ax - (A^*)^* x|| / ||A|| ||x||',
+                verbose=self.verbose) as counter:
             for [name_x, x] in self.operator.domain.examples:
                 opx = self.operator(x)
                 op_adj_adj_x = self.operator.adjoint.adjoint(x)
@@ -185,6 +188,7 @@ class OperatorTest(object):
                     error = 0
                 else:
                     error = (opx - op_adj_adj_x).norm() / denom
+                # TODO: allow specification of tolerance
                 if error > 0.00001:
                     counter.fail('x={:25s} : error={:6.5f}'
                                  ''.format(name_x, error))
@@ -204,7 +208,7 @@ class OperatorTest(object):
             return
 
         if self.verbose:
-            print('\n== Verifying adjoint of operator ==\n')
+            print('\n== Verifying operator adjoint ==\n')
 
         domain_range_ok = True
         if self.operator.domain != self.operator.adjoint.range:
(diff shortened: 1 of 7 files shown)


odl/diagnostics/operator.py
```diff
@@ -53,8 +53,11 @@ def __init__(self, operator, operator_norm=None, verbose=True):
             can be estimated otherwise.
         """
         self.operator = operator
-        self.operator_norm = operator_norm
-        self.verbose = True
+        if operator_norm is None:
+            self.operator_norm = self._opnorm_est(self.operator.domain.one())
```
adler-j: This should be a call to `self.norm()`, no need to have two implementations.

odl/diagnostics/operator.py
```diff
@@ -79,16 +82,19 @@ def norm(self):
...
-
-            operator_norm = max(operator_norm, estimate)
+            operator_norm = max(operator_norm, self._opnorm_est(x))
 
         print('Norm is at least: {}'.format(operator_norm))
         self.operator_norm = operator_norm
         return operator_norm
 
+    def _opnorm_est(self, elem):
+        """Helper to calcualte an operator norm estimate."""
```
adler-j: Perhaps this deserves an upgrade? I.e. use `power_method_opnorm`, especially after #520 is fixed.

odl/diagnostics/operator.py
```diff
@@ -79,16 +82,19 @@ def norm(self):
 
         operator_norm = 0.0
```
adler-j: `operator_norm = max(self._opnorm_est(x) for _, x in samples(self.operator.domain))`

adler-j:
Good idea, minor stuff to fix.

odl/diagnostics/operator.py
kohr-h: `self.norm()` prints out lots of stuff, that's ugly during init.

odl/diagnostics/operator.py
adler-j: Well as long as it doesnt dominate the tests its quite fine. The space has about 10 or so examples, so a number of that magnitude is fine. For the adjoint test, 10^2 evaluations are done, so anything less than 100 wouldnt be to bad, for example.

odl/diagnostics/operator.py
kohr-h: Ok, now I use `self.norm()`.

odl/diagnostics/operator.py
kohr-h: Uses the power method now, with 2 iterations, which should be kind of okay.

odl/diagnostics/operator.py
kohr-h: OK

kohr-h:
I added proper handling of the `verbose` option in the `OperatorTest` class, which required an option to silence the `FailCounter` if no error occurred. Not sure if that's good, seemed easiest to me that way though.

odl/diagnostics/operator.py
```diff
@@ -75,17 +81,16 @@ def norm(self):
...
-        operator_norm = 0.0
-        for _, x in samples(self.operator.domain):
-            result = self.operator(x)
-            x_norm = x.norm()
-            estimate = 0 if x_norm == 0 else result.norm() / x_norm
+        if self.verbose:
+            print('\n== Calculating operator norm ==\n')
 
-            operator_norm = max(operator_norm, estimate)
+        operator_norm = max(power_method_opnorm(self.operator, niter=2,
```
adler-j: Will need to be fixed after #527 goes in.

odl/util/testutils.py
```diff
@@ -298,11 +298,12 @@ class FailCounter(object):
     ``*** FAILED 1 TEST CASE(S) ***``
     """
 
-    def __init__(self, test_name, err_msg=None):
+    def __init__(self, test_name, err_msg=None, quiet=False):
```
adler-j: This should clearly be called `verbose` as well, so that we dont need to do the `quiet = not verbose` all over the board.

adler-j:
Looking good except I'd change `quiet` to `verbose`, so we dont have two different definitions at once.

odl/diagnostics/operator.py
adler-j: Should be cool given that we test for a bunch of elements.

odl/util/testutils.py
kohr-h: That's true. When I started I had the mindset of totally silencing the counter if desired, but that's not what it ended up to be.

kohr-h:
I found a bug triggered by the example in the `diagnostics/operator` module. The non-integer power function on a `DiscreteLpVector` did not work. Fixed in 823c2c067de3d69aaef37484f7c10472f05b42f4.
Also cleanup according to comments & some more.

test/discr/lp_discr_test.py
```diff
@@ -867,6 +867,38 @@ def test_reduction(fn_impl, reduction):
...
+
+
+def test_power(fn_impl, power):
+    space = odl.uniform_discr([0, 0], [1, 1], [2, 2], impl=fn_impl)
+
+    x_arr, x = example_vectors(space, 1)
+    if int(power) != power:
+        # Make input positive to get real result
+        x_arr = np.abs(x_arr) + 0.1
+        x = np.abs(x) + 0.1
```
kohr-h: Will fail for negative powers if there are zeros in the array.

odl/diagnostics/operator.py
```diff
@@ -95,15 +95,15 @@ def norm(self):
...
-
+        """Verify ``<Ax, y> == <x, Ay>``."""
         left_inner_vals = []
         right_inner_vals = []
 
-        quiet = not self.verbose
-        with FailCounter(name, 'error = |(Ax, y) - (x, Ay)| / '
-                         '||A|| ||x|| ||y||', quiet=quiet) as counter:
+        with FailCounter(
+                test_name='Verifying the identity <Ax, y> = <x, Ay>',
```
adler-j: excellent improvement! much more self documenting now.

odl/diagnostics/operator.py
```diff
@@ -75,29 +81,29 @@ def norm(self):
         Wikipedia article on `Operator norm
         <https://en.wikipedia.org/wiki/Operator_norm>`_.
         """
-        print('\n== Calculating operator norm ==\n')
+        if self.verbose:
```
adler-j: This repetative "if self.verbose: print(something)" could be changed to adding a function 

```
def log(self, string):
    if self.verbose:
        print(string)
```

that way we can remove quite some lines of code and also make this extendable in the future.

odl/discr/discretization.py
```diff
@@ -543,6 +543,13 @@ def __init__(self, space, data):
         assert isinstance(space, DiscretizedSpace)
         DiscretizedSetVector.__init__(self, space, data)
 
+    def __ipow__(self, p):
+        """Implement ``self **= p``."""
+        # Falls back to `LinearSpaceVector.__ipow__` if `self.ntuple`
+        # has no own `__ipow__`. The fallback only works for integer `p`.
```
adler-j: "The fallback only works for integer `p`." 

needs not be valid, certain ntuples may chose to override `__pow__` to allow arbitrary `p`. 

Frankly we should make passthroughs for all the arithmetic things in case someone overrides in the ntuple.

odl/diagnostics/operator.py
kohr-h: Was thinking of something like that, but too lazy :-)
Will do that.

odl/discr/discretization.py
kohr-h: The way we implemented it, `__pow__` calls `__ipow__` on a copy, so `__ipow__` has the last word. If an `ntuple` overrides `__pow__`, this `__ipow__` method will never be called (unless that new implementation calls it). If an `ntuple` overrides `__ipow__`, all good. If not, we get `LinearSpace.__ipow__`. I don't see a logical error in the comment.

adler-j: It seems you are correct, needed some thought here :/

Anyway:

> Frankly we should make passthroughs for all the arithmetic things in case someone overrides in the ntuple.

odl/diagnostics/operator.py
```diff
@@ -62,6 +62,11 @@ def __init__(self, operator, operator_norm=None, verbose=True):
 
         self.verbose = bool(verbose)
 
+    def log(self, message):
```
adler-j: Should also be done in `SpaceTest` :)

odl/diagnostics/operator.py
```diff
@@ -75,29 +86,27 @@ def norm(self):
...
+        """Verify ``<Ax, y> == <x, Ay>``."""
         left_inner_vals = []
         right_inner_vals = []
 
-        with FailCounter(name, 'error = ||(Ax, y) - (x, Ay)|| / '
-                         '||A|| ||x|| ||y||') as counter:
+        with FailCounter(
+                test_name='Verifying the identity <Ax, y> = <x, Ay>',
+                err_msg='error = |<Ax, y> - <x, Ay>| / ||A|| ||x|| ||y||',
+                verbose=self.verbose) as counter:
```
adler-j: Now, I'm going to ask you to rewrite this again. Instead of passing verbose here, why not create an argument "out" or something and then pass `self.log`. That way the failcounter can write to the actual log?

kohr-h: Easy one

odl/discr/discretization.py
kohr-h: See #533

kohr-h:
Space tests now also cleaned up

odl/diagnostics/space.py
```diff
@@ -141,160 +149,183 @@ def field(self):
...
 
     def _inverse_element_of_addition(self):
-        """Check additional inverse."""
-        name = 'Inverse element of addition, x + (-x) = 0'
-        zero = self.space.zero()
+        """Verify additive inverse."""
+        try:
+            zero = self.space.zero()
+        except (AttributeError, NotImplementedError):
+            print('*** SPACE HAS NO ZERO VECTOR ***')
```
adler-j: log

odl/diagnostics/space.py
```diff
@@ -567,161 +627,190 @@ def _dist_subadditive(self):
...
     def _multiply_zero(self):
-        """Check vector multiplication with zero is zero."""
-        name = 'Multiplication by zero, x * 0 = 0'
-
-        zero = self.space.zero()
+        """Verify that vector multiplication with zero is zero."""
+        try:
+            zero = self.space.zero()
+        except NotImplementedError:
+            print('*** SPACE HAS NO ZERO VECTOR ***')
```
adler-j: log

odl/diagnostics/space.py
```diff
@@ -567,161 +627,190 @@ def _dist_subadditive(self):
...
+            ``a * (x + y) = a * x + a * y``
+
+            ``x * (y + z) = x * y + x * z``
+        """
+        self.log('\n== Verifying multiplication ==\n')
 
         try:
             zero = self.space.zero()
+        except NotImplementedError:
+            print('*** SPACE HAS NO ZERO VECTOR ***')
```
adler-j: log

odl/diagnostics/space.py
```diff
@@ -783,66 +876,77 @@ def contains(self):
...
                 if not correct:
                     counter.fail('modified y, x changed with x={:25s}'
                                  ''.format(n_x))
 
     def vector_set_zero(self):
         """Verify `LinearSpaceVector.set_zero`."""
+        try:
+            zero = self.space.zero()
+        except NotImplementedError:
+            print('*** SPACE HAS NO ZERO VECTOR ***')
```
adler-j: you get the gist

odl/util/testutils.py
```diff
@@ -317,7 +318,8 @@ def fail(self, string=None):
 
     def __exit__(self, type, value, traceback):
         if self.num_failed == 0:
-            print('{:<70}: Completed all test cases.'.format(self.test_name))
+            self.log('{:<70}: Completed all test cases.'
+                     ''.format(self.test_name))
         else:
             print(self.test_name)
```
adler-j: else log?

odl/diagnostics/space.py
kohr-h: Sure? To me it looks like an error.

kohr-h:
Apart from logging comments, good to go?

odl/util/testutils.py
kohr-h: This `else` is the error case, I think it should print then regardless of verbosity.

adler-j:
Looks good!

PR #529, opened 2016-08-25
ENH: expose ellipse_phantom and utils, closes #519

### LINKED ISSUE #519: Expose utilities to create ellipse phantoms
To enable users to create ellipse-like phantoms 

* The function `ellipse_phantom` should be moved to `odl.phantom.geometric` and exposed.
* `_shepp_logan_ellipses` should be exposed and made more simple to use. For example the `modified` parameter should be added.
* The function `_modified_shepp_logan_ellipses` should return the list

adler-j:


odl/phantom/geometric.py
```diff
@@ -171,23 +171,345 @@ def indicate_proj_axis(space, scale_structures=0.5):
...
+        rectangle [-1, -1] x [1, 1] or equivalent in 3d.
+
+    Notes
+    -----
+    The phantom is created by adding the values of each ellipse. The ellipses
+    are defined by a center point (center_x, center_y, [center_z]) and the
+    length of its principial axes (axis_1, axis_2, [axis_2]) and euler angles.
+
+    This function is heavily optimized, achieving runtimes about 20 times
+    faster than "trivial" implementations. It is therefore recommended to use
```
kohr-h: use it

kohr-h:
What exactly happened in this pull? It's hard to see since stuff has been moved and modified at the same time.

Next time, please make a separate commit for the move only, then change the code. It's also much more better when trying to rebase, since git also has a hard time understanding move-and-modify commits.

adler-j:
> What exactly happened in this pull? It's hard to see since stuff has been moved and modified at the same time.

Basically exposed `ellipse_phantom` and added a `shepp_logan_ellipses` that can be used with it.

> Next time, please make a separate commit for the move only, then change the code. It's also much more better when trying to rebase, since git also has a hard time understanding move-and-modify commits.

Duly noted.

adler-j:
Will merge after CI.

kohr-h:
> Will merge after CI.

Not so fast, I'll have a look at the texts first.

odl/phantom/geometric.py
```diff
@@ -171,23 +171,345 @@ def indicate_proj_axis(space, scale_structures=0.5):
...
+
+def ellipse_phantom(space, ellipses):
+    """Return a phantom given by ellipses.
+
+    Parameters
+    ----------
+    space : `DiscreteLp`
+        Space the phantom should live in, must be 2 or 3 dimensional.
+    ellipses : `sequence` of `sequence`
+        If ``space`` is 2 dimensional each row should contain:
```
kohr-h: 2-dimensional

odl/phantom/geometric.py
```diff
@@ -171,23 +171,345 @@ def indicate_proj_axis(space, scale_structures=0.5):
...
+    return space.element(p)
+
+
+def ellipse_phantom(space, ellipses):
+    """Return a phantom given by ellipses.
+
+    Parameters
+    ----------
+    space : `DiscreteLp`
+        Space the phantom should live in, must be 2 or 3 dimensional.
```
kohr-h: "live in" weird, better "be created in"
The whole sentence could be clearer, like "Space in which the phantom is created, must be 2- or 3-dimensional."

odl/phantom/geometric.py
```diff
@@ -171,23 +171,345 @@ def indicate_proj_axis(space, scale_structures=0.5):
...
+    ellipses : `sequence` of `sequence`
+        If ``space`` is 2 dimensional each row should contain:
+
+        'value', 'axis_1', 'axis_2', 'center_x', 'center_y', 'rotation'
+
+        If space is 3 dimensional each row should contain:
```
kohr-h: 3-dimensional

odl/phantom/transmission.py
```diff
@@ -22,14 +22,17 @@
...
 
-__all__ = ('shepp_logan',)
+__all__ = ('shepp_logan_ellipses', 'shepp_logan')
 
 
 def _shepp_logan_ellipse_2d():
-    """Return ellipse parameters for a 2d Shepp-Logan phantom."""
+    """Return ellipse parameters for a 2d Shepp-Logan phantom.
+
+    This assumes that the ellipses are on the square [-1, -1]x[-1, -1].
```
kohr-h: "are on" -> "are contained in"

odl/phantom/transmission.py
```diff
@@ -44,7 +47,10 @@ def _shepp_logan_ellipse_2d():
 
 
 def _shepp_logan_ellipse_3d():
-    """Return ellipse parameters for a 3d Shepp-Logan phantom."""
+    """Return ellipse parameters for a 3d Shepp-Logan phantom.
+
+    This assumes that the ellipses are on the cube [-1, -1, -1]x[1, 1, 1].
```
kohr-h: "are contained in"

odl/phantom/transmission.py
```diff
@@ -65,12 +71,14 @@ def _modified_shepp_logan_ellipses(ellipses):
     """
     intensities = [1.0, -0.8, -0.2, -0.2, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
 
+    assert len(ellipses) == len(intensities)
```
kohr-h: Should be a proper error message.

odl/phantom/transmission.py
```diff
@@ -95,6 +108,27 @@ def shepp_logan(space, modified=False):
...
+    return ellipses
+
+
+def shepp_logan(space, modified=False):
+    """Standard `Shepp-Logan phantom`_ in 2 or 3 dimensions.
+
+    Parameters
+    ----------
+    space : `DiscreteLp`
+        The 2/3 dimension space that the phantom should be created in.
```
kohr-h: 2- or 3-dimensional

odl/phantom/transmission.py
```diff
@@ -95,6 +108,27 @@ def shepp_logan(space, modified=False):
...
+    modified : `bool`, optional
+        True if the modified Shepp-Logan phantom should be given.
+        The modified phantom has greatly amplified contrast to aid in
```
kohr-h: aid ~~in~~

odl/phantom/transmission.py
```diff
@@ -95,6 +108,27 @@ def shepp_logan(space, modified=False):
...
+        visualization.
+
+    References
+    ----------
+    .. Shepp-Logan phantom: en.wikipedia.org/wiki/Shepp–Logan_phantom
```
kohr-h: `.. _Shepp-Logan phantom` (underscore missing)

kohr-h:
Some minor stuff, after that you can go ahead :-)

odl/phantom/transmission.py
adler-j: internal helpers can do asserts, no? I thought that was policy.

kohr-h: I'm not so sure about that. To me, it depends on the nature of the internal helper. For instance, if the public function delegates everything, including checks, to internal helpers, I'm against asserts since they provide users with very little helpful information. They're not so internal in such cases, after all, and to me, this is such a case.

I'm okay with asserts when a deviation from the check is a bug rather than the result of invalid input. Internal consistency checks can use asserts, input validation should not in my opinion.

adler-j: Well this function should not be called by any external user, ever. The only calls come through from `shepp_logan_ellipses`, and this check is to make sure that the number of ellipses in those are OK, which they currently are.

The only way to make this assert fail is by either calling a private method or by editing the ODL source code. To me that is a typical case for asserts?

kohr-h: So there's  no way for a user to provide own ellipses? In that case it's OK.

kohr-h:
This should also be mentioned in the release notes, BTW.

odl/phantom/transmission.py
adler-j: Not to this one, it's basically a utility so we wont have to write two functions that create the modified ellipses. If a user wants to use this the correct usage would be something like

```python
ellipses = shepp_logan_ellipses(3, modified=True)
# modify ellipses however you want
```

adler-j:
> This should also be mentioned in the release notes, BTW.

It is: 

Expose ``ellipse_phantom`` and ``shepp_logan_ellipses`` to ``odl.phantom``

Issue #530, opened 2016-08-25
Move usage examples from _call to __init__


adler-j:
We currently have documentation on how operators behave in the _call method. This is ok for the webpage, but for editors we need this to show up more simply, and there is no reasonable way to find _call as it is right now.

My suggestion is that we move the `_call` examples into `__init__`. This should reduce some code overhead (since we dont need to init things twice) and also make finding this information much easier.

kohr-h:
Totally agree.

PR #532, opened 2016-08-26
ENH: allow syntax (op, n) in the ProductSpaceOperators, closes #517

### LINKED ISSUE #517: Add syntax BroadcastOperator(op, n)
This would be equivalent to `ProductSpace(space, n)` and do what is expected of it:

```python
>>> op = BroadcastOperator(op, 3)
```

should equal

```python
>>> op = BroadcastOperator(op, op, op)
```

adler-j:
Added option similar to `ProductSpace(space, n)` to `DiagonalOperator` and others. Also did some work on the documentation, related to #530.

### DIFF
--- a/odl/operator/pspace_ops.py
+++ b/odl/operator/pspace_ops.py
@@ -25,6 +25,7 @@ from builtins import super
 
 import numpy as np
 import scipy as sp
+from numbers import Integral
 
 from odl.operator.operator import Operator
 from odl.space import ProductSpace
@@ -39,6 +40,12 @@ class ProductSpaceOperator(Operator):
 
     """A "matrix of operators" on product spaces.
 
+    For example a matrix of operators can act on a vector by::
+
+    ProductSpaceOperator([[A, B], [C, D]])([x, y]) = [A(x) + B(y), C(x) + D(y)]
+
+    Notes
+    -----
     This is intended for the case where an operator can be decomposed
     as a linear combination of "sub-operators", e.g.
 
@@ -83,17 +90,11 @@ class ProductSpaceOperator(Operator):
 
         :math:`[\mathcal{A}(x)]_i = \sum_{j=1}^m \mathcal{A}_{ij}(x_j)`.
 
-    Notes
-    -----
-    In many cases it is of interest to have an operator from a `ProductSpace`
-    to any `LinearSpace`. It that case this operator can be used with a slight
-    modification, simply run
-
-    ``prod_op = ProductSpaceOperator(prod_space, ProductSpace(linear_space))``
-
-    The same can be done for operators `LinearSpace` -> `ProductSpace`
-
-    ``prod_op = ProductSpaceOperator(ProductSpace(linear_space), prod_space)``
+    See Also
+    --------
+    BroadcastOperator : Case when a single argument is used by several ops.
+    ReductionOperator : Calculates sum of operator results.
+    DiagonalOperator : Case where the 'matrix' is diagonal.
     """
 
     def __init__(self, operators, domain=None, range=None):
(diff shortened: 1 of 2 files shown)


odl/operator/pspace_ops.py
```diff
@@ -83,17 +90,11 @@ class ProductSpaceOperator(Operator):
 
         :math:`[\mathcal{A}(x)]_i = \sum_{j=1}^m \mathcal{A}_{ij}(x_j)`.
 
-    Notes
```
adler-j: This is somewhat redundant now that we have the other three operators that cover this case.

odl/operator/pspace_ops.py
```diff
@@ -254,6 +255,7 @@ def _call(self, x, out=None):
             [1.0, 2.0, 3.0]
         ])
         """
+        # TODO: add optimization in case a operator appears repeatedly in a row
```
kohr-h: "an" :-)

odl/operator/pspace_ops.py
```diff
@@ -755,13 +791,53 @@ class ReductionOperator(Operator):
...
+        >>> op.range
+        rn(3)
+
+        Evaluating in a point gives sum
+
+        >>> op([[1.0, 2.0, 3.0],
+        ...     [4.0, 6.0, 8.0]])
+        rn(3).element([9.0, 14.0, 19.0])
+
+        Can also be created using a multiple of a single operator
```
kohr-h: Colons

odl/operator/pspace_ops.py
```diff
@@ -776,7 +852,7 @@ def prod_op(self):
 
     @property
     def operators(self):
-        """A tuple of sub-operators."""
+        """`tuple` of sub-operators."""
```
kohr-h: Tuple, no backticks needed.

odl/operator/pspace_ops.py
```diff
@@ -788,10 +864,10 @@ def _call(self, x, out=None):
...
         Parameters
         ----------
-        x : domain element
-            Input vector that should be used in the reduction
-        out : range element, optional
-            output vector to write result to
+        x : `domain` element
+            Input vector that should be used in the reduction.
+        out : `range` element, optional
+            Output vector to write result to.
```
kohr-h: the result

odl/operator/pspace_ops.py
```diff
@@ -890,28 +957,69 @@ def adjoint(self):
 
 
 class DiagonalOperator(ProductSpaceOperator):
-    """Diagonal 'matrix' of operators
+    """Diagonal 'matrix' of operators.
 
-    For example, if A and B are operators
+    For example, if A and B are operators the diagonaloperator can be seen as a
```
kohr-h: diagonal operator

kohr-h:
Looking good, some of the doc stuff will be dealt with during a proper overhaul.

odl/operator/pspace_ops.py
adler-j: Well, writing `tuple` implies that it actually is of type `tuple` and not simply a tuple of operators. This is again a case of english vs python. I guess this could be made clear by adding a `Returns`, but this is likely enough.

adler-j:
Merge after CI.

odl/operator/pspace_ops.py
kohr-h: On the other hand, since we try to adhere to the principle "be generous in what you accept and strict in what you return", there should be no doubt about the return type if an English word is used that has a corresponding Python type of the same name. 
We should assume that users know what a tuple is, and if we write `Tuple of foo`, what should the return type reasonably be apart from `tuple`? 
Again, with the same line of reasoning, we could easily start flagging all kinds of statements that "could be clearer".

PR #534, opened 2016-08-26
Documentation review, closes #408 (for now)

### LINKED ISSUE #408: Documentation review

kohr-h:
Extremely boring stuff. Most significant is the improvement of the docs and error messages in the `odl.set.space` module, which was kind of old and not up to current standards.

TODO (up to discussion also):

- [x] Double-backtick occurrences of `True`, `False` and `None`, possibly similar other names
- [x] Check if all places of single-vs-double underscore attributes are correct (tests should raise though if anything is wrong)
- [x] Rebase on master (it's only slightly behind, though, worst conflicts are fixed)

### DIFF
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -563,7 +563,7 @@ def dspace_type(space, impl, dtype=None):
         consistent with it.
     impl : string
         Implementation backend for the data space
-    dtype : type, optional
+    dtype : `numpy.dtype`, optional
         Data type which the space is supposed to use. If ``None``, the
         space type is purely determined from ``space`` and
         ``impl``. If given, it must be compatible with the
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -206,10 +206,11 @@ class DiscretizedSet(NtuplesBase):
         Returns
         -------
         equals : bool
-            True if ``other`` is a `DiscretizedSet`
+            ``True`` if ``other`` is a `DiscretizedSet`
             instance and all attributes `uspace`, `dspace`,
             `DiscretizedSet.sampling` and `DiscretizedSet.interpolation`
-            of ``other`` and this discretization are equal, False otherwise.
+            of ``other`` and this discretization are equal, ``False``
+            otherwise.
         """
         if other is self:
             return True
@@ -281,7 +282,7 @@ class DiscretizedSetVector(NtuplesBaseVector):
 
         Parameters
         ----------
-        out : numpy.ndarray, optional
+        out : `numpy.ndarray`, optional
             Array in which the result should be written in-place.
             Has to be contiguous and of the correct dtype.
         """
@@ -293,8 +294,8 @@ class DiscretizedSetVector(NtuplesBaseVector):
         Returns
         -------
         equals : bool
-            True if all entries of ``other`` are equal to this
-            vector's entries, False otherwise.
+            ``True`` if all entries of ``other`` are equal to this
+            vector's entries, ``False`` otherwise.
         """
         return (other in self.space and
                 self.ntuple == other.ntuple)
@@ -304,7 +305,7 @@ class DiscretizedSetVector(NtuplesBaseVector):
 
         Parameters
         ----------
-        indices : int or slice
+        indices : int or `slice`
             The position(s) that should be accessed
 
         Returns
@@ -319,9 +320,9 @@ class DiscretizedSetVector(NtuplesBaseVector):
 
         Parameters
         ----------
-        indices : int or slice
+        indices : int or `slice`
             The position(s) that should be set
-        values : scalar, array-like or `NtuplesBaseVector`
+        values : scalar, `array-like` or `NtuplesBaseVector`
             The value(s) that are to be assigned.
 
             If ``index`` is an int, ``value`` must be single value.
@@ -488,7 +489,7 @@ class DiscretizedSpace(DiscretizedSet, FnBase):
 
     @property
     def is_weighted(self):
-        """Return True if the ``dspace`` is weighted."""
+        """``True`` if the ``dspace`` is weighted."""
         return getattr(self.dspace, 'is_weighted', False)
 
     def _lincomb(self, a, x1, b, x2, out):
@@ -560,10 +561,10 @@ def dspace_type(space, impl, dtype=None):
         Template space from which to infer an adequate data space. If
         it has a `LinearSpace.field` attribute, ``dtype`` must be
         consistent with it.
-    impl : `str`
+    impl : string
         Implementation backend for the data space
-    dtype : `type`, optional
-        Data type which the space is supposed to use. If None, the
+    dtype : type, optional
+        Data type which the space is supposed to use. If ``None``, the
         space type is purely determined from ``space`` and
         ``impl``. If given, it must be compatible with the
         field of ``space``.
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -205,11 +205,11 @@ class DiscretizedSet(NtuplesBase):
 
         Returns
         -------
-        equals : `bool`
-            `True` if ``other`` is a `DiscretizedSet`
+        equals : bool
+            True if ``other`` is a `DiscretizedSet`
             instance and all attributes `uspace`, `dspace`,
             `DiscretizedSet.sampling` and `DiscretizedSet.interpolation`
-            of ``other`` and this discretization are equal, `False` otherwise.
+            of ``other`` and this discretization are equal, False otherwise.
         """
         if other is self:
             return True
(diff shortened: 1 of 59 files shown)


adler-j:
> Double-backtick occurrences of True, False and None, possibly similar other names

I like double (or single) ticking them since it gives a nice distinction to the english language words. I.e.

"Returns 13 if ``something`` is True."

vs

"Returns 13 if `something` is `True`."

I think the second one is more clear in what is python thingies and what is English. This is partly why I like adding ticks to things like `float` over writing float.

kohr-h:
> I like double (or single) ticking them since it gives a nice distinction to the english language words. I.e.
> 
> "Returns 13 if something is True."
> 
> vs
> 
> "Returns 13 if something is True."
> 
> I think the second one is more clear in what is python thingies and what is English. This is partly why I like adding ticks to things like float over writing float.

That's true for texts without hard size limit, like parameter explanations and additional text in docstrings, with the restriction that there should not be too many highlighted parts. If everything is highlighted, nothing is highlighted.
In the type definition lines of parameters, we have a hard limit, and I think we should not waste available characters to backticks there.

Apart from that, I think making hyperlinks to the Python docs for standard notions like `float`, `sequence` or `callable` really goes too far. That's a bit like linking for the sake of linking.

adler-j:
> That's true for texts without hard size limit, like parameter explanations and additional text in docstrings, with the restriction that there should not be too many highlighted parts. If everything is highlighted, nothing is highlighted.
In the type definition lines of parameters, we have a hard limit, and I think we should not waste available characters to backticks there.

Certainly agree that we occationaly go overboard here.

> Apart from that, I think making hyperlinks to the Python docs for standard notions like float, sequence or callable really goes too far. That's a bit like linking for the sake of linking.

Well if I were to ask 10 of our though about users, I bet you 8 would think sequence is a english word and not a well defined python concept. Adding a link here enforced that we actually intend the python concept and lets them google if they get some weird error message.

kohr-h:
> Well if I were to ask 10 of our though about users, I bet you 8 would think sequence is a english word and not a well defined python concept. Adding a link here enforced that we actually intend the python concept and lets them google if they get some weird error message.

Valid point. But where to draw the line? I'd say

- No: `float`, `int`, `bool`, `str`, `list`, `tuple`, `function`, `class`
- Yes: `sequence`, `callable`

adler-j:
I'd draw the line at "general" programming concepts that don't vary notable between languages, i.e.

No: `float`, `int`, `complex`, `object`, `function`, `class`
Yes: `tuple`, `list`, `callable`, `sequence`, `iterable`

Reasoning for e.g. `list` is that it is a linked list in most languages (java, c++) but the python version is actually more like a `ArrayList` in java, for example.

Also only the first time they are named a link is needed, after that users should be able to follow.

kohr-h:
> Reasoning for e.g. list is that it is a linked list in most languages (java, c++) but the python version is actually more like a ArrayList in java, for example.

With the same reasoning, one could also link `float` due to the dynamic way it is handled in Python in contrast to other languages. IMO we should be able to assume at least some basic Python knowledge and familiarity with the very basic concepts. Lists and tuples are among the first things one encounters in tutorials and books, and I don't think there is a need for linking here.
The concepts of sequences and callables are a bit more abstract and come significantly later in a typical Python learning history, so here it's okay to link.

kohr-h:
Rebased and cleaned up according to the last round of discussion. In this version, `list`, `tuple` and `dict` have no backticks, which is my personal viewpoint. All other changes should be conflict-free :-). 

In case we want to continue discussing the matter, I suggest we merge this branch anyway and change again later if necessary. Every other PR introduces new conflicts, so it would be good to get this out of the way.

kohr-h:
Oh man, I forgot `None`. :angry:

kohr-h:
> Oh man, I forgot None. :angry:

Done with that also now.

odl/discr/discretization.py
```diff
@@ -560,10 +561,10 @@ def dspace_type(space, impl, dtype=None):
         Template space from which to infer an adequate data space. If
         it has a `LinearSpace.field` attribute, ``dtype`` must be
         consistent with it.
-    impl : `str`
+    impl : string
         Implementation backend for the data space
-    dtype : `type`, optional
-        Data type which the space is supposed to use. If `None`, the
+    dtype : type, optional
```
adler-j: type basically says nothing. How about "`numpy.dtype`  like", or something?

odl/discr/lp_discr.py
```diff
@@ -1049,34 +1050,34 @@ def uniform_discr_fromintv(interval, nsamples, exponent=2.0, interp='nearest',
     ----------
     interval : `IntervalProd`
         The domain of the uniformly discretized space.
-    nsamples : `int` or `tuple` of `int`
+    nsamples : int or `sequence` of ints
         Number of samples per axis. For dimension >= 2, a tuple is
```
adler-j: perhaps missleading to leave tuple here

odl/discr/discretization.py
kohr-h: One can say `numpy.dtype` directly since it's a type itself, not a function.

odl/discr/lp_discr.py
kohr-h: Definitely misleading! Should be sequence.

odl/operator/default_ops.py
```diff
@@ -44,10 +44,9 @@ def __init__(self, space, scalar):
         Parameters
         ----------
         space : `LinearSpace`
-            Space of elements which the operator is acting on
-        scalar : `LinearSpace.field` `element`
-            An element of the field of the space which vectors are
-            scaled with
+            Space of elements on which this operator acts.
+        scalar : `LinearSpace.field` `element-like`
```
adler-j: I'd prefer ``space.field`` `element-like`

odl/set/domain.py
```diff
@@ -41,21 +41,23 @@ class IntervalProd(Set):
...
     def __init__(self, begin, end):
         """Initialize a new instance.
 
         Parameters
         ----------
-        begin : `array-like` or `float`
-            The lower ends of the intervals in the product
-        end : `array-like` or `float`
-            The upper ends of the intervals in the product
+        begin : `array-like` or float
```
adler-j:
```
begin, end : float or `array-like`
```

I.m.o. put the scalar as the first option (being the most simple) and you can condense this to a single entry with no loss of information.

adler-j:
Damn it this is simply too much to read and it is basically 98% correct and the remaining 2% is very minor style stuff.

I say go ahead.

odl/operator/default_ops.py
kohr-h: Good catch, we should have an eye on all these unspecific types and corresponding checks - many can be replaced by `vec in space` checks.

odl/set/domain.py
kohr-h: I'll take care of that in #452.

kohr-h: Done in  	9768227.

kohr-h:
All comments done, will merge after CI.

Issue #506, opened 2016-08-17
Numpy like utility for versioning


adler-j:
Currently we have a version number of ODL in a few places (I know at least 2, setup.py and __init__.py).

Numpy has a utility which creates a `version.py` file on installation, hence enforcing one and only one version number throughout the package. I think this is something we could use.

kohr-h:
Would be good, agreed. I once looked into their solution, too, but didn't bring enough time to understand it properly.

PR #536, opened 2016-08-27
Build and deploy docs with Travis, fixes #535

### LINKED ISSUE #535: Read the docs is dead
It is down once again due to long computation times. This is becoming a hassle. We have a few options atm:

* Reduce duplication: We currently rebuild the doc even for inherited methods. Thus we build a total of 2072 methods while we only "need" to build 1024 of them. This could buy us "twice the time".
* Somehow move the doc building to the release script and build it ourselves.

At this point, I'm all for alternative 2. It would also allow us to use intersphinx and other things that read the docs currently lack.

kohr-h:
I found [a nice package](https://github.com/Syntaf/travis-sphinx) for automated building and deploying of Sphinx documentation for integration with Travis CI. Good and simple instructions, so setting up was no big deal.

The documentation is put into the `gh-pages` branch of the project, which is a special branch for GitHub hosted project web pages. The big advantage is that we have more control over the build environment, and we're independent from RTD hosting-wise.

A testing version of the docs can be admired [here](https://odlgroup.github.io/odl/index.html). Travis and `travis-sphinx` are configured in such a way that the documentation is built and deployed only when the master branch is pushed to.

If we agree that this is a better solution, then there is one TODO:

- [x] Update all links from RTD to gh-pages.


doc/source/conf.py
```diff
@@ -172,9 +172,9 @@ def setup(app):
 # built documents.
 #
 # The short X.Y version.
-version = 'beta'
+version = '0.4'
```
kohr-h: Comments are misleading, this should be `0.4.0` after all, and `release` would most of the time be identical.

kohr-h: Note that this is one more place to update the version, not that crucial however since it can be fixed afterwards.

adler-j: This should be fixed by simply getting `odl.__version__`, also related to #506.

adler-j:
gh pages is certainly a decent alternative, but I think we need to consider this quite well before we make a decisive change.  Mostly because links will be dead after the change. We need a in person discussion on it.

doc/source/conf.py
```diff
@@ -105,14 +105,14 @@ def __getattr__(cls, name):
...
     # TODO: fix this once RTD updates their intersphinx version
-    extensions += ['sphinx.ext.intersphinx']
+    extensions.append('sphinx.ext.intersphinx')
 
     # Intersphinx to get numpy targets
     intersphinx_mapping = {
-        'python': ('http://python.readthedocs.org/en/latest/', None),
-        'numpy': ('http://numpy.readthedocs.org/en/latest/', None),
+        'python': ('http://python.readthedocs.io/en/latest/', None),
+        'numpy': ('http://numpy.readthedocs.io/en/latest/', None),
```
adler-j: This is dead since several months, we should link the public one, same goes for scipy.

.travis.yml
```diff
@@ -22,16 +27,32 @@ install:
...
-# command to run tests
-script: 
+
+script:
+  # Run tests, including PEP8 check, and produce a coverage report
   - py.test --doctest-modules --cov --cov-report term-missing --pep8
-  
+  # Build the Sphinx doc
+  - cd $TRAVIS_BUILD_DIR/doc/source && python generate_doc.py && cd -
+  - travis-sphinx -n -s $TRAVIS_BUILD_DIR/doc/source build
```
adler-j: This would need to be `make build` to invoke the script, no?

kohr-h: What is "This"?
Which script?

Anyway, I'm not sure what exactly the `travis-sphinx` command does other than `make html`, but if so, this is the more robust variant.

doc/source/conf.py
kohr-h: Not sure what you mean, the `object.inv` files are available there. If there's a better, more reliable source, ok.

doc/source/conf.py
kohr-h: Missed the fenced `odl` import, sure.

kohr-h:
> gh pages is certainly a decent alternative, but I think we need to consider this quite well before we make a decisive change.

Sure, that's the purpose of the PR :-)


> Mostly because links will be dead after the change.

Well, it requires us to update all hyperlinks, but I don't think it''s such a big deal.



> We need a in person discussion on it.

Sure.

doc/source/conf.py
adler-j: Numpy does not seem to support RTD anymore so any link there is going to become outdated, I suppose their official doc is where we should point?

adler-j:
> Well, it requires us to update all hyperlinks, but I don't think it''s such a big deal.

Any external hyperlinks going our way, so our docs will dive in search motors as well. Googling "odl function space"  the top hits are basically all about ODL, Any change means we'll be behind for some time until it catches up. At this time that barely matters but any change now should be somewhat final so that we dont get in trouble later with google pointing to the wrong version of the docs.

kohr-h:
The branch is now up to date again, and if we agree that we go this route, I suggest we merge it rather soon. In particular before #580 goes in, since there were a bunch of broken links, and it's easier to resolve the conflicts than to search again for those links.

We may first want to push a bogus version to master to make RTD build an empty last version, before we merge the branch.

adler-j:
I would be fine with this for now. We are sure that github pages can be uploaded from local as well, if we should chose to change at some point?

kohr-h:
> We are sure that github pages can be uploaded from local as well, if we should chose to change at some point?

I'm pretty sure. It's just another branch in the repo (`gh-pages`), so it can be pushed to from wherever you have push access otherwise.

adler-j:
I say go ahead then!

adler-j:
The new build looks wrong. For example, it has:

https://odlgroup.github.io/odl/generated/odl.discr.lp_discr.DiscreteLpVector.html

Or is this because it has not built yet?

kohr-h:
> Or is this because it has not built yet?

It's still the old one, hang on :-)

adler-j:
Where do I go to find updates, travis?

kohr-h:
Uploaded now:

https://odlgroup.github.io/odl/generated/odl.discr.lp_discr.DiscreteLpElement.html

kohr-h:
> Where do I go to find updates, travis?

What do you mean?

adler-j:
Well is there somewhere I can see the progress of the build of the docs?

kohr-h:
Travis builds the docs, so you can go to the Python 3.5 builds of the master branch. For example [here](https://travis-ci.org/odlgroup/odl/jobs/161561389) for the last build.

PR #537, opened 2016-08-29
API: remove Interval, Rectangle and Cuboid, closes #473

### LINKED ISSUE #473: Remove Interval, Rectangle and Cuboid
These are hardly ever used, follow the wrong naming convention (they are not actual classes) and are not dimension "invariant".

I feel it would be better to simply change all uses of these to `IntervalProd` and be done with it.

adler-j:


kohr-h:
No complaints, go ahead!

adler-j:
Rebase and merge otw.

PR #540, opened 2016-08-29
MAINT: remove python2/3 compatibility code from examples, closes #501

### LINKED ISSUE #501: Python 2/3 compability in examples
We currently have lots of examples where we do the standardized

```
```

in the beginning. I personally feel that this is not good to do in general, mostly because it makes the examples technical and harder to read, as well as forcing users to scroll unnecessarily much to get to the point. My suggestion is to write rewrite the examples such that this is not necessary (where, if anywhere, this is needed) and then simply remove these lines.

adler-j:
Someone (@kohr-h) needs to run `py.test --examples` using python 3 to verify that this works as expected.

### DIFF
--- a/examples/space/simple_r.py
+++ b/examples/space/simple_r.py
@@ -17,11 +17,6 @@
 
 """An example of a very simple space, the real numbers."""
 
-# Imports for common Python 2/3 codebase
-from __future__ import print_function, division, absolute_import
-from future import standard_library
-standard_library.install_aliases()
-from builtins import super
 
 import odl
 
@@ -30,7 +25,7 @@ class Reals(odl.LinearSpace):
     """The real numbers."""
 
     def __init__(self):
-        super().__init__(odl.RealNumbers())
+        odl.LinearSpace.__init__(self, field=odl.RealNumbers())
 
     def _inner(self, x1, x2):
         return x1.__val__ * x2.__val__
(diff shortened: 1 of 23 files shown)


examples/space/simple_r.py
```diff
@@ -45,32 +40,33 @@ def __eq__(self, other):
...
+        return RealNumber(self, value)
 
-    class Vector(odl.LinearSpaceVector):
-        """Real vectors are floats
-        """
 
-        __val__ = None
+class RealNumber(odl.LinearSpaceVector):
+    """Real vectors are floats
+    """
```
kohr-h: Dot and `"""` on the same line

examples/space/simple_rn.py
```diff
@@ -38,7 +29,7 @@ class SimpleRn(odl.space.base_ntuples.FnBase):
     """The real space R^n, non-optimized implmentation."""
 
     def __init__(self, size):
-        super().__init__(size, np.float)
+        odl.space.base_ntuples.FnBase.__init__(self, size, np.float)
```
kohr-h: Perhaps here a `from odl.space.base_ntuples import FnBase` would be more readable.

adler-j:
Did you run the examples?

kohr-h:
Everything works fine except the ones involving proximal and STIR, which I haven't installed. 
All good except for some minor stuff.

adler-j:
merge after CI

PR #547, opened 2016-08-30
Refactor Fourier  and Wavelet modules, make Wavelet n-dimensional

kohr-h:
This un-clutters the `fourier.py` module and puts all the `pyFFTW`-specific stuff in a separate `backends` subpackage, plus the utilities like `reciprocal_space` into a `utils` subpackage. Basically no code changed.

TODOs:

- [x] Improve test coverage. Was probably quite bad before, but now it has even become worse. Needs some parametrization and broader test cases.
- [x] Move remaining `pywt` specific code out of `wavelet.py`. Not that much left, in one or two places maybe.
- [x] Rename `reciprocal` -> `reciprocal_grid` in the Fourier part
- [x] Refactor Fourier tests and parametrize more (could be even more, but anyway)
- [x] Refactor Wavelet tests and parametrize more
- [x] Test wavelet code with more data types (in particular complex)
- [x] Add test comparing with hand-calculated result for wavelet trafo
- [x] Update padding mode names in Pywavelets

Edit: TODOs moved here


odl/trafos/backends/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Bindings to the pyFFTW library."""
```
kohr-h: Wrong text

adler-j:
Doing gods work! I'll review.

kohr-h:
> I'll review.

You can focus on the file structure and the docstrings in the new files, perhaps imports. The rest really is just shuffling code around.

odl/trafos/util/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utility functions for the Fourier transform."""
```
kohr-h: Also wrong text, oops

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utility functions for regular Fourier transforms."""
```
kohr-h: Perhaps "Fourier transforms on regular grids"

kohr-h:
I could actually do the same for `PyWavelets`

adler-j:
It should be done

kohr-h:
I'm already doing it :-)

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    TYPE_MAP_R2C,
+    normalized_scalar_param_list)
+
+
+__all__ = ('reciprocal', 'inverse_reciprocal',
+           'reciprocal_space',
+           'dft_preprocess_data', 'dft_postprocess_data')
+
+
+def reciprocal(grid, shift=True, axes=None, halfcomplex=False):
```
adler-j: Could use being called `reciprocal_grid` now that it's not as obvious what it does.

adler-j:
Looks good, like the initiative.

kohr-h:
Wavelet stuff refactored. I actually picked code from the other wavelet branch, otherwise it would result in tedious rebasing due to moved stuff and changed contents. Apart from that, I also put the common code of `WaveletTransform` and `WaveletTransformInverse` into a common base class, just like for the Fourier transform. Some more refactoring along the way.

TODOs:

- [ ] Improve test coverage. Was probably quite bad before, but now it has even become worse. Needs some parametrization and broader test cases.
- [ ] Move remaining `pywt` specific code out of `wavelet.py`. Not that much left, in one or two places maybe.
- [ ] Rename `reciprocal` -> `reciprocal_grid` in the Fourier part

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+           'PYWT_SUPPORTED_PAD_MODES', 'PYWAVELETS_AVAILABLE')
+
+
+PAD_MODES_ODL2PYWT = {'constant': 'zpd',
+                      'periodic': 'ppd',
+                      'symmetric': 'sym',
+                      'order0': 'cpd',
+                      'order1': 'sp1',
+                      'pywt_periodic': 'per'}
+PAD_MODES_PYWT2ODL = {v: k for k, v in PAD_MODES_ODL2PYWT.items()}
```
kohr-h: Not needed so far, could be removed.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+           'PAD_MODES_ODL2PYWT', 'PAD_MODES_ODL2PYWT',
+           'PYWT_SUPPORTED_PAD_MODES', 'PYWAVELETS_AVAILABLE')
+
+
+PAD_MODES_ODL2PYWT = {'constant': 'zpd',
+                      'periodic': 'ppd',
+                      'symmetric': 'sym',
+                      'order0': 'cpd',
+                      'order1': 'sp1',
+                      'pywt_periodic': 'per'}
```
kohr-h: I removed the function and made it a mapping. The check in the case of constant padding happens in the operator now.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+
+        'sym' : Symlets
+
+        'coif': Coiflets
+
+        'bior': Biorthogonal
+
+        'rbio': Reverse biorthogonal
+
+        'dmey': Discrete FIR approximation of the Meyer wavelet
```
kohr-h: Perhaps this listing could be removed altogether - users can simply click on the link. Now it also doesn't mention that you have to append a certain number to the mentioned strings.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+
+    nscales : int
+        Number of scaling levels to be used in the transform.
+        The maximum number of usable scales can be determined
+        by `pywt.dwt_max_level`.
+    pad_mode : string, optional
```
kohr-h: not optional

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+        the smallest possible number of decomposition coefficients.
+        Only available with ``impl='pywt'``, See `pywt.MODES.modes`.
+
+    Returns
+    -------
+    shape_list : list
+        List containing the shapes of the wavelet (approximation
+        and detail) coefficients at different scaling levels.
+
+        ``shape_list[0]`` = shape of approximation coefficients at
```
kohr-h: the

odl/trafos/backends/pywt_bindings.py
adler-j: Not quite sure If i like this. In the future, we may use a backend that handles `constant != 0` and it will possibly be hard with this. Then again, changing at that time would be trivial.

odl/trafos/backends/pywt_bindings.py
adler-j: I am strongly in favor of keeping this, mostly because "click the link" is not possible from a IDE.

With that said, this list is lacking some things for example `'db` is not a valid parameter, users need to give `dbN` for "some" `N`.

adler-j:
Suggestion: Scrap all wavelet changes from the other branch and only cherrypick the example.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+        Dimensions along which to take the transform. ``None`` means
+        using all axis and is equivalent to ``np.arange(ndim)``.
+    halfcomplex : bool, optional
+        If ``True``, calculate only the negative frequency part along the
+        last axis. If ``False``, calculate the full complex FFT.
+        This option can only be used with real input data.
+
+    Other Parameters
+    ----------------
+    fftw_plan : ``pyfftw.FFTW``, optional
```
adler-j: Add `pyfftw.FFTW` to intersphinx and remove the double ticks.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+            plan_arr_in, array_out, direction=_local_to_pyfftw(direction),
+            flags=flags, planning_timelimit=planning_timelimit,
+            threads=threads, axes=axes)
+    else:
+        fftw_plan = fftw_plan_in
+
+    fftw_plan(array_in, array_out, normalise_idft=normalise_idft)
+
+    if wexport:
+        try:
```
adler-j: Is there no better way to do this?

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+            with open(wexport, 'ab') as wfile:
+                pickle.dump(pyfftw.export_wisdom(), wfile)
+        except TypeError:  # Got file handle
+            pickle.dump(pyfftw.export_wisdom(), wexport)
+
+    return fftw_plan
+
+
+def _pyfftw_to_local(flag):
```
adler-j: docstring, also what is local?

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+    return flag.lstrip('FFTW_').lower()
+
+
+def _local_to_pyfftw(flag):
+    return 'FFTW_' + flag.upper()
+
+
+def _pyfftw_destroys_input(flags, direction, halfcomplex, ndim):
+    """Return ``True`` if FFTW destroys an input array, ``False`` otherwise."""
```
adler-j: Destroy isnt really well defined. Is the pointer freed, or is the data scrambled?

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+        else:
+            in_dtype = arr_out.dtype
+
+        if arr_in.dtype != in_dtype:
+            raise ValueError('expected input dtype {}, got {}'
+                             ''.format(dtype_repr(in_dtype),
+                                       dtype_repr(arr_in.dtype)))
+
+    else:  # Shouldn't happen
+        raise RuntimeError
```
adler-j: This raises the class RunTimeError, not an instance of it (if I'm not wrong?)

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+    aaa = wcoeffs['aaa']
+    aad = wcoeffs['aad']
+    ada = wcoeffs['ada']
+    add = wcoeffs['add']
+    daa = wcoeffs['daa']
+    dad = wcoeffs['dad']
+    dda = wcoeffs['dda']
+    ddd = wcoeffs['ddd']
+
+    details = (aad, ada, add, daa, dad, dda, ddd)
```
adler-j: This should be rewritten

```python
details = [wcoeffs[coeff] for coeff in ['aad', 'ada', 'add', 'daa', 'dad', 'dda', 'ddd']]
```

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+
+        The abbreviations are "a" for "approximation" and "d" for "detail".
+        The letter position encodes the coefficient type in the respective
+        axis. ``N`` stands for the number of scaling levels.
+    """
+    wbasis = pywt_wbasis(wbasis)
+    coeff_list = []
+
+    wcoeffs = pywt.dwtn(arr, wbasis, pywt_mode)
+    aaa = wcoeffs['aaa']
```
adler-j: This is sometimes overwritten (uclear to me how). Should be more clear.

odl/trafos/backends/pywt_bindings.py
adler-j: This can be done in many places it seems.

adler-j:
Very hard to review, I'll have to trust you mostly moved code around :) Agree that we need more tests for wavelets.

odl/trafos/backends/pywt_bindings.py
kohr-h: Probably we should have a helper for all dimensions and use `wavedecn` and `waverecn` in all places. A dimension-independent variant to generate the dictionary keys is
```python
from itertools import product
details = [wcoeffs[''.join(p)] for p in product('ad', repeat=ndim)]
```

kohr-h:
> Very hard to review, I'll have to trust you mostly moved code around :)

Well, for the wavelet part, there are substantial changes, so there a thorough look is more important. Functionality is as before, though, and tests are running.

odl/trafos/backends/pywt_bindings.py
adler-j: Are you sure that generates them in the correct order?

kohr-h: Yes, because this is the way they are generated [in `PyWavelets`](https://github.com/PyWavelets/pywt/blob/5fc0f769dfdd6d21c2d0b2fa4fa8cdb7e887bb9f/pywt/_multidim.py#L267).

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+    """Convert ``wbasis`` a to `pywt.Wavelet` if it is not already one."""
+    if isinstance(wbasis, pywt.Wavelet):
+        return wbasis
+    else:
+        return pywt.Wavelet(wbasis)
+
+
+def pywt_max_level(*args, **kwargs):
+    """Return maximum number of scaling levels."""
+    if not PYWAVELETS_AVAILABLE:
```
adler-j: Disagree with adding these, methods prefixed `pywt_` throwing an error if `pywt` does not exist is rather straightforward.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+
+    # Add a duplicate of the last entry for the approximation coefficients
+    shape_list.append(shape_list[-1])
+
+    # We created the list in reversed order compared to what pywt expects
+    shape_list.reverse()
+    return shape_list
+
+
+def pywt_coeff_to_array(coeff, shape_list):
```
adler-j: This needs a reference on why the f it is in this order and not some other order. Simply saying "this is how pywt orders them" should be fine.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+    flat_sizes = [np.prod(shp) for shp in shape_list[:-1]]
+    ndim = len(shape_list[0])
+    dcoeffs_per_scale = 2 ** ndim - 1
+
+    flat_total_size = flat_sizes[0] + dcoeffs_per_scale * sum(flat_sizes[1:])
+    flat_coeff = np.empty(flat_total_size)
+
+    start = 0
+    stop = flat_sizes[0]
+    flat_coeff[start:stop] = coeff[0].ravel()
```
adler-j: coeff is a sequence but we assume it has `ravel`?

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+        be transformed and on the chosen wavelet basis.
+    """
+    arr = np.asarray(arr)
+    wbasis = pywt_wbasis(wbasis)
+    pywt_mode = PAD_MODES_ODL2PYWT[pad_mode]
+    if arr.ndim == 1:
+        coeff_list = pywt.wavedec(arr, wbasis, pywt_mode, nscales)
+    elif arr.ndim == 2:
+        coeff_list = pywt.wavedec2(arr, wbasis, pywt_mode, nscales)
+    elif arr.ndim == 3:
```
adler-j: This really should be changed to a `arr.ndim >= 3` case.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+        coeff_list = _wavedec3(arr, wbasis, pywt_mode, nscales)
+    else:
+        raise NotImplementedError('no transform available for {} dimensions'
+                                  ''.format(arr.ndim))
+    return pywt_coeff_to_array(coeff_list, shape_list)
+
+
+def pywt_wavelet_recon(coeff, wbasis, pad_mode, shape_list):
+    """Return the discrete wavelet reconstruction from ``coeff``.
```
adler-j: ~~Return the ~~

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+    Returns
+    -------
+    rec : `numpy.ndarray`
+        Wavelet reconstruction from the coefficients.
+    """
+    wbasis = pywt_wbasis(wbasis)
+
+    aaa = coeff_list[0]
+    (aad, ada, add, daa, dad, dda, ddd) = coeff_list[1]
+    coeff_dict = {'aaa': aaa, 'aad': aad, 'ada': ada, 'add': add,
```
adler-j: This should be done with dict comprehension, something like:

```python
detail_names = [''.join(p) for p in product('ad', repeat=ndim)]
coeff_dict = {name: coeff for p in zip(detail_names, [coeff_list[0]] + coeff_list[1]))
```

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+                  'daa': daa, 'dad': dad, 'dda': dda, 'ddd': ddd}
+    for tpl in coeff_list[2:]:
+        aaa = pywt.idwtn(coeff_dict, wbasis, pywt_mode)
```
adler-j: I dont quite get this, are we repeatedly calling `idwtn` for a reason?

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+        (aad, ada, add, daa, dad, dda, ddd) = tpl
+        coeff_dict = {'aaa': aaa, 'aad': aad, 'ada': ada, 'add': add,
```
adler-j: This is repeatedly overwritten, I feel a bit wierd about this code, needs docstrings.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+        In 3D:
+
+        ``[aaaN, (aadN, adaN, addN, daaN, dadN, ddaN, dddN), ...
+        (aad1, ada1, add1, daa1, dad1, dda1, ddd1)]``
+
+        The abbreviations are "a" for "approximation" and "d" for "detail".
+        The letter position encodes the coefficient type in the respective
+        axis. ``N`` stands for the number of scaling levels.
+
+    shape_list : `sequence`
```
adler-j: shouldnt this more appropriately be called `shapes` or perhaps `wavelet_shapes` or something?

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+
+    Parameters
+    ----------
+    arr : `array-like`
+        Input array to the wavelet decomposition.
+    wbasis :  string or `pywt.Wavelet`
+        Specification of the wavelet to be used in the transform.
+        If a string is given, it is converted to a `pywt.Wavelet`.
+    pad_mode: string
+        ODL style boundary condition.
```
adler-j: I guess this could use some documentation written about it at some point.

odl/trafos/wavelet.py
```diff
@@ -21,623 +21,255 @@
...
+    pywt_wbasis, pywt_max_level, pywt_wavelet_decomp,
+    pywt_wavelet_recon, pywt_coeff_shape_list, PYWT_SUPPORTED_PAD_MODES,
+    PYWAVELETS_AVAILABLE)
 
+__all__ = ('WaveletTransform', 'WaveletTransformInverse')
 
-__all__ = ('WaveletTransform', 'WaveletTransformInverse',
-           'PYWAVELETS_AVAILABLE')
 
+_SUPPORTED_WAVELET_IMPL = ()
```
adler-j: is it `IMPL` or `IMPLS` in this? Check what the others do.

adler-j: Seems to be `IMPLS`, so follow that.

odl/trafos/wavelet.py
```diff
@@ -21,623 +21,255 @@
...
-    @property
-    def is_orthogonal(self):
-        """Whether or not the wavelet basis is orthogonal."""
-        return self.wbasis.orthogonal
-
-    @property
-    def is_biorthogonal(self):
-        """Whether or not the wavelet basis is bi-orthogonal."""
-        return self.wbasis.biorthogonal
+        super().__init__(domain, wbasis, nscales, 'forward', pad_mode,
```
adler-j: I prefer named parameters in these cases since it becomes hard to follow.

odl/util/testutils.py
```diff
@@ -217,13 +217,13 @@ def is_subdict(subdict, dictionary):
     )
 
     skip_if_no_pywavelets = pytest.mark.skipif(
-        "not odl.trafos.wavelet.PYWAVELETS_AVAILABLE",
-        reason='Wavelet not available'
+        "not odl.trafos.PYWAVELETS_AVAILABLE",
```
adler-j: Writing this here but it is more general.

I feel that we mix `PYWAVELETS`, `PyWavelets` and `pywt` in a rather unorganized fashion. Is it obvious for users that `PYWAVELETS_AVAILABLE` correlates to `pywt` being installed? Perhaps we should rename this one to `PYWT_AVAILABLE`?

odl/trafos/backends/pywt_bindings.py
kohr-h: Then we need to say instead that the `nscales` thing is only for the `pywt` backend. Or at least write a TODO to make a generic helper if we get a different `impl`.

adler-j: Write a TODO then i.m.o. but I knpow too little about wavelets to know if it is intrinsic or not.

kohr-h: I think this is pretty much independent of implementation since in wavelet decomposition, one computes a downsampled-by-factor-2 convolution of input data with a filter, i.e. the convolution adds a little bit depending on the filter length, and then you halve the size. What one ends up with is `log2(input_len / (filter_len - 1))` as the "sensible" number of levels one can have. After that, the convolved data will have always the same size (something like `2 * (filter_len - 1)`), so there is no more size reduction. After that, one could, of course, just happily go on, but it doesn't make much sense.
Different implementations may or may not allow for this indefinite filtering, but the maximum number of reasonable levels is always the same.
Honestly, one could simply implement the function by hand and ditch the `pywt` helper:
```python
>>> import pywt
>>> for i in range(1000):
...     for f in range(1, 31):
...         if f <= 1 or i < f - 1:
...             n = 0
...         else:
...             n = int(np.log2(i / (f - 1)))
...         m = pywt.dwt_max_level(i, f)
...         if n != m:
...             print('mismatch for i={}, f={}'.format(i, f))

```
No printout.
The above basically replicates [this function](https://github.com/PyWavelets/pywt/blob/master/pywt/_extensions/c/common.c#L70-L75). Seems a bit overkill to implement it in C.

odl/trafos/backends/pywt_bindings.py
kohr-h: We implicitly assume its members to be arrays. If we expect at any point to get array-like members,  we need to use `np.ravel` or `np.asarray(x).ravel()`.

odl/trafos/backends/pywt_bindings.py
kohr-h: why?

odl/trafos/backends/pywt_bindings.py
kohr-h: I'll make a generic implementation. This has always bugged me.

odl/trafos/backends/pywt_bindings.py
kohr-h: `pywt.idwtn` handles only one scaling level at a time, so we have to call it repeatedly.

odl/trafos/backends/pywt_bindings.py
kohr-h: Probably a recursive function would make more sense.

odl/trafos/backends/pywt_bindings.py
kohr-h: Yeah definitely. It was `size_list` before, which I found misleading.

odl/trafos/backends/pywt_bindings.py
adler-j: Well doc says nothing about them being arrays, so I would go for nonmember `ravel`.

kohr-h: In a broader sense, these backend functions hang a little bit in no man's land between public and private. They're kind of private since they're only used internally in the corresponding operators, but we don't prepend an underscore since they anyway feel "potentially useful". 

We should perhaps settle for one interpretation: public if useful outside, private if "too low-level", and in the latter case we prepend an underscore and make more assumptions on the input. In the former case, better documentation and more robust input handling.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,523 @@
...
+
+    flat_total_size = flat_sizes[0] + dcoeffs_per_scale * sum(flat_sizes[1:])
+    flat_coeff = np.empty(flat_total_size)
+
+    start = 0
+    stop = flat_sizes[0]
+    flat_coeff[start:stop] = coeff[0].ravel()
+
+    for fsize, detail_coeffs in zip(flat_sizes[1:], coeff[1:]):
+        if dcoeffs_per_scale == 1:
```
adler-j: This if else could be solved with a nonmember ravel :)

This (should) work:

```python
for fsize, detail_coeffs in zip(flat_sizes, coeff):
    start, stop = stop, stop + fsize
    flat_coeff[start:stop] = np.ravel(detail_coeffs)
```

odl/trafos/backends/pywt_bindings.py
adler-j: I was under the expression that we were trying to avoid "Return the" style docstrings?

odl/trafos/backends/pywt_bindings.py
kohr-h: I admit the construction is a bit brittle. In some way there is the implicit assumption that the code is called in the correct way and that calling code does proper checks. It's again a bit of a scope problem. If we want these functions to be considered public, they need to be more robust, if not it's probably okay to make assumptions like this.

odl/trafos/backends/pywt_bindings.py
kohr-h: That's new to me. My understanding was that we don't use it for `@property` since they look exactly like attributes, and attributes *are* something, in contrast to methods or functions, which *do* something (return, modify, check, ...).

odl/trafos/backends/pywt_bindings.py
kohr-h: I would like to mention that it is advisable for us to keep the one-level-at-a-time kind of implementation since we want to manipulate intermediate results when calculating the adjoint with proper boundary behavior (see issue #400 and PR #421).

adler-j:
How close to working is this atm? We should get it in rather soon since the merge is getting worse and worse.

kohr-h:
I'm working on it right now. The code should be generalized to arbitrary dimension and have a functionality for "one level at a time" since we need to hook in there for the adjoint.

odl/trafos/backends/pywt_bindings.py
adler-j: Well in general it is fine, but in many cases like here, I feel it is simply empty words. And these one setence introductions are the most important part of the doc and what people will be skimming the most.

odl/trafos/backends/pyfftw_bindings.py
kohr-h: How's that work?

odl/trafos/backends/pyfftw_bindings.py
kohr-h: Probably yes, TODO, new issue. No idea now.

odl/trafos/backends/pyfftw_bindings.py
kohr-h: `odl` I guess

odl/trafos/backends/pyfftw_bindings.py
kohr-h: could be expressed as "overwrite"

odl/trafos/backends/pyfftw_bindings.py
kohr-h: AFAIK the string in parentheses is optional, so with this you get the same exception without message. It's handled in an exception-specific way so you always get an instance.
```python
>>> e = RuntimeError
>>> print(e.args)
<attribute 'args' of 'BaseException' objects>
>>> try:
>>>     raise RuntimeError
>>> except RuntimeError as e:
>>>     print(e.args)
()
```

odl/trafos/backends/pywt_bindings.py
kohr-h: I kept this dictionary but revived the function, which in turn uses the dictionary and raises for the invalid cases. The dictionary is useful for quick extensibility.

kohr-h:
Got some serious work done on this PR. Not 100 % there yet coverage-wise but well on the way. 

Suggestions for adequate tests of the operator without manually calculating transforms are very welcome. Apart from that, there are mostly the exceptions which need proper testing.

So this is ready for a new review. No new work went into the Fourier stuff, while the wavelet stuff is essentially all new code, so it can be reviewed as addition.

odl/trafos/backends/pyfftw_bindings.py
adler-j: `conf.py` has a list of intersphinx linkings, you simply add `pyfftw` and a proper URL to it and then links should work.

odl/trafos/backends/pyfftw_bindings.py
adler-j: Agree, leave for now, make a issue and add TODO here

odl/trafos/backends/pyfftw_bindings.py
adler-j: Seems you are right, still I'd add something short, otherwise I prefer `assert False`  as a way of stating "you should never get here".

kohr-h:
I added an issue #614 regarding the bad coverage on Travis.

adler-j:
> Suggestions for adequate tests of the operator without manually calculating transforms are very welcome. 

Very hard to perform. We can ofc test some well known properties of the operator (how does it affect norms etc). Perhaps you can do "scale tests" by testing some "bumb" functions that get increasingly smaller. You can also verify that the mean value is correct. Other than that we'll mostly have to live with "test by review" and make sure that the tests start failing if the code behavior changes.

> Apart from that, there are mostly the exceptions which need proper testing.

Hopefully shouldnt be too many, you can weave this into the others in some cases to reduce the amount of test code needed.

kohr-h:
> Very hard to perform. We can ofc test some well known properties of the operator (how does it affect norms etc). Perhaps you can do "scale tests" by testing some "bumb" functions that get increasingly smaller. You can also verify that the mean value is correct. Other than that we'll mostly have to live with "test by review" and make sure that the tests start failing if the code behavior changes.

Some good ideas there. On the upside, the helper functions for the work under the hood are now tested much more thoroughly with random input and all that. I'll think about your suggestions.

> Hopefully shouldnt be too many, you can weave this into the others in some cases to reduce the amount of test code needed.

There's not that much, no. Mostly standard stuff regarding input, no sneaky corner cases.

odl/trafos/backends/pyfftw_bindings.py
kohr-h: I'll try it, but for the given examples, `intersphinx` downloads a very specific `objects.inv` file from the given URI, and I'm not sure `pyfftw` has such a thing. It's very hard to google also.

adler-j:
Ok putting some thought into this testing thing:

* Test dtypes (specifically pywt supports complex, so should we)
* Obviously simply testing that the wavelet transform and then inverse is the identity operator is something we can always do.
* We expect `n_1 x n_2 x ... x n_k` wavelet transforms to equal `n_1 x n_2 x ... x n_k x 1` etc. Also insert dimension in middle (`n x m` vs `n x 1 x m`)
* Verify linearity, both scaling and sum
* Test locality. Change only parts of the incomming data (in a spatially connected region) and verify that at least some of the output stays exactly the same. (this would by comparsion fail for the Fourier transform)
* Test different choices for the wavelet type and check that the results are actually different.
* Test symmetry. If you change two axes in the input ("transpose the array"), the result should have some kind of symmetry. Same goes for reflections.

I guess you could add more, but these are at least some of them. Some may be to complex, some may fit better in the largescale tests.

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+        raise ValueError("`sign` '{}' not understood".format(sign))
+
+    def _onedim_arr(length, shift):
+        if shift:
+            # (-1)^indices
+            indices = np.arange(length)
+            arr = -2 * np.mod(indices, 2) + 1.0
+        else:
+            indices = np.arange(length)
+            arr = np.exp(-imag * np.pi * indices * (1 - 1.0 / length))
```
adler-j: efficiency question. Why not collect the scalars to one:

```python
arr = np.exp((-imag * np.pi  * (1 - 1.0 / length)) * indices)
```

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+        return arr.astype(out.dtype, copy=False)
```
adler-j: Create a arange of the correct dtype from the start instead, then you can avoid this and optimize above:

```
arr = np.arange(length)
np.exp((-imag * np.pi  * (1 - 1.0 / length)) * arr, out=arr)
```

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    elif sign == '+':
+        imag = 1j
+    else:
+        raise ValueError("`sign` '{}' not understood".format(sign))
+
+    def _onedim_arr(length, shift):
+        if shift:
+            # (-1)^indices
+            indices = np.arange(length)
+            arr = -2 * np.mod(indices, 2) + 1.0
```
adler-j:
```
arr = np.ones(length, dtype=out.dtype)
arr[1::2] = -1
```

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    except TypeError:
+        pass
+
+    onedim_arrs = []
+    for ax, shift, intp in zip(axes, shift_list, interp):
+        x = real_grid.min_pt[ax]
+        xi = recip_grid.coord_vectors[ax]
+
+        # First part: exponential array
+        onedim_arr = (np.exp(imag * x * xi))
```
adler-j: Extra parenthesises

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    else:
+        raise ValueError("`sign` '{}' not understood".format(sign))
+
+    op, op_in = str(op).lower(), op
+    if op not in ('multiply', 'divide'):
+        raise ValueError("kernel `op` '{}' not understood".format(op_in))
+
+    # Make a list from interp if that's not the case already
+    try:
+        interp = [str(interp + '').lower()] * arr.ndim
```
adler-j: split in two lines, this error message will be unreadable for anyone but you

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+        # First part: exponential array
+        onedim_arr = (np.exp(imag * x * xi))
+
+        # Second part: interpolation kernel
+        len_dft = recip_grid.shape[ax]
+        len_orig = real_grid.shape[ax]
+        halfcomplex = (len_dft < len_orig)
+        odd = len_orig % 2
+
+        if shift:
```
adler-j: This if if if construction is a mess to review. I frankly cant tell if it is correct or not. Is there no clearer way?

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+        upper end of the domain (where the half space ends) is chosen to
+        coincide with the grid node.
+    """
+    if not isinstance(space, DiscreteLp):
+        raise TypeError('`space` {!r} is not a `DiscreteLp` instance'
+                        ''.format(space))
+    if not space.is_uniform:
+        raise ValueError('`space` is not uniformly discretized')
+
+    if axes is None:
```
adler-j: This construction happens at least trice, perhaps make function?

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+                         'spaces')
+
+    exponent = kwargs.pop('exponent', None)
+    if exponent is None:
+        exponent = conj_exponent(space.exponent)
+
+    dtype = kwargs.pop('dtype', None)
+    if dtype is None:
+        if is_real_dtype(space.dtype):
+            dtype = TYPE_MAP_R2C[space.dtype]
```
adler-j: This TYPE_MAP_R2C magic should simply be replaces with a utility function `complex_dtype` that works on any dtype, that way you can remove the above if statement as well.

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    # Make a partition with nodes on the boundary in the last transform axis
+    # if `halfcomplex == True`, otherwise a standard partition.
+    if halfcomplex:
+        max_pt = {axes[-1]: recip_grid.max_pt[axes[-1]]}
+        part = uniform_partition_fromgrid(recip_grid, max_pt=max_pt)
+    else:
+        part = uniform_partition_fromgrid(recip_grid)
+
+    recip_spc = uniform_discr_frompartition(part, exponent=exponent,
+                                            dtype=dtype, impl='numpy')
```
adler-j: We never check that the input has impl == 'numpy' so is it really fair to make the output have it? Shouldnt we accept any type of space anyway? At some point we could add CUFFT as a backend for cuda.

odl/trafos/wavelet.py
```diff
@@ -21,623 +21,242 @@
...
-        if dcoeffs_per_scale == 1:
-            start, stop = stop, stop + fsize
-            flat_coeff[start:stop] = detail_coeffs.ravel()
+        impl : {'pywt'}, optional
+            Back-end for the wavelet transform.
+        """
+        if not isinstance(space, DiscreteLp):
+            raise TypeError('`space` {!r} is not a `DiscreteLp` instance.'
+                            ''.format(space))
+        if space.ndim > 3:
```
adler-j: This still holds?

odl/trafos/__init__.py
```diff
@@ -15,13 +15,19 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Function transformations based on ODL."""
+"""Utility functions for transformations."""
```
kohr-h: Seems like a copy & paste error.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+
+    where the summation is taken over all indices
+    ``j = (j[0], ..., j[d-1])`` in the range ``0 <= j < N``
+    (component-wise), with ``N`` being the shape of the input array.
+
+    The output indices ``k`` lie in the same range, except
+    for half-complex transforms, where the last axis ``i`` in ``axes``
+    is shortened to ``0 <= k[i] < floor(N[i]/2) + 1``.
+
+    In the backward transform, sign of the the exponential argument
```
kohr-h: I'm not gonna touch the docstrings in this PR. Its scope is already vastly extended from its original purpose. New issue.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+
+def pywt_wavelet(wavelet):
+    """Convert ``wavelet`` to a ``pywt.Wavelet`` instance."""
+    if isinstance(wavelet, pywt.Wavelet):
+        return wavelet
+    else:
+        return pywt.Wavelet(wavelet)
+
+
+def pywt_mode(pad_mode, pad_const=0):
```
kohr-h: But it's an ODL pad mode.

odl/trafos/backends/pyfftw_bindings.py
kohr-h: See #616.

kohr-h:
> I'm somewhat worried about the fft performance due to the copies we make. How much overhead are we talking about? Have you benchmarked?

No, but I think the copies are somewhat the minimum possible. I can have a look again, but I'm not sure there is very much room for improvement.
There's an issue on that, #617.

odl/trafos/wavelet.py
kohr-h: No, this is actually deprecated. Will be removed.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Bindings to the ``pyFFTW`` backend for Fourier transforms."""
```
kohr-h: OK

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+    """Return a list of coefficient shapes in the specified transform.
+
+    Parameters
+    ----------
+    shape : sequence
+        Shape of an input to the transform.
+    wavelet : string or ``pywt.Wavelet``
+        Specification of the wavelet to be used in the transform.
+        If a string is given, it is converted to a ``pywt.Wavelet``.
+        Use `pywt.wavelist
```
kohr-h: I think intersphinx requires a library to export an `objects.inv` file. I'm not sure PyWavelets does that, googling gives no hits. That doesn't have to mean anything, since googling the same for Numpy also only gives the hits on StackOverflow, not anything on the Numpy page itself.

kohr-h: Apart from that, I'll move the links to a `References` section.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+        i-th level detail coefficients.
+    """
+    shape = tuple(shape)
+    if any(int(s) != s for s in shape):
+        raise ValueError('`shape` may only contain integers, got {}'
+                         ''.format(shape))
+    wavelet = pywt_wavelet(wavelet)
+
+    nlevels, nlevels_in = int(nlevels), nlevels
+    if float(nlevels_in) != nlevels:
```
kohr-h: Yes, it would

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+    -------
+    flat_size : int
+        Size of a flat array containing all approximation and detail
+        coefficients for the specified transform.
+    """
+    shapes = pywt_coeff_shapes(shape, wavelet, nlevels, mode)
+    # 1 x approx coeff and (2**n - 1) * detail coeff
+    ndim = len(shapes[0])
+    return (np.prod(shapes[0]) +
+            sum((2 ** ndim - 1) * np.prod(shape)
```
kohr-h: Ah, now we're talking optimization :-)

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+                for shape in shapes[1:]))
+
+
+def pywt_flat_array_from_coeffs(coeffs):
+    """Return a flat array from a Pywavelets coefficient sequence.
+
+    Related to 1D, 2D and 3D multilevel discrete wavelet transforms.
```
kohr-h: Ah, blast it. Of course deprecated.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+            c.reshape(shape)
+            for c in np.split(arr[start:stop], dcoeffs_per_scale))
+
+        coeff_list.append(detail_coeffs)
+
+    return coeff_list
+
+
+def pywt_dict_keys(length):
+    """Return a list of coefficient dictionary keys as used in ``pywt``."""
```
kohr-h: Will do

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+        try:
+            axes = [int(axes)]
+        except TypeError:
+            axes = list(axes)
+
+    rstride = recip_grid.stride
+    rshape = recip_grid.shape
+
+    # Calculate shape of the output grid by adjusting in axes[-1]
+    irshape = list(rshape)
```
kohr-h: Yes, we are. For some odd reason, `list` has a `copy()` method only on Python 3. This works for both and creates a copy, which I prefer to importing the `copy` module.

kohr-h: The docstring says:
```
list() -> new empty list 
list(iterable) -> new list initialized from iterable’s items
```

odl/trafos/backends/pywt_bindings.py
adler-j: Well the most clear signature is:

```
def pywt_pad_mode(odl_pad_mode, pad_const)
```

Currently `mode` could mean anything. Parallelism mode, error handling mode, etc.

odl/trafos/backends/pywt_bindings.py
adler-j: Find the doc and give it a try locally

odl/trafos/backends/pywt_bindings.py
adler-j: Also readability, I was looking for how it depended on the inner variables and couldn't find it. If it is inside you kinda expect it to be used.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+    array_in : `numpy.ndarray`
+        Array to be transformed
+    array_out : `numpy.ndarray`
+        Output array storing the transformed values, may be aligned
+        with ``array_in``.
+    direction : {'forward', 'backward'}
+        Direction of the transform
+    axes : int or `sequence` of ints, optional
+        Dimensions along which to take the transform. ``None`` means
+        using all axis and is equivalent to ``np.arange(ndim)``.
```
kohr-h: of course

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+        Flag for the amount of effort put into finding an optimal
+        FFTW plan. See the `FFTW doc on planner flags
+        <http://www.fftw.org/fftw3_doc/Planner-Flags.html>`_.
+        Default: 'estimate'
+    planning_timelimit : float or ``None``, optional
+        Limit planning time to roughly this many seconds.
+        Default: ``None`` (no limit)
+    threads : int, optional
+        Number of threads to use.
+        Default: Number of CPUs if the number of data points is larger
```
kohr-h: I'll implement your suggestion below.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,309 @@
...
+        than 1000, else 1.
+    normalise_idft : bool, optional
+        If ``True``, the backward transform is normalized by
```
kohr-h: I'll explain it better in the docstring.

odl/trafos/backends/pywt_bindings.py
kohr-h: Decent suggestion. I chose the "mode" name suffix since that matches the API of PyWavelets. They call the string simply "mode", and I think the names in such a bindings module should be as close as possible to the names in the API that they wrap.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,532 @@
...
+def _check_nlevels(nlevels, data_len, wavelet):
+    """Check if ``nlevels`` is valid for given data length and wavelet."""
+    max_levels = pywt.dwt_max_level(data_len, wavelet.dec_len)
+    if not 0 < nlevels <= max_levels:
+        raise ValueError('`nlevels` must lie between 1 and {}, got {}'
+                         ''.format(max_levels, nlevels))
+
+
+def pywt_coeff_shapes(shape, wavelet, nlevels, mode):
+    """Return a list of coefficient shapes in the specified transform.
```
kohr-h: OK

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    shift_list = normalized_scalar_param_list(shift, length=len(axes),
+                                              param_conv=bool)
+
+    # Make a copy of arr with correct data type if necessary, or copy values.
+    if out is None:
+        if is_real_dtype(arr.dtype) and not all(shift_list):
+            out = np.array(arr, dtype=TYPE_MAP_R2C[arr.dtype], copy=True)
+        else:
+            out = arr.copy()
+    elif out is arr:
```
kohr-h: Yes, this line stems from before my enlightenment in this matter :-)

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+    For a shifted grid, we have :math:``xi[0] =  -pi/s``, thus the
+    array is given by::
+
+        p[k] = (-1)**k
+
+    Parameters
+    ----------
+    arr : `array-like`
+        Array to be pre-processed. If its data type is a real
+        non-floating type, it is converted to 'float64'.
```
kohr-h: No, I don't convert in that case. It only applies for stuff like `int` or `bool`.

odl/trafos/util/ft_utils.py
kohr-h: Yeah this was really some stupid code..

odl/trafos/backends/pywt_bindings.py
kohr-h: Hm, since the function returns the `pywt` padding mode for both styles of modes as input, it would be too strong to call the parameter `odl_pad_mode`. I think I'll stick with just `pad_mode` and change the name of the function to `pywt_pad_mode`.

odl/trafos/backends/pywt_bindings.py
kohr-h: I underestimated sphinx. It seems to always export an `objects.inv` file at the root of the documentation. Works like a charm with both `pywt` and `pyfftw`.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -116,6 +124,27 @@ def pywt_coeff_shapes(shape, wavelet, nlevels, mode):
...
         i-th level detail coefficients.
+
+    See Also
+    --------
+    pywt.dwt_coeff_len : function used to determine coefficient sizes at
+        each scaling level
+
+    Examples
+    --------
+    Determine the coefficient shapes for 2 scaling levels in 3
```
adler-j: in 3 what

odl/trafos/backends/pywt_bindings.py
```diff
@@ -275,6 +338,25 @@ def pywt_coeffs_from_flat_array(arr, shapes):
...
+
+    >>> arr = [1, 1, 2, 2, 3, 3, 4, 4]
+    >>> # approximation and detail coefficient shapes
+    >>> shapes = [(1, 2), (1, 2)]
+    >>>
+    >>> coeffs = pywt_coeffs_from_flat_array(arr, shapes)
+    >>> approx, details = coeffs
+    >>> print(approx)
+    [[1 1]]
+    >>> for detail in details:
```
adler-j: is print(details) not good enough?

odl/trafos/backends/pywt_bindings.py
```diff
@@ -297,7 +379,23 @@ def pywt_coeffs_from_flat_array(arr, shapes):
...
-    """Return a list of coefficient dictionary keys as used in ``pywt``."""
+    """Return a list of coefficient dictionary keys as used in ``pywt``.
+
+    The dictionary keys are composed of the characters ``'a'`` and ``'d'``
+    and stand for "approximation" and "detail" coefficients. The letter
+    i-th position of a key string signalizes which filtering
+    operation the input undergoes in the i-th coordinate axis. For example,
+    ``'ada'`` stands for low-pass filtering (-> approximation) in the
+    first and last axes and high-pass filtering (-> detail) in the
+    middle axis.
```
adler-j: bloody excellent description. This now makes sense to me

odl/trafos/backends/pywt_bindings.py
```diff
@@ -425,19 +614,64 @@ def pywt_multi_level_decomp(arr, wavelet, nlevels, mode):
...
+    Decomposition of a ``(4, 4)`` array using Haar wavelet, zero-padding
+    and 2 scaling levels:
+
+    >>> arr = [[1, 1, 0, 0],
+    ...        [0, 0, 0, 1],
+    ...        [1, 1, 1, 1],
+    ...        [0, 1, 1, 0]]
+    >>>
+    >>> coeffs = pywt_multi_level_decomp(arr, 'haar', 2, 'zpd')
+    >>>
```
adler-j: remove

odl/trafos/backends/pywt_bindings.py
```diff
@@ -425,19 +614,64 @@ def pywt_multi_level_decomp(arr, wavelet, nlevels, mode):
...
+    Examples
+    --------
+    Decomposition of a ``(4, 4)`` array using Haar wavelet, zero-padding
+    and 2 scaling levels:
+
+    >>> arr = [[1, 1, 0, 0],
+    ...        [0, 0, 0, 1],
+    ...        [1, 1, 1, 1],
+    ...        [0, 1, 1, 0]]
+    >>>
```
adler-j: remove

odl/trafos/backends/pywt_bindings.py
```diff
@@ -354,20 +483,64 @@ def pywt_single_level_recon(approx, details, wavelet, mode):
...
+    >>> approx = [[ 1.5, 0.5],
+                  [ 0.5, 0.5]]
+    >>>
+    >>> details = ([[ 0.5, 0.5],
+    ...             [-0.5, 0.5]],
+    ...            [[ 0.5, 0.5],
+    ...             [ 0.5, 0.5]],
+    ...            [[-0.5, 0.5],
+    ...             [-0.5, 0.5]])
+    >>>
```
adler-j: remove

odl/trafos/backends/pywt_bindings.py
```diff
@@ -425,19 +614,64 @@ def pywt_multi_level_decomp(arr, wavelet, nlevels, mode):
...
+    >>> approx2, details2, details1 = coeffs
+    >>> print(approx2)
+    [[ 2.25]]
+    >>>
+    >>> for detail in details2:
+    ...     print(detail)
+    [[ 0.25]]
+    [[-0.75]]
+    [[ 0.25]]
+    >>>
```
adler-j: remove

odl/trafos/backends/pywt_bindings.py
```diff
@@ -425,19 +614,64 @@ def pywt_multi_level_decomp(arr, wavelet, nlevels, mode):
...
+    ...        [1, 1, 1, 1],
+    ...        [0, 1, 1, 0]]
+    >>>
+    >>> coeffs = pywt_multi_level_decomp(arr, 'haar', 2, 'zpd')
+    >>>
+    >>> # "2" -> scaling level 2, "1" -> scaling level 1
+    >>> approx2, details2, details1 = coeffs
+    >>> print(approx2)
+    [[ 2.25]]
+    >>>
```
adler-j: remove

odl/trafos/backends/pywt_bindings.py
```diff
@@ -481,52 +715,100 @@ def pywt_multi_level_recon(coeff_list, recon_shape, wavelet, mode):
...
+    --------
+    Reconstruct the original array from the decomposition in the example
+    in `pywt_multi_level_decomp`:
+
+    >>> # Original array was
+    >>> # [[1, 1, 0, 0],
+    ... #  [0, 0, 0, 1],
+    ... #  [1, 1, 1, 1],
+    ... #  [0, 1, 1, 0]]
+    >>>
```
adler-j: remove

odl/trafos/backends/pywt_bindings.py
kohr-h: I found the output `(array([[2, 2]]), array([[3, 3]]), array([[4, 4]]))` hard to digest in comparison, but that's a matter of taste.

odl/trafos/backends/pywt_bindings.py
kohr-h: Thanks :-)

odl/trafos/backends/pywt_bindings.py
adler-j: I'll have to agree, but I simply felt that it is easier to maintain and this is a internal function anyway so being complicated isn't that bad

odl/trafos/backends/pywt_bindings.py
```diff
@@ -481,52 +715,100 @@ def pywt_multi_level_recon(coeff_list, recon_shape, wavelet, mode):
...
+    >>> details1 = ([[0, -0.5],
+    ...              [-0.5, 0.5]],
+    ...             [[1, -0.5],
+    ...              [0.5, 0.5]],
+    ...             [[0, 0.5],
+    ...              [0.5, -0.5]])
+    >>>
+    >>> coeffs = [approx2, details2, details1]
+    >>> print(pywt_multi_level_recon(coeffs, wavelet='haar', mode='zpd'))
+    [[  1.00000000e+00   1.00000000e+00   3.92523115e-17   0.00000000e+00]
```
kohr-h: Yeah, I'm working on it. I could instead compare to the original with `np.allclose`.

odl/trafos/backends/pywt_bindings.py
kohr-h: Okay, fair enough. I'll change it.

kohr-h: What about [this place](https://github.com/odlgroup/odl/pull/547/files/66fb05044a8a6dccc70f584baed6e0abdaecc39e#diff-a57dd4d9e768f69a8e6f15d547ed456cR438)? The output of the direct print will look like this:
```
(array([[ 0.5,  0.5],
       [-0.5,  0.5]]), array([[ 0.5,  0.5],
       [ 0.5,  0.5]]), array([[-0.5,  0.5],
       [-0.5,  0.5]]))
```
Very ugly IMO.

odl/trafos/util/ft_utils.py
```diff
@@ -0,0 +1,654 @@
...
+        rshape[axes[-1]] = shape[axes[-1]] // 2 + 1
+
+        # - Odd and not shifted or even and shifted -> 0
+        # - Odd and shifted -> - stride / 2
+        # - Even and not shifted -> + stride / 2
+        last_odd = shape[axes[-1]] % 2 == 1
+        last_shifted = shift_list[-1]
+        half_rstride = np.pi / (shape[axes[-1]] * stride[axes[-1]])
+
+        if last_odd:
```
kohr-h: The suggestion is not quite right, it misses the sign difference between the two cases. Need 3 clauses.

odl/trafos/wavelet.py
```diff
@@ -21,623 +21,242 @@
...
+    def is_biorthogonal(self):
+        """Whether or not the wavelet basis is bi-orthogonal."""
+        return self.pywt_wavelet.biorthogonal
 
 
-class WaveletTransform(Operator):
+class WaveletTransform(WaveletTransformBase):
 
-    """Discrete wavelet trafo between discrete L2 spaces."""
+    """Discrete wavelet trafo between discretized Lp spaces."""
```
kohr-h: Actually, the whole operator should be called `DiscreteWaveletTransform` similar to the situation on the Fourier side. That's because this operator is not an approximation of the continuous wavelet transform, which may also have its use cases but isn't yet implemented.

test/trafos/backends/pyfftw_bindings_test.py
```diff
@@ -0,0 +1,422 @@
...
+        return np.random.rand(*shape).astype(dtype)
+    else:
+        return (np.random.rand(*shape).astype(dtype) +
+                1j * np.random.rand(*shape).astype(dtype))
+
+
+# ---- pyfftw_call ---- #
+
+
+def _params_from_dtype(dt):
```
kohr-h: not possible, because it will use the fixture then

kohr-h:
Done with a large portion of the comments. I added 2 TODOs which I feel should be addressed by this PR; everything beyond is better suited for a separate PR.

test/trafos/backends/pywt_bindings_test.py
```diff
@@ -218,5 +220,43 @@ def test_multilevel_decomp_inverts_recon(shape_setup):
...
+    filter_hl = filter_h[:, None] * filter_l[None, :]
+    filter_hh = filter_h[:, None] * filter_h[None, :]
+
+    # Convolve x with 2D filters (implicitly uses zero-padding)
+    conv_ll = convolve(x, filter_ll)
+    conv_lh = convolve(x, filter_lh)
+    conv_hl = convolve(x, filter_hl)
+    conv_hh = convolve(x, filter_hh)
+
+    # Downsampling gives the wavelet coefficients (starting with index 1)
```
kohr-h: That's the choice that PyWavelets took. I think it's rather arbitrary, but I may be wrong and it's prescribed by the theory for discrete wavelet transforms.

odl/util/utility.py
```diff
@@ -258,6 +251,92 @@ def is_complex_floating_dtype(dtype):
...
 
+def real_dtype(dtype, use_default=False, default=None):
+    """Return the real counterpart of ``dtype`` if existing.
+
+    Parameters
+    ----------
+    dtype :
+        Real or complex floating point data type. It can be given in any
+        way the `numpy.dtype` constructor understands.
+    use_default : bool, optional
```
adler-j: Why is this needed. Wouldnt `default=None` imply that a error should be raised and `default != None` mean that it should be used as a default?

test/trafos/backends/pyfftw_bindings_test.py
adler-j: Why would it do that? I'm quite sure it wouldnt since this is not a fixture nor a test, no?

test/trafos/backends/pyfftw_bindings_test.py
```diff
@@ -0,0 +1,413 @@
...
+    shape[axes[-1]] = shape[axes[-1]] // 2 + 1
+    return shape
+
+
+dtype_params = np.sctypes['float'] + np.sctypes['complex']
+dtype_ids = [' dtype = {} '.format(dtype_repr(dt)) for dt in dtype_params]
+
+
+@pytest.fixture(scope="module", ids=dtype_ids, params=dtype_params)
+def dtype(request):
```
adler-j: Make this a "global" fixture in `conftest.py`

test/trafos/backends/pyfftw_bindings_test.py
```diff
@@ -0,0 +1,413 @@
...
+
+        assert all_almost_equal(idft_arr, true_idft)
+
+
+plan_params = ['estimate', 'measure', 'patient', 'exhaustive']
+plan_ids = [" planning = '{}' ".format(p) for p in plan_params]
+
+
+@pytest.fixture(scope="module", ids=plan_ids, params=plan_params)
+def planning(request):
```
adler-j: Move fixtures to top

test/trafos/util/ft_utils_test.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+
+from future import standard_library
+standard_library.install_aliases()
+
+from itertools import product
+from math import pi
```
adler-j: Use np.pi instead

odl/util/utility.py
kohr-h: That would mean that the calling code would need to handle that case and raise the error, which makes it more complex. My thought was to put that complexity here, but I'm also fine with the alternative.

odl/util/testutils.py
```diff
@@ -284,7 +284,7 @@ def noise_array(space):
     """
     from odl.space import ProductSpace
     if isinstance(space, ProductSpace):
-        return np.array([noise_array(si) for si in space])
+        return np.array([noise_array(si) for si in space]).reshape(space.shape)
```
adler-j: I dont get it, this should fail, no? `space.shape` is usually a small number for product spaces but the array is very big.

kohr-h: I was too lazy to test myself, pushed to let travis test ;-)

kohr-h: Failed of course, reverted the commit.

kohr-h:
Okay, so now all comments except the ones with a new issue are taken care of. I also found another bug in the list-array conversions that was hidden by inappropriate test arrays. Fixed now by better parametrization.

Time for another review (sorry!). I can recommend [this view](https://github.com/odlgroup/odl/compare/56cc4de9c62e68d6c2cafb4fecb4dade68e571cb...refactor_fourier) for the changes since [the last "almost done" comment by me](https://github.com/odlgroup/odl/pull/547#issuecomment-249909705). For a more comprehensive view since around the last big review, check out [this view](https://github.com/odlgroup/odl/compare/e62d9b945cfd50a81a5dd38d53935966553fc53c...refactor_fourier).

odl/trafos/wavelet.py
```diff
@@ -21,656 +21,260 @@
...
-        ``N`` refers to the number of scaling levels
+            ``'dmey'``: Discrete FIR approximation of the Meyer wavelet
 
-    size_list : list
-        A list containing the sizes of the wavelet (approximation
-        and detail) coefficients at different scaling levels.
+        nlevels : positive int
+            Number of scaling levels to be used in the decomposition. The
+            maximum number of levels can be calculated with
+            `pywt.dwt_max_level
```
adler-j: Just got for `pywt.dwt_max_level`

odl/trafos/wavelet.py
```diff
@@ -21,656 +21,260 @@
...
-
-        wbasis :  {string, ``pywt.Wavelet``}
-            If a string is given, converts to a ``pywt.Wavelet``.
-            Describes properties of a selected wavelet basis.
-            See PyWavelet `documentation
-            <http://www.pybytes.com/pywavelets/ref/wavelets.html>`_
+        wavelet : string or ``pywt.Wavelet``
+            Specification of the wavelet to be used in the transform.
+            If a string is given, it is converted to a ``pywt.Wavelet``.
+            Use `pywt.wavelist
```
adler-j: `pywt.wavelist`

odl/trafos/wavelet.py
```diff
@@ -21,656 +21,260 @@
...
-            <http://www.pybytes.com/pywavelets/ref/\
-dwt-discrete-wavelet-transform.html#maximum-decomposition-level\
--dwt-max-level>`_ .
-
-        wbasis :  {string, ``pywt.Wavelet``}
-            If a string is given, converts to a ``pywt.Wavelet``.
-            Describes properties of a selected wavelet basis.
-            See PyWavelet `documentation
-            <http://www.pybytes.com/pywavelets/ref/wavelets.html>`_
+        wavelet : string or ``pywt.Wavelet``
```
adler-j: `pywt.Wavelet`

odl/trafos/wavelet.py
```diff
@@ -21,656 +21,260 @@
...
+            ``'rbio'``: Reverse biorthogonal
 
-            Reverse biorthogonal (``rbio``)
+            ``'dmey'``: Discrete FIR approximation of the Meyer wavelet
 
-            Discrete FIR approximation of Meyer wavelet (``dmey``)
+        nlevels : positive int
+            Number of scaling levels to be used in the decomposition. The
+            maximum number of levels can be calculated with
+            `pywt.dwt_max_level
```
adler-j: `pywt.dwt_max_level` etc in other places.

odl/util/normalize.py
```diff
@@ -334,6 +335,54 @@ def normalized_nodes_on_bdry(nodes_on_bdry, length):
     return out_list
 
 
+def normalized_axes_tuple(axes, ndim):
+    """Return a tuple of ``axes`` converted to integers <= ``ndim``.
```
adler-j: dont quite get how they are converted

conftest.py
```diff
@@ -65,6 +73,26 @@ def ntuples_impl(request):
...
+def floating_dtype(request):
+    return request.param
+
+
+dtype_params = floating_dtype_params + np.sctypes['int'] + np.sctypes['uint']
+dtype_ids = [' dtype={} '.format(dtype_repr(dt)) for dt in dtype_params]
+
+
+@pytest.fixture(scope="module", ids=dtype_ids, params=dtype_params)
+def dtype(request):
```
kohr-h: ok

odl/space/base_ntuples.py
```diff
@@ -493,12 +493,18 @@ def __init__(self, size, dtype):
             self.__is_real = True
             self.__real_dtype = self.dtype
             self.__real_space = self
-            self.__complex_dtype = TYPE_MAP_R2C.get(self.dtype, None)
+            try:
+                self.__complex_dtype = complex_dtype(self.dtype)
```
kohr-h: No, that was the reason I had the `use_default` thing before. Now if no dtype is found and `default=None`, it's an error. If we want that behavior, then okay.

odl/trafos/backends/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Bindings to external backends for transformations."""
```
kohr-h: Because it's installed with minimal dependencies and can be seen as "not external". It's a matter of taste, of course, but I saw no point in moving out that stuff without a real need to compartmentalize it.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,313 @@
...
+
+    if not array_out.flags.aligned:
+        raise ValueError('output array not aligned')
+
+    if axes is None:
+        axes = tuple(range(array_in.ndim))
+
+    axes = normalized_axes_tuple(axes, array_in.ndim)
+
+    direction = _pyfftw_to_local(direction)
```
kohr-h: I know. But this was a pure code move, and there is a separate issue #616 on that. Otherwise this branch never goes in.

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -0,0 +1,313 @@
...
+        return True
+    elif (direction in ('backward', 'FFTW_BACKWARD') and halfcomplex and
+          ndim != 1):
+        return True
+    else:
+        return False
+
+
+def _pyfftw_check_args(arr_in, arr_out, axes, halfcomplex, direction):
+    """Raise an error if anything is not ok with in and out."""
```
kohr-h: All the way down here is outside the scope of this PR, will be taken care of in #616 or #617.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,799 @@
...
+        raise ValueError('`shape` may only contain integers, got {}'
+                         ''.format(shape))
+    wavelet = pywt_wavelet(wavelet)
+
+    nlevels, nlevels_in = int(nlevels), nlevels
+    if nlevels_in != nlevels:
+        raise ValueError('`nlevels` must be integer, got {}'
+                         ''.format(nlevels_in))
+    # TODO: adapt for axes
+    _check_nlevels(nlevels, min(shape), wavelet)
```
kohr-h: ok

odl/trafos/backends/pywt_bindings.py
```diff
@@ -0,0 +1,799 @@
...
+    ...            [[-0.5, 0.5],
+    ...             [-0.5, 0.5]])
+    >>>
+    >>> # Gives even shape by default
+    >>> pywt_single_level_recon(approx, details, wavelet='haar', mode='zpd')
+    array([[ 1.,  1.,  1.,  0.],
+           [ 1.,  0.,  0.,  0.],
+           [ 0.,  1.,  1.,  0.],
+           [ 0.,  0.,  0.,  0.]])
+    >>>
```
kohr-h: I actually want a blank line there due to the subsequent comment line.

odl/trafos/wavelet.py
kohr-h: Ah, forgot that file when changing the links.

odl/trafos/backends/pywt_bindings.py
kohr-h: This makes the whole function kind of superfluous. Especially since one needs `max_levels` outside to print the proper error message. Not so bad probably since it's used twice only.

odl/trafos/backends/pywt_bindings.py
adler-j: Blank line is perfectly fine by me You can even go blank line, comment, blank line. That gives a better style imo

kohr-h: Decent suggestion, will do

kohr-h:
Included all suggestions from the review, apart from the `try..except` thing with `complex_dtype` in `base_ntuples.py`. Should be ready for merge.

kohr-h:
I'll clean up the commit history, of course.

adler-j:
Go ahead

adler-j:
All the more reason to get #596 done now.

adler-j:
You put the tests in the wrong folder :O Fixed in 796927e91f8c815a69dd449c17ac58b9c57e8fb2

PR #549, opened 2016-08-30
DOC: move documentation from _call to __init__, closes #530

### LINKED ISSUE #530: Move usage examples from _call to __init__

adler-j:
Quite trivial pull, moved doc from _call to __init__. Should have no changes to actual code.

### DIFF
--- a/odl/discr/diff_ops.py
+++ b/odl/discr/diff_ops.py
@@ -84,6 +84,19 @@ class PartialDerivative(PointwiseTensorFieldOperator):
         edge_order : {1, 2}, optional
             Edge-order accuracy at the boundaries if no padding is used.
             Default: Same order as in the interior.
+
+        Examples
+        --------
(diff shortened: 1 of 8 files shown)


adler-j:
Minor bugs, will fix and rebase after review.

kohr-h:
Were that all examples in `_call`? Was not that bad after all..
Anyway, fix, rebase and merge.

adler-j:
Basically did a regexpr for `_call` and moved all examples to `__init__`, so these are all.

adler-j:
Merge after CI.

PR #554, opened 2016-08-31
Nonlinear conjugate gradient

adler-j:
Work in progress, but I feel that we need this sooner or later.

### DIFF
--- a/odl/solvers/iterative/iterative.py
+++ b/odl/solvers/iterative/iterative.py
@@ -24,7 +24,7 @@ standard_library.install_aliases()
 
 from odl.operator import IdentityOperator, OperatorComp, OperatorSum
 from odl.util import normalized_scalar_param_list
-from odl.solvers.scalar import BacktrackingLineSearch
+from odl.solvers.util import ConstantLineSearch, BacktrackingLineSearch
 
 
 __all__ = ('landweber', 'conjugate_gradient', 'conjugate_gradient_normal',
@@ -284,7 +284,7 @@ Conjugate_gradient_on_the_normal_equations>`_.
 
 
 def conjugate_gradient_nonlinear(f, x, rhs, niter=1, nreset=0,
-                                 line_search=1.0, tol=1e-16, beta_method='FR',
+                                 line_search=None, tol=1e-16, beta_method='FR',
                                  callback=None):
     """Conjugate gradient for nonlinear problems.
 
@@ -331,10 +331,13 @@ def conjugate_gradient_nonlinear(f, x, rhs, niter=1, nreset=0,
         raise TypeError('`x` {!r} is not in the domain of `f` {!r}'
                         ''.format(x, f.domain))
 
-    if not callable(line_search):
-        line_search = BacktrackingLineSearch(f, discount=0.05)
+    if line_search is None:
+        line_search = BacktrackingLineSearch(f, estimate_step=True)
+    elif not callable(line_search):
+        line_search = ConstantLineSearch(line_search)
 
     for rest_nr in range(nreset + 1):
+        # First iteration is done without beta
         dx = -f.gradient(x)
         dir_derivative = -dx.inner(dx)
         if abs(dir_derivative) < tol:
@@ -346,8 +349,10 @@ def conjugate_gradient_nonlinear(f, x, rhs, niter=1, nreset=0,
         s = dx  # for 'HS' and 'DY' beta methods
 
         for _ in range(niter):
+            # Compute dx as -grad f
             dx, dx_old = -f.gradient(x), dx
 
+            # Calculate "beta"
             if beta_method == 'FR':
                 beta = dx.inner(dx) / dx_old.inner(dx_old)
             elif beta_method == 'PR':
@@ -359,14 +364,19 @@ def conjugate_gradient_nonlinear(f, x, rhs, niter=1, nreset=0,
             else:
                 raise ValueError('unknown ``beta_method``')
 
-            if abs(beta) < tol:
-                return
+            # Reset beta if negative.
+            beta = max(0, beta)
 
-            s = dx + beta * s
-            dir_derivative = -s.inner(s)
+            # Update search direction
+            s.lincomb(1, dx, beta, s)  # s = dx + beta * s
+
+            # Find optimal step along s
+            dir_derivative = -dx.inner(s)
             if abs(dir_derivative) < tol:
                 return
             a = line_search(x, s, dir_derivative)
+
+            # Update position
             x.lincomb(1, x, a, s)  # x = x + a * s
 
             if callback is not None:
--- a/odl/solvers/iterative/iterative.py
+++ b/odl/solvers/iterative/iterative.py
@@ -26,6 +26,7 @@ from odl.operator import IdentityOperator, OperatorComp, OperatorSum
 from odl.util import normalized_scalar_param_list
 from odl.solvers.scalar import BacktrackingLineSearch
 
+
 __all__ = ('landweber', 'conjugate_gradient', 'conjugate_gradient_normal',
            'conjugate_gradient_nonlinear', 'gauss_newton', 'kaczmarz')
 
@@ -282,21 +283,10 @@ Conjugate_gradient_on_the_normal_equations>`_.
             callback(x)
 
 
-def conjugate_gradient_nonlinear(op, x, rhs, niter=1, niter_reset=None,
-                                 line_search=None, beta_method='FR',
+def conjugate_gradient_nonlinear(f, x, rhs, niter=1, nreset=0,
+                                 line_search=1.0, tol=1e-16, beta_method='FR',
                                  callback=None):
-    """Optimized implementation of CG for the normal equations.
-
-    This method solves the inverse problem (of the first kind)
-
-    ``A(x) == rhs``
-
-    with a linear `Operator` ``A`` by looking at the normal equations
-
-    ``A.adjoint(A(x)) == A.adjoint(rhs)``
-
-    It uses a minimum amount of memory copies by applying re-usable
-    temporaries and in-place evaluation.
+    """Conjugate gradient for nonlinear problems.
 
     The method is described in a
     `Wikipedia article
@@ -304,7 +294,7 @@ def conjugate_gradient_nonlinear(op, x, rhs, niter=1, niter_reset=None,
 
     Parameters
     ----------
-    op : `Operator`
+    op : `Functional`
         Operator in the inverse problem. If not linear, it must have
         an implementation of `Operator.derivative`, which
         in turn must implement `Operator.adjoint`, i.e.
@@ -316,19 +306,18 @@ def conjugate_gradient_nonlinear(op, x, rhs, niter=1, niter_reset=None,
     rhs : ``op.range`` element
         Right-hand side of the equation defining the inverse problem
     niter : int
-        Maximum number of iterations in total.
-    niter_reset : int, optional
-        Maximum number of iterations before reset. Default: no reset.
-    line_search : `LineSearch`, optional.
-        Line searching method to use. Default: `BacktrackingLineSearch`.
+        Number of iterations per reset.
+    nreset : int, optional
+        Number of times the solver should be reset. Default: no reset.
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
+        fixed step length. Default: `BacktrackingLineSearch`
+    tol : float, optional
+        Tolerance that should be used for terminating the iteration.
     beta_method : {'FR', 'PR', 'HS', 'DY'}
         Method to calculate ``beta`` in the iterates. TODO
     callback : `callable`, optional
         Object executing code per iteration, e.g. plotting each iterate
-
-    Returns
-    -------
-    None
 
     See Also
     --------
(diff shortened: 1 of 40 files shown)


adler-j:
This obviously needs the PR #587 before proceeding.

adler-j:
This is now ready for a review. Only remaining thing to do is that this should be moved out of `iterative` and into `scalar`.

adler-j:
Bump for review

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+        # First iteration is done without beta
+        dx = -f.gradient(x)
+        dir_derivative = -dx.inner(dx)
+        if abs(dir_derivative) < tol:
+            return
+        a = line_search(x, dx, dir_derivative)
+        x.lincomb(1, x, a, dx)  # x = x + a * dx
+
+        dx_old = dx
+        s = dx  # for 'HS' and 'DY' beta methods
```
kohr-h: Not only that, you also use them in the search direction update and the line search, no?

odl/test/solvers/iterative/iterative_test.py
```diff
@@ -49,13 +53,19 @@ def solver(op, x, rhs):
     elif solver_name == 'landweber':
         def solver(op, x, rhs):
             norm2 = op.adjoint(op(x)).norm() / x.norm()
-            odl.solvers.landweber(op, x, rhs, niter=10, omega=0.5 / norm2)
+            odl.solvers.landweber(op, x, rhs, niter=50, omega=0.5 / norm2)
```
kohr-h: reason?

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
             callback(x)
 
 
+def conjugate_gradient_nonlinear(f, x, rhs, niter=1, nreset=0,
+                                 line_search=None, tol=1e-16, beta_method='FR',
+                                 callback=None):
```
adler-j: I'll see what I can do. Overall our solvers could use a look so that the parameters fit better.

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+    """Conjugate gradient for nonlinear problems.
+
+    The method is described in a
+    `Wikipedia article
+    <https://en.wikipedia.org/wiki/Nonlinear_conjugate_gradient_method>`_.
```
adler-j: Will get that done, also short word on algorithm.

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+    x : ``op.domain`` element
+        Vector to which the result is written. Its initial value is
+        used as starting point of the iteration, and its values are
+        updated in each iteration step.
+    rhs : ``op.range`` element
+        Right-hand side of the equation defining the inverse problem
+    niter : int
+        Number of iterations per reset.
+    nreset : int, optional
+        Number of times the solver should be reset. Default: no reset.
```
adler-j: I'll do your first suggestion.

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
+        fixed step length. Default: `BacktrackingLineSearch`
+    tol : float, optional
+        Tolerance that should be used for terminating the iteration.
+    beta_method : {'FR', 'PR', 'HS', 'DY'}
+        Method to calculate ``beta`` in the iterates. TODO
```
adler-j: Yeah I'll add that but that goes up in notes, not here since that would clog this.

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+    callback : `callable`, optional
+        Object executing code per iteration, e.g. plotting each iterate
+
+    See Also
+    --------
+    conjugate_gradient : Optimized solver for linear and symmetric case
+    conjugate_gradient_normal : Equivalent solver but for linear case
+    """
+    # TODO: add a book reference
```
adler-j: Agree

odl/test/solvers/iterative/iterative_test.py
adler-j: It started failing due to some other change.

odl/solvers/iterative/iterative.py
aringh: If you want a book reference there is one in GNS2009 (at least for what wiki calls Flethcer-Reeves update) which we already have in our list of references

odl/solvers/iterative/iterative.py
aringh: As I wrote higher up, we have a book reference for FR: GNS2009

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+        the call ``op.derivative(x).adjoint`` must be valid.
+    x : ``op.domain`` element
+        Vector to which the result is written. Its initial value is
+        used as starting point of the iteration, and its values are
+        updated in each iteration step.
+    rhs : ``op.range`` element
+        Right-hand side of the equation defining the inverse problem
+    niter : int
+        Number of iterations per reset.
+    nreset : int, optional
```
aringh: If we are interested there are more advanced ways to determine when to reset. The one I know about is by Powell from 1977, but maybe there are more recent ones as well: http://link.springer.com/article/10.1007/BF01593790

odl/solvers/iterative/iterative.py
```diff
@@ -290,6 +289,106 @@ def conjugate_gradient_normal(op, x, rhs, niter=1, callback=None):
...
+                beta = dx.inner(dx - dx_old) / dx_old.inner(dx_old)
+            elif beta_method == 'HS':
+                beta = dx.inner(dx - dx_old) / s.inner(dx - dx_old)
+            elif beta_method == 'DY':
+                beta = dx.inner(dx) / s.inner(dx - dx_old)
+            else:
+                raise ValueError('unknown ``beta_method``')
+
+            # Reset beta if negative.
+            beta = max(0, beta)
```
aringh: If you take `beta = 0`, you will effectively reset the conjugate gradient method since `s` just becomes the steepest decent direction. I don't know if that actually matters, but we should think through how we want to handle that.

adler-j: I basically got this from wikipedia, I'm far from an expert. Personally, I'd feel fine with this version for now and then improving it later (if anyone wants to use it).

aringh: I don't consider myself an expert in the area either ;-) and it is a perfectly fine solution for me (if you look at the reference that wikipedia uses, it claims convergence of the PR-version if this is used). I justed wanted to bring to attention that it actually is a restart of the method.

odl/solvers/iterative/iterative.py
adler-j: Did now, but it is now "the same" as the other smooth solvers.

odl/solvers/iterative/iterative.py
adler-j: Second thought, I wont. This can be easily googled and writing it down here would require quite a few lines.

adler-j:
Fixed comments and moved the code into the smooth solvers, guess it needs a re-review.

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,141 @@
...
+    f : `Functional`
+        Functional with ``f.gradient``.
+    x : ``op.domain`` element
+        Vector to which the result is written. Its initial value is
+        used as starting point of the iteration, and its values are
+        updated in each iteration step.
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
+        fixed step length.
+    maxiter : int
```
kohr-h: I guess because the algorithm can terminate early and is not guaranteed to run this number of iterations. Should be cross-checked anyway and changed in the other methods if it applies there, too. I like the distinction.

kohr-h:
> Minor comments, except for a potential sign error.
> 
> If there is a sign error it is a bit weird that the tests pass. I have double checked with the reference on wikipedia for the DY-formula, which contains all four expressions for beta, an it looks like there should indeed be a minus sign for HS and DY.

Could also be a sign error in the Wikipedia reference. Looking at the formulas it seems weird that the last two have this ominous minus sign although the rest looks kind of similar to the first two formulas. The simplest thing would be to flip the sign and see what happens.

aringh:
> The simplest thing would be to flip the sign and see what happens.

I did, and the tests still pass. Which is a bit troublesome.

The reason for the minus signs on wikipedia but that they are not there is in the reference, which is non-obvious from a first look, is that the wikipedia page works with `delta x` which is `- grad` while the reference works with `grad` directly. But someone can double check that I haven't made other sign errors when I tried to check it :-)

odl/solvers/smooth/nonlinear_cg.py
adler-j: Agree with @kohr-h, that is why. Keeping this

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,141 @@
...
+    bfgs_method : Quasi-newton solver for the same problem
+    conjugate_gradient : Optimized solver for linear and symmetric case
+    conjugate_gradient_normal : Equivalent solver but for linear case
+    """
+
+    if x not in f.domain:
+        raise TypeError('`x` {!r} is not in the domain of `f` {!r}'
+                        ''.format(x, f.domain))
+
+    if line_search is None:
```
adler-j: No you are correct, it should be "if isinstance", or perhaps a try catch float conversion.

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,141 @@
...
+            # Compute dx as -grad f
+            dx, dx_old = -f.gradient(x), dx
+
+            # Calculate "beta"
+            if beta_method == 'FR':
+                beta = dx.inner(dx) / dx_old.inner(dx_old)
+            elif beta_method == 'PR':
+                beta = dx.inner(dx - dx_old) / dx_old.inner(dx_old)
+            elif beta_method == 'HS':
+                beta = dx.inner(dx - dx_old) / s.inner(dx - dx_old)
```
adler-j: Where? What source are you using to see that?

adler-j:
Holding on until the sign error is fixed, I'll have a look. Nice catch.

adler-j:
The sign error was indeed a sign error. Fixed it and number of iterations needed went down drastically.

adler-j:
Can I merge this?

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+        used as starting point of the iteration, and its values are
+        updated in each iteration step.
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
+        fixed step length.
+    maxiter : int, optional
+        Maximum number of iterations to perform.
+    nreset : int, optional
+        Number of times the solver should be reset. Default: no reset.
+        fixed step length. Default: `BacktrackingLineSearch`
```
kohr-h: Something got messed up here.

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+    tol : float, optional
+        Tolerance that should be used for terminating the iteration.
+    beta_method : {'FR', 'PR', 'HS', 'DY'}
+        Method to calculate ``beta`` in the iterates.
+        * ``'FR'`` : Fletcher-Reeves
+        * ``'PR'`` : Polak-Ribiere
+        * ``'HS'`` : Hestenes-Stiefel
+        * ``'DY'`` : Dai-Yuan
+    callback : `callable`, optional
+        Object executing code per iteration, e.g. plotting each iterate
```
kohr-h: dot

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+
+    See Also
+    --------
+    bfgs_method : Quasi-newton solver for the same problem
+    conjugate_gradient : Optimized solver for linear and symmetric case
```
kohr-h: "least-squares problem with linear and symmetric operator" maybe? Otherwise not so clear what "linear case" means, could also be "linear functional"

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+    conjugate_gradient_normal : Equivalent solver but for linear case
```
kohr-h: accordingly here

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+from odl.solvers.util import ConstantLineSearch
+
+
+__all__ = ('conjugate_gradient_nonlinear',)
+
+
+def conjugate_gradient_nonlinear(f, x, line_search=1.0, maxiter=1000, nreset=0,
+                                 tol=1e-16, beta_method='FR',
+                                 callback=None):
+    """Conjugate gradient for nonlinear problems.
```
kohr-h: method
also perhaps "general nonlinear problems" since we're also solving a nonlinear problem in the other case (the least squares problem)

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+    Parameters
+    ----------
+    f : `Functional`
+        Functional with ``f.gradient``.
+    x : ``op.domain`` element
+        Vector to which the result is written. Its initial value is
+        used as starting point of the iteration, and its values are
+        updated in each iteration step.
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
```
kohr-h: it is used

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Reset beta if negative.
+            beta = max(0, beta)
+
+            # Update search direction
+            s.lincomb(1, dx, beta, s)  # s = dx + beta * s
+
+            # Find optimal step along s
+            dir_derivative = -dx.inner(s)
+            if abs(dir_derivative) < tol:
```
kohr-h: Strictly speaking it should be `<=` here because "tolerance" in inclusive. It's also relevant if a user sets `tol=0` because in some (toy) cases the update could actually be 0 and the method  should terminate then.

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,139 @@
...
+
+__all__ = ('conjugate_gradient_nonlinear',)
+
+
+def conjugate_gradient_nonlinear(f, x, line_search=1.0, maxiter=1000, nreset=0,
+                                 tol=1e-16, beta_method='FR',
+                                 callback=None):
+    """Conjugate gradient for nonlinear problems.
+
+    Notes
```
adler-j: We do that in some cases, particularily when the description becomes too "thin" otherwise.

adler-j:
Ready for merge?

odl/solvers/smooth/nonlinear_cg.py
kohr-h: Ok, but in the long run I find it preferable to get directly to the parameters without much ado since that's what users have to look at most frequently. The math explanation is useful once or a couple of times, but later it's just in the way to the parameter reference.
So I prefer the note section after Parameters in almost all cases.

kohr-h:
> Ready for merge?

That's for @aringh to decide ;-)

kohr-h:
Perhaps you can also squash the first two commits to get rid of that WIP commit.

adler-j:
Will do once this is accepted

adler-j:
Bump for @aringh attention.

odl/solvers/smooth/nonlinear_cg.py
```diff
@@ -0,0 +1,140 @@
...
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, it is used as
+        a fixed step length.
+    maxiter : int, optional
+        Maximum number of iterations to perform.
+    nreset : int, optional
+        Number of times the solver should be reset. Default: no reset.
+    tol : float, optional
+        Tolerance that should be used for terminating the iteration.
+    beta_method : {'FR', 'PR', 'HS', 'DY'}
```
aringh: `optional`. Mention default value?

kohr-h: We do that only if it's not clear from the signature, i.e., when the parameter is in `kwargs`. Or if the default is `None` and we need to explain what happens in that case.

The convention for "limited-range" parameters is to name the default first. So I'd say it's fine as it is.

adler-j:
We'll @aringh had this innocent comment about missing a "optional" so I made [a script](https://gist.github.com/adler-j/e7aeba3a5db65e4d1a3384ceb0ed0630) to find stuff like this, and needless to say I found like 100+ errors.

Fixed them all so going to need a further review.

aringh:
I had a quick look, and it looks good to me :-)

adler-j:
Merge after CI.

PR #555, opened 2016-09-01
Issue 474  broadcast pspace arithmetic

adler-j:
Quite well needed support for broadcasting, currently only supports "single power spaces", i.e.

```
ProductSpace(ProductSpace(space, 2), 2)
```

will still fail to multiply with `space`. I guess we are fine with that for now, but it should be added.

Also did some other general fixes.

### DIFF
--- a/odl/set/space.py
+++ b/odl/set/space.py
@@ -457,9 +457,9 @@ class LinearSpaceElement(object):
                 # other --> other * space.one()
                 return self.space.lincomb(1, self, other, one(), out=self)
         else:
-            # We do not raise here since we don't want a fallback for inplace.
-            # Otherwise python attempts self = self / other which does not
-            # modify self.
+            # We do not `return NotImplemented` here since we don't want a
+            # fallback for inplace. Otherwise python attempts
+            # `self = self + other` which does not modify self.
             raise TypeError('cannot add {!r} and {!r} in place'
                             ''.format(self, other))
 
@@ -492,9 +492,9 @@ class LinearSpaceElement(object):
             else:
                 return self.space.lincomb(1, self, -other, one(), out=self)
         else:
-            # We do not raise here since we don't want a fallback for inplace.
-            # Otherwise python attempts self = self - other which does not
-            # modify self.
+            # We do not `return NotImplemented` here since we don't want a
+            # fallback for inplace. Otherwise python attempts
+            # `self = self - other` which does not modify self.
             raise TypeError('cannot subtract {!r} and {!r} in place'
                             ''.format(self, other))
 
@@ -538,9 +538,9 @@ class LinearSpaceElement(object):
         elif other in self.space:
             return self.space.multiply(other, self, out=self)
         else:
-            # We do not raise here since we don't want a fallback for inplace.
-            # Otherwise python attempts self = self * other which does not
-            # modify self.
+            # We do not `return NotImplemented` here since we don't want a
+            # fallback for inplace. Otherwise python attempts
+            # `self = self * other` which does not modify self.
             raise TypeError('cannot multiply {!r} and {!r} in place'
                             ''.format(self, other))
 
@@ -565,9 +565,9 @@ class LinearSpaceElement(object):
         elif other in self.space:
             return self.space.divide(self, other, out=self)
         else:
-            # We do not raise here since we don't want a fallback for inplace.
-            # Otherwise python attempts self = self / other which does not
-            # modify self.
+            # We do not `return NotImplemented` here since we don't want a
+            # fallback for inplace. Otherwise python attempts
+            # `self = self + other` which does not modify self.
             raise TypeError('cannot divide {!r} and {!r} in place'
                             ''.format(self, other))
 
--- a/odl/set/space.py
+++ b/odl/set/space.py
@@ -458,7 +458,7 @@ class LinearSpaceElement(object):
                 return self.space.lincomb(1, self, other, one(), out=self)
         else:
             # We do not `return NotImplemented` here since we don't want a
-            # fallback for inplace. Otherwise python attempts
+            # fallback for in-place. Otherwise python attempts
             # `self = self + other` which does not modify self.
             raise TypeError('cannot add {!r} and {!r} in place'
                             ''.format(self, other))
@@ -493,7 +493,7 @@ class LinearSpaceElement(object):
                 return self.space.lincomb(1, self, -other, one(), out=self)
         else:
             # We do not `return NotImplemented` here since we don't want a
-            # fallback for inplace. Otherwise python attempts
+            # fallback for in-place. Otherwise python attempts
             # `self = self - other` which does not modify self.
             raise TypeError('cannot subtract {!r} and {!r} in place'
                             ''.format(self, other))
@@ -539,7 +539,7 @@ class LinearSpaceElement(object):
             return self.space.multiply(other, self, out=self)
         else:
             # We do not `return NotImplemented` here since we don't want a
-            # fallback for inplace. Otherwise python attempts
+            # fallback for in-place. Otherwise python attempts
             # `self = self * other` which does not modify self.
             raise TypeError('cannot multiply {!r} and {!r} in place'
                             ''.format(self, other))
@@ -566,8 +566,8 @@ class LinearSpaceElement(object):
             return self.space.divide(self, other, out=self)
         else:
             # We do not `return NotImplemented` here since we don't want a
-            # fallback for inplace. Otherwise python attempts
-            # `self = self + other` which does not modify self.
+            # fallback for in-place. Otherwise python attempts
+            # `self = self / other` which does not modify self.
             raise TypeError('cannot divide {!r} and {!r} in place'
                             ''.format(self, other))
 
--- a/odl/set/space.py
+++ b/odl/set/space.py
@@ -457,7 +457,11 @@ class LinearSpaceElement(object):
                 # other --> other * space.one()
                 return self.space.lincomb(1, self, other, one(), out=self)
         else:
-            return NotImplemented
+            # We do not raise here since we don't want a fallback for inplace.
+            # Otherwise python attempts self = self / other which does not
+            # modify self.
+            raise TypeError('cannot add {!r} and {!r} in place'
+                            ''.format(self, other))
 
     def __add__(self, other):
         """Return ``self + other``."""
@@ -488,7 +492,11 @@ class LinearSpaceElement(object):
             else:
                 return self.space.lincomb(1, self, -other, one(), out=self)
         else:
-            return NotImplemented
+            # We do not raise here since we don't want a fallback for inplace.
+            # Otherwise python attempts self = self - other which does not
+            # modify self.
+            raise TypeError('cannot subtract {!r} and {!r} in place'
+                            ''.format(self, other))
 
     def __sub__(self, other):
         """Return ``self - other``."""
@@ -508,6 +516,9 @@ class LinearSpaceElement(object):
 
     def __rsub__(self, other):
         """Return ``other - self``."""
+        if other in self.space:
+            tmp = self.space.element()
+            return self.space.lincomb(1, other, -1, self, out=tmp)
         if other in self.space.field:
             one = getattr(self.space, 'one', None)
             if one is None:
@@ -518,7 +529,6 @@ class LinearSpaceElement(object):
                 self.space.lincomb(other, tmp, out=tmp)
                 return self.space.lincomb(1, tmp, -1, self, out=tmp)
         else:
-            # Case `other in self.space` handled by `other`
             return NotImplemented
 
     def __imul__(self, other):
@@ -528,7 +538,11 @@ class LinearSpaceElement(object):
         elif other in self.space:
             return self.space.multiply(other, self, out=self)
         else:
-            return NotImplemented
+            # We do not raise here since we don't want a fallback for inplace.
+            # Otherwise python attempts self = self * other which does not
+            # modify self.
+            raise TypeError('cannot multiply {!r} and {!r} in place'
+                            ''.format(self, other))
 
     def __mul__(self, other):
         """Return ``self * other``."""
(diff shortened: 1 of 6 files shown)


odl/space/pspace.py
```diff
@@ -260,6 +260,8 @@ class instance can be retrieved from the space by the
 
         # Assign spaces and field
         self.__spaces = tuple(spaces)
+        self.__is_power_space = all(spc == self.spaces[0]
```
adler-j: Note on why this is cached

odl/set/space.py
```diff
@@ -475,7 +475,11 @@ def __iadd__(self, other):
                 # other --> other * space.one()
                 return self.space.lincomb(1, self, other, one(), out=self)
         else:
-            return NotImplemented
+            # We do not raise here since we don't want a fallback for inplace.
```
kohr-h: We do raise here?
in-place

odl/set/space.py
```diff
@@ -475,7 +475,11 @@ def __iadd__(self, other):
...
+            # Otherwise python attempts self = self / other which does not
```
kohr-h: `self + other`

odl/space/pspace.py
```diff
@@ -573,6 +575,39 @@ def element_type(self):
         return ProductSpaceVector
 
 
+def _broadcast_arithmetic(op, raises=False):
+    """Return self operator other with broadcasting
```
kohr-h: I don't understand this sentence.

odl/space/pspace.py
```diff
@@ -573,6 +575,39 @@ def element_type(self):
...
+
+    Parameters
+    ----------
+    op : string
+        Name of the operator, e.g. ``'__add__'``.
+    raises : bool
+        True if the method should raise on failure, otherwise returns
+        ``NotImplemented``.
+    """
+
```
kohr-h: no blank line

odl/space/pspace.py
```diff
@@ -573,6 +575,39 @@ def element_type(self):
...
 
 
+def _broadcast_arithmetic(op, raises=False):
+    """Return self operator other with broadcasting
+
+    Parameters
+    ----------
+    op : string
+        Name of the operator, e.g. ``'__add__'``.
+    raises : bool
```
kohr-h: , optional

odl/space/pspace.py
```diff
@@ -573,6 +575,39 @@ def element_type(self):
...
+        True if the method should raise on failure, otherwise returns
```
kohr-h: This sentence is logically wrong. It should be "If true, then ...", while yours means "If ..., then true".
Suggestion:
```
For ``True``, the function raises on failure, otherwise it returns ``NotImplemented``.
```

odl/space/pspace.py
adler-j: Can be rewritten `op(self, other)` now.

adler-j:
What about the design?

odl/space/pspace.py
```diff
@@ -629,6 +664,22 @@ def __setitem__(self, indices, values):
...
+    __isub__ = _broadcast_arithmetic('__isub__')
+    __mul__ = _broadcast_arithmetic('__mul__')
+    __rmul__ = _broadcast_arithmetic('__rmul__')
+    __imul__ = _broadcast_arithmetic('__imul__')
+    __div__ = _broadcast_arithmetic('__div__')
+    __rdiv__ = _broadcast_arithmetic('__rdiv__')
+    __idiv__ = _broadcast_arithmetic('__idiv__')
+    __truediv__ = _broadcast_arithmetic('__truediv__')
+    __rtruediv__ = _broadcast_arithmetic('__rtruediv__')
+    __itruediv__ = _broadcast_arithmetic('__itruediv__')
```
kohr-h: Nice generalization!

odl/space/pspace.py
```diff
@@ -573,6 +575,39 @@ def element_type(self):
...
+    Parameters
+    ----------
+    op : string
+        Name of the operator, e.g. ``'__add__'``.
+    raises : bool
+        True if the method should raise on failure, otherwise returns
+        ``NotImplemented``.
+    """
+
+    def _broadcast_arithmetic_impl(self, other):
```
kohr-h: Some short doc for the developer?

odl/util/testutils.py
```diff
@@ -251,19 +251,37 @@ def _pass(function):
...
 def example_element(space):
-    return space.element(example_array(space))
+    """Construct an example element in a space.
+
+    The element is chosen to be a totally random vector.
+
+    See Also
+    --------
+    example_array
+    Set.examples : More typical examles for the space.
```
kohr-h: examples

odl/util/testutils.py
```diff
@@ -251,19 +251,37 @@ def _pass(function):
...
 
-    return arr.astype(space.dtype, copy=False)
+        return arr.astype(space.dtype, copy=False)
 
 
 def example_element(space):
-    return space.element(example_array(space))
+    """Construct an example element in a space.
+
+    The element is chosen to be a totally random vector.
```
kohr-h: What information does "totally" add?

odl/space/pspace.py
adler-j: could hypthetically be written as a loop:

```python
for op in ['add', 'sub', 'mul', 'div', 'truediv']:
    for modifier in ['', 'r', 'i']:
        name = '__{}{}__'.format(modifier, op)
        setattr(ProductSpaceVector(name), _broadcast_arithmetic(name))
```

I'm not sure if this is better or not.

odl/space/pspace.py
adler-j: Sure

test/space/ntuples_test.py
```diff
@@ -346,112 +347,79 @@ def test_power(fn):
...
+        x_arr, x = example_vectors(fn)
+
+        if scalar == 0 and function in [operator.div,
+                                        operator.idiv,
+                                        operator.truediv,
+                                        operator.itruediv]:
+            # Check for correct zero division behaviour
+            with pytest.raises(ZeroDivisionError):
+                y_arr = function(x_arr, scalar)
+                y = function(x, scalar)
```
kohr-h: This will always raise Numpy's `ZeroDivisionError` and shadow ours, no?

odl/space/pspace.py
adler-j: This parameter is old and will be removed.

test/space/ntuples_test.py
adler-j: Will fix. It's correct in pspace_test i think.

kohr-h: No, probably Numpy creates NaNs in the array. Anyway, the first of the two lines should be obsolete.

odl/space/pspace.py
kohr-h: With a comment explaining what's going on and
```python
setattr(ProductSpaceVector, name, _broadcast_arithmetic(name))
```
in the last line, I think it would be totally okay.

odl/util/testutils.py
kohr-h: While at it, why not call it `random_element`? "example" suggests something reasonable.

kohr-h:
> What about the design?

As far as I can tell from the quick review, the problem you mention above resides in [exactly one place](https://github.com/odlgroup/odl/pull/555/files#diff-37c3edcf62a1d426ec34d06a50d72c22R591), so it's not an extensibility nightmare :-). 
I'm okay with this since it solves the problem at hand. More complete solutions would IMO almost require multi-indexing of the underlying data arrays, allowing a special data structure for power spaces and natural broadcasting on the lowest level.

odl/util/testutils.py
adler-j: Decent suggestion.

adler-j: #557

adler-j:
Ready for re-review I guess

odl/set/space.py
```diff
@@ -571,7 +583,11 @@ def __itruediv__(self, other):
         elif other in self.space:
             return self.space.divide(self, other, out=self)
         else:
-            return NotImplemented
+            # We do not `return NotImplemented` here since we don't want a
+            # fallback for inplace. Otherwise python attempts
```
kohr-h: see above

odl/set/space.py
```diff
@@ -571,7 +583,11 @@ def __itruediv__(self, other):
...
+            # `self = self + other` which does not modify self.
```
kohr-h: self / other

odl/space/pspace.py
```diff
@@ -823,6 +827,60 @@ def show(self, title=None, indices=None, **kwargs):
...
+    Parameters
+    ----------
+    op : string
+        Name of the operator, e.g. ``'__add__'``.
+
+    Returns
+    -------
+    broadcast_arithmetic_op : function
+        Function intended to be used as a method for `ProductSpaceVector`
+        which tries to perform broadcasting if possible.
```
kohr-h: "tries to" and "if possible" mean the same, one can go.

odl/space/pspace.py
```diff
@@ -823,6 +827,60 @@ def show(self, title=None, indices=None, **kwargs):
...
+
+    Notes
+    -----
+    Broadcasting is the operation of "applying a operator multiple times" in
+    some sense. For example:
+
+    .. math::
+        [1, 2] + 1 = [2, 3]
```
kohr-h: Why square brackets?

odl/space/pspace.py
```diff
@@ -823,6 +827,60 @@ def show(self, title=None, indices=None, **kwargs):
...
+
+    is a form of broadcasting. In this implementation, we only allow "single
+    layer" broadcasting. I.e., we do not support broadcasting over several
+    productspaces at once.
```
kohr-h: product spaces

odl/util/testutils.py
```diff
@@ -251,19 +251,37 @@ def _pass(function):
...
 def example_array(space):
     """Generate an example array that is compatible with ``space``."""
-    # Generate numpy vectors, real or complex or int
-    if np.issubdtype(space.dtype, np.floating):
-        arr = np.random.randn(space.size)
-    elif np.issubdtype(space.dtype, np.integer):
-        arr = np.random.randint(-10, 10, space.size)
+    from odl.space import ProductSpace
+    if isinstance(space, ProductSpace):
+        return np.array([example_array(si) for si in space])
```
kohr-h: Renamed `noise_array` upstream if I remember right.

test/space/ntuples_test.py
```diff
@@ -346,112 +347,76 @@ def test_power(fn):
...
+
+            assert all_almost_equal([x, y], [x_arr, y_arr])
+
+        # right op
+        x_arr, x = example_vectors(fn)
+
+        y_arr = function(scalar, x_arr)
+        y = function(scalar, x)
+
+        assert all_almost_equal([x, y], [x_arr, y_arr])
```
kohr-h: This doesn't seem to test if in-place modification works as expected. Or am I missing something?

adler-j: If `function` is `imul` then `y_arr` should be `x`, which numpy gives:

```
>>> import operator
>>> a = np.array([1, 2])
>>> operator.imul(a, 3)
array([3, 6])
>>> a
array([3, 6])
```

And ODL does the same:

```
>>> a = odl.vector([1, 2])
>>> operator.imul(a, 3)
fn(2, 'int').element([3, 6])
>>> a
fn(2, 'int').element([3, 6])
```

the only thing this does not test is that the output is actually the same object as the input, but i guess that's ok.

kohr-h: I'm aware of that behavior. I just wondered if it wouldn't make sense to test proper behavior of the in-place modifying call. As it is now, the test just verifies out-of-place.

adler-j: It should test both. The assert verifies that both the input and output argument is the same after the call, and if the function is in-place, then it gets verified. Or am I missing something?

kohr-h: ~~If we assume that `function` wouldn't work as expected and not write to its `out` parameter (which should be used as keyword arg BTW), then the test would pass since before, `x` and `x_arr` had the same value, and they're both not changed.~~
Okay, so now I finally realized that `function` comes from a fixture by another layer of indirection, and that it uses the magic methods once on the Numpy array and once on the ODL data structure. Either I'm stupid or this could very well use an explaining comment. Otherwise it's going to be magic to anyone not willing to spend 15 minutes on figuring out what's going on.

adler-j: the other layer of indirection is removed in the latest commit. Other than that I'll write some comment.

adler-j:
Updated according to commits and rebased to master. Should be mergable.

odl/space/pspace.py
```diff
@@ -821,6 +825,60 @@ def show(self, title=None, indices=None, **kwargs):
...
+
+    Returns
+    -------
+    broadcast_arithmetic_op : function
+        Function intended to be used as a method for `ProductSpaceVector`
+        which performs broadcasting if possible.
+
+    Notes
+    -----
+    Broadcasting is the operation of "applying a operator multiple times" in
```
kohr-h: an operator :-)

odl/space/pspace.py
```diff
@@ -821,6 +825,60 @@ def show(self, title=None, indices=None, **kwargs):
...
+    some sense. For example:
+
+    .. math::
+        (1, 2) + 1 = (2, 3)
+
+    is a form of broadcasting. In this implementation, we only allow "single
+    layer" broadcasting. I.e., we do not support broadcasting over several
```
kohr-h: , i.e.,

test/space/pspace_test.py
```diff
@@ -566,6 +568,83 @@ def test_element_setitem_fancy():
...
+    # Verify that the statement z=op(x, y) gives equivalent results to NumPy
+    x_arr, x = noise_elements(space, 1)
+    y_arr, y = noise_elements(pspace, 1)
+
+    # non-aliased left
+    if arithmetic_op in [operator.iadd,
+                         operator.isub,
+                         operator.itruediv,
+                         operator.imul]:
+        # Check for correct error since inplace op is not possible here
```
kohr-h: in-place

PR #561, opened 2016-09-05
ENH: Add proximal operator for second kind of KL-divergence.

aringh:
It adds the proximal operator for the convex conjugate of the second kind of the KL-divergence: `f(x) = sum_i (x_i log(x_i) - x_i log(g_i) + g_i - x_i`.

The naming is an issue here... what should we call the two different KL-divergences? The current name is temporary.

The two functionals are highly related as they only difference is that the prior and the variable changes place in expression. However, they are two different functionals with a bit different characteristics.

EDIT: It also contains a few update in the doc of the proximal operator for the previous KL-divergence (typos).

### DIFF
--- a/odl/solvers/advanced/proximal_operators.py
+++ b/odl/solvers/advanced/proximal_operators.py
@@ -31,6 +31,7 @@ standard_library.install_aliases()
 from builtins import super
 
 import numpy as np
+import scipy.special
 
 from odl.operator import (Operator, IdentityOperator, ScalingOperator,
                           ConstantOperator, ResidualOperator, DiagonalOperator)
@@ -45,7 +46,7 @@ __all__ = ('combine_proximals', 'proximal_cconj', 'proximal_translation',
            'proximal_l1', 'proximal_cconj_l1',
            'proximal_l2', 'proximal_cconj_l2',
            'proximal_l2_squared', 'proximal_cconj_l2_squared',
-           'proximal_cconj_kl')
+           'proximal_cconj_kl', 'proximal_cconj_kl_version_two')
 
 
 # TODO: remove diagonal op once available on master
@@ -989,7 +990,7 @@ def proximal_cconj_kl(space, lam=1, g=None):
     that F is defined over whole X. The non-negativity thresholding pos is
     used to define F in the real numbers.
 
-    The proximal operator of the convex conjugate F^* of F is
+    The convex conjugate F^* of F is
 
         F^*(p) = sum_i (-g ln(pos(1_X - p))_i + ind_P(1_X - p)
 
@@ -999,7 +1000,7 @@ def proximal_cconj_kl(space, lam=1, g=None):
     The proximal operator of the convex conjugate of F is
 
         prox[sigma * F^*](x) =
-            1/2 (lam + x - sqrt((x - lam)^2 + 4 lam sigma g)
+            1/2 (lam + x - sqrt((x - lam)^2 + 4 lam sigma g))
 
     with the step size parameter sigma and lam_X is an element of the space X
     with all components set to lam.
@@ -1075,6 +1076,91 @@ def proximal_cconj_kl(space, lam=1, g=None):
 
             # out = 1/2 * out
             out /= 2
+
+    return ProximalCConjKL
+
+
+def proximal_cconj_kl_version_two(space, lam=1, g=None):
+    """Proximal operator factory of the convex conjugate of the KL divergence.
+
+    Function returning the proximal operator of the convex conjugate of the
+    functional F where F is the entropy-type Kullback-Leibler (KL) divergence
+
+        F(x) = sum_i (xln(pos(x)) - xln(g) + g - x)_i + ind_P(x)
+
+    with x and g in X and g non-negative. The indicator function ind_P(x)
+    for the positive elements of x is used to restrict the domain of F such
+    that F is defined over whole X. The non-negativity thresholding pos is
+    used to define F in the real numbers.
+
+    The convex conjugate F^* of F is
+
+        F^*(p) = sum_i (g (exp(p) - 1))_i
+
+    where p is the variable dual to x.
+
+    The proximal operator of the convex conjugate of F is
+
+        prox[sigma * (lam*F)^*](x)_i =
+            x_i - lam W(1/lam sigma g_i exp(1/lam * x_i))
+
+    where sigma is the step size-like parameter, lam the weighting in from of
+    the function F, and W is the Lambert W function (see, for example, the
+    `Wikipedia article <https://en.wikipedia.org/wiki/Lambert_W_function>`_).
+
+    Parameters
+    ----------
+    space : `DiscreteLp` or `ProductSpace` of `DiscreteLp` spaces
+        Space X which is the domain of the functional F
+    g : ``space`` element
+        Data term.
+    lam : positive float
+        Scaling factor.
+
+    Returns
+    -------
+    prox_factory : function
+        Factory for the proximal operator to be initialized.
+
+    Notes
+    -----
+    The functional is not well-defined without a prior g. Hence, if g is
+    omitted this will be interpreted as if g is equal to the one-element.
+
+    Write something about similarities and differences between the two
+    KL-versions?
+    """
+
+    # TODO: Update Notes-section in doc above.
+
+    lam = float(lam)
+
+    if g is not None and g not in space:
+        raise TypeError('{} is not an element of {}'.format(g, space))
+
+    class ProximalCConjKL(Operator):
+
+        """Proximal operator of the convex conjugate of the KL divergence."""
+
+        def __init__(self, sigma):
+            """Initialize a new instance.
+
+            Parameters
+            ----------
+            sigma : positive float
+            """
+            self.sigma = float(sigma)
+            super().__init__(domain=space, range=space, linear=False)
+
+        def _call(self, x, out):
+            """Apply the operator to ``x`` and stores the result in ``out``."""
+
+            if g is None:
+                out.assign(x - lam * scipy.special.lambertw(
+                    1.0 / lam * self.sigma * np.exp(1.0 / lam * x)))
+            else:
+                out.assign(x - lam * scipy.special.lambertw(
+                    1.0 / lam * self.sigma * g * np.exp(1.0 / lam * x)))
 
     return ProximalCConjKL
 
--- a/odl/solvers/advanced/proximal_operators.py
+++ b/odl/solvers/advanced/proximal_operators.py
@@ -46,7 +46,7 @@ __all__ = ('combine_proximals', 'proximal_cconj', 'proximal_translation',
            'proximal_l1', 'proximal_cconj_l1',
            'proximal_l2', 'proximal_cconj_l2',
            'proximal_l2_squared', 'proximal_cconj_l2_squared',
-           'proximal_cconj_kl', 'proximal_cconj_kl_version_two')
+           'proximal_cconj_kl', 'proximal_cconj_kl_cross_entropy')
 
 
 # TODO: remove diagonal op once available on master
@@ -1005,7 +1005,7 @@ def proximal_cconj_kl(space, lam=1, g=None):
 
     See Also
     --------
-    proximal_cconj_kl_version_two : proximal for releated functional
+    proximal_cconj_kl_cross_entropy : proximal for releated functional
 
     Notes
     -----
@@ -1053,7 +1053,15 @@ def proximal_cconj_kl(space, lam=1, g=None):
     the converged solution will be non-negative. Non-negative intermediate
     image estimates can be enforced by adding an indicator function ind_P
     the primal objective.
+
+    This functional :math:`F`, described above, is related to the
+    Kullback-Leibler cross entropy functional. The KL cross entropy is the one
+    diescribed in `this Wikipedia article
+    <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_, and
+    the functional :math:`F` is obtained by switching place of the prior and
+    the varialbe in the KL cross entropy functional. See the See Also section.
     """
+
     lam = float(lam)
 
     if g is not None and g not in space:
@@ -1107,11 +1115,12 @@ def proximal_cconj_kl(space, lam=1, g=None):
     return ProximalCConjKL
 
 
-def proximal_cconj_kl_version_two(space, lam=1, g=None):
-    """Convex conjugate proximal factory of the "second kind" KL divergence.
+def proximal_cconj_kl_cross_entropy(space, lam=1, g=None):
+    """Proximal factory of the convex conjugate of cross entropy KL divergence.
 
-    Function returning the proximal operator of the convex conjugate of the
-    functional F where F is the entropy-type Kullback-Leibler (KL) divergence
+    Function returning the proximal facotry of the convex conjugate of the
+    functional F, where F is the corss entorpy Kullback-Leibler (KL)
+    divergence given by
 
         F(x) = sum_i (x_i ln(pos(x_i)) - x_i ln(g_i) + g_i - x_i) + ind_P(x)
 
@@ -1136,7 +1145,7 @@ def proximal_cconj_kl_version_two(space, lam=1, g=None):
 
     See Also
     --------
-    proximal_cconj_kl : proximal for releated functional
+    proximal_cconj_kl : proximal for related functional
 
     Notes
     -----
@@ -1181,24 +1190,26 @@ def proximal_cconj_kl_version_two(space, lam=1, g=None):
     :math:`\\lambda` and :math:`g` are positive, the argument of :math:`W`
     will always be positive.
 
-    Write something about similarities and differences between the two
-    KL-versions?
-
     `Wikipedia article on Kullback Leibler divergence
-    <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_
+    <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_.
+    For further information about the functional, see for example `this article
+    <http://ieeexplore.ieee.org/document/1056144/?arnumber=1056144>`_.
 
+    The KL cross entropy functional :math:`F`, described above, is related to
+    another functional functional also know as KL divergence. This functional
+    is often used as data discrepancy term in inverse problems, when data is
+    corrupted with Poisson noise. This functional is obtained by changing place
+    of the prior and the variable. See the See Also section.
     """
-
-    # TODO: Update Notes-section in doc above.
 
     lam = float(lam)
 
     if g is not None and g not in space:
         raise TypeError('{} is not an element of {}'.format(g, space))
 
-    class ProximalCConjKLSecondKind(Operator):
+    class ProximalCConjKLCrossEntropy(Operator):
 
-        """Proximal operator of convex conjugate of 2nd kind KL divergence."""
+        """Proximal operator of conjugate of cross entropy KL divergence."""
 
         def __init__(self, sigma):
             """Initialize a new instance.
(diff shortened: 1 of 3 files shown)


odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+    """
+
+    # TODO: Update Notes-section in doc above.
+
+    lam = float(lam)
+
+    if g is not None and g not in space:
+        raise TypeError('{} is not an element of {}'.format(g, space))
+
+    class ProximalCConjKL(Operator):
```
aringh: This name should not be the same as in the other implementation (copy-past error)

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +466,76 @@ def test_proximal_convconj_kl_product_space():
...
+    # Apply the proximal operator returning its optimal point
+    prox(x, x_opt)
+
+    # Explicit computation:
+    x_verify = x - lam * lambertw(1.0 / lam * sigma * g * np.exp(
+        1.0 / lam * x))
+
+    assert all_almost_equal(x_opt, x_verify, HIGH_ACC)
+
+    # Numerically solving min_x KL(x | g)
```
aringh: Is this test needed, given that we have the more advanced one a few lines below?

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+            sigma : positive float
+            """
+            self.sigma = float(sigma)
+            super().__init__(domain=space, range=space, linear=False)
+
+        def _call(self, x, out):
+            """Apply the operator to ``x`` and stores the result in ``out``."""
+
+            if g is None:
+                out.assign(x - lam * scipy.special.lambertw(
```
adler-j: Is the choice of branch relevant?

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+                    1.0 / lam * self.sigma * np.exp(1.0 / lam * x)))
```
adler-j: group terms: `(self.sigma / lam)` to reduce number of multiplications

test/solvers/advanced/proximal_operator_test.py
adler-j: In general this file shouldnt have to much solutions happening since it will make the tests slow in the long run.

adler-j:
Should be added to slow tests as well. We have tests for proximals there.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+    -------
+    prox_factory : function
+        Factory for the proximal operator to be initialized.
+
+    Notes
+    -----
+    The functional is not well-defined without a prior g. Hence, if g is
+    omitted this will be interpreted as if g is equal to the one-element.
+
+    Write something about similarities and differences between the two
```
adler-j: Yes, and also make sure to add a `See Also` part

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+    The convex conjugate F^* of F is
+
+        F^*(p) = sum_i (g (exp(p) - 1))_i
+
+    where p is the variable dual to x.
+
+    The proximal operator of the convex conjugate of F is
+
+        prox[sigma * (lam*F)^*](x)_i =
+            x_i - lam W(1/lam sigma g_i exp(1/lam * x_i))
```
adler-j: might as well shorten some things:

`1 /lam * x_i` -> `x_i / lam` etc

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+        F(x) = sum_i (xln(pos(x)) - xln(g) + g - x)_i + ind_P(x)
+
+    with x and g in X and g non-negative. The indicator function ind_P(x)
+    for the positive elements of x is used to restrict the domain of F such
+    that F is defined over whole X. The non-negativity thresholding pos is
+    used to define F in the real numbers.
+
+    The convex conjugate F^* of F is
+
+        F^*(p) = sum_i (g (exp(p) - 1))_i
```
adler-j: prefer moving indices inside, i.e. `sum_i g_i (exp(p_i) - 1)`

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
     return ProximalCConjKL
 
 
+def proximal_cconj_kl_version_two(space, lam=1, g=None):
+    """Proximal operator factory of the convex conjugate of the KL divergence.
+
+    Function returning the proximal operator of the convex conjugate of the
+    functional F where F is the entropy-type Kullback-Leibler (KL) divergence
+
+        F(x) = sum_i (xln(pos(x)) - xln(g) + g - x)_i + ind_P(x)
```
adler-j: prefer moving indices inside and use spacing, i.e. 

`sum_i g_i x_i ln(pos(x_i)) - x_i ln(g_i) + g_i - x_i`

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+        prox[sigma * (lam*F)^*](x)_i =
+            x_i - lam W(1/lam sigma g_i exp(1/lam * x_i))
+
+    where sigma is the step size-like parameter, lam the weighting in from of
+    the function F, and W is the Lambert W function (see, for example, the
+    `Wikipedia article <https://en.wikipedia.org/wiki/Lambert_W_function>`_).
+
+    Parameters
+    ----------
+    space : `DiscreteLp` or `ProductSpace` of `DiscreteLp` spaces
```
adler-j: Why not any `FnBase` space? Also `ProductSpace` will fail since `lambertw` will scream about it.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+        F^*(p) = sum_i (g (exp(p) - 1))_i
+
+    where p is the variable dual to x.
+
+    The proximal operator of the convex conjugate of F is
+
+        prox[sigma * (lam*F)^*](x)_i =
+            x_i - lam W(1/lam sigma g_i exp(1/lam * x_i))
+
+    where sigma is the step size-like parameter, lam the weighting in from of
```
adler-j: We tend to write 

```
... where ``sigma`` is ...
```

to make clear what is a variable

odl/solvers/advanced/proximal_operators.py
kohr-h: I would prefer having just this definition in the top-level docstring and moving the deeper details to the `Notes` section with proper math typesetting. This long section pushes down the programming-wise more important `Parameters` and `Returns` sections too far down.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
     return ProximalCConjKL
 
 
+def proximal_cconj_kl_version_two(space, lam=1, g=None):
```
kohr-h: Would it make sense to speak of "first kind" and "second kind"? Alternatively, one could use the terms "data matching" and "prior matching" with some explanation.
However, I would probably keep the name of the existing KL divergence since it is somewhat the "default" for inverse problems, and use an additional tag for this slightly more exotic version as it is now.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1080,91 @@ def _call(self, x, out):
...
+    """Proximal operator factory of the convex conjugate of the KL divergence.
```
kohr-h: Should also be adapted somewhat.
```
"""Factory for the convex conjugate proximal of the "second kind" KL divergence."""
```
perhaps? Could be too long though. Alternatively "Convex conjugate proximal factory of ..." shortens a bit.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +466,76 @@ def test_proximal_convconj_kl_product_space():
     assert all_almost_equal(x_verify, x_opt)
 
 
+def test_proximal_convconj_kl_version_two():
+    """Proximal factory of convex conjugate for 2nd version KL divergence."""
```
kohr-h: Test for ...

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +466,76 @@ def test_proximal_convconj_kl_product_space():
...
+
+    # Allocate an output vector
+    x_opt = space.element()
+
+    # Apply the proximal operator returning its optimal point
+    prox(x, x_opt)
+
+    # Explicit computation:
+    x_verify = x - lam * lambertw(1.0 / lam * sigma * g * np.exp(
+        1.0 / lam * x))
```
kohr-h: All comments up to here can be removed, not necessary in tests. On the other hand, the variable names should tell the story. `x_verify` goes in the right direction, I would deem `true_prox_val` or `expected_...` slightly better though.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +466,76 @@ def test_proximal_convconj_kl_product_space():
...
+
+    # Element in image space where the proximal operator is evaluated
+    x = space.element(np.arange(-5, 5))
+
+    # Data
+    g = space.element(np.arange(10, 0, -1))
+
+    # Factory function returning the proximal operator
+    lam = 2
+    prox_factory = proximal_cconj_kl_version_two(space, lam=lam, g=g)
```
kohr-h: `lam=2` directly, don't use a variable.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +466,76 @@ def test_proximal_convconj_kl_product_space():
...
+    sigma = 0.25
+    prox = prox_factory(sigma)
+
+    assert isinstance(prox, odl.Operator)
+
+    # Allocate an output vector
+    x_opt = space.element()
+
+    # Apply the proximal operator returning its optimal point
+    prox(x, x_opt)
```
kohr-h: I'd write `prox_val = prox(x)` and not use `out`. If you specifically want to test that feature, make it a separate test (not in a new test function).

test/solvers/advanced/proximal_operator_test.py
kohr-h: Agree. What you could do, though, is to move this test to the `largescale` suite and make the problem bigger.

odl/solvers/advanced/proximal_operators.py
aringh: I thought that the different branches stemmed from the ambiguity of the complex logarithm, so for real data that would have been fine. However this seems to NOT be the case...

However, in [the wikipedia article](https://en.wikipedia.org/wiki/Lambert_W_function), second paragraph in the introduction, it seems like it is only in the interval `-1/e < x < 0` where there are two branches. This should not cause a problem here, since `e^(x) > 0`, `lam > 0` (at least in normal cases...), `sigma > 0` and `g > 0`.

odl/solvers/advanced/proximal_operators.py
aringh: This is copy-past from the other KL-proximal. Change to `LinearSpace`?

Also: any idea of how to handle `ProductSpace`?

odl/solvers/advanced/proximal_operators.py
aringh: The other one is only default in data-matching term... this version has been used as regularizing functional in different settings under the name maximum entropy or cross entropy:

- http://ieeexplore.ieee.org/document/1056144/?arnumber=1056144&tag=1
- http://mnras.oxfordjournals.org/content/211/1/111.short
- http://ieeexplore.ieee.org/document/1456693/?arnumber=1456693
- https://www.osapublishing.org/josa/abstract.cfm?uri=josa-62-4-511
- http://ieeexplore.ieee.org/document/771159/?arnumber=771159
- http://iopscience.iop.org/article/10.1088/0266-5611/7/6/004/meta

and this is only the tip of the iceberg.

With that said, we can still settle for keeping the name of the other and giving a new name to this one. `cross_entorpy`?

odl/solvers/advanced/proximal_operators.py
adler-j: This would be a general change to the whole `proximals` module. Basically retype this mathy stuff in typeset mathematics under the notes section. I made an issue for it #562.

odl/solvers/advanced/proximal_operators.py
adler-j: `LinearSpace` will fail. We need a space whose vectors implement `__array__` and `__array_wrap__`, and currently that is only the `FnBase` spaces (Note that `Discretization` inherits from `FnBase`). Productspaces could be handled by us solving #83 at some point.

odl/solvers/advanced/proximal_operators.py
adler-j: Perhaps add a `#` comment about this then, or perhaps even a comment to the `Notes` section. We are bound to forget this later.

test/solvers/advanced/proximal_operator_test.py
adler-j: If that is done I suggest generalizing it with a `fixture` so that we can test the others as well. Do however note that this test may become circular since this depends on `douglas_rachford_pd` working as expected.

odl/solvers/advanced/proximal_operators.py
aringh: So doc should say `FnBase` for now?

test/solvers/advanced/proximal_operator_test.py
aringh: I'm gonna keep this, since it is used in the explicit computations to verify the evaluation further down.

odl/solvers/advanced/proximal_operators.py
adler-j: Yes

aringh:
Updated according to comments. Left to do

- Resolve naming. Calling the new one `cross_entropy` or `kl_cross_entropy`?
- Update doc with interlinking comments. I will do this when the name issue is resolved.

aringh:
The naming issue remains: what to call it? And what to call the other one? I vote for calling this one `kl_cross_entropy`. Any other suggestions?

After naming, is this one ok to merge?

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1107,125 @@ def _call(self, x, out):
...
+
+    Write something about similarities and differences between the two
+    KL-versions?
+
+    `Wikipedia article on Kullback Leibler divergence
+    <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_
+
+    """
+
+    # TODO: Update Notes-section in doc above.
```
adler-j: I guess this should be done.

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1107,125 @@ def _call(self, x, out):
...
+    Lambert W function (see, for example, the
+    `Wikipedia article <https://en.wikipedia.org/wiki/Lambert_W_function>`_).
+
+    For real-valued input x, the Lambert :math:`W` function is defined only for
+    :math:`x \\geq -1/e`, and it has two branches for values
+    :math:`-1/e \\leq x < 0`. However, for inteneded use-cases, where
+    :math:`\\lambda` and :math:`g` are positive, the argument of :math:`W`
+    will always be positive.
+
+    Write something about similarities and differences between the two
```
adler-j: Yes

odl/solvers/advanced/proximal_operators.py
```diff
@@ -1079,6 +1107,125 @@ def _call(self, x, out):
...
+
+            if g is None:
+                # If g is None, it is taken as the one element
+                # Different branches of lambertw is not an issue, see Notes
+                out.assign(x - lam * scipy.special.lambertw(
+                    (self.sigma / lam) * np.exp(x / lam)))
+            else:
+                # Different branches of lambertw is not an issue, see Notes
+                out.assign(x - lam * scipy.special.lambertw(
+                    ((self.sigma * g) / lam) * np.exp(x / lam)))
```
adler-j: `(self.sigma / lam) * g`

adler-j: Might as well do lincomb here for minor efficiency gains.

```
out.lincomb(1, x,
            -lam, scipy.special.lambertw(
                (self.sigma / lam) * g * np.exp(x / lam))
```

test/largescale/solvers/advanced/proximal_operator_slow_test.py
```diff
@@ -31,9 +31,11 @@
 from odl.solvers.advanced.proximal_operators import (
     proximal_l1, proximal_cconj_l1,
     proximal_l2, proximal_cconj_l2,
-    proximal_l2_squared, proximal_cconj_l2_squared)
-from odl.util.testutils import noise_element
+    proximal_l2_squared, proximal_cconj_l2_squared,
+    proximal_cconj_kl, proximal_cconj_kl_version_two)
+from odl.util.testutils import (noise_element, all_almost_equal)
 
+from scipy.special import lambertw
```
adler-j: simply `import scipy` and then do `scipy.special.lambertw` below. More clear where it comes form.

test/largescale/solvers/advanced/proximal_operator_slow_test.py
```diff
@@ -183,5 +221,48 @@ def test_proximal_defintion(proximal_and_function):
 
         assert f_prox_x <= f_y
 
+
+def test_proximal_convconj_kl_version_two_solving_opt_problem():
```
adler-j: Nice addition!

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +465,42 @@ def test_proximal_convconj_kl_product_space():
...
+
+    # Explicit computation:
+    x_verify = x - lam * lambertw(1.0 / lam * sigma * g * np.exp(
+        1.0 / lam * x))
+
+    assert all_almost_equal(prox_val, x_verify, HIGH_ACC)
+
+    # Test inplace evaluation
+    x_opt = space.element()
+    prox(x, x_opt)
```
adler-j: use `out=x_opt` to make it clear what happens.

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +465,42 @@ def test_proximal_convconj_kl_product_space():
...
+    # Initialize the proximal operator of F^*
+    sigma = 0.25
+    prox = prox_factory(sigma)
+
+    assert isinstance(prox, odl.Operator)
+
+    prox_val = prox(x)
+
+    # Explicit computation:
+    x_verify = x - lam * lambertw(1.0 / lam * sigma * g * np.exp(
```
adler-j:
```python
x_verify = x - lam * lambertw(
    1.0 / lam * sigma * g * np.exp(1.0 / lam * x))
```

test/solvers/advanced/proximal_operator_test.py
```diff
@@ -464,5 +465,42 @@ def test_proximal_convconj_kl_product_space():
...
 
 
+def test_proximal_convconj_kl_version_two():
+    """Test for proximal of the conjugate of the "2nd" kind KL divergence."""
+
+    # Image space
+    space = odl.uniform_discr(0, 1, 10)
+
+    # Element in image space where the proximal operator is evaluated
+    x = space.element(np.arange(-5, 5))
```
adler-j: move this down to where it is evaluated

adler-j:
Gave some nit picking comments but I guess input is always good. Once done, do the rename (and make sure you also rename tests etc) and you can merge.

aringh:
Updated according to comments, and renamed to `kl_cross_entropy`. Merging after the checks are done

kohr-h:
@aringh Please write a release note on this.

Issue #563, opened 2016-09-08
Solvers like CG and Landweber assumes the range is a linear space


aringh:
Both `conjugate_gradient_normal` and `landweber` assumes the range of the operator is a `LinearSpace`. For `Functional`s this will typically be `RealNumbers` and they both give errors:
```
>>> odl.solvers.conjugate_gradient_normal(my_func, x, my_func.range.element(0), 100)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[...]/odl/solvers/iterative/iterative.py", line 266, in conjugate_gradient_normal
    d.lincomb(1, rhs, -1, d)               # d = rhs - A x
AttributeError: 'float' object has no attribute 'lincomb'
```
and
```
>>> odl.solvers.landweber(my_func, x, my_func.range.element(0), 100)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[...]/odl/solvers/iterative/iterative.py", line 114, in landweber
    op(x, out=tmp_ran)
  File "/home/aringh/git/odl/odl/operator/operator.py", line 688, in __call__
    raise TypeError('`out` parameter cannot be used '
TypeError: `out` parameter cannot be used when range is a field
```
The bugs/errors were found in connection to the functional-branch #246 #498, but these errors would arise also for "normal" `Operator`s if the `range` is not a linear vector space.

adler-j:
This is true for all of our solvers. At what point would you want to use these with non-spaces as ranges?

aringh:
Finding a zero for a functional? That was the test I was trying to implement as an example to show that "everything that works for operators also works for functionals".

adler-j:
The whole issue here is that `float`s in python are immutable. Hence any syntax like `out=out` will fail miserably.

There are basically two approaches to solving this

* Extend `Field` significantly so that `Field.element` returns an element of some type `ODLScalar` (name tbd) that is actually mutable, but that also implements e.g. `__float__` and otherwise masquerades as a usual float. With this change we could make `Field` a `LinearSpace` itself, and all this would be solved.

* Change the solvers so they have options (somehow) for handling non-mutable data types.

adler-j:
Added a suggested solution to this according to option 1 above. Spoiler: it aint very pretty, but it (should) work.

Please go ahead and try it. I suppose @kohr-h will have opinions on if this is a good idea.

kohr-h:
Right now I don't have any strong opinions about this solution :-)

[...]

PR #565, opened 2016-09-08
WIP: example showing Field as a LinearSpace added

adler-j:
Example showing how #563 could be solved. If you want to try this @aringh, simply cherry-pick this pull and try using the solver as you did before. It should work.


adler-j:
This won't go in as is, closing.

PR #574, opened 2016-09-14
ENH: Allow show with coord as subset of axis

adler-j:
Slight functionality extension to show, can now properly display subsets of an image.

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -727,6 +727,9 @@ class DiscreteLpElement(DiscretizedSpaceElement):
             where ``None`` means all values along that dimension. For
             example, ``[None, None, 0.5]`` shows all values in the first
             two dimensions, with the third coordinate equal to 0.5.
+            If a sequence is provided, it specifies the minimum and maximum
+            point to be shown, i.e. ``[None, [0, 1]]`` shows all of the
+            first axis and values between 0 and 1 in the second.
             This option is mutually exclusive to ``indices``.
 
         indices : index expression, optional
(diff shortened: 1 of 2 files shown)


odl/discr/lp_discr.py
```diff
@@ -775,19 +778,43 @@ def show(self, title=None, method='', coords=None, indices=None,
...
             for axis, (xmin, xmax, n, coord) in enumerate(
                     zip(interv.min_pt, interv.max_pt, shape, coords)):
-                if coord is None:
-                    indices += [slice(None)]
+                try:
+                    coord_minp, coord_maxp = coord
+                except TypeError:
+                    coord_minp = coord_maxp = coord
+
+                # Validate input
```
kohr-h: Cast `coord_minp` and `coord_maxp` already here to float. Otherwise, the comparison may raise and the error is less conclusive.

adler-j:
Fixed comment, merge?

odl/discr/lp_discr.py
```diff
@@ -775,19 +778,47 @@ def show(self, title=None, method='', coords=None, indices=None,
...
                         raise ValueError('in axis {}: coordinate {} not in '
                                          'the valid range {}'
                                          ''.format(axis, coord, [xmin, xmax]))
 
-                    if xmin == xmax:
-                        indices += [0]
+                if xmin == xmax:  # trivial cases
+                    indices += [0]
+                elif coord_minp is not None and coord_minp == coord_maxp:
+                    normalized_pos = ((float(coord_minp) - xmin) /
```
kohr-h: Minor thing, you can remove the cast here and further down also. Merge after that change.

kohr-h:
@adler-j  Please write a release note on this.

PR #576, opened 2016-09-14
WIP: add ufunc operators

adler-j:
Initial shell for ufunc operators. Works for all ufuncs that we have, but currently lacks derivatives etc. Not quite sure how to add that in a nice manner, but worst case we could monkey patch it individually. I guess we only really need derivative for the following, which should be "doable".
- sin
- sinh
- cos
- cosh
- tan
- tanh
- negative
- sqrt
- square
- log
- exp
- arccos
- arccosh
- arcsin
- arcsinh
- arctan
- arctanh
- absolute

Edit:

Something that is missing here is the functional equivalents of these, basically `R -> R`. I suppose they would be useful in some cases.

### DIFF
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -411,6 +411,10 @@ class ProductSpace(LinearSpace):
         if inp in self:
             return inp
 
+        if len(inp) != len(self):
+            raise ValueError('length of `inp` {} does not match lenght of '
+                             'space {}'.format(len(inp), len(self)))
+
         if (all(isinstance(v, LinearSpaceElement) and v.space == space
                 for v, space in zip(inp, self.spaces))):
             parts = list(inp)
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -350,15 +350,16 @@ class ProductSpace(LinearSpace):
 
         Raises
         ------
-        AttributeError : If any of the subspaces does not implement `dtype` or
-            if the dtype of the subspaces does not match.
+        AttributeError
+            If any of the subspaces does not implement `dtype` or if the dtype
+            of the subspaces does not match.
         """
         dtypes = [space.dtype for space in self.spaces]
 
         if all(dtype == dtypes[0] for dtype in dtypes):
             return dtypes[0]
         else:
-            raise AttributeError("`dtype`'s of subspaces is not equal")
+            raise AttributeError("`dtype`'s of subspaces not equal")
 
     def element(self, inp=None, cast=True):
         """Create an element in the product space.
@@ -412,7 +413,7 @@ class ProductSpace(LinearSpace):
             return inp
 
         if len(inp) != len(self):
-            raise ValueError('length of `inp` {} does not match lenght of '
+            raise ValueError('length of `inp` {} does not match length of '
                              'space {}'.format(len(inp), len(self)))
 
         if (all(isinstance(v, LinearSpaceElement) and v.space == space
--- a/odl/ufunc_ops/ufunc_ops.py
+++ b/odl/ufunc_ops/ufunc_ops.py
@@ -78,12 +78,24 @@ def gradient_factory(name):
     elif name == 'log':
         def gradient(self):
             """Return the gradient operator."""
-            return FunctionalQuotient(ConstantFunctional(self.domain, 1.0),
-                                      IdentityFunctional(self.domain))
+            return reciprocal(self.domain)
     elif name == 'exp':
         def gradient(self):
             """Return the gradient operator."""
             return self
+    elif name == 'reciprocal':
+        def gradient(self):
+            """Return the gradient operator."""
+            return FunctionalQuotient(ConstantFunctional(self.domain, -1.0),
+                                      square(self.domain))
+    elif name == 'sinh':
+        def gradient(self):
+            """Return the gradient operator."""
+            return cosh(self.domain)
+    elif name == 'cosh':
+        def gradient(self):
+            """Return the gradient operator."""
+            return sinh(self.domain)
     else:
         # Fallback to default
         gradient = Functional.gradient
@@ -125,6 +137,20 @@ def derivative_factory(name):
         def derivative(self, point):
             """Return the derivative operator."""
             return MultiplyOperator(self(point))
+    elif name == 'reciprocal':
+        def derivative(self, point):
+            """Return the derivative operator."""
+            point = self.domain.element(point)
+            return MultiplyOperator(-self(point) ** 2)
+    elif name == 'sinh':
+        def derivative(self, point):
+            """Return the derivative operator."""
+            point = self.domain.element(point)
+            return MultiplyOperator(cosh(self.domain)(point))
+    elif name == 'cosh':
+        def derivative(self, point):
+            """Return the derivative operator."""
+            return MultiplyOperator(sinh(self.domain)(point))
     else:
         # Fallback to default
         derivative = Operator.derivative
@@ -166,7 +192,7 @@ def ufunc_class_factory(name, nargin, nargout, docstring):
         Operator.__init__(self, domain=domain, range=range, linear=linear)
 
     def _call(self, x, out=None):
-        """return ``self(x)``."""
+        """Return ``self(x)``."""
         if out is None:
             if nargin == 1:
                 return getattr(x.ufunc, name)()
@@ -243,7 +269,7 @@ def ufunc_functional_factory(name, nargin, nargout, docstring):
         Functional.__init__(self, space=field, linear=linear)
 
     def _call(self, x):
-        """return ``self(x)``."""
+        """Return ``self(x)``."""
         if nargin == 1:
             return getattr(np, name)(x)
         else:
@@ -296,14 +322,12 @@ Examples
 RAW_UFUNC_FACTORY_FUNCTIONAL_DOCSTRING = """
 Create functional with domain/range as real numbers:
 
->>> import odl
 >>> func = odl.ufunc_ops.{name}()
 """
 
 RAW_UFUNC_FACTORY_OPERATOR_DOCSTRING = """
 Create operator that acts pointwise on a `FnBase`
 
->>> import odl
 >>> space = odl.rn(3)
 >>> op = odl.ufunc_ops.{name}(space)
 """
--- a/odl/ufunc_ops/ufunc_ops.py
+++ b/odl/ufunc_ops/ufunc_ops.py
@@ -23,11 +23,14 @@ from future import standard_library
 standard_library.install_aliases()
 
 import numpy as np
-from odl.set.space import LinearSpace
+from odl.set import LinearSpace, RealNumbers, Field
 from odl.space import ProductSpace, fn
+from odl.space.base_ntuples import FnBase
 from odl.operator import Operator, MultiplyOperator
 from odl.util.utility import is_int_dtype
 from odl.util.ufuncs import UFUNCS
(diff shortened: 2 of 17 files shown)


adler-j:
Tests now run properly and we have derivative for some of the basic ones, as well as proper flagging of linear ufuncs. Ready for a review.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+from odl.util.utility import is_int_dtype
+from odl.util.ufuncs import UFUNCS
+
+__all__ = ()
+
+
+def _is_integer_only_ufunc(name):
+    return 'shift' in name or 'bitwise' in name or name == 'invert'
+
+LINEAR_UFUNCS = ['conj', 'negative', 'rad2deg', 'deg2rad', 'add', 'subtract']
```
adler-j: This list needs to be verified, perhaps I missed some.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+--------
+>>> import odl
+>>> space = odl.{space}
+>>> op = {name}(space)
+>>> print(op({arg}))
+{result}
+"""
+
+
+class UfuncOperator(Operator):
```
adler-j: I had more hopes for this than we seem to need. Perhaps remove it and let the ops directly subclass `Operator`.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+
+    """Base class for all ufunc operators."""
+
+
+def derivative_factory(name):
```
adler-j: The style of this is a freaking monster, but I dont know how to do it better.

odl/space/pspace.py
```diff
@@ -339,6 +339,24 @@ def is_weighted(self):
         """Return ``True`` if weighting is not `ProductSpaceNoWeighting`."""
         return not isinstance(self.weighting, ProductSpaceNoWeighting)
 
+    @property
+    def dtype(self):
```
adler-j: Was needed to find dtype for ufuncs.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+        return derivative
+    elif name == 'tan':
+        def derivative(self, point):
+            """Return the derivative operator."""
+            return MultiplyOperator(1 + self(point) ** 2)
+        return derivative
+    elif name == 'sqrt':
+        def derivative(self, point):
+            """Return the derivative operator."""
+            return MultiplyOperator(0.5 / self(point))
```
aringh: Do we want tests for `point > 0`, or is that handled elsewhere?

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+        def derivative(self, point):
+            """Return the derivative operator."""
+            point = self.domain.element(point)
+            return MultiplyOperator(2.0 * point)
+        return derivative
+    elif name == 'log':
+        def derivative(self, point):
+            """Return the derivative operator."""
+            point = self.domain.element(point)
+            return MultiplyOperator(1.0 / point)
```
aringh: Same as above

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+        return derivative
+    else:
+        # Fallback to default
+        return UfuncOperator.derivative
+
+
+def ufunc_class_factory(name, nargin, nargout, docstring):
+    """Create a UfuncOperator from a given specification."""
+
+    assert 0 <= nargin <= 2
```
aringh: What is `nargin` suppose to contain, since it is only allowed to be `0 <=` and `<= 2`?

odl/ufunc_ops/ufunc_ops.py
aringh: Looks correct to me

odl/ufunc_ops/ufunc_ops.py
aringh: This one contains nothing different from an ordinary `Operator`, right? But on the other hand maybe it would be interesting at some point to know that it is a ``UfuncOperator`? I think that you @adler-j and @kohr-h are more suited to sort that out.

odl/ufunc_ops/ufunc_ops.py
adler-j: this is the range of values currently supported and currently available in numpy. I simply added it here in case numpy adds more ufuncs in the future so we don't forget to update it.

odl/ufunc_ops/ufunc_ops.py
adler-j: See above

odl/ufunc_ops/ufunc_ops.py
kohr-h: Well, it IS ugly, but nobody is ever going to see it, so I think it's okay :-)

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, space):
+        """Initialize an instance."""
+        if not isinstance(space, LinearSpace):
+            raise TypeError('`space` {!r} not a `LinearSpace`'.format(space))
+
+        if _is_integer_only_ufunc(name) and not is_int_dtype(space.dtype):
+            raise ValueError('`space` {!r} not a `LinearSpace`'.format(space))
```
kohr-h: wrong error message

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+
+        linear = name in LINEAR_UFUNCS
+        UfuncOperator.__init__(self, domain=domain, range=range, linear=linear)
+
+    def _call(self, x, out=None):
+        """return ``self(x)``."""
+        if out is None:
+            if nargin == 1:
+                return getattr(x.ufunc, name)()
+            elif nargin == 2:
```
kohr-h: Why not `else`? The implementation will extend to that case, provided Numpy implements it this way, too. Alternatively, `nargin == 2` should be implemented explicitly and the `else` case raise a `NotImplementedError`. That will give us a chance to catch an unsupported upstream implementation.

odl/ufunc_ops/ufunc_ops.py
kohr-h: Doesn't add anything right now. Once we need a specific behavior of such a guy, we can still add it as a layer.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,196 @@
...
+                  "_call": _call,
+                  "derivative": derivative_factory(name),
+                  "__repr__": __repr__,
+                  "__doc__": full_docstring}
+
+    return type(name, (UfuncOperator,), attributes)
+
+# Create an operator for each ufunc
+for name, nargin, nargout, docstring in UFUNCS:
+    globals()[name] = ufunc_class_factory(name, nargin, nargout, docstring)
```
kohr-h: Hacking the namespace, eh? ;-)

adler-j:
Updated with UFunc functionals and exampels, needs final review.

adler-j:
Will rebase and fix this for a review.

adler-j:
This is now up to date and kinda "final". Not sure if i really want to write actual tests for this or if we can let the doctests do the work but I'm leaning towards just doctests.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,359 @@
...
+        if nargout == 1:
+            range = space
+        else:
+            range = ProductSpace(space, nargout)
+
+        linear = name in LINEAR_UFUNCS
+        Operator.__init__(self, domain=domain, range=range, linear=linear)
+
+    def _call(self, x, out=None):
+        """return ``self(x)``."""
```
kohr-h: Return

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,359 @@
...
+            return ScalingFunctional(self.domain, 2.0)
+    elif name == 'log':
+        def gradient(self):
+            """Return the gradient operator."""
+            return FunctionalQuotient(ConstantFunctional(self.domain, 1.0),
+                                      IdentityFunctional(self.domain))
+    elif name == 'exp':
+        def gradient(self):
+            """Return the gradient operator."""
+            return self
```
kohr-h: Cool, I like the simple gradient formulas.
Further simple additions are `cosh` and `sinh` since they are each other's derivatives. Also `tanh` with derivative `1/tanh^2`.
I think also `reciprocal` would be very useful, you could use it for the `tanh` thing.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,359 @@
...
+Examples
+--------
+{operator_example}
+{functional_example}
+"""
+
+RAW_UFUNC_FACTORY_FUNCTIONAL_DOCSTRING = """
+Create functional with domain/range as real numbers:
+
+>>> import odl
```
kohr-h: Not needed?

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -0,0 +1,359 @@
...
+>>> func = odl.ufunc_ops.{name}()
+"""
+
+RAW_UFUNC_FACTORY_OPERATOR_DOCSTRING = """
+Create operator that acts pointwise on a `FnBase`
+
+>>> import odl
```
kohr-h: Not needed?

kohr-h:
Ah, and I guess the failing tests are from the time when Travis was down.

adler-j:
I'll try to dedicate some time this week to getting the review comments to this fixed so we may get it in.

odl/ufunc_ops/ufunc_ops.py
adler-j: Will do, especial :+1: on `reciprocal`

adler-j:
Updated according to comments, rebased and fixed the failing tests. Merge after CI.

kohr-h:
Oh dear, printing stuff seems to be dangerous wrt Py2/3..

kohr-h:
> Well how the f am I supposed to solve that in python 3? Doesnt print call str anymore? In python 2 it works perfectly.

It does, but I think the default float formatting is different (less digits)... What happens if you use `repr`, i.e., omit the `print`?

adler-j:
> It does, but I think the default float formatting is different (less digits)

How can it be different? `__str__` doesnt take any arguments

> What happens if you use repr, i.e., omit the print?

Rounding errors

adler-j:
Frankly don't know how to solve this and I don't have the time to debug with a 10 min debug cycle. Putting this of for later, worst case I'll change the implementation.

kohr-h:
> It does, but I think the default float formatting is different (less digits)
> 
> How can it be different? __str__ doesnt take any arguments

My guess is that the builtin `float.__str__` implementation has changed.

Probably the easiest fix is to compare to an expected value instead of printing, if that's possible in an automated fashion, probably against the value of the raw ufunc?. That could also improve readability because it's more explicit. But I do agree it's annoying.

adler-j:
Last try, could be due to the return type being numpy.float64, which apparently has another __str__ than float does (but only in python 3, to make matters more fun).

Issue #577, opened 2016-09-15
Readthedoc build failed: No module named 'scipy.special'


aringh:
If I knew how to fix it, I would just do it. Unfortunately I don't, and hence I post and issue. Here is the build log: http://readthedocs.org/projects/odl/builds/4408684/

adler-j:
Two ways to solve this. 

* Either it needs to be added to the list of mock modules in the doc conf:
       https://github.com/odlgroup/odl/blob/38c4f90fbeafce850db3833a8209f9da90386c98/doc/source/conf.py#L50-L55

* Or you remove all `import scipy.special` calls and use `import scipy` and later do `scipy.special.???`. I would recommend this approach.

aringh:
Closed via #579

PR #578, opened 2016-09-15
BUG: import scipy instead of scipy.special.

aringh:
The later caused Readthedoc to not build. See #577

### DIFF
--- a/odl/solvers/advanced/proximal_operators.py
+++ b/odl/solvers/advanced/proximal_operators.py
@@ -31,6 +31,7 @@ standard_library.install_aliases()
 from builtins import super
 
 import numpy as np
+import scipy.special
 
 from odl.operator import (Operator, IdentityOperator, ScalingOperator,
                           ConstantOperator, ResidualOperator, DiagonalOperator)
@@ -45,7 +46,7 @@ __all__ = ('combine_proximals', 'proximal_cconj', 'proximal_translation',
            'proximal_l1', 'proximal_cconj_l1',
            'proximal_l2', 'proximal_cconj_l2',
            'proximal_l2_squared', 'proximal_cconj_l2_squared',
-           'proximal_cconj_kl')
+           'proximal_cconj_kl', 'proximal_cconj_kl_version_two')
 
 
 # TODO: remove diagonal op once available on master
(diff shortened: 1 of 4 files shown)


odl/solvers/advanced/proximal_operators.py
```diff
@@ -31,7 +31,7 @@
 from builtins import super
 
 import numpy as np
-import scipy.special
+import scipy
```
adler-j: Did you not use this one?

aringh: I think I used `scipy.special.lambertw`further down anyway. But now that the build fails I will have to double check...

aringh: The failure is due to `AttributeError: module 'scipy' has no attribute 'special'`. But the tests run on my machine... and in all places I do `scipy.special.lambertw`

aringh: Also when I switch machine all tests pass locally. I don't understand why Travis fails?

adler-j: What version of scipy are you using?

Scipy uses lazy imports (you may need to google what it means), and thus *if* `special` is imported lazily, it will not be available via `scipy.special.lambertw`, and you need to do `import scipy.special`. There is some text on this [here](http://docs.scipy.org/doc/scipy/reference/api.html), but it's rather vague.

adler-j: Anyway, it seems you need to add `scipy.special` to the mock list. But you should still do `import scipy.special` up top, and then call `scipy.special.lambertw` later on, since this is more clear to readers. 1000 lines into a file it's hard to remember what was imported.

aringh:
This is weird...

I have changed back in `proximal_operators` to `from scipy.special import lambertw`, which is what causes the Readthedocs build to fail. Then tests run, also in Travis.

However if I only do `import scipy` in `proximal_operators`, then Travis fails with the error `AttributeError: module 'scipy' has no attribute 'special'` on line 1235 (or 1230) in `proximal_operators`, when doing the call `scipy.special.lambertw`. However, that is the exact same call  which is done in both test (both the slow and `proximal_operator_test`) in both cases.

Moreover, the test crash is only Travis,  on the two local machines I have tried any of the combination works.

Any ideas @adler-j?

adler-j:
Looks very weird, but simply do my suggestion in the comment and you should be fine.

aringh:
For anyone how finds this, it was replaced by #579

PR #579, opened 2016-09-15
BUG: fix import of scipy.special and add it to doc/conf.py. Importing …

aringh:
…scipy.special caused Readthedoc to not build.


aringh:
After the slight mess on #578, I simply closed that one and made a new branch.

Issue #358, opened 2016-04-22
Cleanup top level namespace


adler-j:
Many functions/classes such as `odl.weighted_norm`, `odl.OpDomainError` and `odl.FnMatrixWeighting` should probably not be exposed in the main interface, but left in some other (deeper) namespace.

adler-j:
I've mad a list of all operators in the `odl` namespace, and I'll put a checkmark in those I feel should be moved out of the top level namespace:

- [ ] BroadcastOperator
- [ ] CartesianProduct
- [ ] ComplexNumbers
- [ ] ComponentProjection
- [ ] ComponentProjectionAdjoint
- [ ] ConstantOperator
- [ ] Cuboid
- [ ] DiagonalOperator
- [ ] DiscreteLp
- [ ] DiscreteLpVector
- [ ] DiscretizedSet
- [ ] DiscretizedSetVector
- [ ] DiscretizedSpace
- [ ] DiscretizedSpaceVector
- [ ] DistOperator
- [ ] Divergence
- [ ] EmptySet
- [ ] FN_IMPLS
- [ ] Field
- [ ] FunctionSet
- [ ] FunctionSetMapping
- [ ] FunctionSetVector
- [ ] FunctionSpace
- [ ] FunctionSpaceVector
- [ ] FunctionalLeftVectorMult
- [ ] Gradient
- [ ] IdentityOperator
- [ ] InnerProductOperator
- [ ] Integers
- [ ] Interval
- [ ] IntervalProd
- [ ] Laplacian
- [ ] LinCombOperator
- [ ] LinearInterpolation
- [ ] LinearSpace
- [ ] LinearSpaceTypeError
- [ ] LinearSpaceVector
- [ ] MatVecOperator
- [ ] MultiplyOperator
- [ ] NTUPLES_IMPLS
- [ ] NearestInterpolation
- [ ] NormOperator
- [ ] NumpyFn
- [ ] NumpyFnConstWeighting
- [ ] NumpyFnMatrixWeighting
- [ ] NumpyFnVector
- [ ] NumpyFnVectorWeighting
- [ ] NumpyNtuples
- [ ] NumpyNtuplesVector
- [ ] OpDomainError
- [ ] OpNotImplementedError
- [ ] OpRangeError
- [ ] OpTypeError
- [ ] Operator
- [ ] OperatorComp
- [ ] OperatorLeftScalarMult
- [ ] OperatorLeftVectorMult
- [ ] OperatorPointwiseProduct
[...]

PR #587, opened 2016-09-19
Issue 569: Migrate solvers to use functionals

adler-j:
Migraded douglas rachford to use functionals as an example. Obviously WIP, but @ozanoktem should have a look. This can work as a example branch of this.

This version removes all proximals etc, but does not notably change the format of the solvers (you do not "give it a optimization problem").

The current functionals do not work with complex domains, which means that I cannot migrate the MRI example properly.

Two of the commits have been moved to other pull requests so we can clean this out.

##### Ongoing issues:

- [ ] How should we name the parameters previously called ``prox_cc_l`` and similar. I simply changed them  to `l`, but as @kohr-h pointed out this is not a long term solution.


adler-j:
Obviously WIP, but @ozanoktem should have a look. This can work as a example branch of this.

This version removes all proximals etc, but does not notably change the format of the solvers (you do not "give it a optimization problem").

examples/solvers/douglas_rachford_primal_dual_heron.py
```diff
@@ -47,23 +47,22 @@
...
 # g is the distance function. Here, the l2 distance
-prox_cc_g = [odl.solvers.proximal_cconj_l2(space)] * len(rectangles)
+g = [odl.solvers.L2Norm(space)] * len(rectangles)
 
 # l are the indicator functions on the rectangles.
-prox_l = [odl.solvers.proximal_box_constraint(space, minp, maxp)
-          for minp, maxp in rectangles]
-# We want the proximal of the convex conjugate, so we need to convert to that.
-prox_cc_l = [odl.solvers.proximal_cconj(prox) for prox in prox_l]
+l = [odl.solvers.IndicatorBox(space, minp, maxp) for minp, maxp in rectangles]
```
kohr-h: `l` is a really uninspired name for a collection of functionals.

examples/solvers/douglas_rachford_primal_dual_heron.py
```diff
@@ -47,23 +47,22 @@
...
 sigma = [1.0] * len(rectangles)
 
+
 # The lam parameter can be used to accelerate the convergence rate
-lam = lambda n: 1.0 + 1.0 / (n + 1)
+def lam(n):
+    return 1.0 + 1.0 / (n + 1)
 
 
 def print_objective(x):
```
kohr-h: Next line should say "print it"

examples/solvers/douglas_rachford_primal_dual_heron.py
```diff
@@ -76,9 +75,9 @@ def print_objective(x):
 
 # Solve
 x = space.zero()
-odl.solvers.douglas_rachford_pd(x, prox_f, prox_cc_g, lin_ops,
+odl.solvers.douglas_rachford_pd(x, f, g, lin_ops,
                                 tau=tau, sigma=sigma, niter=20, lam=lam,
-                                callback=print_objective, prox_cc_l=prox_cc_l)
+                                callback=print_objective, l=l)
```
kohr-h: When I see this, I also don't like the `l` parameter of the solver. But it's kind of hard to come up with a better name. Something more telling perhaps, like `inf_conv` or, slightly longer, `inf_conv_funcs`?

odl/solvers/advanced/douglas_rachford.py
```diff
@@ -51,27 +51,27 @@ def douglas_rachford_pd(x, prox_f, prox_cc_g, L, tau, sigma, niter,
...
     x : `LinearSpaceElement`
         Initial point, updated in-place.
-    prox_f : `callable`
+    f : `Functional`
         `proximal factory` for the function ``f``.
-    prox_cc_g : `sequence` of `callable`'s
-        Sequence of `proximal factory` for the convex conjuates of the
-        functions ``g_i``.
+    g : `sequence` of `Functional`'s
+        Sequence of of the functions ``g_i``. Needs to have
```
kohr-h: Should say `functionals` rather than `functions` (across the board I guess).

odl/solvers/advanced/douglas_rachford.py
```diff
@@ -51,27 +51,27 @@ def douglas_rachford_pd(x, prox_f, prox_cc_g, L, tau, sigma, niter,
...
+        ``convex_conj.proximal``.
     L : `sequence` of `Operator`'s
-        Sequence of `Opeartor`s` with as many elements as ``prox_cc_gs``.
+        Sequence of `Opeartor`s` with as many elements as ``g``.
     tau : float
-        Step size parameter for ``prox_f``.
+        Step size parameter for ``f``.
```
kohr-h: Needs update (step-size-like)

odl/solvers/functional/default_functionals.py
```diff
@@ -464,3 +464,67 @@ def __init__(self, space):
...
         """
         super().__init__(space=space, constant=0)
+
+
+class IndicatorBox(Functional):
+
+    """Indicator on some box shaped domain."""
+
+    def __init__(self, space, lower, upper):
+        """Initialize an instance
```
kohr-h: .

odl/solvers/functional/default_functionals.py
```diff
@@ -464,3 +464,67 @@ def __init__(self, space):
...
+
+        Parameters
+        ----------
+        space : `LinearSpace`
+            Domain of the functional G(x)
```
kohr-h: Dot missing, G(x) makes no sense.

odl/solvers/functional/default_functionals.py
```diff
@@ -464,3 +464,67 @@ def __init__(self, space):
...
+        >>> func([0, 1, 3])  # one point outside
+        inf
+        """
+        Functional.__init__(self, space, linear=False)
+        self.lower = lower
+        self.upper = upper
+
+    def _call(self, x):
+        """Apply the functional to the given point."""
+        out = self.domain.element()
```
kohr-h: `out` is a weird name, why not `tmp`?

odl/solvers/functional/functional.py
```diff
@@ -884,3 +885,44 @@ def convex_conj(self):
         """
         return self.functional.convex_conj.translated(
             self.linear_term)
+
+
+class FunctionalDefaultConvexConjugate(Functional):
+
+    """The ``Functional`` representing ``f^*(.)``.
```
kohr-h: Functional in single ticks.

odl/solvers/functional/functional.py
```diff
@@ -884,3 +885,44 @@ def convex_conj(self):
...
+
+    Does not provide a way in which to evaluate the functional, intended to
```
kohr-h: This class does not ..., it is rather intended to ...
~~in which to~~

odl/solvers/functional/functional.py
kohr-h: Also: the description should make sense of `f`. Like 
```
"""The `Functional` representing ``F^*``, the convex conjugate of ``F``."""
```
Abbreviate conj. or remove "convex" if it doesn't fit.

odl/solvers/functional/functional.py
```diff
@@ -884,3 +885,44 @@ def convex_conj(self):
...
+    be used for its `proximal`.
+    """
+
+    def __init__(self, func):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        func : `Functional`
+            Function corresponding to F.
```
kohr-h: Functional <-> Function. Settle for one (Functional). Also `F` here and `f^*` above, both don't mean anything.

odl/solvers/functional/functional.py
```diff
@@ -884,3 +885,44 @@ def convex_conj(self):
...
+        """
+        if not isinstance(func, Functional):
+            raise TypeError('`func` {} is not a `Functional` instance'
+                            ''.format(func))
+
+        super().__init__(space=func.domain,
+                         linear=func.is_linear)
+
+        self.__functional = func
```
kohr-h: Why double underscore here, but not [here](https://github.com/odlgroup/odl/pull/587/files#diff-50e0484cf75f430f06878917e0952f87R499)?

test/solvers/advanced/douglas_rachford_test.py
```diff
@@ -40,40 +40,40 @@ def test_primal_dual_input_handling():
     space1 = odl.uniform_discr(0, 1, 10)
 
     lin_ops = [odl.ZeroOperator(space1), odl.ZeroOperator(space1)]
-    prox_cc_g = [odl.solvers.proximal_const_func(space1),  # Identity operator
-                 odl.solvers.proximal_const_func(space1)]  # Identity operator
-    prox_f = odl.solvers.proximal_const_func(space1)  # Identity operator
+    g = [odl.solvers.ZeroFunctional(space1),  # Identity operator
+         odl.solvers.ZeroFunctional(space1)]  # Identity operator
+    f = odl.solvers.ZeroFunctional(space1)  # Identity operator
```
kohr-h: Superfluous comments

adler-j:
> Apart from that, I think the interface is okay. It was probably quite hard to add a proximal before, and adding a functional is only slightly harder. 

Agree

> The question is if we want to provide a simple way to wrap an "old style" proximal factory or convex conjugate or proximal of convex conjugate into a dummy functional instance compatible with this interface.

I would say no, this is quite easy to hack if someone wants to do this

```
>>> obj = object()
>>> obj.cconj = object()
>>> obj.cconj.proximal = my_proximal
```

Otherwise actually defining it is a few lines. I would wait until we see a use case.

kohr-h:
> I would say no, this is quite easy to hack if someone wants to do this
> 
> ```
> >>> obj = object()
> >>> obj.cconj = object()
> >>> obj.cconj.proximal = my_proximal
> ```
> 
> Otherwise actually defining it is a few lines. I would wait until we see a use case.

That's the poor man's way of doing it, which is fine as long as we really never check that we got a `Functional`. That implies situations where we need to distinguish between different types of input in the same parameter. It means that we always have to check
```python
if isinstance(parameter, the_other_type):
    # something happens
else:
    # `Functional` case
```
and we'll get crashes with our thing if we ever happen to code it the other way around.
That said, I don't quite see a use case of all this, and I agree that we can wait until there is one.

odl/solvers/functional/functional.py
adler-j: Needed to "rename it" here but not in the other case, basically lazyness.

odl/solvers/advanced/douglas_rachford.py
```diff
@@ -51,27 +51,27 @@ def douglas_rachford_pd(x, prox_f, prox_cc_g, L, tau, sigma, niter,
     ----------
     x : `LinearSpaceElement`
         Initial point, updated in-place.
-    prox_f : callable
+    f : `Functional`
         `proximal factory` for the function ``f``.
```
aringh: `Function``f``. Need to have f.proximal`

adler-j:
So guys! Just saying, dont get ahead of yourselves in reviewing this in depth.

This is a "discussion", aka it is done to show what the effects of this change would be. If we decide to go for it we can trim the doc.

The main questions is if we are fine with this, or if it causes some unforseen problems.

aringh:
I think it looks nice :) It does create a bit of overhead in some sense, since the user need to send a functional that implements only `g.convex_conj.proximal`. But on the other hand, if one wants to solve `min_x f(x) + sum_i g_i(L_i x)` it feels intuitive to give the solver `f`, `g` and `L`.

Input from @ozanoktem and/or @hilding79 ?

kohr-h:
> The main questions is if we are fine with this, or if it causes some unforseen problems.

I already said I'm fine with this as it is now, and I can't foresee any unforeseen problems :-). Or one could also call them [unknown unknowns](https://en.wikipedia.org/wiki/There_are_known_knowns).

adler-j:
I fear the biggest problem will be in things that are supposed to be relatively simple, like the "old style" solvers. In there, feeding a full functional just to grab its `gradient` could perhaps be overkill.

adler-j:
I Decided to go for the Broyden solvers next, they seemed to be abit outdated (#599) but should now be working. No example of using them available atm, but after #597 goes it we could add it.

adler-j:
The reason for it is that the changes broke the tests as well, so the option was either to commit with failing tests, or to reimplement the tests only to throw them away in the next commit. Neither was particularly tempting.

I do certainly agree though and I try to avoid it.

kohr-h:
Yeah well, the choice to mess up `git blame` vs. `git bisect`...
The easier workflow is first move and adapt paths, then the changes.
Don't put more work into it now.

adler-j:
This PR is now to my knowledge feature complete, time to review.

Experience from implementing this:

### Interfaces become *much* cleaner.

In some cases ridiculously so. Lots of complexity can be hidden.

Compare the following examples of solving the Heron problem using the `douglas_rachford_pd` solver (stripped of comments for brevity):

OLD CODE:

```python
rectangles = [[[0, 0], [1, 1]],
              [[0, 2], [1, 3]],
              [[2, 2], [3, 3]]]

space = odl.rn(2)
lin_ops = [odl.IdentityOperator(space)] * len(rectangles)
prox_f = odl.solvers.proximal_const_func(space)
prox_cc_g = [odl.solvers.proximal_cconj_l2(space)] * len(rectangles)
prox_l = [odl.solvers.proximal_box_constraint(space, minp, maxp)
          for minp, maxp in rectangles]
prox_cc_l = [odl.solvers.proximal_cconj(prox) for prox in prox_l]

tau = 1.0 / len(rectangles)
sigma = [1.0] * len(rectangles)

x = space.zero()
odl.solvers.douglas_rachford_pd(x, prox_f, prox_cc_g, lin_ops,
                                tau=tau, sigma=sigma, niter=20, lam=lam,
                                callback=print_objective, prox_cc_l=prox_cc_l)
```

NEW CODE:

```python
rectangles = [[[0, 0], [1, 1]],
              [[0, 2], [1, 3]],
              [[2, 2], [3, 3]]]

space = odl.rn(2)
lin_ops = [odl.IdentityOperator(space)] * len(rectangles)
f = odl.solvers.ZeroFunctional(space)
g = [odl.solvers.L2Norm(space)] * len(rectangles)
l = [odl.solvers.IndicatorBox(space, minp, maxp) for minp, maxp in rectangles]

tau = 1.0 / len(rectangles)
sigma = [1.0] * len(rectangles)

x = space.zero()
odl.solvers.douglas_rachford_pd(x, f, g, lin_ops,
                                tau=tau, sigma=sigma, niter=20, l=l)
```

I mean would you look at that readability improvement? No more magical calls to `proximal_cconj` suddenly happening etc, just state your problem at it's solved.

### Adding stuff becomes harder

It becomes harder to simply add a new proximal. Perhaps I'm noticing this extra much since I have lots of nice proximals but not lots of fancy functionals, but most of the work on this branch has been in adding these anchilary functions. Issues #590, #598, #611, #612 and #613 have all been cases of me needing something added to do this nicely.

I think once #613 gets solved this will improve, but before that this actually is a bit hard to manoeuvre.

### The solvers package needs some love

Especially the smooth solvers are way outdated, the Broyden solvers did not seem to work at all (issue #599), and this is totally understandable since we never use them, but now that the functionals are in using them becomes a bit more simple and once @ozanokten solves #597 we could actually start using them even for TV problems in some cases.

The remaining work that should be done here is a documentation review, which I've partly done, and particularly a set of examples. We currently lack any example of smooth optimization! That has to be fixed.

### Overall this is a needed change

This change severely reduces the step to start using ODL and I guess it is a needed change for our audience. Overall it is a step away from algorithm developers towards algorithm users, and we need to be very careful here to lot lose track of our "ideals".

Input welcome, @ozanoktem, @aringh, @kohr-h, @hilding79, @chongchenmath.

adler-j:
rebased on master for easier review.

odl/solvers/functional/default_functionals.py
```diff
@@ -576,29 +577,23 @@ def __init__(self, space):
 class KullbackLeibler(Functional):
 
     """The Kullback-Leibler divergence functional.
-
```
adler-j: Something bad happened in the merge here.

adler-j:
This has become the new conebeam branch, we need to get it in shortly and then issue a 0.5.0 release.

odl/operator/operator.py
```diff
@@ -31,7 +31,7 @@
 from odl.set import LinearSpace, LinearSpaceElement, Set, Field
 
 
-__all__ = ('Operator', 'OperatorComp', 'OperatorSum',
+__all__ = ('Operator', 'OperatorComp', 'OperatorSum', 'OperatorVectorSum',
```
kohr-h: Do we need all those expression types in the top level namespace? In my view, they clutter substantially, and most users don't want to see them since they never use them directly.

odl/operator/operator.py
```diff
@@ -1158,6 +1155,97 @@ def __str__(self):
...
+        """The operator to apply."""
+        return self.__operator
+
+    @property
+    def vector(self):
+        """The constant operator range element to add."""
+        return self.__vector
+
+    def _call(self, x, out=None):
+        """Evaluate the residual at ``x`` and write to ``out`` if given."""
```
kohr-h: outdated

odl/solvers/functional/default_functionals.py
```diff
@@ -1029,6 +1028,68 @@ def convex_conj(self):
...
+    The separable sum is thus defined for any collection of functionals with
+    the same range.
+    """
+
+    def __init__(self, *functionals):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        functionals1, ..., functionalsN : `Functional`
```
kohr-h: `functional1, ..., functionalN`

odl/solvers/functional/default_functionals.py
```diff
@@ -1029,6 +1028,68 @@ def convex_conj(self):
...
+            The functionals in the sum.
+        """
+        domains = [func.domain for func in functionals]
+        domain = ProductSpace(*domains)
+        linear = all(func.is_linear for func in functionals)
+
+        self.functionals = functionals
```
kohr-h: property?

odl/solvers/functional/default_functionals.py
```diff
@@ -1029,6 +1028,68 @@ def convex_conj(self):
...
+
+        super().__init__(space=domain, linear=linear)
+
+    def _call(self, x):
+        """Return the squared L2-norm of ``x``."""
```
kohr-h: copy-paste

odl/solvers/functional/default_functionals.py
```diff
@@ -1029,6 +1028,68 @@ def convex_conj(self):
...
+class SeparableSum(Functional):
+
+    """The functional corresponding to separable sum of functionals.
+
+    The separable sum of functionals ``f_1, f_2, ..., f_n`` is given by::
+
+        h(x_1, x_2, ..., x_n) = sum_i^n f_i(x_i)
+
+    The separable sum is thus defined for any collection of functionals with
+    the same range.
```
kohr-h: Write something about gradient, proximal and convex conjugate, which are the interesting properties here.

odl/solvers/functional/functional.py
```diff
@@ -1072,6 +1072,77 @@ def proximal(self):
...
+        Gradient operator of the convex conjugate functional
+    cconj_prox : `proximal factory`, optional
+        Proximal factory for the convex conjugate functional.
+    cconj_grad_lip : float, optional
+        lipschitz constant of the convex conjugate functional.
+    linear : bool, optional
+        True if the operator is linear
+    """
+    class SimpleFunctional(Functional):
+        """A simplified functional for examples."""
```
kohr-h: blank line above

odl/solvers/functional/functional.py
```diff
@@ -1072,6 +1072,77 @@ def proximal(self):
...
+        Function to evaluate when calling the convex conjugate functional.
+    cconj_grad : `Operator`, optional
+        Gradient operator of the convex conjugate functional
+    cconj_prox : `proximal factory`, optional
+        Proximal factory for the convex conjugate functional.
+    cconj_grad_lip : float, optional
+        lipschitz constant of the convex conjugate functional.
+    linear : bool, optional
+        True if the operator is linear
+    """
```
kohr-h: A small doctest example would be nice, perhaps a very simple one from scratch, and one using properties of an existing functional and extending it.

odl/solvers/functional/functional.py
```diff
@@ -1072,6 +1072,77 @@ def proximal(self):
...
+    Users may specify as many properties as is needed by the application.
+
+    Parameters
+    ----------
+    space : `LinearSpace`
+        Space that the functional should act on.
+    fcall : `callable`, optional
+        Function to evaluate when calling the functional.
+    grad : `Operator`, optional
+        Gradient operator of the functional
```
kohr-h: dot

odl/solvers/nonsmooth/__init__.py
```diff
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
+"""Solvers for non-smooth optimization."""
```
kohr-h: optimization problems

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -28,7 +28,7 @@
 __all__ = ('douglas_rachford_pd',)
 
 
-def douglas_rachford_pd(x, prox_f, prox_cc_g, L, tau, sigma, niter,
+def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
```
kohr-h: We should definitely look over the names in this interface. It becomes even worse in the forward-backward example where an additional `h` pops up. Even `x` is up to discussion IMO. @ozanoktem, @hilding79, @aringh?

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""(Quasi-)Newton schemes to find zeros of functions."""
```
kohr-h: functionals

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+import numpy as np
+from odl.operator import IdentityOperator
+from odl.solvers.util import ConstantLineSearch
+from odl.solvers.iterative.iterative import conjugate_gradient
+
+
+__all__ = ('newtons_method', 'bfgs_method', 'broydens_method')
+
+
+# TODO: update all docs
```
kohr-h: Done?

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+def _bfgs_multiply(s, y, x, hessinv_estimate=None):
+    """Computes ``Hn^-1(x)`` for the L-BFGS method.
```
kohr-h: Compute

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+    for i in range(len(s)):
+        beta = rhos[i] * (y[i].inner(r))
+        r.lincomb(1, r, alphas[i] - beta, s[i])
+
+    return r
+
+
+def newtons_method(f, x, line_search=1.0, maxiter=1000, tol=1e-16,
+                   cg_iter=None, callback=None):
+    """Newton's method for minimizing a functional.
```
kohr-h: update or make a new issue on it

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+        Starting point of the iteration
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
+        fixed step length.
+    maxiter : int, optional
+        Maximum number of iterations.
+        ``tol``.
+    tol : float, optional
+        Tolerance that should be used for terminating the iteration.
+    maxcor : int, optional
```
kohr-h: Name makes no sense until one reads this doc. Becomes slightly better with an additional "r".
How about using something like "store" in the name to relate it to the storage thing? Like `nstore`, `num_store` or `maxstore`.

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+        Maximum number of correction factors to store. If ``None``, the method
+        is the regular BFGS method. If an integer, the method becomes the
```
kohr-h: For an integer

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+    line_search : float or `LineSearch`, optional
+        Strategy to choose the step length. If a float is given, uses it as a
+        fixed step length.
+    maxiter : int, optional
+        Maximum number of iterations.
+        ``tol``.
+    tol : float, optional
+        Tolerance that should be used for terminating the iteration.
+    maxcor : int, optional
+        Maximum number of correction factors to store. If ``None``, the method
```
kohr-h: For None

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+    ----------
+    s : `sequence` of `LinearSpaceElement`
+        The ``s`` coefficients in the BFGS update, see Notes.
+    y : `sequence` of `LinearSpaceElement`
+        The ``y`` coefficients in the BFGS update, see Notes.
+    x : `LinearSpaceElement`
+        Point in which to evaluate the product.
+    hessinv_estimate : `Operator`, optional
+        Initial estimate of the hessian ``H0^-1``.
+
```
kohr-h: Returns

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+from odl.solvers.iterative.iterative import conjugate_gradient
+
+
+__all__ = ('newtons_method', 'bfgs_method', 'broydens_method')
+
+
+# TODO: update all docs
+
+
+def _bfgs_multiply(s, y, x, hessinv_estimate=None):
```
kohr-h: bfgs_direction?

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+    if not callable(line_search):
+        line_search = ConstantLineSearch(line_search)
+
+    ys = []
+    ss = []
+
+    grad_x = grad(x)
+    for _ in range(maxiter):
+        # Determine a stepsize using line search
+        search_dir = -_bfgs_multiply(ys, ss, grad_x, hessinv_estimate)
```
kohr-h: `bfgs_direction` would make sense for the function name

odl/solvers/smooth/newton.py
```diff
@@ -0,0 +1,380 @@
...
+        dir_deriv = search_dir.inner(grad_x)
+        if np.abs(dir_deriv) < tol:
+            return  # we found an optimum
+        step = line_search(x, direction=search_dir, dir_derivative=dir_deriv)
+
+        # Update x
+        x_update = search_dir
+        x_update *= step
+        x += x_update
```
kohr-h: Use lincomb?

odl/solvers/util/steplen.py
```diff
@@ -146,8 +124,21 @@ def __call__(self, x, direction, dir_derivative):
         step : float
             The computed step length
         """
-        alpha = 1.0
         fx = self.function(x)
+        dir_derivative = float(dir_derivative)
+
+        if dir_derivative == 0:
+            # TODO: raise instead?
+            raise ValueError('dir_derivative == 0, no decent can be found')
```
kohr-h: descent

odl/solvers/util/steplen.py
```diff
@@ -146,8 +124,21 @@ def __call__(self, x, direction, dir_derivative):
...
+        if not self.estimate_step:
+            alpha = 1.0
+        else:
+            alpha = self.alpha
+
+        if dir_derivative > 0:
+            # We need to move backwards if the direction is a
```
kohr-h: an

test/solvers/nonsmooth/chambolle_pock_test.py
```diff
@@ -61,12 +61,12 @@ def test_chambolle_pock_solver_simple_space():
     # Dual variable required to resume iteration
     discr_dual = op.range.zero()
 
-    # Proximal operator, use same the factory function for F^* and G
-    prox = proximal_const_func(space)
+    # Functional, use same the factory function for F^* and G
```
kohr-h: outdated

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test for the Chambolle-Pock solver."""
```
kohr-h: copy-paste

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+# External
```
kohr-h: remove

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+import pytest
+
+# Internal
```
kohr-h: remove

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+from odl.operator import OpNotImplementedError
+from odl.util.testutils import noise_element
+
+
+@pytest.fixture(scope="module", params=['l2_squared', 'l2_squared_scaled'])
+def functional(request):
+    """Return a functional whose optimal value is at 0."""
+    name = request.param
+
+    # TODO: Add rosenbrock (#600) and quadratic (#606) functionals
```
kohr-h: status?

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+        scaling = odl.MultiplyOperator(space.element([1, 2, 3, 5, 5]),
+                                       domain=space)
+        return odl.solvers.L2NormSquared(space) * scaling
+    else:
+        assert False
+
+
+@pytest.fixture(scope="module", params=['constant', 'backtracking'])
+def functional_and_linesearch(request, functional):
+    """Return a functional whose optimal value is at 0."""
```
kohr-h: What?

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+    x = functional.domain.zero()
+    odl.solvers.newtons_method(functional, x, maxiter=50, tol=1e-4,
+                               line_search=line_search)
+
+    # Assert x is close to the optimum at [1, 1]
+    assert functional(x) < 1e-3
+
+
+def test_bfgs_solver(functional_and_linesearch):
+    """Test the BFGS quasi-newton solver."""
```
kohr-h: Newton (throughout)

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+    x = noise_element(functional.domain)
+    odl.solvers.bfgs_method(functional, x, maxiter=50, tol=1e-4,
+                            line_search=line_search, maxcor=5)
+
+    assert functional(x) < 1e-3
+
+
+def test_broydens_method(functional_and_linesearch, broyden_impl):
+    """Test a quasi-newton solver."""
```
kohr-h: Why unspecific here?

test/solvers/smooth/smooth_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+    x = noise_element(functional.domain)
+    odl.solvers.steepest_descent(functional, x, maxiter=50, tol=1e-4,
+                                 line_search=line_search, callback=print)
+
+    assert functional(x) < 1e-3
+
+
+if __name__ == '__main__':
+    pytest.main(str(__file__.replace('\\', '/') + ' -v'))
```
kohr-h: comma instead of plus

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: I have also grown to like the very short `f`, `g` etc mostly because frankly any name we come up with wont mean much unless you have read the documentation, and these can at least easily be tied with the research article.

The largest change I'd do is to make something imply that the g's can be a list (frankly we should add some code that relaxes this assumption and allows single functions as input.) Perhaps `gi`, `gs` etc.

Regarding `x` this is in following with the convention of `x0` used in scipy. Anything longer would clog all the interfaces with minimal gain, especially since this is almost universally called `x` in the optimization literature.

adler-j:
Great review dispite the obvious difficulties with moving things. Will get to it.

examples/solvers/lbfgs_tomography.py
```diff
@@ -0,0 +1,101 @@
...
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+data = ray_trafo(discr_phantom)
+data += odl.phantom.white_noise(ray_trafo.range) * np.mean(data) * 0.1
+
+# --- Set up optimization problem and solve --- #
+
+# Create objective functional
+obj_fun = odl.solvers.L2NormSquared(ray_trafo.range) * (ray_trafo - data)
```
aringh: I guess one should be able to do `odl.solvers.L2NormSquared(ray_trafo.range).translate(data) * ray_trafo` as well. But I don't know that is clearer.

odl/solvers/nonsmooth/douglas_rachford.py
aringh: I agree with @adler-j. I don't think we can name the functionals anything that will make it more clear what they are, if you do not read the documentation. And this naming is (at least, if I recall correctly) consistent with the original paper which we cite further down.

kohr-h: I find it hard remembering the difference between `f`, `g`, `h`, `l` and the other stuff. If you need to consult the documentation every single time, the names are not good. Especially when you see the function "in the wild", you need to go back every time and check which name was which. 

And matching paper notions is secondary IMO. What if we find another variant of the method in another paper that uses different notions, and we want to put it into the same function, like in the case of memory limited BFGS?

I think it would be good to give at least some hint about the nature of the variables in the names. They typically differ in attributes like "smooth", "composed", "inf-convolved", which would set them apart quite nicely.

kohr-h:
I also saw that some of the documentation still says "proximal" where it should say "functional", and that the step sizes also refer to the old variable names. Please double-check also there.

adler-j:
Also the example in the doc is old and needs an update.

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: Decent suggestion for names, although id go for `inf_conv_funcs` for consistency.

examples/solvers/lbfgs_tomography.py
adler-j: I found this to be more clear, e.g. first take residual and then take norm.

odl/operator/default_ops.py
```diff
@@ -791,8 +765,9 @@ def __init__(self, constant, domain=None, range=None):
         if range is None:
             range = constant.space
 
-        super().__init__(domain, range)
-        self.__constant = self.range.element(constant)
+        self.__constant = range.element(constant)
+        linear = self.constant.norm() == 0
```
adler-j: No

adler-j: Read this through, nowhere is the operator implied to work on scalars, so this wont work. We have `ConstantFunctional` for that.

odl/operator/default_ops.py
```diff
@@ -834,99 +816,71 @@ def __str__(self):
...
-        Examples
-        --------
-        >>> import odl
-        >>> r3 = odl.rn(3)
-        >>> op = IdentityOperator(r3)
-        >>> res = ResidualOperator(op, r3.element([1, 2, 3]))
-        >>> x = r3.element([4, 5, 6])
-        >>> res.derivative(x)(x)
-        rn(3).element([4.0, 5.0, 6.0])
+        The zero operator is self adjoint.
```
adler-j: Also it is not self adjoint if the range and domain are not matched, fixing.

odl/operator/operator.py
adler-j: We have an issue for this, #358, I refer this discussion there for now.

odl/solvers/functional/default_functionals.py
adler-j: Did a Notes writeup on that

odl/solvers/smooth/newton.py
adler-j: I guess yes

odl/solvers/smooth/newton.py
adler-j: This is taken directly from scipy [here](https://github.com/scipy/scipy/blob/v0.18.1/scipy/optimize/lbfgsb.py#L217-L221), but I do agree. It seems scipy also uses `m` which is likely even worse of. I like `num_store`.

odl/solvers/smooth/newton.py
adler-j: Fine fine :(

odl/solvers/smooth/newton.py
adler-j: No, I use the scaled guy later so I think this may be fast enough, using lincomb could shave of a very small time diff but would add a non-obvious step term a few lines further down.

odl/solvers/util/steplen.py
adler-j: Well no decent direction either, just shitty ones :/

test/solvers/smooth/smooth_test.py
adler-j: Should be add-able for rosenbrock, quadratic still needs review.

odl/solvers/smooth/newton.py
kohr-h: You mean ":-)"?
Don't feel forced, I just found the single use case for the function to be the direction computation.

test/solvers/smooth/smooth_test.py
adler-j: Just did this which resulted in a major "review" of the quasi newton solvers. They now have a primitive resetting functionality since they refused to converge due to instabilities due to scale differences.

adler-j:
Fixed all comments except the pesky naming issue. Had a discussion with @ozanoktem about this and we both tend towards the simple `f`, `g` etc naming, mostly because it makes reading the documentation so much easier. Without it, we would need a translation table:

```
smooth_func : Functional
    The `h` functional in the problem statement. Lorem ipsum.
```

kohr-h:
> Fixed all comments except the pesky naming issue. Had a discussion with @ozanoktem about this and we both tend towards the simple f, g etc naming, mostly because it makes reading the documentation so much easier.

Okay, it seems I'm on lost ground here. I'll write my final remark and shut up.

Of course, an interface that matches names in the math documentation is a good thing. But in my opinion, an even better interface is one that relieves you from consulting the documentation in the first place. If you use keyword arguments like
```python
forward_backward_pd(x, smooth_func=myfunc1, comp_funcs=[g1, g2], ...)
```
it is trivial to exchange stuff without having to go back to the documentation and find out which argument was which. Compare:
```python
forward_backward_pd(x, h=myfunc1, g=[g1, g2], ...)
```
Will you know in 3 months whether `h` was the smooth or the non-smooth functional? I find the latter interface harder to digest because the names give you no hint about their meanings.

Okay, now if you are 3 in favor of `f, g, h` against 1, I give up. :see_no_evil:

odl/solvers/functional/default_functionals.py
```diff
@@ -1038,29 +1038,63 @@ class SeparableSum(Functional):
 
     The separable sum is thus defined for any collection of functionals with
     the same range.
+
+    Notes
+    -----
+    The separable sum of functionals :math:`f_1, f_2, ..., f_n` is given by
+
+    .. math::
+        h(x_1, x_2, ..., x_n) = \sum_i^n f_i(x_i)
```
kohr-h: `\sum_{i=1}^n`

odl/solvers/functional/default_functionals.py
```diff
@@ -1038,29 +1038,63 @@ class SeparableSum(Functional):
...
+    gradient is a `DiagonalOperator`:
+
+    .. math::
+        [\\nabla h](x_1, x_2, ..., x_n) =
+        [\\nabla f_1(x_i), \\nabla f_2(x_i), ..., \\nabla f_n(x_i)]
+
+    The convex conjugate distributes:
+
+    .. math::
+        [h^*](y_1, y_2, ..., y_n) = \sum_i^n f_i^*(y_i)
```
kohr-h: `\sum_{i=1}^n`

odl/solvers/functional/functional.py
```diff
@@ -1085,24 +1085,71 @@ def simple_functional(space, fcall=None, grad=None, prox=None, grad_lip=np.nan,
...
+    >>> space = odl.rn(3)
+    >>> func = simple_functional(space, f, grad=dfdx)
+    >>> func.domain
+    rn(3)
+    >>> func.range
+    RealNumbers()
+    >>> func([1, 2, 3])
+    14.0
+    >>> func.gradient([1, 2, 3])
+    rn(3).element([2.0, 4.0, 6.0])
```
kohr-h: Nice example!

odl/solvers/smooth/newton.py
```diff
@@ -336,16 +395,19 @@ def broydens_method(f, x, line_search=1.0, impl='first', maxiter=1000,
...
     for _ in range(maxiter):
         # find step size
-        search_dir = -hess(grad_x)
+        search_dir = -_broydens_direction(ss, ys, grad_x,
+                                          hessinv_estimate, impl)
         dir_deriv = search_dir.inner(grad_x)
         if np.abs(dir_deriv) < tol:
             return  # we found an optimum
         step = line_search(x, search_dir, dir_deriv)
+        print(x, search_dir, f(x), step)
```
kohr-h: leftover print

odl/solvers/functional/example_funcs.py
```diff
@@ -75,20 +75,32 @@ def __init__(self, space, scale=100.0):
...
+        101.0
 
         The functional can also be used in higher dimensions
 
         >>> r5 = odl.rn(5)
         >>> functional = RosenbrockFunctional(r5)
         >>> functional([1, 1, 1, 1, 1])
         0.0
+
+        We can change how much the function is ill behaved via ``scale``
```
kohr-h: colon

odl/solvers/smooth/newton.py
```diff
@@ -323,23 +324,37 @@ def bfgs_method(f, x, line_search=1.0, maxiter=1000, tol=1e-16, num_store=None,
...
+        if np.abs(y_inner_s) < tol:
+            if grad_x.norm() < tol:
+                return
+            else:
+                # Reset if needed
+                ys = []
+                ss = []
+                continue
+
+        if i == 0 and hessinv_estimate is None and False:
```
kohr-h: left-over `and False`?

odl/solvers/smooth/newton.py
```diff
@@ -323,23 +324,37 @@ def bfgs_method(f, x, line_search=1.0, maxiter=1000, tol=1e-16, num_store=None,
...
+            # Estimate scale
+            scale = x_update.inner(x_update) / grad_diff.inner(x_update)
+            hessinv_estimate = MultiplyOperator(scale,
+                                                grad.domain, grad.domain)
+        else:
+            # Update Hessian
+            ys += [grad_diff]
+            ss += [x_update]
```
kohr-h: Minor thing: better use `ys.append(grad_diff)` and similar

test/solvers/smooth/smooth_test.py
```diff
@@ -25,16 +25,17 @@
...
 from odl.operator import OpNotImplementedError
-from odl.util.testutils import noise_element
 
 
 @pytest.fixture(scope="module", params=['l2_squared', 'l2_squared_scaled',
                                         'rosenbrock'])
 def functional(request):
-    """Return a functional whose optimal value is 0."""
+    """Return a functional whose optimal value,
+    which is in the point 0, is 0.
```
kohr-h: functional with optimum 0 at 0?

odl/solvers/smooth/newton.py
adler-j: Yes

test/solvers/smooth/smooth_test.py
adler-j: Much better

odl/solvers/smooth/newton.py
adler-j: Totally removed it now, don't know enough about scaling to get it right.

adler-j:
Monster merge performed. Waiting for CI then i'll go to master. We should fix 0.5.0 now.

PR #588, opened 2016-09-19
ENH: Add default convex conjugate, closes #570

### LINKED ISSUE #570: Add a default convex conjugate
Currently in the functional class, `convex_conj` raises `NotImplementedError`. This is problematic for proximals where some algebra exists.

Consider if I implement a functional with a proximal

```python
class MyFancyFunctional(Functional):
    ...
    @property
    def proximal(self):
        return MyProximal
```

and I want to use this with a solver that takes the proximal of its convex conjugate. Here I would expect this to work due to the Moreau identity

```python
f = MyFancyFunctional(...)
my_solver(f.convex_conj.proximal)
```

But right now I simply get `NotImplementedError`.

We should implement a class, like `DefaultConvexConjugate`, which does not implement `_call`, but which implements `proximal`.

adler-j:
Cherry picked from #587, should be simple to review.

Things that need to be done:

* Add a test

Things that could be done:

* Add some intelligence so that this can only be created if it is useful (i.e. if the functional has a proximal). This could be nice in some cases I guess, but in others it could also give hard to follow logic.
* Perhaps add some logic to evaluate the functional in some sense (perhaps solve the optimization problem, somehow). Not sure at all if this is a good idea.

### DIFF
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -29,7 +29,7 @@ from odl.operator.operator import (
 from odl.operator.default_ops import (IdentityOperator, ConstantOperator)
 from odl.solvers.advanced import (proximal_arg_scaling, proximal_translation,
                                   proximal_quadratic_perturbation,
-                                  proximal_const_func)
+                                  proximal_const_func, proximal_cconj)
 
 
 __all__ = ('Functional', 'FunctionalLeftScalarMult',
@@ -160,7 +160,7 @@ class Functional(Operator):
 
         For literature references see, e.g., [Lue1969]_, [Roc1970]_.
         """
-        raise NotImplementedError
+        return FunctionalDefaultConvexConjugate(self)
 
     def derivative(self, point):
         """Return the derivative operator in the given point.
@@ -430,7 +430,8 @@ class FunctionalLeftScalarMult(Functional, OperatorLeftScalarMult):
     def convex_conj(self):
         """Convex conjugate functional of the scaled functional.
 
-        `Functional.__rmul__` takes care of the case scalar = 0."""
+        `Functional.__rmul__` takes care of the case scalar = 0.
+        """
         if self.scalar <= 0:
             raise ValueError('scaling with nonpositive values have no convex '
                              'conjugate. Current value: {}.'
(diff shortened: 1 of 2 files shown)


test/solvers/functional/functional_test.py
```diff
@@ -280,11 +280,9 @@ def test_functional_sum(space):
                             func2.gradient(x).inner(p)),
                         places=PLACES)
 
-    # Test that prox and convex conjugate is not known
+    # Test that prox is not known
     with pytest.raises(NotImplementedError):
         (func1 + func2).proximal
-    with pytest.raises(NotImplementedError):
```
aringh: The convex conjugate should not be known for a sum. Wouldn't you still like to check something here, like that `(func1 + func2).convex_conj` is a `FunctionalDefaultConvexConjugate` or that  `(func1 + func2).convex_conj(x)` raises `NotImplemetedError`? Or are those kind of test just redundant?

odl/solvers/functional/functional.py
```diff
@@ -500,7 +501,7 @@ def __init__(self, func, scalar):
             raise TypeError('`func` {!r} is not a `Functional` instance'
                             ''.format(func))
 
-        scalar = func.range.element(scalar)
+        scalar = func.domain.field.element(scalar)
```
aringh: This should be the same, since `func` is checked to be a `functional` on the line above, and for `functional` we have `range = domain.field`.

kohr-h: Currently yes, but once we got the math straight, we might in fact want allow spaces over the complex numbers as `domain`, and in that case, we want to cast to the field of the domain, not the range (which we want to keep as "always real", I suppose).

test/solvers/functional/functional_test.py
kohr-h: The thing is that with the default convex conjugate added, this line won't raise any more but rather return an object that cannot be evaluated, but still asked for its convex conjugate and proximal.

kohr-h:
> Looks very good to me! :)

Same here.

test/solvers/functional/functional_test.py
adler-j: So this test should simply be changed to testing that we cannot evaluate it

kohr-h:
@adler-j  Please write a release note on this.

PR #591, opened 2016-09-20
BUG: fix functional failures with cuda, closes #538

### LINKED ISSUE #538: edge_order should not be a separate argument to finite_diff
It makes no sense to have a specified edge order AND a padding mode. The edge order is (i.m.o) simply a special case of `pad_mode` and should be baked into that parameter.

While at it, we could perhaps expose this in the cases of `Gradient` etc where this parameter is currently not exposed.

adler-j:
Basically removed `PLACES` everywhere and let `almost_equal` figure it out.

### DIFF
--- a/test/solvers/functional/default_functionals_test.py
+++ b/test/solvers/functional/default_functionals_test.py
@@ -25,9 +25,6 @@ import pytest
 import odl
 from odl.util.testutils import all_almost_equal, almost_equal, noise_element
 
-# Places for the accepted error when comparing results
-PLACES = 8
-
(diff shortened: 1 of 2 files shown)


test/solvers/functional/default_functionals_test.py
```diff
@@ -106,7 +100,7 @@ def test_indicator_lp_unit_ball(space):
     one_elem = space.one()
 
     # Used du to numerical errors
```
kohr-h: due to

PR #592, opened 2016-09-21
REL: empty doc for RTD

kohr-h:
Removes the doc and replaces it with a referring note so that RTD stops showing an outdated page. I suggest we merge this, then switch RTD off, revert this merge and then merge the #536 PR. Okay @adler-j ?


adler-j:
Looks fine by me

Issue #600, opened 2016-09-22
Add famous example functionals


adler-j:
Several famous "test" functionals exist. [Wikipedia has a long list](https://en.wikipedia.org/wiki/Test_functions_for_optimization) and there is a good [literature study](https://arxiv.org/pdf/1308.4008.pdf).

Some that we should add to ODL, mostly for testing purposes, are:

* [Rosenbrock function](https://en.wikipedia.org/wiki/Rosenbrock_function)
* More?

aringh:
We have a simple version of a 2d rosenbrock function [here.](https://github.com/odlgroup/odl/blob/7dad80ed4342346126b0df36f931556ada5596a0/test/solvers/vector/newton_test.py#L40)

aringh:
@ozanoktem also sent me these suggestions a couple of months ago, but I never had time to really look into it:
 
- L1 test pack: https://www.tu-braunschweig.de/iaa/personal/lorenz/l1testpack
- SPEAR project: http://wwwopt.mathematik.tu-darmstadt.de/spear/
- Global Optimization Test Problems: http://www.mat.univie.ac.at/~neum/glopt/test.html
- Handbook of Test Problems in Local and Global Optimization: http://titan.princeton.edu/TestProblems/
- The book "A Collection of Test Problems for Constrained Global Optimization Algorithms” by Floudas C A and Pardalos P M, Springer, 1990.

adler-j:
Closed this since rosenbrock is in, if we need more we should probably make an issue for each since it is a bit of work to add them.

ozanoktem:
One may here also consider test suits developed specifically for sparse inverse problems. One such suit is [SPARCO](http://www.cs.ubc.ca/labs/scl/sparco/), another is [COPS](http://www.mcs.anl.gov/~more/cops/).

PR #602, opened 2016-09-22
ENH: Add rosenbrock functional, see #600

adler-j:
Minor yet probably useful addition. We could add more examples if we want.

### DIFF
--- a/odl/solvers/functional/example_funcs.py
+++ b/odl/solvers/functional/example_funcs.py
@@ -1,0 +1,160 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example functionals used in optimization."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import super
+
+import numpy as np
+
+from odl.solvers.functional.functional import Functional
+from odl.operator.operator import Operator
+from odl.space.base_ntuples import FnBase
+from odl.space.npy_ntuples import MatVecOperator
+
+
+__all__ = ('RosenbrockFunctional',)
+
+
+class RosenbrockFunctional(Functional):
+
+    """The well-known `Rosenbrock function`_ on ``R^n``.
+
+    This function is usually used as a test problem in smooth optimization.
+
+    Notes
+    -----
+    The functional is defined for :math:`x` in :math:`\\mathbb{R}^n`,
+    :math:`n \\geq 2`, as
+
+    .. math::
+        \sum_{i=1}^{n - 1} c (x_{i+1} - x_i^2)^2 + (1 - x_i)^2
+
+    Where :math:`c` is a constant usually set to 100 which determines how "ill-
+    behaved" the function should be.
+    It has a minimum at :math:`x = [1, \\dots, 1]`, independent of :math:`c`.
+
+    There are two definitions of the n-dimensional Rosenbrock function found in
+    the literature. One is the product of 2-dimensional Rosenbrock functions,
+    which is not the one used here. This one extends the pattern of the 2d
+    Rosenbrock function so all dimensions depend on each other in sequence.
+
+    References
+    ----------
+    .. _Rosenbrock function: en.wikipedia.org/wiki/Rosenbrock_function
+    """
+
+    def __init__(self, space, scale=100.0):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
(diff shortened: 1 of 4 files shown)


odl/solvers/functional/example_functionals.py
```diff
@@ -0,0 +1,149 @@
...
+from odl.space.base_ntuples import FnBase
+from odl.space.npy_ntuples import MatVecOperator
+
+
+__all__ = ('RosenbrockFunctional',)
+
+
+class RosenbrockFunctional(Functional):
+
+    """The famous Rosenbrock funciton on ``R^n``.
```
adler-j: funciton

odl/solvers/functional/example_functionals.py
```diff
@@ -0,0 +1,149 @@
...
+        scale : positive float, optional
+            The scale ``c`` in the functional, determines how ill posed the
+            functional should be.
+
+        Examples
+        --------
+        >>> import odl
+        >>> r2 = odl.rn(2)
+        >>> functional = RosenbrockFunctional(r2)
+        >>> functional([1, 1])  # optimum is 0 at [0, 0]
```
aringh: The optimum is at `[1, 1]`.

odl/solvers/functional/example_functionals.py
aringh: The Rosenbrock function seems to have [two different versions](https://en.wikipedia.org/wiki/Rosenbrock_function#Multidimensional_generalisations). Should we clarify this?

adler-j: Perhaps we should, but the first one is not feasible for a general implementation since it only works for spaces of even dimension. I'll write this in the notes

kohr-h: Also typo "funciton"

odl/solvers/functional/example_functionals.py
```diff
@@ -0,0 +1,149 @@
...
+    The functional is defined for :math:`x` in :math:`\\mathbb{R}^n`,
+    :math:`n \\geq 2`, as
+
+    .. math::
+        \sum_{i=1}^{n - 1} c (x_{i+1} - x_i^2)^2 + (1 - x_i)^2
+
+    Where :math:`c` is a constant usually set to 100 which determines how ill-
+    posed the function should be.
+    It has a minimum at :math:`x = [1, \\dots, 1]`, independent of :math:`c`.
+    """
```
kohr-h: Formula for the gradient?

odl/solvers/functional/example_functionals.py
```diff
@@ -0,0 +1,149 @@
...
+                              4 * c * (x[i + 1] - x[i]**2) * x[i] -
+                              2 * (1 - x[i]))
+                out[0] = (-4 * c * (x[1] - x[0] ** 2) * x[0] +
+                          2 * (x[0] - 1))
+                out[-1] = 2 * c * (x[-1] - x[-2] ** 2)
+
+            def derivative(self, x):
+                """The derivative of the gradient.
+
+                This is also known as the hessian.
```
kohr-h: Formula in math at some place?

odl/solvers/functional/example_functionals.py
adler-j: It's a mess and I doubt anyone would read it, users can open the code if they are truly interested.

odl/solvers/functional/example_functionals.py
```diff
@@ -0,0 +1,149 @@
...
+        if space.size < 2:
+            raise ValueError('`space` must be at least two dimensional')
+        super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
+
+    def _call(self, x):
+        """Return ``self(x)``."""
+        result = 0
+        for i in range(0, self.domain.size - 1):
+            result += self.scale * ((x[i + 1] - x[i] ** 2) ** 2 +
+                                    (x[i] - 1) ** 2)
```
adler-j: The most common case will be R2 and such, there it only makes this harder to read and slower, I doubt it's worth it. Any performance gains would be in improving the hessian to be more diagonal.

odl/solvers/functional/example_functionals.py
kohr-h: OK

odl/solvers/functional/example_functionals.py
```diff
@@ -0,0 +1,149 @@
...
+        >>> functional = RosenbrockFunctional(r5)
+        >>> functional([1, 1, 1, 1, 1])
+        0.0
+        """
+        self.scale = float(scale)
+        if not isinstance(space, FnBase):
+            raise ValueError('`space` must be a `FnBase`')
+        if space.size < 2:
+            raise ValueError('`space` must be at least two dimensional')
+        super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
```
kohr-h: Right, I keep forgetting that we're on the whole space.

odl/solvers/functional/example_functionals.py
adler-j: Even worse than the gradient, for these things I think it's better to just let users read the code (it is readily available) if they are interested. Easier to maintain as well.

Adding it would only be helpfull (imo) if we also added a deriviation of the result, which would take way too much space.

adler-j:
Updated according to comments and rebased.

odl/solvers/functional/example_funcs.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+__all__ = ('RosenbrockFunctional',)
+
+
+class RosenbrockFunctional(Functional):
+
+    """The well-known Rosenbrock function on ``R^n``.
+
+    This function is usually used as a test problem in smooth optimization.
```
kohr-h: Please add a link to Wikipedia here.

adler-j:
Merge after CI

PR #615, opened 2016-09-23
TST: add pywt and pyfftw to installed packages on Travis

kohr-h:
My suggestion to get better (honest) test coverage. The test times don't go up significantly.

Question on the side, while we're at it: Travis now tests building the doc for every build with Python 3.5, which takes about 3 minutes alone. Since we seldom make changes to the doc infrastructure that break it, I could imagine doing that build only on the `master` branch, i.e. when PRs are merged or when someone pushes to `master` directly. This behavior is currently only realized for the "deploy" part of that building process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ python:
 install:
   - sudo apt-get update
 
+  # Install FFTW for pyFFTW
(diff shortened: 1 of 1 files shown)


kohr-h:
+2.3 % coverage for free :tada:

Issue #623, opened 2016-09-27
Add numerical gradient and hessian


adler-j:
These are often useful when testing code and checking that a computed result is correct, kindoff like the `matrix_representation` function.

If we make a free `NumericalGradient` operator and add a `gradient.setter` to functional we could even do things like

```python
>>> f = MyExtremelyComplicatedFunctional()
>>> f.gradient = odl.solvers.NumericalGradient(f)
```

We should also add a free function:

```python
>>> f = MyExtremelyComplicatedFunctional()
>>> odl.solvers.numerical_gradient(f, point)
... #result
```

In the same vein we could add a numerical hessian computation and let `NumericalGradient.derivative` use it.

PR #624, opened 2016-09-27
ENH: Add NumericalGradient, see #623

adler-j:
TODO:

* ~~Add hessian~~

Ready for review, I'd say I'm actually quite happy about how this turned out.

### DIFF
--- a/odl/solvers/functional/derivatives.py
+++ b/odl/solvers/functional/derivatives.py
@@ -1,0 +1,159 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utilities for computing the gradient and Hessian of functionals."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+import numpy as np
+
+from odl.solvers.functional.functional import Functional
+from odl.operator.operator import Operator
+from odl.space.base_ntuples import FnBase
+
+
+__all__ = ('NumericalGradient',)
+
+
+class NumericalGradient(Operator):
+
+    """The gradient of a `Functional` computed by finite differences."""
+
+    def __init__(self, functional, method='forward', step=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        functional : `Functional`
+            The functional whose gradient should be computed. Its domain must
+            be a `FnBase` space.
+        method : {'backward', 'forward', 'central'}
+            The method to use to compute the gradient.
+        step : `None` or float
+            The step length used in the gradient computation.
+            `None` is interpreted as selecting the step according to the dtype
+            of the space. ``step = 1e-8`` for ``float64`` spaces and
+            ``step = 1e-4`` for ``float32`` spaces.
+
+        Examples
+        --------
+        >>> import odl
+        >>> space = odl.rn(3)
+        >>> func = odl.solvers.L2NormSquared(space)
+        >>> grad = NumericalGradient(func)
+        >>> grad([1, 1, 1])
+        rn(3).element([2.0, 2.0, 2.0])
+
+        The gradient gives the correct value with sufficiently small step size:
+
+        >>> grad([1, 1, 1]) == func.gradient([1, 1, 1])
+        True
+
+        If the step is too large the result is not correct:
+
+        >>> grad = NumericalGradient(func, step=0.5)
+        >>> grad([1, 1, 1])
+        rn(3).element([2.5, 2.5, 2.5])
+
+        But it can be improved by using the more accurate ``method='central'``:
+
+        >>> grad = NumericalGradient(func, method='central', step=0.5)
+        >>> grad([1, 1, 1])
+        rn(3).element([2.0, 2.0, 2.0])
+
+        Notes
+        -----
+        If the function is :math:`f(x)` and stepsize :math:`h` is used, the
+        gradient is computed as follows.
+
+        If ``method='backward'`` is used:
+
+        .. math::
+            \\frac{f(x) - f(x - h)}{h}
+
+        if ``method='forward'`` is used:
+
+        .. math::
+            \\frac{f(x + h) - f(x)}{h}
+
+        if ``method='central'`` is used:
+
+        .. math::
+            \\frac{f(x + h/2) - f(x - h/2)}{h}
+
+        The number of function evaluations is ``functional.domain.size + 1`` if
+        ``'backward'`` or ``'forward'`` is used and
+        ``2 * functional.domain.size`` if ``'central'`` is used.
+        """
+        if not isinstance(functional, Functional):
+            raise TypeError('`functional` has to be a `Functional` instance')
+
+        if not isinstance(functional.domain, FnBase):
+            raise TypeError('`functional.domain` has to be a `FnBase` '
+                            'instance')
+
+        self.functional = functional
+        if step is None:
+            self.step = np.sqrt(np.finfo(functional.domain.dtype).eps)
+        else:
+            self.step = float(step)
+
+        self.method = str(method).lower()
+        if method not in ('backward', 'forward', 'central'):
+            raise ValueError('`method` not understood')
+
+        Operator.__init__(self, functional.domain, functional.domain,
+                          linear=functional.is_linear)
+
+    def _call(self, x):
+        """Return ``self(x)``."""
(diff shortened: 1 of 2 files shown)


odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,323 @@
...
+            The method to use to compute the derivative.
+        step : `None` or float
+            The step length used in the derivative computation.
+            `None` is interpreted as selecting the step according to the dtype
+            of the space. ``step = 1e-8`` for ``float64`` spaces and
+            ``step = 1e-4`` for ``float32`` spaces.
+
+        Examples
+        --------
+        Compute a numerical estimate of the derivative of the L2^2 norm:
```
adler-j: derivative of the gradient aka hessian

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,323 @@
...
+
+        .. math::
+            \\frac{A(x + h) - A(x)}{h}
+
+        if ``method='central'`` is used:
+
+        .. math::
+            \\frac{A(x + h/2) - A(x - h/2)}{h}
+
+        The number of function evaluations is ``functional.domain.size + 1`` if
```
adler-j: Copy paste error, in this case it is always two evaluations.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,323 @@
...
+        if step is None:
+            self.step = np.sqrt(np.finfo(operator.domain.dtype).eps)
+        else:
+            self.step = float(step)
+
+        self.method = str(method).lower()
+        if method not in ('backward', 'forward', 'central'):
+            raise ValueError('`method` not understood')
+
+        # TODO: should we really set linear to True here?
```
adler-j: Edit: as this is computed it is actually linear, at least under scalar multiplications, so I'd keep this. It is also supposed to masquerade as a linear operator in general.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,326 @@
...
+
+from odl.solvers.functional.functional import Functional
+from odl.operator.operator import Operator
+from odl.space.base_ntuples import FnBase
+
+
+__all__ = ('NumericalDerivative', 'NumericalGradient',)
+
+
+class NumericalDerivative(Operator):
```
adler-j: This is technically not functional related, perhaps it should be moved to `odl.operators`

odl/solvers/functional/derivatives.py
kohr-h: skip the "aka"

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,326 @@
...
+        Examples
+        --------
+        Compute a numerical estimate of the derivative of the L2^2 norm:
+
+        >>> import odl
+        >>> space = odl.rn(3)
+        >>> func = odl.solvers.L2NormSquared(space)
+        >>> grad = NumericalGradient(func)
+        >>> hess = grad.derivative([1, 1, 1])
+        >>> hess([1, 0, 0])
```
kohr-h: This example is not correct in the way I described above. The Hessian is usually the bilinear functional that appears as the second term in the Taylor expansion of an operator (or function) around a fixed point. Most of the time, this functional is identified with its operator representation, which in turn is identified with its matrix representation (for a finite-dimensional space) in the sense that
```
H(p)(x, y) = <H_op(p) x, y> = x_vec^T H_mat(p) y_vec
```
The relation to the gradient is as follows: if `S` is our functional, then its derivative is represented by its gradient via
```
S'(p) x = <grad S(p), x>
```
and in the Taylor expansion one has
```
S(p + h * x) ~ S(p) + h S'(p) x + o(h) = S(p) + h <grad S(p), x> + o(h)
```
Now, the second derivative of `S` in `p` is the derivative of `S'(p)` *in the same point `p`*, in order to make sense in the Taylor formula:
```
S(p + h * x) ~ S(p) + h S'(p) x + h^2 S''(p)(x, x) + o(h^2)
```
(Bilinear operators are uniquely determined by their diagonal values.)
How does that relate to the gradient? Let's differentiate it: for all `y`, it is
```
[S'(p + h * x)y - S'(p)y] / h --> S''(p)(x, y)
```
which can be translated to gradients:
```
<grad S(p + h * x) - grad S(p), y> / h --> <(grad S)'(p) x, y>
```
This shows that the derivative of the gradient of `S` in `p` is the same as the bilinear functional `S''(p)` in the sense of identification with the operator in an operator-weighted inner product. However, this only holds if we choose the same point `p` in the last step!

When I think about it, this would be a good fit for the math documentation, at the very least it would be useful for us to recall the proper settings of gradients and derivatives of functionals in Hilbert spaces.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,326 @@
...
+               [ 0.,  2.,  0.],
+               [ 0.,  0.,  2.]])
+        """
+        return NumericalDerivative(self, point,
+                                   method=self.method, step=np.sqrt(self.step))
+
+if __name__ == '__main__':
+    # pylint: disable=wrong-import-position
+    from odl.util.testutils import run_doctests
+    run_doctests()
```
kohr-h: The doctests are all good, but some tests with more complex examples and especially data types (including complex!) feel necessary.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,326 @@
...
+    --------
+    NumericalGradient : Compute gradient of a functional
+    """
+
+    def __init__(self, operator, point, method='forward', step=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        functional : `Operator`
```
kohr-h: operator

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,326 @@
...
+            range must be `FnBase` spaces.
+        point : ``functional.domain`` `element-like`
+            The point to compute the derivative in.
+        method : {'backward', 'forward', 'central'}
+            The method to use to compute the derivative.
+        step : `None` or float
+            The step length used in the derivative computation.
+            `None` is interpreted as selecting the step according to the dtype
+            of the space. ``step = 1e-8`` for ``float64`` spaces and
+            ``step = 1e-4`` for ``float32`` spaces.
```
kohr-h: Alternatively, be less specific here with numbers, you could just state the implementation, namely the square root of the data type precision.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,326 @@
...
+
+        Notes
+        -----
+        If the operator is :math:`A(x)` and stepsize :math:`h` is used, the
+        gradient is computed as follows.
+
+        If ``method='backward'`` is used:
+
+        .. math::
+            \\frac{A(x) - A(x - h)}{h}
```
kohr-h: Or roughly at least.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,323 @@
...
+        >>> grad = NumericalGradient(func, step=0.5)
+        >>> grad([1, 1, 1])
+        rn(3).element([2.5, 2.5, 2.5])
+
+        But it can be improved by using the more accurate ``method='central'``:
+
+        >>> grad = NumericalGradient(func, method='central', step=0.5)
+        >>> grad([1, 1, 1])
+        rn(3).element([2.0, 2.0, 2.0])
+
```
adler-j: Would probably go into the examples folder

odl/solvers/functional/derivatives.py
adler-j: > we don't have an adjoint anyway

I wonder if we can rectify this, I'll think about it.

Other than that I agree.

odl/solvers/functional/derivatives.py
adler-j: This rather length writeup certainly could make do for the math documentation. Also we should write down some of the rules that we make use of (the chain rule for the frechet derivative is not obvious to everyone).

I opened an issue on this here #632.

adler-j:
Updated according to above comments, only lacking a more complicated example but I'd personally leave that for now.

odl/solvers/functional/derivatives.py
kohr-h: > Write something about this choice in the class documentation, at the very top since it's important.

How about that? A single sentence would be enough.

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,330 @@
...
+from odl.operator import Operator
+from odl.space.base_ntuples import FnBase
+
+
+__all__ = ('NumericalDerivative', 'NumericalGradient',)
+
+
+class NumericalDerivative(Operator):
+
+    """The derivative of a operator by forward differences.
```
kohr-h: outdated
an

odl/solvers/functional/derivatives.py
```diff
@@ -0,0 +1,330 @@
...
+
+        ``method='forward'``:
+
+        .. math::
+            \\frac{f(x + h) - f(x)}{h}
+
+        ``method='central'``:
+
+        .. math::
+            \\frac{f(x + h/2) - f(x - h/2)}{h}
```
kohr-h: This is not the same as the code. What happens is (here for backward)
```
(\\nabla f(x))_i = \\frac{f(x) - f(x - h e_i)}{h}
```
for all standard unit vectors `e_i`.

odl/solvers/functional/derivatives.py
adler-j: Will do

adler-j:
Rebased from master, merge after CI.

PR #625, opened 2016-09-28
TST: avoid random failures in functional tests

### LINKED ISSUE #622: Tests fail for indicator of Lp unit ball
From [this build log](https://travis-ci.org/odlgroup/odl/jobs/163033638):
```
    def test_indicator_lp_unit_ball(space):
        """Test for indicator function on unit ball."""
        x = noise_element(space)
        one_elem = space.one()

        # Used due to numerical errors
        accuracy = 1 - 10**(-4)

        # Exponent = 1
        exponent = 1
        func = odl.solvers.IndicatorLpUnitBall(space, exponent)

        # Evaluation of the functional
        one_norm = np.abs(x).inner(one_elem)
 
        norm_larger_than_one = 1.1 * x / one_norm
        assert func(norm_larger_than_one) == np.inf

        norm_less_than_one = 0.9 * x / one_norm
        assert func(norm_less_than_one) == 0

        # Slightly less than 1 due to potential numerical errors
        norm_equal_to_one = accuracy * x / one_norm
        assert func(norm_equal_to_one) == 0

        # Negative, noninteger power
        exponent = -1 * np.random.rand()
        func = odl.solvers.IndicatorLpUnitBall(space, exponent)

        # Evaluation of the functional
        p_norm_x = np.power(np.power(np.abs(x), exponent).inner(one_elem),
                            1 / exponent)

        norm_larger_than_one = 1.1 * x / p_norm_x
        assert func(norm_larger_than_one) == np.inf

        norm_less_than_one = 0.9 * x / p_norm_x
>       assert func(norm_less_than_one) == 0
E       assert inf == 0
E        +  where inf = IndicatorLpUnitBall: rn(10) -> RealNumbers()(rn(10).element([inf, inf, -inf, ..., -inf, inf, -inf]))

test/solvers/functional/default_functionals_test.py:134: AssertionError

----------------------------- Captured stderr call -----------------------------
/home/travis/build/odlgroup/odl/odl/set/space.py:564: RuntimeWarning: divide by zero encountered in double_scalars
  return self.space.lincomb(1.0 / other, self, out=tmp)
/home/travis/build/odlgroup/odl/odl/solvers/functional/default_functionals.py:187: RuntimeWarning: divide by zero encountered in power
  x_norm = np.power(tmp.inner(self.domain.one()), 1 / self.exponent)
```
It seems like this randomized exponent thing is a bad idea and can result in arbitrary failures in case of bad luck. Better to test a fixed non-integer value.

kohr-h:
Using constants or fixtures for tests where scalar values matter. Random values are prone to lead to random failures (especially for `sigma` in proximals and exponent values).

While testing the final changes, I got another random failure in the `diff_ops_test`, which was, as usual, a too strict test for CUDA (i.e. single precision arithmetic). We should try to solve that problem in general in some elegant way.

Closes #622

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@ target/
 .spyproject/
 .spyderproject
 .spyderworkspace
+.spyproject/
 
 # PyCharm project files
 .idea
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -1062,21 +1062,22 @@ class OperatorSum(Operator):
         rn(3).element([2.0, 4.0, 6.0])
         """
         if left.range != right.range:
-            raise TypeError('operator ranges {!r} and {!r} do not match'
-                            ''.format(left.range, right.range))
+            raise OpTypeError('operator ranges {!r} and {!r} do not match'
+                              ''.format(left.range, right.range))
         if not isinstance(left.range, (LinearSpace, Field)):
-            raise TypeError('`left.range` {!r} not a `LinearSpace` or `Field` '
-                            'instance'.format(left.range))
+            raise OpTypeError('`left.range` {!r} not a `LinearSpace` or '
+                              '`Field` instance'.format(left.range))
         if left.domain != right.domain:
-            raise TypeError('operator domains {!r} and {!r} do not match'
-                            ''.format(left.domain, right.domain))
+            raise OpTypeError('operator domains {!r} and {!r} do not match'
+                              ''.format(left.domain, right.domain))
 
         if tmp_ran is not None and tmp_ran not in left.range:
-            raise TypeError('`tmp_ran` {!r} not an element of the operator '
-                            'range {!r}'.format(tmp_ran, left.range))
+            raise OpRangeError('`tmp_ran` {!r} not an element of the operator '
+                               'range {!r}'.format(tmp_ran, left.range))
         if tmp_dom is not None and tmp_dom not in left.domain:
-            raise TypeError('`tmp_dom` {!r} not an element of the operator '
-                            'domain {!r}'.format(tmp_dom, left.domain))
+            raise OpDomainError('`tmp_dom` {!r} not an element of the '
+                                'operator domain {!r}'
+                                ''.format(tmp_dom, left.domain))
 
         super().__init__(left.domain, left.range,
                          linear=left.is_linear and right.is_linear)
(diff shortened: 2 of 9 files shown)


test/discr/diff_ops_test.py
```diff
@@ -287,6 +287,8 @@ def test_part_deriv(space, method, padding):
 def test_gradient(space, method, padding):
     """Discretized spatial gradient operator."""
 
+    places = 2 if space.dtype == np.float32 else 4
```
adler-j: will fail for complex64. Perhaps we need to create a utility for this?

test/solvers/functional/functional_test.py
```diff
@@ -145,244 +171,268 @@ def test_arithmetic():
...
 
-    # Test that we cannot evaluate the convex conjugate
+    # Test the convex conjugate raises
     with pytest.raises(NotImplementedError):
-        (func1 + func2).convex_conj(x)
+        func_sum.convex_conj(x)
 
 
 def test_functional_plus_scalar(space):
     """Test for sum of functioanl and scalar."""
```
adler-j: functioanl

test/solvers/functional/functional_test.py
```diff
@@ -433,19 +488,23 @@ def test_multiplication_with_vector(space):
...
     expected_result = y_other_space * func(x)
-    assert all_almost_equal(y_other_times_func(x), expected_result)
+    assert all_almost_equal(y_other_times_func(x), expected_result,
+                            places=places)
 
 
 def test_functional_linear_perturb(space):
     """Test for the functional f(.) + <y, .>."""
+    # Less strict checking for single precision
+    places = 3 if space.dtype == np.float32 else 5
```
adler-j: getting even more strongly in favor of some kind of utility here.

adler-j:
How about changing the `all_almost_equal` from taking `places=` (or perhaps in addition to) taking a `precision = 'low'` etc flag that scales with the dtype?

kohr-h:
I'm a bit puzzled by the failure of the `IndicatorLpUnitBall` tests for Python 2 while Python 3 works as expected. I changed the factor for "element slightly smaller in norm than 1" from 0.9 to 0.99, and similarly from 1.1 to 1.01 for slightly larger. It seems to fail rather consistently for Python 2. Any idea why this is the case? Something different in the automatic float precision behavior  in Py3?
I can change back to 0.9 and 1.1, but I actually think the tests shouldn't fail. See [here](https://travis-ci.org/odlgroup/odl/jobs/163383792#L827) for the start of the fail log.

kohr-h:
> How about changing the all_almost_equal from taking places= (or perhaps in addition to) taking a precision = 'low' etc flag that scales with the dtype?

That's doable, I'll test. The same won't work for scalar results and `almost_equal`, though, since results are often times cast to Python `float` or `complex`, and the function has no chance to know in what precision the computation was performed. We have to stick with manual `places` (which should probably rather be called `significant_digits` or a shorter version of that) in those cases.

adler-j:
The discussion on changing to "precision" is out of scope for this pull imo. Rebase and merge. I also agree we should also make a new issue for `significant_digits`.

Issue #569, opened 2016-09-12
Migrate solvers to use functionals?


kohr-h:
In the same spirit as #566: Solvers should probably allow one or several functionals to be passed instead of passing proximals, gradients or similar directly. We need to assess the feasibility of the approach for the individual methods, but if possible, I think an additional optional argument to provide full-fledged functionals would be the way to go.

adler-j:
The hard part here is how to handle the two ways of calling the solvers. 

* One option is to simply force users to create a functional (even if it is a empty shell) and only accept functionals as arguments. We do get into issues with forcing users to create whole towers of them for some common cases however, e.g. `fun.convex_conjugate.proximal`.
* Another option is to close this issue and leave it as is.
* A third option is to allow users a choice, i.e. allowing them to provide the parameter in two ways. For example:
    * Allowing both in the same argument:
    ```
    f_cconj_prox : `Functional` or `callable`
        Functional with convex conjugate with proximal or proximal of convex conjugate of ``f``
    ```
    The problem here is obviously to handle determining which one is which.
    * Another option is to create two parameters and let users choose. Since they are supposed to be exclusive we need to make them both keyword arguments.
    ```
    f : `Functional` or `callable`
        Functional with convex conjugate. Mutually exclusive with ``f_cconj_prox``.
    f_cconj_prox : `Functional` or `callable`
        Proximal of convex conjugate of ``f``. Mutually exclusive with ``f``.
    ```

kohr-h:
> One option is to simply force users to create a functional (even if it is a empty shell) and only accept functionals as arguments. We do get into issues with forcing users to create whole towers of them for some common cases however, e.g. fun.convex_conjugate.proximal.

That's not an option for me, frankly. It puts unnecessary load on users with no substantial gain.



> Another option is to close this issue and leave it as is.

Surely a viable option.



> Allowing both in the same argument:

That becomes hard when you have lists of functionals / proximals and need to determine which is which.



> Another option is to create two parameters and let users choose. Since they are supposed to be exclusive we need to make them both keyword arguments.

That was the option I had in mind. It still requires some handling of different cases and error raising, but is the easiest way to distinguish between the different call patterns.

adler-j:
Full implementation of this depends at least partially on #564. Does anyone want to look at that? Specifically we need to figure out how to handle the L1 norm case, others can wait.

aringh:
[...]

PR #626, opened 2016-09-29
ENH: Add Limited memory BFGS, closes #621

### LINKED ISSUE #621: limited memory BFGS
We should implement the [Limited-memory BFGS](https://en.wikipedia.org/wiki/Limited-memory_BFGS) algorithm.

adler-j:
Added memory limited option to BFGS solver.

This is a merge into the #569 branch since it needs functionality from there.

Edit:
Added more functionality:

* Fixed legacy issues including a bug with the derivative of `ConstantOperator`. `ZeroOperator` is now a free standing operator.
* Added LBFGS example.


odl/operator/default_ops.py
```diff
@@ -834,99 +816,71 @@ def __str__(self):
...
-    def vector(self):
-        """The constant operator range element to subtract."""
-        return self.__vector
+        super().__init__(domain, range, linear=True)
 
     def _call(self, x, out=None):
-        """Evaluate the residual at ``x`` and write to ``out`` if given."""
-        if out is None:
-            out = self.operator(x)
+        """Return the constant vector or assign it to ``out``."""
```
aringh: Explicit would be to say `zero vector` and not `constant vector`.

odl/operator/operator.py
```diff
@@ -1158,6 +1155,97 @@ def __str__(self):
...
+        super().__init__(operator.domain, operator.range)
+
+    @property
+    def operator(self):
+        """The operator to apply."""
+        return self.__operator
+
+    @property
+    def vector(self):
+        """The constant operator range element to subtract."""
```
aringh: In the definition above and the `_call` below it looks like the vector is added.

odl/operator/operator.py
```diff
@@ -1158,6 +1155,97 @@ def __str__(self):
...
+        ----------
+        point : `domain` element
+            Any element in the domain where the derivative should be taken
+
+        Examples
+        --------
+        >>> import odl
+        >>> r3 = odl.rn(3)
+        >>> op = odl.IdentityOperator(r3)
+        >>> res = odl.OperatorVectorSum(op, r3.element([1, 2, 3]))
```
aringh: Very minor: I guess `res` is with "residual" in mind. However this operator is the sum.

odl/solvers/smooth/newton.py
```diff
@@ -34,6 +34,54 @@
...
+    Parameters
+    ----------
+    s : `sequence` of `LinearSpaceElement`
+        The ``s`` coefficients in the BFGS update, see Notes.
+    y : `sequence` of `LinearSpaceElement`
+        The ``y`` coefficients in the BFGS update, see Notes.
+    x : `LinearSpaceElement`
+        Point in which to evaluate the product.
+    hessinv_estimate : `Operator`, optional
+        Initial estimate of the hessian ``H0^-1``.
```
aringh: What happens if `hessinv_estimate` is not give? Taken as identity?

Edit: I guess this is only called from the bfgs method, and there it says that identity operator is default. Maybe that is enough...

adler-j:
Merged this into the other big branch (after fixing @aringh comments) review can be done over there instead.

PR #627, opened 2016-09-29
Issue 612  kullback leibler functionals

aringh:
This adds the kullback-leibler and the kullback-leibler corss entropy functionals. (Naming convention is taken in accordance with the previously implemented proximal operators).

It also implements the corresponding convex conjugate functionals. These are explicitly called `ConvexConjugate`, which have be criticized before in other functionals. However, I don't know what to call them: for me their "most basic property" is that they are the "convex conjugate functional"s of other, more well-known functionals. Any other suggestions for names?

Also: for kl the gradient is only defined for `x > 0`, since the const function is considered infinite for `x < 0`. However, the expression for the gradient makes sense as long as `0 no in x`. Is there an efficient way to test if all components of `x` are `> 0`? (similarly for kl convex conj, where I want to check that all components are `< 1`).

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -612,7 +612,7 @@ class KullbackLeibler(Functional):
         ----------
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
-        prior : ``space`` element, optional
+        prior : ``space`` `element-like`, optional
             Data term, positive.
             Default: if None it is take as the one-element.
         """
@@ -711,8 +711,6 @@ class KullbackLeiblerConvexConj(Functional):
     where :math:`g` is the prior, and :math:`I_{1_X - x \geq 0}(x)` is the
     indicator function on :math:`x \leq 1`.
 
-    This is the convex conjugate functional of `KullbackLeibler`.
-
     See Also
     --------
     KullbackLeibler : convex conjugate functional
@@ -725,7 +723,7 @@ class KullbackLeiblerConvexConj(Functional):
         ----------
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
-        g : ``space`` element, optional
+        g : ``space`` `element-like`, optional
             Data term, positive.
             Default: if None it is take as the one-element.
         """
@@ -846,7 +844,7 @@ class KullbackLeiblerCrossEntropy(Functional):
         ----------
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
-        prior : ``space`` element, optional
+        prior : ``space`` `element-like`, optional
             Data term, positive.
             Default: if None it is take as the one-element.
         """
@@ -952,8 +950,6 @@ class KullbackLeiblerCrossEntropyConvexConj(Functional):
 
     where :math:`g` is the prior.
 
-    This is the convex conjugate functional of `KullbackLeiblerCrossEntropy`.
-
     See Also
     --------
     KullbackLeiblerCrossEntropy : convex conjugate functional
@@ -966,7 +962,7 @@ class KullbackLeiblerCrossEntropyConvexConj(Functional):
         ----------
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
-        g : ``space`` element, optional
+        g : ``space`` `element-like`, optional
             Data term, positive.
             Default: if None it is take as the one-element.
         """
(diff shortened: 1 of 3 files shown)


odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,467 @@ def __init__(self, space):
...
+
+                The gradient is not defined in for points with components less
+                than or equal to zero.
+                """
+                if functional.prior is None:
+                    tmp = np.log(x)
+                else:
+                    tmp = np.log(x / functional.prior)
+
+                if any(np.isnan(tmp)) or -np.inf in tmp:
```
aringh: I use the `nan`-test to catch if any `x` is `< 0` and the `-inf`-test to catch if any `x` is = 0 (`np.log` is defined differently in these cases).

aringh:
Issues addressed in the new commit. Good to go? :)

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
+    The functional is given by
+
+    .. math::
+        \\sum_i \left(-g_i \ln(pos({1_X}_i - x_i)) \\right) +
+        I_{1_X - x \geq 0}(x)
+
+    where :math:`g` is the prior, and :math:`I_{1_X - x \geq 0}(x)` is the
+    indicator function on :math:`x \leq 1`.
+
+    This is the convex conjugate functional of `KullbackLeibler`.
```
adler-j: Obvious from the See Also

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
+    """
+
+    def __init__(self, space, prior=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        g : ``space`` element, optional
```
adler-j: Might as well go for 

```
``space`` `element-like`
```

and call `self.__prior = self.domain.element(prior)` further down, instead of the "if prior in space" thing.

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
+    """
+
+    def __init__(self, space, prior=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        prior : ``space`` element, optional
```
adler-j: See above comment on space element-like

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
```
adler-j: See above comment on space element-like

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
+    """
+
+    def __init__(self, space, prior=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        g : ``space`` element, optional
```
adler-j: See above comment on space element-like

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
+    Notes
+    -----
+    The functional is given by
+
+    .. math::
+        \\sum_i g_i (exp(x_i) - 1)
+
+    where :math:`g` is the prior.
+
+    This is the convex conjugate functional of `KullbackLeiblerCrossEntropy`.
```
adler-j: "duplicate" from See Also

odl/solvers/functional/default_functionals.py
```diff
@@ -570,6 +573,465 @@ def __init__(self, space):
...
+        else:
+            tmp = (self.prior * (np.exp(x) - 1)).inner(self.domain.one())
+        return tmp
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional."""
+        functional = self
+
+        class KLCrossEntCCGradient(Operator):
```
adler-j: Add a todo of replacing this once the `UFuncOperators` come in. Also that branch needs a review :/

aringh:
Updated :) I will merge after test.

PR #630, opened 2016-09-30
ENH: Add poisson_noise, closes #629

### LINKED ISSUE #629: Add function for poisson noise
We should have an option to create a vector with Poisson noise.

adler-j:


odl/phantom/noise.py
```diff
@@ -25,17 +25,62 @@
...
+__all__ = ('white_noise', 'poisson_noise')
 
 
 def white_noise(space):
     """Standard gaussian noise in space, pointwise N(0, 1).
 
+    Parameters
+    ----------
+    space : `FnBase`
+        The space that the noise should be created in
```
kohr-h: dot

kohr-h: Convoluted sentence - reads easier as `The space in which the noise should be created.`
BTW, why "should be"? Why not "is"?

odl/phantom/noise.py
```diff
@@ -25,17 +25,62 @@
 import numpy as np
 
 
-__all__ = ('white_noise',)
+__all__ = ('white_noise', 'poisson_noise')
 
 
 def white_noise(space):
     """Standard gaussian noise in space, pointwise N(0, 1).
```
kohr-h:
```
``N(0, 1)``
```

kohr-h: Can't we have at least a `sigma` or `stddev` parameter so users don't have to think about proper scaling all the time? Shifting is trivial and not so important as feature. Both would be a simple change by using `np.random.normal(loc=mean, scale=stddev, size=space.shape)`.

odl/phantom/noise.py
```diff
@@ -25,17 +25,62 @@
...
+        The space that the noise should be created in
+
+    Returns
+    -------
+    white_noise : ``space`` element
+
+    See Also
+    --------
+    poisson_noise
+    numpy.random.randn
```
kohr-h: `numpy.random.normal`

odl/phantom/noise.py
```diff
@@ -25,17 +25,62 @@
...
+        The intensity (usually called lambda) parameter of the noise.
+
+    Returns
+    -------
+    poisson_noise : ``intensity.space`` element
+        Poisson distributed random variable.
+
+    Notes
+    -----
+    The probability density function for value :math:`k` and intensity
```
kohr-h: Weird formulation. Why speak of a function when it is a discrete probability measure? My suggestion is to use a phrasing like 
```
For a Poisson distributed random variable :math:`X` with intensity
:math:`\\lambda`, the probability of it taking the value 
:math:`k \\in \mathbb{N}_0` is given by

    .. math::
        P(X = k) = \\frac ...
```

odl/phantom/noise.py
adler-j: I guess we can do that.

PR #636, opened 2016-10-04
REL: add test files to MANIFEST.in, closes #634

### LINKED ISSUE #633: odl.test() without pytest throws ImportError
Currently you get a nondescript `no package named pytest` error if it is not installed. Perhaps we should catch this and give a more descriptive error.

### LINKED ISSUE #634: Tests not installed by pip
`pip install odl` does not install the test package, only the `odl` package. Because of this we cannot do `python -c "import odl; odl.test()"` as written in the introduction.

kohr-h:
Obviously, we want to make a new minor version for this and upload to PyPI.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,21 @@
 include CONTRIBUTING.md
 include LICENSE
+include pytest.ini
 include README.md
 include requirements.txt
 include setup.cfg
 include setup.py
 include test_requirements.txt
+
+exclude .coveragerc
+exclude .pylintrc
+exclude .travis.yml
+exclude .landscape.yml
+exclude appveyor.yml
+
+recursive-include test *.py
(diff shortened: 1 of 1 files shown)


kohr-h:
Actually the fix is not as easy as this. The install location for `odl` is `path/to/lib/python<ver>/site-packages/`, where the contents of the `odl` folder are installed (no top-level stuff). So there is no natural place to put the tests -- they make no sense in the site-packages dir, and one level down in `odl` they obstruct the package layout.
I suggest closing this PR and instead removing the hint from the documentation about testing with the pip-installed package.

adler-j:
I suggest we do it like other packages tend to do, move the `test` folder to `odl/test`. That way it gets installed naturally.

adler-j:
Closed, see #638 for an alternative implementation.

PR #638, opened 2016-10-06
MAINT: Move tests to odl/test folder, closes #634

### LINKED ISSUE #634: Tests not installed by pip

adler-j:
An alternative to #636. This solves the problem by moving tests into the odl folder, similar to how most other major packages seem to do it.


kohr-h:
Looks good, tests run on my machine. Go ahead.

adler-j:
Merge after CI.

PR #653, opened 2016-10-14
Issue 515  pytest 30 take 2

adler-j:
New PR instead of #516.


conftest.py
```diff
@@ -67,11 +79,9 @@ def fn_impl(request):
 @pytest.fixture(scope="module", ids=ntuples_impl_ids,
                 params=ntuples_impl_params)
 def ntuples_impl(request):
+    """String with a available `NtuplesBase` implementation name."""
```
kohr-h: an

conftest.py
```diff
@@ -94,20 +105,47 @@ def floating_dtype(request):
...
+
+    Returns
+    -------
+    name : `str`
+        Name of the ufunc.
+    n_in : `int`
+        Number of input values of the ufunc.
+    n_out : `int`
+        Number of output values of the ufunc.
+    doc : `str`
```
kohr-h: No ticks for `int` and `str`

conftest.py
```diff
@@ -94,20 +105,47 @@ def floating_dtype(request):
...
 
 @pytest.fixture(scope="module", ids=reduction_ids, params=reduction_params)
 def reduction(request):
+    """Tuple with information on a reduction.
+
+    Returns
+    -------
+    name : `str`
+        Name of the reduction.
+    doc : `str`
```
kohr-h: same here

kohr-h:
Update also `.travis.yml` and `setup.py` (find in files `py\.test`)

adler-j:
Should be nothing in .travis.yml and setup.py already fixed

adler-j:
Merge after CI

kohr-h:
> Should be nothing in .travis.yml

There's a call to `py.test` which should be changed to `pytest`. Do that on master.

adler-j:
py.test is still valid, not sure if they are scrapping it but sure will do.

PR #664, opened 2016-10-17
Issue 121  install docs

kohr-h:
More extensive and updated installation docs, and a restructured "Getting started" section.


doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+If you choose to use your system Python interpreter (the "pip install as user" variant), it may be a good idea to stick with the default one, i.e. the one invoked by ``python``.
+Otherwise, we recommend using Python 3, since Python 2 support will be discontinued in 2020.
+
+Development enviroment
+----------------------
+Since ODL is object-oriented, using an Integrated Development Environment (IDE) is recommended, but not required.
+The most popular ones are `Spyder`_ which works on all major platforms and can be installed through both ``conda`` and ``pip``, and `PyCharm`_ which can be integrated with any text editor of your choice, such as Emacs or Vim.
+
+
+Alternative 1: Installing a release using ``conda`` (recommended for users)
```
adler-j: Have you tried this on a VM? Does it work on windows?

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+===========================================================================
+
+TL;DR
+-----
+- Download and install `Miniconda`_
```
adler-j: Why not the full anaconda? If you need a guide like this you might as well get it.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+
+    $ conda install scikit-image
+
+- To run unit tests:
+
+  .. code-block:: bash
+
+    $ conda install pytest
+
+
```
adler-j: Lots of newlines

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+- Install ODL and dependencies:
+
+  .. code-block:: bash
+
+    $ pip install odl[show,fftw,pywavelets,scikit,testing]
+
+Install a Python interpreter
+----------------------------
+Open a terminal and type ``python`` + Enter.
+If a Python prompt appears, you already have an interpreter installed and can skip this step.
```
adler-j: Perhaps hint that they would exit, we dont want users to get stuck here :D

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+
+In the unlikely event that Python is not installed, consult your distro package manager.
+
+**On MacOS:**
+
+Get the latest release (2 or 3) for MacOS `here <https://www.python.org/downloads/mac-osx/>`_ and install it.
+
+**On Windows:**
+
+Installers can be downloaded from `this link <https://www.python.org/downloads/windows/>`_.
```
adler-j: We should be clear that anaconda is even more recommended on windows since installing numpy etc is hard as f.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+Install ODL and its dependencies
+--------------------------------
+You need to `install pip`_ to be able to install ODL and its dependencies from the `Python Package Index`_ (PyPI).
+Check if running ``pip`` (alternatively: ``pip3``) shows a help message or an error.
+
+For basic installation without extra dependencies, run
+
+.. code-block:: bash
+
+   $ pip install --user odl
```
adler-j: This will fail on windows with the python above (to my knowledge) quite sure the dependencies to build numpy are not available.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+
+**On Windows:**
+
+Installers can be downloaded from `this link <https://www.python.org/downloads/windows/>`_.
+You should pick the latest release for your favorite version (2 or 3).
+
+
+Install ODL and its dependencies
+--------------------------------
+You need to `install pip`_ to be able to install ODL and its dependencies from the `Python Package Index`_ (PyPI).
```
adler-j: http://stackoverflow.com/questions/4750806/how-do-i-install-pip-on-windows

> Good news! Python 3.4 (released March 2014) and Python 2.7.9 (released December 2014) ship with Pip. This is the best feature of any Python release.

So anyone installing python themselves should have pip, this is only a problem on old distros (like ubuntu 14). We should be clear in this, i.e. "You **may** need to"

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,397 @@
...
+For basic installation without extra dependencies, run
+
+.. code-block:: bash
+
+   $ pip install --user odl
+
+
+**Extra dependencies**
+
+Additional dependencies can be specified in square brackets, separated by commas (no spaces!):
```
adler-j: Be clear that these are not needed (see Frank trying to install all of them)

doc/source/guide/numpy_guide.rst
```diff
@@ -77,35 +77,35 @@ The fact that the ``x.ufunc.negative()`` interface is needed is a known issue wi
...
+   ...         self.vector = vector
+   ...
+   ...         # Initialize operator base class.
+   ...         # This operator maps from the space of vector to the same space and is linear
+   ...         odl.Operator.__init__(self, domain=vector.space, range=vector.space,
+   ...                               linear=True)
+   ...
+   ...     def _call(self, x):
+   ...         # The output of an Operator is automatically cast to an ODL vector
+   ...         return np.convolve(x, self.vector, mode='same')
```
adler-j: It seems you have removed all my changes here. This should be fftconvolve

doc/source/getting_started/installing.rst
kohr-h: I don't have a VM with a Windows more up to date than XP, so I was hoping that some Windows user could perhaps test it?^^

doc/source/getting_started/installing.rst
kohr-h: I think it's just waste of time and bandwidth. When installing ODL it will update the relevant packages anyway, so you can as well go with the small download.

doc/source/getting_started/installing.rst
kohr-h: Will be formatted away, but I'll delete one for you.

doc/source/getting_started/installing.rst
kohr-h: Haha, I'll add a sentence.

kohr-h:
Okay, all comments fixed -- as far as I understood them. I also emphasized some things more by using `.. warning::` and `.. note::` directives, gives nice boxes in the final doc. I also tried to be more clear in places that were unclear before, and hopefully the structure is now easier to spot.

Merge?

doc/source/getting_started/about_odl.rst
```diff
@@ -158,29 +138,24 @@ right-hand side is 0.14, so we choose 0.1.
     >>> for i in range(niter):
     ...     residual = matrix_op(reco) - data
     ...     reco -= 0.1 * matrix_op.adjoint(residual)
-    ...     print(residual.norm())
-    1.41421356237
-    0.583095189485
-    0.240416305603
-    0.0991261822124
-    0.0408707719526
+    ...     print('{:.3}'.format(residual.norm()))
```
kohr-h: This was needed since Python 2 and 3 never print the same way automatically.

doc/source/getting_started/first_steps.rst
```diff
@@ -206,23 +207,13 @@ Perhaps a bit better, but far from excellent.
...
-Consulting the `douglas_rachford_pd` documentation we see that it solves problems of the form:
 
 .. math::
-    \min_x f(x) + \sum_{i=1}^n g_i(L_i x),
 
-where :math:`f`, :math:`g_i` are convex functions, :math:`L_i` are linear `Operator`'s. 
-By identification, we see that the above problem can be written in this form if we let `f` be the indicator function on :math:`[0, 1]`, 
-:math:`g_1` be the squared l2 distance :math:`\| \cdot - g\|_2^2`, 
-:math:`g_2` be the norm :math:`\| \cdot \|_1`, 
-:math:`L_1` be the convolution operator and :math:`L_2` be the gradient operator.
```
kohr-h: Damnit, this got auto-merged in the wrong way. I'll fix it.

doc/source/guide/numpy_guide.rst
kohr-h: This is only a small example, I'll keep `np.convolve`. Mostly because otherwise the results below are messed up, e.g.
```
File "/home/hkohr/git/odl/doc/source/guide/numpy_guide.rst", line 114, in numpy_guide.rst
Failed example:
    composed_op(x)
Expected:
    26.0
Got:
    26.000000000000004
```

doc/source/getting_started/first_steps.rst
kohr-h: > All of it is wrong, look further down.

What do you mean? Is d1c2a49 not getting it right? I don't see any duplicates.

doc/source/guide/numpy_guide.rst
adler-j: ok!

doc/source/getting_started/first_steps.rst
```diff
@@ -232,27 +233,27 @@ e.g. `forward_backward_pd`, `chambolle_pock_solver`, etc in the ODL `examples/so
...
-   g_funcs = [odl.solvers.L2NormSquared(space).translated(g),
-              a * odl.solvers.L1Norm(grad.range)]
-   
-   # Functional of the bound constraint 0 <= f <= 1
-   f = odl.solvers.IndicatorBox(space, 0, 1)
-   
+
+   # Create proximals operators corresponding to the convex conjugate (cconj)
+   # of the l2 squared and l1 norms.
+   prox_cc_g = [odl.solvers.proximal_cconj_l2_squared(space, g=g),
```
adler-j: This is where I meant that the merge failed, you fell back to using proximals

doc/source/getting_started/first_steps.rst
```diff
@@ -232,27 +233,27 @@ e.g. `forward_backward_pd`, `chambolle_pock_solver`, etc in the ODL `examples/so
...
    opnorm_A = odl.power_method_opnorm(A, xstart=g)
    opnorm_grad = odl.power_method_opnorm(grad, xstart=g)
    sigma = [1 / opnorm_A**2, 1 / opnorm_grad**2]
    tau = 1.0
-   
+
    # Solve using the Douglas-Rachford Primal-Dual method
    x = space.zero()
-   odl.solvers.douglas_rachford_pd(x, f, g_funcs, lin_ops,
+   odl.solvers.douglas_rachford_pd(x, prox_f, prox_cc_g, lin_ops,
```
adler-j: Also here

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+.. _alternative_2:
+
+*************************************************
+Alternative 2: Installing a release using ``pip``
+*************************************************
+
+TL;DR
+=====
+- Install `pip`_
+- Install `FFTW`_
```
adler-j: Does this really need to be in the TLDR? I find it very un-important since it is only an optimization that rather advanced users could take advantage of, numpy has the same functionality.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+You may need to `install pip`_ to be able to install ODL and its dependencies from the `Python Package Index`_ (PyPI).
+If running ``pip`` (alternatively: ``pip3``) an shows a help message, it is installed -- otherwise you need to install it first.
+
+For basic installation without extra dependencies, run
+
+.. code-block:: bash
+
+   $ pip install --user odl
+
+.. warning::
```
adler-j: Perhaps a single warning (the one above) is enough.

doc/source/getting_started/installing.rst
kohr-h: Not necessarily, but then they cannot use `pip install odl[all]`.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+- ``scikit`` : Install `scikit-image`_ as a simple backend for ray transforms.
+- ``proximal``: Install the `ProxImaL`_ convex optimization package.
+
+
+These dependencies are optional and may not be easy to install on your system (especially on Windows).
+In general, a clean ODL installation is enough for most users' initial needs.
+
+
+------
+
```
adler-j: .. alternative 3_

doc/source/getting_started/installing.rst
kohr-h: Yeah, it becomes a bit too much perhaps. Downgrade to note or remove?

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+   $ pip install --editable .
+
+Using only ``pip``
+==================
+Enter the top-level directory of the cloned repository and run
+
+.. code-block:: bash
+
+   $ pip install --user --editable .
```
adler-j: We changed from not having --user to having it, any reason?

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+Unit tests in ODL are based on `pytest`_, and coverage reports are created by the `coverage`_ module.
+
+Installing testing dependencies
+===============================
+
+Using ``conda``:
+----------------
+.. code-block:: bash
+
+    $ conda install pytest pytest-cov pytest-pep8
```
adler-j: Dislike this. Why not simply install the ODL dependencies? This way we can enforce dependencies etc. The way this is phrased atm would say "ok" if the user has pytest 2.x, but then the next step would fail since odl requires pytest>3.0.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+
+Using ``pip``:
+--------------
+
+.. code-block:: bash
+
+    $ pip install --user pytest pytest-cov pytest-pep8
```
adler-j: See above, we should do `pip install --user odl[testing]` and then perhaps add pytest-cov and pytest-pep8 to the testing requirements.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+
+Testing the code
+================
+Now you can check that everything was installed properly by running
+
+.. code-block:: bash
+
+   $ python -c "import odl; odl.test()"
+
+**anywhere but** in the top-level directory of an ODL clone.
```
adler-j: This comment only applies for users that have installed from source, so perhaps move it to after the "if you have installed from source" thing?

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+
+CUDA backend for linear arrays
+==============================
+The `odlcuda`_ backend for fast array calculations on CUDA requires the `CUDA toolkit`_ (on Linux: use your distro package manager) and a CUDA capable graphics card with compute capability of at least 5.0.
+Search `this table <https://en.wikipedia.org/wiki/CUDA#GPUs_supported>`_ for your model.
+
+Installation using ``conda``
+----------------------------
+.. note::
+    ``odlcuda`` is currently only available for Linux and Python 3.5.
```
adler-j: only available in conda for

doc/source/getting_started/installing.rst
kohr-h: How would you get pytest 2.x with those instructions?

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+
+After that, follow the `build instructions there <https://github.com/odlgroup/odlcuda.git>`_.
+
+
+ASTRA for X-ray tomography
+==========================
+To calculate fast forward and backward projections for image reconstruction in X-ray tomography, install the `ASTRA tomography toolbox`_.
+ASTRA projectors are fully supported in ODL.
+
+You can try using the conda package, but we can give no guarantee that it works out of the box:
```
adler-j: A link to their github for "for more help, please see..."

doc/source/getting_started/installing.rst
kohr-h: Not really, if you installed a release and run this in some (unrelated) clone, it will run the tests in the repo, not the installed ones.

adler-j:
Requested some improvements for clarity, looks great overall but obviously needs to be tested against reality by new people.

doc/source/getting_started/installing.rst
adler-j: Well then you have installed ODL from source *at some point*. I simply feel that this comment is not something people who use conda or pip need to read.

doc/source/getting_started/installing.rst
adler-j: Just wondering why we have different suggestions here

doc/source/getting_started/first_steps.rst
```diff
@@ -232,27 +233,27 @@ e.g. `forward_backward_pd`, `chambolle_pock_solver`, etc in the ODL `examples/so
...
                                    tau=tau, sigma=sigma, niter=100)
-   x.show('TV Douglas-Rachford', show=True)
-
+   x.show('TV Douglas-Rachford')
```
kohr-h: Damnit!

doc/source/getting_started/installing.rst
adler-j: I'm frankly in favor of removing `pip install odl[all]`, people like Frank seem to think that they should do it while it is likely the last thing any normal user should do.

doc/source/getting_started/installing.rst
kohr-h: Not even that. I suffices that you cloned the repo, it doesn't have to be installed.
Anyway, I'll make it a not and clarify that it only applies to source installations (move it).

doc/source/getting_started/installing.rst
adler-j: Well this line of code is independent of ODL, it doesnt mention it anywhere. Hence updating our dependencies would not make any difference here. Checking in code is indeed an option (that we should do), but it would only lead to users running this, getting an error and then having to run update anyway.

doc/source/getting_started/installing.rst
adler-j: Remove imo

doc/source/getting_started/installing.rst
kohr-h: Oh, but if it's in the dependencies then this step becomes obsolete altogether. The pytest stuff will be installed along with odl in the first place.

doc/source/getting_started/installing.rst
adler-j: My argument is that users should not install pyfftw unless they:

* Know what they are doing
* Have already tried the numpy fft and find it to be slow

doc/source/getting_started/installing.rst
adler-j: Valid point i guess. Then I'd also add matplotlib.

doc/source/getting_started/installing.rst
adler-j: See https://github.com/odlgroup/odl/issues/665, I suggest we remove [all]

doc/source/getting_started/installing.rst
kohr-h: Sure, maybe the `conda` package doesn't have to be that minimal.

doc/source/getting_started/installing.rst
kohr-h: Okay, that's the most robust solution.

doc/source/getting_started/about_odl.rst
```diff
@@ -1,60 +1,53 @@
...
 
 
 Set
 ===
 
-A `Set` is the fundamental building block of ODL objects. It mirrors the mathematical concept of a
-`set`_ in that it can tell if an object belongs to it or not:
+A `Set` is the fundamental building block of ODL objects. It mirrors the mathematical concept of a `set`_ in that it can tell if an object belongs to it or not:
 
     >>> interv = odl.IntervalProd(0, 1)
```
adler-j: Would be great with python typesetting of these, is that possible?

doc/source/getting_started/about_odl.rst
```diff
@@ -63,12 +56,9 @@ or :math:`\mathbb{C}`), and :math:`n` a positive integer. Example:
...
-and range (more accurately: `codomain`_), where :math:`\mathcal{Y}` is a vector space. The ODL
-implementation `FunctionSpace` covers only the cases :math:`\mathcal{Y} = \mathbb{R}` or
-:math:`\mathbb{C}` since the general case has large overlaps with `Operator`. Note that we do
-not make a distinction between different types of function spaces with respect to regularity,
-integrability etc. on an *abstract* level since there is no obvious way to check it.
+A `function space`_ is a set of functions :math:`f: \mathcal{X} \to \mathcal{Y}` with fixed domain and range (more accurately: `codomain`_), where :math:`\mathcal{Y}` is a vector space.
+The ODL implementation `FunctionSpace` covers only the cases :math:`\mathcal{Y} = \mathbb{R}` or :math:`\mathbb{C}` since the general case has large overlaps with `Operator`.
+Note that we do not make a distinction between different types of function spaces with respect to regularity, integrability etc. on an *abstract* level since there is no obvious way to check it.
 
 As linear spaces, function spaces support some interesting operations:
```
adler-j: While reading the example i found this line:

>>> ratio_func = exp_plus_one / exp  # x -> exp(x) / (exp(x) + 1)

The comment has to be a misstake, right?

doc/source/getting_started/about_odl.rst
```diff
@@ -102,51 +89,44 @@ functions.
...
-``f(x) = exp(-x)``, we can simply pass it to the ``element()`` method:
+Discretizations have a large number of useful functionality, for example the direct and vectorized sampling of continuously defined functions.
+If we, for example, want to discretize the function ``f(x) = exp(-x)``, we can simply pass it to the ``element()`` method:
 
     >>> exp_discr = l2_discr.element(lambda x: np.exp(-x))
     >>> type(exp_discr)
-    odl.discr.lp_discr.DiscreteLpElement
+    <class 'odl.discr.lp_discr.DiscreteLpElement'>
     >>> exp_discr.shape
     (5,)
```
adler-j: Here printing its values would probably satisfy the readers curiosity :)

doc/source/getting_started/about_odl.rst
```diff
@@ -158,29 +138,24 @@ right-hand side is 0.14, so we choose 0.1.
...
+    0.0991
+    0.0409
 
-If we now exchange ``matrix_op`` and ``data`` with a tomographic projector and line integral data,
-not a single line of code in the reconstruction method changes since the operator interface is
-exactly the same.
+If we now exchange ``matrix_op`` and ``data`` with a tomographic projector and line integral data, not a single line of code in the reconstruction method changes since the operator interface is exactly the same.
 
 
 Further features
```
adler-j: Here we should have doc linkage

doc/source/getting_started/about_odl.rst
```diff
@@ -158,29 +138,24 @@ right-hand side is 0.14, so we choose 0.1.
...
 * A unified structure for representing tomographic acquisition geometries
-* Interfaces to fast external libraries, e.g. ASTRA for X-ray tomography, STIR for emission
-  tomography (preliminary), pyFFTW for fast Fourier transforms, ...
+* Interfaces to fast external libraries, e.g. ASTRA for X-ray tomography, STIR for emission tomography (preliminary), pyFFTW for fast Fourier transforms, ...
 * A growing number of "must-have" operators like `Gradient`, `FourierTransform`, `WaveletTransform`
-* Several solvers for variational inverse problems, ranging from simple gradient methods to
-  the Chambolle-Pock method (more to come...)
-* Standardized tests for the correctness of implementations of operators and spaces, e.g. does
-  the adjoint operator fulfill its defining relation?
+* Several solvers for variational inverse problems, ranging from simple gradient methods to the Chambolle-Pock method (more to come...)
```
adler-j: Chambolle-pock isnt our most complicated. I'd say something about "nonsmooth convex solvers such as (list a few)" also remove "more to come".

doc/source/getting_started/first_steps.rst
```diff
@@ -1,8 +1,9 @@
...
-###############
+.. _first_steps:
+
+##################
+First steps
+##################
 
-Welcome to the ODL getting started guide.
 This guide is intended to give you a simple introduction to ODL and how to work with it.
 If you need help with a specific function you should look at the `ODL API reference <https://odlgroup.github.io/odl/odl.html>`_.
```
adler-j: A few lines down we write "whatver", spelling error.

adler-j: Also found while reading this: We use conjugate_gradient_normal, but the operator is self adjoint and we could use conjugate_gradient. Do we want to fix this?

doc/source/getting_started/first_steps.rst
adler-j: bumpty

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
+.. _installing_odl:
+
+##############
+Installing ODL
+##############
+
```
adler-j: A few words here or remove the "Introduction" title below. I find the style

Big Title
Small Title
text

To be quite hard to read and it feels like something was missed.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+Introduction
+************
+
+Installing ODL is intended to be straightforward, and this guide is meant for new users.
+For a minimal installation you should perform the following steps:
+
+1. Install a Python interpreter
+2. Install ODL and its dependencies
+3. (optional) Run the tests
+
```
adler-j: Here we should be more clear that if the reader already has a python installation that they are happy with, they can skip the "install python" steps and go straight to the install ODL step

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+TL;DR
+=====
+- Download and install `Miniconda`_
+- Create conda environment:
+
+  .. code-block:: bash
+
+    $ conda create -c odlgroup -n odl-py35 python=3.5 odl odlcuda matplotlib pytest scikit-image spyder
+
```
adler-j: Does the user need to activate the environment as well?

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+.. warning::
+
+    Correctly installing ODL's dependencies on Windows, especially Numpy and Scipy, can be quite a hassle, and we therefore discourage this variant.
+    You should really consider using Anaconda instead, see :ref:`alternative_1`.
+
+
+Installing ODL and its dependencies
+===================================
+You may need to `install pip`_ to be able to install ODL and its dependencies from the `Python Package Index`_ (PyPI).
+If running ``pip`` (alternatively: ``pip3``) an shows a help message, it is installed -- otherwise you need to install it first.
```
adler-j: ``pip2``/``pip3``

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,458 @@
...
+The following optional packages extend the functionality of ODL.
+They can be specified as keywords in square brackets, separated by commas (no spaces!):
+
+.. code-block:: bash
+
+   $ pip install odl[dep1,dep2]
+
+Possible choices:
+
+- ``all`` : Install with all extra dependencies.
```
adler-j: I.m.o. get the #665 in here as well while at it.

doc/source/getting_started/installing.rst
kohr-h: Ah, but further up we're in a `conda` environment where it makes no sense to use `--user` since the default install location is user-writable.

doc/source/getting_started/installing.rst
kohr-h: Frankly we can skip the cov and pep8 stuff here. They're only really interesting to developers.

doc/source/getting_started/about_odl.rst
kohr-h: Typo of course, good spot.

doc/source/getting_started/about_odl.rst
kohr-h: Nothing simpler that that :-)

doc/source/getting_started/first_steps.rst
kohr-h: Perhaps not. The normal equation variant is perhaps less common and appears more useful since it's more general. There's also the danger that users think `conjugate_gradient` makes no assumptions. I think this leads more into the right direction, even if it's not necessary.

doc/source/getting_started/installing.rst
kohr-h: I actually don't want to do that. Most Linux people have a Python interpreter, but I don't want them to miss the Anaconda recommendation. And they should know that they can skip things that they have installed themselves manually at some point.

kohr-h:
OK, I feel I have addressed all issues raised above. Maybe you want to check the conda env related thing @adler-j. Otherwise I'm quite happy with the overhaul, we should get this in asap.

doc/source/getting_started/installing.rst
adler-j: I would still add a line that implies it, even a fancy sounding:

> - Activate the conda environment and start working!

would be nice.

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,461 @@
...
+You may choose to download the `full Anaconda <https://www.continuum.io/downloads>`_ variant, but we recommend the slim `Miniconda`_ distribution since many of the packages included in full Anaconda are out of date anyway and need to be updated.
+Note that the choice of Python version (2 vs. 3) of the Anaconda installer is not very important since you can later choose to create conda environments with any Python version (see below).
+
+Make sure that during installation, your ``PATH`` variable is adapted such that ``conda`` and other scripts can be found by your shell::
+
+    Do you wish the installer to prepend the Miniconda3 install location
+    to PATH in your /home/user/.bashrc ? [yes|no]
+    [no] >>> yes
+
+After opening a new terminal (for the changed ``PATH`` variable to take effect), you can run
```
adler-j: I prefer `exec bash` to opening a new terminal

doc/source/getting_started/installing.rst
```diff
@@ -0,0 +1,461 @@
...
+
+    $ pip install proximal
+
+- To run unit tests:
+
+  .. code-block:: bash
+
+    $ conda install pytest
+
+
```
adler-j: Where did astra go here? They even have a conda package

doc/source/guide/chambolle_pock_guide.rst
```diff
@@ -0,0 +1,176 @@
...
+Now that the problem is set up, we need to select some optimization parameters.
+For the Chambolle-Pock method, there is one main rule that we can use:
+The product of the primal step :math:`\tau`, the dual step :math:`\sigma` and the squared operator norm :math:`\|L\|^2` has to be smaller than 1, :math:`\tau \sigma \|L\|^2 < 1`.
+Apart from this, there are no clear rules on how to select :math:`\tau` and :math:`\sigma` -- basically we're left with trial and error.
+We decide to pick them both equal to :math:`1 / \|L\|^2`.
+To calculate an estimate of the operator norm, we have the tool `power_method_opnorm` which performs the simple `power iteration <https://en.wikipedia.org/wiki/Power_iteration>`_ to approximate the largest singular value of :math:`L`:
+
+.. code-block:: python
+
+   >>> op_norm = 1.1 * odl.power_method_opnorm(L, maxiter=4, xstart=phantom)
```
adler-j: Much better to use `xstart=data` here, both because it is more realistic and because it will give a better answer

adler-j:
Minor stuff, fix and merge.

doc/source/getting_started/installing.rst
kohr-h: Okay, convinced :-)

doc/source/getting_started/installing.rst
kohr-h: Also, Ubuntu for example has its own shell "dash", so fixing bash would be hard-coding a non-obvious choice. And I don't know too much about shells on OSX.

doc/source/guide/chambolle_pock_guide.rst
```diff
@@ -0,0 +1,176 @@
...
+    >>> l2_norm_squared = odl.solvers.L2NormSquared(space).translated(data)
+    >>> l1_norm = 0.0003 * odl.solvers.L1Norm(grad.range)
+    >>> f = odl.solvers.SeparableSum(l2_norm_squared, l1_norm)
+
+.. note::
+    We don't need to take extra care of the :math:`L^1` norm being a vectorial norm since `L1Norm` also works on product spaces.
+
+Finally, we define the functional for the nonnegativity constraint, available as the functional `IndicatorNonnegativity`:
+
+    >>> g = odl.solvers.functional.IndicatorNonnegativity(space)
```
kohr-h: right

doc/source/guide/chambolle_pock_guide.rst
```diff
@@ -0,0 +1,176 @@
...
+
+Calling the solver
+^^^^^^^^^^^^^^^^^^
+Now that the problem is set up, we need to select some optimization parameters.
+For the Chambolle-Pock method, there is one main rule that we can use:
+The product of the primal step :math:`\tau`, the dual step :math:`\sigma` and the squared operator norm :math:`\|L\|^2` has to be smaller than 1, :math:`\tau \sigma \|L\|^2 < 1`.
+Apart from this, there are no clear rules on how to select :math:`\tau` and :math:`\sigma` -- basically we're left with trial and error.
+We decide to pick them both equal to :math:`1 / \|L\|^2`.
```
kohr-h: oops

doc/source/guide/chambolle_pock_guide.rst
kohr-h: But that happens to work only in this case since the operator domain and range are the same. I find it misleading to use a range element as input to the norm estimator.
Or - if we want to do that - we could accept a range element and start with the adjiont evaluation. Would be a new issue, though.

doc/source/getting_started/installing.rst
adler-j: I guess you are right. But perhaps go for "opening a new" to "restarting"? That kinda implies that `exec bash` is OK.

doc/source/getting_started/installing.rst
kohr-h: I'll add it with a link to the section all the way down.

PR #669, opened 2016-10-21
Distribute tests

kohr-h:
Unfortunately I realized too late that the tests are not included, neither in the `pip` nor the `conda` packages. The problem is basically that `odl/test` is not a package and thus not included. We actually explicitly excluded anything containing "test" in `setup.py`, ahem.

Anyway, it turns out that the `python setup.py sdist` and `bdist_wheel` commands can rather easily be convinced to include extra stuff using the `package_data` parameter AND recursively adding `odl/test` in `MANIFEST.in`. However, for `python setup.py install`, you also have to give the `include_package_data=True` option (of course, why would I want to install with data when I explicitly give `package_data`???).

Next issue: `conftest.py` is not installed, and without it, the extra plugins (`--largescale` and such) don't work. No test runs since the options in the corresponding files are not recognized. Putting `conftest.py` one level below didn't work either for a reason I don't remember. Here the solution was to go for the second item [here](http://doc.pytest.org/en/latest/writing_plugins.html#plugin-discovery-order-at-tool-startup) and follow the instructions [here](http://doc.pytest.org/en/latest/writing_plugins.html#making-your-plugin-installable-by-others) to make an entry point for `pytest` -- which now resides in `odl.util.pytest_plugins.py`. Here the description was actually very good, and the approach worked immediately.

So now that this is done, I think we need a new release 0.5.1 (sorry about that). I'll put the correct tag into `conda/meta.yaml` and merge when we agree on the changes.


kohr-h:
Perhaps it's not too bad that Travis CI was down due to connection problems at GitHub (hard to say whose fault), so the new installation doc is not yet deployed. Seems fixed now.

kohr-h:
See [Travis CI status](https://www.traviscistatus.com/)

kohr-h:
Another remark: Sudden pytest warnings are not my fault:

https://github.com/pytest-dev/pytest/issues/2005

adler-j:
We need to look into how numby solves this, in the end this really is something we need.

adler-j:
Above comment was me thinking this was a issue ^^.

Anyway nice pull, but does removing `conftest.py` not screw up running `pytest`, or does it find it somehow?

setup.py
```diff
@@ -124,8 +136,11 @@ def run_tests(self):
 
     keywords='research development mathematics prototyping imaging tomography',
 
-    packages=find_packages(exclude=['*test*']),
+    packages=find_packages(),
     package_dir={'odl': 'odl'},
+    package_data={'odl': find_tests()},
+    include_package_data=True,
+    entry_points={'pytest11': ['odl_plugins = odl.util.pytest_plugins']},
```
adler-j: Does this make pytest find it? What does 11 mean?

adler-j:
How about the pytest.ini file?

doc/source/getting_started/installing.rst
```diff
@@ -467,5 +467,5 @@ If that does not help, `make an issue on GitHub <https://github.com/odlgroup/odl
 .. _CVXPY: http://www.cvxpy.org/en/latest/
 .. _odlcuda: https://github.com/odlgroup/odlcuda
 .. _CUDA toolkit: https://developer.nvidia.com/cuda-toolkit
-.. _ASTRA tomography toolbox: https://github.com/astra-toolbox/astra-toolbox
+.. _ASTRA: https://github.com/astra-toolbox/astra-toolbox
```
adler-j: This link has to be wrong now?

setup.py
```diff
@@ -61,6 +56,23 @@ def run_tests(self):
...
+test_path = os.path.join(root_path, 'odl/test')
+
+
+def find_tests():
+    tests = []
+    for path, _, filenames in os.walk(os.path.join(root_path, test_path)):
+        for filename in filenames:
+            basename, suffix = os.path.splitext(filename)
+            if (suffix == '.py' and
+                    (basename.startswith('test') or
```
adler-j: It is actually supposed to be `_test` and `test_`

See:

http://doc.pytest.org/en/latest/goodpractices.html

setup.py
```diff
@@ -61,6 +56,23 @@ def run_tests(self):
         errno = pytest.main(self.pytest_args)
         sys.exit(errno)
 
+
+test_path = os.path.join(root_path, 'odl/test')
```
adler-j: os.path.join(root_path, 'odl', test')

adler-j:
Looks good except for the minor comments. Are we sure this works? If we we need to make a 5.1 release, perhaps you can look into pypi test (https://wiki.python.org/moin/TestPyPI) before the actual release.

doc/source/getting_started/installing.rst
kohr-h: No, the other one is inline now, I don't see a point in 3 different targets linking to the same page.

kohr-h:
> How about the pytest.ini file?

It can be included as extra package file, however only under `odl`, not at top repo level (in this case one can specify the file in `MANIFEST.in`, but it won't be put into the final package since it's counted to metadata).
The price we pay is either a duplicate at top level or that we need to invoke `pytest odl` in the future, which is also okay IMO.

setup.py
kohr-h: Oh, of course.

setup.py
kohr-h: Yeah, I had that first, it was correct. Dunno why I changed it.

kohr-h:
Everything is fixed, and I tested all combinations of Python 2/3 and source/wheel/conda packages. Works smoothly.
You can try the packages yourself with
```
pip install -i https://testpypi.python.org/pypi odl
```
Good suggestion with the test server.

kohr-h:
Bumpety

adler-j:
Fine by me!

adler-j:
5.1 release comming?

kohr-h:
Done!

adler-j:
Trying install in a VM

adler-j:
Worked perfectly from `pip install` at least! Roughly 3300 tests, no failures etc. Good work!

kohr-h:
Phew! :-)

Issue #675, opened 2016-10-24
Error in show when using upper boundary for coords


kohr-h:
Minimal example:

``` python
>>> space = odl.uniform_discr([0, 0], [1, 1], [10, 10])
>>> x = space.one()
>>> x.show(coords=[1, None])
Traceback (most recent call last):

  File "<ipython-input-6-8b47643bf30d>", line 1, in <module>
    x.show(coords=[1, 1])

  File "/home/hkohr/git/odl/odl/discr/lp_discr.py", line 856, in show
    part = self.space.partition[indices].squeeze()

  File "/home/hkohr/git/odl/odl/discr/partition.py", line 451, in __getitem__
    int_to_slice=True)

  File "/home/hkohr/git/odl/odl/util/normalize.py", line 257, in normalized_index_expression
    ''.format(idx, i, n))

IndexError: Index 10 is out of bounds for axis 0 with size 10.
```

adler-j:
This could be fixed by using the `index_of` thing I implemented recently.

Issue #676, opened 2016-10-25
Indexing partitions by dimension


adler-j:
Indexing of partitions is currently done on a cell basis, but is often of interest to do indexing by dimension, i.e. `partition.bydimension[0]` would be the sub-partition along dimension 0.

kohr-h:
You can do that with lists: `partition[[0]]`.

adler-j:
Wait what? How does that even work?

kohr-h:
It's NumPy style indexing. Lists refer to axes, tuples to... something else.

adler-j:
I feel like you are missunderstanding the question then, there is no numpy equivalent of what I want to do here. I'll give an example.

Suppose I have the partition

```
>>> p = odl.uniform_partition([0, 0], [1, 2], [3, 4])
```

The partition can be seen as the product of sub-partitions

```
>>> p1 = odl.uniform_partition(0, 1, 3)
>>> p2 = odl.uniform_partition(0, 2, 4)
>>> p_alt = p1.append(p2)
```

I want to extract these sub-partitions, i.e.

```
>>> p.bydimension[0] == p1
True
>>> p.bydimension[1] == p2
True
```

your syntax does not do this:

```
>>> p[[0]]
uniform_partition([0.0, 0.0], [0.33333333333333331, 2.0], [1, 4])
>>> p[[0]] == p1
False
```

In fact,

```
>>> p[[0]] == p[0]
True
>>> p[[1]] == p[1]
True
```

kohr-h:
Right, I was too quick. Then we need an extra method.
The alternative is to not do the Numpy style indexing for single integers (or other "incomplete" slices) and let `p[0]` do what you want to do. But on the other hand, the list indexing thing could be very handy at some point.

PR #678, opened 2016-10-25
Issue 675  show upper bound

adler-j:
- Fixes issue #675, error when the upper bound is given to `show(coords=[...])`
- Add `RectPartition.index`
- Add `RectPartition.bydimension`

### DIFF
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -546,9 +546,9 @@ class RectPartition(object):
         newset = self.set[self.grid.inondeg]
         return RectPartition(newset, self.grid.squeeze())
 
-    def index_of(self, value, floating=False):
+    def index(self, value, floating=False):
         """Return the index of a value in the domain.
-        
+
         Parameters
         ----------
         value : ``self.set`` element
@@ -557,46 +557,46 @@ class RectPartition(object):
             True if the index should also give the position inside the voxel.
             This is given by returning the integer valued index of the voxel
             plus the partial distance from the left to the right boundary.
-            
+
         Returns
         -------
         index : float
             Index of the value, as counted from the left.
-            
+
         Examples
         --------
         Get the indices of start and end:
-        
+
         >>> p = uniform_partition(0, 2, 5)
-        >>> p.index_of(0)
+        >>> p.index(0)
         (0,)
-        >>> p.index_of(2)
+        >>> p.index(2)
         (4,)
-        
+
         Also supports points inside voxels, in that case returning the index
         of the voxel:
-        
-        >>> p.index_of(0.2)
+
+        >>> p.index(0.2)
         (0,)
-        
+
         By using the ``floating`` argument, partial positions inside the voxels
         can instead be determined:
-        
-        >>> p.index_of(0.2, floating=True)
+
+        >>> p.index(0.2, floating=True)
         (0.5,)
-            
-        These indexes work with indexing, extracting the voxel the point lies 
+
+        These indexes work with indexing, extracting the voxel the point lies
         in:
-        
-        >>> p[p.index_of(0.1)]
+
+        >>> p[p.index(0.1)]
         uniform_partition(0.0, 0.4, 1)
-        
+
         Index also works in higher dimensions:
-        
-        >>> p = uniform_partition([0, -1], [1, 2], (4, 1))      
-        >>> p.index_of([0.5, 2])
+
+        >>> p = uniform_partition([0, -1], [1, 2], (4, 1))
+        >>> p.index([0.5, 2])
         (2, 0)
-        >>> p[p.index_of([0.5, 2])]
+        >>> p[p.index([0.5, 2])]
         uniform_partition([0.5, -1.0], [0.75, 2.0], [1, 1])
         """
         value = np.atleast_1d(self.set.element(value))
@@ -617,10 +617,10 @@ class RectPartition(object):
                     result += (ind,)
                 else:
                     result += (ind - 1,)
-          
+
         return result
-                
-        
+
+
     def __str__(self):
         """Return ``str(self)``."""
         return 'partition of {} using {}'.format(self.set, self.grid)
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -561,17 +561,18 @@ class RectPartition(object):
         Parameters
         ----------
         value : ``self.set`` element
-            The point to find the index of.
+            Point whose index to find.
         floating : bool
-            True if the index should also give the position inside the voxel.
-            This is given by returning the integer valued index of the voxel
-            plus the partial distance from the left to the right boundary.
+            If True, then the index should also give the position inside the
+            voxel. This is given by returning the integer valued index of the
+            voxel plus the distance from the left cell boundary as a fraction
+            of the full cell size.
 
         Returns
         -------
-        index : int/float or tuple of int/float
+        index : int, float, tuple of int or tuple of float
             Index of the value, as counted from the left.
-            If ``self.ndim>1`` the result is a tuple, else a scalar.
+            If ``self.ndim > 1`` the result is a tuple, else a scalar.
             If ``floating=True`` the scalar is a float, else an int.
 
         Examples
@@ -584,8 +585,7 @@ class RectPartition(object):
         >>> p.index(2)
         4
 
-        Also supports points inside voxels, in that case returning the index
-        of the voxel:
+        For points inside voxels, the index of the containing cell is returned:
 
         >>> p.index(0.2)
         0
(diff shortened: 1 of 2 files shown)


odl/discr/partition.py
```diff
@@ -625,6 +625,60 @@ def index(self, value, floating=False):
...
+                Examples
+                --------
+                >>> p = uniform_partition(0, 1, 5)
+                >>> p.bydimension
+                uniform_partition(0, 1, 5).bydimension
+                """
+                return '{!r}.bydimension'.format(partition)
+
+        return RectPartitionByDimension()
+
```
adler-j: I disagree, see reply in #676

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+                else:
+                    result += (ind - 1,)
+
+        if self.ndim == 1:
+            result = result[0]
+
+        return result
+
+    @property
+    def bydimension(self):
```
kohr-h: I'd prefer the name `byaxis`, it's more in line with the terms we use in the rest of the package.

adler-j: I certainly agree with `byaxis`, I was hoping for something better.

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
         newset = self.set[self.grid.inondeg]
         return RectPartition(newset, self.grid.squeeze())
 
+    def index(self, value, floating=False):
```
kohr-h: Bloody useful function, I have to say

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+        """Return the index of a value in the domain.
+
+        Parameters
+        ----------
+        value : ``self.set`` element
+            The point to find the index of.
```
kohr-h: "Point whose index to find." would be my preference

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+        floating : bool
+            True if the index should also give the position inside the voxel.
```
kohr-h: Wrong logic, should be "If \`\`True\`\`, then ..." not "True if".

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+            This is given by returning the integer valued index of the voxel
+            plus the partial distance from the left to the right boundary.
```
kohr-h: Almost perfect, just at the end it's better to speak of "distance from the left cell boundary" only, since the distance from the left to the right boundary is the cell size by definition. Perhaps one could be even clearer by adding "in terms of multiples of the cell sizes" or something like that, to make clear that it's not an absolute quantity.

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+
+        Returns
+        -------
+        index : int/float or tuple of int/float
```
kohr-h: tuple of ints/floats

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+            Index of the value, as counted from the left.
+            If ``self.ndim>1`` the result is a tuple, else a scalar.
```
kohr-h: self.ndim > 1

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+                else:
+                    # interpolate between
+                    csize = float(cell_bdry_vec[ind] - cell_bdry_vec[ind - 1])
+                    result += (ind - (cell_bdry_vec[ind] - val) / csize,)
+            else:
+                if cell_bdry_vec[ind] == val and ind != len(cell_bdry_vec) - 1:
+                    # Value is on top of edge, but not last edge
+                    result += (ind,)
+                else:
+                    result += (ind - 1,)
```
kohr-h: This is stylistically not so great (I mean incrementing tuples). It's not a question of efficiency, obviously, but the mutable data structure for this purpose is a list, which should be `append`'ed to. At the very end, `return tuple(result)` then.

odl/discr/partition.py
kohr-h: You could also use a name like `as_outer` since you represent the partition as an outer product.

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+
+            See Also
+            --------
+            RectPartition.bydimension
+            """
+
+            def __getitem__(self, dim):
+                """Return ``self[dim]``."""
+                slc = np.zeros(partition.ndim, dtype=object)
+                slc[dim] = np.s_[:]
```
kohr-h: slice(None)

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+            """Helper class for accessing `RectPartition` by dimension.
+
+            See Also
+            --------
+            RectPartition.bydimension
+            """
+
+            def __getitem__(self, dim):
+                """Return ``self[dim]``."""
+                slc = np.zeros(partition.ndim, dtype=object)
```
kohr-h: Why this complicated? Is this any different from `[0] * partition.ndim`?

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+
+        return result
+
+    @property
+    def bydimension(self):
+        """Return the subpartition defined along one or several dimensions.
+
+        Examples
+        --------
+        Can get each dimension independently:
```
kohr-h: Start sentence differently

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+        Get the indices of start and end:
+
+        >>> p = uniform_partition(0, 2, 5)
+        >>> p.index(0)
+        0
+        >>> p.index(2)
+        4
+
+        Also supports points inside voxels, in that case returning the index
+        of the voxel:
```
kohr-h: For points inside voxels, the index of the containing cell is returned:

would be my favorite

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+        >>> p.index(0.2)
+        0
+
+        By using the ``floating`` argument, partial positions inside the voxels
+        can instead be determined:
+
+        >>> p.index(0.2, floating=True)
+        0.5
+
+        These indexes work with indexing, extracting the voxel the point lies
```
kohr-h: I think "indexes" is valid, too, but we used exclusively "indices" before.

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+        in:
```
kohr-h: the voxel in which the point lies

odl/discr/partition.py
```diff
@@ -546,6 +555,139 @@ def squeeze(self):
...
+
+        >>> p[p.index(0.1)]
+        uniform_partition(0.0, 0.4, 1)
+
+        Index also works in higher dimensions:
```
kohr-h: "The same principle" maybe, or explicitly \`\`index\`\`

odl/discr/partition.py
adler-j: This allows numpy indexing, i.e.

```
>>> slc = np.zeros(3, dtype=object)
>>> slc[[0, 2]] = 2
>>> slc
array([2, 0, 2], dtype=object)
>>> lst = [0] * 3
>>> lst[[0, 2]] = 2
TypeError: list indices must be integers, not list
```

Also, I find your code a-bit scary since it implicitly relies on the immutability of `0`. For immutable types, magic stuff like

```
>>> lst = [[0]] * 3
>>> lst[0][0] = 2
>>> lst
[[2], [2], [2]]
```

can occur.

kohr-h: > This allows numpy indexing, i.e.

Fair enough, but what about the `dtype=object` thing?



> Also, I find your code a-bit scary since it implicitly relies on the immutability of 0. For immutable types, magic stuff like
> 
> >>> lst = [[0]] * 3
> >>> lst[0][0] = 2
> >>> lst
> [[2], [2], [2]]
> 
> can occur.

That's not a problem of the 0, rather a problem of the inner list not being copied, i.e. all entries refer to the same value. If the inner thing is a Numpy array, it's exactly the same:
```python
>>> lst = [np.array([0])] * 3
>>> lst[0][0] = 2
>>> lst

[array([2]), array([2]), array([2])]
```
But no need to go further into this topic.

adler-j: > but what about the dtype=object thing?

Without it I cannot do

```
slc[dim] = np.s_[:]
```

later on. If the dtype was `int`, then assigning a `slice` would fail.

kohr-h: Ah of course, I didn't see it as a whole.

odl/discr/partition.py
adler-j: Sadly no, it can be:
`int`, `float`, tuple of int and tuple of float

adler-j: I'll go for

index : int, float, tuple of int or tuple of float

odl/discr/partition.py
adler-j: Also fits perfectly on the line

adler-j:
Fixed all comments, merge?

odl/discr/partition.py
kohr-h: That's what I meant, i.e. the suggestion applied to the second part only. Another possibility would be
```
index : int, float or tuple of these
```
but it's not quite as good.

PR #680, opened 2016-10-25
BUG: fix bugs caused by moving the conftest and pytest files

adler-j:
This has to be fixed rather soon.


conftest.py
```diff
@@ -52,7 +52,7 @@ def add_doctest_np_odl(doctest_namespace):
 
 
 this_dir = os.path.dirname(__file__)
-odl_root = os.path.abspath(os.path.join(this_dir, os.pardir, os.pardir))
+odl_root = os.path.abspath(this_dir)
```
kohr-h: This is not a solution, it will make it impossible to ship the tests.

Issue #693, opened 2016-10-31
Vector field space of complex space


chongchenmath:
I met with an interesting phenomenon:

``space = odl.uniform_discr(min_pt=[-16, -16], max_pt=[16, 16], shape=[256, 256], dtype='float32', interp='linear')``
``shifts = [not s % 2 for s in space.shape]``
``ft_op = odl.trafos.FourierTransform(space, halfcomplex=False, shift=shifts)``

The `dtype` is different below:

``ft_op.range.vector_field_space``
Output: ``ProductSpace(uniform_discr([-25.230915999143026, -25.230915999143026], [25.034566458293664, 25.034566458293664], [256, 256], dtype='float32'), 2)``

``odl.ProductSpace(ft_op.range, ft_op.range.ndim)``
Output: ``ProductSpace(uniform_discr([-25.230915999143026, -25.230915999143026], [25.034566458293664, 25.034566458293664], [256, 256], dtype='complex64'), 2)``

Why?

adler-j:
`ft_op.range.vector_field_space` is supposed to be the tangent bundle (ish) of the space `ft_op.range`, while `odl.ProductSpace(ft_op.range, ft_op.range.ndim)` would be the set of (`ft_op.range.ndim`) vectors on the range. This was a concious design decision at some point, see the discussion about it [here](https://github.com/odlgroup/odl/pull/488#discussion-diff-70392722).

chongchenmath:
The following is the definition of `vector_field_space`.

````
@property
    def vector_field_space(self):
        """Space of vector fields based on ``self``."""
        return ProductSpace(self.real_space, self.ndim)
````
It just supports `real_space`, not `complex_space`. I think that is the reason. Right?

adler-j:
It explicitly wants the real space, that was one of the motivating reasons behind introducing it.

Otherwise tests like [this test](https://github.com/odlgroup/odl/blob/master/odl/test/deform/linearized_deform_test.py#L35-L37) would fail with complex dtypes.

adler-j:
I guess the naming is the big issue at hand here. Perhaps it should be renamed to `tangent_boundle`. Thoughts @kohr-h, @ozanoktem?

ozanoktem:
[...]

PR #702, opened 2016-11-03
Issue 693  vector field space rename

adler-j:
Solves issue #693 and two other bugs found when running `pytest --examples`

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -237,8 +237,10 @@ class DiscreteLp(DiscretizedSpace):
     def tangent_bundle(self):
         """The tangent bundle associated with `domain` using `partition`.
 
-        The tangent bundle of a n-dimensional space can be seen as the set of
-        n-vectors on the same domain.
+        The tangent bundle of a space ``X`` of functions ``R^d --> F`` can be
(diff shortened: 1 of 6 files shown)


odl/discr/lp_discr.py
```diff
@@ -234,9 +234,13 @@ def exponent(self):
         return self.__exponent
 
     @property
-    def vector_field_space(self):
-        """Space of vector fields based on ``self``."""
-        return ProductSpace(self.real_space, self.ndim)
+    def tangent_bundle(self):
+        """The tangent bundle associated with `domain` using `partition`.
+
+        The tangent bundle of a n-dimensional space can be seen as the set of
```
kohr-h: an n-dimensional

odl/discr/lp_discr.py
```diff
@@ -234,9 +234,13 @@ def exponent(self):
...
+        n-vectors on the same domain.
```
kohr-h: I'd write it like this: 
```
The tangent bundle of a space `X` of functions `R^d --> F` can be
interpreted as the space of vector-valued functions `R^d --> F^d`.
This space can be identified with the power space `X^d` as used
in this implementation.
```

adler-j:
Merge after CI.

PR #703, opened 2016-11-03
ENH: Add fbp_op utility with tests

adler-j:
Tests not yet done using ASTRA (Since I lack it at this machine) but the overall structure should stand.

### DIFF
--- a/examples/tomo/filtered_backprojection_cone_3d.py
+++ b/examples/tomo/filtered_backprojection_cone_3d.py
@@ -16,11 +16,11 @@
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Example using a filtered backprojection in 2d using the ray transform and a
-ramp filter. The ramp filter is implemented in fourier space.
+Example using a filtered backprojection (FBP) in cone-beam 3d using `fbp_op`.
 
-See https://en.wikipedia.org/wiki/Radon_transform#Inversion_formulas for
-more information.
+Note that the FBP is only approximate in 3d, but still gives a decent
+reconstruction that can be used as an initial guess in more complicated
+methods.
 """
 
 import numpy as np
@@ -30,19 +30,20 @@ import odl
 # --- Set-up geometry of the problem --- #
 
 
-# Discrete reconstruction space: discretized functions on the rectangle
-# [-20, 20]^2 with 300 samples per dimension.
+# Discrete reconstruction space: discretized functions on the cube
+# [-20, 20]^3 with 300 samples per dimension.
 reco_space = odl.uniform_discr(
-    min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='float32')
+    min_pt=[-20, -20, -20], max_pt=[20, 20, 20], shape=[300, 300, 300],
+    dtype='float32')
 
-# Angles: uniformly spaced, n = 1000, min = 0, max = pi
-angle_partition = odl.uniform_partition(0, np.pi, 1000)
-
-# Detector: uniformly sampled, n = 500, min = -30, max = 30
-detector_partition = odl.uniform_partition(-30, 30, 500)
-
-# Make a parallel beam geometry with flat detector
-geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+# Make a circular cone beam geometry with flat detector
+# Angles: uniformly spaced, n = 360, min = 0, max = 2 * pi
+angle_partition = odl.uniform_partition(0, 2 * np.pi, 360)
+# Detector: uniformly sampled, n = (558, 558), min = (-30, -30), max = (30, 30)
+detector_partition = odl.uniform_partition([-30, -30], [30, 30], [558, 558])
+geometry = odl.tomo.CircularConeFlatGeometry(
+    angle_partition, detector_partition, src_radius=500, det_radius=100,
+    axis=[1, 1, 1])
 
 
 # --- Create FilteredBackProjection (FBP) operator --- #
@@ -51,19 +52,8 @@ geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
 # Ray transform (= forward projection). We use the ASTRA CUDA backend.
 ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
 
-# Fourier transform in detector direction
-fourier = odl.trafos.FourierTransform(ray_trafo.range, axes=[1])
-
-# Create ramp in the detector direction
-ramp_function = fourier.range.element(lambda x: np.abs(x[1]) / (2 * np.pi))
-
-# Create ramp filter via the convolution formula with fourier transforms
-ramp_filter = fourier.inverse * ramp_function * fourier
-
-# Create filtered backprojection by composing the backprojection (adjoint) with
-# the ramp filter.
-fbp = ray_trafo.adjoint * ramp_filter
-
+# Create FBP operator using utility function
+fbp = odl.tomo.fbp_op(ray_trafo)
 
 # --- Show some examples --- #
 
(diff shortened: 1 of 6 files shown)


odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -0,0 +1,80 @@
...
+__all__ = ('fbp_op',)
+
+
+def fbp_op(ray_trafo):
+    """Create Filtered BackProjection from a ray transform.
+
+    Parameters
+    ----------
+    ray_trafo : `RayTransform`
+
```
adler-j: Perhaps add some options for users to specify one of the commonly used filters here.

adler-j:
Working implementation done, works for a range of test cases, as shown in the test.

Needs review

adler-j:
Now happy with the implementation, should be ready to go in (barring any misstakes etc).

odl/test/largescale/tomo/analytic_slow_test.py
```diff
@@ -0,0 +1,188 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test reconstruction with ASTRA."""
```
adler-j: doh

odl/test/largescale/tomo/analytic_slow_test.py
```diff
@@ -0,0 +1,188 @@
...
+        # Ray transform
+        return tomo.RayTransform(discr_reco_space, geom,
+                                 impl=impl)
+    else:
+        raise ValueError('param not valid')
+
+
+@skip_if_no_largescale
+def test_fbp_reconstruction(projector):
+    """Test discrete Ray transform using ASTRA for reconstruction."""
```
adler-j: wrong

examples/tomo/filtered_backprojection_cone_3d.py
```diff
@@ -0,0 +1,74 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Example using a filtered backprojection (FBP) in cone-beam 3d using `fbp_op`.
+
+Note that the FBP is only approximate in 3d, but still gives a decent
```
adler-j: Well if the kernel is uniform, i.e. a filter, then we get the wrong answer, no? We would need to have some kind of weighting which would make the algorithm e.g. FDK or something else.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -0,0 +1,135 @@
...
+    ray_trafo : `RayTransform`
+        The ray transform (forward operator) whose approximate inverse should
+        be computed. Its geometry has to be any of the following
+        Parallel2DGeometry : Exact reconstruction
+        Parallel3dAxisGeometry : Exact reconstruction
+        FanFlatGeometry : Approximate reconstruction, correct in limit of fan
+                          angle = 0.
+        CircularConeFlatGeometry : Approximate reconstruction, correct in limit
+                                   of fan angle = 0.
+        HelicalConeFlatGeometry : Very approximate.
```
adler-j: Certainly should have backticks.

Regarding the formulation I wanted to convey that this is actually quite bad (as in a factor of 2 of at times). Not sure how to do it better.

examples/tomo/filtered_backprojection_cone_3d.py
kohr-h: Okay, if the kernel is by definition independent of the reco point, then yes. Wasn't too serious a remark anyway. The second part was serious, though.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -0,0 +1,135 @@
...
+    if ray_trafo.domain.ndim == 2:
+        # Define ramp filter
+        def fft_filter(x):
+            return np.sqrt(x[1]**2) / (2 * alen)
+
+        # Define (padded) fourier transform
+        if padding:
+            # Define padding operator
+            ran_shp = (ray_trafo.range.shape[0],
+                       ray_trafo.range.shape[1] * 2 - 1)
```
adler-j: This is what is theoretically needed, but as you say it is hardly the case in reality. I think I'll let this stay for now though since it should be doable for most cases, worst case users can simply pick `padding=False`. If this becomes much used I'll optimize it.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -0,0 +1,135 @@
...
+            fourier = FourierTransform(resizing.range, axes=1, impl=impl)
+            fourier = fourier * resizing
+        else:
+            fourier = FourierTransform(ray_trafo.range, axes=1, impl=impl)
+
+    elif ray_trafo.domain.ndim == 3:
+        # Find the direction that the filter should be taken in
+        du = ray_trafo.geometry.det_init_axes[0]
+        dv = ray_trafo.geometry.det_init_axes[1]
+        axis = ray_trafo.geometry.axis
```
adler-j: For the geometries this supports, yes. Perhaps should have a `isinstance` but this is basically duck-typed to some axis oriented geometry and the error would be quite clear "no `geometry.axis` exists".

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -0,0 +1,135 @@
...
+            return np.abs(c[0] * x[1] + c[1] * x[2]) / (2 * alen)
+
+        # Define (padded) fourier transform
+        if padding:
+            # Define padding operator
+            ran_shp = (ray_trafo.range.shape[0],
+                       ray_trafo.range.shape[1] * 2 - 1 if used_axes[0]
+                       else ray_trafo.range.shape[1],
+                       ray_trafo.range.shape[2] * 2 - 1 if used_axes[1]
+                       else ray_trafo.range.shape[2])
```
adler-j: Will do.

odl/tomo/analytic/filtered_back_projection.py
kohr-h: No need for instance checks, but you could emphasize that the named geometries are the ones that are supported.

adler-j:
> I guess you did some speed-testing also? Perhaps the padding cost could be lowered a bit by padding less.

Well the padding causes this to take about 4 times longer, so perhaps 10-20 seconds for a reconstruction of realistic CBCT data. I'd say that is acceptable since this is not used iteratively. If we want to optimize it at some point we can put in the effort. I'd rather put that effort into getting FDK/Katsevich method in if I have the time.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -0,0 +1,135 @@
...
+        dv = ray_trafo.geometry.det_init_axes[1]
+        axis = ray_trafo.geometry.axis
+        det_normal = np.cross(du, dv)
+        rot_dir = np.cross(axis, det_normal)
+        c = np.array([np.vdot(rot_dir, du), np.vdot(rot_dir, dv)])
+        cnorm = np.linalg.norm(c)
+        assert cnorm != 0
+        c /= cnorm
+
+        used_axes = c != 0
```
kohr-h: Oh, I was rather thinking of "snapping" cases close to 0 to the axis-aligned case. It will rather run faster then.

adler-j: Sure, but we risk getting numerical errors. I'll leave this until someone comes up with a case where this is an issue. Less code to test etc.

odl/tomo/analytic/filtered_back_projection.py
adler-j: Will add that.

adler-j:
Merge after CI.

kohr-h:
> dismissed kohr-h’s review 14 hours ago

Scandal!

Issue #667, opened 2016-10-21
Write documentation on testing


adler-j:
We currently lack documentation on how testing of the package works, now that we'll probably get some new developers we should write doc on this.

PR #704, opened 2016-11-03
DOC: Initial work in testing documentation, see #667

adler-j:
Adds documentation on how tests work in ODL along with implementing new ones.

### DIFF
--- a/doc/source/dev/dev.rst
+++ b/doc/source/dev/dev.rst
@@ -15,9 +15,11 @@ take a look at our `issue tracker <https://github.com/odlgroup/odl/issues>`_, an
 existing issue dealing with your problem, create a new one. Don't be shy -- use the issue tracker to
 ask questions, too.
 
-If you are a developer and want to contribute code, we recommend that you read through the
+If you are a developer and want to contribute code, you may want to read through the
 subsequent instructions. If you are experienced with Git, you may want to skip directly to
 :ref:`the development workflow section <development-workflow>`.
+
+In order to properly follow the ODL style, we recommend that you read the `How to document`_ and `Testing in ODL`_ sections.
 
 Contents
 --------
@@ -27,4 +29,5 @@ Contents
 
    extend
    document
+   testing
    gitwash/index
--- a/doc/source/dev/dev.rst
+++ b/doc/source/dev/dev.rst
@@ -19,7 +19,7 @@ If you are a developer and want to contribute code, you may want to read through
 subsequent instructions. If you are experienced with Git, you may want to skip directly to
 :ref:`the development workflow section <development-workflow>`.
 
-In order to properly follow the ODL style, we recommend that you read the `How to document`_ and `Testing in ODL`_ sections.
+In order to properly follow the ODL style, we recommend that you read the `How to document`_:ref:`dev_document` and :ref:`dev_testing` sections.
 
 Contents
 --------
--- a/doc/source/dev/testing.rst
+++ b/doc/source/dev/testing.rst
@@ -1,0 +1,71 @@
+##############
+Testing in ODL
+##############
+
+ODL is tested using pytest_ and has four main types of tests and verification measures that are supposed to test the code in different ways. These are listed below along with the command to run them.
+
+============  =====================  =======
+Name          Command                Description
+============  =====================  =======
+Unittests     `pytest`               Test "micro-features" of the code, like testing that each parameter combination works, that the correct exceptions are raced and that the code works correctly in corner cases and with a wide range of input. 
+Largescale    `pytest --largescale`  Verify that the functionality works well in more complicated and realistic conditions.
+Doctests      `pytest`               Meant as simple examples of how to use a functionality for users reading the documentation. 
+Examples      `pytest --examples`    Copy-paste friendly examples on how to use a function in a more complete context.
+============  =====================  =======
+
+Unittests
+~~~~~~~~~
+All unit-tests in ODL are contained in the `odl/tests` folder, where each ODL sub-package has a test file of its own. Any major ODL functionality should have unittests covering all of the use cases that are implied in the documentation. In addition to this, the tests should be quick to run, preferably at most a few milliseconds per tests. This is because the tests should be run as often as possible in order to verify that no functionality has been broken.
+
+A short example of testing a function is given below, for more information consult the pytest documentation and look at existing tests in the test folder.
+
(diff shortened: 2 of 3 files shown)


adler-j:
This has actually come quite far and could use a review on the content level, lets get it out of the way.

adler-j:
Bumping this.

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+
+.. code:: python
+
+    import numpy as np
+    import pytest
+
+
+    def myfunction(x):
+        """Convert ``x`` to a integer and add 1"""
+        return int(x) + 1
```
kohr-h: This is nice and simple, but IMO it tests the wrong thing and gives a slightly wrong impression of what testing is supposed to achieve. The problem is that the tests are entirely testing the builtin `int` constructor, which is not exactly the purpose of testing. Testing against bad input makes more sense if our own implementation contains checks and we want to make sure they are effective. Of course it's also valid to check that the functionality we rely upon behaves in the way we expect, but it would be nice to have at least one self-implemented aspect that is checked.

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+            myfunction('non-integer')
+
+        with pytest.raises(TypeError):
+            myfunction(object())
+
+        with pytest.raises(OverflowError):
+            myfunction(np.inf)
+
+
+Largescale
```
kohr-h: Large-scale

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+~~~~~~~~~~
+Largescale test that functions work well even in realistic conditions and with a wide range of input, they live in `odl/tests/largescale`. Not all functionality needs largescale tests, in fact, most doesn't. However for some functions accurate tests even in realistic conditions are needed and these cannot be run very quickly.
```
kohr-h: Large-scale tests verify that ...
"wide range of input" was already mentioned in the unit test row in the table. Perhaps "even wider range of input than in the standard unit tests"
New sentence starting after "input"
`odl/tests/largescale` link probably dead
Second sentence leaves open what does need large-scale tests. Suggestion:
"This type of test makes most sense for (1) functionality that has a complex implementation where it's easy to make mistakes that the code slow (regression tests) and (2) features that take too much time to be tested broadly in the standard suite.
For the second type, the unit tests should include only a couple of tests that can run fast, and the full range of inputs can be tested in the large-scale suite."
Basically you write the second part below.

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+
+It may also be the case that some functions accept a very large number of possible input configurations, in this case, testing the most common configuration in the regular unittest and testing the others in a largescale test is acceptable.
+
+Doctests
+~~~~~~~~
+Doctests are the most simple type of test used in ODL, and are mainly meant as examples on how to use functionality. They can be included by using the Examples header in an usual docstring, as shown below:
```
kohr-h: most simple -> simplest?
"mainly meant as examples on how to use functionality" -> "snippets of code that document the usage of functions and classes and can be run as small tests at the same time"

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+        2
+
+        The function also works with floats:
+
+        >>> myfunction(1.3)
+        2
+        """
+        return int(x) + 1
+
+Dispite simply looking like documentation, doctests are actual pieces of python code and will be executed when the ``pytest`` command is invoked. All ODL source files also contain the lines
```
kohr-h: Despite
You could add a reference to the `doctest` module here.

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+
+    if __name__ == '__main__':
+        from odl.util.testutils import run_doctests
+        run_doctests()
+
+which mean that if a ODL source file is executed in isolation, all the doctests in the file are run. This can be useful during development in order to quickly see if some functionality works as expected.
+
+Examples
+~~~~~~~~
+Examples, while not technically tests in the traditional sense, still constitute part of the test framework for ODL by showing how different parts of ODL work togeather and by ensuring that functions that depend on each other work as expected.
```
kohr-h: constitute a part
together
You could mention here that the examples reflect more the user's perspective, and that keeping them in a working state ensures that ODL is usable from a user's point of view.

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+
+It is even possible to run all examples as part of the test suite by running ``pytest --examples``, but be aware that this requires all ODL dependencies to be installed.
```
kohr-h: and that plotting windows can be opened during execution

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,106 @@
...
+
+For examples on how to write examples, please consult the examples directory.
```
kohr-h: This is like a triple tautology :-)
"Consult the `exampls`(a link) directory for an impression of the style in which ODL examples are written."

doc/source/dev/testing.rst
adler-j: While I agree with you, it is really hard to make something that fits both. In this case, the "own implemented" thing would be the + 1

adler-j:
Updated by comments, I guess this needs a final round and then we get it in.

doc/source/dev/testing.rst
kohr-h: I tried to think of something but ended up doing the same, just more complicated.. Perhaps just leave it as it is and add a short comment like "Note that this is a toy example that basically only tests the builtin `int`, but you get the idea."

doc/source/dev/dev.rst
```diff
@@ -19,7 +19,7 @@ If you are a developer and want to contribute code, you may want to read through
 subsequent instructions. If you are experienced with Git, you may want to skip directly to
 :ref:`the development workflow section <development-workflow>`.
 
-In order to properly follow the ODL style, we recommend that you read the `How to document`_ and `Testing in ODL`_ sections.
+In order to properly follow the ODL style, we recommend that you read the `How to document`_:ref:`dev_document` and :ref:`dev_testing` sections.
```
kohr-h: Double link

PR #708, opened 2016-11-08
Issue 639  callback show and save

### LINKED ISSUE #639: `callback` for saving intermediate iterates?
We have `callback`s for showing intermediate iterations, but is there an easy way to save these plots to disc?

There is a `callback` for applying functions to intermediate iterates, and I assume that one could write some kind of `lambda`-function for it there. Do we want a "standard implementation" of that in odl?

aringh:
Callback for showing and saving the intermediate iterates. Closes #639

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -374,6 +374,9 @@ class CallbackShow(SolverCallback):
         ----------
         display_step : positive int, optional
             Number of iterations between plots. Default: 1
+        saveto : string, optional
+            Path to where the figures are to be saved. If not given, the
+            figures are not saved. Default: None
 
         Other Parameters
         ----------------
@@ -384,6 +387,7 @@ class CallbackShow(SolverCallback):
         self.kwargs = kwargs
         self.fig = kwargs.pop('fig', None)
         self.display_step = kwargs.pop('display_step', 1)
+        self.saveto = kwargs.pop('saveto', None)
         self.iter = 0
         self.space_of_last_x = None
 
@@ -395,8 +399,15 @@ class CallbackShow(SolverCallback):
         self.space_of_last_x = x_space
 
         if (self.iter % self.display_step) == 0:
-            self.fig = x.show(*self.args, fig=self.fig,
-                              update_in_place=update_in_place, **self.kwargs)
+            if self.saveto is None:
+                self.fig = x.show(*self.args, fig=self.fig,
+                                  update_in_place=update_in_place,
+                                  **self.kwargs)
+            else:
+                self.fig = x.show(*self.args, fig=self.fig,
+                                  saveto=(self.saveto + '_' + str(self.iter)),
+                                  update_in_place=update_in_place,
+                                  **self.kwargs)
 
         self.iter += 1
 
@@ -408,55 +419,14 @@ class CallbackShow(SolverCallback):
     def __repr__(self):
         """Return ``repr(self)``."""
 
-        return '{}(display_step={}, fig={}, *{!r}, **{!r})'.format(
+        return '{}(display_step={}, saveto={}, fig={}, *{!r}, **{!r})'.format(
             self.__class__.__name__,
+            self.saveto,
             self.display_step,
             self.fig,
             self.args,
             self.kwargs)
 
-
-class CallbackShowAndSave(SolverCallback):
-
-    """Show and save the iterates.
-
-    See Also
-    --------
-    odl.discr.lp_discr.DiscreteLpElement.show
-    odl.space.base_ntuples.NtuplesBaseVector.show
-    """
-
-    def __init__(self, file_prefix, *args, **kwargs):
-        """Initialize a new instance.
-
-        Additional parameters are passed through to the ``show`` method.
-
-        Parameters
-        ----------
-        file_prefix : string
-            Path to where the figure is to be saved
-        display_step : positive int, optional
-            Number of iterations between plots/saves. Default: 1
-
-        Other Parameters
-        ----------------
-        kwargs :
-            Optional arguments passed on to ``x.show``
-        """
-        self.file_prefix = file_prefix
-        self.args = args
-        self.kwargs = kwargs
-        self.fig = kwargs.pop('fig', None)
-        self.display_step = kwargs.pop('display_step', 1)
-        self.iter = 0
-
-    def __call__(self, x):
-        """Show and save the current iterate."""
-        if (self.iter % self.display_step) == 0:
-            self.fig = x.show(*self.args, fig=self.fig, **self.kwargs,
-                              saveto=(self.file_prefix + '_' + str(self.iter)))
-
-        self.iter += 1
 
 if __name__ == '__main__':
     # pylint: disable=wrong-import-position
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -416,6 +416,48 @@ class CallbackShow(SolverCallback):
             self.kwargs)
 
 
+class CallbackShowAndSave(SolverCallback):
+
+    """Show and save the iterates.
+
+    See Also
+    --------
+    odl.discr.lp_discr.DiscreteLpElement.show
+    odl.space.base_ntuples.NtuplesBaseVector.show
+    """
+
+    def __init__(self, file_prefix, *args, **kwargs):
+        """Initialize a new instance.
+
+        Additional parameters are passed through to the ``show`` method.
+
+        Parameters
+        ----------
+        file_prefix : string
+            Path to where the figure is to be saved
+        display_step : positive int, optional
+            Number of iterations between plots/saves. Default: 1
+
+        Other Parameters
+        ----------------
+        kwargs :
+            Optional arguments passed on to ``x.show``
+        """
+        self.file_prefix = file_prefix
+        self.args = args
+        self.kwargs = kwargs
+        self.fig = kwargs.pop('fig', None)
+        self.display_step = kwargs.pop('display_step', 1)
+        self.iter = 0
+
+    def __call__(self, x):
+        """Show and save the current iterate."""
+        if (self.iter % self.display_step) == 0:
+            self.fig = x.show(*self.args, fig=self.fig, **self.kwargs,
+                              saveto=(self.file_prefix + '_' + str(self.iter)))
+
+        self.iter += 1
+
 if __name__ == '__main__':
     # pylint: disable=wrong-import-position
     from odl.util.testutils import run_doctests
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -23,7 +23,6 @@ from future import standard_library
 standard_library.install_aliases()
 
 import time
-import os
 
 
 __all__ = ('CallbackStore', 'CallbackApply',
@@ -387,11 +386,6 @@ class CallbackShow(SolverCallback):
         ----------------
         kwargs :
             Optional arguments passed on to ``x.show``
-
-        Raises
-        ------
-        ValueError
-           If ``saveto`` is not a valid path
 
         Examples
         --------
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -23,6 +23,7 @@ from future import standard_library
 standard_library.install_aliases()
 
 import time
+import os
 
 
 __all__ = ('CallbackStore', 'CallbackApply',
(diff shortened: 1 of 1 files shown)


adler-j:
Disagree on solution, I'd simply add a parameter to the existing `CallbackShow`. I.e. 

```
file_prefix : string, optional
    Path to where the figure is to be saved. If `None`, no file is saved.
```

and then add an if statement in the call:

```python
def __call__(self, x):
    """Show and save the current iterate."""
    if self.file_prefix is not None:
        saveto = self.file_prefix + '_' + str(self.iter))
    else:
        saveto = None

    if (self.iter % self.display_step) == 0:
        self.fig = x.show(*self.args, fig=self.fig, **self.kwargs,
                          saveto=saveto)
    self.iter += 1
```

i.e. copy paste this code in a nice manner into the old callback.

Reason for this is to avoid redundancy and also make it easer to change these things on the fly.

adler-j:
Also needs to get the failing tests fixed

adler-j:
After more thought, perhaps simply over-ride the `saveto` keyword in the callback. I.e. change `file_prefix` to `saveto` above.

aringh:
And just have a flag with `True`/`False` if each shown iterate should also be saved?

adler-j:
No, you can use the existence of the path as a flag. If a path is given, then we should save, else not.

aringh:
Updated according to comments. Tests should also work now :-)

odl/solvers/util/callback.py
```diff
@@ -325,12 +328,18 @@ def __init__(self, *args, **kwargs):
         self.kwargs = kwargs
         self.fig = kwargs.pop('fig', None)
         self.display_step = kwargs.pop('display_step', 1)
+        self.saveto = saveto
```
adler-j: This should perhaps be pushed into kwargs similar to display_step

adler-j:
Basically only one comment, else looking great!

odl/solvers/util/callback.py
```diff
@@ -315,6 +315,9 @@ def __init__(self, *args, **kwargs):
         ----------
         display_step : positive int, optional
             Number of iterations between plots. Default: 1
+        saveto : string, optional
```
adler-j: move into kwargs

adler-j: Also should be a path, not a string. See https://docs.python.org/2/library/os.path.html

odl/solvers/util/callback.py
```diff
@@ -315,6 +315,9 @@ def __init__(self, *args, **kwargs):
...
+            Path to where the figures are to be saved. If not given, the
```
adler-j: Clarify what happens if the path does not exist

adler-j: Also be clear that this is the path to a directory, not the path to a file.

aringh:
Fixed. I also cleaned the commit history a bit.

I will give @kohr-h a chance to have a look if he wants to, and merge later today or tomorrow.

odl/solvers/util/callback.py
```diff
@@ -325,12 +328,18 @@ def __init__(self, *args, **kwargs):
...
 
     def __call__(self, x):
         """Show the current iterate."""
         if (self.iter % self.display_step) == 0:
-            self.fig = x.show(*self.args, fig=self.fig, **self.kwargs)
+            if self.saveto is None:
+                self.fig = x.show(*self.args, fig=self.fig, **self.kwargs)
+            else:
+                self.fig = x.show(*self.args, fig=self.fig,
+                                  saveto=(self.saveto + '_' + str(self.iter)),
```
kohr-h: Two comments:

- Add a format suffix. Right now it's storing in ??? format, which is whatever matplotlib defaults to. Perhaps use another kwarg to specify the format (suffix), defaulting to `'png'`.
- If possible, use an enumeration scheme that preserves order on disk. If you're unlucky, you get `iter_10` before `iter_2`. I'm not sure how to do it best, but at least test if this enumeration gives the right ordering.

adler-j: On most OSs, `iter_10` comes before `iter_2`, frankly I dont know any where it doesn't. I also dont know a proper solution to this except the dicom approach:

`iter_0000010`
`iter_0000002`

kohr-h: Suggestion: Offer the user an "advanced" variant where it is possible to specify a format string as `saveto` argument (something like `reco_iter_{:>03}` to enforce at least 3 digits), where we simply fill in the iteration number via `self.saveto.format(self.iter)`. That would require a way to determine if a format is present, which should be simple -- just check that there are 0 to 1 pairs `{` and `}`. For 0, replace the `saveto` by `saveto + '_{}' and then go ahead as in the "1" case.

adler-j: Excellent idea! Go with that one.

odl/solvers/util/callback.py
aringh: It is moved into kwargs, but the arguments listed here are the arguments which are not just directly piped into kwargs of show, which in turn I think pipes most of it into matplotlib.

aringh:
I have (finally) tried to update according to comments. Two comments:
1) I'm not sure what the difference between a path an a string is. Hence I'm not sure how to update according to one of @adler-j 's comments.
2) Regarding the advanced saving in order to be able to keep ordering of the saved files: I'm not sure exactly how that works. I have updated according to the comment of @kohr-h but since I'm not sure of what it actually does I do not know how to document this feature.

odl/solvers/util/callback.py
```diff
@@ -315,6 +316,13 @@ def __init__(self, *args, **kwargs):
         ----------
         display_step : positive int, optional
             Number of iterations between plots. Default: 1
+        saveto : string, optional
+            Path to a directory where the figures are to be saved, potentially
+            followed by a file name. If the directory name does not exist, a
+            ``ValueError`` is raised. If not given, the figures are not saved.
```
adler-j: Be explicit in that "not given" means "If `None`."

odl/solvers/util/callback.py
```diff
@@ -315,6 +316,13 @@ def __init__(self, *args, **kwargs):
...
+            Default: None
+        suffix : string, optinal
```
adler-j: Why is suffix a separate argument? Wouldnt getting it from the saveto string be enough?

odl/solvers/util/callback.py
```diff
@@ -325,12 +333,38 @@ def __init__(self, *args, **kwargs):
...
         self.fig = kwargs.pop('fig', None)
         self.display_step = kwargs.pop('display_step', 1)
+        self.saveto = kwargs.pop('saveto', None)
+        self.suffix = kwargs.pop('suffix', 'png')
         self.iter = 0
 
+        if self.saveto is not None:
+            head, tail = os.path.split(self.saveto)
+            if not os.path.exists(head) or head == '':
+                raise ValueError('head of the `saveto` {} path not found'
```
adler-j: I have never heard the word head in this sense before, what do we mean more explicitly?

odl/solvers/util/callback.py
```diff
@@ -325,12 +333,38 @@ def __init__(self, *args, **kwargs):
...
+                                 ''.format(head))
+            if ((not tail.count('{') == tail.count('}'))):
+                raise ValueError('tail of `saveto` {} does not contain the '
+                                 ' same number of ``{`` as ``}``'.format(tail))
+            elif tail.count('{') == 1:
```
adler-j: else:
    self.adv_saveto = tail.count('{') == 1

adler-j: Frankly second thought, dont even make an if/else here. Just make this assignment first and then validate the input on the next line.

adler-j:
Did a short review, answers to questions:

> I'm not sure what the difference between a path an a string is. Hence I'm not sure how to update according to one of @adler-j 's comments.

Well a string is any string "banana", "I like ODL", "" etc. A path is a location on your harddrive (as represented by a string). Perhaps it is obvious from the doc now though.

> Regarding the advanced saving in order to be able to keep ordering of the saved files: I'm not sure exactly how that works. I have updated according to the comment of @kohr-h but since I'm not sure of what it actually does I do not know how to document this feature.

This is enough, users can now give a "smart" format string and that should do this.

adler-j:
Anyway, the advanced formatting option is not documented, and it would be great to have a doctext on initializing one of these (without writing to any file) to demonstrate that how it works.

odl/solvers/util/callback.py
kohr-h: The suffix thing was from my initial proposal but then superseded by the format-type string suggestion. That wasn't clear.

odl/solvers/util/callback.py
kohr-h: I'm against just giving a directory because the user then has to guess as to how the system is gonna name the files. "Explicit is better than implicit."
I would prefer a ``saveto_fmt`` parameter in which users can specify *both location and file naming pattern*. The canonical example would be ``'/path/to/image/files/image-{:03}.png'`` for less than 1000 files. Link also to [this page]{https://docs.python.org/3/library/string.html#format-examples}, it gives some instructive examples regarding usage of format strings.

Finally we shouldn't raise a `ValueError` for an I/O error. That should be handled by the system. Just let the underlying error propagate.

odl/solvers/util/callback.py
kohr-h: IMO don't check at all. Just use the format as-is, the user's gonna get what she deserves. :-)

adler-j: Frankly I agree, no need to be too smart here.

adler-j: Also the user may actually want these:

```
>>> 'The result was {{ {}'.format('correct')
'The result was { correct'
```

odl/solvers/util/callback.py
aringh: Getting the suffix from the `saveto` would mean that we have to strip the `saveto` of a potential `.myFormat`, add the iteration counts and then put it back. That seems cumbersome to me...

Regarding the format-type string: since I was not sure of how/why that works, I was not sure of how to implement this :p I still don't know how to implement that in an efficient way, if it is to include a file format too...?

aringh:
The advanced formatting is not documented, since I'm not sure of what exactly it does and therefore even less sure of how to document it ;-). I can take a look at it later, to try to understand what it actually does.

kohr-h:
So the way this works is just via standard python string formatting. The curly brackets are replaced by arguments to the `format` string method, by default in order of appearance, but that can be changed, too.

So my preferred version of this feature would be that we use the string we get in the form `string.format(iteration_number)`, which should then also be mentioned in the documentation like this. Now the user can specify what `string` is. For example, if `string = 'image_{}.png'` then formatting will just substitute the number in some default format for `{}`, resulting in `image_0.png`, `image_10.png` and similar.

But this gives the user all freedom to specify paths and suffixes, for example `my_subdir/proj_{}.jpg`, and formats like `proj_{:03}` for advanced number formatting. I don't think we need extensive documentation on this, people can read up themselves. We should just give an example that combines all these aspects so users can make a good guess from there to their cases. Here I would literally write something `/path/to/my/images/image_{:03}.png` and explain in one sentence what it does.

adler-j:
Bumping this, very close to competition so lets not forget it!

aringh:
Rebased and fixed comment. Ready to go if all checks pass?

odl/solvers/util/callback.py
```diff
@@ -23,6 +23,7 @@
 standard_library.install_aliases()
 
 import time
+import os
```
adler-j: Unused import?

odl/solvers/util/callback.py
```diff
@@ -374,16 +375,47 @@ def __init__(self, *args, **kwargs):
...
+            iteration number. If the directory name does not exist, a
+            ``ValueError`` is raised. If ``None``, the figures are not saved.
+            Default: ``None``
 
         Other Parameters
         ----------------
         kwargs :
             Optional arguments passed on to ``x.show``
+
+        Raises
```
adler-j: This does not actually raise, right? In that case we can actually drop this section.

odl/solvers/util/callback.py
aringh: True :-) The use of it was removed when the implementation was simplified

odl/solvers/util/callback.py
aringh: This class does not raise anything, that is true (but the `show`-method will).

aringh:
Closes #639

Issue #666, opened 2016-10-20
Improving the readme


adler-j:
"now" that we have a very nice installation documentation, I suggest we simplify the README.

Removing the "Optional Dependencies" and "Compatibility" headings, moving perhaps the second one to the installation instructions. We could leave the "ODL is compatible to Python 2 and 3 through the future library. It is intended to work on all major platforms (GNU/Linux / Mac / Windows)." comment under installation, if we want.

Adding a "Code example" heading that does something truly amazing and mind blowing that shows the strength of ODL, input @ozanoktem, @aringh etc on this would be great.

Optionally, we could also want to remove the contributors list from the README and move it to a separate file, but I'd leave it for now.

adler-j:
Some of this has been done, but we still need to discuss if we want a code example or not. If we don't thats totally fine and we cna close this.

kohr-h:
If we can come up with a short but non-trivial example, that would be good. I can't think of anything mind-blowing from the top of my head. If there are no inputs in the near future I say we close this.

kohr-h:
Seems no-one has a killer example for the README. Shall we close?

adler-j:
I agree, perhaps it is hard to give one.

PR #710, opened 2016-11-09
Removed Optional Dependencies and Compatibility from readme, see #666

adler-j:
Basically a cleanup of things that made it to hard to read, now that the install doc is in, users can find this information there.


kohr-h:
Agree, merge!

PR #716, opened 2016-11-14
ENH: add reader for MRC2014 EM data

kohr-h:
Initial pull - a writer is on the TODO list. What's the best way of doing it? A separate writer class? Or a general `BlaFile` class? Having to open a new file in different mode from the original one would speak against it. Thoughts?

TODOs:
- [x] fix tests (PEP8 probably)
- [x] fix the stuff below
- [x] add writer
- [x] add utility to create headers
- [x] make classes context managers (reader is done)
- [x] process extended header at least for FEI type MRC format
- [x] add generic example for file format definition & header creation
- [x] add example for MRC
- [x] add test for raw I/O
- [x] add test for MRC combining read/write


odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+                spec_fields(MRC_2014_SPEC, id_key=MRC_SPEC_KEYS['id']),
+                keys=MRC_SPEC_KEYS, dtype_map=MRC_DTYPE_TO_NPY_DTYPE)
+        super().__init__(file, header_fields)
+
+    def _set_attrs_from_header(self):
+        """Set the following attributes of ``self`` from ``self.header``:
+
+            - ``data_shape`` : Shape of the (full) data.
+            - ``data_dtype`` : Data type of the data.
+            - ``cell_sides`` : Size of the unit cell in meters.
```
kohr-h: Add `cell_sides_angstrom`

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # cell_sides
+        self.cell_sides_angstrom = np.asarray(self.header['cella']['value'],
+                                              dtype=float)
+        self.cell_sides = self.cell_sides_angstrom * ANGSTROM
+
+        # mrc_version
+        nversion = self.header['nversion']['value']
+        maj_ver, min_ver = nversion // 10, nversion % 10
+        self.mrc_version = (maj_ver, min_ver)
```
kohr-h: In one step

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+import csv
+import numpy as np
+import struct
+
+
+__all__ = ('FileReaderUncompressedBinary',)
+
+
+# TODO: add rest
+DTYPE_MAP_NPY2STRUCT = {
```
kohr-h: Some obviously missing, like int64 and float64.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        Specification given as a string containing a definition table.
+
+    Returns
+    -------
+    lines : tuple of strings
+        Table lines with leading and trailing '|' stripped and lines
+        containing '...' removed.
+    """
+    return tuple(line[1:-1].rstrip() for line in spec.splitlines()
+                 if line.startswith('|') and '...' not in line)
```
kohr-h: This function is used only once in the `spec_fields` function. It's simple enough to be just inlined.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+              this shape.
+            - ``'description'`` : A human-readable description of the field.
+
+            For the default ``None``, the MRC2014 format is used, see
+            ``MRC2014_SPEC``.
+        """
+        if header_fields is None:
+            header_fields = standardized_fields(
+                spec_fields(MRC_2014_SPEC, id_key=MRC_SPEC_KEYS['id']),
+                keys=MRC_SPEC_KEYS, dtype_map=MRC_DTYPE_TO_NPY_DTYPE)
```
kohr-h: This is redundant, see comment [below](https://github.com/odlgroup/odl/pull/716/files#r87861953).

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        new_field = {}
+
+        # Name and description: lowercase name, copy description
+        new_field['name'] = field[keys['name']].lower()
+        new_field['description'] = field[keys['description']]
+
+        # Get offset from ID range
+        num_range = field[keys['id']].split('-')
+        nstart = int(num_range[0])
+        nend = int(num_range[-1])  # 0 for range of type 3-
```
kohr-h: Not correct: `num_range[-1] == ''` in this case, which raises an error when passed to `int`.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+            raise ValueError(
+                'in row {}: byte range {} not a multiple of itemsize {} '
+                'of the data type {}.'
+                ''.format(row + 1, field[keys['byte_range']], dtype.itemsize,
+                          field[keys['dtype']]))
+
+        new_field['dtype'] = dtype
+        new_field['size_bytes'] = byte_size
+        # Assuming 1d arrangement of multiple elements
+        # TODO: find way to handle 2d fields
```
kohr-h: Probably not necessary - better to let users reshape themselves.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+            new_field['dshape'] = (1,)
+
+        conv_list.append(new_field)
+
+    return tuple(conv_list)
+
+
+class FileReaderUncompressedBinary(object):
+
+    """Reader for uncompressed binary files including a header."""
```
kohr-h: Mention `from_raw_file` to skip the header stuff.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        try:
+            f = open(file, 'rb')
+        except TypeError:
+            f = file
+
+        if f.mode != 'rb':
+            raise ValueError("`file` must be opened in 'rb' mode, but mode "
+                             "'is {}'".format(f.mode))
+
+        self.file = f
```
kohr-h: Use `self.file` above

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+            is allowed to be already opened in 'rb' mode.
+        header_bytes : int
+            Size of the header in bytes.
+        dtype :
+            Data type specifier for the data field. It must be
+            understood by the `numpy.dtype` constructor.
+
+        Returns
+        -------
+        reader : `MRCFileReader`
```
kohr-h: Wrong class

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+            Raw reader for the given MRC file.
```
kohr-h: ~~MRC~~

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        try:
+            f = open(file, 'rb')
+        except TypeError:
+            f = file
```
kohr-h: Use `file` in try clause and `pass` in the except clause.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        filesize_bytes = f.seek(0, 2)
+        data_bytes = filesize_bytes - header_bytes
+        f.seek(0)
+
+        if header_bytes >= filesize_bytes:
+            raise ValueError('`header_bytes` is larger or equal to file size '
+                             '({} >= {})'.format(header_bytes, filesize_bytes))
+
+        if dtype is None:
+            raise TypeError('`dtype` cannot be `None`')
```
kohr-h: To prevent the `np.dtype(None) == float` behavior.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        data_size = data_bytes / dtype.itemsize
+
+        instance = cls(f, header_fields=[])
+        instance.data_dtype = dtype
+        instance.data_shape = (data_size,)
+        instance.header_bytes = header_bytes
+        return instance
+
+    def read_header(self):
+        """Read the header from the reader's file.
```
kohr-h: \`\`self.file\`\`

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def _set_attrs_from_header(self):
+        """Abstract method for setting attributes from ``self.header``.
+
+        The minimum attributes to set by this method are:
+
+            - ``data_shape`` : Shape of the (full) data.
+            - ``data_dtype`` : Data type of the data.
+
+        Subclasses must override this method. They may define additional
```
kohr-h: Overriding is not a must, but otherwise the `read_header` method is not usable.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+            f = file
+
+        if f.mode != 'rb':
+            raise ValueError("`file` must be opened in 'rb' mode, but mode "
+                             "'is {}'".format(f.mode))
+
+        self.file = f
+
+        # Initialize some attributes to default values, plus the header
+        self.data_shape = -1  # Makes reshape a no-op
```
kohr-h: Set `self.data_dtype` to something.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        self.file.seek(dstart_abs)
+        # TODO: use byteorder according to header
+        data = np.fromfile(self.file, dtype=self.data_dtype,
+                           count=int(num_elems))
+
+        if dstart_abs == self.header_bytes and dend_abs == filesize_bytes:
+            # Full dataset read, reshape to stored shape.
+            # Use 'F' order in reshaping since it's the native MRC data
+            # ordering.
+            data = data.reshape(self.data_shape, order='F')
```
kohr-h: Order should be a method parameter. This is still hard-coded for MRC.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+                             ''.format(dend_abs, filesize_bytes))
+
+        num_elems = (dend_abs - dstart_abs) / self.data_dtype.itemsize
+        if num_elems != int(num_elems):
+            raise ValueError('trying to read {} bytes, corresponding to '
+                             '{} elements of type {}'
+                             ''.format(dend_abs - dstart_abs, num_elems,
+                                       self.data_dtype))
+        self.file.seek(dstart_abs)
+        # TODO: use byteorder according to header
```
kohr-h: Make `byte_order` a method parameter.

odl/tomo/data/uncompr_bin.py
adler-j: A file with this name and module should, i.m.o. have absolutely no references to MRC at all. Easiest if you want to keep a MRC reference (for this pull) is to simply move this class into the MRC module and name it something like `MRC...`.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        raise NotImplementedError
+
+    def read_data(self, dstart=None, dend=None):
+        """Read the data from the reader's file and store if desired.
+
+        Parameters
+        ----------
+        dstart : int, optional
+            Offset in bytes of the data field. By default, it is equal
+            to ``header_size``. Negative values are added to the file
```
adler-j: Negative values ...

Could say something like

Supports python style negative indexing.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+    'Int32': np.dtype('int32'),
+    'Char': np.dtype('S1')}
+
+MRC_MODE_TO_NPY_DTYPE = {
+    0: np.dtype('uint8'),
+    1: np.dtype('int16'),
+    2: np.dtype('float32'),
+    6: np.dtype('uint16')}
+
+ANGSTROM = 1e-10
```
adler-j: Dislike raw numbers, it should always be conversions:

`ANGSTROM_IN_METERS = 1e-10`

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
++---------+-------+---------+--------+-------------------------------+
+|9        |33-36  |Int32    |MY      |Number of intervals along Y of |
+|         |       |         |        |the "unit cell"                |
++---------+-------+---------+--------+-------------------------------+
+|10       |37-40  |Int32    |MZ      |Number of intervals along Z of |
+|         |       |         |        |the "unit cell"                |
++---------+-------+---------+--------+-------------------------------+
+|11-13    |41-52  |Float32  |CELLA   |Cell dimension in angstroms    |
++---------+-------+---------+--------+-------------------------------+
+|...      |       |         |        |                               |
```
adler-j: ...?

adler-j:
Also would be nice if you could get someone at CWT who is more knowledgeable about this application and data format to do a review!

odl/tomo/data/mrc2014.py
kohr-h: To indicate we're skipping fields - not really needed but makes it a bit more clear.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+    6: np.dtype('uint16')}
+
+ANGSTROM = 1e-10
+
+
+class FileReaderMRC(FileReaderUncompressedBinary):
+
+    """Reader for the MRC file format(s).
+
+    By default, the MRC2014 format is used, see ``MRC_2014_SPEC`` for
```
kohr-h: I can add it to `__all__` but it's not possible to make it clickable that way.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+    details. See also [Che+2015]_ or the `explanations on the CCP4 homepage
+    <http://www.ccpem.ac.uk/mrc_format/mrc2014.php>`_ for the
+    text of the specification.
+
+    References
+    ----------
+    [Che+2015] Cheng, A, Henderson, R, Mastronarde, D, Ludtke, S J,
```
kohr-h: Right.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,197 @@
...
+    'description': 'Description'}
+
+# Add more if needed
+MRC_DTYPE_TO_NPY_DTYPE = {
+    'Float32': np.dtype('float32'),
+    'Int32': np.dtype('int32'),
+    'Char': np.dtype('S1')}
+
+MRC_MODE_TO_NPY_DTYPE = {
+    0: np.dtype('uint8'),
```
kohr-h: Wrong, should be signed

odl/tomo/data/uncompr_bin.py
kohr-h: I don't know - the only thing that could be more specific is that it's supposed to be a fixed-size header. Other than that, it's pretty generic.

odl/tomo/data/mrc2014.py
adler-j: Hmm, you are correct. Irritating, not quite sure how to handle this. Is it not possible to add module variables to the doc?

odl/tomo/data/uncompr_bin.py
kohr-h: A suggestion would be to use a data type `Float32(3,2)` if a certain shape is desired. The function should in any case explain how the table is supposed to look like.

odl/tomo/data/uncompr_bin.py
adler-j: Well some examples that this doesnt handle:

* The binary format used by EMRecon, where they give `n` data points and then a time stamp, i.e.

  ```
  data1
  data2
  timestamp1
  data3
  timestamp2
  ```

* Non-numpy based data:

  ```
  binary_pickle1
  binary_pickle2
  ```
  
* Formats where the interpretation depends on the header

  ```
  dtype=int
  [binary int 1]
  [binary int 2]
  ```
  
  vs
  
  
  ```
  dtype=float
  [binary float 1]
  [binary float 2]
  ```
  
* Formats where each object uses more than one line

  ```
  object1_part1
  object1_part2
  object2_part1
  object2_part2
  ```
  
* Formats where there are sub-headers for each object

  ```
  object1_header
  object1_data1  
  object1_data2  
  object1_data3 
  object2_header
  object2_data1
  ```
  
I could probably go on, but I'm just trying to say that there are heaps of formats that fit the name of this class that do not work with it.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        self.data_shape = -1  # Makes reshape a no-op
+        self.header_bytes = None
+        self.header_fields = header_fields
+        self.data = None
+        if self.header_fields:
+            self.header = self.read_header()
+        else:
+            self.header = None
+
+    @classmethod
```
kohr-h: I'll see if I can use the standard initializer for this case, too. 

Generally class methods `from...` are the standard way of having alternative constructors for classes, see e.g. [dict.fromkeys](https://docs.python.org/3.5/library/stdtypes.html#dict.fromkeys). Those alternative constructors are used whenever the construction pattern (mostly input parameters) contradict the one of the standard constructor, which may either be inherited or designed for subclassing (like in this case). In fancy words, alternative constructors are needed whenever they would violate the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) - good practice is to compartmentalize these violators.

That said, I'll see if we really need it.

adler-j: Ok, I'm fine with having this if it would make the default construction weird or convoluted. Just want to make sure that is actually the case.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+                    "field '{}': format '{}' has results in {} bytes, but "
+                    "`size_bytes` is {}"
+                    "".format(name, fmt, struct.calcsize(fmt), size_bytes))
+
+            self.file.seek(offset_bytes)
+            packed_value = self.file.read(size_bytes)
+            value = np.array(struct.unpack_from(fmt, packed_value),
+                             dtype=dtype)
+
+            if dtype == np.dtype('S1'):
```
kohr-h: For me this is the simplest way of making the string case consistent with the other cases. The specification says "4 bytes of type Char", which is equivalent to `np.array(..., dtype='S1').shape == (4,)`. I may need to not `struct.unpack` them if they are to be read as strings. Some testing required.

odl/tomo/data/uncompr_bin.py
kohr-h: IMO, that version says much less that the text right now - "python style" is not a well-defined notion.

odl/tomo/data/uncompr_bin.py
kohr-h: > The binary format used by EMRecon, where they give n data points and then a time stamp, i.e.

Okay, I'll add **one** header and **one** block of data.

> Non-numpy based data:

The reader does in no way rely on Numpy's format. Depending on how it's pickled it actually is supported. Numpy's `.npy` format happens to be a header-followed-by-data-block format, and that is supported by the reader.

> Formats where the interpretation depends on the header

Supported. Actually, that's exactly what happens in MRC. It has a `mode` header entry which specifies how the data is to be understood, i.e. the data type.

> Formats where each object uses more than one line

I'm not sure what you mean by that. The reader is not line-based.

> Formats where there are sub-headers for each object

Okay, that's the same case as in the first example.

> I could probably go on, but I'm just trying to say that there are heaps of formats that fit the name of this class that do not work with it.

With the "one header, one data block" restriction, what is left that isn't supported?

Regarding the name, I'm not sure how to make it adequate and not ridiculously long at the same time. `FileReaderSingleHeaderUncompressedBinaryData`?

odl/tomo/data/uncompr_bin.py
kohr-h: Yes, I think there is no elegant way around this. What the constructor needs to do is to use a raw specification `header_bytes` and `dtype` and bypass the usual constructor completely. I see no way of doing that by calling `__init__` with some special case of arguments. The problem is that the size of the header is calculated on the fly since it depends on an entry in the header that gives the size of the extended header (MRC). Such a thing seems to be used quite often (e.g. also in the `.npy` format).

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,472 @@
...
+        if f.mode != 'rb':
+            raise ValueError("`file` must be opened in 'rb' mode, got '{}'"
+                             "".format(f.mode))
+
+        header_bytes = int(header_bytes)
+        if header_bytes < 0:
+            raise ValueError('`header_bytes` must be nonnegative, got {}.'
+                             ''.format(header_bytes))
+
+        filesize_bytes = f.seek(0, 2)
```
kohr-h: The 2 means "backwards from the end of the file". Unfortunately it cannot be used as keyword argument, so a comment will have to explain this.

kohr-h:
> My biggest isuse with this as is is that it is hanging code. Frankly, I dont see how I could use it from the code as is, and no other code is using it. We need
> 
>  - Example of using this

Added to the TODO list.

>  - Example of actually loading data with this and creating a geometry etc (I assume that is the end goal?)
> 
> Why not also make a "end to end" function that takes a file and gives you a geometry and data? Much easier on users.

The problem is that the information in the file is not sufficient to create a geometry. There is, for example, no information about the angles, at least not in the standard header. You also don't know the tilt axis, which is fine if it agrees with *some* coordinate axis, but actually it's often slightly rotated. Sometimes it's stored in the additional text labels, but that's not standardized. Then, the pixel size of the detector is usually stored as `actual_size / magnification` such that it's on the same scale as the reconstruction, but unfortunately the magnification is stored nowhere. Then there are many important experimental parameters that are not stored at all, and you have to get them from some other source. Of course, most people don't care about this since they only use FBP or SIRT with plain ray transform.

So, unless there is added functionality in form of a config file and a parser for it (which I have lying around somewhere, I guess), no end-to-end function is possible.

odl/tomo/data/uncompr_bin.py
kohr-h: `np.issubdtype(dtype, np.dtype('S'))`

odl/tomo/data/uncompr_bin.py
kohr-h: Now that I think again about this, I actually see an alternative. This method can be scrapped completely, the header size set to 0, and users be advised to use the `dstart` parameters in the `read_data` method. It does the same.

adler-j:
> So, unless there is added functionality in form of a config file and a parser for it (which I have lying around somewhere, I guess), no end-to-end function is possible.

How about:

```python
def make_data_and_geometry(file_path, angles, tilt_axis, pixel_size, ...):
    stuff
```
?

But perhaps it is fine once the example has been added.

odl/tomo/data/uncompr_bin.py
adler-j: Oh damn, well just make a comment then.

odl/tomo/data/mrc2014.py
kohr-h: There seems to be [a solution](https://stackoverflow.com/questions/20227051/how-to-document-a-module-constant-in-python#20227174), and [the Sphinx documentation](http://www.sphinx-doc.org/en/1.4.8/ext/autodoc.html#directive-autodata) describes this usage of an `.. autodata::` directive for this purpose. But I have no idea where to put this directive. Would it have to go into the `generate_doc.py` script?

odl/tomo/data/mrc2014.py
```diff
@@ -20,72 +20,75 @@
...
+|         |        |          |        |identify file type             |
++---------+--------+----------+--------+-------------------------------+
+|54       |213-216 |String    |MACHST  |Machine stamp                  |
++---------+--------+----------+--------+-------------------------------+
+|55       |217-220 |Float32   |RMS     |RMS deviation of map from mean |
+|         |        |          |        |density                        |
++---------+--------+----------+--------+-------------------------------+
+|56       |221-224 |Int32     |NLABL   |Number of labels being used    |
++---------+--------+----------+--------+-------------------------------+
+|57-256   |225-1024|String(10)|LABEL   |10 80-character text labels    |
```
adler-j: String(80)?

odl/tomo/data/uncompr_bin.py
```diff
@@ -22,12 +22,13 @@
 import struct
 
 
-__all__ = ('FileReaderUncompressedBinary',)
+__all__ = ('FileReaderUncompressedBinary', 'header_fields_from_table')
 
 
-# TODO: add rest
 DTYPE_MAP_NPY2STRUCT = {
```
adler-j: These look scarily much like regular format characters:

https://docs.python.org/3/library/struct.html#format-characters

odl/tomo/data/mrc2014.py
kohr-h: Makes more sense, right. I should fill to the left.

odl/tomo/data/mrc2014.py
```diff
@@ -20,72 +20,75 @@
...
-+---------+-------+---------+--------+-------------------------------+
-|53       |209-212|Char     |MAP     |Character string 'MAP' to      |
-|         |       |         |        |identify file type             |
-+---------+-------+---------+--------+-------------------------------+
-|54       |213-216|Char     |MACHST  |Machine stamp                  |
-+---------+-------+---------+--------+-------------------------------+
-|55       |217-220|Float32  |RMS     |RMS deviation of map from mean |
-|         |       |         |        |density                        |
-+---------+-------+---------+--------+-------------------------------+
++---------+--------+----------+--------+-------------------------------+
```
kohr-h: Because the table will look much better when used in documentation (once I figure out how to add it). And it's not that much harder to parse it than to process the list.

adler-j: Still rather sceptical that we should force all future users to give input in this rather complicated format. If a user wants to dynamically set stuff this will become hard.

Why not add the parameter as I wrote it, and then have the table as a docstring for it?

adler-j: Fine by me as well!

kohr-h: Actually, I think this is simple enough that users can take that way of specifying a format directly. I'll do the following:
- shorten the keys a bit (remove the `'bytes'`)
- use conversion on the values (to integer, `np.dtype`)
- make an example with a super-simple file format (2d, shape, cell size and origin information)

kohr-h:
About naming of the uncompressed binary thing: any suggestions?

adler-j:
I'm fine with something like `FileReaderUncompressedBinaryWithHeader`

kohr-h:
> I'm fine with something like FileReaderUncompressedBinaryWithHeader

Yeah, I think I'll shorten it to `FileReaderRawBinaryWithHeader`

odl/tomo/data/mrc2014.py
```diff
@@ -86,10 +92,21 @@
...
 
+
+def print_mrc2014_spec():
+    """Print the MRC2014 specification table.
+
+    The specification table is as follows:
+    """
+    print(MRC_2014_SPEC_TABLE)
+
+print_mrc2014_spec.__doc__ += MRC_2014_SPEC_TABLE
```
adler-j: For adding this to the documentation I'm fine with this approach.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,663 @@
...
+            self.__header = self.read_header()
+        else:
+            self.__header = None
+
+    @property
+    def file(self):
+        """File object from which ``self`` reads."""
+        return self.__file
+
+    def close(self):
```
adler-j: How about letting users use this class with `with`? Cleaner than requiring that they call this.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,663 @@
...
+            else:
+                value = np.array(struct.unpack_from(fmt, packed_value),
+                                 dtype=dtype)
+                entry['value'] = value.reshape(shape)
+
+            header[name] = entry
+
+        # Store information gained from the header
+        self.__header = header
+        if self.__set_attrs_from_header:
```
adler-j: wut?

adler-j: Is the naming truly the best we have here? This one has a implied questionmark (**should I** `__set_attrs_from_header?`) while the other one is a implication "**i want you to** `_set_attrs_from_header`". Got to this line as was like, wut?

kohr-h: A bit ugly, admittedly. Probably the better approach is to make the convenience attributes properties and check for existence there.

examples/tomo/data/mrc_io.py
```diff
@@ -0,0 +1,102 @@
...
+(temporary) file is written to disk.
+This simulates the start and end of a reconstruction workflow, where
+first projection data is read from disk, and finally the reconstructed
+volume is written back to disk, using a different header.
+"""
+
+from __future__ import print_function
+
+try:
+    import matplotlib.pyplot as plt
```
adler-j: Lets not have stuff like this, the whole library basically has matplotlib as a dependency, skipping it is if you truly want to go minimal.

kohr-h:
All done, time for another review.

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
   - "3.5"
+cache: pip
```
adler-j: eh?

kohr-h: Yeah, dunno. You can [cache pip files](https://docs.travis-ci.com/user/caching/#pip-cache) between builds so they don't have to be built again every time. Not sure if that makes any difference time-wise or if it doesn't screw up stuff.

kohr-h:
Rebased and history cleaned up slightly. Good for another round of scrutiny.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+        -------
+        data : `numpy.ndarray`
+            The data read from `file`.
+        """
+        data = super().read_data(dstart, dend, reshape_order='F')
+        if swap_axes:
+            data = np.transpose(data, axes=self.data_axis_order)
+        return data
+
+    # TODO: read extended header for the standard flavors, see the spec
```
adler-j: this done?

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+
+    References
+    ----------
+    [Che+2015] Cheng, A et al. *MRC2014: Extensions to the MRC format header
+    for electron cryo-microscopy and tomography*. Journal of Structural
+    Biology, 129 (2015), pp 146--150.
+    """
+
+    print_mrc2014_spec = staticmethod(print_mrc2014_spec)
+
```
adler-j: no print extended here?

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+    @property
+    def header_size(self):
```
adler-j: isnt this duplicated, shouldnt we refactor this to a shared base class?

adler-j: Applies to many of these functions

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+            dstart = int(dstart)
+
+        if dstart < self.header_size:
+            raise ValueError('invalid `dstart`, resulting in absolute '
+                             '`dstart` < `header_size` ({} < {})'
+                             ''.format(dstart, self.header_size))
+
+        data = np.asarray(data, dtype=self.data_dtype).reshape(shape)
+        if swap_axes:
+            data = np.transpose(data, axes=self.data_axis_order)
```
adler-j: Shouldnt this be the inverse permutation? If we want this to match the other guy

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+
+    # Keyword args
+    extent = kwargs.pop('extent', num_cells)
+    axis_order = kwargs.pop('axis_order', (0, 1, 2))
+    if tuple(axis_order) not in permutations((0, 1, 2)):
+        raise ValueError('`axis_order` must be a permutation of (0, 1, 2), '
+                         'got {}'.format(axis_order))
+    dmin = kwargs.pop('dmin', 1.0)
+    dmax = kwargs.pop('dmax', 0.0)
+    dmean = kwargs.pop('dmean', min(dmin, dmax) - 1.0)
```
adler-j: I dont get the default

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+        raise ValueError('labels cannot have more than 80 characters each')
+
+    # Convert to header-friendly form. Names are required to match
+    # exactly the header field names, and all of them must exist,
+    # so that `eval` below succeeds for all fields.
+    nx, ny, nz = [np.array(n, dtype='int32').reshape([1]) for n in num_cells]
+    mode = np.array(NPY_DTYPE_TO_MRC_MODE[np.dtype(dtype)],
+                    dtype='int32').reshape([1])
+    mx, my, mz = nx, ny, nz
+    cella = (np.array(extent).reshape([3])).astype('float32')
```
adler-j: remove extra ()

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+    text_labels = [str(label).ljust(80) for label in text_labels_in]
+    if any(len(label) > 80 for label in text_labels):
+        raise ValueError('labels cannot have more than 80 characters each')
+
+    # Convert to header-friendly form. Names are required to match
+    # exactly the header field names, and all of them must exist,
+    # so that `eval` below succeeds for all fields.
+    nx, ny, nz = [np.array(n, dtype='int32').reshape([1]) for n in num_cells]
+    mode = np.array(NPY_DTYPE_TO_MRC_MODE[np.dtype(dtype)],
+                    dtype='int32').reshape([1])
```
adler-j: what does the reshape do?

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+
+    # Make the header
+    # We use again the specification to set the values
+    header_fields = header_fields_from_table(
+        MRC_2014_SPEC_TABLE, MRC_SPEC_KEYS, MRC_DTYPE_TO_NPY_DTYPE)
+
+    header = OrderedDict()
+    for field in header_fields:
+        header[field['name']] = {'offset': field['offset'],
+                                 'value': eval(field['name'])}
```
adler-j: ough, eval? no other way?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+import struct
+
+
+__all__ = ('FileReaderRawBinaryWithHeader',
+           'FileWriterRawBinaryWithHeader',
+           'header_fields_from_table')
+
+
+def _fields_from_table(spec_table, id_key):
+    """Read the specification and return a list of fields.
```
adler-j: read **a** specification?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+    Returns
+    -------
+    fields : tuple of dicts
+        Field list of the specification with combined multi-line entries.
+        Each field corresponds to one (multi-)line of the spec.
+    """
+    # Reformat the table, throwing away lines not starting with '|' or
+    # containing '...'.
+    spec_lines = [line[1:-1].rstrip() for line in spec_table.splitlines()
+                  if line.startswith('|') and '...' not in line]
```
adler-j: this can bug out if some description contains "...", do we accept that?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+"""Utilities and class for reading uncompressed binary files with header."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import int, object, open, str
+
+from collections import OrderedDict
+import csv
```
adler-j: is this a core library?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+    The specification table is assumed to be in
+    `reST grid table format
+    <http://docutils.sourceforge.net/docs/user/rst/quickref.html#tables>`_.
+    It must have the following 5 columns:
+
+        1. ID : an arbitrary unique identifier, e.g., a number.
+        2. Byte range : Bytes in the file covered by this field, given
+           as number range (e.g. ``15-24``). The byte values start at
+           1 (not 0), and the upper value of the range is included.
+        3. Data type : Field values are stored in this format. For multiple
```
adler-j: mention that this should be a numpy dtype

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+    """
+
+    def __init__(self, file, header_fields=(), dtype=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        file : file-like or str
+            Stream or filename from which to read the data. The stream
+            is allowed to be already opened in ``'rb'`` mode.
```
adler-j: is allowed to _> should be?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        header_fields_from_table :
+            Function to parse a specification table, returning a field
+            sequence usable as ``header_fields`` parameter.
+        """
+        # Need those attrs in subsequent code
+        file_attrs = ('mode', 'seek', 'read', 'close')
+        is_file = all(hasattr(file, attr) for attr in file_attrs)
+        if is_file:
+            self.__file = file
+            self.__file_shared = True
```
adler-j: change to positive: __owns_file

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        -------
+        data : `numpy.ndarray`
+            The data read from `file`.
+
+        See Also
+        --------
+        read_header
+        """
+        self.file.seek(0, 2)  # 2 means "from the end"
+        filesize_bytes = self.file.tell()
```
adler-j: what type does filesize_bytes have here?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+                             '`dstart` < `header_size` ({} < {})'
+                             ''.format(dstart_abs, self.header_size))
+        if dend_abs > filesize_bytes:
+            raise ValueError('invalid `dend`, resulting in absolute '
+                             '`dend` > `filesize_bytes` ({} < {})'
+                             ''.format(dend_abs, filesize_bytes))
+
+        num_elems = (dend_abs - dstart_abs) / self.data_dtype.itemsize
+        if num_elems != int(num_elems):
+            raise ValueError('trying to read {} bytes, corresponding to '
```
adler-j: This error does not (to me at least) say what went wrong.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+                             '{} elements of type {}'
+                             ''.format(dend_abs - dstart_abs, num_elems,
+                                       self.data_dtype))
+        self.file.seek(dstart_abs)
+        # Numpy determines byte order by itself
+        data = np.fromfile(self.file, dtype=self.data_dtype,
+                           count=int(num_elems))
+
+        if dstart_abs == self.header_size and dend_abs == filesize_bytes:
+            # Full dataset read, reshape to stored shape.
```
adler-j: Wait, this is NOT obvious. Sure, the doc mentions it but i guarantee that this will catch 98% of users of guard anyway (mostly because the parameter is an optional parameter and they may not even be using it). No other way to solve this?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+    """Writer for uncompressed binary files using an optional header.
+
+    This class can be used to write a single binary header followed by a
+    single block of uncompressed binary data to a file.
+    Alternatively, the header can be bypassed and data blocks can be
+    written directly using `write_data`, which allows to write arbitrary
+    portions.
+    """
+
+    def __init__(self, file, header=OrderedDict()):
```
adler-j: The default cannot be a mutable. See e.g.

```python
from collections import OrderedDict

def f(name, x=OrderedDict()):
    x[name] = 'value'
    return x

print(f('1'))  # OrderedDict([('1', 'value')])
print(f('2'))  # OrderedDict([('1', 'value'), ('2', 'value')])
```

You need to use `None` and handle it inside.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        file_attrs = ('mode', 'seek', 'write', 'close')
+        is_file = all(hasattr(file, attr) for attr in file_attrs)
+        if is_file:
+            self.__file = file
+            self.__file_shared = True
+        else:
+            self.__file = open(file, 'wb')
+            self.__file_shared = False
+
+        if 'b' not in self.file.mode:
```
adler-j: what about 'w'?

adler-j: Also this should perhaps be inside the if statement above.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+    @property
+    def header_size(self):
+        """Size of `header` in bytes."""
+        if not self.header:
+            return 0
+
+        # Determine header size by finding the largest offset and the
+        # value of the corresponding entry. The header size is the
+        # offset plus the size of the entry.
+        max_offset = 0
```
adler-j: Here you basically re-implement max. Why not something like

```python
max_key = max(self.header, key=lambda key: self.header[key]['offset'])
max_offset = self.header[max_key]['offset']
max_value = self.header[max_key]['value']
```

also makes it more obvious what you're aiming at.

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        self.write_data(data)
+
+    def write_header(self):
+        """Write `header` to `file`.
+
+        See Also
+        --------
+        read_data
+        """
+        for name, properties in self.header.items():
```
adler-j: name un-used? I guess you should look over values then?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        """
+        self.write_header()
+        self.write_data(data)
+
+    def write_header(self):
+        """Write `header` to `file`.
+
+        See Also
+        --------
+        read_data
```
adler-j: write_data?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        ----------
+        data : `array-like`
+            Data that should be written to `file`.
+        dstart : non-negative int, optional
+            Offset in bytes of the start position of the written data.
+            By default, it is taken to be `header_size`.
+        reshape_order : {'C', 'F', 'A'}, optional
+            Value passed as ``order`` parameter to `numpy.reshape`.
+            Reshaping is only done in case the whole data block is read.
+        """
```
adler-j: See also?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        data = np.asarray(data).reshape(-1, order=reshape_order)
+        if dstart is None:
+            dstart = int(self.header_size)
+        elif dstart < 0:
```
adler-j: Why dont we do the usual negative indexing here?

odl/tomo/data/uncompr_bin.py
```diff
@@ -0,0 +1,716 @@
...
+        """Initializer for the context manager."""
+        return self
+
+    def __exit__(self, *exc):
+        """Cleanup before on exiting the context manager."""
+        if not self.__file_shared:
+            self.file.close()
+
+    @property
+    def header_size(self):
```
kohr-h: I was thinking about that. I'm counting 5 common properties. Does that warrant a base class? Apart from that, they're not that terribly similar. Probably it's better to use a common mixin class instead of a common base.

odl/tomo/data/uncompr_bin.py
kohr-h: Integer.

odl/tomo/data/uncompr_bin.py
adler-j: Anything goes for me, just pointing this out. For inheritance, I agree common base classes add even more complexity so perhaps should be skipped.

odl/tomo/data/uncompr_bin.py
kohr-h: Well, another option is to not reshape at all. Obviously reshaping is not possible if you read an arbitrary chunk of data. On the other hand, if the header specifies a shape of the data, you probably want your data to be in that shape after reading it from the file.
An alternative would be a boolean parameter reshape yes/no, but that feels a bit redundant and won't play well with the chunk reading situation (ignore? raise exception?).
Yet another alternative would be to never reshape and let the user do it. But then, as usual, they can get the order wrong, and when axis ordering enters the game, things get really awful.

So I would very much want to offer the possibility to reliably get the data reshaped and transposed correctly from the information in the header. It's not so relevant here in the base class, but in MRC it definitely is.

odl/tomo/data/uncompr_bin.py
kohr-h: Sure

odl/tomo/data/uncompr_bin.py
kohr-h: The file could be opened in `r+` mode which would also be writable. I had a check for `file.is_writable()` before, but temporary files don't have that method. So basically I'm relying on the OS to tell the user whether or not they can write. Not everything has to be double-checked.

examples/tomo/data/raw_binary_io_with_header.py
```diff
@@ -0,0 +1,115 @@
...
+# In practice, these could come from a `DiscreteLp` space as `mid_pt`
+# and `cell_sides` properties.
+image = scipy.misc.ascent()
+shape = np.array(image.shape, dtype='int32')
+origin = np.array([-1.0, 0.0], dtype='float32')
+px_size = np.array([0.1, 0.1], dtype='float32')
+# To make it storable as binary data, we take the string version of the data
+# type with a fixed size of 10 characters and encode it as array of single
+# bytes.
+dtype = np.fromiter(str(image.dtype).ljust(10), dtype='S1')
```
kohr-h: To make it work with `array` this is required:
```python
np.array(list(str(np.dtype('float32')).ljust(10)), dtype='S1')
```
Is that any simpler?

kohr-h: Perhaps in two steps:
```python
dtype_str = str(image.dtype).ljust(10)
dtype = np.array(list(dtype_str), dtype='S1')
```

examples/tomo/data/raw_binary_io_with_header.py
```diff
@@ -0,0 +1,115 @@
...
+
+# Create the header
+# Use an OrderedDict for the header to have a predictable order when
```
kohr-h: It is -- otherwise the header is stored in random order in the file. The header writing routine loops over `header.items()` and that order is unpredictable for regular `dict`s.

odl/test/tomo/data/mrc2014_test.py
```diff
@@ -0,0 +1,243 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Tests for the MRC I/O routines."""
```
kohr-h: Yes, or older versions for that matter.

odl/test/tomo/data/mrc2014_test.py
```diff
@@ -0,0 +1,243 @@
...
+    for v in string_vars:
+        assert v.dtype == np.dtype('S1')
+
+
+def test_mrc_header_from_params_kwargs():
+    """Test the utility function for the minimal required parameters."""
+    ncells = (10, 20, 30)
+    dtype = np.dtype('int8')
+    kind = 'projections'
+    kwargs = {}
```
kohr-h: More DRY

odl/test/tomo/data/mrc2014_test.py
```diff
@@ -0,0 +1,243 @@
...
+    ispg = header['ispg']['value']
+    assert ispg == true_ispg
+    nsymbt = header['nsymbt']['value']
+    assert nsymbt == 0
+    exttype = header['exttype']['value']
+    assert np.array_equal(exttype, np.fromiter('    ', dtype='S1'))
+    nversion = header['nversion']['value']
+    assert nversion == 20140
+    origin = header['origin']['value']
+    assert np.array_equal(origin, np.zeros(3, dtype=int))
```
kohr-h: `label` is a `(10, 80)` array of empty single byte strings, so your suggestions doesn't work. Also the `all_equal` thing doesn't work well with Numpy arrays that are not 1D.

odl/test/tomo/data/uncompr_bin_test.py
```diff
@@ -0,0 +1,111 @@
...
+axis_order_params = list(permutations((0, 1, 2)))
+axis_order_ids = [' axis_order = {} '.format(p) for p in axis_order_params]
+
+
+@pytest.fixture(scope='module', ids=axis_order_ids, params=axis_order_params)
+def axis_order(request):
+    return request.param
+
+
+shape_params = [(5, 10, 20), (1, 5, 6), (10, 1, 1), (1, 1, 1)]
```
kohr-h: Good idea, done (44e4de09d2bc970517e9f0ebca1259c00aa1f4e6).

examples/tomo/data/raw_binary_io_with_header.py
kohr-h: Sure, that reads better than `fromiter`. The `ljust(10)` has to stay, though.

odl/tomo/data/uncompr_bin.py
kohr-h: It's part of the standard library, yes.

odl/tomo/data/uncompr_bin.py
kohr-h: Right

odl/tomo/data/uncompr_bin.py
kohr-h: It's not 100 % watertight, I agree. The best thing is to remove that "feature" and to just squeeze the table.

odl/tomo/data/fei_extended_header.py
```diff
@@ -0,0 +1,101 @@
...
+    print(MRC_FEI_EXT_HEADER_SECTION)
+
+print_fei_ext_header_spec.__doc__ += MRC_FEI_EXT_HEADER_SECTION
+
+
+MRC_FEI_SECTION_SIZE = 128
+MRC_FEI_MAX_SECTIONS = 1024
+
+ANGSTROM_IN_METERS = 1e-10
+MICRON_IN_METERS = 1e-6
```
kohr-h: Well.. the initial thought was that it's somewhat optional so perhaps the functionality can be outsourced, but now all that's left is the specification and some constant definitions. I guess it can go into the main file.

odl/tomo/data/uncompr_bin.py
kohr-h: Did this now:
```python
        max_entry = max(self.header.values(),
                        key=lambda val: val['offset'])
        return max_entry['offset'] + max_entry['value'].nbytes
```

odl/tomo/data/uncompr_bin.py
kohr-h: `file.seek` does not support negative offsets.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+        """
+        if header_fields is None:
+            header_fields = header_fields_from_table(
+                spec_table=MRC_2014_SPEC_TABLE,
+                keys=MRC_SPEC_KEYS,
+                dtype_map=MRC_DTYPE_TO_NPY_DTYPE)
+
+        super().__init__(file, header_fields)
+
+    print_mrc2014_spec = staticmethod(print_mrc2014_spec)
```
kohr-h: This preserves the docstring. It's equivalent to decorating an inline function definition with `@staticmethod`.

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+        else:
+            return tuple(int(m) - 1 for m in (mapc, mapr, maps))
+
+    @property
+    def cell_sides_angstrom(self):
+        """Array of sizes of a unit cell in Angstroms.
+
+        The value is determined from the ``'cella'`` entry in `header`.
+        """
+        return (np.asarray(self.header['cella']['value'], dtype=float) /
```
kohr-h: IMO the readability of
```python
        return np.asarray(self.header['cella']['value'],
                          dtype=float) / self.num_cells
```
is worse than
```python
        return (np.asarray(self.header['cella']['value'], dtype=float) /
                self.num_cells)
```
In the latter case, you split at the operator, in contrast to an arbitrary point in the function call as in the first case.
Third option, dunno:
```python
        return (np.asarray(self.header['cella']['value'],
                           dtype=float) /
                self.num_cells)
```
Kind of ugly?

odl/tomo/data/mrc2014.py
```diff
@@ -0,0 +1,846 @@
...
+           http://www.2dx.unibas.ch/documentation/mrc-software/fei-\
+extended-mrc-format-not-used-by-2dx
+        """
+        ext_header_type = str(force_type).upper() or self.extended_header_type
+        if ext_header_type != 'FEI1':
+            raise ValueError("extended header type '{}' not supported"
+                             "".format(self.extended_header_type))
+
+        groupby, groupby_in = str(groupby).lower(), groupby
+        if groupby not in ('field', 'section'):
```
kohr-h: ok

odl/tomo/data/mrc2014.py
kohr-h: Yeah, nah, I don't know. I was too lazy to implement the writer for the extended header. We won't need it anyway, quite sure. Or?

odl/tomo/data/mrc2014.py
kohr-h: Okay, I'll explain it to myself:
According to the documentation, we have the mapping
```python
num_cells[i] == data_shape[data_axis_order[i]]
```
which means that the x axis corresponds to the `data_axis_order[0]` axis in the data (and similar for y, z) *before transposing*, i.e. when reading raw from the file and reshaping into `data_shape`.
That means, in order to move the x axis into position 0, y to 1 and z to 2,  we need to permute the axes like in this example from the doc of `np.transpose`:
```python
>>> x = np.ones((1, 2, 3))
>>> np.transpose(x, (1, 0, 2)).shape
(2, 1, 3)
```
Since we need to transpose from `data_shape` to `num_cells`, we need the inverse behavior, since the argument to `transpose` is the "to" layout, you're right.
But then, the writer has to use the forward permutation instead. They can never be the same except for involutary permutations.

examples/tomo/data/raw_binary_io_with_header.py
adler-j: Ok cool!

odl/tomo/data/mrc2014.py
adler-j: I was thinking

```python
total_volume = np.asarray(self.header['cella']['value'], dtype=float)
return total_volume / self.num_cells
```

odl/tomo/data/mrc2014.py
adler-j: Frankly I hardly know what we need the writer for at all so that is up to you to decide :)

odl/tomo/data/mrc2014.py
adler-j: Perhaps it is enough if we are clear about this in the documentation

odl/tomo/data/uncompr_bin.py
adler-j: Looks great to me!

odl/tomo/data/mrc2014.py
kohr-h: Okay, some more thinking went into this. Actually it's the other way around. Here's the portion from the specification table (corrected the typos, ahem):
```
|17       |65-68   |Int32     |MAPC    |axis corresponding to columns  |
|         |        |          |        |(1,2,3 for X,Y,Z)              |
+---------+--------+----------+--------+-------------------------------+
|18       |69-72   |Int32     |MAPR    |axis corresponding to rows     |
|         |        |          |        |(1,2,3 for X,Y,Z)              |
+---------+--------+----------+--------+-------------------------------+
|19       |73-76   |Int32     |MAPS    |axis corresponding to sections |
|         |        |          |        |(1,2,3 for X,Y,Z)              |
```
This means that the x axis (columns) is the data axis given by the `MAPC` parameter, and so on. Hence, this defines a mapping
```
(x, y, z) axes --> storage axes
```
or, in other words (using the notions above),
```
data_shape[i] = num_cells[data_axis_order[i]]
```
Hence, in order to transpose from `data_shape` to `num_cells`, we need the forward permutation. This is the case when reading. For writing, things are the other way around.

This is confusing...

kohr-h: Maybe the names are not so good. Change as follows?

- `data_shape` -> `data_storage_shape` since it's the shape as in actual storage
- `num_cells` -> `data_shape` since it's the shape of the data as we actually work with it

odl/tomo/data/mrc2014.py
kohr-h: It's according to the standard: If the mean value is smaller than both min and max, this signals "not defined".

kohr-h: Note also that min and max have switched ordering by default, which signals the same undefined thing.

odl/tomo/data/mrc2014.py
kohr-h: Makes it a 1d array, not 0d. This is the easiest way of saying "I want a 1d array of exactly 1 element".

odl/tomo/data/mrc2014.py
kohr-h: Fixed this now, but not very much has changed in the end. I had to get a better idea of the relation between `axis_order` and `(mapc, mapr, maps)`. I understand the two concept as follows:

- `axis_order[1] = 0` : the data axis 0 corresponds to storage axis 1 
- `mapr = 0` : rows map to axis 0, i.e. storage axis 1 corresponds to data axis 0

So they're the same. Perhaps one should call it `storage_axis_order` for clarity.

Then, there is a necessary asymmetry between reader and writer. Reader needs to swap from storage shape to data shape, writer the other way around. Now the `np.transpose` function needs as `axes` argument the permutation "which axis goes where", i.e. the result of an axis permutation *to* the desired result. Since `axis_order` maps *storage order to data order*, we can use it directly in the reader and its inverse for the writer.

kohr-h:
Fixed everything according to comments, except for the reshaping thing. My suggestion would be to not reshape at all for the basic reader but do it for the MRC stuff. It just gets too annoying with all the axis ordering and stuff.
Change that and merge?

adler-j:
Frankly don't feel like re-reviewing 2000 lines :) Your answers were nice, and no major changes were proposed so I say you are free to do as you wish, if you feel done go ahead.

kohr-h:
Okay, fixed the `simple_fixture` thing and removed the reshaping in the raw reader. Will merge after CI.

odl/tomo/data/mrc2014.py
kohr-h: Overlooked this -- it does give warnings about unused stuff, but saves 20-ish lines of code of the annoying type
```python
header['bla'] = {'offset': header_fields['bla']['offset'], 'value': bla}
```
I could have done it with an explicit dictionary `{'bla': bla}` for all fields. That's fine with me if this looks too "dirty".

odl/tomo/data/mrc2014.py
kohr-h: Well, the writer can be handy if you want to store recos in MRC format to process them with some other EM tool. But recos don't need an extended header, and I've never written any tilt series, so I figured the extended header would be close to useless in the writer. If somebody wants to have it, I can add it, no big deal. But not now :-)

PR #707, opened 2016-11-08
Issue 657, Add IndicatorZero remove ConstantFunctionalConvexConj

adler-j:

PR #722, opened 2016-11-18
Revert "Merge branch 'test_deformations'"

adler-j:
This reverts commit ff4b13593a1650c339868d6734db9f7ce1d9df17, reversing changes made to f762820760d22587ef2ee8678b36bd9ad35f84b0.

Master was kinda butchered, but I cannot find the PR that was accepted for #454 only #518. This should clean up master anyway. We also need to make new pull requests for the other branches.

Could you @kohr-h check if you find any leftovers of these pulls on master?

P.S. Should we lock pull requests to me/@kohr-h?


adler-j:
So, I've made two changes to the ODL repo, effective immediately.

* Me and @kohr-h are now the only ones who can push straight to master without making a pull request.
* Pull requests can only be merged into master after they have been reviewed (with an accepted status) before they can be merged.

kohr-h:
I had a similar thought once, but then I figured "nah, we're getting too strict". Seems that it's a decent policy anyway. Chances are too high that stuff gets pushed by mistake to the wrong branch (e.g. master), and then everything is just messed up and life is bad.
Anyway, I'll check if the branch is clean.

But why didn't "revert merge" do the deal in the first place?

adler-j:
> But why didn't "revert merge" do the deal in the first place?

It didnt work. It seems the merge was a mix of two other merges, so simple revert was not possible.

adler-j:
> I had a similar thought once, but then I figured "nah, we're getting too strict". Seems that it's a decent policy anyway. Chances are too high that stuff gets pushed by mistake to the wrong branch (e.g. master), and then everything is just messed up and life is bad.
Anyway, I'll check if the branch is clean.

While I agree with you, I also feel that the extra overhead of others having to make a simple pull request is quite minimal, and now that we're getting more users having a f-ed up master is actually a bad thing. Had this happened last week when I was in china, who knows when it was going to get fixed.

adler-j:
By request, I added @ozanoktem to the odl-core group.

kohr-h:
So the code is clean on the reverted branch, but this will leave behind a totally messed up commit history. I think we should rather take the last "proper" commit, which was f762820760d22587ef2ee8678b36bd9ad35f84b0 (right after merging #707), and force-push it to master.

adler-j:
I'll do that.

adler-j:
Done.

PR #728, opened 2016-11-22
BUG: Fix failing unompr_bin tests on windows

adler-j:


adler-j:
Now working on both windows and linux it seems, should be a simple merge, basically changed two lines.

Issue #732, opened 2016-11-24
Bug: no update with conjugate gradient, os-dependency? 


davlars:
I'm running odl on the following system: 

Ubuntu 16.04.1 LTS
`odl.__version__ : 0.5.2`
`numpy.__version__: 1.11.2`
`python.__version__: 2.7.12`

With the following example, trying to reconstruct with conjugate gradient on normal equations (`odl.solvers.conjugate_gradient_normal`) gives no visual update. Instead the reconstruction is stuck on what seems to be the first iterate. This is however not the case when trying with e.g. landweber (wihch seems to work fine).

To make it even more interesting, this is not reproducible on a windows-system. 

Could someone try to re-run it on a linux-os to see if the error is local? 

```python
# Test run
import odl
import numpy as np

# Set geometry parameters
volumeSize = np.array([10, 10, 10]) #mm?
volumeOrigin = np.array([-5, -5, -5.0]) #mm?
nVoxels = np.array([100,100,100])
detectorSize = np.array([30.0, 3.0])
nPixels = [300, 30]
detectorOrigin = -detectorSize/2

# Discrete reconstruction space
discr_reco_space = odl.uniform_discr(volumeOrigin,volumeOrigin + volumeSize,
                                     nVoxels, dtype='float32')

#Define projection geometry
angle_partition = odl.uniform_partition(0, 2 * np.pi, 360)

# Detector: uniformly sampled, n = nPixels, min = detectorOrigin, max = detectorOrigin+detectorSize
detector_partition = odl.uniform_partition(detectorOrigin, detectorOrigin+detectorSize, nPixels)

# Spiral has a pitch of pitch_mm, we run n_turns rounds (due to max angle = 8 * 2 * pi)
geom = odl.tomo.HelicalConeFlatGeometry(angle_partition, 
                                        detector_partition, 
                                        src_radius=250, 
                                        det_radius=250,
                                        pitch=0,
                                        pitch_offset=0)

A = odl.tomo.RayTransform(discr_reco_space, geom, impl='astra_cuda')
# A = odl.trafos.FourierTransform(discr_reco_space) also fails
phantom = odl.phantom.cuboid(A.domain)
rhs = A(phantom)

# Reconstruct
partialPrintIter = odl.solvers.CallbackPrintIteration() #Print iterations
partialShowIter = odl.solvers.CallbackShow(coords=[None, 2, None]) #Show parital reconstructions

# THIS FAILS

[...]

PR #734, opened 2016-11-24
Issue 733  wavelet trafo

adler-j:
Work in progress. Fixes some tests but not all. Need input on @kohr-h since you were on the last time.

TODO:

- [ ] Fix adjoint of inverse FourierTransform being wrong.
- [ ] Fix corner cases of WaveletTransform.adjoint
- [ ] Add test for and verify correctness of FT.adjoint with odd sized domain. Currently fails.
- [ ] Add test of `FourierTransform.inverse.adjoint` and `FourierTransform.adjoint.inverse`

### DIFF
--- a/odl/test/trafos/fourier_test.py
+++ b/odl/test/trafos/fourier_test.py
@@ -34,38 +34,75 @@ from odl.trafos.fourier import (
     FourierTransform)
 from odl.util import (all_almost_equal, never_skip, skip_if_no_pyfftw,
                       noise_element,
-                      is_real_dtype, conj_exponent, complex_dtype)
-
+                      is_real_dtype, conj_exponent, complex_dtype,
+                      simple_fixture)
 # --- pytest fixtures --- #
 
-impl_params = [never_skip('numpy'), skip_if_no_pyfftw('pyfftw')]
-impl_ids = [' impl = {} '.format(impl.args[1]) for impl in impl_params]
+# Fixture for FFT implementations
+impl = simple_fixture('impl', [never_skip('numpy'),
+                               skip_if_no_pyfftw('pyfftw')])
+
+# Fixture for space exponents
+exponent = simple_fixture('p', [2.0, 1.0, float('inf'), 1.5])
+
+# Fixture for the sign of the Fourier transform
+sign = simple_fixture('sign', ['-', '+'])
 
 
-@pytest.fixture(scope="module", ids=impl_ids, params=impl_params)
-def impl(request):
-    """Fixture for FFT implementations."""
-    return request.param
+fouriertrafo_params = ['1d real-halfcomplex, odd n, sign+',
+                       '1d real-halfcomplex, odd n, sign-',
+                       '1d real-halfcomplex, even n',
+                       '1d real-complex, odd n',
+                       '1d real-complex, even n',
+                       '1d complex-complex odd n',
+                       '1d complex-complex even n',
+                       '2d real-halfcomplex',
+                       '2d real-complex',
+                       '2d complex-complex']
+fouriertrafo_ids = [" ft='{}' ".format(m)
+                    for m in fouriertrafo_params]
 
 
-exp_params = [2.0, 1.0, float('inf'), 1.5]
-exp_ids = [' p = {} '.format(p) for p in exp_params]
+@pytest.fixture(scope='module',
+                params=fouriertrafo_params, ids=fouriertrafo_ids)
+def fouriertrafo(request, impl):
+    name = request.param
 
-
-@pytest.fixture(scope="module", ids=exp_ids, params=exp_params)
-def exponent(request):
-    """Fixture for space exponents."""
-    return request.param
-
-
-sign_params = ['-', '+']
-sign_ids = [" sign='{}' ".format(p) for p in sign_params]
-
-
-@pytest.fixture(scope='module', ids=sign_ids, params=sign_params)
-def sign(request):
-    """Fixture for the sign of the Fourier transform."""
-    return request.param
+    if name == '1d real-halfcomplex, odd n, sign+':
+        discr = odl.uniform_discr(-2, 2, 11, impl='numpy', dtype='float32')
+        return FourierTransform(discr, sign='+', impl=impl, halfcomplex=True)
+    elif name == '1d real-halfcomplex, odd n, sign-':
+        discr = odl.uniform_discr(-2, 2, 13, impl='numpy', dtype='float64')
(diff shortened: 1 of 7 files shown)


odl/test/trafos/fourier_test.py
```diff
@@ -631,6 +631,53 @@ def char_rect(x):
             assert all_almost_equal(ft.adjoint(ft(char_rect)), discr_rect)
 
 
+def _test_adjoint(operator):
+    """Test adjoint of operator satisfies the definition."""
+    x = noise_element(operator.domain)
+    y = noise_element(operator.range)
```
kohr-h: These are expected to fail. The adjoint is set equal to the inverse, not (necessarily) the adjoint of the discretized operator. So for functions that fill the entire space, FT is expected to be produce errors. The inverse tests would probably also fail for these cases.
But perhaps the choice of setting adjoint = inverse needs a revisit.

adler-j: Well the tests are wildly off, not like 5% or so. With that said, I changed the phantom to be a cuboid, which should be nicer.

adler-j: I seem to be getting a factor 2 across the board, which seems to be the same as we saw in the bug report #732

adler-j: Example:

```python
>           assert AtAxy.real == pytest.approx(AxAy.real, rel=1e-2, abs=1e-4)
E           assert 1.9999999999999998 == 0.989876379296 ± 9.9e-03
```

adler-j:
Tests are now in place and I am "happy" with them, so to say. Would be great if you could throw an eye on how to fix this @kohr-h.

odl/test/trafos/fourier_test.py
kohr-h: > I seem to be getting a factor 2 across the board, which seems to be the same as we saw in the bug report #732

That could come from the roughly halved range in the real-to-complex case. Maybe that case needs a scaling? It's not exactly 2, it's more like `n / (n // 2 + 1)` probably.

adler-j: `n / (n // 2 + 1)` sounds very feasible, I saw a change when changing the level of discretization points.

adler-j:
Added  a new commit which makes all the fourier tests run happily, yet we lack a test of `FourierTransform.inverse.adjoint`, which still has the wrong scaling, not to mention `FourierTransform.adjoint.inverse`. We need to fix that.

adler-j:
Also, we have a huge problem with halfcomplex transforms and odd sized domains, which I cannot explain.

odl/trafos/fourier.py
```diff
@@ -1365,13 +1366,25 @@ def _call_pyfftw(self, x, out, **kwargs):
         return out
 
     @property
+    def adjoint(self):
+        """The adjoint Fourier transform."""
+        if self.domain.exponent != 2.0 or self.range.exponent != 2.0:
+            raise NotImplementedError(
+                'Adjoint only implemented for exponent=2')
```
kohr-h: At some point I was considering scrapping these checks. Why should we perform them? In a generalized sense the adjoint relation is still correct even if `p != 2`, so why forbidding it? Just a thought (again).

ozanoktem: L^p is a Hilbert space only for p=2, so the notion of adjoint when 1 < p and p !=2 would have to be based on the Banach space adjoint. I have no idea what happens with these topological issues when considering a discretization, i.e. what happens with the Banach space adjoint in a discretization.

kohr-h: My intuition (somewhat informed since I worked on that stuff in my PhD) is that you get the exact same expressions for the L^p case (1 <= p < infty) since the replacement of the inner product has the same form (integral), and the defining equation for the adjoint is the same.
The only notable difference is that the domain of the operator would be smaller, making it a closed operator only, no longer bounded (for p > 2), and the equality `<F f, g>_(L^q x L^p) = <f, F^* g>_(L^p x L^q)` would only hold for `f, g` such that `f in domain(F)` and `g in domain(F^*)`.

Anyway, regarding warnings I don't know. They are extremely annoying and interfere with screen output a lot. If at all, we would need a method to switch them off.

adler-j:
Further improvement that should be done here (should I do it?) is to change the implementation of the Fourier tests so that we have a fixture that returns `FourierTransform`s that can be used in other tests, i.e. in the tests of the inverse and of the adjoint.

This fixture should also include at least one domain with an odd number of points.

odl/trafos/fourier.py
adler-j: My suggestion is to leave this as is and perhaps make a TODO and an issue with this exact discussion quoted. In fact, i just did this: #741.

kohr-h:
> Further improvement that should be done here (should I do it?) is to change the implementation of the Fourier tests so that we have a fixture that returns FourierTransforms that can be used in other tests, i.e. in the tests of the inverse and of the adjoint.
> 
> This fixture should also include at least one domain with an odd number of points.

Good suggestion. I think it makes sense then that we don't work on the same branch simultaneously but push to our respective forks and rebase when necessary.

adler-j:
Pushed the fixtures which adds a few tests and not very surprisingly also causes some failures. This increases the number of tests however, so we may want to move this to the largescale tests (once we get the tests to not fail anymore).

adler-j:
Are you doing any work on your branch with this @kohr-h? Otherwise I would honestly be fine with a partial submission that fixes some of the worst offenders (i.e. at least solves the fourier transform).

kohr-h:
I haven't come around to this lately. The whole thing became a larger refactor on the FT side which is partly done. Forward works but not inverse or adjoint, needs some more massage. I'd say put in partial fixes and perhaps some xfails in the tests where we know things are broken, and we fix those later.

adler-j:
Is this officially on your table now @kohr-h, in that case, do we want to move this to a private branch of yours?

kohr-h:
I'd say no, because I already have [a branch](https://github.com/kohr-h/odl/tree/restructure_ft) on my fork that is based on this and deals with the larger issue of the FT re-organization. This is basically your initial work only. However, I won't have time to look into this in the *near* future. Couple of weeks, maybe.

adler-j:
So we can close this?

kohr-h:
I don't know. Wasn't there some talking about merging the wavelet-specific stuff? I have I just forgotten that this has already been done?
If that's the case then I'd say yes, close this and remove the branch.

adler-j:
The wavelet specific stuff got a quick fix at some point. Closing.

PR #738, opened 2016-11-25
WIP: CPU tensors

kohr-h:
Work-in-progress PR for a larger change to the underlying data containers.

TODOs:
- [x] Implement basic structures for tensors and tensor spaces
- [x] Replicate functionality of `NtuplesBase`, `FnBase`, `NumpyNtuples` and `NumpyFn` in the corresponding tensor classes
- [x] Some doctests
- [ ] More complete doctests
- [ ] Unit tests
- [ ] Extra documentation
- [ ] Cleanup (old stuff, commit history, ...)
- [x] Remove `Ntuples` and related stuff
- [x] Add former `Ntuples` and the likes as special case to tensor classes
- [ ] Check documentation for outdated information, e.g. regarding 1D data layout
- [ ] Port `show` methods (1D from `Ntuples`, 2D from `DiscreteLp`)
- [ ] Find place for `MatVecOperator`
- [ ] Use "rank" instead of "number of dimensions" to better distinguish between `space.ndim` and `space.size`

Edit: ~~There is some overlap of this branch with #739. I'll make a separate branch that only contains the base classes and rebase them both on it (and keep doing that later on).~~

Edit 2: It became too annoying. Instead I'll use this as base branch and rebase #739 on this one since they are basically orthogonal apart from the `base_tensors` stuff, which they should share.

PR #739, opened 2016-11-25
WIP: GPU tensors

kohr-h:
Second WIP PR. ~~Much earlier stage than #738 but maybe some comments already applicable.~~

TODOs:
- [x] Make the basic structures (space, element, weighting) work
- [x] Implement ufuncs and reductions
- [x] Write optimized kernels for inner, dist and norm
- [x] Fence imports and dependencies on `pygpu`
- [ ] Adapt the unit tests (complex dtypes etc.) and make them work for GPU arrays
- [x] Do some benchmarking to find out thresholds for Numpy vs GPU
- [x] Check if reductions in all axes run faster on flattened arrays. ~~They should.~~ No, because `reshape` of `GpuArray` objects copy the array, in contrast to Numpy.
- [ ] Update documentation
- [ ] Large-scale tests
- [ ] What to do with the doctests? They fail if `pygpu` is not installed. Should the package be mandatory?
     Update: manual doctests are fixed by adding a conditional in the `__main__` section, need to fix runs by `pytest` also.
- [x] Expose function to precompile kernels for linear space methods? Also add more of them.
- [x] Keep `gpuary_tensors.py` namespace clean by making context initialization code a function only returning the relevant stuff.


odl/space/gpuary_ntuples.py
```diff
@@ -0,0 +1,2302 @@
...
+                lico(1, x1.data, b, x2.data, out.data)
+            else:
+                if b == 1:  # No scaling in x2 =>  out <-- a * x1 + x2
+                    # TODO: make optimized version or change axpy
+                    lico(a, x1.data, 1, x2.data, out.data)
+                else:  # General case  out <-- a * x1 + b * x2
+                    lico(a, x1.data, b, x2.data, out.data)
+
+
+class GpuArrayTensorSpace(FnBase, GpuArrayRawTensorSpace):
```
kohr-h: From here on, the rest is currently copy-paste from `npy_ntuples.py`. Ignore it.

adler-j:
Great news:

https://github.com/Theano/Theano/issues/4705#issuecomment-272188594

> Just a note, we are doing conda packages to help install pygpu. This will
make this much easier to install on windows.

kohr-h:
[I know](https://github.com/Theano/libgpuarray/pull/311) - but good to hear that they are trying to make it work for Windows, too.

kohr-h:
So this one also approaches completion. All doctests go green on my machine (although they are not that exhaustive).

I add a TODO list at the first post.

adler-j:
Status on this, when will it be up to date?

kohr-h:
Not quite reviewable right now, mainly since not all of the docs are up to date yet. Unit tests are the other big TODO, after that a review will be possible.

kohr-h:
Some timing of the linear space methods, just for reference.

All tests were run with flat arrays of `N = 10 ** k` entries, with `k` as in the table. Times are in milliseconds.

## `scal`

k | GPU | CPU | BLAS  | GPU to CPU
--|--------|-------|----------|----------------
1 | 0.016|7.2e-7|1.6e-7|0.027
2 | 0.016|7.6e-7|1.8e-7|0.027
3 | 0.016|9.6e-7|2.4e-7|0.028
4 | 0.014|2.2e-4|9.5e-7|0.038
5 | 0.015|0.014|0.008  |0.16
6 | 0.042|0.12  |0.082  |1.2
7 | 0.41  | 2.5   |2.4      |19
8 | 4.3    |27     |27       |194

GPU: custom kernel
CPU: `np.multiply(a, x, out=x)`
BLAS: `scal(a, x)`
GPU to CPU:
```python
x_host = a * np.asarray(x_dev)
x_dev[:] = x_host
```

## `lico`

k | GPU | CPU |GPU to CPU
--|--------|---------|--------------
1 | 0.022|0.0018|0.036
2 | 0.022|0.0018|0.036
3 | 0.022|0.0026|0.039
4 | 0.022|0.0068|0.058
5 | 0.022|0.049  |0.26
6 | 0.062|0.90    |2.3
7 | 0.60  | 16      |54
8 | 6.1    |155     |530

GPU: custom kernel
CPU:
```python
np.multiply(a, x, out=out)
out += b * y
```
GPU to CPU:
```python
out_host = a * np.asarray(x_dev)
out_host += b * np.asarray(y_dev)
out_dev[:] = out_host
```

## Some conclusions

- It is never faster to copy from GPU to CPU and run the loop there. Hence, no need to distinguish between small and large arrays.
- The constant overhead for copies from GPU to CPU lies in the order of 14 microseconds.
- The constant overhead of custom kernel calls is in the order of 20 microseconds.
- BLAS is always faster than the Numpy ufunc, regardless of size (call overhead is negligible).
- The GPU speedup is small for simple stuff like scaling (max 6x), and medium for slightly more complex things like linear combinations (max 25x).

PR #745, opened 2016-11-30
DOC: Improve doc of KL functionals, see #743

adler-j:
Improved the math documentation of the KL functionals to be more readable to the "layman".

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -993,12 +993,18 @@ class KullbackLeibler(Functional):
     diescribed in `this Wikipedia article
     <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_, and
     the functional :math:`F` is obtained by switching place of the prior and
-    the varialbe in the KL cross entropy functional. See the See Also section.
+    the varialbe in the KL cross entropy functional.
+
+    For a theoretical exposition, see `Csiszar1991`_.
 
     See Also
     --------
     KullbackLeiblerConvexConj : the convex conjugate functional
     KullbackLeiblerCrossEntropy : related functional
+
+    References
+    ----------
+    .. _Csiszar1991:  http://www.jstor.org/stable/2241918
     """
 
     def __init__(self, space, prior=None):
@@ -1223,15 +1229,15 @@ class KullbackLeiblerCrossEntropy(Functional):
 
     Notes
     -----
-    The functional :math:`F` with prior :math:`g>0` is given by:
+    The functional :math:`F` with prior :math:`x>0` is given by:
 
     .. math::
-        F(x)
+        F(g)
         =
         \\begin{cases}
             \\sum_{i} \left( g_i - x_i + x_i \log \left( \\frac{x_i}{g_i}
             \\right) \\right)
-            & \\text{if } x_i > 0 \\forall i
+            & \\text{if } g_i > 0 \\forall i
             \\\\
             +\\infty & \\text{else}
         \\end{cases}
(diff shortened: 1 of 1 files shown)


odl/solvers/functional/default_functionals.py
```diff
@@ -1217,18 +1223,23 @@ class KullbackLeiblerCrossEntropy(Functional):
...
-    For further information about the functional, see for example `this article
+        F(x)
+        =
+        \\begin{cases}
+            \\sum_{i} \left( g_i - x_i + x_i \log \left( \\frac{x_i}{g_i}
+            \\right) \\right)
+            & \\text{if } x_i > 0 \\forall i
+            \\\\
+            +\\infty & \\text{else}
+        \\end{cases}
```
kohr-h: I'm confused. This definition is the same as for the usual KL divergence, which is correct, but shouldn't one of them be a function of `x` and the other one a function of `g`?

adler-j: I think this is a naming issue. Paging @aringh who wrote the original documentation. I simply ported the old doc to more readable mathematics (and added what I presumed was the free variable, given that we had "where g is the prior" before.)

kohr-h: I assume that this one should be `F(g)` instead for fixed `x`. Or you switch roles completely.

aringh: The definition is the same, up to the fact that we switch which of the two arguments is seen as prior (i.e., is fixed) and which is seen as variable (i.e., the one we optimize over).

If you see it as a functional of two arguments
```
F(x, y) = \\begin{cases}
             \\sum_{i} \left( y_i - x_i + x_i \log \left( \\frac{x_i}{y_i}
             \\right) \\right)
             & \\text{if } x_i > 0 \\forall i \text{and } y_i > 0 \\forall i
             \\\\
             +\\infty & \\text{else}
         \\end{cases}
```
the one of the functionals is given by fixing `x`, the other by fixing `y`.

adler-j: So then the conclusion is that this text is correct?

adler-j: Above comment is "clearly" wrong (my comment, that is). Anyway, this is then a function of `g` I suppose? What would we call `x` in that case?

adler-j:
Updated according to comments and added reference. Ok now? Need a review from @aringh I think.

odl/solvers/functional/default_functionals.py
aringh: No, I think what you had was correct. The naming in the two functionals where different: `x` is always the variable and `g` is always the prior. As it says now, when `g` is the variable in ` KullbackLeiblerCrossEntropy`, the two functionals are the same.

odl/solvers/functional/default_functionals.py
```diff
@@ -1217,30 +1229,41 @@ class KullbackLeiblerCrossEntropy(Functional):
...
-        \\sum_{i} \left( g_i - x_i + x_i \log \left( \\frac{x_i}{ pos(g_i) }
-        \\right) \\right) + I_{x \\geq 0}(x)
-
-    where :math:`g` is the prior, and :math:`I_{x \\geq 0}(x)` is the indicator
-    function on nonnegative elements.
-
-    `Wikipedia article on Kullback Leibler divergence
-    <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_.
-    For further information about the functional, see for example `this article
+        F(g)
```
aringh: Just change to a function of `x`: `F(x)`.

odl/solvers/functional/default_functionals.py
```diff
@@ -1356,12 +1379,11 @@ class KullbackLeiblerCrossEntropyConvexConj(Functional):
 
     Notes
     -----
-    The functional is given by
+    The functional :math:`F^*` with prior :math:`x>0` is given by
 
     .. math::
-        \\sum_i g_i (exp(x_i) - 1)
+        F^*(g) = \\sum_i g_i \\left(e^{x_i} - 1\\right)
```
aringh: Just change to a function of `x`: `F(x)`.

adler-j:
> I also saw when looking at the code that the convex conjugate functionals are not exposed in all at the top. Do we want them to be that?

Leaving them as is for the moment, the interested user can fish them out. I have yet to find a use-case for them.

adler-j:
Merge after CI

Issue #737, opened 2016-11-25
Functionals for approximate data matching


aringh:
For exact data matching the convex conjugate of the Zero Functional is actually of interest (see issue #657). Normal ways of doing data matching is also expressions of the form
```
||Ax - b||_2^2 <= epsilon.
```
The convex conjugate of this functional, being defined as a infinity outside "the tolerance region" and zero inside, can be derived and also the proximal operator of this.

My suggestions is to implement this as a separate functional, since this opens up for using this kind of expressions in the Douglas-Rachford and primal-dual forward-backward solver.

adler-j:
Example of how to do this via the convjex conjugate of l2 or using IndicatorLpUnitBall:

```python
import odl
import numpy as np

r2 = odl.rn(2)
A = odl.IdentityOperator(r2)
b = r2.one()

epsilon = 0.5

func = ((1 / np.sqrt(epsilon)) * odl.solvers.L2Norm(A.domain) ).convex_conj.translated(b)
func2 = odl.solvers.IndicatorLpUnitBall(r2, 2).translated(b) * (1 / np.sqrt(epsilon))  # note right multiplication
```

PR #746, opened 2016-11-30
Douglas rachford tomography example

adler-j:
Added an example of tomography with the douglas rachford solver using TV minimization but with a strict data discrepancy term.

I quite like this example since it shows some capabilities of ODL that you probably would not expect at first glance.


adler-j:
It replicates this paper rather strictly:

http://statweb.stanford.edu/~candes/papers/ExactRecovery.pdf

Using the same numbers etc. @aringh and @hilding79 may be interested given their interest in issue #737.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+import odl
+
+# Parameters
+lam = 0.01
+
+# --- Create spaces, forward operator and simulated data ---
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 128 samples per dimension.
```
adler-j: Change 128 -> 512

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+
+# --- Create functionals for solving the optimization problem ---
+
+# Gradient for TV regularization
+gradient = odl.Gradient(space)
+
+# Functional to enforce 0 <= x <= 1
+f = odl.solvers.IndicatorBox(space, 0, 1)
+
+# Functional for I_{Ax = g}
```
adler-j: Technically `I_{. = g}` here, should probably be discussed.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+where I_{.} is the indicator function, which is zero if ``Ax = g`` and infinity
+otherwise. This is a standard convex optimization problem that can
+be solved with the `douglas_rachford_pd` solver.
+
+In this example, the angles are highly under-sampled and the problem is solved
+using only 6 angles. Dispite this, we get a perfect reconstruction.
+A filtered backprojection (pseudoinverse) reconstruction is also shown at the
+end for comparsion.
+
+Note that the ``Ax = g`` condition could be relaxed to ``||Ax - g|| < eps`` for
```
chongchenmath: ``||Ax - g|| <= eps``?

adler-j: I suppose that is more correct.

chongchenmath: If not ``<=``, the optimization might have no solution for the differentiation of ``inf`` and ``min``.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Total variation sparse angle tomography using the Douglas-Rachford solver.
+
+Solves the optimization problem
+
+    min_{0 <= x <= 1, Ax = g} lam ||grad(x)||_x
+
+where ``A`` is ray transform operator, ``g`` the given noisy data, ``eps`` is
```
kohr-h: There is no `eps` in the above problem definition.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+a small number, ``grad`` is the spatial gradient and || . ||_x is the so called
+cross-norm, giving rise to isotropic Total Variation.
+
+We do this by rewriting the problem on the form
```
kohr-h: Do what? The beginning speaks of "solves", which I read as "this example solves", not as "we solves".
You may want to write "This problem can be rewritten in the form" here.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+
+    min_{0 <= x <= 1} lam ||grad(x)||_x + I_{Ax = g}
+
+where I_{.} is the indicator function, which is zero if ``Ax = g`` and infinity
+otherwise. This is a standard convex optimization problem that can
+be solved with the `douglas_rachford_pd` solver.
+
+In this example, the angles are highly under-sampled and the problem is solved
+using only 6 angles. Dispite this, we get a perfect reconstruction.
```
kohr-h: Despite

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+
+We do this by rewriting the problem on the form
+
+    min_{0 <= x <= 1} lam ||grad(x)||_x + I_{Ax = g}
+
+where I_{.} is the indicator function, which is zero if ``Ax = g`` and infinity
+otherwise. This is a standard convex optimization problem that can
+be solved with the `douglas_rachford_pd` solver.
+
+In this example, the angles are highly under-sampled and the problem is solved
```
kohr-h: Perhaps the other way around:
"In this example, the problem is solved with only 6 angles available, which is highly under-sampled data."

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+using only 6 angles. Dispite this, we get a perfect reconstruction.
+A filtered backprojection (pseudoinverse) reconstruction is also shown at the
+end for comparsion.
+
+Note that the ``Ax = g`` condition could be relaxed to ``||Ax - g|| < eps`` for
+some small eps in order to account for noise. This would be done using the
```
kohr-h: \`\`eps\`\`

examples/solvers/douglas_rachford_pd_tomography_tv.py
kohr-h: `I_{A(.) = g}` but I'm not so fond of the dot notation in general, it sometimes makes things quite unreadable. We shouldn't overdo it.
Optionally, write this in words, like "Functional enforcing Ax = g"

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+
+# --- Create spaces, forward operator and simulated data ---
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 128 samples per dimension.
+space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20], shape=[512, 512])
+
+# Make a parallel beam geometry with flat detector
+# Angles: uniformly spaced, n = 22, min = 0, max = pi
+angle_partition = odl.uniform_partition(0, np.pi, 22)
```
kohr-h: This doesn't look like 6 angles to me.

examples/solvers/douglas_rachford_pd_tomography_tv.py
kohr-h: A comment showing how to switch this for the indicator of the unit ball as mentioned above would be nice.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+# Functional for TV minimization
+cross_norm = lam * odl.solvers.GroupL1Norm(gradient.range)
+
+# --- Create functionals for solving the optimization problem ---
+
+# Assemble operators and functionals for the solver
+lin_ops = [ray_trafo, gradient]
+g = [indicator_data, cross_norm]
+
+# Create callback that prints iteration number and shows partial results
```
kohr-h: the iteration number

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -0,0 +1,114 @@
...
+callback = (odl.solvers.CallbackShow('iterates',
+                                     display_step=20, clim=[0, 1]) &
+            odl.solvers.CallbackPrintIteration())
+
+# Solve with initial guess x = 0.
+# Step size parameters are selected to ensure convergece.
```
kohr-h: convergece

examples/solvers/douglas_rachford_pd_tomography_tv.py
adler-j: It's 22 in the paper, so I increased from 6 to 22 (and also increased size of domain to compensate, thats why I had the 128-512 error).

examples/solvers/douglas_rachford_pd_tomography_tv.py
adler-j: I'll add that.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -84,7 +83,7 @@
 # Functional to enforce 0 <= x <= 1
 f = odl.solvers.IndicatorBox(space, 0, 1)
 
-# Functional for I_{Ax = g}
+# Functional to enforce Ax = g
```
kohr-h: Now that I look at it again, it may also confusing for a user to see `Ax = g` here but no `A` in the functional. So it would be good to add a comment that due to splitting, the operator A is not included here.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -83,12 +84,22 @@
...
-indicator_data = odl.solvers.IndicatorZero(ray_trafo.range).translated(data)
+if data_matching == 'exact':
+    # Functional to enforce Ax = g
+    indicator_zero = odl.solvers.IndicatorZero(ray_trafo.range)
+    indicator_data = indicator_zero.translated(data)
+elif data_matching == 'inexact':
+    # Functional to enforce ||Ax - g||_2 < eps
+    eps = 1e-4
+    indicator_l2_ball = odl.solvers.IndicatorLpUnitBall(ray_trafo.range, 2)
+    indicator_data = indicator_l2_ball.translated(data / eps) * (1 / eps)
```
kohr-h: Nice, but perhaps make it a bit digestible for users? :-)

adler-j: Added more information text.

examples/solvers/douglas_rachford_pd_tomography_tv.py
```diff
@@ -86,20 +86,29 @@
...
     indicator_data = indicator_zero.translated(data)
 elif data_matching == 'inexact':
     # Functional to enforce ||Ax - g||_2 < eps
+    # We do this by rewriting the condition on the form
+    # f(x) = 0 if ||A(x/eps) - (g/eps)||_2 < 1, infinity otherwise
+    # That function (with A handled separately, as mentioned above) is
+    # implemented in ODL as the IndicatorLpUnitBall function.
+    # Note that we use right multiplication in order to scale in input argument
+    # instead of the result of the functional, as would be the case with left
+    # multiplication.
```
kohr-h: Nice!

adler-j:
Merge after PR.

This was quite a educative example for me to write, didn't know ODL could do this type of stuff but it goes like a charm.

PR #749, opened 2016-11-30
BUG: fix scaling for current ASTRA version

### LINKED ISSUE #724: Scaling in ASTRA 3d geometries
In recent changes in the ASTRA code, scalings in the 3d cases were changed. I suspect [this pull request](https://github.com/astra-toolbox/astra-toolbox/pull/41). We need to test this explicitly, but I guess the changes fix the scaling inconsistencies that were there before and that we [fix on our side](https://github.com/odlgroup/odl/blob/master/odl/tomo/backends/astra_cuda.py#L235-L262). As it seems, there is nothing to fix any more, and I see failures in tests with current ASTRA versions due to this.

So how are we going to fix this issue? We cannot check (yet) "astra version before this: fix scaling else: don't" because the fixed scaling code is on master and not in a released version. Also, master internally works with the same version number as the last release so we cannot use that either.

kohr-h:
Fixes #724.

I left the testing code in the examples so you can test the changes. When done, I'll rebase and remove it again.
Test the code with the latest ASTRA git version or take the newest conda package:
```
conda install -c wjpalenstijn astra-toolbox
```
There is still an issue with all cone beam geometries, namely that the density of lines is not taken into account in the backprojection. Basically the backprojectors interpolate the value in a voxel from the two closest lines to the center (ish?), but that does not differentiate between regions with low and high line density. That also explains the large tolerance required especially for the `cone3d` tests because there the line density is least homogeneuous. For helical the situation is better and the error smaller.

This issue is known and on the agenda here :-).

### DIFF
--- a/odl/test/tomo/operators/ray_trafo_test.py
+++ b/odl/test/tomo/operators/ray_trafo_test.py
@@ -22,13 +22,13 @@ from __future__ import print_function, division, absolute_import
 from future import standard_library
 standard_library.install_aliases()
 
-# External
-import pytest
 import numpy as np
+from pkg_resources import parse_version
+import pytest
 
-# Internal
 import odl
 import odl.tomo as tomo
+from odl.tomo.backends import ASTRA_VERSION
 from odl.tomo.util.testutils import (skip_if_no_astra, skip_if_no_astra_cuda,
                                      skip_if_no_scikit)
 from odl.util.testutils import almost_equal, all_almost_equal
@@ -190,10 +190,13 @@ def test_projector(projector, in_place):
 
 def test_adjoint(projector):
     """Test discrete Ray transform backward projection."""
-
-    # TODO: this needs to be improved
-    # Accept 10% errors
-    places = 1
+    # Relative tolerance, still rather high due to imperfectly matched
+    # adjoint in the cone beam case
+    if (parse_version(ASTRA_VERSION) < parse_version('1.8rc1') and
+            isinstance(projector.geometry, odl.tomo.HelicalConeFlatGeometry)):
+        rtol = 0.1
+    else:
+        rtol = 0.05
 
     # Create Shepp-Logan phantom
     vol = odl.phantom.shepp_logan(projector.domain, modified=True)
@@ -205,7 +208,7 @@ def test_adjoint(projector):
     # Verified the identity <Ax, Ax> = <A^* A x, x>
     result_AxAx = proj.inner(proj)
     result_xAtAx = backproj.inner(vol)
-    assert almost_equal(result_AxAx, result_xAtAx, places=places)
+    assert result_AxAx == pytest.approx(result_xAtAx, rel=rtol)
 
 
 def test_adjoint_of_adjoint(projector):
--- a/odl/tomo/backends/astra_cuda.py
+++ b/odl/tomo/backends/astra_cuda.py
@@ -23,6 +23,7 @@ from future import standard_library
 standard_library.install_aliases()
 
 import numpy as np
+from pkg_resources import parse_version
 try:
     import astra
     ASTRA_CUDA_AVAILABLE = astra.astra.use_cuda()
@@ -31,6 +32,7 @@ except ImportError:
 
 from odl.discr import DiscreteLp, DiscreteLpElement
 from odl.tomo.backends.astra_setup import (
+    ASTRA_VERSION,
     astra_projection_geometry, astra_volume_geometry, astra_projector,
     astra_data, astra_algorithm)
 from odl.tomo.geometry import (
@@ -44,7 +46,6 @@ __all__ = ('astra_cuda_forward_projector', 'astra_cuda_back_projector',
 
 
 # TODO: use context manager when creating data structures
-# TODO: is magnification scaling at the right place?
 
 def astra_cuda_forward_projector(vol_data, geometry, proj_space, out=None):
     """Run an ASTRA forward projection on the given data using the GPU.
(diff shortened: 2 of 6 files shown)


odl/tomo/backends/astra_cuda.py
```diff
@@ -228,37 +229,72 @@ def astra_cuda_back_projector(proj_data, geometry, reco_space, out=None):
...
+            scaling_factor *= float(reco_space.cell_volume)
+            # Magnification correction
+            src_radius = geometry.src_radius
+            det_radius = geometry.det_radius
+            scaling_factor *= ((src_radius + det_radius) / src_radius)
+        elif isinstance(geometry, Parallel3dAxisGeometry):
+            # Scales with cell volume
+            # currently only square voxels are supported
+            scaling_factor /= reco_space.cell_volume
+        elif isinstance(geometry, CircularConeFlatGeometry):
```
adler-j: `CircularConeFlatGeometry` is a special case of `HelicalConeFlatGeometry`, why would be need two versions of the code?

The only difference is that `CircularConeFlatGeometry` has `pitch=0`.

odl/tomo/backends/astra_cuda.py
```diff
@@ -228,37 +229,72 @@ def astra_cuda_back_projector(proj_data, geometry, reco_space, out=None):
 
     # Angular integration weighting factor
     # angle interval weight by approximate cell volume
-    extent = float(geometry.motion_partition.extent())
-    size = float(geometry.motion_partition.size)
-    scaling_factor = extent / size
+    angle_extent = float(geometry.motion_partition.extent())
+    num_angles = float(geometry.motion_partition.size)
+    scaling_factor = angle_extent / num_angles
```
adler-j: This should likely be factored out to a free function. It is becoming to big.

odl/tomo/backends/astra_cuda.py
```diff
@@ -228,37 +229,72 @@ def astra_cuda_back_projector(proj_data, geometry, reco_space, out=None):
...
+
+    astra_ver_num = astra.astra.version()
```
adler-j: Perhaps move this stuff to when we import astra up top?

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -191,9 +191,9 @@ def test_projector(projector, in_place):
 def test_adjoint(projector):
     """Test discrete Ray transform backward projection."""
 
-    # TODO: this needs to be improved
-    # Accept 10% errors
-    places = 1
+    if isinstance(projector.geometry, odl.tomo.CircularConeFlatGeometry):
```
adler-j: See comment further down, `CircularConeFlatGeometry` is a special case of helical and should never be treated as a separate class in ODL. If you want ot handle it, check if it is a helical with pitch 0.

odl/tomo/backends/astra_cuda.py
```diff
@@ -228,37 +229,72 @@ def astra_cuda_back_projector(proj_data, geometry, reco_space, out=None):
...
+            # Scales with 1 / cell_volume
+            # In 1.7, only cubic voxels are supported
+            voxel_stride = reco_space.cell_sides[0]
+            scaling_factor /= float(voxel_stride)
+            # Magnification correction
+            src_radius = geometry.src_radius
+            det_radius = geometry.det_radius
+            scaling_factor *= ((src_radius + det_radius) / src_radius) ** 2
+    else:
+        # TODO: check 2D
```
kohr-h: This is done.

odl/tomo/backends/astra_cuda.py
kohr-h: You mean the whole thing or just these 3 lines? The whole thing would make sense.

kohr-h:
Did all the requested changes, plus a conditional relative tolerance in the unit test. I can test it with an old ASTRA version, with the newest it works.

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -190,10 +190,13 @@ def test_projector(projector, in_place):
...
     """Test discrete Ray transform backward projection."""
-
-    # TODO: this needs to be improved
-    # Accept 10% errors
-    places = 1
+    # Relative tolerance, still rather high due to imperfectly matched
+    # adjoint in the cone beam case
+    if (parse_version(ASTRA_VERSION) < parse_version('1.8rc1') and
+            isinstance(projector.geometry, odl.tomo.HelicalConeFlatGeometry)):
+        rtol = 1e-1
```
adler-j: Write this out instead, 0.1 is much easier to read.

odl/tomo/backends/astra_cuda.py
```diff
@@ -273,6 +241,83 @@ def astra_cuda_back_projector(proj_data, geometry, reco_space, out=None):
     return out
 
 
+def astra_cuda_bp_scaling_factor(reco_space, geometry):
+    """Volume scaling accounting for differing adjoint definitions.
```
adler-j: Great that this is now commented better as well!

odl/tomo/backends/astra_cuda.py
```diff
@@ -273,6 +241,83 @@ def astra_cuda_back_projector(proj_data, geometry, reco_space, out=None):
...
+            # Scales with cell volume
+            scaling_factor /= reco_space.cell_volume
+            # Magnification correction
+            src_radius = geometry.src_radius
+            det_radius = geometry.det_radius
+            scaling_factor *= ((src_radius + det_radius) / src_radius) ** 2
+
+            # Correction for scaled 1/r^2 factor in ASTRA's density weighting
+            det_px_area = geometry.det_partition.cell_volume
+            scaling_factor *= (src_radius ** 2 * det_px_area ** 2 /
```
adler-j: I dont get it. You scale by 1/src_radius^2 above and src_radius^2 here. Why not simply remove it?

odl/tomo/backends/astra_setup.py
```diff
@@ -52,14 +54,26 @@
...
+try:
+    # Available from 1.8rc1
+    ASTRA_VERSION = astra.__version__
+except AttributeError:
+    astra_ver_num = astra.astra.version()
+    ASTRA_VERSION = '.'.join([str(astra_ver_num // 100),
+                              str(astra_ver_num % 100)])
+
+if parse_version(ASTRA_VERSION) < parse_version('1.7'):
+    raise RuntimeError('ASTRA version < 1.7 not supported')
```
adler-j: Do we prefer this over ASTRA_AVAILABLE = False?

adler-j:
Works with ASTRA 1.7

odl/tomo/backends/astra_cuda.py
kohr-h: I don't understand the scaling perfectly but I didn't want to mix the geometrical magnification thing with the 1/r^2 correction. If you think they're actually related I can just remove the canceling factors but for me it was clearer to keep the contributions separate. It's only a scalar calculation anyway.

kohr-h:
Just fixed a bug that made import fail without astra. Will merge after CI then (or perhaps remove the canceling factors as discussed above).

Issue #50, opened 2015-11-19
Common step length class


adler-j:
Copied from [odl-solvers issue 5](https://github.com/odlgroup/odl-solvers/issues/5)

I suggest that we write a common class for step length calculation which tells the algorithm what it needs (which part of the history), such that an algorithm can determine which iterates and gradients (what else?) to keep.

A simple approach would be to give the `StepLength` class attributes

- `iter_idcs` (iterates)
- `func_idcs` (function evaluations)
- `grad_idcs` (gradient evaluations / approximations)
- `hess_idcs` (Hessian evaluations / approximations)
- `step_idcs` (step lengths)

which all return lists of indices of items to keep. An algorithm could manage lists of all these objects and pass its lists (possibly kept in a dictionary) to the step length call, which in turn does its thing to calculate the step length.

adler-j:
We have already added a class that does some of this. Do we think that anything this advanced will be done, ever?

kohr-h:
I don't think there will be much activity here anytime soon. The current design of step length rules is limited, though.

PR #752, opened 2016-12-06
ENH: Add PredefinedLineSearch

adler-j:
Add option to have a line search where the step length depends on the iteration count.

### DIFF
--- a/odl/solvers/util/steplen.py
+++ b/odl/solvers/util/steplen.py
@@ -29,7 +29,7 @@ from odl.util.utility import with_metaclass
 
 
 __all__ = ('LineSearch', 'BacktrackingLineSearch', 'ConstantLineSearch',
-           'PredefinedLineSearch')
+           'LineSearchFromIterNum')
 
 
 class LineSearch(with_metaclass(ABCMeta, object)):
@@ -201,9 +201,12 @@ class ConstantLineSearch(LineSearch):
         return self.constant
 
 
-class PredefinedLineSearch(LineSearch):
+class LineSearchFromIterNum(LineSearch):
 
-    """Line search object that returns a step length from a function."""
+    """Line search object that returns a step length from a function.
+
+    The returned step length is ``func(iter_count)``.
+    """
 
     def __init__(self, func):
         """Initialize a new instance.
@@ -211,12 +214,12 @@ class PredefinedLineSearch(LineSearch):
         Parameters
         ----------
         func : callable
-            Function that when called with an iteration count. Iterations are
-            zero indexed, i.e. the first iteration is the zero:th.
+            Function that when called with an iteration count should return the
+            step length. The iteration count starts at 0.
 
         Examples
         --------
-        Make a step size that is 1.0 for the first 5 iterations, then 0.1
+        Make a step size that is 1.0 for the first 5 iterations, then 0.1:
 
         >>> def step_length(iter):
         ...     if iter < 5:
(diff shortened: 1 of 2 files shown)


odl/solvers/util/steplen.py
```diff
@@ -195,21 +196,48 @@ def __init__(self, constant):
...
         Parameters
         ----------
-        x : `LinearSpaceElement`
-            The current point
-        direction : `LinearSpaceElement`
-            Search direction in which the line search should be computed
-        dir_derivative : float
-            Directional derivative along the ``direction``
+        func : callable
+            Function that when called with an iteration count. Iterations are
```
kohr-h: Read again

odl/solvers/util/steplen.py
```diff
@@ -195,21 +196,48 @@ def __init__(self, constant):
...
+            zero indexed, i.e. the first iteration is the zero:th.
```
kohr-h: "start at 0" simply?
No Swedish enumeration please :-)

odl/solvers/util/steplen.py
```diff
@@ -195,21 +196,48 @@ def __init__(self, constant):
...
+
+        Examples
+        --------
+        Make a step size that is 1.0 for the first 5 iterations, then 0.1
```
kohr-h: :

odl/solvers/util/steplen.py
```diff
@@ -195,21 +196,48 @@ def __init__(self, constant):
...
+
+        >>> def step_length(iter):
+        ...     if iter < 5:
+        ...         return 1.0
+        ...     else:
+        ...         return 0.1
+        >>> line_search = PredefinedLineSearch(step_length)
+        """
+        if not callable(func):
+            raise TypeError('`func` must be a callable.')
```
kohr-h: The more I see these checks, the more I doubt their value, in particular in cases like this where the failure will come immediately anyway. The only gain here is that the user knows from the message which argument was not callable and that it's not an internal bug that causes the failure.

It's just a thought that popped into my mind, no need to change anything here. But we should keep in mind not to overdo it, otherwise we have redundant checks all over the place.

odl/solvers/util/steplen.py
```diff
@@ -195,21 +196,48 @@ def __init__(self, constant):
...
         """Calculate the step length at a point.
 
+        All arguments are ignored and are only added to fit the interface.
+        """
+        return self.constant
+
+
+class PredefinedLineSearch(LineSearch):
+
+    """Line search object that returns a step length from a function."""
```
kohr-h: Until I looked at the example, I didn't know what this class does. A simple additional sentence like "The returned step length is \`\`func(iter_count)\`\`." would enhance clarity a great deal.

adler-j:
The name `PredefinedLineSearch` is not very good btw. Any suggestions on it?

kohr-h:
> The name PredefinedLineSearch is not very good btw. Any suggestions on it?

Yeah I was trying to think of something better but couldn't quite come up with anything. A proper name would be `FunctionOfIterationNumberLineSearch` which is horribly long. How about `LineSearchFromIterNum`?

As a side remark, we currently tend to name classes of a family as "family name + specialization", so probably `LineSearch` should come first here, too.
I also find the name `LineSearch` a bit restrictive since it excludes any other kind of step length calculation. But that's an old story, see #50.

adler-j:
> As a side remark, we currently tend to name classes of a family as "family name + specialization", so probably LineSearch should come first here, too.

Agree, but I will not change this in this branch due to

> I also find the name LineSearch a bit restrictive since it excludes any other kind of step length calculation. But that's an old story, see #50.

which we should get back to at some point.

adler-j:
Fixed comments and also added some tests! Good to go?

kohr-h:
Looks good to me! We can revive the BB code later when we have a more general rule for step lengths.

PR #754, opened 2016-12-07
Add pointwise sum

adler-j:
Added PointwiseSum utility by using the PointwiseInner functionality.

In the future, we can optimize this.

### DIFF
--- a/odl/discr/tensor_ops.py
+++ b/odl/discr/tensor_ops.py
@@ -395,9 +395,7 @@ class PointwiseInnerBase(PointwiseTensorFieldOperator):
             product of an input vector field
         weight : `array-like` or float, optional
             Weighting array or constant for the norm. If an array is
-            given, its length must be equal to ``domain.size``, and
-            all entries must be positive. A provided constant must be
-            positive.
+            given, its length must be equal to ``domain.size``.
             By default, the weights are is taken from
             ``domain.weighting``. Note that this excludes unusual
             weightings with custom inner product, norm or dist.
@@ -433,16 +431,9 @@ class PointwiseInnerBase(PointwiseTensorFieldOperator):
                                  'not define a weighting vector or constant'
                                  ''.format(vfspace.weighting))
         elif np.isscalar(weight):
-            if weight <= 0:
-                raise ValueError('weighting constant must be positive, got '
-                                 '{}'.format(weight))
             self._weights = float(weight) * np.ones(vfspace.size)
         else:
             self._weights = np.asarray(weight, dtype='float64')
-            if (not np.all(self.weights > 0) or
-                    not np.all(np.isfinite(self.weights))):
-                raise ValueError('weighting array {} contains invalid '
-                                 'entries'.format(weight))
         self._is_weighted = not np.array_equiv(self.weights, 1.0)
 
     @property
@@ -498,8 +489,7 @@ class PointwiseInner(PointwiseInnerBase):
         weight : `array-like` or float, optional
             Weighting array or constant for the norm. If an array is
             given, its length must be equal to ``domain.size``, and
-            all entries must be positive. A provided constant must be
-            positive.
+            all entries must be positive.
             By default, the weights are is taken from
             ``domain.weighting``. Note that this excludes unusual
             weightings with custom inner product, norm or dist.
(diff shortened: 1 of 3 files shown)


odl/discr/tensor_ops.py
```diff
@@ -669,6 +669,67 @@ def adjoint(self):
...
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        vfspace : `ProductSpace`
+            Space of vector fields on which the operator acts.
+            It has to be a product space of identical spaces, i.e. a
+            power space.
+        weight : `array-like` or float, optional
+            Weighting array or constant for the norm. If an array is
```
adler-j: norm -> sum. General problem.

odl/discr/tensor_ops.py
```diff
@@ -669,6 +669,67 @@ def adjoint(self):
...
+# TODO: Optimize this to an optimized operator on its own.
+class PointwiseSum(PointwiseInner):
+
+    """Take the point-wise sum of a vector field.
+
+    This operator takes the (weighted) sum
+
+        ``sum(F(x)) = [ sum_j( w_j * F_j(x) ) ]
+
+    where ``F`` is a vector field. This implies that
```
kohr-h: No objections against "vector field"? Here it wouldn't be a big pain to change to "vector-valued function" which is a more generous term. The variable name could even stay :-)

odl/discr/tensor_ops.py
```diff
@@ -669,6 +669,67 @@ def adjoint(self):
...
+    def __init__(self, vfspace, weight=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        vfspace : `ProductSpace`
+            Space of vector fields on which the operator acts.
+            It has to be a product space of identical spaces, i.e. a
+            power space.
+        weight : `array-like` or float, optional
```
kohr-h: positive float
Then remove the corresponding sentence from the text below.

adler-j:
> Fix the comment you had yourself and that one I had. There's also the more general question about "vector field" against "vector-valued function". Cool otherwise.

Well this whole file is positioned wrong, since it actually works on any linear space and has nothing to do with discretizations, frankly it would play better inside productspace operators or something like that. Will fix comments.

kohr-h:
This will become even more interesting when we have tensors. Because then, all these operators are just standard operators that only act along the first axis.

odl/discr/tensor_ops.py
adler-j: Vector valued function implies it is actually a function in a stronger sense than vector field does, at least to me. This function accepts anything, i.e. a productspace of rn spaces is fine. So we dont want a very strong notion of function in there.

odl/discr/tensor_ops.py
adler-j: I went by simply removing the positivity constraint. It is not really a strict constraint anyway, everything works with negative values.

adler-j:
Updated by comments, but changed to remove the positivity condition that was rather arbitrary anyway. Needs new review I guess.

kohr-h:
> Updated by comments, but changed to remove the positivity condition that was rather arbitrary anyway. Needs new review I guess.

With "unbiased" operators like sum or inner product the constraint really was arbitrary and can be removed, but for anything that has the name "norm" in it I'm against allowing negative weigths. That will catch users off guard.

adler-j:
> With "unbiased" operators like sum or inner product the constraint really was arbitrary and can be removed, but for anything that has the name "norm" in it I'm against allowing negative weigths. That will catch users off guard.

Agree and that was my plan, will see I went over board.

adler-j:
> Agree and that was my plan, will see I went over top.

It seems I actually succeeded. Any more comments?

odl/test/discr/tensor_ops_test.py
```diff
@@ -89,12 +89,6 @@ def test_pointwise_norm_init_properties():
     with pytest.raises(ValueError):
         PointwiseNorm(vfspace, exponent=0.5)  # < 1 not allowed
 
-    with pytest.raises(ValueError):
-        PointwiseNorm(vfspace, weight=-1)  # < 0 not allowed
-
-    with pytest.raises(ValueError):
-        PointwiseNorm(vfspace, weight=[1, 0, 1])  # 0 invalid
```
kohr-h: Add these back

kohr-h:
Otherwise LGTM.

adler-j:
Merge after CI

PR #755, opened 2016-12-09
BUG: Fix uncompr_bin string fail on CentOS

adler-j:
We were getting the error

```
>           if struct.calcsize(fmt) != size_bytes:
E           TypeError: Struct() argument 1 must be string, not newstr
```

this is fixed if we only use the "default" string type of the system.

Tested on CentOS with Python 2.7.


kohr-h:
Yeah, makes sense.

PR #756, opened 2016-12-09
Add filter to fbp

adler-j:
* Adds filters, the same as in matlab, to `fbp_op`. Frequency response as follows:

filters

* Adds `tam_danielson_window` which allows good helical reconstructions.

Helical reconstruction before windowing:

filtered_backprojection

After windowing:

windowed_filtered_backprojection

as you can see the results are basically "perfect".

### DIFF
--- a/examples/tomo/filtered_backprojection_helical_3d.py
+++ b/examples/tomo/filtered_backprojection_helical_3d.py
@@ -15,15 +15,14 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""
-Example using a filtered backprojection (FBP) in helical 3d using `fbp_op`.
+"""Example using a FBP in helical 3D geometry using `fbp_op`.
 
 Note that the FBP is only approximate in this geometry, but still gives a
-decent reconstruction that can be used as an initial guess in more complicated
+decent reconstruction that can be used as an initial guess in more complex
 methods.
 
-In helical geometries, the data is in general over-sampled which causes
-streaking artefacts and a wrong scaling. This can be reduced using a
+In helical geometries, the data are in general over-sampled which causes
+streak artefacts and a wrong scaling. This can be reduced using a
 Tam-Danielson window.
 """
 
@@ -31,10 +30,10 @@ import numpy as np
 import odl
 
 
-# --- Set-up geometry of the problem --- #
+# --- Set up geometry of the problem --- #
 
 
-# Discrete reconstruction space: discretized functions on the cube
+# Reconstruction space: discretized functions on the cube
 # [-20, 20]^2 x [0, 40] with 300 samples per dimension.
 reco_space = odl.uniform_discr(
     min_pt=[-20, -20, 0], max_pt=[20, 20, 40], shape=[300, 300, 300],
@@ -42,23 +41,27 @@ reco_space = odl.uniform_discr(
 
 # Make a helical cone beam geometry with flat detector
 # Angles: uniformly spaced, n = 2000, min = 0, max = 8 * 2 * pi
+# This gives 8 full turns of the helix.
 angle_partition = odl.uniform_partition(0, 8 * 2 * np.pi, 2000)
-# Detector: uniformly sampled, n = (558, 60), min = (-30, -4), max = (30, 4)
+# Detector: uniformly sampled with a small height,
+# n = (558, 60), min = (-30, -4), max = (30, 4)
 detector_partition = odl.uniform_partition([-40, -4], [40, 4], [558, 60])
-# Spiral has a pitch of 5, we run 8 rounds (due to max angle = 8 * 2 * pi)
+# Create geometry
 geometry = odl.tomo.HelicalConeFlatGeometry(
     angle_partition, detector_partition, src_radius=100, det_radius=100,
     pitch=5.0)
 
 
-# --- Create FilteredBackProjection (FBP) operator --- #
+# --- Create Filtered Back-Projection (FBP) operator --- #
 
 
 # Ray transform (= forward projection). We use the ASTRA CUDA backend.
 ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
 
 # Unwindowed fbp
-fbp = odl.tomo.fbp_op(ray_trafo, filter_type='Hann', filter_cutoff=0.8)
+# We select a Hamming filter, and only use the lowest 80% of frequencies to
+# avoid high frequency noise.
+fbp = odl.tomo.fbp_op(ray_trafo, filter_type='Hamming', frequency_scaling=0.8)
 
 # Create Tam-Danielson window to improve result
 windowed_fbp = fbp * odl.tomo.tam_danielson_window(ray_trafo)
@@ -67,20 +70,20 @@ windowed_fbp = fbp * odl.tomo.tam_danielson_window(ray_trafo)
 # --- Show some examples --- #
 
 
-# Create a discrete Shepp-Logan phantom (modified version)
+# Create a Shepp-Logan phantom (modified version)
 phantom = odl.phantom.shepp_logan(reco_space, modified=True)
 
 # Create projection data by calling the ray transform on the phantom
 proj_data = ray_trafo(phantom)
 
-# Calculate filtered backprojection of data
+# Calculate FBP reconstructions, once without window, once with window
 fbp_reconstruction = fbp(proj_data)
 w_fbp_reconstruction = windowed_fbp(proj_data)
 
-# Shows a slice of the phantom, projections, and reconstruction
+# Show a slice of phantom, projections, and reconstruction
 phantom.show(title='Phantom')
-proj_data.show(title='Projection data (sinogram)')
-fbp_reconstruction.show(title='Filtered backprojection',
+proj_data.show(title='Simulated data (sinogram)')
+fbp_reconstruction.show(title='Filtered back-projection',
                         coords=[0, None, None], clim=[-0.1, 1.1])
-w_fbp_reconstruction.show(title='Windowed filtered backprojection',
+w_fbp_reconstruction.show(title='Windowed filtered back-projection',
                           coords=[0, None, None], clim=[-0.1, 1.1])
--- a/examples/tomo/filtered_backprojection_helical_3d.py
+++ b/examples/tomo/filtered_backprojection_helical_3d.py
@@ -15,10 +15,20 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Example using the ray transform with helical cone beam geometry."""
+"""
+Example using a filtered backprojection (FBP) in cone-beam 3d using `fbp_op`.
+
+Note that the FBP is only approximate in this geometry, but still gives a
+decent reconstruction that can be used as an initial guess in more complicated
+methods.
+"""
 
 import numpy as np
 import odl
+
+
+# --- Set-up geometry of the problem --- #
+
 
 # Discrete reconstruction space: discretized functions on the cube
 # [-20, 20]^2 x [0, 40] with 300 samples per dimension.
@@ -30,14 +40,29 @@ reco_space = odl.uniform_discr(
 # Angles: uniformly spaced, n = 2000, min = 0, max = 8 * 2 * pi
 angle_partition = odl.uniform_partition(0, 8 * 2 * np.pi, 2000)
 # Detector: uniformly sampled, n = (558, 60), min = (-30, -3), max = (30, 3)
-detector_partition = odl.uniform_partition([-30, -3], [30, 3], [558, 60])
+detector_partition = odl.uniform_partition([-40, -3], [40, 3], [558, 60])
 # Spiral has a pitch of 5, we run 8 rounds (due to max angle = 8 * 2 * pi)
 geometry = odl.tomo.HelicalConeFlatGeometry(
     angle_partition, detector_partition, src_radius=100, det_radius=100,
     pitch=5.0)
 
-# Ray transform (= forward projection). We use ASTRA CUDA backend.
+
+# --- Create FilteredBackProjection (FBP) operator --- #
+
+
+# Ray transform (= forward projection). We use the ASTRA CUDA backend.
 ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+
+# Unwindowed fbp
+unwindowed_fbp = odl.tomo.fbp_op(ray_trafo,
+                                 filter_type='Hann', filter_cutoff=0.8)
+
+# Create Tam-Danielson window to improve result
+windowed_fbp = unwindowed_fbp * odl.tomo.tam_danielson_window(ray_trafo)
+
+
+# --- Show some examples --- #
+
 
 # Create a discrete Shepp-Logan phantom (modified version)
 phantom = odl.phantom.shepp_logan(reco_space, modified=True)
@@ -45,13 +70,14 @@ phantom = odl.phantom.shepp_logan(reco_space, modified=True)
 # Create projection data by calling the ray transform on the phantom
 proj_data = ray_trafo(phantom)
 
-# Back-projection can be done by simply calling the adjoint operator on the
-# projection data (or any element in the projection space).
-backproj = ray_trafo.adjoint(proj_data)
+# Calculate filtered backprojection of data
+uw_fbp_reconstruction = unwindowed_fbp(proj_data)
+w_fbp_reconstruction = windowed_fbp(proj_data)
 
 # Shows a slice of the phantom, projections, and reconstruction
-phantom.show(indices=np.s_[:, :, 150], title='Phantom, middle z slice')
-proj_data.show(indices=np.s_[1000, :, :], title='Projection 1000')
-proj_data.show(indices=np.s_[1500, :, :], title='Projection 1500')
-backproj.show(indices=np.s_[:, :, 150],
-              title='back-projection, middle z slice')
+phantom.show(title='Phantom')
+proj_data.show(title='Projection data (sinogram)')
+uw_fbp_reconstruction.show(title='Filtered backprojection un-windowed',
+                           coords=[0, None, None], clim=[-0.1, 1.1])
+w_fbp_reconstruction.show(title='Filtered backprojection windowed',
+                           coords=[0, None, None], clim=[-0.1, 1.1])
--- a/examples/tomo/filtered_backprojection_helical_3d.py
+++ b/examples/tomo/filtered_backprojection_helical_3d.py
@@ -16,11 +16,15 @@
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Example using a filtered backprojection (FBP) in cone-beam 3d using `fbp_op`.
+Example using a filtered backprojection (FBP) in helical 3d using `fbp_op`.
 
 Note that the FBP is only approximate in this geometry, but still gives a
 decent reconstruction that can be used as an initial guess in more complicated
 methods.
+
+In helical geometries, the data is in general over-sampled which causes
+streaking artefacts and a wrong scaling. This can be reduced using a
+Tam-Danielson window.
 """
 
 import numpy as np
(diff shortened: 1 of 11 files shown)


adler-j:
Now also added to large scale tests. The biggest issue there is that these take freaking forever (basically does 5 * 2 * 2 * 5 = 100 filtered backprojection reconstructions, many in 3d.

adler-j:
Reduced lots of the redundancy in the test, now only runs 22 tests and does it in reasonable times.

adler-j:
This now works rather well. I also added a Tam-Danielson window for proper helical reconstructions which actually allows us to do decent helical reconstructions using ODL:

Helical reconstruction before windowing:

filtered_backprojection

After windowing:

windowed_filtered_backprojection

as you can see the results are basically "perfect".

adler-j:
Only major thing lacking now, and then we have a good FBP suite, is proper handling of CBCT data. For that, we need

* FDK: I.e. scaling in z direction
* Parker weighting: Scaling of data to avoid oversampling in rotation direction. Especially important for half-angle scans.

adler-j:
filters

Frequency response of the added filters, as generated by the "if __main__" code.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+                         ''.format(filter_type))
+
+    return filt
+
+
+def tam_danielson_window(ray_trafo, smoothing_width=0.05):
+    """Create Tam-Danielson window from a `RayTransform`.
+
+    The Tam-Danielson window is an indicator function on the minimal set of
+    data needed to reconstruct a given data. It is useful in analytic
```
kohr-h: we don't reconstruct data

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    reconstruction methods such as FBP to give a more accurate reconstruction.
+
+    Parameters
+    ----------
+    ray_trafo : `RayTransform`
+        The ray transform that the window should be computed for.
+    smoothing_width : float
```
kohr-h: optional

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    """Create Tam-Danielson window from a `RayTransform`.
+
+    The Tam-Danielson window is an indicator function on the minimal set of
+    data needed to reconstruct a given data. It is useful in analytic
+    reconstruction methods such as FBP to give a more accurate reconstruction.
+
+    Parameters
+    ----------
+    ray_trafo : `RayTransform`
+        The ray transform that the window should be computed for.
```
kohr-h: Unnecessarily hard to digest due to the split logic ("for" at the very end). More compact:
"The ray transform for which the window should be computed."  or
"The ray transform for which to compute the window."

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    smoothing_width : float
+        Relative width of the smoothing applied to the windows edges.
```
kohr-h: window edges? window's edges?
Can't be "windows" since there's only one window.

Relative to what?

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+
+    Returns
+    -------
+    tam_danielson_window : ``ray_trafo.range`` element
+
+    See Also
+    --------
+    fbp_op : Filtered back-projection from `RayTransform`
```
kohr-h: Filtered back-projection operator

We have both spellings "back-projection" and "backprojection" all over the place. Needs fix.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    HelicalConeFlatGeometry : The geometry this is most useful for.
```
kohr-h: See above. Why not "The primary use case for this window function."?

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    """
+
```
kohr-h: remove blank

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    # Extract parameters
+    src_radius = ray_trafo.geometry.src_radius
+    det_radius = ray_trafo.geometry.det_radius
+    pitch = ray_trafo.geometry.pitch
+    dx = ray_trafo.range.meshgrid[1].ravel()
+
+    # Find the direction that the filter should be taken in
```
kohr-h: again

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    --------
+    fbp_op : Filtered back-projection from `RayTransform`
+    HelicalConeFlatGeometry : The geometry this is most useful for.
+    """
+
+    # Extract parameters
+    src_radius = ray_trafo.geometry.src_radius
+    det_radius = ray_trafo.geometry.det_radius
+    pitch = ray_trafo.geometry.pitch
+    dx = ray_trafo.range.meshgrid[1].ravel()
```
kohr-h: `ray_trafo.geometry.det_grid.coord_vectors[0]`
makes your intention clearer

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+
+    # Find the direction that the filter should be taken in
+    axis_proj = _axis_in_detector(ray_trafo.geometry)
+
+    # Compute angles
```
kohr-h: which angles? Seems like horizontal between piercing point and detector points?

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+        filt = indicator * np.cos(norm_freq * np.pi / (2 * filter_cutoff)) ** 2
+    else:
+        raise ValueError('unknown `filter_type` ({})'
+                         ''.format(filter_type))
+
+    return filt
+
+
+def tam_danielson_window(ray_trafo, smoothing_width=0.05):
+    """Create Tam-Danielson window from a `RayTransform`.
```
kohr-h: A reference would be nice.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+
+    The Tam-Danielson window is an indicator function on the minimal set of
+    data needed to reconstruct a given data. It is useful in analytic
+    reconstruction methods such as FBP to give a more accurate reconstruction.
```
kohr-h: 2 or 3 words about the window shape would be good. It looks like the windowing is purely vertical, for example.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -50,27 +196,44 @@ def fbp_op(ray_trafo, padding=True):
         `CircularConeFlatGeometry` : Approximate reconstruction, correct in
         limit of fan angle = 0 and cone angle = 0.
 
-        `HelicalConeFlatGeometry` : Very approximate.
+        `HelicalConeFlatGeometry` : Very approximate unless a
+        `tam_danielson_window` is used. Accurate with the window.
```
kohr-h: More accurate

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -50,27 +196,44 @@ def fbp_op(ray_trafo, padding=True):
...
         be corrupted due to the circular convolution used. Using padding makes
         the algorithm slower.
+    filter_type : string, optional
+        The type of filter to be used. The options are, approximate order from
+        most noise senstive to least noise sensitive: 'Ram-Lak', 'Shepp-Logan',
+        'Cosine', 'Hamming' and 'Hann'.
+    filter_cutoff : float, optional
+        Relative cutoff frequency for the filter, a scalar in the range (0, 1].
+        The normalized frequencies are rescaled so that they fit into the range
+        [0, filter_cutoff].
```
kohr-h: Can you use this, or a simplified version of it, in the private helpers? Helps developers.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -109,12 +265,20 @@ def fft_filter(x):
             scale = (ray_trafo.geometry.src_radius /
                      (ray_trafo.geometry.src_radius +
                       ray_trafo.geometry.det_radius))
+
+            if ray_trafo.geometry.pitch != 0:
+                # In helical each projection hits the detector less than once.
```
kohr-h: Unclear what this means. In my world view, a projection is exactly what the detector spits out for a certain angle or whatever parametrizes the geometry.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -109,12 +265,20 @@ def fft_filter(x):
...
+                scale *= alen / (np.pi)
         else:
             scale = 1.0
 
         # Define ramp filter
         def fft_filter(x):
```
kohr-h: why "fft"?

examples/tomo/filtered_backprojection_helical_3d.py
```diff
@@ -0,0 +1,86 @@
...
+# Create projection data by calling the ray transform on the phantom
+proj_data = ray_trafo(phantom)
+
+# Calculate filtered backprojection of data
+fbp_reconstruction = fbp(proj_data)
+w_fbp_reconstruction = windowed_fbp(proj_data)
+
+# Shows a slice of the phantom, projections, and reconstruction
+phantom.show(title='Phantom')
+proj_data.show(title='Projection data (sinogram)')
```
adler-j: Axis is the x direction in the image shown so we are fine?

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+
+
+def _rotation_direction_in_detector(geometry):
+    """A vector in the detector plane that points in the rotation direction."""
+    du = geometry.det_init_axes[0]
+    dv = geometry.det_init_axes[1]
+    axis = geometry.axis
+    det_normal = np.cross(du, dv)
+    rot_dir = np.cross(axis, det_normal)
+    c = np.array([np.vdot(rot_dir, du), np.vdot(rot_dir, dv)])
```
adler-j: We only use absolute value of it, but you may indeed be correct. Fixing.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -21,14 +21,160 @@
...
+    Returns
+    -------
+    smoothing_filter : `numpy.ndarray`
+
+    References
+    ----------
+    http://oftankonyv.reak.bme.hu/tiki-index.php?page=Reconstruction
+    """
+    if not (0 < filter_cutoff <= 1):
+        raise ValueError('`filter_cutoff` ({}) not in the interval (0, 1]'
```
adler-j: Matlab has this requirement but frankly it doesn't add anything. Removing it.

examples/tomo/filtered_backprojection_helical_3d.py
kohr-h: Wait, what's the default for `show`? Is it `n // 2` in the last axis? Then it's actually a sinogram. 
But my point is actually that some people would be confused if you *also* call it a projection, since that's the detector output for a given angle, and those two notions are usually kept apart. Perhaps just call it "Simulated data (sinogram)" then.

adler-j: Can do, but this is in basically all examples :)

odl/tomo/analytic/filtered_back_projection.py
kohr-h: The function could be useful for other situations as well where no absolute value is taken, so 👍 for fixing it.

examples/tomo/filtered_backprojection_helical_3d.py
kohr-h: In 2D it's equivalent ;-), so needs fix only in 3D

odl/tomo/analytic/filtered_back_projection.py
adler-j: Extremely hard to write down without the source, here I guess the interested reader will have to read the reference. I added one above:

http://iopscience.iop.org/article/10.1088/0031-9155/43/4/028

odl/tomo/analytic/filtered_back_projection.py
adler-j: Well actually exact except at the edges, so I guess that is quite accurate?

adler-j:
Fixed according to comments.

adler-j:
Also added the defrise phantom and intended to use it in the helical test to better visualize the difference, but it seems it was not much better so I'll just leave it as is, anyway a nice phantom to have added.

adler-j:
So, quite happy with this now. Anything else to add? In particular regarding the `defrise` phantom.

odl/tomo/analytic/filtered_back_projection.py
kohr-h: Yeah, that's my favorite. Thanks for fixing it.

odl/phantom/geometric.py
```diff
@@ -95,6 +95,92 @@ def phantom(x):
     return space.element(phantom)
 
 
+def defrise(space, nellipses=8, alternating=False):
+    """Phantom with regularily spaced ellipses.
```
kohr-h: Ah, now that I see this: One feedback I got was the term "ellipses" that doesn't make sense in 3D. We should call things "ellipsoid" throughout. But let's make a new issue for that. (Read: I'll create a new issue.)

odl/phantom/geometric.py
```diff
@@ -95,6 +95,92 @@ def phantom(x):
...
+
+    This phantom is often used to verify cone-beam algorithms.
```
kohr-h: Is there a canonical reference?

odl/phantom/geometric.py
```diff
@@ -95,6 +95,92 @@ def phantom(x):
...
+
+def defrise_ellipses(ndim, nellipses=8, alternating=False):
+    """Ellipses for the standard Defrise phantom in 2 or 3 dimensions.
+
+    Parameters
+    ----------
+    ndim : {2, 3}
+        Dimension of the space the ellipses should be in.
+    nellipses : int, optional
+        Number of ellipses
```
kohr-h: .

odl/phantom/geometric.py
adler-j: Frankly I've really looked and cannot find one, it seems to be "standard". If you google you find like half a million answers.

odl/phantom/geometric.py
```diff
@@ -95,6 +95,92 @@ def phantom(x):
...
+            center_x = center_y = 0.0
+            center_z = -1 + 2.0 / (nellipses + 1.0) * (i + 1)
+            rotation_phi = rotation_theta = rotation_psi = 0
+
+            ellipses.append(
+                [value, axis_1, axis_2, axis_3,
+                 center_x, center_y, center_z,
+                 rotation_phi, rotation_theta, rotation_psi])
+
+    return ellipses
```
kohr-h: Hm, all I can find on the web about "Defrise" phantoms refers to disk phantoms, basically 3D stacks of identical cylinders with gaps. How's that relate to this thing?

ozanoktem: We should also add Lambda-filters :-)

examples/tomo/filtered_backprojection_helical_3d.py
```diff
@@ -0,0 +1,90 @@
...
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Example using a FBP in helical 3D geometry using `fbp_op`.
```
kohr-h: an FBP (throughout)
or here just FBP

odl/phantom/geometric.py
adler-j: > We should also add Lambda-filters :-)

Is this comment related to @kohr-h comment, or to this PR in general?

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -109,12 +285,22 @@ def fft_filter(x):
             scale = (ray_trafo.geometry.src_radius /
                      (ray_trafo.geometry.src_radius +
                       ray_trafo.geometry.det_radius))
+
+            if ray_trafo.geometry.pitch != 0:
+                # In helical geometry the whole volume is not in by each
+                # projection, ideally each point in the volume effects only by
+                # the projections in a half rotation.
```
kohr-h: Check that sentence again. I think I understand what you mean, but there are several versions in one sentence :-)

odl/phantom/geometric.py
ozanoktem: See update of comment. Misplaced and not relevant.

kohr-h: > That is exactly this thing. If that wasn't obvious I guess I'll add some more text.

To me, this looks like stacked ellipsoids, not cylinders. Am I missing something?

odl/tomo/analytic/filtered_back_projection.py
kohr-h: So why do people use complicated formulas like Katsevich (in this particular geometry) when this gives in principle exact results? And which edges do you mean? The edges of the cone, or the volume, or what else?

odl/phantom/geometric.py
kohr-h: Yeah well then screw it.

kohr-h:
So apart from general comments and questions I think this can be merged.

odl/phantom/geometric.py
adler-j: Well if you google you get lots of answers. One is this:



which is basically this phantom (with a slightly higher `nellipses` and `alternating=True`). 

Here is also an example, with `nellipses=5`:





As you see it is not particularily well defined, and as you say some take the Defrise phantom to mean with cylinders, but it seems that is mostly when people produce a physical copy of it.

Anyway I'll improve the doc.

kohr-h: Yeah, just keep it as it is and make sure to convey that this uses ellipsoids. I actually only found the physical versions of the phantom (using Duckduckgo), and I guess the cylinders are just easier to manufacture. Anyway, nice phantom to have for validation.

odl/tomo/analytic/filtered_back_projection.py
adler-j: Well this gives the wrong result for the "truncated part", otherwise it is all fine. Anyway it is still slightly wrong since by simply right-multiplying with the window we get a problem since we later want to take the derivative along the "u" direction of the detector, but since we have added this window that derivative will get a big jump which causes artefacts. You can inspect that by using the window with `smoothing_width=0`, if you want.

Anyway those artefacts do not go away as we improve the discretization, but persist. In this code we do as everyone else does and smooths the edges (via smoothing_width) which causes that derivative to "level out nicely" and we dont get any major artefacts. The problem with this is that by doing  this smoothing we lose mathematical exactness.

Katsevich solves this by proposing a method that does not need a "smoothing" style hack by doing something else than simply multiplying with a window.

Anyway, we an arbitrarily good discretization this method can get arbitrarily close to being exact, something that for example cannot be said about FDK.

adler-j: Perhaps @zickert could write more on this since you have been reading about Katsevich?

adler-j: I used [this code](https://gist.github.com/adler-j/d1eae02978296ed710994e8f5359676b) to generate an example image showing the artefact:

windowed_filtered_back-projection

odl/phantom/geometric.py
adler-j: New issue: #759

I'll not touch this for now then.

adler-j:
Merge after CI.

odl/tomo/analytic/filtered_back_projection.py
kohr-h: Thanks for the explanation, I was just curious, no need to go into extreme detail in the doc. But to me this windowing seems like a cheap trick to get rid of unwanted influences in the data, at the cost of either artefacts from the jump or inexactness due to the smoothing. And I guess noise doesn't improve the situation.. Anyway, we can close this chapter.

adler-j: I did some more thinking about this, I realized that adding more discretization points will **not** allow this cheap trick to converge to the correct solution. Even in the continuum limit, this method is inexact (but still rather good).

Anyway, as you see from the example the error is quite small.

Regarding noise, that is actually the main reason people are still doing something like this still in the clinic. The Katsevich method is even more sensitive to noise, and especially to geometry errors.

PR #758, opened 2016-12-13
Add ISTA and FISTA to algorithms

adler-j:
Frankly not the most useful algorithms since they cannot solve the TV problem or similar ones, since that would require the proximal of the TV operator. Nonetheless, I added them and a simple example of using them.

Main question is: Should these be merged into a single algorithm?

Algorithms follow the slides by Radu, I can send to anyone who's interested.


odl/solvers/util/steplen.py
```diff
@@ -118,16 +138,20 @@ def __call__(self, x, direction, dir_derivative):
             The current point
         direction : `LinearSpaceElement`
             Search direction in which the line search should be computed
-        dir_derivative : float
+        dir_derivative : float, optional
             Directional derivative along the ``direction``
+            Default: ``function.gradient(x).inner(direction)``
```
kohr-h: This requires `function.gradient` to exist, i.e. `function` to be a functional. That's obvious here, but users usually don't see this docstring, so a hint in `__init__` at the explanation of `function` would be better.

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,242 @@
...
+    """Accelerated proximal gradient algorithm for convex optimization.
+
+    The method is known as "Fast Iterative Soft-Thresholding Algorithm"
+    (FISTA).
+
+    Solves the convex optimization problem::
+
+        min_{x in X} f(x) + g(x)
+
+    where ``g`` is differentiable and the proximal operator of ``f`` is known.
```
kohr-h: Can you point out the difference to the non-accelerated version in one sentence? Like writing out the update rule for `t`, and what it is.

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,242 @@
...
+        tmp.lincomb(1, x, -gamma, g.gradient(x))
+
+        # Update x
+        x.lincomb(1 - lam_k, x, lam_k, f_prox(tmp))
+
+        if callback is not None:
+            callback(x)
+
+
+def accelerated_proximal_gradient(x, f, g, gamma, niter, callback=None,
```
kohr-h: See comments above. I'm not sure if they should be lumped into one function. The iterations are rather different.

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,242 @@
...
+    tmp = x.space.element()
+    y = x.copy()
+    t = 1
+
+    for k in range(niter):
+        # Update t
+        t, t_old = (1 + np.sqrt(1 + 4 * t ** 2)) / 2, t
+        alpha = (t_old - 1) / t
+
+        # x - tau grad_g (x)
```
kohr-h: y

kohr-h:
> Main question is: Should these be merged into a single algorithm?

I wouldn't, there are significant differences in the iterations, and that just leads to a lot of if else.



> Frankly not the most useful algorithms since they cannot solve the TV problem or similar ones, since that would require the proximal of the TV operator.

They are used e.g. in [this paper](http://epubs.siam.org/doi/abs/10.1137/090769521) for the solution of a partial problem, so not super interesting in themselves, but good to have in the arsenal anyway.

examples/solvers/proximal_gradient_tomography.py
```diff
@@ -0,0 +1,119 @@
...
+
+# --- Generate artificial data --- #
+
+
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+data = ray_trafo(discr_phantom)
+data += odl.phantom.white_noise(ray_trafo.range) * np.mean(data) * 0.1
```
adler-j: Well how would you pose denoising? We cannot do TV with this solver :/

kohr-h: You could just make up something, like L2 regularizer with a very small parameter. In this case it would probably just serve to make the solution unique, in contrast to least-squares minimization where the unregularized problem has a unique solution.

kohr-h: Alternatively, you could make a wavelet L1 regularization.

kohr-h: Don't we support compositions with unitary or "almost unitary" operators `A^*A = a I`?

adler-j: That could actually be true, but since #734 is up in the air we don't really have any of those :/

adler-j:
Made a quick-fix of the wavelet trafo and added an example with wavelet regularized reconstruction here now. Performs "OK" I guess. Needs more review of that.

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,242 @@
...
+        indexed) and return the corresponding step size.
+        Default: 1.0
+
+    Notes
+    -----
+    The problem of interest is
+
+    .. math::
+
+        \\min_{x \\in X} f(x) + g(x),
```
adler-j: If we do this it should be a global change to keep the style.

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,242 @@
...
+        raise TypeError('`x` {!r} is not in the domain of `g` {!r}'
+                        ''.format(x, g.domain))
+
+    # Step size parameter
+    gamma, gamma_in = float(gamma), gamma
+    if gamma <= 0:
+        raise ValueError('`gamma` must be positive, got {}'.format(gamma_in))
+
+    # Number of iterations
+    if not isinstance(niter, int) or niter < 0:
```
adler-j: We should do this is many places, I agree.

odl/solvers/util/steplen.py
adler-j: I don't see an issue with this. In general all of our solvers do not use this default and instead provides this manually since it is easier.

adler-j:
Finished the questions in the last review as well.

odl/trafos/wavelet.py
```diff
@@ -286,6 +286,28 @@ def _call(self, x):
         else:
             raise RuntimeError("bad `impl` '{}'".format(self.impl))
 
+    def scales(self):
```
adler-j: The design here could probably be improved. Perhaps this can somehow be moved to the Base class. In addition to that, this should have a doctest to show more clearly what the scales actually are.

odl/trafos/wavelet.py
```diff
@@ -301,7 +323,8 @@ def adjoint(self):
             if `is_orthogonal` is ``False``
         """
         if self.is_orthogonal:
-            return self.inverse
+            scale = 1 / self.domain.partition.cell_volume
```
adler-j: Another option for this is to move this into the `_call` method so that the method actually becomes orthogonal, I guess that is preferable?

examples/solvers/proximal_gradient_wavelet_tomography.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Create wavelet operator
+W = odl.trafos.WaveletTransform(space, wavelet='haar', nlevels=5)
+
+# The wavelets bases are normalized to constant norm regardless of scale.
+# since we want to penalize "small" wavelets more than "large" ones, we need
+# to weight by the scale of the wavelets.
+# The "area" of the wavelets scales as 2 ^ scale, but we use a slightly smaller
+# number in order to allow some high frequencies.
+scales = W.inverse.get_scales()
```
kohr-h: bug

examples/solvers/proximal_gradient_wavelet_tomography.py
```diff
@@ -0,0 +1,118 @@
...
+Wtrafo = W.inverse * (1 / (np.power(1.7, 1 + scales)))
+
+# Create regularizer as l1 norm
+regularizer = 0.0002 * odl.solvers.L1Norm(W.range)
+
+# l2-squared norm of residual
+l2_norm_sq = odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+
+# Compose from the right with ray transform and wavelet transform
+data_discrepancy = l2_norm_sq * ray_trafo * Wtrafo
```
kohr-h: great names here

examples/solvers/proximal_gradient_denoising.py
```diff
@@ -0,0 +1,82 @@
...
+callback.reset()
+x_acc = space.zero()
+odl.solvers.accelerated_proximal_gradient(
+    x_acc, f=data_discrepancy, g=regularizer, niter=50, gamma=gamma,
+    callback=callback)
+
+# Display images
+data.show(title='Data')
+x.show(title='L1 regularized reconstruction')
+x_acc.show(title='L1 regularized reconstruction (accelerated)', show=True)
```
adler-j: `force_show`, perhaps? To make it clear that it could be shown otherwise anyway

odl/phantom/noise.py
```diff
@@ -105,6 +108,82 @@ def poisson_noise(intensity):
...
+    """Add salt and pepper noise to vector.
+
+    Salt and pepper noise replaces random elements in ``vector`` with
+    ``low_val`` or ``high_val``.
+
+    Parameters
+    ----------
+    vector : `FnBase` or `ProductSpace`
+        The vector that noise should be added to.
+    amount : float, optional
```
kohr-h: "amount" sounds absolute, maybe the more relative-sounding "fraction"? Dunno

adler-j: `noise_fraction`?

odl/phantom/noise.py
```diff
@@ -105,6 +108,82 @@ def poisson_noise(intensity):
...
+        The propotion of elements in vector whose element should be converted
+        to noise.
+    salt_vs_pepper : float, optional
+        Relative aboundance of salt (high) vs pepper (low) noise. A high value
+        means more salt than pepper noise.
+    low_val : float, optional
+        The "pepper" color in the noise.
+        Default: minimum value of ``vector``. For product spaces, takes min for
+        each sub-space.
```
kohr-h: the minimum value per subspace is taken

If you write "for each", it could be mistaken for "across all"

odl/phantom/noise.py
```diff
@@ -105,6 +108,82 @@ def poisson_noise(intensity):
...
+        raise ValueError('`salt_vs_pepper` ({}) should be a float in the '
+                         'interval [0, 1]'.format(salt_vs_pepper_in))
+
+    if isinstance(vector.space, ProductSpace):
+        values = [salt_pepper_noise(subintensity, amount, salt_vs_pepper,
+                                    low_val, high_val)
+                  for subintensity in vector]
+    else:
+        # Extract vector of values
+        values = as_flat_array(vector).copy()
```
kohr-h: Necessary? The product space case is already handled separately.

odl/phantom/noise.py
```diff
@@ -105,6 +108,82 @@ def poisson_noise(intensity):
     return intensity.space.element(values)
 
 
+def salt_pepper_noise(vector, amount=0.05, salt_vs_pepper=0.5,
+                      low_val=None, high_val=None):
```
kohr-h: great!

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Notes
+    -----
+    The problem of interest is
+
+    .. math::
+
+        \\min_{x \\in X} f(x) + g(x),
+
+    where the technical conditions are that
```
kohr-h: ping

odl/solvers/util/steplen.py
kohr-h: comment?

odl/solvers/nonsmooth/proximal_gradient_solvers.py
adler-j: Agree

odl/solvers/nonsmooth/proximal_gradient_solvers.py
```diff
@@ -0,0 +1,236 @@
...
+    :math:`f : X \\rightarrow \mathbb{R}` is proper, convex and
+    lower-semicontinuous, and :math:`g : X \\rightarrow \mathbb{R}` is
+    differentiable and :math:`\\nabla g` is :math:`1 / \\beta`-Lipschitz
+    continuous.
+
+    Convergence is only guaranteed if the step length :math:`\\gamma` satisfies
+
+    .. math::
+
+       0 < \\gamma < 2 \\beta.
```
adler-j: Rest does not apply, this method does not let users chose an overrelaxation since the algorithm does not work in that case.

kohr-h: Ah  okay, missed that.

odl/phantom/noise.py
adler-j: Yeah you are right its redundant

odl/solvers/util/steplen.py
kohr-h: Well, currently the code anyway assumes that `function` is a `Functional` due to [this line](https://github.com/odlgroup/odl/pull/758/files/26b163d8775a73c9227761fc1f983552692729d5#diff-4af2b36cbd22929eb22d8c1dd57afe8eR126). If this is fixed, then the user sees this when trying to call the line search instance:
backtrack
If we focus on the interpreter & inspector part we see (already fading out) the help tooltip for the signature of the object:
backtrack
So `dir_derivative` is optional, right? No trace of the doc explaining *when* it is optional. And, of course, leaving it empty will fail:
```
s(0, 1)

  File "<ipython-input-4-3201524aa66e>", line 1, in <module>
    s(0, 1)

  File "/export/scratch1/kohr/git/odl/odl/solvers/util/steplen.py", line 153, in __call__
    dir_derivative = self.function.gradient(x).inner(direction)

AttributeError: 'function' object has no attribute 'gradient'
```
I don't think this is very user-friendly, even towards advanced users.

Of course, Spyder *should* display the doc when calling the object, so there's a bug, but IMO restrictions on the input parameters should *always* be documented in `__init__`, which also happens to be displayed more reliably.

odl/trafos/wavelet.py
kohr-h: A doctest would indeed be nice, it took me a minute to understand what this function does. And yes, it would make sense to move it into the base class. The only "tricky" part is to determine the wavelet space, everything else is just the same code.

kohr-h:
Did another review without knowing it from the beginning, ended up as a bunch of single comments.. Anyway, I'll approve the changes, and you can merge after this final round (and CI).

examples/solvers/proximal_gradient_wavelet_tomography.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Run the algorithm (FISTA)
+x = data_discrepancy.domain.zero()
+odl.solvers.accelerated_proximal_gradient(
+    x, f=regularizer, g=data_discrepancy, niter=200, gamma=gamma,
+    callback=callb)
+
+# Display images
+data.show(title='Data')
+x.show(title='Wavelet coefficients')
```
adler-j: Bold suggestion: How about we add some "`default_axis_labels`" thing to `DiscreteLp`, and then use that when appropriate (I guess the obvious cases are for the ray transform and the fourier transform). That way all of these will be beautiful :D

kohr-h: Hm, interesting suggestion. How would you go about it? Using `x`, `y`, `z` for up to 3d and `[x{}.format(i + 1) for i in range(self.ndim)]` for higher dimensions? The show implementation already selects the correct labels depending on the way you slice, so that wouldn't require any new code.

adler-j: #769

examples/solvers/proximal_gradient_denoising.py
adler-j: #768

odl/phantom/transmission.py
```diff
@@ -75,6 +75,11 @@ def _modified_shepp_logan_ellipses(ellipses):
     """
     intensities = [1.0, -0.8, -0.2, -0.2, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
 
+    # Add minimal numbers to ensure that the result is nowhere negative.
+    # This is needed due to numerical issues.
+    intensities[2] += 5e-17
+    intensities[3] += 5e-17
```
adler-j: This guy doesn't know about the data type and the numerical error seems to only appear for double precision. 

```python
>>> import odl
>>> space = odl.uniform_discr([0, 0], [1, 1], [100, 100], dtype='float32')
>>> np.all(np.greater_equal(odl.phantom.shepp_logan(space), 0))
```

kohr-h: I really hate that it's impossible on GitHub to simply look a bit farther up in these snippets. So goddamn hard to know in which  function you are, and because I'm lazy, I just guess.
In the emails you get it's actually much better.

Anyway, if the fix does the trick then whatever.

odl/phantom/noise.py
adler-j: I want to access the vector as if it was flat, and asarray could return a square array

adler-j:
Merge after CI

PR #770, opened 2016-12-19
ENH: Add axis_labels option to DiscreteLp, use in RayTrafo and Fourier

adler-j:
Adds axis labels and defaults, mostly intended to be used with stuff like there RayTransform where `x` for an angle makes absolutely no sense.

Example 1, fourier transform. Note that when the fourier transform is only along one axis this is clearly indicated in the plots now.

fourier_transform_along_axis_0

Example 2, ray transform. former "x" variable is now the much more readable theta, even properly typeset.

projection_data_ sinogram

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -157,9 +157,9 @@ class DiscreteLp(DiscretizedSpace):
         axis_labels = kwargs.pop('axis_labels', None)
         if axis_labels is None:
             if self.ndim <= 3:
-                self.axis_labels = ['x', 'y', 'z'][:self.ndim]
+                self.axis_labels = ['$x$', '$y$', '$z$'][:self.ndim]
             else:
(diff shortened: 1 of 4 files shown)


kohr-h:
Awesome! Just two tiny questions:
- Any chance to get the TeX'ed symbols a bit larger? They look quite small. Maybe you can simply add a `\normalsize` qualifier or something to the labels.
- Make the standard labels TeX as well?

kohr-h:
Oh, and I think usually physicists write units in square brackets, that is `theta [rad]`.

adler-j:
Changed all labels to be math-style, also simply removed the unit (it should be obvious, and interfers with hacks like the one we do for the Fourier transform).

Finally, I added an option to change font size (and used it to set it to 16 by default, which should be fine).

adler-j:
Merge?

kohr-h:
Great!

PR #773, opened 2016-12-19
BUG: Add test for ray transform angle and offset

adler-j:
This is frankly rather bad, added a better test for this and we're not consistent at all.


adler-j:
Close this in favor of #968?

kohr-h:
> Close this in favor of #968?

Yes, I have added the test there.

Issue #767, opened 2016-12-19
Why is `ProductSpaceElement` mutable?


kohr-h:
While doing stuff on tensors, I saw that entries in a `ProductSpaceElement` are [stored in a list](https://github.com/odlgroup/odl/blob/0a609ea89e2000d1dfbdad1622c959ee9d575a43/odl/space/pspace.py#L605). Didn't make much sense to me, so I changed it to a tuple. That apparently blew the hat off a couple of tests, e.g. [this line](https://github.com/odlgroup/odl/blob/0a609ea89e2000d1dfbdad1622c959ee9d575a43/odl/test/space/pspace_test.py#L571) that assumes the storage is mutable. 
I don't get it. Why? Why would it ever be a good idea to make the storage mutable? Actually, in this same test, one could insert arbitrary garbage into the list, and [the code](https://github.com/odlgroup/odl/blob/0a609ea89e2000d1dfbdad1622c959ee9d575a43/odl/space/pspace.py#L651-L657) would just happily run.

So either I'm retarded or we should change this and re-implement `__setitem__` to assign *in-place* instead of replacing the whole element in the list.

adler-j:
You are totally correct in that this is bad design. It should be changed to assigning in place.

kohr-h:
Absolutely, we shouldn't require users to add the extra `[:]`. They should think of `x[i]` as a view of the i-th part, which is itself mutable, but not replaceable by something else.

PR #774, opened 2016-12-19
ENH: implement pspace assignment broadcasting

kohr-h:
In `ProductSpaceElement`:
- Make x.parts a tuple
- Re-implement `__setitem__` with broadcasting
- Adapt tests

Closes: #767

### DIFF
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -644,9 +644,11 @@ class ProductSpaceElement(LinearSpaceElement):
             return self.parts[indices]
         elif isinstance(indices, slice):
             return self.space[indices].element(self.parts[indices])
-        else:
+        elif isinstance(indices, list):
             out_parts = [self.parts[i] for i in indices]
             return self.space[indices].element(out_parts)
+        else:
+            raise TypeError('bad index type {}'.format(type(indices)))
 
     def __setitem__(self, indices, values):
         """Implement ``self[indices] = values``."""
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -602,12 +602,12 @@ class ProductSpaceElement(LinearSpaceElement):
     def __init__(self, space, parts):
         """Initialize a new instance."""
         super().__init__(space)
-        self._parts = list(parts)
+        self.__parts = tuple(parts)
 
     @property
     def parts(self):
         """Parts of this product space element."""
-        return self._parts
+        return self.__parts
 
     @property
     def size(self):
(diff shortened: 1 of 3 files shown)


kohr-h:
Ended up implementing broadcasting when fixing the `__setitem__` implementation of product space elements. Thought it was an open issue, but it wasn't apparently.

odl/space/pspace.py
```diff
@@ -650,11 +650,41 @@ def __getitem__(self, indices):
...
+        # Get the parts to which we assign values
+        if isinstance(indices, Integral):
+            indexed_parts = (self.parts[indices],)
+            values = (values,)
+        elif isinstance(indices, slice):
+            indexed_parts = self.parts[indices]
+        elif isinstance(indices, list):
+            indexed_parts = tuple(self.parts[i] for i in indices)
+        else:
+            raise TypeError('bad index type {}'.format(type(indices)))
```
kohr-h: Question here: This is more restrictive than `__getitem__` which happily iterates over anything that allows it, including tuples. That's not Numpy-style behavior -- indexing with tuples means something very different there. (They also have indexing with arrays, but that's not applicable here)

odl/space/pspace.py
```diff
@@ -602,12 +602,12 @@ class ProductSpaceElement(LinearSpaceElement):
     def __init__(self, space, parts):
         """Initialize a new instance."""
         super().__init__(space)
-        self._parts = list(parts)
+        self.__parts = tuple(parts)
```
adler-j: About time here

odl/space/pspace.py
adler-j: I agree that we should do it like you do here.

odl/space/pspace.py
```diff
@@ -650,11 +650,41 @@ def __getitem__(self, indices):
...
-            for i, index in enumerate(indices):
-                self.parts[index] = values[i]
+            # `values` is not iterable, assume it can be assigned to
+            # all indexed parts
+            for p in indexed_parts:
+                p[:] = values
+        else:
+            # `values` is iterable; it could still represent a single
+            # element of a power space.
+            if self.space.is_power_space and values in self.space[0]:
```
adler-j: What about stuff like assigning a list? I think that works in rn, i.e.

```
>>> r2 = odl.rn(2)
>>> el = r2.element()
>>> el[:] = [1, 2]
```

don't we want something like that to work here?

odl/space/pspace.py
kohr-h: Then I'll put similar code into `__getitem__`.

kohr-h:
Finally did this last (trivial) change, merge after CI.

PR #777, opened 2016-12-20
BUG: avoid zero division in ellipse phantom for shape=1 axes

### LINKED ISSUE #776: Ellipse phantoms don't work in "degenerate" spaces
Due to [division by zero](https://github.com/odlgroup/odl/blob/master/odl/phantom/geometric.py#L434-L435), ellipse phantoms don't work if the space has shape of, e.g., `(300, 300, 1)`.

kohr-h:
Closes #776

### DIFF
--- a/odl/phantom/geometric.py
+++ b/odl/phantom/geometric.py
@@ -317,6 +317,9 @@ def _ellipse_phantom_2d(space, ellipses):
     grid = []
     for i in range(2):
         meani = (minp[i] + maxp[i]) / 2.0
+        # Where space.shape = 1, we have minp = maxp, so we set diffi = 1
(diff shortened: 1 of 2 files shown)


kohr-h:
Comments added, okay?

adler-j:
Damn thats an essay :D anyway fine with me

kohr-h:
> Damn thats an essay :D anyway fine with me

Yes, when I started writing the comment I realized that this was more of a special case than I had realized before, since usually we "squeeze" the phantom into the available space, while here we only get a partial phantom. Would you think this should be mentioned in the doc?

kohr-h:
Just went ahead and added a sentence in the docs where it fits. Merge after CI.

PR #778, opened 2016-12-20
MAINT: stop using np.matrix

kohr-h:
Trivial change in tomography utility functions, replacing `numpy.matrix` instances with `numpy.ndarray` ones.


adler-j:
Great

PR #789, opened 2017-01-05
Improve show efficiency

adler-j:
Made some fixes to the most common use pattern of `show_discrete_data`. Now actually uses the "correct" methods to update the plot instead of re-drawing the full figure in each iteration. This behavior needs to be enabled with a flag, which is currently only used by `CallbackShow`, which is the prime example of updating plots repeatedly.

For a relatively simple use case, a 200x200 pixel image, the run-time went from ~0.3s down to 0.04s for a single update. This allows "almost real time" updating, as compared to the rather clunky version before.

In response to the changes, I also increased the FPS of some examples.

The main issue with this that is yet to be solved is that Ubuntu (for some reason) identifies the plot as un-responsive after a few seconds and greys it out. I have no idea how to fix this personally without changing the behavior on OS level (which I did).

### DIFF
--- a/examples/solvers/chambolle_pock_deconvolve.py
+++ b/examples/solvers/chambolle_pock_deconvolve.py
@@ -32,7 +32,7 @@ import numpy as np
 import odl
 
 # Discretization parameters
-n = 50
+n = 128
 
 # Discretized spaces
 space = odl.uniform_discr([0, 0], [n, n], [n, n])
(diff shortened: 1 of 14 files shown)


adler-j:
Did some fixes for windows. Could you please run this on your platform and try all examples in the examples/visualization folder and the solver examples that utilize CallbackShow? See that they work as expected before we move on.

kohr-h:
> Did some fixes for windows. Could you please run this on your platform and try all examples in the examples/visualization folder and the solver examples that utilize CallbackShow? See that they work as expected before we move on.

> Edit: Please try both with and without interactive mode, I just found a bug when running from commandline.

Those fixes have broken it for Linux. The Chambolle-Pock denoising example worked before in interactive mode, now the figure isn't updated until the very end.

adler-j:
Now validated on windows and ubuntu with: Spyder, Spyder+IPython and python from console.

Can you try again?

kohr-h:
Tried the first ~5 solvers examples and the new update example, works in interactive and in non-interactive mode.
I noticed that the 1D deconvolution example runs way too fast now, the result is immediately there. Would be good to have a `CallbackSleep` for that purpose.

adler-j:
Well that's a minor issue. How about we merge this now and perhaps look at that later?

kohr-h:
Yes, definitely. Made a new issue #792.

PR #796, opened 2017-01-08
MAINT: Improve performance of operator algebra derivative if linear

adler-j:
Small (actually quite noticeable for OperatorComp) improvement to the derivative computation given linear operators.

### DIFF
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -1112,8 +1112,12 @@ class OperatorSum(Operator):
         x : `domain` `element-like`
             Evaluation point of the derivative
         """
-        return OperatorSum(self.left.derivative(x), self.right.derivative(x),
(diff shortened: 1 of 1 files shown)


odl/operator/operator.py
```diff
@@ -1328,11 +1332,14 @@ def derivative(self, x):
             Evaluation point of the derivative. Needs to be usable as
             input for the ``right`` operator.
         """
-        left_deriv = self.left.derivative(self.right(x))
-        right_deriv = self.right.derivative(x)
+        if self.is_linear:
+            return self
+        else:
+            left_deriv = self.left.derivative(self.right(x))
```
adler-j: This is where it amkes the largest difference, since we need to call `self.right(x)`, which gives an extra operator evaluation.

kohr-h:
Totally makes sense.

PR #802, opened 2017-01-11
WIP: Add caching to astra cuda projectors

adler-j:
Work in progress.

Adds caching of the CUDA memory for astra projectors, resulting in rather immense speedups. On my Titan X card I can get speed-up of about x10.

TODO:

- [x] cleanup
- [x] add test
- [x] Optimize back-projection as well
- [x] Fix 3d case.
- [x] Fix GPU memory leak. Clarification: It seems GPU memory use goes in a "zig-zag" pattern, increasing for a while then dropping to "zero". Edit: Fixed, forgot to remove a ASTRA memory handle.

### DIFF
--- a/examples/tomo/backends/astra_performance_cpu_parallel_2d_cg.py
+++ b/examples/tomo/backends/astra_performance_cpu_parallel_2d_cg.py
@@ -1,0 +1,116 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ODL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Performance example of running native astra vs using ODL for reconstruction.
+
+In this example, a 512x512 image is reconstructed using the Conjugate Gradient
+Least Squares method on the CPU.
+
+In general, ASTRA is fater than ODL since it does not need to perform any
+copies and all arithmetic is performed in place. Dispite this, ODL is not much
+slower. In this example, the overhead is about x1.2, depending on the
+hardware used.
+"""
+
+import astra
+import numpy as np
+import matplotlib.pyplot as plt
+import scipy
+import odl
+
+
+# Common geometry parameters
+
+domain_size = np.array([512, 512])
+n_angles = 180
+det_size = 362
+niter = 20
+data = np.rot90(scipy.misc.ascent().astype('float'), -1)
+
+
+# --- ASTRA ---
+
+
+# Define ASTRA geometry
+vol_geom = astra.create_vol_geom(domain_size[0], domain_size[1])
+proj_geom = astra.create_proj_geom('parallel',
+                                   np.linalg.norm(domain_size) / det_size,
+                                   det_size,
+                                   np.linspace(0, np.pi, n_angles))
+
+# Create ASTRA projector
+proj_id = astra.create_projector('line', proj_geom, vol_geom)
+
+# Create sinogram
+sinogram_id, sinogram = astra.create_sino(data, proj_id)
+
+# Create a data object for the reconstruction
+rec_id = astra.data2d.create('-vol', vol_geom)
+
+# Set up the parameters for a reconstruction algorithm using the CUDA backend
+cfg = astra.astra_dict('CGLS')
+cfg['ReconstructionDataId'] = rec_id
+cfg['ProjectionDataId'] = sinogram_id
+cfg['ProjectorId'] = proj_id
+
+# Create the algorithm object from the configuration structure
+alg_id = astra.algorithm.create(cfg)
+
+with odl.util.Timer('Astra run'):
+    # Run the algorithm
+    astra.algorithm.run(alg_id, niter)
+
+# Get the result
+rec = astra.data2d.get(rec_id)
+
+# Clean up.
+astra.algorithm.delete(alg_id)
+astra.data2d.delete(rec_id)
+astra.data2d.delete(sinogram_id)
+astra.projector.delete(proj_id)
+
+# --- ODL ---
+
+# Create reconstruction space
+reco_space = odl.uniform_discr(-domain_size/2, domain_size/2, domain_size)
+
+# Create geometry
+geometry = odl.tomo.parallel_beam_geometry(reco_space, n_angles, det_size)
+
+# Create ray transform
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cpu')
+
+# Create sinogram
+rhs = ray_trafo(data)
+
+# Solve with CGLS (aka CGN)
+x = reco_space.zero()
+with odl.util.Timer('ODL run'):
+    odl.solvers.conjugate_gradient_normal(ray_trafo, x, rhs, niter=niter)
+
+# Display results for comparsion
+plt.figure('data')
+plt.imshow(data.T, origin='lower', cmap='bone')
+plt.figure('ASTRA sinogram')
+plt.imshow(sinogram.T, origin='lower', cmap='bone')
+plt.figure('ASTRA reconstruction')
+plt.imshow(rec.T, origin='lower', cmap='bone')
+plt.figure('ODL sinogram')
+plt.imshow(rhs.asarray().T, origin='lower', cmap='bone')
+plt.figure('ODL reconstruction')
+plt.imshow(x.asarray().T, origin='lower', cmap='bone')
+plt.show()
--- a/examples/tomo/backends/astra_performance_cpu_parallel_2d_cg.py
+++ b/examples/tomo/backends/astra_performance_cpu_parallel_2d_cg.py
@@ -15,13 +15,13 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Performance example of running native astra vs using ODL for reconstruction.
+"""Performance example of running native ASTRA vs using ODL for reconstruction.
 
 In this example, a 512x512 image is reconstructed using the Conjugate Gradient
 Least Squares method on the CPU.
 
-In general, ASTRA is fater than ODL since it does not need to perform any
-copies and all arithmetic is performed in place. Dispite this, ODL is not much
+In general, ASTRA is faster than ODL since it does not need to perform any
+copies and all arithmetic is performed in place. Despite this, ODL is not much
 slower. In this example, the overhead is about x1.2, depending on the
 hardware used.
 """
@@ -61,7 +61,7 @@ sinogram_id, sinogram = astra.create_sino(data, proj_id)
 # Create a data object for the reconstruction
 rec_id = astra.data2d.create('-vol', vol_geom)
 
-# Set up the parameters for a reconstruction algorithm using the CUDA backend
+# Set up the parameters for a reconstruction algorithm using the CPU backend
 cfg = astra.astra_dict('CGLS')
 cfg['ReconstructionDataId'] = rec_id
 cfg['ProjectionDataId'] = sinogram_id
@@ -70,7 +70,7 @@ cfg['ProjectorId'] = proj_id
 # Create the algorithm object from the configuration structure
 alg_id = astra.algorithm.create(cfg)
 
-with odl.util.Timer('Astra run'):
+with odl.util.Timer('ASTRA run'):
     # Run the algorithm
     astra.algorithm.run(alg_id, niter)
 
@@ -86,7 +86,7 @@ astra.projector.delete(proj_id)
 # --- ODL ---
 
 # Create reconstruction space
-reco_space = odl.uniform_discr(-domain_size/2, domain_size/2, domain_size)
+reco_space = odl.uniform_discr(-domain_size / 2, domain_size / 2, domain_size)
 
 # Create geometry
 geometry = odl.tomo.parallel_beam_geometry(reco_space, n_angles, det_size)
@@ -102,7 +102,7 @@ x = reco_space.zero()
 with odl.util.Timer('ODL run'):
     odl.solvers.conjugate_gradient_normal(ray_trafo, x, rhs, niter=niter)
 
-# Display results for comparsion
+# Display results for comparison
 plt.figure('data')
 plt.imshow(data.T, origin='lower', cmap='bone')
 plt.figure('ASTRA sinogram')
--- a/examples/tomo/backends/astra_performance_cuda_parallel_2d_cg.py
+++ b/examples/tomo/backends/astra_performance_cuda_parallel_2d_cg.py
@@ -15,13 +15,13 @@
 # You should have received a copy of the GNU General Public License
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Performance example of running native astra vs using ODL for reconstruction.
+"""Performance example of running native ASTRA vs using ODL for reconstruction.
 
 In this example, a 512x512 image is reconstructed using the Conjugate Gradient
 Least Squares method on the GPU.
 
 In general, ASTRA is fater than ODL since it does not need to perform any
-copies and all arithmetic is performed on the GPU. Dispite this, ODL is not
+copies and all arithmetic is performed on the GPU. Despite this, ODL is not
 much slower. In this example, the overhead is about x2, depending on the
 hardware used.
 """
@@ -70,7 +70,7 @@ cfg['ProjectorId'] = proj_id
 # Create the algorithm object from the configuration structure
 alg_id = astra.algorithm.create(cfg)
 
-with odl.util.Timer('Astra run'):
+with odl.util.Timer('ASTRA run'):
     # Run the algorithm
     astra.algorithm.run(alg_id, niter)
 
@@ -86,7 +86,7 @@ astra.projector.delete(proj_id)
 # --- ODL ---
 
 # Create reconstruction space
-reco_space = odl.uniform_discr(-domain_size/2, domain_size/2, domain_size)
+reco_space = odl.uniform_discr(-domain_size / 2, domain_size / 2, domain_size)
 
 # Create geometry
 geometry = odl.tomo.parallel_beam_geometry(reco_space, n_angles, det_size)
@@ -102,7 +102,7 @@ x = reco_space.zero()
 with odl.util.Timer('ODL run'):
     odl.solvers.conjugate_gradient_normal(ray_trafo, x, rhs, niter=niter)
 
-# Display results for comparsion
+# Display results for comparison
 plt.figure('data')
 plt.imshow(data.T, origin='lower', cmap='bone')
 plt.figure('ASTRA sinogram')
--- a/examples/tomo/backends/astra_performance_cuda_parallel_2d_cg.py
+++ b/examples/tomo/backends/astra_performance_cuda_parallel_2d_cg.py
@@ -1,0 +1,116 @@
+# Copyright 2014-2016 The ODL development group
+#
+# This file is part of ODL.
+#
+# ODL is free software: you can redistribute it and/or modify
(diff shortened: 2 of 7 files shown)


adler-j:
This is now a priority, it makes a huge difference. Any input in how to design this would be great, but the current idea is actually quite descent (imo).

adler-j:
With this and the lincomb caching fixed, tomography problems will run almost as fast as "native" i.e. calling astra and other stuff by hand, frankly faster since we call BLAS etc which most users would not do.

adler-j:
```python
import odl

space = odl.uniform_discr([-129, -129], [129, 129], [400, 400])

geometry = odl.tomo.parallel_beam_geometry(space)

ray_trafo = odl.tomo.RayTransform(space, geometry, impl='astra_cuda', use_cache=False)  # cache true or false.

grad = odl.Gradient(space)

phantom = odl.phantom.shepp_logan(space)
b = ray_trafo(phantom)

l2_sq = odl.solvers.L2NormSquared(ray_trafo.range).translated(b)
l1 = odl.solvers.GroupL1Norm(grad.range)
f = odl.solvers.ZeroFunctional(space)

g = [l2_sq, 3.0 * l1]
L = [ray_trafo, grad]
tau = 2
sigma = [1.0 / odl.power_method_opnorm(ray_trafo)**2,
         1.0 / odl.power_method_opnorm(grad)**2]

callback = (odl.solvers.CallbackPrintTiming())

fbp_op = odl.tomo.fbp_op(ray_trafo,
                         filter_type='Hann', frequency_scaling=0.7)
x = fbp_op(b)
odl.solvers.douglas_rachford_pd(x, f, g, L, tau, sigma, niter=5000,
                                callback=callback)
```

With cache false:

```
Time elapsed = 0.348 s
Time elapsed = 0.225 s
Time elapsed = 0.210 s
Time elapsed = 0.230 s
Time elapsed = 0.243 s
Time elapsed = 0.238 s
Time elapsed = 0.232 s
Time elapsed = 0.235 s
Time elapsed = 0.240 s
Time elapsed = 0.245 s
```

with cache true:

```
Time elapsed = 0.145 s
Time elapsed = 0.067 s
Time elapsed = 0.068 s
Time elapsed = 0.062 s
Time elapsed = 0.068 s
Time elapsed = 0.070 s
Time elapsed = 0.067 s
Time elapsed = 0.068 s
Time elapsed = 0.065 s
Time elapsed = 0.075 s
```

Improvements seem to depend on the data size and projector type. For some reason it seems bigger for fan beam geometries.

kohr-h:
The diff is a bit hard to interpret since Git(Hub) does a poor job understanding the changes. So do I understand it correctly that the classes have ASTRA keep a persistent reference of (memory for)
- volume
- data
- projector
- algorithm
and that the only major remaining overhead is axis swapping for 3 dimensions? If that's true, great!

Regarding design, I don't think I can come up with a better solution. ASTRA's Python wrapper does implement reference counting, so we don't leak memory if our element is deleted, right? That would be my only worry, otherwise it's probably simplest to let ASTRA hold on to the memory.

adler-j:
> The diff is a bit hard to interpret since Git(Hub) does a poor job understanding the changes. So do I understand it correctly that the classes have ASTRA keep a persistent reference of (memory for)
    - volume
    - data
    - projector
    - algorithm
    and that the only major remaining overhead is axis swapping for 3 dimensions? If that's true, great!

Yes, with some overhead for copying data in each step (which seems "small"). As of now a forward projection is about 70-80% spent inside `astra.run`

> Regarding design, I don't think I can come up with a better solution. ASTRA's Python wrapper does implement reference counting, so we don't leak memory if our element is deleted, right? That would be my only worry, otherwise it's probably simplest to let ASTRA hold on to the memory.

Great! I'll go ahead and clean this up then.

adler-j:
This is now ready for an actual review. Style is perhaps not 100%, but it should be reviewable before final touches.

### Performance test
I ran a test to compare performance to master, running [this script](https://gist.github.com/adler-j/01d9f97cf593e07dbf07ea7b44e23698) i got the following results:

|  Setup                 | Before | After|
|----------------------- | ------ | -----|
|   par2d single forward | 0.005  | 0.005|
|  par2d single backward | 0.005  | 0.008|
|  cone2d single forward | 0.048  | 0.047|
| cone2d single backward | 0.027  | 0.028|
|   par3d single forward | 0.077  | 0.085|
|  par3d single backward | 0.048  | 0.050|
|  cone3d single forward | 0.085  | 0.105|
| cone3d single backward | 0.060  | 0.050|
| helical single forward | 0.102  | 0.130|
|helical single backward | 0.073  | 0.068|
|      par2d 100 forward | 0.670  | 0.202|
|     par2d 100 backward | 0.605  | 0.110|
|     cone2d 100 forward | 2.739  | 0.212|
|    cone2d 100 backward | 2.748  | 0.145|
|      par3d 100 forward | 4.861  | 1.643|
|     par3d 100 backward | 5.339  | 1.540|
|     cone3d 100 forward | 4.726  | 1.637|
|    cone3d 100 backward | 4.963  | 1.607|
|    helical 100 forward | 6.711  | 2.798|
|   helical 100 backward | 6.133  | 1.538|

As you can see it seems my new implementation may have lost some minor performance in single call situations (in some cases at least), about 10%. However, in multiple-call situations we have speedups that are simply huge.

We have the following speedups per iteration:

|  Setup  | Speedup |
|---------| ------- |
|   par2d | 300%    |
|  cone2d | 1400%   |
|   par3d | 220%    |
|  cone3d | 200%    |
| helical | 200%    |

That is ofc given that users can sacrifice enough ram and GPU memory to "waste" on the projector in each iteration.

An obvious follow-up to this is to somehow allow re-use of the memory, but that is for later.

adler-j:
Now also added examples that show ODL performance as compared to astra, went from x10 to x2 in the coda case, so we don't have to feel as ashamed anymore.

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+    def __init__(self, geometry, reco_space, proj_space, use_cache):
+        """Run an ASTRA forward projection on the given data using the GPU.
 
-        # Wrap the array in correct dtype etc if needed
-        with writable_array(out, dtype='float32', order='C') as arr:
-            sino_id = astra_data(proj_geom, datatype='projection', data=arr)
+        Parameters
+        ----------
+        geometry : `Geometry`
+            Geometry defining the tomographic setup
```
kohr-h: dot

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+        out : ``proj_space`` element, optional
+            Element of the projection space to which the result is written. If
+            ``None``, an element in ``proj_space`` is created.
```
kohr-h: Update with correct parameter list.

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+
+    def __init__(self, geometry, reco_space, proj_space, use_cache):
+        """Run an ASTRA forward projection on the given data using the GPU.
+
+        Parameters
+        ----------
+        geometry : `Geometry`
+            Geometry defining the tomographic setup
+        reco_space : `DiscreteLp`
+            Space to which the calling operator maps
```
kohr-h: - dot
- `proj_space` missing

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+
+    """Thin wrapper around ASTRA."""
+
+    algo_id = None
+    vol_id = None
+    sino_id = None
+    proj_id = None
+
+    def __init__(self, geometry, reco_space, proj_space, use_cache):
+        """Run an ASTRA forward projection on the given data using the GPU.
```
kohr-h: Wrong line

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
-
-    # Create projector
-    proj_id = astra_projector('nearest', vol_geom, proj_geom, ndim,
-                              impl='cuda')
-
-    if ndim == 2:
-        if out is None:
-            out = proj_space.element()
+    def __init__(self, geometry, reco_space, proj_space, use_cache):
+        """Run an ASTRA forward projection on the given data using the GPU.
```
kohr-h: Wrong line

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+
+        Parameters
+        ----------
+        geometry : `Geometry`
+            Geometry defining the tomographic setup
+        reco_space : `DiscreteLp`
+            Space to which the calling operator maps
+        use_cache : bool
+            True if data should be cached. Default: True
```
kohr-h: see above

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+    sino_id = None
+    proj_id = None
+
+    def __init__(self, geometry, reco_space, proj_space, use_cache):
+        """Run an ASTRA forward projection on the given data using the GPU.
+
+        Parameters
+        ----------
+        geometry : `Geometry`
+            Geometry defining the tomographic setup
```
kohr-h: dot

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+
+        if not isinstance(proj_space, DiscreteLp):
+            raise TypeError('projection space {!r} is not a DiscreteLp '
+                            'instance'.format(proj_space))
+        self.proj_space = proj_space
+
+        self.use_cache = bool(use_cache)
+
+        if self.use_cache:
+            self.create_ids()
```
kohr-h: See above

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+
+    def call_backward(self, proj_data, out=None):
+        """Run an ASTRA backward projection on the given data using the GPU.
```
kohr-h: back-projection

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+            raise TypeError('projection data {!r} is not a DiscreteLpElement '
+                            'instance'.format(proj_data))
+        if self.reco_space.ndim != self.geometry.ndim:
+            raise ValueError('dimensions {} of reconstruction space and {} of '
+                             'geometry do not match'.format(
+                                 self.reco_space.ndim, self.geometry.ndim))
+        if out is not None:
+            if not isinstance(out, DiscreteLpElement):
+                raise TypeError('`out` {} is neither None nor a '
+                                'DiscreteLpElement instance'.format(out))
```
kohr-h: See above

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+        self.sino_id = astra_data(proj_geom, datatype='projection',
+                                  data=self.in_array,
+                                  allow_copy=False)
 
-    return out
+        # Create projector
+        self.proj_id = astra_projector('nearest', vol_geom, proj_geom,
+                                       ndim, impl='cuda')
+
+        # Wrap the array in correct dtype etc if needed
```
kohr-h: wrong comment?

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+            self.vol_id = None
+        if self.sino_id is not None:
+            adata.delete(self.sino_id)
+            self.sino_id = None
+        if self.proj_id is not None:
+            aproj.delete(self.proj_id)
+            self.proj_id = None
+
+    def __del__(self):
+        self.delete_ids()
```
kohr-h: see above

odl/tomo/operators/ray_trafo.py
```diff
@@ -72,6 +72,8 @@ def __init__(self, discr_domain, geometry, impl='astra_cpu', **kwargs):
         discr_range : `DiscreteLp`
             Discretized space, the range of the forward projector.
             Default: Infered from parameters.
+        use_cache : bool
+            True if data should be cached. Default: True
```
kohr-h: see above (default correct though)

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+            raise TypeError('projection space {!r} is not a DiscreteLp '
+                            'instance'.format(proj_space))
+        self.proj_space = proj_space
+
+        self.use_cache = bool(use_cache)
+
+        if self.use_cache:
+            self.create_ids()
+
+    def call_forward(self, vol_data, out=None):
```
kohr-h: Why not implement this with `__call__`?

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+            raise TypeError('projection space {!r} is not a DiscreteLp '
+                            'instance'.format(proj_space))
+        self.proj_space = proj_space
+
+        self.use_cache = bool(use_cache)
+
+        if self.use_cache:
+            self.create_ids()
+
+    def call_backward(self, proj_data, out=None):
```
kohr-h: `__call__`?

odl/tomo/operators/ray_trafo.py
```diff
@@ -240,6 +256,8 @@ def __init__(self, discr_range, geometry, impl='astra_cpu', **kwargs):
         discr_domain : `DiscreteLp`
             Discretized space, the range of the forward projector.
             Default: Infered from parameters.
+        use_cache : bool
+            True if data should be cached. Default: True
```
kohr-h: see above

odl/tomo/operators/ray_trafo.py
```diff
@@ -321,8 +344,13 @@ def _call(self, x, out=None):
...
             elif data_impl == 'cuda':
-                return astra_cuda_back_projector(x, self.geometry,
-                                                 self.range, out)
+                backproj = getattr(self, 'astra_backprojector', None)
+                if backproj is None:
+                    self.astra_backprojector = AstraCudaBackProjectorImpl(
+                        self.geometry, self.range, self.domain,
+                        use_cache=self.use_cache)
+
+                return self.astra_backprojector.call_backward(x, out)
```
kohr-h: This would look much better with just `self.astra_backprojector(x, out)`.

kohr-h:
Needless to say that this is a huge improvement, really nice work!

examples/tomo/backends/astra_performance_cpu_parallel_2d_cg.py
```diff
@@ -0,0 +1,116 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Performance example of running native astra vs using ODL for reconstruction.
+
+In this example, a 512x512 image is reconstructed using the Conjugate Gradient
+Least Squares method on the CPU.
+
+In general, ASTRA is fater than ODL since it does not need to perform any
```
adler-j: Well here it doesn't use the same things. We both share the projector, but astra then performs arithmetic interally, while we do it externally

examples/tomo/backends/astra_performance_cuda_parallel_2d_cg.py
```diff
@@ -0,0 +1,116 @@
...
+"""Performance example of running native astra vs using ODL for reconstruction.
+
+In this example, a 512x512 image is reconstructed using the Conjugate Gradient
+Least Squares method on the GPU.
+
+In general, ASTRA is fater than ODL since it does not need to perform any
+copies and all arithmetic is performed on the GPU. Dispite this, ODL is not
+much slower. In this example, the overhead is about x2, depending on the
+hardware used.
+"""
```
adler-j: Felt like this was one of the questions that people kept asking every time I did an ODL introduction: "How much overhead is it?" etc. An example like this would be nice to show them.

odl/tomo/backends/astra_cuda.py
adler-j: I had a plan of having a single object for both the forward and back projection, that way we get half the memory overhead. Naming like this is a step for enabling that later.

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+            raise TypeError('volume data {!r} is not a DiscreteLpElement '
+                            'instance'.format(vol_data))
+        if vol_data.ndim != self.geometry.ndim:
+            raise ValueError('dimensions {} of volume data and {} of geometry '
+                             'do not match'
+                             ''.format(vol_data.ndim, self.geometry.ndim))
+        if out is not None:
+            if not isinstance(out, DiscreteLpElement):
+                raise TypeError('`out` {} is neither None nor a '
+                                'DiscreteLpElement instance'.format(out))
```
adler-j: Will make them asserts

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+                out = self.proj_space.element()
+
+            out[:] = np.rollaxis(self.out_array, 0, 3)
+        else:
+            raise RuntimeError('unknown ndim')
+
+        # Fix inconsistent scaling
+        if isinstance(self.geometry, Parallel2dGeometry):
+            # parallel2d scales with pixel stride
+            out *= 1 / float(self.geometry.det_partition.cell_sides[0])
```
adler-j: That is not the same as this line, your value is ~the square of this.

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+        self.out_array = np.empty(astra_proj_shape,
+                                  dtype='float32', order='C')
+
+        # Create ASTRA data structures
+        vol_geom = astra_volume_geometry(self.reco_space)
+        proj_geom = astra_projection_geometry(self.geometry)
+        self.vol_id = astra_data(vol_geom, datatype='volume',
+                                 data=self.in_array, allow_copy=False)
+
+        self.proj_id = astra_projector('nearest', vol_geom, proj_geom,
```
adler-j: For CUDA it is the only option

odl/tomo/backends/astra_cuda.py
```diff
@@ -38,207 +38,340 @@
...
+            self.sino_id = None
+        if self.proj_id is not None:
+            aproj.delete(self.proj_id)
+            self.proj_id = None
+
+        self.in_array = None
+        self.out_array = None
+
+    def __del__(self):
+        self.delete_ids()
```
adler-j: inside `delete_ids`? Should already be done, but I'll have a look.

odl/tomo/operators/ray_trafo.py
adler-j: Agree, but as said there is a long term goal of enabling memory re-use here.

examples/tomo/backends/astra_performance_cpu_parallel_2d_cg.py
kohr-h: Oh right, you also use their *algorithm*, overlooked that.

odl/tomo/backends/astra_cuda.py
adler-j: One option is to make these actual `odl.Operator` classes and hide the "deep stuff" in here for later, but I'll stay with this for this PR.

adler-j:
Thanks for review,will get it done asap.

odl/tomo/backends/astra_cuda.py
kohr-h: No I mean in `__del__`.

odl/tomo/operators/ray_trafo.py
kohr-h: Well I was guessing that it was intended for making it one class but rather that had been the goal once and then abandoned. It's totally fine in the current case.

odl/tomo/backends/astra_cuda.py
adler-j: I don't get it, why would I need to do that?

adler-j:
Fixed according to comments!

odl/tomo/backends/astra_cuda.py
kohr-h: Okay, never mind. The reference is deleted automatically when removing the object.
But then, `del` is not really recommended in the sense of "guarantee the things get deleted". 

So I would prefer having a more general `clear_cache` method that also sets the array references to `None`. It's a clearer indicator to the user that this is intended to clear the cache. You never know if `__del__` is implemented or not, but you clearly see the public methods.

kohr-h:
Well then, click the button!

adler-j:
Lunch is tasty!

PR #804, opened 2017-01-12
ENH: Allow users to get materials from phantom.forbild, closes #803

### LINKED ISSUE #803: Add utility to get forbild by material type
Currently `phantom.forbild` returns the density map of the forbild phantom. It would be useful if it also had the option to return the material indices.

adler-j:


odl/phantom/transmission.py
```diff
@@ -330,7 +349,28 @@ def transposeravel(arr):
...
+        # less_dense_sphere
+        materials[(image > 1.047) & (image < 1.048)] = 2
+        # brain
+        materials[(image > 1.048) & (image < 1.052)] = 3
+        # denser_sphere
+        materials[(image > 1.052) & (image < 1.053)] = 4
+        # blood
+        materials[(image > 1.053) & (image < 1.056)] = 5
+        # eye
+        materials[(image > 1.058) & (image < 1.062)] = 6
```
kohr-h: gap between 1.056 and 1.058 - not that it makes a difference, but looks like a gotcha.

odl/phantom/transmission.py
```diff
@@ -253,12 +253,27 @@ def _analytical_forbild_phantom(resolution, ear):
...
+    =========================  =====  =======
+    Air                        0      0.0000
+    Cerebrospinal fluid (CSF)  1      1.0450
+    Small less dense sphere    2      1.0475
+    Brain                      3      1.0500
+    Small more dense sphere    4      1.0525
+    Blood                      5      1.0550
+    Eyes                       6      1.0600
+    Bone                       7      1.8000
+    =========================  =====  =======
```
kohr-h: nice!

odl/phantom/transmission.py
adler-j: Shouldn't be a problem, the reason i use < instead of == is floating point rounding.

kohr-h: I know the code is correct. It's more a psychological thing - you see all the nicely matching bounds before, and then "oops, this looks wrong". I'd just change the 56 for 58 to avoid questions.

adler-j: Will fix :)

adler-j:
Fixed comments. Merge?

odl/phantom/transmission.py
```diff
@@ -317,9 +317,9 @@ def transposeravel(arr):
     # Rescale points to the default grid.
     # The forbild phantom is defined on [-12.8, 12.8] x [-12.8, 12.8]
     xcoord, ycoord = space.points().T
-    xcoord = (xcoord - np.min(xcoord)) / (np.max(xcoord) - np.min(xcoord))
+    xcoord = (xcoord - space.min_pt[0]) / (space.max_pt[0] - space.min_pt[0])
     xcoord = 25.8 * xcoord - 12.8
-    ycoord = (ycoord - np.min(ycoord)) / (np.max(ycoord) - np.min(ycoord))
+    ycoord = (ycoord - space.min_pt[1]) / (space.max_pt[1] - space.min_pt[1])
```
kohr-h: We really need `.grid` here in between, now you get the partition min/max instead of the grid min/max.
```python
>>> space = odl.uniform_discr(0, 1, 5)
>>> space.min_pt
array([ 0.])
>>> space.grid.min_pt
array([ 0.1])
>>> np.min(space.points())
0.10000000000000001
```

adler-j: But we want to use the partition coordinates. I.e. after rescaling the minimum point should not be exactly 0, but slightly larger. No?

kohr-h: That's quite possible. I just looked at before vs. after, and the behavior before was the same as with `grid`. If it should always have been the partition coordinates, then fine.

PR #815, opened 2017-01-15
Issue 800  product space getitem

adler-j:


odl/operator/pspace_ops.py
```diff
@@ -354,9 +358,47 @@ def adjoint(self):
         adj_matrix = sp.sparse.coo_matrix((adjoint_ops, indices), shape)
         return ProductSpaceOperator(adj_matrix, self.range, self.domain)
 
+    def __getitem__(self, index):
+        """Get sub-operator by index.
+
+        Returns
+        -------
+        suboperator : `Operator` or ``None``
+            If there is an operator at [row, col], returns the operator,
```
adler-j: > Unclear how [row, col] relate to index.

Will fix

>Also, BroadcastOperator and ReductionOperator should support 1d indexing, how about that?

They already do. Only one missing is DiagonalOperator

kohr-h: > They already do.

Separate implementation? Otherwise it's a mystery to me when the matrix isn't squashed.

adler-j: They aren't even `ProductSpaceOperators`, they only use them as an internal representation.

kohr-h: Ah, mystery solved.

adler-j:
Fixed comments, merge?

An obvious followup is to allow "single index access" of `ProductSpaceOperator` which would return a `ReductionOperator` or a `BroadcastOperator`. But we'll leave that for now since the uses cases are not really abundant.

adler-j:
Bump

odl/operator/pspace_ops.py
```diff
@@ -354,9 +358,51 @@ def adjoint(self):
...
+        Parameters
+        ----------
+        index : tuple of int
+            A pair of integers given as (row, col).
+
+        Returns
+        -------
+        suboperator : `Operator` or ``0``
+            If there is an operator at (row, col), the operator is returned,
+            otherwise returns ``0``.
```
kohr-h: ~~returns~~

odl/operator/pspace_ops.py
```diff
@@ -354,9 +358,51 @@ def adjoint(self):
...
+        else:
+            return self.ops.data[int(linear_index)]
+
     def __repr__(self):
         """Return ``repr(self)``."""
-        return 'ProductSpaceOperator({!r})'.format(self.ops)
+        aslist = [[0] * self.domain.size for _ in range(self.range.size)]
+        for i, j, op in zip(self.ops.row, self.ops.col, self.ops.data):
+            aslist[i][j] = op
+        return 'ProductSpaceOperator({!r})'.format(aslist)
```
kohr-h: `self.__class__.__name__`

adler-j:
Merge after CI.

kohr-h:
Do that, it's already mentioned in the release notes :-) #821

PR #816, opened 2017-01-15
MAINT: Change ellipse to ellipsoid where appropriate

adler-j:


adler-j:
God damnit, the update button sucks, it doesn't give warnings about when it will merge and when it wont. Any reason we got it now?

kohr-h:
> God damnit, the update button sucks, it doesn't give warnings about when it will merge and when it wont. Any reason we got it now?

I didn't even know it was there, always do it on the command line.
But it said "merge current master into blabla" a couple of times recently. Perhaps they changed internally to always merge, even for fast-forward?

PR #823, opened 2017-01-17
MAINT: Convert tests to use simple_fixture where appropriate

adler-j:
Booring, but could actually use a proper readhthrough since it was lots of manual copy-paste, misstakes could have happened.

### DIFF
--- a/odl/test/deform/linearized_deform_test.py
+++ b/odl/test/deform/linearized_deform_test.py
@@ -27,24 +27,13 @@ import pytest
 
 import odl
 from odl.deform import LinDeformFixedTempl, LinDeformFixedDisp
-from odl.util.testutils import almost_equal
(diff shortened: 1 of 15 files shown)


odl/test/deform/linearized_deform_test.py
```diff
@@ -54,7 +43,7 @@ def space(request, ndim, interp, dtype, fn_impl):
     Generates example spaces with various implementations, dimensions, dtypes
     and interpolations.
     """
-    if dtype not in odl.FN_IMPLS[fn_impl].available_dtypes():
+    if np.dtype(dtype) not in odl.FN_IMPLS[fn_impl].available_dtypes():
```
adler-j: This was an old misstake that caused these tests to be skipped.

adler-j:
Merge after CI

PR #826, opened 2017-01-17
Issue 822  ray trafo impl default

adler-j:
I'm keeping with the theme of the day of removing code, i.m.o. this is much cleaner. An issue is that users are not given sufficient information that they *REALLY* want to install ASTRA.

### DIFF
--- a/odl/tomo/operators/ray_trafo.py
+++ b/odl/tomo/operators/ray_trafo.py
@@ -30,7 +30,7 @@ from odl.operator import Operator
 from odl.space import FunctionSpace
 from odl.tomo.geometry import Geometry, Parallel2dGeometry
 from odl.tomo.backends import (
-    ASTRA_AVAILABLE, ASTRA_CUDA_AVAILABLE,
+    ASTRA_AVAILABLE, ASTRA_CUDA_AVAILABLE, SCIKIT_IMAGE_AVAILABLE,
     astra_cpu_forward_projector, astra_cpu_back_projector,
     AstraCudaProjectorImpl, AstraCudaBackProjectorImpl,
     scikit_radon_forward, scikit_radon_back_projector)
@@ -47,7 +47,7 @@ class RayTransform(Operator):
 
     """Discrete Ray transform between L^p spaces."""
 
-    def __init__(self, discr_domain, geometry, impl='astra_cpu', **kwargs):
+    def __init__(self, discr_domain, geometry, **kwargs):
         """Initialize a new instance.
 
         Parameters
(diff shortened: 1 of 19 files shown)


odl/tomo/operators/ray_trafo.py
```diff
@@ -57,18 +57,19 @@ def __init__(self, discr_domain, geometry, impl='astra_cpu', **kwargs):
...
+        ----------------
+        impl : {`None`, 'astra_cuda', 'astra_cpu', 'scikit'}, optional
             Implementation back-end for the transform. Supported back-ends:
-            'astra_cpu': ASTRA toolbox using CPU, only 2D
-            'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
-            'scikit': scikit-image, only 2D parallel with square domain
+            * 'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
+            * 'astra_cpu': ASTRA toolbox using CPU, only 2D
+            * 'scikit': scikit-image, only 2D parallel with square domain
+            If `None` is given, prefers the fastest available backend.
```
kohr-h: \`\`None\`\`
the fastest available back-end is used

odl/tomo/operators/ray_trafo.py
```diff
@@ -246,17 +260,18 @@ def __init__(self, discr_range, geometry, impl='astra_cpu', **kwargs):
...
+        Other Parameters
+        ----------------
         impl : {'astra_cpu', 'astra_cuda', 'scikit'}, optional
             Implementation back-end for the transform. Supported back-ends:
-            'astra_cpu': ASTRA toolbox using CPU, only 2D
-            'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
-            'scikit': scikit-image, only 2D parallel with square domain
+            * 'astra_cuda': ASTRA toolbox, using CUDA, 2D or 3D
+            * 'astra_cpu': ASTRA toolbox using CPU, only 2D
+            * 'scikit': scikit-image, only 2D parallel with square domain
```
kohr-h: put strings in double backticks

adler-j:
> Small stuff, easy to fix

Any comments on ASTRA? Should we write a few more words in the readme perhaps?

adler-j:
Rebased from master, waiting for CI.

kohr-h:
> Any comments on ASTRA? Should we write a few more words in the readme perhaps?

Hm, perhaps something like "By default, the fastest available back-end is chosen." wouldn't hurt.

adler-j:
> Hm, perhaps something like "By default, the fastest available back-end is chosen." wouldn't hurt.
Not sure if it makes sense to put it in the front page readme.

Simply feels weird to have comments like that everywhere, yet it will probably be the first question people have "ODL is slow, what do?" Perhaps we leave it as is now and think a bit about it?

kohr-h:
> Simply feels weird to have comments like that everywhere, yet it will probably be the first question people have "ODL is slow, what do?" Perhaps we leave it as is now and think a bit about it?

Fine with me, go ahead.

kohr-h:
Also, users that don't do anything non-standard won't have a single back-end installed by default, so they rather ask "Tomography don't work, what do?".

Issue #440, opened 2016-06-03
Integration operator


ozanoktem:
We need to implement an integration functional which accepts a real value discretized function and evaluates, using some quadrature rule, its integral.  It should account for boundary behavior and support integration along axes in multidimensional functions. Such functionals are needed when defining many regularization functionals and currently we use the norm of the underlying space, which can lead to working results. One idea would be to use the [scipy.integrate](http://docs.scipy.org/doc/scipy/reference/tutorial/integrate.html) package. One may also check out the [Python wrapper for the Cubature algorithm](http://github.com/saullocastro/cubature) for multidimensional integration (cubature) of vector-valued functions over hypercubes.

adler-j:
One obvious comment is that the integral  needs to match the inner product and the likes, that is (on unweighted spaces):

```python
x.norm() == (a ** 2).integrate()
x.inner(y) == (x * y).integrate()
```

Another is that we will likely need the gradient of the integral (that is, the weights).

kohr-h:
> One obvious comment is that the integral needs to match the inner product and the likes

That actually depends. Perhaps you want to have the vanilla integral even if your reconstruction space is weighted? Or, in other words, sometimes you want to manually set the weighting in the integral functional.

adler-j:
> That actually depends. Perhaps you want to have the vanilla integral even if your reconstruction space is weighted? Or, in other words, sometimes you want to manually set the weighting in the integral functional.

Well, if the integral is a property of the vector, then it should match, if it is a freestanding functional, then we can ofc give options. Anyway, my comment was specifically on *unweighted* spaces.

kohr-h:
> One idea would be to use the scipy.integrate package.

That one is intended for integration of a continuously given function where you can freely select the quadrature points. Possible in principle, but arbitrarily slow.



> One may also check out the Python wrapper for the Cubature algorithm for multidimensional integration (cubature) of vector-valued functions over hypercubes.

Same here, although it may be faster by a more sophisticated handling of the integrand function. But why would we want to depend on such a package?

kohr-h:
> We need to implement an integration functional which accepts a real value discretized function and evaluates, using some quadrature rule, its integral.

If it's already discretized (in the way we understand now), there is no choice of quadrature rule.

ozanoktem:
[...]

PR #834, opened 2017-01-19
ENH: Add integration functionality to spaces

adler-j:
Adds integration operation to spaces

TODO:

- [ ] Optimize for lp discr (does not need to come in this branch)
- [ ] Add tests


ozanoktem:
Has this anything to do with #440 ?

adler-j:
Yes, the branch is named "issue-440__...".

kohr-h:
This would be nice to have. Generally I would consider putting this on the `Weighting` class since it also has to make use of weightings to be correct in weighted spaces. That will also give you the `DiscreteLp` implementation for free.

PR #840, opened 2017-01-20
Kaczmarzs method and __hash__

adler-j:
Closes an old friend.

Also adresses #843, adding `__hash__` to spaces.

Todo:

- [x] Add hash test to grid
- [x] Add hash test to partition
- [x] Add hash test to function space

### DIFF
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -465,6 +465,10 @@ class RectPartition(object):
         # Optimized version for exact equality
         return self.set == other.set and self.grid == other.grid
 
+    def __hash__(self):
+        """Return ``hash(self)``."""
+        return hash(type(self)) ^ hash(self.set) ^ hash(self.grid)
+
     def __ne__(self, other):
         """Return ``self != other``."""
         return not (self == other)
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -463,7 +463,10 @@ class RectPartition(object):
             return True
 
         # Optimized version for exact equality
-        return self.set == other.set and self.grid == other.grid
+        return (isinstance(other, type(self)) and
+                isinstance(self, type(other)) and
+                self.set == other.set and
+                self.grid == other.grid)
 
     def __hash__(self):
         """Return ``hash(self)``."""
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -467,7 +467,7 @@ class RectPartition(object):
 
     def __hash__(self):
         """Return ``hash(self)``."""
-        return hash(type(self)) ^ hash(self.set) ^ hash(self.grid)
+        return hash((type(self), self.set, self.grid))
 
     def __ne__(self, other):
         """Return ``self != other``."""
--- a/odl/set/domain.py
+++ b/odl/set/domain.py
@@ -225,7 +225,8 @@ class IntervalProd(Set):
 
     def __hash__(self):
         """Return ``hash(self)``."""
-        return hash(type(self)) ^ hash(self.min_pt) ^ hash(self.max_pt)
+        return (hash(type(self)) ^
+                hash(list(self.min_pt)) ^ hash(list(self.max_pt)))
 
     def approx_contains(self, point, atol):
         """Return ``True`` if ``point`` is "almost" contained in this set.
--- a/odl/set/domain.py
+++ b/odl/set/domain.py
@@ -225,8 +225,7 @@ class IntervalProd(Set):
 
     def __hash__(self):
         """Return ``hash(self)``."""
-        return (hash(type(self)) ^
-                hash(list(self.min_pt)) ^ hash(list(self.max_pt)))
+        return hash((type(self), tuple(self.min_pt), tuple(self.max_pt)))
 
     def approx_contains(self, point, atol):
         """Return ``True`` if ``point`` is "almost" contained in this set.
--- a/odl/set/domain.py
+++ b/odl/set/domain.py
@@ -223,6 +223,10 @@ class IntervalProd(Set):
         return (np.all(self.min_pt == other.min_pt) and
                 np.all(self.max_pt == other.max_pt))
 
+    def __hash__(self):
+        """Return ``hash(self)``."""
+        return hash(type(self)) ^ hash(self.min_pt) ^ hash(self.max_pt)
+
     def approx_contains(self, point, atol):
         """Return ``True`` if ``point`` is "almost" contained in this set.
 
--- a/odl/solvers/iterative/iterative.py
+++ b/odl/solvers/iterative/iterative.py
@@ -43,9 +43,7 @@ def landweber(op, x, rhs, niter=1, omega=1, projection=None, callback=None):
     Parameters
     ----------
     op : `Operator`
-        Operator in the inverse problem. It must have a `Operator.derivative`
-        property, which returns a new operator which in turn has an
-        `Operator.adjoint` property, i.e. ``op.derivative(x).adjoint`` must be
+        Operator in the inverse problem. ``op.derivative(x).adjoint`` must be
         well-defined for ``x`` in the operator domain.
     x : ``op.domain`` element
         Element to which the result is written. Its initial value is
@@ -54,7 +52,7 @@ def landweber(op, x, rhs, niter=1, omega=1, projection=None, callback=None):
     rhs : ``op.range`` element
         Right-hand side of the equation defining the inverse problem.
     niter : int, optional
-        Maximum number of iterations.
+        Number of iterations.
     omega : positive float, optional
         Relaxation parameter in the iteration.
     projection : callable, optional
@@ -155,7 +153,7 @@ def conjugate_gradient(op, x, rhs, niter=1, callback=None):
     rhs : ``op.range`` element
         Right-hand side of the equation defining the inverse problem.
     niter : int, optional
-        Maximum number of iterations.
+        Number of iterations.
     callback : callable, optional
         Object executing code per iteration, e.g. plotting each iterate.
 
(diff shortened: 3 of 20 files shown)


adler-j:
Added `__hash__` which should solve the python 3 build error and also solves #843. Review ready.

odl/space/weighting.py
```diff
@@ -594,6 +603,10 @@ def __eq__(self, other):
         return (super().__eq__(other) and
                 self.array is getattr(other, 'array', None))
 
+    def __hash__(self):
+        """Return ``hash(self)``."""
+        return super().__hash__() ^ hash(self.array)
```
kohr-h: `ndarray`'s can't be hashed. But perhaps we can have this anyway to let the error propagate.

adler-j: Nice catch, trying another way.

odl/set/domain.py
```diff
@@ -224,7 +224,8 @@ def __eq__(self, other):
 
     def __hash__(self):
         """Return ``hash(self)``."""
-        return hash(type(self)) ^ hash(self.min_pt) ^ hash(self.max_pt)
+        return (hash(type(self)) ^
+                hash(list(self.min_pt)) ^ hash(list(self.max_pt)))
```
kohr-h: lists are not hashable

kohr-h: `tostring` would be the way to go here, too

odl/space/weighting.py
adler-j: Fixed now

odl/set/domain.py
adler-j: Build system noticed that as well, went with tuple, but perhaps should do tostring. Felt tuple is more readable however.

odl/solvers/iterative/iterative.py
```diff
@@ -404,6 +394,129 @@ def gauss_newton(op, x, rhs, niter=1, zero_seq=exp_zero_seq(2.0),
...
+                        ''.format(x, domain))
+
+    if len(ops) != len(rhs):
+        raise ValueError('`number of `ops` {} does not match number of '
+                         '`rhs` {}'.format(len(ops), len(rhs)))
+
+    omega = normalized_scalar_param_list(omega, len(ops), param_conv=float)
+
+    # Reusable elements in the range, one per type of space
+    ranges = [opi.range for opi in ops]
```
adler-j: This type of optimization could likely be done in several places.

adler-j:
Ready for another review

adler-j:
Fixed the outstanding issues.

odl/discr/partition.py
```diff
@@ -451,7 +451,14 @@ def __eq__(self, other):
             return True
 
         # Optimized version for exact equality
-        return self.set == other.set and self.grid == other.grid
+        return (isinstance(other, type(self)) and
```
adler-j: This was an old bug where we did not properly test that other had the required properties, which gave an exception. We didn't find ti since we lacked tests (we now do).

Anyway, can someone remind me why we use the pattern:

```
isinstance(other, type(self)) and isinstance(self, type(other))
```

instead of the much more simple

```
type(self) == type(other)
```

odl/test/set/domain_test.py
```diff
@@ -179,7 +179,22 @@ def test_element():
...
+    assert x == y
+    assert not x != y
+    assert hash(x) == hash(y)
+
+
+def _test_neq(x, y):
+    """Test non-equality of x and y."""
+    assert x != y
+    assert not x == y
+    assert hash(x) != hash(y)
```
adler-j: This is, strictly speaking, not a requirement. But we didn't get any issues in the tests and it should hold "in general", so I'll keep it.

odl/discr/partition.py
kohr-h: Main reason was that it was flagged by Pylint as a code smell - and unidiomatic way of doing things. The point is not that it's unidiomatic regarding what we want to achieve (checking equivalence of types) but rather that the fact of enforcing an exact type is deemed unidiomatic by the linter.
However, since we need exactly that here, we can as well go with the simpler code and ignore the flag.

odl/test/set/domain_test.py
kohr-h: I hope hash collisions are unlikely enough in Python :-)
Anyway, if the purpose of this test is to make sure there's no obvious mistake in the hashing method, it's fine.

odl/solvers/iterative/iterative.py
```diff
@@ -404,6 +394,129 @@ def gauss_newton(op, x, rhs, niter=1, zero_seq=exp_zero_seq(2.0),
...
+    coincides with the Landweber method for a single operator.
+
+    Parameters
+    ----------
+    ops : sequence of `Operator`'s
+        Operators in the inverse problem. They must have a
+        `Operator.derivative` property, which returns a new operator which in
+        turn has an `Operator.adjoint` property, i.e.
+        ``op[i].derivative(x).adjoint`` must be well-defined for ``x`` in the
+        operator domain and for all ``i``.
```
kohr-h: Comes across a bit convoluted text-wise. I think the part after "i.e." makes perfectly clear what the requirements are and is much more readable; the first part is not needed.

odl/solvers/iterative/iterative.py
```diff
@@ -404,6 +394,129 @@ def gauss_newton(op, x, rhs, niter=1, zero_seq=exp_zero_seq(2.0),
...
+    x : ``op.domain`` element
+        Element to which the result is written. Its initial value is
+        used as starting point of the iteration, and its values are
+        updated in each iteration step.
+    rhs : sequence of ``ops[i].range`` elements
+        Right-hand side of the equation defining the inverse problem.
+    niter : int, optional
+        Maximum number of iterations
```
kohr-h: dot

odl/solvers/iterative/iterative.py
```diff
@@ -404,6 +394,129 @@ def gauss_newton(op, x, rhs, niter=1, zero_seq=exp_zero_seq(2.0),
...
+    omega : positive float or sequence of positive float, optional
```
kohr-h: sequence of positive float**s**

odl/solvers/iterative/iterative.py
```diff
@@ -404,6 +394,129 @@ def gauss_newton(op, x, rhs, niter=1, zero_seq=exp_zero_seq(2.0),
...
+
+    for a (Frechet-) differentiable operator
+    :math:`\mathcal{A}: \mathcal{X} \\to \mathcal{Y}` between Hilbert
+    spaces :math:`\mathcal{X}` and :math:`\mathcal{Y}`. The method
+    starts from an initial guess :math:`x_0` and uses the
+    iteration
+
+    .. math::
+        x_{k+1} = x_k - \omega_{[i]} \ \partial \mathcal{A}_{[i]}(x)^*
+                                 (\mathcal{A}_{[i]}(x_k) - y_{[i]}),
```
kohr-h: Something's not quite right here.

If `k` should be the "outer" iteration number, then I don't think it can be written this way. Then one needs inner iterates `x_{k, i+1} = x_{k,i} - ...` and set `x_{k,0} = x_k` and `x_{k+1} = x_{k, n}`.

Alternatively, count all iterations by writing `x_{i+1} = x_i - ...` basically replacing `k` with `[i]` on the RHS. Then it would be good to mention that the outer iteration number is given as `k = \lfloor i / n \rfloor`.

odl/solvers/iterative/iterative.py
```diff
@@ -404,6 +394,129 @@ def gauss_newton(op, x, rhs, niter=1, zero_seq=exp_zero_seq(2.0),
...
+    :math:`0 < \omega_{[i]} < 2/\\lVert \mathcal{A}_{[i]}^2\\rVert` guarantees
+    convergence, where :math:`\|\mathcal{A}_{[i]}\|` stands for the
+    operator norm of :math:`\mathcal{A}_{[i]}`.
+
+    This implementation uses a minimum amount of memory copies by
+    applying re-usable temporaries and in-place evaluation.
+
+    The method is also described in a
+    `Wikipedia article
+    <https://en.wikipedia.org/wiki/Kaczmarz_method>`_.
```
kohr-h: Here I'd definitely add a reference to one of the Natterer books, they do a very good job describing the method.

odl/solvers/iterative/iterative.py
kohr-h: Yes, now that we can hash and pick the unique ones :-)

kohr-h:
Oh crap, the submodule pointer for sphinxext now points into void, #846 needs to go in.

odl/discr/partition.py
adler-j: I made an issue, #848 where we can discuss this further.

odl/solvers/iterative/iterative.py
adler-j: Well there are two ways of writing the iteration, and this "swallows" the outer iterate into the `[i]` syntax. I personally feel that it is more clear this way, since we only get a single index that increases, it also makes it clear when the callbacks will be called etc.

This is your second idea.

Anyway, I'll add a comment about "other" iterates.

Will fix the argument to the derivative.

odl/solvers/iterative/iterative.py
adler-j: Will do.

adler-j:
Fixed comments and rebased from master.

adler-j:
Rebased again. Merging.

MatthiasJE:
The terminology "Kaczmarz" might be misleading. The implementation here is explicit, while the "Kaczmarz" method is implicit as defined on https://en.wikipedia.org/wiki/Kaczmarz_method. This may not be obvious from Algorithm 1 and 2 but is clear for Algorithm 3. To me, this implementation is more like stochastic forward-backward splitting or so.

PR #841, opened 2017-01-20
MAINT: rename TensorGrid->RectGrid and remove RegularGrid

### LINKED ISSUE #670: Remove RegularGrid
We have a distinction `RegularGrid` vs `TensorGrid`, but no such distinction for `RectPartition`. I propose we scrap `RegularGrid` and rename `TensorGrid` to `RectGrid`. Any opinions?

kohr-h:
Closes #670.

Also some maintenance changes regarding the `ideg` and `inondeg` properties, now replaced by a single `nondegen_byaxis` property that returns a full boolean array.

### DIFF
--- a/odl/__init__.py
+++ b/odl/__init__.py
@@ -16,18 +16,16 @@
 # along with ODL.  If not, see <http://www.gnu.org/licenses/>.
 
 
-"""ODL is a functional analysis library with a focus on discretization.
+"""ODL (Operator Discretization Library)
 
-ODL suppors abstract sets, linear vector spaces defined on such
-and Operators/Functionals defined on these sets. It is intended
-to be used to write general code and faciliate code reuse.
+ODL is a Python library for fast prototyping focusing on (but not
+restricted to) inverse problems.
 """
 
 from __future__ import absolute_import
 
 __version__ = '0.5.4.dev0'
-__all__ = ('diagnostics', 'discr', 'operator', 'set', 'space', 'solvers',
-           'tomo', 'trafos', 'util', 'phantom', 'deform', 'ufunc_ops')
+__all__ = ()
 
 # Propagate names defined in __all__ of all submodules into the top-level
 # module
--- a/odl/__init__.py
+++ b/odl/__init__.py
@@ -25,24 +25,25 @@ restricted to) inverse problems.
 from __future__ import absolute_import
 
 __version__ = '0.5.4.dev0'
-__all__ = ()
+__all__ = ('diagnostics', 'discr', 'operator', 'set', 'space', 'solvers',
+           'tomo', 'trafos', 'util', 'phantom', 'deform', 'ufunc_ops')
 
 # Propagate names defined in __all__ of all submodules into the top-level
 # module
-from . import diagnostics
-
-from .discr import *
-__all__ += discr.__all__
-
-from .operator import *
-__all__ += operator.__all__
-
 from .set import *
 __all__ += set.__all__
+
+# operator must come before space because npy_ntuples imports Operator
+from .operator import *
+__all__ += operator.__all__
 
 from .space import *
 __all__ += space.__all__
 
+from .discr import *
+__all__ += discr.__all__
+
+from . import diagnostics
 from . import solvers
 from . import trafos
 from . import tomo
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -106,8 +106,6 @@ class RectGrid(Set):
             [1.0, 2.0, 5.0],
             [-2.0, 1.5, 2.0]
         )
-        >>> print(g)
-        grid [1.0, 2.0, 5.0] x [-2.0, 1.5, 2.0]
         >>> g.ndim  # number of axes
         2
         >>> g.shape  # points per axis
@@ -208,7 +206,7 @@ class RectGrid(Set):
         # Uniformity, setting True in degenerate axes
         diffs = [np.diff(v) for v in self.coord_vectors]
         self.__is_uniform_byaxis = tuple(
-            True if diff.size == 0 else np.allclose(diff, diff[0])
+            (diff.size == 0) or np.allclose(diff, diff[0])
             for diff in diffs)
 
     # Attributes
@@ -392,6 +390,11 @@ class RectGrid(Set):
     def stride(self):
         """Step per axis between neighboring points of a uniform grid.
 
+        Raises
+        ------
+        NotImplementedError
+            if the grid is not uniform
+
         Examples
         --------
         >>> rg = uniform_grid([-1.5, -1], [-0.5, 3], (2, 3))
@@ -570,10 +573,12 @@ class RectGrid(Set):
         # Optimization for some common cases
         if other is self:
             return True
-        if not (isinstance(other, RectGrid) and
-                np.all(self.shape <= other.shape) and
-                np.all(self.min_pt >= other.min_pt - atol) and
-                np.all(self.max_pt <= other.max_pt + atol)):
+        if not isinstance(other, RectGrid):
+            return False
+        if not all(self.shape[i] <= other.shape[i] and
+                   self.min_pt[i] >= other.min_pt[i] - atol and
+                   self.max_pt[i] <= other.max_pt[i] + atol
+                   for i in range(self.ndim)):
             return False
 
         if self.is_uniform and other.is_uniform:
@@ -599,7 +604,7 @@ class RectGrid(Set):
                 # vec_s. If there is no almost zero entry in each row,
                 # return False.
                 vec_o_mg, vec_s_mg = sparse_meshgrid(vec_o, vec_s)
-                if not np.all(np.any(np.abs(vec_s_mg - vec_o_mg) <= atol,
+                if not np.all(np.any(np.isclose(vec_s_mg, vec_o_mg, atol=atol),
                                      axis=0)):
                     return False
             return True
@@ -701,7 +706,7 @@ class RectGrid(Set):
         )
 
         """
-        nondegen_indcs = np.where(self.nondegen_byaxis)[0]
+        nondegen_indcs = np.flatnonzero(self.nondegen_byaxis)
         coord_vecs = [self.coord_vectors[axis] for axis in nondegen_indcs]
         return RectGrid(*coord_vecs)
 
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -31,12 +31,11 @@ import numpy as np
 
 from odl.set import Set, IntervalProd
 from odl.util import (
-    normalized_index_expression, array1d_repr, array1d_str, safe_int_conv,
-    signature_string, indent_rows)
+    normalized_index_expression, normalized_scalar_param_list, safe_int_conv,
+    array1d_repr, array1d_str, signature_string, indent_rows)
 
 
-__all__ = ('RectGrid', 'uniform_grid', 'uniform_sampling_fromintv',
-           'uniform_sampling')
+__all__ = ('RectGrid', 'uniform_grid', 'uniform_grid_fromintv')
 
 
 def sparse_meshgrid(*x):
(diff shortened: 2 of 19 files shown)


odl/discr/partition.py
```diff
@@ -480,19 +492,7 @@ def __getitem__(self, indices):
...
-        vectors):
-
-        >>> part[0, 0, 0, 0]
-        nonuniform_partition(
-            [-1.0],
-            [2.0],
-            [5.0],
-            [2.0],
-            min_pt=[-1.0, 1.0, 4.0, 2.0], max_pt=[-0.5, 3.0, 5.0, 3.0]
-        )
```
kohr-h: This now became a uniform partition and printed out an enormous line, had to remove it.

odl/test/tomo/backends/astra_setup_test.py
```diff
@@ -168,12 +168,6 @@ def test_astra_projection_geometry():
     apart = odl.uniform_partition(0, 2 * np.pi, 5)
     dpart = odl.uniform_partition(-40, 40, 10)
 
-    # motion sampling grid, detector sampling grid but not RegularGrid
-    dpart_0 = odl.RectPartition(odl.IntervalProd(0, 0), odl.TensorGrid([0]))
-    geom_p2d = odl.tomo.Parallel2dGeometry(apart, dpart=dpart_0)
-    with pytest.raises(ValueError):
-        odl.tomo.astra_projection_geometry(geom_p2d)
```
kohr-h: This check no longer made sense since the grid was in fact uniform, hence the error condition was gone.

odl/discr/grid.py
```diff
@@ -129,16 +119,58 @@ def __init__(self, *coord_vectors):
...
+            v_1 = (-1, 0, 2),\ v_2 = (0, 1)
+
+        the corresponding rectilinear grid :math:`G` is the set of all
+        2d points whose first component is from :math:`v_1` and the
+        second component from :math:`v_2`:
+
+        .. math::
+            G = \{(-1, 0), (-1, 1), (0, 0), (0, 1), (2, 0), (2, 1)\}
+
+        Here is a graphical representation::
```
adler-j: Does this typeset well?

odl/discr/grid.py
```diff
@@ -327,12 +377,43 @@ def max(self, **kwargs):
...
+        Examples
+        --------
+        >>> rg = uniform_grid([-1.5, -1], [-0.5, 3], (2, 3))
+        >>> rg.stride
+        array([ 1.,  2.])
+        """
+        if not self.is_uniform:
+            raise NotImplementedError('`stride` not defined for non-uniform '
+                                      'grid')
+        strd = np.zeros(self.ndim)
```
adler-j: might as well write out stride instead of strd

odl/discr/grid.py
```diff
@@ -327,12 +377,43 @@ def max(self, **kwargs):
...
+        >>> rg = uniform_grid([-1.5, -1], [-0.5, 3], (2, 3))
+        >>> rg.mid_pt
+        array([-1.,  1.])
+        """
+        return (self.max_pt + self.min_pt) / 2
+
+    @property
+    def stride(self):
+        """Step per axis between neighboring points of a uniform grid.
+
```
adler-j: raises ...

odl/discr/grid.py
```diff
@@ -451,54 +532,85 @@ def __contains__(self, other):
...
         False
-        >>> g_sub.is_subgrid(g, atol=0.15)
+        >>> rg_sub.is_subgrid(rg, atol=0.02)
         True
         """
         # Optimization for some common cases
         if other is self:
             return True
-        if not (isinstance(other, TensorGrid) and
+        if not (isinstance(other, RectGrid) and
```
adler-j: Why is this if statement still there?

odl/discr/grid.py
```diff
@@ -451,54 +532,85 @@ def __contains__(self, other):
...
-        return True
+        else:
+            # Array version of the fuzzy subgrid test, about 3 times faster
+            # than the loop version.
+            for vec_o, vec_s in zip(other.coord_vectors, self.coord_vectors):
+                # Create array of differences of all entries in vec_o and
+                # vec_s. If there is no almost zero entry in each row,
+                # return False.
+                vec_o_mg, vec_s_mg = sparse_meshgrid(vec_o, vec_s)
+                if not np.all(np.any(np.abs(vec_s_mg - vec_o_mg) <= atol,
```
adler-j: np allclose?

odl/discr/grid.py
```diff
@@ -564,17 +689,22 @@ def squeeze(self):
...
-        TensorGrid([0.0, 1.0], [-1.0, 0.0, 2.0])
+        RectGrid(
+            [0.0, 1.0],
+            [-1.0, 0.0, 2.0]
+        )
+
         """
-        coord_vecs = [self.coord_vectors[axis] for axis in self.inondeg]
-        return TensorGrid(*coord_vecs)
+        nondegen_indcs = np.where(self.nondegen_byaxis)[0]
```
adler-j: use np.flatnonzero

adler-j: Also why do we only use the non-degenerate axes?

odl/discr/grid.py
```diff
@@ -802,346 +947,104 @@ def __array__(self, dtype=None):
...
+            inner_str = signature_string(posargs, [])
+            return '{}({})'.format(constructor, inner_str)
+        else:
+            constructor = self.__class__.__name__
+            posargs = [array1d_repr(v) for v in self.coord_vectors]
+            inner_str = signature_string(posargs, [], sep=[',\n', ', ', ', '],
+                                         mod=['!s', ''])
+            return '{}(\n{}\n)'.format(constructor, indent_rows(inner_str))
 
     def __str__(self):
```
adler-j: I frankly prefer the repr over this. It is more readable.

odl/discr/grid.py
```diff
@@ -802,346 +947,104 @@ def __array__(self, dtype=None):
...
-            return minmax_contained and checkpt_contained
-
-        elif isinstance(other, TensorGrid):
-            # other is a TensorGrid, we need to fall back to full check
-            self_tg = TensorGrid(*self.coord_vectors)
-            return self_tg.is_subgrid(other)
-
-    def insert(self, index, other):
-        """Insert another regular grid before the given index.
+def uniform_grid(min_pt, max_pt, shape):
```
adler-j: Good addition.

odl/discr/grid.py
```diff
@@ -802,346 +947,104 @@ def __array__(self, dtype=None):
...
-        """Return ``repr(self)``."""
-        inner_str = '{}, {}, {}'.format(list(self.min_pt), list(self.max_pt),
-                                        tuple(self.shape))
-        return '{}({})'.format(self.__class__.__name__, inner_str)
+    coord_vecs = [np.linspace(mi, ma, num, endpoint=True, dtype=np.float64)
+                  for mi, ma, num in zip(min_pt, max_pt, shape)]
+    return RectGrid(*coord_vecs)
 
 
 def uniform_sampling_fromintv(intv_prod, shape, nodes_on_bdry=True):
```
adler-j: uniform_grid_fromintv now I guess

odl/discr/grid.py
```diff
@@ -1264,7 +1168,7 @@ def uniform_sampling_fromintv(intv_prod, shape, nodes_on_bdry=True):
             gmin.append(xmin + (xmax - xmin) / (2 * n))
             gmax.append(xmax - (xmax - xmin) / (2 * n))
 
-    return RegularGrid(gmin, gmax, shape)
+    return uniform_grid(gmin, gmax, shape)
 
 
 def uniform_sampling(min_pt, max_pt, shape, nodes_on_bdry=True):
```
adler-j: How does this compare to uniform_grid?

odl/discr/partition.py
adler-j: I dont get it, why did it change and how does it look now?

odl/test/tomo/backends/astra_setup_test.py
adler-j: Well, perhaps we should change the grid then?

odl/discr/grid.py
kohr-h: I think it looks decent. It's treated as code, so it uses a monospace font.

odl/discr/partition.py
kohr-h: Now by definition, any axis with two cells or less is uniform, which is kind of hard to argue against. The printed return value was roughly
```
uniform_partition([-1.0, 1.0, 4.0, 2.0], [-1.0, 1.0, 4.0, 2.0], (1, 1, 1, 1), nodes_on_bdry=((True, False), False, (False, True), (True, False))
```
modulo error in the booleans.

odl/test/tomo/backends/astra_setup_test.py
kohr-h: Too lazy :-)
Okay I'll do it. I had to get it out because gates close at 7, didn't want to be locked in.

odl/discr/grid.py
kohr-h: Hesitant to shadow the function name, feels wrong.

odl/discr/grid.py
kohr-h: To return `False` for wrong type instead of raising an `AttributeError`?

odl/discr/grid.py
adler-j: The function name isn't shadowed by it since it wasn't available in the first place? The only way to access it is via `self.stride`

odl/discr/grid.py
adler-j: well any non rect grid should be false, I was thinking about the following lines but perhaps I missed the logic (hard to read with changelogs)

odl/discr/grid.py
kohr-h: Oh well yeah, that was some old code I just copied.

odl/discr/grid.py
kohr-h: Okay, "shadow" was the wrong term, "simultaneously highlighted by the editor" would be more precise.

odl/discr/grid.py
kohr-h: It would probably be easier to read as
```python
if not isinstance(other, RectGrid) or
not ...
```
It's easy to mismatch parens I guess.

odl/discr/grid.py
kohr-h: We're in the `sqeeze` method here, that's kind of its purpose :-)

odl/discr/grid.py
kohr-h: Hm, I thought I had started an attempt at erasing this earlier but hit some resistance. When putting together this PR it tickled again in the code-deleting finger, but I resisted the temptation.
But I'm happy to ditch this function.

odl/discr/grid.py
```diff
@@ -802,346 +947,104 @@ def __array__(self, dtype=None):
...
     def __str__(self):
         """Return ``str(self)``."""
-        grid_str = ' x '.join(array1d_str(vec) for vec in self.coord_vectors)
+        posargs = [array1d_repr(v) for v in self.coord_vectors]
+        grid_str = signature_string(posargs, [], sep=[' x ', ', ', ', '],
+                                    mod=['!s', ''])
         return 'grid ' + grid_str
 
 
-class RegularGrid(TensorGrid):
```
kohr-h: Meh, not really..

odl/discr/grid.py
kohr-h: Better name eh? I'll change it.

odl/discr/grid.py
adler-j: Oh god damnit xD

odl/discr/grid.py
kohr-h: This uses the interval product min/max, while the other uses the grid min/max directly. Subtle difference.

odl/discr/partition.py
adler-j: I see why that was removed.

odl/discr/grid.py
adler-j: Nice then!

odl/discr/grid.py
adler-j: I guess editor confusion is also a reason, stay with this then.

odl/discr/grid.py
adler-j: Well

```
if not isinstance(other, RectGrid):
    return False

# other logic here
```

is our standard pattern.

odl/discr/grid.py
adler-j: Great!

odl/discr/grid.py
adler-j: Oh god you're right, but with nodes_on_bdry=True they are equal, no?

kohr-h: Oh yes, of course. So perhaps one of them can go. I'll look into it.

odl/discr/grid.py
kohr-h: No, `allclose` is not suitable since the two vectors are broadcast and form a matrix (like a distance matrix). Then it is checked that in each column there is at least one entry close enough to zero.
AFAICS the only thing I can do is replace the boolean abs expression with `np.isclose`.

kohr-h:
@adler-j you may want to look at 86acb74 and check if that's okay. I find it nice because it's now the package that solely defines the version.

odl/__init__.py
```diff
@@ -16,18 +16,16 @@
...
+ODL is a Python library for fast prototyping focusing on (but not
+restricted to) inverse problems.
 """
 
 from __future__ import absolute_import
 
 __version__ = '0.5.4.dev0'
-__all__ = ('diagnostics', 'discr', 'operator', 'set', 'space', 'solvers',
-           'tomo', 'trafos', 'util', 'phantom', 'deform', 'ufunc_ops')
+__all__ = ()
```
adler-j: Are we sure all of them should be removed? How about stuff like util that is not default imported?

odl/discr/grid.py
```diff
@@ -72,34 +73,19 @@ def sparse_meshgrid(*x):
...
-
-    The storage need for a tensor grid is only the sum of the lengths
+    The storage need for a rectilinear grid is only the sum of the lengths
     of the coordinate vectors, while the total number of points is
-    the product of these lengths. This class makes use of this
-    sparse storage.
+    the product of these lengths. This class makes use of that
+    sparse storage scheme.
+
+    See ``Notes`` for details.
```
adler-j: Where did notes go?

adler-j: Edit: Into Init, i'd keep it at class level.

odl/discr/grid.py
```diff
@@ -169,10 +198,16 @@ def __init__(self, *coord_vectors):
...
-                                  if len(vecs[i]) != 1])
+
+        # Non-degenerate axes
+        self.__nondegen_byaxis = np.fromiter(
+            (len(v) > 1 for v in self.coord_vectors), dtype=bool)
+
+        # Uniformity, setting True in degenerate axes
+        diffs = [np.diff(v) for v in self.coord_vectors]
+        self.__is_uniform_byaxis = tuple(
+            True if diff.size == 0 else np.allclose(diff, diff[0])
```
adler-j: `diff.size == 0 or np.allclose(diff, diff[0])`

odl/discr/grid.py
```diff
@@ -327,12 +374,49 @@ def max(self, **kwargs):
...
+        ------
+        NotImplementedError
+            if the grid is not uniform
+
+        Examples
+        --------
+        >>> rg = uniform_grid([-1.5, -1], [-0.5, 3], (2, 3))
+        >>> rg.stride
+        array([ 1.,  2.])
+
```
adler-j: remove blankline

odl/discr/grid.py
```diff
@@ -451,54 +535,86 @@ def __contains__(self, other):
...
         # Optimization for some common cases
         if other is self:
             return True
-        if not (isinstance(other, TensorGrid) and
-                np.all(self.shape <= other.shape) and
-                np.all(self.min_pt >= other.min_pt - atol) and
-                np.all(self.max_pt <= other.max_pt + atol)):
+        if not isinstance(other, RectGrid):
+            return False
+        if not(np.all(self.shape <= other.shape) and
```
adler-j: little reason to use np.all for length ~3 arrays, more readable with simply all (and probably faster for this small case)

odl/discr/partition.py
```diff
@@ -629,10 +629,11 @@ def squeeze(self):
 
         See Also
         --------
-        TensorGrid.squeeze
+        RectGrid.squeeze
         IntervalProd.squeeze
         """
-        newset = self.set[self.grid.inondeg]
+        nondegen_indcs = np.where(self.grid.nondegen_byaxis)[0]
```
adler-j: np.flatnonzero to skip the [0]

odl/set/domain.py
```diff
@@ -102,7 +101,7 @@ def ndim(self):
     @property
     def true_ndim(self):
         """Number of non-degenerate (positive-length) intervals."""
-        return len(self.inondeg)
+        return sum(self.nondegen_byaxis.astype(int))
```
adler-j: `return np.count_nonzero(self.nondegen_byaxis)`

odl/set/domain.py
```diff
@@ -431,7 +432,7 @@ def measure(self, ndim=None):
         elif ndim > self.true_ndim:
             return 0.0
         else:
-            return np.prod((self.max_pt - self.min_pt)[self.inondeg])
+            return np.prod(self.extent()[self.nondegen_byaxis])
```
adler-j: Remind me, why is extent not a property?

odl/__init__.py
kohr-h: As I understand it, the imports further down put all the subpackages in the module dictionary, so that's not the purpose of adding them to `__all__` anyway. The `__all__` here is only effective if someone does `from odl import *`, and then I guess we don't want the subpackages *themselves* to be included, only their `__all__` contents. Or do we?

odl/discr/grid.py
kohr-h: But then in IDEs you get class docstring -> Notes -> Parameters, which IMO is not the best possible order. As I've argued earlier, and that's of course debatable, I think parameters should come up as quickly as possible, with all the extra stuff afterwards. That's why I moved the notes into `__init__` after parameters and examples.

odl/set/domain.py
kohr-h: Probably because the thought was that for some grid or domain structures in the future, this might be costly to compute. If we're sure that it's always gonna be cheap, we can make it a property (but then what to do with `min` and `max` which are methods, too?).

odl/__init__.py
adler-j: Well if someone does `from odl import all` I guess they should get something called `util`? Otherwise they cant access everything. For subpackages that are included via `import *`, we don't need to import them specifically.

odl/discr/grid.py
adler-j: Still not convinced that we should change the logical structure of the doc in order to please a specific IDE typesetting, other IDEs may do this "properly" (i.e. distinguish between a class and its initializer in a better way than spyder). Very few will read this anyway.

adler-j: To continue on the above, this is mostly dev doc and for developers those notes are very important, more so than the init doc.

odl/set/domain.py
adler-j: `min` and `max` are methods in order to conform to the numpy interface, which allows these to be used with numpy in a duck-typing style. Otherwise we provide `min_pt` etc which do the same.

I'll propose a change.

adler-j: Issue: #847

odl/discr/grid.py
kohr-h: That may be true, but *how often* is that important compared to the parameter explanations? 

The math notes are important to understand what this represents, and that's usually when you read the doc for the first time. After that, you perhaps occasionally check the math but not often.

On the other hand, the parameters are important on a day-to-day basis when you use the function or class in practice. Of course the signature is printed first (usually, sometimes it doesn't work but that's nothing to count in), but when there are `kwargs` or if you need details about the parameters you need that section. And then at least I find it annoying if there are too many other things in the way that make me scroll every time I need info about the arguments.

So even from a dev perspective I feel that in the long run, parameters are more important than other details in the notes.



> other IDEs may do this "properly" (i.e. distinguish between a class and its initializer in a better way than spyder)

Not sure what a better way would look like, or how other IDEs handle this. To me it seems reasonable to show the class docstring and the initialization doc if a user asks for documentation on a class.
The alternative would be to let the user choose, with pop-up menus like (apparently) in PyCharm. I personally find that annoying.

odl/__init__.py
kohr-h: Hm, right. But that basically means that everything has to stay, since the same applies to the other subpackages. Currently you get everything from the subpackages' `__all__` but the other stuff is not accessible, at least not directly.

That also means that we have to be careful not to include any module or package into `__all__` that conflicts with another name somewhere in the library, or at least make sure that they're not imported into the same namespace.

odl/discr/grid.py
kohr-h: It's also wrong since the statement in parenthesis is already a boolean (compares tuples).

odl/discr/grid.py
kohr-h: Made a new issue #854.

PR #865, opened 2017-01-25
Issue 863  finite set

adler-j:
Adds three rather basic sets

### DIFF
--- a/odl/set/sets.py
+++ b/odl/set/sets.py
@@ -25,7 +25,7 @@ from odl.util.utility import with_metaclass
 
 __all__ = ('Set', 'EmptySet', 'UniversalSet', 'Field', 'Integers',
            'RealNumbers', 'ComplexNumbers', 'Strings', 'CartesianProduct',
-           'SetUnion', 'SetIntersection')
+           'SetUnion', 'SetIntersection', 'FiniteSet')
 
 
 class Set(with_metaclass(ABCMeta, object)):
@@ -650,12 +650,12 @@ class SetUnion(Set):
                      if not isinstance(set_, Set)]
             raise TypeError('{!r} not Set instance(s)'.format(wrong))
 
-        self._sets = tuple(sets)
+        self.__sets = tuple(sets)
 
     @property
     def sets(self):
         """The factors (sets) as a tuple."""
-        return self._sets
+        return self.__sets
 
     def __contains__(self, other):
         """Return ``other in self``.
@@ -704,6 +704,10 @@ class SetUnion(Set):
                 pass
         raise NotImplementedError('element not implemented for any of the '
                                   'subsets')
+
+    def __len__(self):
+        """Return ``len(self)``."""
+        return len(self.sets)
 
     def __getitem__(self, indcs):
         """Return ``self[indcs]``.
@@ -766,12 +770,12 @@ class SetIntersection(Set):
                      if not isinstance(set_, Set)]
             raise TypeError('{!r} not Set instance(s)'.format(wrong))
 
-        self._sets = tuple(sets)
+        self.__sets = tuple(sets)
 
     @property
     def sets(self):
         """The factors (sets) as a tuple."""
-        return self._sets
+        return self.__sets
 
     def __contains__(self, other):
         """Return ``other in self``.
@@ -821,6 +825,10 @@ class SetIntersection(Set):
                                           'of the subsets')
         return inp
 
+    def __len__(self):
+        """Return ``len(self)``."""
+        return len(self.sets)
+
     def __getitem__(self, indcs):
         """Return ``self[indcs]``.
 
@@ -849,6 +857,112 @@ class SetIntersection(Set):
         """
         sets_str = ', '.join(repr(set_) for set_ in self.sets)
         return '{}({})'.format(self.__class__.__name__, sets_str)
+
+
+class FiniteSet(Set):
+    """A set given by a finite tuple of elements.
+
+    Notes
+    -----
+    Mathematically, this is simply the set given via listing the members of the
+    set:
+
+    .. math::
+        X = \{x_1, x_2, \dots, x_n\}
+    """
+
+    def __init__(self, *elements):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        set1, ..., setN : `Set`
+            The elements in the set.
+
+        Examples
+        --------
+        >>> set = odl.FiniteSet(1, 'string')
+        """
+        self.__elements = tuple(elements)
+
+    @property
+    def elements(self):
+        """The elements as a tuple."""
+        return self.__elements
+
+    def __contains__(self, other):
+        """Return ``other in self``.
+
+        Returns
+        -------
+        contains : `bool`
+            `True` if `other` is an element in `elements`, `False` otherwise.
+
+        Examples
+        --------
+        >>> set = odl.FiniteSet(1, 'string')
+        >>> 1 in set
+        True
+        >>> 2 in set
+        False
+        >>> 'string' in set
+        True
+        """
+        return other in self.elements
+
+    def __eq__(self, other):
+        """Return ``self == other``.
+
+        Returns
+        -------
+        equals : `boolean`
+            `True` if `other` is a `SetUnion` instance, and
+            has the same subsets as this set, `False` otherwise.
+        """
+        return (type(self) == type(other) and
+                len(other) == len(self) and
+                all(eo == es for eo, es in zip(other.elements, self.elements)))
+
+    def element(self, inp=None):
+        """Create a new element.
+
+        First tries calling the first set, then the second, etc.
+
+        For more specific control, use set[i].element() to pick which subset to
+        use.
+        """
+        if inp in self.elements:
+            return inp
+        else:
+            return Set.element(self, inp)
+
+    def __getitem__(self, indcs):
+        """Return ``self[indcs]``.
+
+        Examples
+        --------
+
+        >>> set = odl.FiniteSet(1, 2, 3, 'string')
+        >>> set[:3]
+        FiniteSet(1, 2, 3)
+        >>> set[3]
+        'string'
+        """
+        if isinstance(indcs, slice):
+            return FiniteSet(*self.elements[indcs])
+        else:
+            return self.elements[indcs]
+
+    def __repr__(self):
+        """Return ``repr(self)``.
+
+        Examples
+        --------
+        >>> odl.FiniteSet(1, 'string')
+        FiniteSet(1, 'string')
+        """
+        elements_str = ', '.join(repr(el) for el in self.elements)
+        return '{}({})'.format(self.__class__.__name__, elements_str)
 
 
 if __name__ == '__main__':
--- a/odl/set/sets.py
+++ b/odl/set/sets.py
@@ -19,7 +19,7 @@ from abc import ABCMeta, abstractmethod
 from numbers import Integral, Real, Complex
 import numpy as np
 
-from odl.util import is_int_dtype, is_real_dtype, is_scalar_dtype
+from odl.util import is_int_dtype, is_real_dtype, is_scalar_dtype, unique
 from odl.util.utility import with_metaclass
 
 
@@ -518,7 +518,7 @@ class CartesianProduct(Set):
 
     @property
     def sets(self):
-        """Factors (sets) as a tuple."""
+        """The sets of this carthesian product as a tuple."""
         return self.__sets
 
     def __contains__(self, other):
@@ -553,9 +553,8 @@ class CartesianProduct(Set):
             has the same length as this Cartesian product and all sets
             with the same index are equal, ``False`` otherwise.
         """
-        return (isinstance(other, CartesianProduct) and
-                len(other) == len(self) and
-                all(so == ss for so, ss in zip(other.sets, self.sets)))
+        return (type(self) == type(other) and
+                self.sets == other.sets)
 
     def __hash__(self):
         """Return ``hash(self)``."""
@@ -597,12 +596,12 @@ class CartesianProduct(Set):
 
         Examples
         --------
-        >>> emp, univ = EmptySet(), UniversalSet()
-        >>> prod = CartesianProduct(emp, univ, univ, emp, emp)
-        >>> prod[2]
-        UniversalSet()
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> prod = odl.CartesianProduct(reals, complexnrs, complexnrs, reals)
+        >>> prod[1]
+        ComplexNumbers()
         >>> prod[2:4]
-        CartesianProduct(UniversalSet(), EmptySet())
+        CartesianProduct(ComplexNumbers(), RealNumbers())
         """
         if isinstance(indices, slice):
             return CartesianProduct(*self.sets[indices])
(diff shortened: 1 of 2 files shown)


odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+            has the same subsets as this set, `False` otherwise.
+        """
+        return (type(self) == type(other) and
+                len(other) == len(self) and
+                all(so == ss for so, ss in zip(other.sets, self.sets)))
+
+    def element(self, inp=None):
+        """Create a new element.
+
+        First tries calling the first set, then the second, etc.
```
adler-j: update

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+    def __len__(self):
+        """Return ``len(self)``."""
+        return len(self.sets)
+
+    def __getitem__(self, indcs):
+        """Return ``self[indcs]``.
+
+        Examples
+        --------
+        >>> emp, univ = odl.EmptySet(), odl.UniversalSet()
```
adler-j: Perhaps use some other example sets

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+            has the same subsets as this set, `False` otherwise.
+        """
+        return (type(self) == type(other) and
+                len(other) == len(self) and
+                all(eo == es for eo, es in zip(other.elements, self.elements)))
+
+    def element(self, inp=None):
+        """Create a new element.
+
+        First tries calling the first set, then the second, etc.
```
adler-j: update

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+            return inp
+        else:
+            return Set.element(self, inp)
+
+    def __getitem__(self, indcs):
+        """Return ``self[indcs]``.
+
+        Examples
+        --------
+
```
adler-j: Remove blank row

adler-j:
bump

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+        Parameters
+        ----------
+        set1, ..., setN : `Set`
+            The elements in the set.
+
+        Examples
+        --------
+        >>> set = odl.FiniteSet(1, 'string')
+        """
+        self.__elements = tuple(elements)
```
kohr-h: How about duplicates?

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+
+        First tries calling the first set, then the second, etc.
+
+        For more specific control, use set[i].element() to pick which subset to
+        use.
+        """
+        if inp in self.elements:
+            return inp
+        else:
+            return Set.element(self, inp)
```
kohr-h: Huh? Does this raise an error or what?

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+
+        Returns
+        -------
+        equals : `boolean`
+            `True` if `other` is a `SetUnion` instance, and
+            has the same subsets as this set, `False` otherwise.
+        """
+        return (type(self) == type(other) and
+                len(other) == len(self) and
+                all(eo == es for eo, es in zip(other.elements, self.elements)))
```
kohr-h: `other.elements == self.elements`??

odl/set/sets.py
```diff
@@ -621,6 +622,352 @@ def __repr__(self):
...
+        if not all(isinstance(set_, Set) for set_ in sets):
+            wrong = [set_ for set_ in sets
+                     if not isinstance(set_, Set)]
+            raise TypeError('{!r} not Set instance(s)'.format(wrong))
+
+        self.__sets = tuple(sets)
+
+    @property
+    def sets(self):
+        """The factors (sets) as a tuple."""
```
adler-j: Any suggestion?

adler-j:
Thanks for the review, I'll got most of it fixed asap. Anyway regarding the wider comments, yes this is largely a "because we can", but also it is so fundamental to mathematics that I guess our users would assume that we support it.

The reason this is made instead of the built-in `set` and `tuple` is largely that we need a thin wrapper in order to use those with methods that only work on sets (see e.g. an operator with output `True` or `False`, could be given range `FiniteSet(True, False)`.)

odl/set/sets.py
kohr-h:
```python
docstrings = ["The sets of this {} as a tuple.".format(x)
              for x in ('union', 'intersection')]
```
😉

kohr-h:
> Thanks for the review, I'll got most of it fixed asap. Anyway regarding the wider comments, yes this is largely a "because we can", but also it is so fundamental to mathematics that I guess our users would assume that we support it.

Well, there's a lot more fundamental math to cover. But in a sense, because we already have abstract `Set`s, adding a simple special case and some basic operations on it is not totally far-fetched.

I'm trying hard to think of a case where *within the ODL ecosystem* you would need these classes. Functions to finite sets are not very useful (in principle yes, but not with this generality) because vectorization is impossible, or at least checks like `contains_all` become quite expensive.
Functionals... I don't know.
Operators... maybe


> The reason this is made instead of the built-in set and tuple is largely that we need a thin wrapper in order to use those with methods that only work on sets (see e.g. an operator with output True or False, could be given range FiniteSet(True, False).)

That was my suspicion 😉 An example doing this would certainly be more convincing.

odl/set/sets.py
adler-j: Also `__eq__` is order dependent :/

adler-j:
Fixed comments, need another review.

adler-j:
Bump

odl/util/utility.py
```diff
@@ -23,15 +23,17 @@
 standard_library.install_aliases()
 
 from functools import wraps
+from collections import OrderedDict
 import numpy as np
 
 
+
```
kohr-h: remove extra line

odl/set/sets.py
```diff
@@ -521,7 +521,7 @@ def __init__(self, *sets):
 
     @property
     def sets(self):
-        """Factors (sets) as a tuple."""
+        """The sets of this carthesian product as a tuple."""
```
kohr-h: wrong spelling

odl/set/sets.py
```diff
@@ -642,31 +637,32 @@ def __init__(self, *sets):
         ----------
         set1, ..., setN : `Set`
             The sets whose union should be taken.
+            Any duplicates are be ignored.
```
kohr-h: ~~be~~

odl/set/sets.py
```diff
@@ -684,13 +680,17 @@ def __eq__(self, other):
...
         """
         return (type(self) == type(other) and
-                len(other) == len(self) and
-                all(so == ss for so, ss in zip(other.sets, self.sets)))
+                all(set_ in other for set_ in self) and
+                all(set_ in self for set_ in other))
+
+    def __hash__(self):
+        """Return ``hash(self)``."""
+        return hash((type(self), set(self.sets)))
```
kohr-h: Why convert to `set` before hashing?

odl/set/sets.py
```diff
@@ -717,12 +717,12 @@ def __getitem__(self, indcs):
...
         Examples
         --------
-        >>> emp, univ = odl.EmptySet(), odl.UniversalSet()
-        >>> prod = odl.SetUnion(emp, univ, univ, emp, emp)
-        >>> prod[2]
-        UniversalSet()
-        >>> prod[2:4]
-        SetUnion(UniversalSet(), EmptySet())
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> prod = odl.SetUnion(reals, complexnrs)
```
kohr-h: `prod` not a good name

odl/set/sets.py
```diff
@@ -762,31 +758,32 @@ def __init__(self, *sets):
         ----------
         set1, ..., setN : `Set`
             The sets whose intersection should be taken.
+            Any duplicates are be ignored.
```
kohr-h: ~~be~~

odl/set/sets.py
```diff
@@ -837,12 +822,12 @@ def __getitem__(self, indcs):
...
         Examples
         --------
-        >>> emp, univ = odl.EmptySet(), odl.UniversalSet()
-        >>> prod = odl.SetIntersection(emp, univ, univ, emp, emp)
-        >>> prod[2]
-        UniversalSet()
-        >>> prod[2:4]
-        SetIntersection(UniversalSet(), EmptySet())
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> prod = odl.SetIntersection(reals, complexnrs)
```
kohr-h: `prod` not good

odl/set/sets.py
```diff
@@ -918,33 +895,38 @@ def __eq__(self, other):
 
         Returns
         -------
-        equals : `boolean`
-            `True` if `other` is a `SetUnion` instance, and
-            has the same subsets as this set, `False` otherwise.
+        equals : bool
+            ``True`` if ``other`` is a `SetUnion` instance, and
```
kohr-h: update

odl/util/utility.py
```diff
@@ -707,6 +709,44 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
...
+    Parameters
+    ----------
+    seq : sequence
+        Sequence with (possibly duplicate) elements.
+
+    Returns
+    -------
+    unique : list
+        Unique elements of seq. Order is guaranteed to be the same as in seq.
```
kohr-h: \`\`seq\`\`

odl/util/utility.py
```diff
@@ -707,6 +709,44 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
...
+
+    Examples
+    --------
+    >>> unique([1, 2, 3, 3])
+    [1, 2, 3]
+
+    >>> unique((1, 'str', 'str'))
+    [1, 'str']
+
+    Also works with unhashable types:
```
kohr-h: Make a sentence that doesn't start with "Also".

odl/set/sets.py
kohr-h: Done?

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+
+        For more specific control, use set[i].element() to pick which subset to
+        use.
+        """
+        for set in self.sets:
+            try:
+                return set.element(inp)
+            except NotImplementedError:
+                pass
+        raise NotImplementedError('element not implemented for any of the '
```
kohr-h: \`element\`

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+    def __hash__(self):
+        """Return ``hash(self)``."""
+        return hash((type(self), set(self.sets)))
+
+    def element(self, inp=None):
+        """Create a new element.
+
+        First tries calling the first set, then the second, etc.
+
+        For more specific control, use set[i].element() to pick which subset to
```
kohr-h: typeset `set[i].element()`

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+        """Return ``len(self)``."""
+        return len(self.sets)
+
+    def __getitem__(self, indcs):
+        """Return ``self[indcs]``.
+
+        Examples
+        --------
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> prod = odl.SetUnion(reals, complexnrs)
```
kohr-h: union

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
         return '{}({})'.format(self.__class__.__name__, sets_str)
 
 
+class SetUnion(Set):
+    """The union of several subsets.
```
kohr-h: blank line above

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> odl.SetUnion(reals, complexnrs)
+        SetUnion(RealNumbers(), ComplexNumbers())
+        """
+        sets_str = ', '.join(repr(set_) for set_ in self.sets)
+        return '{}({})'.format(self.__class__.__name__, sets_str)
+
+
+class SetIntersection(Set):
+    """The intersection of several subsets.
```
kohr-h: blank above

Check with the docstring of `CartesianProduct`.

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+        """Return ``len(self)``."""
+        return len(self.sets)
+
+    def __getitem__(self, indcs):
+        """Return ``self[indcs]``.
+
+        Examples
+        --------
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> prod = odl.SetIntersection(reals, complexnrs)
```
kohr-h: intersection

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+        >>> reals, complexnrs = odl.RealNumbers(), odl.ComplexNumbers()
+        >>> odl.SetIntersection(reals, complexnrs)
+        SetIntersection(RealNumbers(), ComplexNumbers())
+        """
+        sets_str = ', '.join(repr(set_) for set_ in self.sets)
+        return '{}({})'.format(self.__class__.__name__, sets_str)
+
+
+class FiniteSet(Set):
+    """A set given by a finite number of elements."""
```
kohr-h: blank above

odl/set/sets.py
```diff
@@ -621,6 +621,335 @@ def __repr__(self):
...
+            has the same subsets as this set, ``False`` otherwise.
+        """
+        # Need to loop since order could be different
+        return (type(self) == type(other) and
+                all(el in other for el in self) and
+                all(el in self for el in other))
+
+    def __hash__(self):
+        """Return ``hash(self)``."""
+        return hash((type(self), set(self.elements)))
```
kohr-h: Why go through `set`?

odl/util/utility.py
```diff
@@ -707,6 +709,44 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
...
+
+    >>> unique((1, [1], [1]))
+    [1, [1]]
+    """
+    # First check if all elements are hashable, if so O(n) can be done
+    try:
+        return list(OrderedDict.fromkeys(seq))
```
kohr-h: Is there no `OrderedSet` or similar?

kohr-h:
Status on this? Test failure is only a PEP8 thing.

adler-j:
I'll try to get this out of the way quite soon, but as you note there are quite some work.

odl/util/utility.py
adler-j: It seems not:
https://stackoverflow.com/questions/1653970/does-python-have-an-ordered-set

odl/set/sets.py
adler-j: To ignore ordering

odl/set/sets.py
adler-j: Uniqueness again

PR #867, opened 2017-01-26
MAINT: remove old LaTeX doc

kohr-h:
A bit of clean-up


adler-j:
Give a link, if you have one, to where this can be found now. If someone Googles.

kohr-h:
This was mostly internal stuff and never meant to be bundled here. What could be relevant is the acquisition stuff, but I realized it can be strongly simplified and is not really relevant in its current form. So nothing here is worth linking to I think.

PR #871, opened 2017-02-02
ENH: change in CallbackPrintIteration.

### LINKED ISSUE #870: Let CallbackPrintIteration not printing every iteration
When I am running many iterations to solve a problem, I want to have iteration print in the terminal in order to see that the things are running. However, I might not need to see all of them. I suggest that we implement a `display_step` like in `CallbackShow`. What do you think?

aringh:
The user can specify how often the iteration number should be printed.

Edit: this fixes #870

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -236,20 +236,25 @@ class CallbackPrintIteration(SolverCallback):
 
     _default_text = 'iter ='
 
-    def __init__(self, text=None):
+    def __init__(self, text=None, display_step=1):
         """Initialize a new instance.
 
         Parameters
         ----------
         text : string
             Text to display before the iteration count. Default: 'iter ='
+        display_step : positive int, optional
+            Number of iterations between output. Default: 1
         """
+        self.display_step = int(display_step)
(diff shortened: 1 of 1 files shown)


odl/solvers/util/callback.py
```diff
@@ -236,20 +236,25 @@ class CallbackPrintIteration(SolverCallback):
...
-    def __init__(self, text=None):
+    def __init__(self, text=None, display_step=1):
         """Initialize a new instance.
 
         Parameters
         ----------
         text : string
             Text to display before the iteration count. Default: 'iter ='
+        display_step : positive int, optional
+            Number of iterations between plots. Default: 1
```
kohr-h: "plots" doesn't make sense here. Use "output" or "print" or something like that.

aringh:
Fixed :-) I am not sure I can merge, or maybe after two approved reviews?

kohr-h:
You should be able to, just wait until CI has finished its job :-)

aringh:
> You're not authorized to merge this pull request.

:p

kohr-h:
Too bad :-)

PR #893, opened 2017-02-07
BUG: Fix FileReaderRawBinaryWithHeader ignoring extended header size,…

### LINKED ISSUE #892: Read mrc file
We use the ``FileReaderMRC`` to read mrc file following the example in ``examples/tomo/data/mrc_io.py``. Just read the given file ``test.mrc``, no problem. But if we read the interesting mrc file, get the following error:

>Traceback (most recent call last):

>  File "<ipython-input-6-8bf8cb54b126>", line 1, in <module>
>    runfile('/home/chchen/odl/odl/deform/read_mrc_data.py', wdir='/home/chchen/odl/odl/deform')

>  File "/home/chchen/miniconda2/lib/python2.7/site-packages/spyder/utils/site/sitecustomize.py", line > 866, in runfile
>    execfile(filename, namespace)

>  File "/home/chchen/miniconda2/lib/python2.7/site-packages/spyder/utils/site/sitecustomize.py", line > 94, in execfile
>   builtins.execfile(filename, *where)

>  File "/home/chchen/odl/odl/deform/read_mrc_data.py", line 114, in <module>
>    header, data = reader.read()

>  File "/home/chchen/odl/odl/tomo/data/uncompr_bin.py", line 385, in read
>    return self.read_header(), self.read_data()

>  File "/home/chchen/odl/odl/tomo/data/mrc.py", line 632, in read_data
>    order='F')

>ValueError: total size of new array must be unchanged
> > > >

adler-j:
… closes #892

### DIFF
--- a/odl/tomo/data/mrc.py
+++ b/odl/tomo/data/mrc.py
@@ -423,7 +423,7 @@ class MRCHeaderProperties(object):
             return labels[:nlabels]
 
 
-class FileReaderMRC(FileReaderRawBinaryWithHeader, MRCHeaderProperties):
+class FileReaderMRC(MRCHeaderProperties, FileReaderRawBinaryWithHeader):
 
     """Reader for the MRC file format.
 
@@ -473,8 +473,6 @@ class FileReaderMRC(FileReaderRawBinaryWithHeader, MRCHeaderProperties):
 
     print_mrc2014_spec = staticmethod(print_mrc2014_spec)
     print_fei_ext_header_spec = staticmethod(print_fei_ext_header_spec)
-    data_storage_shape = MRCHeaderProperties.data_storage_shape
-    data_dtype = MRCHeaderProperties.data_dtype
 
     def read_extended_header(self, groupby='field', force_type=''):
         """Read the extended header according to `extended_header_type`.
@@ -636,7 +634,7 @@ extended-mrc-format-not-used-by-2dx
         return data
 
 
-class FileWriterMRC(FileWriterRawBinaryWithHeader, MRCHeaderProperties):
+class FileWriterMRC(MRCHeaderProperties, FileWriterRawBinaryWithHeader):
 
     """Writer for the MRC file format.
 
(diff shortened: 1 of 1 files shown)


kohr-h:
I disagree with the fix. Extended header is a property of the MRC file format and has nothing to do with the general uncompressed binary reader.
What you call `total_header_size` is supposed to be `header_size` to begin with, since it makes no sense to call something "header size" if the actual header is larger. And that's actually what is supposed to happen already now, see [here](https://github.com/odlgroup/odl/blob/8c8070ddc5e330b511cee25a1cbcda651669902a/odl/tomo/data/mrc.py#L215-L232).
My guess is that the order of inheritance bites here, and the MRC reader needs to explicitly override the `header_size` property with the one from the mixin class.
That means that inserting a `header_size = MRCHeaderProperties.header_size` at [this](https://github.com/odlgroup/odl/blob/8c8070ddc5e330b511cee25a1cbcda651669902a/odl/tomo/data/mrc.py#L477) should fix the issue.

adler-j:
Wouldn't this whole thing be fixed by changing the order of inheritance so that the properties thing comes first?

kohr-h:
Yes that's right, in particular since the mixin class has no init, that shouldn't pose any difficulty.

adler-j:
Prefer this way? Now everything seems to be working properly.

odl/tomo/data/mrc.py
```diff
@@ -473,8 +473,6 @@ def __init__(self, file, header_fields=None):
 
     print_mrc2014_spec = staticmethod(print_mrc2014_spec)
```
adler-j: Should this line be moved to `MRCHeaderProperties`?

kohr-h: Yes why not, together with the next one.

kohr-h:
> Prefer this way? Now everything seems to be working properly.

Much  better :-)

adler-j:
Need an approval. Then merge.

adler-j:
New issue on the test: #894

PR #901, opened 2017-02-10
Issue 899  broadcast op len

adler-j:


odl/operator/pspace_ops.py
```diff
@@ -365,14 +366,16 @@ def __getitem__(self, index):
...
-        index : tuple of int
+        index : int or tuple of int
             A pair of integers given as (row, col).
 
         Returns
         -------
-        suboperator : `Operator` or ``0``
-            If there is an operator at (row, col), the operator is returned,
-            otherwise ``0``.
+        suboperator : `ReductionOperator`, `Operator` or ``0``
```
kohr-h: Style question: backticks around 0 or not? I'd say no.

odl/operator/pspace_ops.py
```diff
@@ -365,14 +366,16 @@ def __getitem__(self, index):
...
+            If index is an integer, return the row given by the index.
+
+            If index is a tuple, and if there is an operator at (row, col),
```
kohr-h: backticks around `(row, col)`

odl/operator/pspace_ops.py
```diff
@@ -390,14 +393,47 @@ def __getitem__(self, index):
...
+
+            for i in range(len(self.domain)):
+                if ops[i] is None:
+                    ops[i] = ZeroOperator(self.domain[i])
+
+            return ReductionOperator(*ops)
+
+    @property
+    def shape(self):
+        """Return the shape of the matrix of operators."""
```
kohr-h: ~~Return the~~
Same for every other property

odl/operator/pspace_ops.py
```diff
@@ -390,14 +393,47 @@ def __getitem__(self, index):
...
+                                          (self.ops.col == col))
+            if linear_index.size == 0:
+                return 0
+            else:
+                return self.ops.data[int(linear_index)]
         else:
-            return self.ops.data[int(linear_index)]
+            ops = [None] * len(self.domain)
+            for op, col, row in zip(self.ops.data, self.ops.col, self.ops.row):
+                if row == index:
```
kohr-h: Perhaps better to convert to integer first, otherwise you can get these "comparison is ambitous blabla" messages from Numpy.

odl/operator/pspace_ops.py
```diff
@@ -494,6 +530,18 @@ def adjoint(self):
...
         return ComponentProjectionAdjoint(self.domain, self.index)
 
+    def __repr__(self):
+        """Return ``repr(self)``.
+
+        Examples
+        --------
+        >>> X = odl.ProductSpace(odl.rn(1), odl.rn(2))
+        >>> odl.ComponentProjection(X, 0)
+        ComponentProjection(ProductSpace(rn(1), rn(2)), 0)
```
kohr-h: Actually not a bad idea to have a simple doctest for repr, that prevents it from being totally broken at least.

adler-j: For me this was a case of test driven development and I figured leaving it here is easier than moving it to the full tests.

Agree it helps in general.

odl/operator/pspace_ops.py
adler-j: Personally I'd say yes, mostly because it typesets and thus it is more clear that we mean the python value 0 instead of just the "mathematical concept 0", if you see what I mean.

odl/operator/pspace_ops.py
adler-j: Also catches some other errors

adler-j:
Fixed comments, merge after CI.

PR #902, opened 2017-02-13
Issue 319  linearspace more permissive

adler-j:
Thus turned out to be harder than I anticipiated, since stuff like (rn vector) + (operator) started giving wierd stuff. Thankfully I (think) I've solved it, but this is non-trivial and could use a thorough review.

Closes three issues anyhow!

### DIFF
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -1158,7 +1158,8 @@ class OperatorSum(Operator):
 
     def __repr__(self):
         """Return ``repr(self)``."""
-        return '({!r} + {!r})'.format(self.left, self.right)
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.left, self.right)
 
     def __str__(self):
         """Return ``str(self)``."""
@@ -1250,7 +1251,8 @@ class OperatorVectorSum(Operator):
 
     def __repr__(self):
         """Return ``repr(self)``."""
-        return '({!r} + {!r})'.format(self.operator, self.vector)
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.operator, self.vector)
 
     def __str__(self):
         """Return ``str(self)``."""
@@ -1382,7 +1384,8 @@ class OperatorComp(Operator):
 
     def __repr__(self):
         """Return ``repr(self)``."""
-        return '{!r} * {!r}'.format(self.left, self.right)
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.left, self.right)
 
     def __str__(self):
         """Return ``str(self)``."""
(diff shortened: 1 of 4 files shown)


odl/operator/operator.py
```diff
@@ -1158,8 +1158,7 @@ def adjoint(self):
 
     def __repr__(self):
         """Return ``repr(self)``."""
-        return '{}({!r}, {!r})'.format(self.__class__.__name__,
-                                       self.left, self.right)
+        return '({!r} + {!r})'.format(self.left, self.right)
```
kohr-h: Not sure I like this change. For long and messy output (which is like 90 % of the time), there's no way to use multiple lines for better visual appearance. Also the "+" sign will be hard to find in the output, whereas `OperatorSum` is clear from the beginning.

For `__str__`, okay.

odl/set/space.py
```diff
@@ -536,33 +549,56 @@ def __add__(self, other):
...
+            return self.__add__(other)
 
     def __isub__(self, other):
         """Implement ``self -= other``."""
         if other in self.space:
             return self.space.lincomb(1, self, -1, other, out=self)
+        elif isinstance(other, LinearSpaceElement):
+            # We do not `return NotImplemented` here since we don't want a
+            # fallback for in-place. Otherwise python attempts
+            # `self = self + other` which does not modify self.
```
kohr-h: \-

odl/set/space.py
```diff
@@ -588,66 +633,112 @@ def __rsub__(self, other):
...
         if other in self.space.field:
             return self.space.lincomb(1.0 / other, self, out=self)
         elif other in self.space:
             return self.space.divide(self, other, out=self)
-        else:
+        elif isinstance(other, LinearSpaceElement):
             # We do not `return NotImplemented` here since we don't want a
             # fallback for in-place. Otherwise python attempts
-            # `self = self / other` which does not modify self.
+            # `self = self * other` which does not modify self.
```
kohr-h: /

odl/test/space/ntuples_test.py
```diff
@@ -428,6 +428,39 @@ def test_binary_operator(fn, arithmetic_op):
...
+    Verifies that the statement z=op(x, y) gives equivalent results to NumPy.
+    """
+    [x_arr, y_arr], [x, y] = noise_elements(fn, 2)
+
+    # non-aliased left
+    y_arr_cpy = y_arr.copy()
+    z_arr = arithmetic_op(x_arr, y_arr_cpy)
+    z = arithmetic_op(x, y_arr)
+
+    assert isinstance(z, x.__class__)
```
kohr-h: `type(x)` easier to read

odl/operator/operator.py
adler-j: Reverting then

adler-j:
Fixed failing test and comments, reverted the repr change. Merge after CI.

PR #912, opened 2017-02-16
ENH: Add 3d cone cude performance example with ASTRA

adler-j:
Minor example. Clearly needs another example where we use pygpu on the ODL side.

### DIFF
--- a/examples/tomo/backends/astra_performance_cuda_cone_3d_cg.py
+++ b/examples/tomo/backends/astra_performance_cuda_cone_3d_cg.py
@@ -1,0 +1,108 @@
+"""Performance example of running native ASTRA vs using ODL for reconstruction.
+
+In this example, a 256x256x256 image is reconstructed using the Conjugate
+Gradient Least Squares method on the GPU.
+
+In general, pure ASTRA is faster than ODL since it does not need to perform any
+copies and all arithmetic is performed on the GPU. Despite this, ODL is not
+much slower. In this example, the overhead is about x3, depending on the
+hardware used.
+"""
+
+import astra
+import numpy as np
+import matplotlib.pyplot as plt
+import odl
+
+
+# Common geometry parameters
+
+domain_size = np.array([256, 256, 256])
+n_angles = 360
+det_size = 512
+niter = 10
+
+# Create reconstruction space
+reco_space = odl.uniform_discr(-domain_size / 2, domain_size / 2, domain_size)
+
+# Create geometry
+apart = odl.uniform_partition(0, 2 * np.pi, n_angles)
+dpart = odl.uniform_partition([-500, -500], [500, 500],
+                              [det_size, det_size])
+geometry = odl.tomo.CircularConeFlatGeometry(apart, dpart,
+                                             src_radius=500, det_radius=500)
+
+
+data = odl.phantom.shepp_logan(reco_space, modified=True).asarray()
+
+# --- ASTRA ---
+
+# Define ASTRA geometry
+astra_vol_geom = astra.create_vol_geom(*domain_size)
+det_row_count = geometry.det_partition.shape[1]
+det_col_count = geometry.det_partition.shape[0]
+vec = odl.tomo.backends.astra_setup.astra_conebeam_3d_geom_to_vec(geometry)
+astra_proj_geom = astra.create_proj_geom('cone_vec', det_row_count,
+                                         det_col_count, vec)
+
+# Create ASTRA projector
+proj_cfg = {}
+proj_cfg['type'] = 'cuda3d'
+proj_cfg['VolumeGeometry'] = astra_vol_geom
+proj_cfg['ProjectionGeometry'] = astra_proj_geom
+proj_cfg['options'] = {}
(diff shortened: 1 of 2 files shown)


examples/tomo/backends/astra_performance_cuda_cone_3d_cg.py
```diff
@@ -0,0 +1,108 @@
...
+
+# Solve with CGLS (aka CGN)
+x = reco_space.zero()
+with odl.util.Timer('ODL run'):
+    odl.solvers.conjugate_gradient_normal(ray_trafo, x, rhs, niter=niter)
+
+coords = (slice(None), slice(None), 128)
+
+# Display results for comparison
+plt.figure('data')
```
kohr-h: wrong title (or wrong input to plot)

adler-j:
My runtime diff is about the same as you with astra but 3 times that with ODL, try running the profiler on your machine and see what's wrong.

kohr-h:
Here's the culprit:

profile

It's the call in [this line](https://github.com/odlgroup/odl/blob/05873fdc6e1bd8769135608ce871ef46b19cf8a4/odl/tomo/backends/astra_cuda.py#L274-L275).

adler-j:
I have nothing like that on my machine, weird. Can you try to fix it on yours?

kohr-h:
Same runtime in the console, so no issue with Spyder.

kohr-h:
Same result in a fresh conda environment.

No idea what's going on.

kohr-h:
Okay, I had a suspicion and it was correct. There's some disk swapping going on in my home directory, which is on a network drive. Therefore it's bloody slow.

adler-j:
How much RAM do you have? Does the problem not fit in your RAM?

kohr-h:
> How much RAM do you have? Does the problem not fit in your RAM?

I've got 64G so that shouldn't be a problem. Also swap is on a hard disk somewhere else, so it's not really swapping. But when I monitor the the disk usage I see changes in my home dir (on a network FS) during the run.

adler-j:
Does ASTRA access the disk at any point?

kohr-h:
I don't think so. The pure ASTRA code is roughly as fast as it should be.

wjp:
I get roughly identical numbers as @kohr-h on a machine at CWI without a networked home directory. Using conda with odl-0.5.3-py35_2 and astra-toolbox-1.8-np111py35_1 .

adler-j:
Very weird stuff going on there. I get

```
                     ASTRA run :      9.586 
                       ODL run :     30.476 
```

where for me `ascontiguousarray` takes a total of 9 seconds. Could you expand it further so we see if there are any weird calls under the hood? And can you run with `ODL 0.5.4.dev0`?

Anyway, it seems the call to `ascontiguousarray` is not needed and removing it I get a slight speedup

```
                     ASTRA run :      9.510 
                       ODL run :     27.949 
```

I pushed that up here, see how it goes for you now.

adler-j:
Bump for input

wjp:
With ODL from this branch, on an i7-3930K + GTX1080, I get timings around

```
                     ASTRA run :      5.181 
                       ODL run :     32.031 
```

However, this was not triggered by your latest commit here, but by some other commit between 0.5.3 and this branch. (I can bisect if that would be interesting or useful, but haven't yet.)

adler-j:
Well bisecting is not useful, but it would be nice to see if the latest commit helped, anyway that number is at least better than the one before.

kohr-h:
I'm currently working on other stuff, but I'll also rerun the example. Still getting the weird oscillations between 30 and 85 seconds @wjp?

wjp:
For context: I saw the oscillations kohr-h is referring to with odl 0.5.3. Usually I got timings around 85 seconds, but maybe 20% of the time it was around 35 seconds instead.

I didn't see any large changes in timings caused by c1e6a4d72471cccfafdb8b613bb5b61d0c5e297f, but I'll re-run the script a few times to get some accurate numbers.

adler-j:
Extremely weird. Can you do some memory profiling and see if you occationally run into swapping? I.e. 30 sec could be when you run on the RAM, while 85 may be when you have to run on the disk.

For reference I've tried this on both linux and windows and I'm not seeing it. My setups are:

* Windows 7, Titan X GPU
* Ubuntu 14.04, GTX 970 GPU

wjp:
I just re-ran this script a few times with 3 versions of ODL.

There's no swapping (or any other noticable disk access).

3d23aad : 32.123 (average of 5 runs)
c1e6a4d: 32.322 (average of 5 runs)
ee326c5 (v5.3.0):  84.750 (average of 3 runs)

adler-j:
Thank you for the input!

It would seem that 32 seconds is at least much better than 85 seconds. So perhaps that is actual overhead and i guess we can live with it until the CUDA level integration is added.

Ok to merge?

kohr-h:
I guess we'll never know :-) 
Merging is fine with me.

kohr-h:
Thanks @wjp for exploring!

adler-j:
Agree, thanks @wjp for the help!

PR #929, opened 2017-02-21
ENH: add step to CallbackApply and fix reprs

kohr-h:


odl/solvers/util/callback.py
```diff
@@ -42,8 +42,8 @@ def __call__(self, result):
 
         Parameters
         ----------
-        result : `LinearSpaceElement`
-            Partial result after n iterations
+        iterate : `LinearSpaceElement`
```
adler-j: Also change in the signature

odl/solvers/util/callback.py
```diff
@@ -196,35 +196,66 @@ def __repr__(self):
...
+        >>> def func(x):
+        ...     print(np.max(x))
+        >>> callback = CallbackApply(func)
+        >>> x = odl.rn(3).element([1, 2, 3])
+        >>> callback(x)
+        3.0
+        >>> callback(x)
+        3.0
+
+        To apply only to each n-th iterate, supply ``step=n``:
```
adler-j: Perhaps we should rename the others to ``step`` as well?

odl/solvers/util/callback.py
kohr-h: Oops, of course

odl/solvers/util/callback.py
kohr-h: Would make it more consistent, right.

kohr-h:
Fixed, merging after CI

kohr-h:
Oh, wait, I need to change the examples, too.

PR #930, opened 2017-02-21
MAINT: improve some error messages for not implemented methods

kohr-h:


odl/solvers/functional/functional.py
```diff
@@ -125,7 +127,9 @@ def proximal(self):
         i.e., a function returning a proximal operator. See for example
         `forward_backward_pd`.
         """
-        raise NotImplementedError
+        raise NotImplementedError(
+            'no proximal operator implemented for functional {!r}'
+            ''.format(self))
```
kohr-h: This one was particularly annoying when you made a mistake in setting up the functionals. You would get an unspecific `NotImplementedError` because you did something that destroyed the proximal, but no clue what exactly.

Issue #931, opened 2017-02-21
space element > 0?


MatthiasJE:
Should the use of "`x > 0`" be allowed where x is some element of a space X, e.g. `X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[128, 128])`? Most spaces are not ordered so you could argue that it should not be defined. However, it is common to expect as an output an element of `X` where each location checks if `x > 0`, e.g. `X.element(x.asarray() > 0))`. In any case, currently ODL gives out a single `bool`. Is this a bug or do you want this behavior? I find it rather counter intuitive and would prefer the latter behavior (in line with numpy and MATLAB for vectors, matrices, etc)

Edit: @adler-j improved formatting.

adler-j:
### Clarification of the problem

The current behaviour can be seen from this Minimal Working Example:

```python
>>> X = odl.rn(3)
>>> x = X.one()
>>> x < 0
True
>>> x < 'some string'
True
```

This is clearly a bug that should be fixed. Specifically this is related to [a very weird part of the python 2 standard](https://docs.python.org/2/reference/datamodel.html#object.__cmp__) that specifies 

> If no `__cmp__()`, `__eq__()` or `__ne__()` operation is defined, class instances are compared by object identity (“address”)

It seems this is "fixed" in python 3. I.e. while Python 2 gives

```python
>>> 2 < '1'
True
```

Python 3 gives the much more sensible answer:

```python
>>> 2 < '1'
Traceback (most recent call last):
  File "python", line 1, in <module>
TypeError: unorderable types: int() < str()
```

### Regarding `x < 1`

After the above bug has been fixed, the question of how to handle expressions like `x < 1` still remain. As @MatthiasJE notes, both matlab and numpy handle this by returning the pointwise answer for all comparsions.

We have an interesting question of consistency in that one would expect `x < y` to be similar to `x == y`, but for equality, ODL actually has a well defined value and it is a single bool, in contrast to numpy. At the time we added this, it was selected given that ODL should emphasize that vectors are objects "in their own right" and not just tuples of objects.

So we do, to me, basically have three options

* Disable `x < 1`, let users use `np.less(x, 1)` instead if they want pointwise comparsions.
* Implement `x < 1` as pointwise comparsion, i.e. `np.less(x, 1)`.
* Implement `x < 1` as global comparsion, i.e.`np.all(np.less(x, 1))`.

I'll have to think of which one is the best, and any input is ofc nice.

[...]

PR #933, opened 2017-02-22
Issue 931  linearspace cmp

adler-j:
Fixes the first half of issue #931, does not add functionality for comparison.

### DIFF
--- a/odl/set/space.py
+++ b/odl/set/space.py
@@ -804,14 +804,10 @@ class LinearSpaceElement(object):
         return self.copy()
 
     def __cmp__(self, other):
-        """Comparsion not implemented.
-
-        Raises
-        ------
-        TypeError
-        """
-        raise TypeError('unorderable type `{}`'
-                        ''.format(self.__class__.__name__))
+        """Comparsion not implemented."""
+        # Stops python 2 from allowing comparsion of arbitrary objects
+        raise TypeError('unorderable types: {}, {}'
+                        ''.format(self.__class__.__name__, type(other)))
 
     # Metric space method
     def __eq__(self, other):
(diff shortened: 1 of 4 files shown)


odl/set/space.py
```diff
@@ -803,6 +803,16 @@ def __pos__(self):
...
 
+    def __cmp__(self, other):
+        """Comparsion not implemented.
+
+        Raises
+        ------
+        TypeError
+        """
+        raise TypeError('unorderable type `{}`'
+                        ''.format(self.__class__.__name__))
```
kohr-h: Can you try to mimic the upstream error message? It says, for example, `unorderable types: int() < str()`. Perhaps replace the comparison operator by a comma, i.e.
```python
raise TypeError('unorderable types: {}, {}'.format(self.__class__.__name__, type(other)))
```

adler-j:
Added a more controversial change, related to comparsion of sets. Please have a look.

kohr-h:
That's indeed more controversial. Do we want to have things like `grid <= partition`, `grid <= intv_prod` and the likes?

In some sense I feel that comparison operations `<=` should only compare within the same "category" while `contains_set` explicitly tells "compare in the sense of sets".
For instance, for grids, I think `<=`, if implemented at all, should invoke `is_subgrid` (not we have reversed logic here, we should go for one, either `is_subset` or `contains_grid`).

adler-j:
Ok, I'll remove that part then and fix your comment, the merge.

adler-j:
Also disallowed comparsion of `Set` instances in python 2 and added a test for it.

kohr-h:
Okay, good fix.

Issue #936, opened 2017-02-22
Sampling Operator (sample coefficients of discretised function)


MatthiasJE:
I want to implement a sampling operator that takes an element of a vector space, e.g. X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[128, 128]) and maps it to a list of coefficients, e.g. Xd = odl.space.rn(N) for instance by taking every other index for N = 128**2/2 or just N times the first index. How can I achieve this within ODL? 

The reason why I want to do this is two fold. First, for convenience it might be good to map images into vectors by "reshaping" them. Also this is a good way to model a sampling operator that is used for instance in MRI. When I visited you in Stockholm last August we came up with the following solution for MRI:

```python
n = 128
s_image = (n,n)

# Create a space
#space = odl.uniform_discr([-1, -1], [1, 1], s_image, dtype='complex')
space = odl.uniform_discr([0, 0], [1, 1], s_image, dtype='float32')

# Create MRI operator. First fourier transform, then subsample
ft = odl.trafos.FourierTransform(space, halfcomplex=False)

sampling_array = np.zeros(ft.range.shape)

num_lines = 20
line_indcs = np.random.normal(loc=n//2, scale=n/4, size=num_lines).astype(int)
line_indcs[line_indcs < 0] = 0
line_indcs[line_indcs >= n] = n - 1

sampling_array[line_indcs, :] = 1
sampling_array[n//2-2:n//2+3, :] = 1
#sampling_array[50:70, 50:70] = 1
sampling_vector = ft.range.element(sampling_array)

sampling_vector.show('sampling mask')

mri_op = sampling_vector * ft
```

which was sufficient for our purposes but here I want a little more. 

First, the output is still in the same space (which might be very much larger than the initial space) and thus memory inefficient. Second, it does not allow to sample coefficients multiple times. Do you have any idea how to do this? For the latter I don't even have a good python code at the moment as it's adjoint sums over elements that have been sampled multiple times. I am sure there is a smart way as MATLAB has a neat way of doing this, too.

MatthiasJE:
I have started to code some operator to do the sampling (see end of comment) which seems to work fine (please ignore the comments in the code as they are still the one from the template I used). When testing it with

```
X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[128, 128], weighting="None")
ind = range(0,100)
A = odl.operator.SamplingOperator(X, ind)
x = X.one()
[...]

PR #861, opened 2017-01-25
Tensor spaces, second try

kohr-h:
Copied from #753

TODOs:
- [x] Implement basic structures for tensors and tensor spaces
- [x] Replicate functionality of `NtuplesBase`, `FnBase`, `NumpyNtuples` and `NumpyFn` in the corresponding tensor classes
- [x] Remove `Ntuples` and related stuff
- [x] Add former `Ntuples` and the likes as special case to tensor classes
- [x] Change names with "tensor" to something more lightweight
- [x] Find place for ~~`MatVecOperator`~~`MatrixOperator`, see #584. Solved in #911 
- [x] Improve pretty-printing of elements
- [x] Fix broken downstream code (`lp_discr`, ...)
- [x] Change code that was restricted to 1D data containers
- [x] Port unit tests
- [x] Write some doctests
- [x] Write more doctests
- [x] Remove `TensorSet` and others (see #860) duck-type space properties
- [x] Check documentation for outdated information, e.g. regarding 1D data layout (Update: can't find anything explicit)
- [x] ~~Use notion "rank" instead of "number of dimensions" to better distinguish between `space.ndim` and `space.size`~~ Not doing this
- [x] Implement indexing of `TensorSpace`. ~~Done, but kind of wrong, see [here](https://github.com/odlgroup/odl/pull/861#issuecomment-275454215) and [here](https://github.com/odlgroup/odl/pull/861#issuecomment-275540310)~~ Done properly now
- [x] ~~Write extra documentation on tensors~~ Should be part of #908
- [x] Port `show` methods (1D from `Ntuples`, 2D from `DiscreteLp`)
- [ ] Add unit tests for unusual data types and the likes
- [x] Simplify doctests of tensor spaces
- [x] Fix `FunctionSpaceElement` in 1D (again)
- [x] Implement tensor-valued functions, see #908 
- [x] Fix interpolation from array of points, related to #908 
- [x] Cleanup (old stuff, commit history, ...)
- [ ] Fix large-scale tests
- [x] Make test with power function more general
- [ ] Add more tests for product space multi-indexing
- [ ] Implement indexing of `FunctionSpace` as 
  - [ ] indexing the components of the output (i.e. raise an error for scalar-valued functions) in `__getitem__`
  - [ ] by axis of the domain using a `byaxis` attribute
- [ ] Implement indexing of `DiscreteLp` as in `FunctionSpace`
[...]

PR #940, opened 2017-02-24
initial version of sampling and embedding operators

MatthiasJE:
This is the current version of the code that we discussed in #936. Please have a look and give feedback. Thanks!

### DIFF
--- a/odl/operator/fn_ops.py
+++ b/odl/operator/fn_ops.py
@@ -1,0 +1,406 @@
+## Copyright 2014-2016 The ODL development group
+##
+## This file is part of ODL.
+##
+## ODL is free software: you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published by
+## the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## ODL is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with ODL.  If not, see <http://www.gnu.org/licenses/>.
+#
+#"""Default operators defined on any (reasonable) space."""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import super
+
+from odl.operator.operator import Operator
+from odl.space import fn
+from odl.set import LinearSpace, Field
+#from odl.util import writable_array
+
+import numpy as np
+
+__all__ = ('SamplingOperator', 'ZerofillingOperator', 
+           'VectoriseOperator', 'EmbeddingOperator')
+
+class SamplingOperator(Operator):
+
+    """Operator that samples coefficients.
+
+        ``SamplingOperator(f)(i) == cell_volume * f(i) \approx \int_cell f(x) dx``
+    """
+
+    def __init__(self, domain, indices):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain : `LinearSpace` or `Field`
+            Set of elements on which this operator acts.
+        indices : list
+            List of indices where to sample the function
+
+        Examples
+        --------
+        >>> X = odl.uniform_discr(min_pt=[0, 0], max_pt=[1, 1], shape=[2, 2])
+        >>> x = X.element(range(X.size))
+        >>> ind = list()
+        >>> ind.append(range(2))
+        >>> ind.append([1,1])
+        >>> A = odl.operator.SamplingOperator(X, ind)
+        >>> A(x)
+        rn(2).element([0.25, 0.75])
+        >>> A(x)  # Out-of-place
+        rn(2).element([0.25, 0.75])
+        """
+        if not isinstance(domain, (LinearSpace, Field)):
+            raise TypeError('`space` {!r} not a `LinearSpace` or `Field` '
+                            'instance'.format(domain))
+
+        self.__indices = np.asarray(indices, dtype=int)
+        # Converts a single sequence to an array of integers
+        # and a sequence of arrays to a vertically stacked array
+        if self.indices.ndim > 1:
+            # Strides = increment in linear indices per axis
+            strides = np.concatenate((np.cumprod(domain.shape[1:])[::-1], [1]))
+            self.__flat_indices = np.sum(self.indices * strides[:, None], axis=0)
+        
+        range = fn(self.flat_indices.size, dtype=domain.dtype)
+
+        super().__init__(domain, range, linear=True)
+
+    @property
+    def indices(self):
+        """indices where to sample the function."""
+        return self.__indices
+
+    @property
+    def flat_indices(self):
+        """indices where to sample the function."""
+        return self.__flat_indices
+
+    def _call(self, x, out=None):
+        """Collect indices weighted with the cell volume."""
+        c = getattr(self.domain, 'cell_volume', 1.0) # TODO: This only works for a subset of function spaces
+        if out is None:
+            out = x[self.flat_indices]*c
+        else:
+            out[:] = x[self.flat_indices]*c
+        return out
+
+    @property
+    def adjoint(self):
+        """Adjoint of the sampling is 'zero filling'.
+
+        Returns
+        -------
+        adjoint : `ZerofillingOperator`
+        """
+        return ZerofillingOperator(self.domain, self.indices)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.domain, self.indices)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{} * I'.format(self.indices)
+
+class ZerofillingOperator(Operator):
+
+    """Operator of multiplication with a scalar.
+
+        ``ZerofillingOperator(f)(x) == sum_{y in I, y == x} f(y), 0 if x not in I``
+    """
+
+    def __init__(self, range, indices):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        range : `LinearSpace` or `Field`
+            Set of elements into which this operator maps.
+        sample_indices : list
+            List of indices (I in the formula above)
+
+        Examples
+        --------
+        >>> X = odl.uniform_discr(min_pt=[0, 0], max_pt=[1, 1], shape=[2, 2])
+        >>> x = X.one()
+        >>> ind = list()
+        >>> ind.append(range(2))
+        >>> ind.append([1,1])
+        >>> A = odl.operator.SamplingOperator(X, ind)
+        >>> A.adjoint(A(x))
+        uniform_discr([0.0, 0.0], [1.0, 1.0], (2, 2)).element([[0.0, 0.25], [0.0, 0.25]])
+        >>> A.adjoint(A(x))  # Out-of-place
+        uniform_discr([0.0, 0.0], [1.0, 1.0], (2, 2)).element([[0.0, 0.25], [0.0, 0.25]])
+
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[4, 3])
+        >>> x = X.element(range(1,X.size+1))
+        >>> ind = list()
+        >>> ind.append(range(3))
+        >>> ind.append(range(3))
+        >>> A = odl.operator.SamplingOperator(X, ind)
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x))
+        -1.7763568394002505e-15
+        >>> A.adjoint(A(x))  # Out-of-place
+        -1.7763568394002505e-15
+        """
+
+        self.__indices = np.asarray(indices, dtype=int)
+        # Converts a single sequence to an array of integers
+        # and a sequence of arrays to a vertically stacked array
+        if self.indices.ndim > 1:
+            # Strides = increment in linear indices per axis
+            strides = np.concatenate((np.cumprod(range.shape[1:])[::-1], [1]))
+            self.__flat_indices = np.sum(self.indices * strides[:, None], axis=0)
+
+        self.__flat_indices_unique = np.unique(self.flat_indices)
+
+        domain = fn(self.flat_indices.size, dtype=range.dtype)
+        super().__init__(domain, range, linear=True)
+            
+    @property
+    def indices(self):
+        """indices where to sample the function."""
+        return self.__indices
+
+    @property
+    def flat_indices(self):
+        """flattened set of indices where to sample the function."""
+        return self.__flat_indices
+
+    @property
+    def flat_indices_unique(self):
+        """unique set of indices where to sample the function."""
+        return self.__flat_indices_unique
+
+    def _call(self, x, out=None):
+        """sum all values if indices are given multiple times."""        
+        y = np.bincount(self.flat_indices, weights=x, minlength=self.range.size)
+            
+        if out is None:
+            out = self.range.element(y)
+        else:
+            out[:] = y
+
+        return out
+
+    @property
+    def adjoint(self):
+        """Adjoint of the sampling is 'sampling'.
+
+        Returns
+        -------
+        adjoint : `SamplingOperator`
+        """
+        return SamplingOperator(self.domain, self.indices)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.domain, self.indices)
+                
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{} * I'.format(self.indices)
+
+
+class VectoriseOperator(Operator):
+
+    """Operator that reshapes the object as a column vector.
+
+        ``VectoriseOperator(f) == f[:]``
+    """
+
+    def __init__(self, domain, scaling=1):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain : `LinearSpace` or `Field`
+            Set of elements on which this operator acts.
+
+        Examples
+        --------
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> x = X.element(range(X.size))
+        >>> A = odl.operator.VectoriseOperator(X)
+        >>> A(x)
+        rn(12).element([0.0, 1.0, 2.0, ..., 9.0, 10.0, 11.0])
+        >>> A(x)  # Out-of-place
+        rn(12).element([0.0, 1.0, 2.0, ..., 9.0, 10.0, 11.0])
+
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> x = X.element(range(X.size))
+        >>> A = odl.operator.VectoriseOperator(X)
+        >>> (A.inverse(A(x)) - x).norm()
+        0.0
+        >>> (A.inverse(A(x)) - x).norm() # Out-of-place
+        0.0
+        """
+        if not isinstance(domain, (LinearSpace, Field)):
+            raise TypeError('`space` {!r} not a `LinearSpace` or `Field` '
+                            'instance'.format(domain))
+        
+        self.__scaling = scaling        
+        range = fn(domain.size, dtype=domain.dtype)
+        super().__init__(domain, range, linear=True)
+
+    def _call(self, x, out=None):
+        """Collect indices."""
+        if out is None:
+            out = self.scaling*x[:]
+        else:
+            out[:] = self.scaling*x[:]
+        return out
+
+    @property
+    def scaling(self):
+        """scaling factor."""
+        return self.__scaling
+    
+    @property
+    def adjoint(self):
+        """Adjoint of the vectorising is creating an element with these components
+        with an appropriate weighting.
+
+        Returns
+        -------
+        adjoint : `VectoriseOperatorAdjoint`
+        """
+        c = getattr(self.domain, 'cell_volume', 1.0) # TODO: This only works for a subset of function spaces
+        return EmbeddingOperator(self.domain, self.scaling/c)
+
+    @property
+    def inverse(self):
+        """Inverse of the vectorising is creating an element with these components.
+
+        Returns
+        -------
+        adjoint : `EmbeddingOperator`
+        """
+        return EmbeddingOperator(self.domain, 1./self.scaling)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r})'.format(self.__class__.__name__, self.domain)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{}'
+
+
+class EmbeddingOperator(Operator):
+
+    """Operator that creates an element of a vector space given its coefficients.
+
+        ``EmbeddingOperator(f) == X.element(f)``
+        
+        For the adjoint it is important that the element is scaled properly.
+    """
+
+    def __init__(self, range, scaling=1):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        range : `LinearSpace` or `Field`
+            Space that the operator should map to.
+
+        Examples
+        --------
+        Test for the adjoint
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> A = odl.operator.VectoriseOperator(X)
+        >>> x = A.domain.element(range(A.domain.size))
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x))
+        0.0
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x)) # Out-of-place
+        0.0
+
+        Test for the adjoint of the adjoint
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> A = odl.operator.VectoriseOperator(X).adjoint
+        >>> x = A.domain.element(range(A.domain.size))
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x))
+        0.0
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x)) # Out-of-place
+        0.0
+
+        Test for the adjoint of the inverse
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> A = odl.operator.VectoriseOperator(X).inverse
+        >>> x = A.domain.element(range(A.domain.size))
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x))
+        0.0
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x)) # Out-of-place
+        0.0
+        """
+        if not isinstance(range, (LinearSpace, Field)):
+            raise TypeError('`space` {!r} not a `LinearSpace` or `Field` '
+                            'instance'.format(range))
+
+        self.__scaling = scaling        
+        domain = fn(range.size, dtype=range.dtype)
+        super().__init__(domain, range, linear=True)
+
+    def _call(self, x, out=None):
+        """Create a new element with the known coefficients."""
+        if out is None:
+            out = self.range.element(self.scaling*x)
+        else:
+            out[:] = self.scaling*x[:]
+        return out
+
+    @property
+    def scaling(self):
+        """scaling factor."""
+        return self.__scaling
+
+    @property
+    def adjoint(self):
+        """Adjoint of the embedding is vectorising with an appropriate scaling'.
+
+        Returns
+        -------
+        adjoint : `VectoriseOperator`
+        """
+        c = getattr(self.range, 'cell_volume', 1.0) # TODO: This only works for a subset of function spaces
+        return VectoriseOperator(self.range, self.scaling/c)
+
+    @property
+    def inverse(self):
+        """Inverse of the embedding is vectorising'.
+
+        Returns
+        -------
+        adjoint : `VectoriseOperator`
+        """
+        return VectoriseOperator(self.range, 1./self.scaling)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r})'.format(self.__class__.__name__, self.domain)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{}'
+
+if __name__ == '__main__':
+    # pylint: disable=wrong-import-position
+    from odl.util.testutils import run_doctests
+    run_doctests()
--- a/odl/operator/fn_ops.py
+++ b/odl/operator/fn_ops.py
@@ -41,7 +41,7 @@ class SamplingOperator(Operator):
         ``SamplingOperator(f, indices) == cell_volume * f[indices]``
     """
 
-    def __init__(self, domain, indices):
+    def __init__(self, domain, indices, weighting=None):
         """Initialize a new instance.
 
         Parameters
@@ -50,6 +50,8 @@ class SamplingOperator(Operator):
             Set of elements on which this operator acts.
         indices : list
             List of indices where to sample the function
+        weighting : {None, 'cell_volume'}
+            weighting scheme
 
         Examples
         --------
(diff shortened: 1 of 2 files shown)


odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
+## Copyright 2014-2016 The ODL development group
```
adler-j: Somehow snuck in dubble # here, remove one. Also it is 2017 now so might as well fix that while at it.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+import numpy as np
+
+__all__ = ('SamplingOperator', 'ZerofillingOperator', 
+           'VectoriseOperator', 'EmbeddingOperator')
+
+class SamplingOperator(Operator):
+
+    """Operator that samples coefficients.
+
+        ``SamplingOperator(f)(i) == cell_volume * f(i) \approx \int_cell f(x) dx``
```
adler-j: Use math typesetting or equation typsetting, try to mix. I.e

```
.. math::
    SamplingOperator(f)_i = cell_volume * f_{indices_i} \approx \int_{\text{cell}_{indices_i}} f(x) dx
```
    
OR

```
The operator is defined by::
    SamplingOperator(f, indices) == cell_volume * f[indices]
```

I would suggest the second option myself.

Further, I would personally prefer if this operator did not scale by `cell_volume`, i.e.

```
The operator is defined by::
    SamplingOperator(f, indices) == f[indices]
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    """
+
+    def __init__(self, domain, indices):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        domain : `LinearSpace` or `Field`
```
adler-j: I guess `domain` needs to be a `FnBase` or `DiscreteLp`

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+            Set of elements on which this operator acts.
+        indices : list
+            List of indices where to sample the function
+
+        Examples
+        --------
+        >>> X = odl.uniform_discr(min_pt=[0, 0], max_pt=[1, 1], shape=[2, 2])
+        >>> x = X.element(range(X.size))
+        >>> ind = list()
```
adler-j: Rather just write 

```python
>>> ind = [[0, 1], 
...        [1, 1]]
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        >>> ind.append(range(2))
+        >>> ind.append([1,1])
+        >>> A = odl.operator.SamplingOperator(X, ind)
```
adler-j: If you add `fn_ops` [here](https://github.com/odlgroup/odl/blob/master/odl/__init__.py), you could write this as `odl.SamplingOperator`.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        >>> A(x)
+        rn(2).element([0.25, 0.75])
+        >>> A(x)  # Out-of-place
```
adler-j: Duplicate from above

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        return self.__indices
+
+    @property
+    def flat_indices(self):
+        """indices where to sample the function."""
+        return self.__flat_indices
+
+    def _call(self, x, out=None):
+        """Collect indices weighted with the cell volume."""
+        c = getattr(self.domain, 'cell_volume', 1.0) # TODO: This only works for a subset of function spaces
```
adler-j: As noted above, I'd prefer if this was not done.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        if out is None:
+            out = x[self.flat_indices]*c
```
adler-j: This would fail [PEP8](https://www.python.org/dev/peps/pep-0008/) and should be written as

```
out = x[self.flat_indices] * c
```

To verify that your code follows pep8, you can run the whole test suite:

```
$ pytest --pep8
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+
+    @property
+    def adjoint(self):
+        """Adjoint of the sampling is 'zero filling'.
+
+        Returns
+        -------
+        adjoint : `ZerofillingOperator`
+        """
+        return ZerofillingOperator(self.domain, self.indices)
```
adler-j: If you change the scaling of the operator to be zero, it would mean you have to add a scaling here instead.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+
```
adler-j: Perhaps also add (pseudo-)`inverse`

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.domain, self.indices)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{} * I'.format(self.indices)
```
adler-j: I'd prefer `return repr(self)`

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+
+class ZerofillingOperator(Operator):
+
+    """Operator of multiplication with a scalar.
+
+        ``ZerofillingOperator(f)(x) == sum_{y in I, y == x} f(y), 0 if x not in I``
```
adler-j: See above comment on math vs code

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.domain, self.indices)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{} * I'.format(self.indices)
+
+class ZerofillingOperator(Operator):
+
+    """Operator of multiplication with a scalar.
```
adler-j: Old doc

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        range : `LinearSpace` or `Field`
+            Set of elements into which this operator maps.
+        sample_indices : list
+            List of indices (I in the formula above)
+
+        Examples
+        --------
+        >>> X = odl.uniform_discr(min_pt=[0, 0], max_pt=[1, 1], shape=[2, 2])
+        >>> x = X.one()
+        >>> ind = list()
```
adler-j: See above, i.e.

```python
>>> ind = [[0, 1], 
...        [1, 1]]
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        >>> ind.append(range(2))
+        >>> ind.append([1,1])
+        >>> A = odl.operator.SamplingOperator(X, ind)
```
adler-j: In general we try to have the examples of each operator present the operator in isolation, i.e. unless it doesnt make sense as anything but an adjoint we present the operator "in its own right".

In this case, that would mean creating a `ZerofillingOperator` directly.

The test of the adjoint could then be performed in an Example under the `adjoint` property.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+            self.__flat_indices = np.sum(self.indices * strides[:, None], axis=0)
+
+        self.__flat_indices_unique = np.unique(self.flat_indices)
+
+        domain = fn(self.flat_indices.size, dtype=range.dtype)
+        super().__init__(domain, range, linear=True)
+            
+    @property
+    def indices(self):
+        """indices where to sample the function."""
```
adler-j: General comment:

Full sentence, so start with capital letter

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        return self.__indices
+
+    @property
+    def flat_indices(self):
```
adler-j: A more transparent naming scheme (that works better with an IDE) would perhaps be

```
indices
indices_flat
indices_flat_unique
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    def flat_indices_unique(self):
+        """unique set of indices where to sample the function."""
+        return self.__flat_indices_unique
+
+    def _call(self, x, out=None):
+        """sum all values if indices are given multiple times."""        
+        y = np.bincount(self.flat_indices, weights=x, minlength=self.range.size)
+            
+        if out is None:
+            out = self.range.element(y)
```
adler-j: This is automatically performed by ODL, you could merely do:

```
out = y
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.domain, self.indices)
+                
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{} * I'.format(self.indices)
```
adler-j: See above, i.e.

```
return repr(self.indices)
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+
+    @property
+    def adjoint(self):
+        """Adjoint of the sampling is 'sampling'.
+
+        Returns
+        -------
+        adjoint : `SamplingOperator`
+        """
+        return SamplingOperator(self.domain, self.indices)
```
adler-j: If you fix the scaling comment above, you may need to add scaling here. Also consider adding an `inverse`

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{} * I'.format(self.indices)
+
+
+class VectoriseOperator(Operator):
+
+    """Operator that reshapes the object as a column vector.
+
+        ``VectoriseOperator(f) == f[:]``
```
adler-j: This looks like matlab syntax to me, since in python this is a no-op.

I suspect you mean

```
VectoriseOperator(f) == np.ravel(f)
```

or something like it?

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    """
+
+    def __init__(self, domain, scaling=1):
```
adler-j: Avoid having scaling here, instead simply use operator arithmetic if you need it at some point, i.e.

```
>>> op = VectoriseOperator(space)
>>> scaled_op = scaling * op
```

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    @property
+    def adjoint(self):
+        """Adjoint of the vectorising is creating an element with these components
+        with an appropriate weighting.
+
+        Returns
+        -------
+        adjoint : `VectoriseOperatorAdjoint`
+        """
+        c = getattr(self.domain, 'cell_volume', 1.0) # TODO: This only works for a subset of function spaces
```
adler-j: Note my above comment: try to to do weighting inside and use operator arithmetic

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        return out
+
+    @property
+    def scaling(self):
+        """scaling factor."""
+        return self.__scaling
+    
+    @property
+    def adjoint(self):
+        """Adjoint of the vectorising is creating an element with these components
```
adler-j: If possible, fit this on one line.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        return '{}'
+
+
+class EmbeddingOperator(Operator):
+
+    """Operator that creates an element of a vector space given its coefficients.
+
+        ``EmbeddingOperator(f) == X.element(f)``
+        
+        For the adjoint it is important that the element is scaled properly.
```
adler-j: Indentation

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{}'
+
+
+class EmbeddingOperator(Operator):
+
+    """Operator that creates an element of a vector space given its coefficients.
+
+        ``EmbeddingOperator(f) == X.element(f)``
```
adler-j: Write this as

```
The operator is defined by::
    EmbeddingOperator(X)(f) == X.element(f)
```

This typesets better

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        Parameters
+        ----------
+        range : `LinearSpace` or `Field`
+            Space that the operator should map to.
+
+        Examples
+        --------
+        Test for the adjoint
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> A = odl.operator.VectoriseOperator(X)
```
adler-j: See above comment on `__all__`, then do `A = odl.VectoriseOperator(X)`

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        range : `LinearSpace` or `Field`
+            Space that the operator should map to.
+
+        Examples
+        --------
+        Test for the adjoint
```
adler-j: Add a blank row after this

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[3, 4])
+        >>> A = odl.operator.VectoriseOperator(X)
+        >>> x = A.domain.element(range(A.domain.size))
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x))
+        0.0
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x)) # Out-of-place
+        0.0
+
+        Test for the adjoint of the adjoint
```
adler-j: Blank row after (do everywhere in similar cases)

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+class EmbeddingOperator(Operator):
+
+    """Operator that creates an element of a vector space given its coefficients.
+
+        ``EmbeddingOperator(f) == X.element(f)``
+        
+        For the adjoint it is important that the element is scaled properly.
+    """
+
+    def __init__(self, range, scaling=1):
```
adler-j: See above comment on scaling.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        """
+        return VectoriseOperator(self.range, 1./self.scaling)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r})'.format(self.__class__.__name__, self.domain)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{}'
```
adler-j: `return repr(self)`.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+
+if __name__ == '__main__':
+    # pylint: disable=wrong-import-position
+    from odl.util.testutils import run_doctests
+    run_doctests()
```
adler-j: Missing newline at end.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        --------
+        >>> X = odl.uniform_discr(min_pt=[0, 0], max_pt=[1, 1], shape=[2, 2])
+        >>> x = X.one()
+        >>> ind = list()
+        >>> ind.append(range(2))
+        >>> ind.append([1,1])
+        >>> A = odl.operator.SamplingOperator(X, ind)
+        >>> A.adjoint(A(x))
+        uniform_discr([0.0, 0.0], [1.0, 1.0], (2, 2)).element([[0.0, 0.25], [0.0, 0.25]])
+        >>> A.adjoint(A(x))  # Out-of-place
```
adler-j: Duplicate

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,406 @@
...
+        uniform_discr([0.0, 0.0], [1.0, 1.0], (2, 2)).element([[0.0, 0.25], [0.0, 0.25]])
+
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[4, 3])
+        >>> x = X.element(range(1,X.size+1))
+        >>> ind = list()
+        >>> ind.append(range(3))
+        >>> ind.append(range(3))
+        >>> A = odl.operator.SamplingOperator(X, ind)
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x))
+        -1.7763568394002505e-15
```
adler-j: Approximate values like this cause bugs and failing tests, I'd prefer doing an approximate test, e.g.

```
>>> abs(A.adjoint(A(x)).inner(x) - A(x).inner(A(x))) < 1e-10
True
```

odl/operator/fn_ops.py
MatthiasJE: Thanks, done.

odl/operator/fn_ops.py
MatthiasJE: I agree about the syntax and I have changed that. 

About the cell_volume I disagree as we are trying to approximate \int_{\text{cell}_{indices_i}} f(x) dx ??? But then this should depend on how the interpolation is done as well which currently it does not. The current version is the right formula if we assume that f is constant over the cell.

odl/operator/fn_ops.py
MatthiasJE: OK, changed

odl/operator/fn_ops.py
MatthiasJE: Changed

odl/operator/fn_ops.py
MatthiasJE: fn_ops was already added to __init__ so no need to write it like this. Thanks!

odl/operator/fn_ops.py
MatthiasJE: I tried a bit and I cannot run pytest successfully on ODL. I also get error messages like:  
"fixture 'fn_impl' not found"

However, the ODL parts that I am using run fine and seem to do the right thing.

odl/operator/fn_ops.py
MatthiasJE: I am not sure I follow what you mean. Is this related to the scaling I introduced for the other operators?

odl/operator/fn_ops.py
MatthiasJE: Good idea. Not sure what it is though. Probably something like the average over all sampled components. However, I would not know how to implement this easily.

odl/operator/fn_ops.py
MatthiasJE: This is just more or less copy pasted from another ODL example. Changed it now to "return repr(self)"

odl/operator/fn_ops.py
MatthiasJE: I do see your point but I would not know how to write this easily with code. Do you?

odl/operator/fn_ops.py
MatthiasJE: changed

odl/operator/fn_ops.py
MatthiasJE: Good point. Changed.

odl/operator/fn_ops.py
MatthiasJE: Changed. I hope I found all relevant cases ... :)

odl/operator/fn_ops.py
MatthiasJE: Good point, changed it. The attribute "indices_flat_unique" is gone now anyways due to comments of @kohr-h

odl/operator/fn_ops.py
MatthiasJE: Good to know. Can you quickly explain how you managed that?

MatthiasJE: Ah, I see. You want me to write

+        if out is None:
+            out = y

right?

MatthiasJE: If so, then this is now corrected

odl/operator/fn_ops.py
MatthiasJE: We should probably discuss the scaling separately. For now I would say I leave it the way it is. Regarding the inverse, see my comment above.

odl/operator/fn_ops.py
MatthiasJE: Changed

odl/operator/fn_ops.py
MatthiasJE: True, MATLAB caught be here.

odl/operator/fn_ops.py
MatthiasJE: The problem is that the adjoint and inverse differ by different scaling. I though it would be more elegant to have one operator that can be scaled differently depending on whether we want to have the adjoint or the inverse

odl/operator/fn_ops.py
MatthiasJE: OK, how many character shall I use per line? Is there a strict upper limit in ODL? This one already uses 83 (> 79).

odl/operator/fn_ops.py
MatthiasJE: OK

odl/operator/fn_ops.py
MatthiasJE: Done

odl/operator/fn_ops.py
MatthiasJE: Sure, fixed

odl/operator/fn_ops.py
MatthiasJE: Fixed

odl/operator/fn_ops.py
MatthiasJE: Changed

odl/operator/fn_ops.py
MatthiasJE: Changed

MatthiasJE:
Thanks @adler-j for all the feedback. I changed the code accordingly. There are still some open questions:
+ pseudo inverse of sampling operator. This is tricky when a component is sampled more than once.
+ scaling of the sampling operator by the cell volume. I suppose we need to agree on that the sampling operator is supposed to represent. For me I am thinking of integrating over cells/pixels/voxel. For nearest neighbor interpolation it then results in the code that we have so far. Obviously it would be good if the code was interpolation independent.
+ scaling of the vectorization operator. Currently the scaling is used to avoid writing too much code as the adjoint and inverse are the same up to some scaling.

What do we do about these things?

MatthiasJE:
I now implemented your scaling option @adler-j. There is now an option to switch between the two modes.

MatthiasJE:
Any news about this @adler-j, are you happy about it?

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,454 @@
...
+
+    """Operator that creates an element of a vector space given its coefficients.
+    The operator is defined by:
+
+        ``EmbeddingOperator(X)(f) == X.element(f)``
+        
+    For the adjoint it is important that the element is scaled properly.
+    """
+
+    def __init__(self, range, scaling=1):
```
kohr-h: Slightly weird asymmetry that you need to give the range here. No need to fix it now, but both guys should have the flat space as an optional argument so you can use a `DiscreteLp` as a flat space, for example.
Write a TODO comment.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,454 @@
...
+        return out
+
+    @property
+    def scaling(self):
+        """scaling factor."""
+        return self.__scaling
+
+    @property
+    def adjoint(self):
+        """Adjoint of the embedding is vectorising with an appropriate scaling'.
```
kohr-h: leftover `'` at the end

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,454 @@
...
+        >>> x = X.element(range(X.size))
+        >>> ind = [[0, 1], [1, 1]]
+        >>> A = odl.SamplingOperator(X, ind)
+        >>> A(x)
+        rn(2).element([1.0, 3.0])
+        >>> A(x)  # Out-of-place
+        rn(2).element([1.0, 3.0])
+        """
+        if not isinstance(domain, (FnBase, DiscreteLp)):
+            raise TypeError('`space` {!r} not a `FnBase` or `DiscreteLp` '
```
kohr-h: domain

odl/operator/fn_ops.py
MatthiasJE: Thanks. I wasn't aware of that spyder feature but I like it a lot! It makes life so much easier.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,454 @@
...
+            strides = np.concatenate((np.cumprod(domain.shape[1:])[::-1], [1]))
+            self.__indices_flat = np.sum(self.indices * strides[:, None], axis=0)
+        else:
+            self.__indices_flat = self.indices
+
+        self.__weighting = weighting        
+        if self.weighting is None:
+            self.__weights = 1
+        elif self.weighting == 'cell_volume':
+            self.__weights = getattr(domain, 'cell_volume', 1.0)
```
MatthiasJE: I could do this, but then the call is more complicated as we need the if statement there then. I prefer it to be this way to be honest.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,454 @@
...
+        return repr(self)
+
+class EmbeddingOperator(Operator):
+
+    """Operator that creates an element of a vector space given its coefficients.
+    The operator is defined by:
+
+        ``EmbeddingOperator(X)(f) == X.element(f)``
+        
+    For the adjoint it is important that the element is scaled properly.
```
MatthiasJE: What do you mean?

MatthiasJE:
Thanks @kohr-h for all your comments. I think the file is now much, much better than before. Please have a look and let me know what you think.

odl/operator/fn_ops.py
kohr-h: Ah, I meant that this almost looks like a note to yourself during implementation. I wasn't sure what you want to say with this sentence.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,486 @@
...
+        --------
+        >>> X = odl.uniform_discr(min_pt=[0, 0], max_pt=[1, 1], shape=[2, 2])
+        >>> x = X.element(range(X.size))
+        >>> sampling_points = [[0, 1], [1, 1]]
+        >>> A = odl.SamplingOperator(X, sampling_points, 'point_eval')
+        >>> A(x)
+        rn(2).element([1.0, 3.0])
+        >>> A = odl.SamplingOperator(X, sampling_points, 'integrate')
+        >>> A(x)
+        rn(2).element([0.25, 0.75])
```
kohr-h: Top notch!

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,486 @@
...
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x)) < 1e-10
+        True
+        >>> A = odl.SamplingOperator(X, sampling_points, 'integrate')
+        >>> A.adjoint(A(x)).inner(x) - A(x).inner(A(x)) < 1e-10
+        True
+        """
+        if self.variant == 'point_eval':
+            variant = 'dirac'
+        elif self.variant == 'integrate':
+            variant = 'char_fun'
```
kohr-h: Oh right, that's actually a good way of seeing it.

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,486 @@
...
+
+    @property
+    def adjoint(self):
+        """Adjoint of this operation is flattening with an appropriate scaling,
+        see FlatteningOperator.
+
+        Examples
+        --------
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[2, 3])
+        >>> A = odl.EmbeddingOperator(X)
```
kohr-h: update

odl/operator/fn_ops.py
```diff
@@ -0,0 +1,486 @@
...
+        ----------
+        range : `LinearSpace` or `Field`
+            Space that the operator should map to.
+
+        Examples
+        --------
+        General usage example:
+
+        >>> X = odl.uniform_discr(min_pt=[-1, -1], max_pt=[1, 1], shape=[1, 2])
+        >>> A = odl.EmbeddingOperator(X)
```
kohr-h: update

kohr-h:
As far as I can see, all of @adler-j's comments have been addressed. Regarding the placement, I'll say it's okay for now since the operators are not totally general in the sense that they work on any linear space. In the long run they will probably end up in `tensor_ops.py` when the `fn` becomes `tensor_space`.
Since @adler-j is not available for response currently (if I remember correctly), I'll dismiss his review and merge.

kohr-h:
@MatthiasJE During work at the monster-PR #861 that will bring a lot of cool stuff with multi-indexing, I needed to adapt your code so it works with the new syntax (replacing `fn` and such). While at it I went over it and changed some things that I considered nitpicking back when merging this PR and that would have delayed the process unnecessarily.

So for you to study, I brought it a bit more in line with the style of the rest of ODL and fixed some minor things in the docstrings. The main changes are in the commit 3ab415a6879c6af422af07149b6201c71e6c3ce9. Here's a list, just take it as an exemplary style guide:

- I wrote a function to normalize the input indices and factored it out since both `SamplingOperator` and its adjoint use it.
- It is now allowed to pass a single index, in 1D as integer, in nD as single array. In 1D, several indices can be provided in a single array, no need to wrap into a sequence (I think that also worked before).
- In general, arguments are normalized ("defensive programming") and checked.
- ~~I chose not to store the linear indices in the forward version since the index arrays can be used directly without significant speed loss.~~ I measured times and it actually matters, so I put the linear indices back in, see a89ceb6841730751105ce3db5eab1d586bc18338.
- Fixed some minor issues with docs, for instance single-tick markup instead of double ticks. The former tries to create an internal link to a known name, e.g. another operator in the same module. Strings, parameter names and the likes should use double ticks, then they are just highlighted.
  Then there were summary lines spanning multiple lines, fixed that, too.
- The doctests now start simple in 1D, explain what's going on (e.g. with the weighting) and move from simple to more complicated. Also names like `X` for spaces are replaced (in general: variables are written in `snake_case`, i.e., lowercase with underscores).
- I think the `indices_flat` don't need to be public, they're an implementation detail. I removed them from the public interface.
- The `_call` methods don't really need the `out` parameter. The variant where `out` is given did exactly the same as the wrapping `__call__` of `Operator`, namely create a new `out` and assign to it. Therefore we can simply drop `out` here. (As an aside, I tested if `np.take(x, indices, out=out)` gives a speed benefit, but it's actually slower.)
- Minor thing: the "not yet supported" message for wrong `variant` "promises" too much, I replaced it with our standard formulation.
- In doctests: To make them well comprehensible, they should be as *visual* as possible, in particular one should avoid things like `space.element(range(1, space.size + 1))`. It's just a clever but much harder to read way to write `space.element([1, 2, 3, 4])` or so. We also write 2D arrays row by row with rows that are aligned as well as possible, to make them easy to read.
  For the same reason, it's good to start with simple cases (1D, default parameters) and find settings where the output is nice and readable (avoid `0.333333333`).
- For `__repr__` we have a nice tool to format the signature part from the arguments, the `signature_string` helper function. You can pass it the required arguments and the optional arguments with defaults, customize how to separate them (`', '` or `',\n'` basically) and add modifiers for printing. This is much more extensible than to manually add arguments to the printed string (it's also simple to forget by postponing, in particular when it's tedious work).
- In the `WeightedSumSamplingOperator`, I find it more natural to use the unscaled variant by default, like in `SamplingOperator`. The correspondence of adjoints goes criss-cross, but since this is an operator in its own right, the natural choice should be the default.
- In the `FlatteningOperator`, I used a trick that you may not have seen before, called *closure*. In this case it means that you can define a class within a method of an existing class and *use names that are defined in the enclosing scope*.
  It's the same as in this code snippet:
  ```python
  >>> c = 1
  >>> def f(x):
  ...     return x + c
  >>> f(2)
  3
  ```
  The function `f` can access its global namespace containing `c` since it is attached to the function object.
  In our case, I implement `FlatteningOperatorInverse` as an inner class since the code is actually very short *if you have access to the original operator*. This is possible here by storing the operator instance (`self` in the outer scope) under a different name (the `op = self` line), which makes it accessible also in the inner operator, where `self` has a different meaning. For instance, the `inverse` property of the inverse can just `return op`.
  In general, this technique can be used for `adjoint` or `inverse` or `derivative` operators that are simple enough to be implement in a couple of lines given the original operator, and which are not very interesting as operators in their own right.
- The adjoint is simply a scaled version of the inverse, largely as before.

MatthiasJE:
Looks very good, @kohr-h. Thanks for that.

PR #952, opened 2017-03-13
Issue 689  ipython spyder doctest

adler-j:


odl/util/testutils.py
```diff
@@ -679,6 +682,19 @@ def run_doctests(skip_if=False):
...
+    if run_from_ipython():
+        try:
+            import spyder
+        except ImportError:
+            pass
+        else:
+            if parse_version(spyder.__version__) < parse_version('3.1.4'):
+                warnings.warn('A bug with IPython and Spyder < 3.1.4 '
+                              'sometimes cause doctests to fail to run. '
+                              'Please upgrade spyder or use another '
```
kohr-h: Spyder

odl/util/utility.py
```diff
@@ -707,6 +708,15 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
     return part_sep.join(parts)
 
 
+def run_from_ipython():
+    """If the process is run from IPython."""
+    try:
+        __IPYTHON__
+        return True
+    except NameError:
+        return False
```
kohr-h: Why not `return '__IPYTHON__' in globals()`?

adler-j: Because globals :D

adler-j: Jokes aside, do you think it would be better?

kohr-h: But your implementation is 100 % equivalent.

kohr-h: Not sure if better, but shorter and more to the point. In the end, your version also tries to look up the name `'__IPYTHON__` in the global (module) dictionary, and if present, returns `True`, otherwise `False`. So unless I'm missing something regarding `globals()` I think it's the same.

kohr-h: To follow up, according to the discussion [here](https://stackoverflow.com/questions/12693606/reason-for-globals-in-python), `globals()` is (slightly misleadingly) "`locals()` at module level". So my guess was right.

adler-j:
Fixed comments and rebased, will merge

PR #955, opened 2017-03-13
ENH: Allow callable title and saveto in callbacks

adler-j:


odl/solvers/util/callback.py
```diff
@@ -440,13 +440,20 @@ def __init__(self, *args, **kwargs):
...
 
                 filename = saveto.format(cur_iter_num)
 
             where ``cur_iter_num`` is the current iteration number.
+
+            If saveto is a callable, the file name is generated as::
+
+                filename = saveto(cur_iter_num)
+
             If the directory name does not exist, a ``ValueError`` is raised.
```
kohr-h: What is this for anyway?

odl/solvers/util/callback.py
```diff
@@ -440,13 +440,20 @@ def __init__(self, *args, **kwargs):
             where ``cur_iter_num`` is the current iteration number.
         step : positive int, optional
             Number of iterations between plots. Default: 1
-        saveto : string, optional
+        saveto : string or callable, optional
             Format string for the name of the file(s) where
-            iterates are saved. The file name is generated as
+            iterates are saved.
+
+            If saveto is a string, the file name is generated as::
```
kohr-h: \`\`saveto\`\`

Leave a whitespace between "as" and "::", then the colon is not printed.

kohr-h: Leave a whitespace between "as" and "::", then the colon is not printed.

odl/solvers/util/callback.py
```diff
@@ -440,13 +440,20 @@ def __init__(self, *args, **kwargs):
...
 
                 filename = saveto.format(cur_iter_num)
 
             where ``cur_iter_num`` is the current iteration number.
+
+            If saveto is a callable, the file name is generated as::
```
kohr-h: Same here

odl/solvers/util/callback.py
```diff
@@ -482,8 +489,20 @@ def __init__(self, *args, **kwargs):
...
+            self.saveto = saveto.format
+        except TypeError:
+            self.saveto = saveto
+
+        title = kwargs.pop('title', 'Iterate {}')
+        try:
+            title = str(title + '')
+            self.title = title.format
+        except TypeError:
+            self.title = title
```
kohr-h: That's a nice idea, but I prefer to save the input parameters directly, to make repr and pickling easier. Either add attributes or move this logic to `__call__`.

odl/solvers/util/callback.py
```diff
@@ -482,8 +489,20 @@ def __init__(self, *args, **kwargs):
         self.kwargs = kwargs
         self.fig = kwargs.pop('fig', None)
         self.step = kwargs.pop('step', 1)
-        self.saveto = kwargs.pop('saveto', None)
-        self.title = kwargs.pop('title', 'Iterate {}')
+        saveto = kwargs.pop('saveto', None)
+        try:
+            saveto = str(saveto + '')
+            self.saveto = saveto.format
```
kohr-h: Append something like `_formatter`

odl/solvers/util/callback.py
```diff
@@ -482,8 +489,20 @@ def __init__(self, *args, **kwargs):
...
+        except TypeError:
+            self.saveto = saveto
+
+        title = kwargs.pop('title', 'Iterate {}')
+        try:
+            title = str(title + '')
+            self.title = title.format
```
kohr-h: `_formatter`

odl/solvers/util/callback.py
```diff
@@ -569,7 +588,13 @@ def __init__(self, saveto, step=1, impl='pickle', **kwargs):
         >>> callback = CallbackSaveToDisk(saveto='my_path/my_iterate_{}',
         ...                               step=5, impl='numpy')
         """
-        self.saveto = saveto
+        saveto = kwargs.pop('saveto', None)
+        try:
+            saveto = str(saveto + '')
+            self.saveto = saveto.format
+        except TypeError:
+            self.saveto = saveto
```
kohr-h: Same here

odl/solvers/util/callback.py
adler-j: We're explicitly saying we will not create the directory, I find it nice.

odl/solvers/util/callback.py
adler-j: Added an extra `title_formatter` attribute that solves this and the above comment.

adler-j:
Fixed the comments, needs a re-review.

odl/solvers/util/callback.py
kohr-h: The note is helpful, yes. But we do whatever matplotlib does, we shouldn't be too specific here I think. Perhaps change it to something like "Note that an error will be raised when trying to create a new directory."

odl/solvers/util/callback.py
```diff
@@ -483,7 +492,17 @@ def __init__(self, *args, **kwargs):
         self.fig = kwargs.pop('fig', None)
         self.step = kwargs.pop('step', 1)
         self.saveto = kwargs.pop('saveto', None)
+        try:
+            self.saveto_formatter = self.saveto.format
+        except AttributeError:
+            self.saveto_formatter = self.saveto
```
adler-j: It seems it was caused by upgrading scipy to 0.19, at least on my local machine it works for scipy 0.18, but fails for 0.19.

adler-j: Verified to work with 0.18.1 but failing with 0.19, all other packages kept constant.

Not quite sure how to handle this on our side.

kohr-h: Well if they decide to fix it we can blacklist scipy 0.19 for the time being. Although I suspect that this issue (amazing it hasn't come up earlier) can potentially pop up at any place any time.
Otherwise I guess what we can only either stop using `newint` or `np.isscalar`.

adler-j: My opinion is we stop using newint, I don't see where we need it, do you have a usecase?

We cannot stop using `np.isscalar` here since it is scipy calling it. 

Blacklisting scipy 0.19 is, imo, not an option given that this error only occurs in this really minor function. In that case I'd prefer an x_fail.

kohr-h: Unless we really have big ints at some place, which I don't see currently, abandoning `newint` shouldn't be a problem.

PR #956, opened 2017-03-13
ENH: Add FunctionalComp.gradient.derivative

adler-j:
Simple implementation for when it exists and we can "easily" find it.

### DIFF
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -576,6 +576,17 @@ class FunctionalComp(Functional, OperatorComp):
                 """Apply the gradient operator to the given point."""
                 return op.derivative(x).adjoint(func.gradient(op(x)))
 
+            def derivative(self, x):
+                """The derivative in point ``x``.
+
+                This is only defined
+                """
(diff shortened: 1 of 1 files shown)


odl/solvers/functional/functional.py
```diff
@@ -576,6 +576,17 @@ def _call(self, x):
...
                 return op.derivative(x).adjoint(func.gradient(op(x)))
 
+            def derivative(self, x):
+                """The derivative in point ``x``.
+
+                This is only defined
+                """
+                if not op.is_linear:
+                    raise NotImplementedError('derivative only implemented '
+                                              'linear opertors.')
```
kohr-h: "for" missing

adler-j:
Fixed comment, will merge after CI.

adler-j:
It seems I wont. Any idea what's up with the error message?

adler-j:
I've reproduced it locally, it is because travis upgraded scipy to 0.19, which apparently has a breaking change here, related to `future.newint`, see this issue at numpy: https://github.com/numpy/numpy/issues/8787

Issue #938, opened 2017-02-23
can I pass (and process) a complex data sinogram into fbp_op?


rshpeley:
I need a gpu based fbp algorithm to reconstruct from complex sinogram data. I've done this in matlab by passing complex floats and have got a complex reconstruction from it. I'd like to do this in odl running on the gpu.

Looking at the astra source code I don't see a way to use a complex sinogram as input to the reconstruction. It looks like fbp_op breaks out this functionality, but can I use complex data?

rshpeley:
This is the error I get when attempting with `ray_trafo.adjoint()`, so I guess not...

```python
/home/shpeley/miniconda3/lib/python3.6/site-packages/numpy/core/numeric.py:482: ComplexWarning: Casting complex values to real discards the imaginary part
  return array(a, dtype, copy=False, order=order)
```

Any suggestions? Rewrite astra?

kohr-h:
No, currently you have to manually apply the operators to real and imaginary parts separately. There would be an obvious way to wrap this in ODL, of course, but we haven't quite come around to doing that.

Given your complex sinogram, you can simply get `sinogram.real` and `sinogram.imag` as ODL vectors, and there's similar functionality to split the spaces.

adler-j:
A full example would be:

```python
import odl

# Create ray transform
space = odl.uniform_discr([-1, -1], [1, 1], [128, 128])
geometry = odl.tomo.parallel_beam_geometry(space)
rt = odl.tomo.RayTransform(space, geometry)

# Create complex data
data_space = rt.range.complex_space
data = data_space.element(lambda x: x[0] + 1j * x[1])

# Apply FBP
fbp = odl.tomo.fbp_op(rt)
image_real = fbp(data.real)
image_imag = fbp(data.imag)

# Visualize
image_real.show('real part')
image_imag.show('imaginary part')
```

Is this sufficient for you?

adler-j:
Another option is to create a FBP operator that works on complex values. The easiest way is by using the `RealPart`, `ImagPart` and `ComplexEmbedding` operators as such:

```python
import odl

# Create ray transform
space = odl.uniform_discr([-1, -1], [1, 1], [128, 128])
geometry = odl.tomo.parallel_beam_geometry(space)
rt = odl.tomo.RayTransform(space, geometry)

# Create complex data
data_space = rt.range.complex_space
data = data_space.element(lambda x: x[0] + 1j * x[1])

# Create complex FBP
[...]

PR #970, opened 2017-04-13
Issue 969  ray trafo complex

### LINKED ISSUE #969: RayTransform for complex spaces
The `RayTransform` operator should support complex spaces. Some thought needs to go into caching, but the functionality should be very simple to get right.

kohr-h:
Closes #969 

Also renames "scikit" -> "skimage". I felt "scikit" is such a generic name for anything that we should be more specific. "skimage" is the official module name for scikit-image.

TODO:
- [x] unit test for complex


odl/tomo/backends/astra_cpu.py
```diff
@@ -104,26 +105,47 @@ def astra_cpu_forward_projector(vol_data, geometry, proj_space, out=None):
     proj_id = astra_projector(vol_interp, vol_geom, proj_geom, ndim,
                               impl='cpu')
 
-    # Create ASTRA data structures
-    vol_id = astra_data(vol_geom, datatype='volume', data=vol_data,
-                        allow_copy=True)
+    if vol_data.space.is_rn:
```
adler-j: Strongly disagree with this implementation for two reasons:

* Double the implementation effort
* Double the memory overhead at runtime

A better way would (imo) be to do all of this inside the main "RayTransform" class and call the backends with real arrays in all cases.

kohr-h: Hm, right. Check out the new version. However, I'm not sure the memory overhead is really less. The `writable_array` function will make copies since real and imag parts are not contiguous as needed by Astra.

odl/tomo/operators/ray_trafo.py
```diff
@@ -214,26 +211,54 @@ def geometry(self):
...
     def _call(self, x, out=None):
         """Forward project ``x`` and store the result in ``out`` if given."""
+        if self.domain.is_rn:
+            return self._call_real_fp(x, out)
+
+        elif self.domain.is_cn:
+            result_parts = []
+            for part in ['real', 'imag']:
+                x_part = getattr(x, part)
+                out_part = getattr(out, part, None)
```
adler-j: won't this fail if out is None?

odl/tomo/operators/ray_trafo.py
```diff
@@ -214,26 +211,54 @@ def geometry(self):
...
+                result_parts.append(self._call_real_fp(x_part, out_part))
```
adler-j: Do we really want a for loop here? I'd probably prefer it unrolled for clarity

odl/tomo/operators/ray_trafo.py
```diff
@@ -386,28 +409,58 @@ def geometry(self):
...
+                out = self.range.element()
+                out.real = result_parts[0]
+                out.imag = result_parts[1]
+
+            return out
 
+        else:
+            raise RuntimeError('bad domain {!r}'.format(self.domain))
+
+    def _call_real_bp(self, x_real, out_real):
```
adler-j: how about simply `_call_real`?

examples/tomo/ray_trafo_parallel_2d_complex.py
```diff
@@ -0,0 +1,42 @@
...
+detector_partition = odl.uniform_partition(-30, 30, 558)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# Ray transform (= forward projection). The backend is set explicitly -
+# possible choices are 'astra_cpu', 'astra_cuda' and 'skimage'.
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='skimage')
+
+# Create a discretized phantom that is a Shepp-Logan phantom in the real
+# part and a cuboid in the imaginary part
+phantom = odl.phantom.shepp_logan(reco_space, modified=True)
```
adler-j: Perhaps:

```
phantom = (odl.phantom.shepp_logan(reco_space, modified=True) +
           1j * odl.phantom.cuboid(reco_space))
```

examples/tomo/ray_trafo_parallel_2d_complex.py
```diff
@@ -0,0 +1,42 @@
...
+phantom.imag = odl.phantom.cuboid(reco_space.real_space)
+
+# Create projection data by calling the ray transform on the phantom
+proj_data = ray_trafo(phantom)
+
+# Back-projection can be done by simply calling the adjoint operator on the
+# projection data (or any element in the projection space).
```
adler-j: Perhaps modify this to mention how this "works" for complex data

examples/tomo/ray_trafo_parallel_2d_complex.py
```diff
@@ -0,0 +1,42 @@
...
+separately.
+"""
+
+import numpy as np
+import odl
+
+# Reconstruction space: discretized complex-valued functions on the rectangle
+# [-20, 20]^2 with 300 samples per dimension.
+reco_space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='complex128')
```
adler-j: `complex128` -> `complex`, easier to read

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -229,5 +235,23 @@ def test_angles(projector):
     assert almost_equal(np.fmod(maximum_angle, np.pi), np.pi / 2, places=1)
 
 
+def test_complex(impl):
```
adler-j: You could also add `complex` as a dtype for the largescale tests

odl/tomo/operators/ray_trafo.py
```diff
@@ -101,25 +101,22 @@ def __init__(self, discr_domain, geometry, **kwargs):
...
                 raise ValueError('no valid `impl` installed')
 
         impl, impl_in = str(impl).lower(), impl
         if impl not in _SUPPORTED_IMPL:
-            raise ValueError('`impl` {!r} not supported'
-                             ''.format(impl_in))
+            raise ValueError('`impl` {!r} not understood'.format(impl_in))
+        impl_available = globals()[impl.upper() + '_AVAILABLE']
+        if not impl_available:
+            raise ValueError('{!r} back-end not available'.format(impl))
```
adler-j: Perhaps a word on what happens if NO backends are available?

odl/tomo/operators/ray_trafo.py
kohr-h: Yeah, I've rewritten it without loop. The original was a mix of copy from the old implementation and some show-off with `getattr` Kung-Fu. Much simpler now.

odl/tomo/operators/ray_trafo.py
kohr-h: Done

odl/tomo/operators/ray_trafo.py
kohr-h: No, but now it's obsolete.

odl/tomo/operators/ray_trafo.py
kohr-h: Added

examples/tomo/ray_trafo_parallel_2d_complex.py
kohr-h: Actually I wanted `'complex64'` to match the real space example, this was only to test if other dtypes work, too.

examples/tomo/ray_trafo_parallel_2d_complex.py
kohr-h: Yeah, easier on the eyes.

examples/tomo/ray_trafo_parallel_2d_complex.py
kohr-h: OK

examples/tomo/ray_trafo_parallel_2d_complex.py
```diff
@@ -0,0 +1,45 @@
+"""Example computing the 2d parallel beam ray transform of a complex phantom.
+
+This example demonstrates that complex-valued functions work for both
+forward and back projections by application to real and imaginary parts
+separately.
```
adler-j: The "separately" part here implies that this is in some sense done by the user, or in a way visible to the user. I'd simply say:

"This example demonstrates that forward and back projections work with complex-valued functions."""

examples/tomo/ray_trafo_parallel_2d_complex.py
```diff
@@ -0,0 +1,45 @@
...
+# possible choices are 'astra_cpu', 'astra_cuda' and 'skimage'.
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='skimage')
+
+# Create a discretized phantom that is a Shepp-Logan phantom in the real
+# part and a cuboid in the imaginary part
+phantom = (odl.phantom.shepp_logan(reco_space, modified=True) +
+           1j * odl.phantom.cuboid(reco_space))
+
+# Create projection data by calling the ray transform on the phantom.
+# This is equivalent to evaluating the real-space ray transform on the
```
adler-j: remove "real-space" adds little if anything but feels confusing

odl/tomo/operators/ray_trafo.py
```diff
@@ -31,180 +31,227 @@
...
 
         if not isinstance(geometry, Geometry):
-            raise TypeError('`geometry` {!r} is not a `Geometry` instance'
-                            ''.format(geometry))
-
+            raise TypeError('`geometry` must be a `Geometry` instance, got '
+                            '{!r}'.format(geometry))
+
+        # Handle backend choice
+        if not any(globals()[imp.upper() + '_AVAILABLE']
```
adler-j: do we truly need this implementation? If we cant do anything else, then fine, but I'd rather not.

kohr-h: As in "this type of code" or "this check at all"?

examples/tomo/ray_trafo_parallel_2d_complex.py
kohr-h: ok, will do

examples/tomo/ray_trafo_parallel_2d_complex.py
kohr-h: Right, this was before the comments were improved.

odl/tomo/operators/ray_trafo.py
adler-j: Even though it is 2 AM, `global()` hurts my eyes :D

kohr-h: Hehe, truly painful, I actually agree. How about replacing it with `eval(imp.upper() + '_AVAILABLE')`? Just kidding.
But what's the alternative? I think only manual check of each backend.

adler-j: I would frankly prefer an explicit list somewhere and then just check if it's empty

kohr-h:
> Also, does this fix #938?

I just added an example for this case, and indeed, it just works.

Will merge after CI.

examples/tomo/filtered_backprojection_parallel_2d_complex.py
```diff
@@ -0,0 +1,70 @@
+"""Example creating a filtered back-projection (FBP) in 2d for complex data.
```
adler-j: This is a "manual" FBP, how does it work with `fbp_op`?

kohr-h: Just tested it, works as well.

adler-j: Perhaps use it in the example instead?

kohr-h: OK, done

PR #971, opened 2017-04-13
BUG: fix min/max_pt init with dict

kohr-h:
Simple fix for a bug in the normalization of the input dicts. The bug was triggered by CPython 3.6 preserving the insertion order of dicts (not guaranteed though, implementation detail).

### DIFF
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -974,15 +974,16 @@ def uniform_partition_fromgrid(grid, min_pt=None, max_pt=None):
     >>> part.cell_boundary_vecs[1]
     array([-0.25,  0.25,  0.75,  3.  ])
     """
-    # Make dictionaries from min_pt and max_pt and fill with None where no
-    # value is given.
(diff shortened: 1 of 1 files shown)


odl/discr/partition.py
```diff
@@ -982,15 +982,17 @@ def uniform_partition_fromgrid(grid, min_pt=None, max_pt=None):
         min_pt = np.atleast_1d(min_pt)
         min_pt = {i: float(v) for i, v in enumerate(min_pt)}
     else:
-        min_pt.update({i: None for i in range(grid.ndim) if i not in min_pt})
+        min_pt.update({i: None for i in range(grid.ndim)
+                       if i not in min_pt and i - grid.ndim not in min_pt})
```
adler-j: I don't quite follow what this does, add a comment above

kohr-h: Well the comment farther above explains what happens, clarified it a bit now. Basically the new code does what the old comment says, in contrast to the old code :-)

PR #974, opened 2017-04-13
BLD: change to no-sudo Travis build, closes #927

### LINKED ISSUE #927: Stop using sudo in .travis.yml
Instead of `sudo apt-get package` we should do
```yaml
sudo: false
addons:
    apt:
        package1
        package2
```
I heard rumors that it's faster, too.

kohr-h:
WIP PR to check if this works and how fast it runs. Allegedly faster because docker is used in case of `sudo:false`, but I have to unconditionally install the LaTeX packages.. we'll see.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,69 +1,70 @@
 language: python
+
 python:
   - "2.7"
   - "3.5"
 
+sudo: false
+
+addons:
+    apt:
+        packages:
+            libfftw3-dev
+            texlive-latex-base
+            texlive-latex-recommended
+            texlive-latex-extra
+            dvipng
+
+# Enable cache for $HOME/.cache/pip (wheels and such)
+cache: pip
+
 # Setup numpy + scipy using miniconda
 # See http://conda.pydata.org/docs/travis.html
 install:
-  - sudo apt-get update
+    # Install miniconda according to Python version of the build (saves downloading if versions match)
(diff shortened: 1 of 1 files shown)


adler-j:
Just make sure the build runs, then I'll review.

kohr-h:
Works now. It doesn't really run faster per build - in fact it's slower for Python 2.7 since I now pull the LaTeX packages unconditionally. But it has a much better upstart time (hardly any delay between push and build), and worker availability seems to be better.
I also experimented a bit with caching. Enabling cache for `pip` brings some speedup (~20 s), in particular for pyfftw and pywavelets become faster. Caching miniconda, on the other hand, cuts down total time by meager 4 seconds. I disabled it again.

.travis.yml
```diff
@@ -1,69 +1,70 @@
...
-  - conda update -q conda
-  # Useful for debugging any issues with conda
-  - conda info -a
+# Enable cache for $HOME/.cache/pip (wheels and such)
+cache: pip
 
-  # Install dependencies and enter test environment
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy future nomkl sphinx sphinx_rtd_theme
-  - source activate test-environment
+# Setup numpy + scipy using miniconda
```
kohr-h: From here down no changes except `py.test` -> `pytest`. Just messed up by the GH diff.

PR #976, opened 2017-04-18
ENH: allow anisotropic voxels in 3d

kohr-h:
This is partially supported in upstream - 3d works for all geometries, 2d still fails.

### DIFF
--- a/examples/tomo/ray_trafo_parallel_3d.py
+++ b/examples/tomo/ray_trafo_parallel_3d.py
@@ -9,17 +9,11 @@ reco_space = odl.uniform_discr(
     min_pt=[-20, -20, -20], max_pt=[20, 20, 20], shape=[300, 300, 300],
     dtype='float32')
 
-# Make a parallel beam geometry with flat detector
+# Make a 3d single-axis parallel beam geometry with flat detector
 # Angles: uniformly spaced, n = 180, min = 0, max = pi
 angle_partition = odl.uniform_partition(0, np.pi, 180)
 # Detector: uniformly sampled, n = (558, 558), min = (-30, -30), max = (30, 30)
 detector_partition = odl.uniform_partition([-30, -30], [30, 30], [558, 558])
-# Discrete reconstruction space
-
-# Some versions of ASTRA cannot handle axis-aligned origin_to_det unless
-# it is aligned with the third coordinate axis, see issue #18 at ASTRA's
-# GitHub. This is fixed in ASTRA v1.7.2.; with older versions, the following
-# may produce a zero result.
 geometry = odl.tomo.Parallel3dAxisGeometry(angle_partition, detector_partition)
 
 # Ray transform (= forward projection).
@@ -38,6 +32,6 @@ backproj = ray_trafo.adjoint(proj_data)
 # Show the slice z=0 of phantom and backprojection, as well as a projection
 # image at theta=0 and a sinogram at v=0 (middle detector row)
 phantom.show(coords=[None, None, 0], title='Phantom, middle z slice')
+backproj.show(coords=[None, None, 0], title='Back-projection, middle z slice')
 proj_data.show(coords=[0, None, None], title='Projection at theta=0')
 proj_data.show(coords=[None, None, 0], title='Sinogram, middle slice')
-backproj.show(coords=[None, None, 0], title='Back-projection, middle z slice')
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -211,7 +211,7 @@ class DiscreteLp(DiscretizedSpace):
 
     @property
     def is_uniform(self):
-        """``True`` if ``self.partition`` is uniform."""
+        """``True`` if `partition` is uniform."""
         return self.partition.is_uniform
 
     @property
--- a/odl/discr/partition.py
+++ b/odl/discr/partition.py
@@ -229,8 +229,26 @@ class RectPartition(object):
 
     @property
     def is_uniform(self):
-        """``True`` if ``self.grid`` is uniform."""
+        """``True`` if `grid` is uniform."""
         return self.grid.is_uniform
+
+    @property
+    def has_isotropic_cells(self):
+        """``True`` if `grid` is uniform and `cell sides` are all equal.
+
+        Always ``True`` for 1D partitions.
+
+        Examples
+        --------
+        >>> part = uniform_partition([0, -1], [1, 1], (5, 10))
+        >>> part.has_isotropic_cells
+        True
+        >>> part = uniform_partition([0, -1], [1, 1], (5, 5))
+        >>> part.has_isotropic_cells
+        False
+        """
+        return self.is_uniform and np.allclose(self.cell_sides[:-1],
+                                               self.cell_sides[1:])
 
     @property
     def ndim(self):
--- a/odl/test/tomo/operators/ray_trafo_test.py
+++ b/odl/test/tomo/operators/ray_trafo_test.py
@@ -45,7 +45,7 @@ def geometry(request):
         return tomo.Parallel2dGeometry(apart, dpart)
     elif geom == 'par3d':
         apart = odl.uniform_partition(0, np.pi, n_angles)
-        dpart = odl.uniform_partition([-30] * 2, [30] * 2, [m] * 2)
+        dpart = odl.uniform_partition([-30, -30], [30, 30], (m, m))
         return tomo.Parallel3dAxisGeometry(apart, dpart)
     elif geom == 'cone2d':
         apart = odl.uniform_partition(0, 2 * np.pi, n_angles)
@@ -54,12 +54,12 @@ def geometry(request):
                                     det_radius=100)
     elif geom == 'cone3d':
         apart = odl.uniform_partition(0, 2 * np.pi, n_angles)
-        dpart = odl.uniform_partition([-60] * 2, [60] * 2, [m] * 2)
+        dpart = odl.uniform_partition([-60, -60], [60, 60], (m, m))
         return tomo.CircularConeFlatGeometry(apart, dpart, src_radius=200,
                                              det_radius=100)
     elif geom == 'helical':
         apart = odl.uniform_partition(0, 8 * 2 * np.pi, n_angles)
-        dpart = odl.uniform_partition([-30, -3], [30, 3], [m] * 2)
+        dpart = odl.uniform_partition([-30, -3], [30, 3], (m, m))
         return tomo.HelicalConeFlatGeometry(apart, dpart, pitch=5.0,
                                             src_radius=200, det_radius=100)
     else:
@@ -272,7 +272,7 @@ def test_complex(impl):
     assert all_almost_equal(data.imag, true_data_im)
 
 
-def test_aniso_voxels(geometry):
+def test_anisotropic_voxels(geometry):
     """Test projection and backprojection with anisotropic voxels."""
     ndim = geometry.ndim
     shape = [10] * (ndim - 1) + [5]
--- a/odl/test/tomo/operators/ray_trafo_test.py
+++ b/odl/test/tomo/operators/ray_trafo_test.py
@@ -29,6 +29,42 @@ impl_params = [skip_if_no_astra('astra_cpu'),
                skip_if_no_skimage('skimage')]
 impl = simple_fixture('impl', impl_params, fmt=" {name} = '{value.args[1]}' ")
 
+geometry_params = ['par2d', 'par3d', 'cone2d', 'cone3d', 'helical']
+geometry_ids = [' geometry = {} '.format(p) for p in geometry_params]
+
+
+@pytest.fixture(scope='module', ids=geometry_ids, params=geometry_params)
+def geometry(request):
+    geom = request.param
+    m = 100
+    n_angles = 100
+
+    if geom == 'par2d':
+        apart = odl.uniform_partition(0, np.pi, n_angles)
+        dpart = odl.uniform_partition(-30, 30, m)
+        return tomo.Parallel2dGeometry(apart, dpart)
+    elif geom == 'par3d':
+        apart = odl.uniform_partition(0, np.pi, n_angles)
+        dpart = odl.uniform_partition([-30] * 2, [30] * 2, [m] * 2)
+        return tomo.Parallel3dAxisGeometry(apart, dpart)
+    elif geom == 'cone2d':
+        apart = odl.uniform_partition(0, 2 * np.pi, n_angles)
+        dpart = odl.uniform_partition(-30, 30, m)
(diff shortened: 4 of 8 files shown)


examples/tomo/ray_trafo_parallel_3d.py
```diff
@@ -16,10 +16,10 @@
...
 # Discrete reconstruction space
 
-# Astra cannot handle axis aligned origin_to_det unless it is aligned
-# with the third coordinate axis. See issue #18 at ASTRA's github.
-# This is fixed in new versions of astra, with older versions, this could
-# give a zero result.
+# Some versions of ASTRA cannot handle axis-aligned origin_to_det unless
+# it is aligned with the third coordinate axis, see issue #18 at ASTRA's
+# GitHub. This is fixed in ASTRA v1.7.2.; with older versions, the following
+# may produce a zero result.
```
adler-j: Perhaps we should give a warning in this case?

examples/tomo/ray_trafo_parallel_3d_aniso_voxels.py
```diff
@@ -0,0 +1,47 @@
+"""Example for ray transform with 3d parallel beam and anisotropic voxels.
+
+Anisotropic voxels are supported in ASTRA v1.8 and upwards; lower versions
```
adler-j: Lower -> earlier

examples/tomo/ray_trafo_parallel_3d.py
adler-j: I.e. move this to RayTransform and remove the comment here

examples/tomo/ray_trafo_parallel_3d_aniso_voxels.py
```diff
@@ -0,0 +1,47 @@
...
+# Back-projection can be done by simply calling the adjoint operator on the
+# projection data (or any element in the projection space).
+backproj = ray_trafo.adjoint(proj_data)
+
+# Show the slice y=0 of phantom and backprojection, as well as a projection
+# image at theta=0 and a sinogram at v=0 (middle detector row)
+phantom.show(coords=[None, 0, None], title='Phantom, middle y slice')
+proj_data.show(coords=[0, None, None], title='Projection at theta=0')
+proj_data.show(coords=[None, None, 0], title='Sinogram, middle slice')
+backproj.show(coords=[None, 0, None], title='Back-projection, middle y slice')
```
adler-j: Order these in the same order as they are presented above.

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -39,6 +39,42 @@
...
+    m = 100
+    n_angles = 100
+
+    if geom == 'par2d':
+        apart = odl.uniform_partition(0, np.pi, n_angles)
+        dpart = odl.uniform_partition(-30, 30, m)
+        return tomo.Parallel2dGeometry(apart, dpart)
+    elif geom == 'par3d':
+        apart = odl.uniform_partition(0, np.pi, n_angles)
+        dpart = odl.uniform_partition([-30] * 2, [30] * 2, [m] * 2)
```
adler-j: [m, m] is more clear than [m] * 2

odl/tomo/backends/astra_setup.py
```diff
@@ -74,6 +74,70 @@
...
+
+
+def astra_supports(feature):
+    """Return bool indicating whether current ASTRA supports ``feature``.
+
+    Parameters
+    ----------
+    feature : str
+        String standing for a potential feature of ASTRA.
+        See ``ASTRA_FEATURES`` for possible values.
```
adler-j: Single ticks

odl/tomo/backends/astra_setup.py
```diff
@@ -74,6 +74,70 @@
...
+                  }
+
+
+def astra_supports(feature):
+    """Return bool indicating whether current ASTRA supports ``feature``.
+
+    Parameters
+    ----------
+    feature : str
+        String standing for a potential feature of ASTRA.
```
adler-j: Name of ...

odl/tomo/backends/astra_setup.py
```diff
@@ -74,6 +74,70 @@
...
+# Dictionary specifying which ASTRA versions support which feature. The dict
+# values can be given as exact version '==1.7' or '1.7', as range '1.7-1.7.4',
+# as inequality '>=1.7' or as comma-separated composition of such (the items
+# will be OR-ed together), e.g. '<=1.7, >=1.7.2'.
+ASTRA_FEATURES = {'anisotropic_voxels_2d': None,
+                  'anisotropic_voxels_3d': '>=1.8',
+                  }
+
+
+def astra_supports(feature):
```
adler-j: Do we truly need this much code for this? Isn't there some built In feature? If so, use it.  Else move the generic parts of this to the util library.

odl/tomo/backends/astra_setup.py
```diff
@@ -117,14 +181,16 @@ def astra_volume_geometry(discr_reco):
...
-        # values for the volume extent, but projector creation fails
+        # values for the volume extent, but running the algorithm fails
         # if voxels are non-isotropic. We raise an exception here in
         # the meanwhile.
-        if not np.allclose(discr_reco.partition.cell_sides[1:],
-                           discr_reco.partition.cell_sides[:-1]):
-            raise NotImplementedError('non-isotropic voxels not supported by '
-                                      'ASTRA')
-        # given a 2D array of shape (x, y), a volume geometry is created as:
+        if (not np.allclose(discr_reco.partition.cell_sides[1:],
```
adler-j: We should add something like "is_isotropic_cell_sides" to partition instead of this.

odl/tomo/backends/astra_setup.py
kohr-h: I very much doubt there is something like this in ASTRA. And although this is a bit of code, I think it makes it easier to specify features of versions, compared to the alternative (that I started with) of hard-coding variables like
```
ASTRA_SUPPORTS_ANISO_VOXELS_3D = (
    parse_version(ASTRA_VERSION) >= parse_version('1.8'))
```
which becomes messy with broken intermediate versions like 1.7.1 which screwed up axis-aligned `origin_to_det`, for instance.

I'll see if I can move it to a util module, however it should still be under `backends/`.

adler-j: What I mean is that there is a generic component to this, I.e. "comparing versions", that should be in odl.util, this function can then stay but should call that function.

kohr-h: Ah I see. Dunno, I'll scan `pkg_resources`, `setuptools` and the likes.

kohr-h: Found `parse_requirements` from `pkg_resources`. Now the utility function is in `odl.util.utility`.

odl/tomo/backends/astra_setup.py
kohr-h: That produces a broken link, unfortunately, since module constants are not available for linking.

odl/test/tomo/operators/ray_trafo_test.py
kohr-h: Hm, that was largely copy-paste, but yeah I'll change it.

examples/tomo/ray_trafo_parallel_3d.py
kohr-h: What exactly was the issue here? Did this occur only in parallel beam 3d? And does axis-aligned mean the case where the rotation axis was the same as the vector from origin to detector at start?
Wouldn't that be a silly setup?

odl/tomo/backends/astra_setup.py
```diff
@@ -64,16 +63,50 @@
...
 
+# Dictionary specifying which ASTRA versions support which feature. The dict
+# values must be specified as valid setuptools package requirements without
+# package name, e.g., as exact version '==1.7', as inequality '>=1.7' or as
+# several requirements that need to be satisfied simultaneously, e.g.,
+# '>=1.8,<=2.0'.
+# To give multiple requirements that should be OR-ed together, use a
+# sequence instead of a single string in the dictionary, e.g.,
+# ['==1.7', '>=1.8,<=2.0'].
+ASTRA_FEATURES = {'anisotropic voxels 2d': None,
```
adler-j: I would prefer if these labels were given in snake case, `anisotropic_voxels_2d`, simply because it is more clear that they belong together.

kohr-h: Really weird, I had the same feeling while changing it ... I'll revert that change.

odl/tomo/operators/ray_trafo.py
```diff
@@ -146,6 +149,21 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+
+            # Filter axis-alinged initial detector position in parallel 3d
+            # geometry since it's broken for some ASTRA versions
+            if isinstance(geometry, Parallel3dAxisGeometry):
+                axis = geometry.axis
+                init_vec = (geometry._det_init_pos /
+                            np.linalg.norm(geometry._det_init_pos))
+                # Almost the same
+                if (np.isclose(np.dot(axis, init_vec), 1) and
+                        not astra_supports('par3d_axis_aligned_det_init_pos')):
```
kohr-h: This check is not correct. I just installed ASTRA 1.7.1 which shows the issue with the default setup in [the 3d parallel beam example](https://github.com/odlgroup/odl/blob/master/examples/tomo/ray_trafo_parallel_3d.py). Here it is `axis = [0, 0, 1]` and `det_init_pos = `[1, 0, 0]`.
So `axis_aligned_det_init_pos` (taken from the comment in the example) is the wrong name. What is the exact fault scenario? Do you remember, @adler-j? The ASTRA issue is [here](https://github.com/astra-toolbox/astra-toolbox/issues/18) for reference.

adler-j: I do not, but we can reverse engineer it via the fix: 

https://github.com/astra-toolbox/astra-toolbox/commit/86ed34e9a5fa408c9338cd5c2871f7f7953806b7

kohr-h: Okay, I just did that. The exact error scenario is that `geometry.det_to_src(angle, mid_pt)` in [this line](https://github.com/odlgroup/odl/blob/4d886be4cf761eb297e257b5487abc7634ba5754/odl/tomo/backends/astra_setup.py#L304) has 0 component along `geometry.axis`, i.e., when the two are perpendicular. In that case, [this line](https://github.com/astra-toolbox/astra-toolbox/blob/86ed34e9a5fa408c9338cd5c2871f7f7953806b7/src/CompositeGeometryManager.cpp#L228) in ASTRA, and in particular [this division](https://github.com/astra-toolbox/astra-toolbox/blob/71db6331f8dd0d5abbeee92977af01293be4f427/src/GeometryUtil3D.cpp#L171) in `backprojectPointX` yields +-infinity or NaN(?) (the TODO above the function is telling :-)). The `fRayX` member refers to the [first column of the vectors in the vec geometries](https://github.com/astra-toolbox/astra-toolbox/blob/71db6331f8dd0d5abbeee92977af01293be4f427/src/ParallelVecProjectionGeometry3D.cpp#L105), which I checked will be zero in the error case.

kohr-h:
All comments addressed, maybe another quick look won't hurt (and approval is needed anyway).

examples/tomo/ray_trafo_parallel_3d_aniso_voxels.py
```diff
@@ -0,0 +1,41 @@
+"""Example for ray transform with 3d parallel beam and anisotropic voxels.
```
adler-j: Rename the file to anisotropic_voxels, more clear that way

odl/discr/lp_discr.py
```diff
@@ -211,10 +211,15 @@ def is_uniform_byaxis(self):
 
     @property
     def is_uniform(self):
-        """``True`` if ``self.partition`` is uniform."""
+        """``True`` if `partition` is uniform."""
         return self.partition.is_uniform
 
     @property
+    def has_isotropic_cells(self):
```
adler-j: Not quite sure if we need this passthrough, accessing the partition directly should be ok

odl/discr/partition.py
```diff
@@ -229,10 +229,26 @@ def is_uniform_byaxis(self):
 
     @property
     def is_uniform(self):
-        """``True`` if ``self.grid`` is uniform."""
+        """``True`` if `grid` is uniform."""
         return self.grid.is_uniform
 
     @property
+    def has_isotropic_cells(self):
+        """``True`` if `grid` is uniform and `cell sides` are all equal.
```
adler-j: A word on the 1d case

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -246,5 +272,38 @@ def test_complex(impl):
     assert all_almost_equal(data.imag, true_data_im)
 
 
+def test_aniso_voxels(geometry):
```
adler-j: write out anisotropic

odl/tomo/operators/ray_trafo.py
```diff
@@ -146,6 +149,27 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+                axis = geometry.axis
+                mid_pt = geometry.det_params.mid_pt
+                for i, angle in enumerate(geometry.angles):
+                    if abs(np.dot(axis,
+                                  geometry.det_to_src(angle, mid_pt))) < 1e-4:
+                        warnings.warn(
+                            'angle {}: detector midpoint normal {} is '
+                            'perpendicular to the geometry axis {} in '
+                            '`Parallel3dAxisGeometry`; this is broken in '
+                            'ASTRA v{}, please upgrade'
```
adler-j: ASTRA v{} < [NEEDED VERSION]

odl/tomo/backends/astra_setup.py
```diff
@@ -44,7 +43,7 @@
                                   str(astra_ver_num % 100)])
 
     if parse_version(ASTRA_VERSION) < parse_version('1.7'):
```
adler-j: Perhaps this can be made a special case of the general astra supports stuff?

odl/discr/lp_discr.py
kohr-h: It's definitely less important than `is_uniform`, although related. I'll remove it then.

odl/tomo/backends/astra_setup.py
kohr-h: Not sure how to do this meaningfully. If I add `astra_supports('odl')` or so, it's not accurate since *we* decide not to support versions older than some number. I think it's okay to hard-code here since it's the only place. Later on it will still be easy to drop support for more old versions.

kohr-h:
Done, will merge when CI is happy

PR #977, opened 2017-04-18
REL: change license to MPL-2.0, closes #967

### LINKED ISSUE #967: Change ODL licence to Mozilla Public License, version 2.0
Bearing in mind the discussions we have had with partners, we should change the ODL licence from the current one, which is [GNU General Public License v3.0](http://www.gnu.org/licenses/gpl-3.0.en.html), to [Mozilla Public License, version 2.0](http://www.mozilla.org/en-US/MPL/2.0/). The motivation for choosing such a weak copyleft license are the following:

- We welcome commercial usage of ODL and want to minimize obstacles for such usage.
- We appreciate others forking and improving our work.
- We want to prevent license-incompatible forks of ODL, be it proprietary versions or versions under a stronger copyleft license.

Due to the last point, permissive licenses are of less interest, and the first point excludes strong copyleft licenses like GPL.

kohr-h:
Here we go, nothing interesting code-wise. I cleaned up a few test files and fixed some docs, but nothing big.

I guess this would require a new release -- version 0.6.0?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,40 +3,44 @@ How to contribute
 
 You're reading this, so you actually so you actually want to contribute. Good!
 
-This is some quick information to get you started. At the bottom you find a list of useful links to get you further.
+This is some quick information to get you started. At the bottom you find a list of useful links for further reference.
 
 As a user
 ---------
+If you are using the library and find a bug in the code, an error in the [documentation](https://odl.readthedocs.org/index.html), or simply have a question:
 
-If you are using the library and find a bug in the code, an error in the [documentation](https://odl.readthedocs.org/index.html), would like to have a feature which does not exist currently, or simply have a question:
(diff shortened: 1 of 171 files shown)


ozanoktem:
Yes this requires a new release.

adler-j:
Please fix the failing tests and I'll give this a look

kohr-h:
Tests fixed, ready for review.

kohr-h:
Btw, the failing tests were triggered by deleted `from __future__ import division`, maybe we want to have that in all test files to prevent future failures?

kohr-h:
> Looks fine to me, frankly I'll not verify that you changed in all places and simply trust that you did.

I hope, too :-)
But grepping for GNU or GPL doesn't yield any results, so at least there's no old header left in. At maximum some files without header.



> General import changes etc are good.

So I'll go ahead and add `from __future__ import division` to the test files, and after that I'll merge.

PR #981, opened 2017-04-19
BUG: fix multiple title args in CallbackShow

kohr-h:
"Emergency bugfix" for `CallbackShow` which raised an exception for some of the examples. The `*args` mess up everything.

This needs to go in before the version bump.

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -482,25 +482,25 @@ class CallbackShow(SolverCallback):
     odl.space.base_ntuples.NtuplesBaseVector.show
     """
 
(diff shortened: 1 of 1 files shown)


odl/solvers/util/callback.py
```diff
@@ -516,11 +516,10 @@ def __init__(self, *args, **kwargs):
 
             If the directory name does not exist, a ``ValueError`` is raised.
             If ``saveto is None``, the figures are not saved.
-            Default: ``None``
 
         Other Parameters
         ----------------
-        kwargs :
+        args, kwargs :
```
kohr-h: no args

PR #982, opened 2017-04-19
BUG: fence pytest import in tomo testutils

kohr-h:
Another thing that needs last-minute fix. Import in a new environment without `pytest` installed would fail.

### DIFF
--- a/odl/tomo/util/testutils.py
+++ b/odl/tomo/util/testutils.py
@@ -16,14 +16,8 @@ __all__ = ('skip_if_no_astra', 'skip_if_no_astra_cuda', 'skip_if_no_skimage')
 try:
     import pytest
 except ImportError:
-    # Make trivial decorator
-    from odl.util import OptionalArgDecorator
-
-    class ident(OptionalArgDecorator):
-        @staticmethod
-        def _wrapper(f, *a, **kw):
-            return f
-
+    # Use the identity decorator (default of OptionalArgDecorator)
+    from odl.util import OptionalArgDecorator as ident
     skip_if_no_astra = skip_if_no_astra_cuda = skip_if_no_skimage = ident
 
 else:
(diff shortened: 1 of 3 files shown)


odl/tomo/util/testutils.py
```diff
@@ -11,17 +11,25 @@
...
-
-skip_if_no_skimage = pytest.mark.skipif("not odl.tomo.SKIMAGE_AVAILABLE",
-                                        reason='skimage not available')
+try:
+    import pytest
+except ImportError:
+    # Make trivial decorator
+    from odl.util import OptionalArgDecorator
+
+    class ident(OptionalArgDecorator):
```
adler-j: Why would we need to overwrite this, shouldn't this be the default implementation?

kohr-h: You mean the `_wrapper`? That would make sense, yes.

kohr-h:
Needs approval

odl/tomo/util/testutils.py
adler-j: Yeah the _wrapper

odl/tomo/util/testutils.py
```diff
@@ -11,17 +11,23 @@
...
 
+try:
+    import pytest
+except ImportError:
+    # Make trivial decorator
+    from odl.util import OptionalArgDecorator
 
-skip_if_no_astra = pytest.mark.skipif("not odl.tomo.ASTRA_AVAILABLE",
-                                      reason='ASTRA not available')
+    class ident(OptionalArgDecorator):
```
adler-j: Why even do this? Instead do 

```
from odl.util import OptionalArgDecorator as ident
```

Fix in other places as well

kohr-h: Ha!

PR #987, opened 2017-04-26
BUG: Fix a bug with MultiplyOperator.adjoint for scalars

ozanoktem:


odl/operator/default_ops.py
```diff
@@ -298,17 +298,24 @@ def adjoint(self):
         >>> op.adjoint(x)
         rn(3).element([1.0, 4.0, 9.0])
 
-        Multiply by a scalar:
+        Multiply a scalar by vector:
 
         >>> op2 = MultiplyOperator(x, domain=r3.field)
         >>> op2.adjoint(x)
         14.0
+
```
kohr-h: Try to convince your editor to remove trailing whitespace when saving.
In Spyder, this is under Tools -> Preferences -> Editor -> Advanced settings -> Automatically remove trailing spaces when saving files

Also run `pytest --pep8 odl/operator/default_ops.py` to run the doctests in that specific file. You'll need to `conda install pytest-pep8` for that. `pip install autopep8` is also a useful tool.

odl/operator/default_ops.py
```diff
@@ -298,17 +298,24 @@ def adjoint(self):
         >>> op.adjoint(x)
         rn(3).element([1.0, 4.0, 9.0])
 
-        Multiply by a scalar:
+        Multiply a scalar by vector:
```
kohr-h: Something like "Multiply scalars with a fixed vector:" would be more understandable.

odl/operator/default_ops.py
```diff
@@ -298,17 +298,24 @@ def adjoint(self):
...
 
         >>> op2 = MultiplyOperator(x, domain=r3.field)
         >>> op2.adjoint(x)
         14.0
+        
+        Multiply a vector by scalar:
```
kohr-h: See above

kohr-h:
How about this?

adler-j:
I thought it'd be a good idea if @ozanoktem finished this himself :)

adler-j:
I got this fixed since @ozanoktem isn't, will merge soon.

PR #989, opened 2017-05-02
ENH: Add example of finding optimal parameters for solution method

bgris:


examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,189 @@
...
+
+        l2_norm = odl.solvers.L2NormSquared(ray_trafo.range).translated(proj_data)
+        l1_norm = lam * odl.solvers.GroupL1Norm(gradient.range)
+        f = odl.solvers.SeparableSum(l2_norm, l1_norm)
+
+        # Select solver parameters
+        op_norm = 1.1 * odl.power_method_opnorm(op)
+
+        niter = 200  # Number of iterations
+        tau = 1.0 / op_norm  # Step size for the primal variable
```
adler-j: Remove "# Step size for the primal variable" style parameters, move these into the solver call

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,189 @@
...
+    # See lbfgs_tomograhpy_tv.py for more information.
+    def reconstruction(proj_data, parameters):
+        # Extract the separate parameters
+        lam, sigma = parameters
+
+        print('lam = {}, sigma = {}'.format(lam, sigma))
+
+        # We do not allow negative paramters, so return a bogus result
+        if lam <= 0 or sigma <= 0:
+            return 1000 * space.one()
```
adler-j: np.inf * space.one()

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,189 @@
...
+    # See chambolle_pock_tomography.py for more information.
+
+    def reconstruction(proj_data, lam):
+        lam = float(lam)
+
+        print('lam = {}'.format(lam))
+
+        # We do not allow negative paramters, so return a bogus result
+        if lam <= 0:
+            return 1000 * space.one()
```
adler-j: np.inf * space.one()

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,189 @@
...
+        Initial guess for the parameters
+    """
+    def func(lam):
+        # Function to be minimized by scipy
+        return sum(fom(reconstruction(datai, lam), phantomi)
+                   for phantomi, datai in zip(phantoms, data))
+
+    # Pick resolution to fit the one used by the space
+    xtol = ftol = np.finfo(phantom.space.dtype).resolution * 10
+
```
adler-j: Here we could add an optimization in case the parameter is a scalar,

```python
    initial_param = np.asarray(initial_param)
    if initial_param.size == 1:
        bracket = [initial_param - tol, initial_param + tol]
        result = scipy.optimize.minimize_scalar(func,
                                                bracket=bracket,
                                                tol=tol,
                                                bounds=bounds,
                                                options={'disp': False})
        return result.x
    else:
        # Use a gradient free method to find the best parameters
        parameters = scipy.optimize.fmin_powell(func, initial_param,
                                                xtol=tol,
                                                ftol=tol,
                                                disp=False)
        return parameters
```

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,189 @@
...
+    initial_param : array-like
+        Initial guess for the parameters
+    """
+    def func(lam):
+        # Function to be minimized by scipy
+        return sum(fom(reconstruction(datai, lam), phantomi)
+                   for phantomi, datai in zip(phantoms, data))
+
+    # Pick resolution to fit the one used by the space
+    xtol = ftol = np.finfo(phantom.space.dtype).resolution * 10
```
adler-j: This references `phantom` which is not defined

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def optimal_parameters(reconstruction, fom, phantoms, data,
+                       initial_param=0):
+    """Find the optimal parameters for a reconstruction method.
+
+    Notes
+    -----
+    For a forward operator :math:`A : X \to Y`, a reconstruction operator
+     parametrized by :math:`\theta` is some operator :math:`R_\theta : Y \to X`
```
adler-j: seems like there is a extra space in the start of this line

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+
+        return x
+
+    initial_param = 0.1
+else:
+    raise RuntimeError('unknown reconstruction method')
+
+
+def fom(I0,I1):
+    gradient = odl.Gradient(I0.space)
```
adler-j: Move the definition of the gradient to outside of the function, also rename arguments to something like `x1`, `x2` since this is the standard naming. Also perhaps write a line on what is happening here.

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+            x, f, g, op, tau=1.0 / op_norm, sigma=1.0 / op_norm, niter= 200, gamma= 0.3)
+
+        return x
+
+    initial_param = 0.1
+else:
+    raise RuntimeError('unknown reconstruction method')
+
+
+def fom(I0,I1):
```
adler-j: PEP8-ify this. Overall I'd recommend doing a pep8 check on the whole file, but some obvious things are `a,b` should be `a, b`. You can enable pep8 checks in Spyder.

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+    gradient = odl.Gradient(I0.space)
+    return gradient(I0-I1).norm() + I0.space.dist(I0,I1)
+
+
+# Find optimal lambda
+optimal_parameters = optimal_parameters(reconstruction,  fom,
+                                        phantoms, data,
+                                        initial_param=initial_param)
+
+reconstruction(data[0], initial_param).show(reconstruction_method+' , initial parameter')
```
adler-j: Also a pep8 error, should be spaces around `+`, also quite sure this line is far beyond 80 chars.

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+        # Function to be minimized by scipy
+        return sum(fom(reconstruction(datai, lam), phantomi)
+                   for phantomi, datai in zip(phantoms, data))
+
+    # Pick resolution to fit the one used by the space
+    tol= np.finfo(phantoms[0].space.dtype).resolution * 10
+
+    initial_param = np.asarray(initial_param)
+
+    if initial_param.size == 1:
```
adler-j: Add a word on what is happening here (we use a faster optimizer for the one parameter case).

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+            * true_image
+
+        and returns a scalar Figure of Merit.
+    phantoms : sequence
+        True images.
+    data : sequence
+        The data to be reconstructed.
+    initial_param : array-like
+        Initial guess for the parameters.
+    """
```
adler-j: Add a `Returns` field

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+        R_\theta(A(x)) \approx x.
+
+    The optimal choice of :math:`\theta` is given by
+
+    .. math::
+        \theta = \argmin_\theta fom(R(A(x) + noise), x)
+
+    where :math:`fom : X \times X \to \mathbb{R}` is a Figure of Merit.
+
+
```
adler-j: Remove extra newline

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
+"""
+Example of how to use optimization in order to pick reconstruction parameters.
```
kohr-h: To the first line

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+
+In this example, we solve the tomographic inversion problem with different
+regularizers (fbp, huber and tv) and pick the "best" regularization parameter
```
kohr-h: FBP, Huber-TV and TV

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+for each method w.r.t. a set of reference data.
+
+To find the "best" parameter we use Powells method to optimize a figure of
```
kohr-h: Powell's

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
+"""
+Example of how to use optimization in order to pick reconstruction parameters.
+
+In this example, we solve the tomographic inversion problem with different
+regularizers (fbp, huber and tv) and pick the "best" regularization parameter
+for each method w.r.t. a set of reference data.
```
kohr-h: with respect to

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+     parametrized by :math:`\theta` is some operator :math:`R_\theta : Y \to X`
+     such that
+
+    .. math::
+        R_\theta(A(x)) \approx x.
+
+    The optimal choice of :math:`\theta` is given by
+
+    .. math::
+        \theta = \argmin_\theta fom(R(A(x) + noise), x)
```
kohr-h: I suspect that some of these formulas don't render properly, at least not in Spyder. That's because they contain escape sequences like `\t` that are interpreted. Those need to be double-escaped, like `\\t`. Further candidates are `\n`, `\l`, `\r`. You can check in Spyder by putting the cursor over a function definition and hitting Ctrl+I (capital "i"), and the formulas should be rendered correctly.

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+
+    where :math:`fom : X \times X \to \mathbb{R}` is a Figure of Merit.
```
kohr-h: We have "figure of merit" and "Figure of Merit", choose one

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+        Function that takes two parameters:
+
+            * reconstructed_image
+            * true_image
+
+        and returns a scalar Figure of Merit.
+    phantoms : sequence
+        True images.
+    data : sequence
+        The data to be reconstructed.
```
kohr-h: I know this is a very widely used term, to "reconstruct the data", but IMO it's wrong because it's not the data that is reconstructed, but the unknown quantity ("volume" or "image"). Should rather be "data to reconstruct from" or so.

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+    # lam the regularization strength
+    #   (multiplying the smoothed l1 in the objectiv function)
+    # See lbfgs_tomograhpy_tv.py for more information.
+    def reconstruction(proj_data, parameters):
+        # Extract the separate parameters
+        lam, sigma = parameters
+
+        print('lam = {}, sigma = {}'.format(lam, sigma))
+
+        # We do not allow negative paramters, so return a bogus result
```
kohr-h: bogus :-) okay

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,234 @@
...
+        op = odl.BroadcastOperator(ray_trafo, gradient)
+
+        g = odl.solvers.ZeroFunctional(op.domain)
+
+        l2_norm = odl.solvers.L2NormSquared(ray_trafo.range).translated(proj_data)
+        l1_norm = lam * odl.solvers.GroupL1Norm(gradient.range)
+        f = odl.solvers.SeparableSum(l2_norm, l1_norm)
+
+        # Select solver parameters
+        op_norm = 1.1 * odl.power_method_opnorm(op)
```
kohr-h: This will run up to 100 iterations by default I think, maybe you want to limit it to a lower number (~10) and multiply by a larger factor (1.5, say) to compensate for the lower accuracy of the estimate.

examples/solvers/find_optimal_parameters.py
kohr-h: Or names that tell a bit more about their meanings. Here it would make sense to name one `true_image` and the other `reco` or similar.

examples/solvers/find_optimal_parameters.py
kohr-h: My suggestion here is to bind the reconstruction to a new name `reco_0 = reconstruction(...)` and then do `reco_0.show(...)`.

adler-j:
Nice catch on adutils, I thought we had it properly removed (actually did a rebase already) but it seems I thought it was added in another commit than it actually was. I'll help you out with that @bgris. Great comments otherwise

adler-j:
Any progress on this?

kohr-h:
> Nice catch on adutils, I thought we had it properly removed (actually did a rebase already) but it seems I thought it was added in another commit than it actually was.

To get this out properly, use `git rebase -i` and `e` for "stop for amending" of the respective commit.
Alternatively, this can also be done by [git filter-branch](https://git-scm.com/docs/git-filter-branch).

examples/solvers/find_optimal_parameters.py
bgris: Sorry, I do not really understand what you mean here.

examples/solvers/find_optimal_parameters.py
bgris: Sorry, I do not really understand what you mean here.

adler-j: He means that instead of

```
"""
My comment starts here
```

The "correct" format is

```
"""My comment starts here
```

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,241 @@
...
+    phantoms : sequence
+        True images.
+    data : sequence
+        The data to reconstruct from.
+    initial_param : array-like
+        Initial guess for the parameters.
+
+    Returns
+    -------
+    parameters : 'array'
```
adler-j: parameters : `numpy.ndarray`

examples/solvers/find_optimal_parameters.py
adler-j: These formulas are intended to be read typeset, e.g. if you click `Ctrl+I` as @kohr-h mentioned you get:

image

Now, this is because of something called "escaped characters", or in particular, the lack of them. Escaped characters exist because some combination of letters already carry a meaning. For example, `\t` means a tab, `\n` means a newline, etc. So if you simply write `\theta` the typesetter will read this as "[tab]eta" and it will look weird. To tell the typesetter that you truly mean "`\t`", literally, you need to write `\\t`. Here, `\\` is an escaped character indicating that you really mean the character `\\` and not any special magic to happen. This concept exists in basically all typesetting standards, including HTML, markdown, etc. Even `LaTeX` has it (try writing `%` in latex, you need to do `\%` to escape it).

If you fix this, everything typesets nicely:

image

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,241 @@
...
+
+        return x
+
+    initial_param = 0.1
+else:
+    raise RuntimeError('unknown reconstruction method')
+
+
+def fom(reco, true_image):
+    gradient = odl.Gradient(reco.space)
```
adler-j: This choice of `fom` deserves some further explanation I think, at least a sentence:

```
"""Sobolev type FoM enforcing both gradient and absolute similarity."""
```

examples/solvers/find_optimal_parameters.py
```diff
@@ -0,0 +1,241 @@
...
+    return gradient(reco-true_image).norm() + reco.space.dist(reco, true_image)
+
+
+# Find optimal lambda
+optimal_parameters = optimal_parameters(reconstruction,  fom,
+                                        phantoms, data,
+                                        initial_param=initial_param)
+
+reco_0 = reconstruction(data[0], initial_param)
+reco_0.show(reconstruction_method + ' , initial parameter')
```
adler-j: Extra space before `,`

adler-j:
Also need to fix the extra large thing, but I suggest we squash the commits before a merge which will solve this, OK @kohr-h?

kohr-h:
> Also need to fix the extra large thing, but I suggest we squash the commits before a merge which will solve this, OK @kohr-h?

Sure.

examples/solvers/find_optimal_parameters.py
bgris: Ok I did not know thqt, thqnk you !!

PR #990, opened 2017-05-09
ENH: make MRC reader more robust wrt broken axis mapping

kohr-h:
Simple fix for a case when MRC files don't properly define the axis mapping fields.

### DIFF
--- a/odl/tomo/data/mrc.py
+++ b/odl/tomo/data/mrc.py
@@ -18,6 +18,7 @@ from collections import OrderedDict
 from itertools import permutations
 import numpy as np
 import struct
+import warnings
 
 from odl.tomo.data.uncompr_bin import (
     FileReaderRawBinaryWithHeader, FileWriterRawBinaryWithHeader,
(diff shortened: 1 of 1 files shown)


odl/tomo/data/mrc.py
```diff
@@ -346,7 +346,14 @@ def data_axis_order(self):
         except KeyError:
             return (0, 1, 2)
         else:
-            return tuple(int(m) - 1 for m in [mapc, mapr, maps])
+            axis_order = tuple(int(m) - 1 for m in [mapc, mapr, maps])
+            if (len(set(axis_order)) != 3 or
+                    any(m not in (0, 1, 2) for m in axis_order)):
+                # Ignore invalid entries in the header, e.g. 0, 0, 0.
```
adler-j: I am strongly against silently improving data. Add something like

```
warnings.warn('Invalid axis_order, using (0, 1, 2)')
```

odl/tomo/data/mrc.py
```diff
@@ -346,7 +346,14 @@ def data_axis_order(self):
         except KeyError:
             return (0, 1, 2)
         else:
-            return tuple(int(m) - 1 for m in [mapc, mapr, maps])
+            axis_order = tuple(int(m) - 1 for m in [mapc, mapr, maps])
+            if (len(set(axis_order)) != 3 or
```
adler-j: `if sorted(axis_order) != (0, 1, 2):` 

would perhaps be easier to read?

odl/tomo/data/mrc.py
```diff
@@ -346,7 +346,14 @@ def data_axis_order(self):
...
+                    any(m not in (0, 1, 2) for m in axis_order)):
+                # Ignore invalid entries in the header, e.g. 0, 0, 0.
+                # Some MRC files out there are like that.
+                return (0, 1, 2)
```
adler-j: I would prefer going from here as

```
    axis_order = (0, 1, 2)

return axis_order
```

more readable with single return.

odl/tomo/data/mrc.py
kohr-h: Fair enough

kohr-h:
Fixed, merging after CI.

PR #992, opened 2017-05-10
Add datasets module

adler-j:
This is a proposal for how we could start including actual datasets into ODL.

The architecture works similarily to the datasets in `scikit-learn`, and it basically means data is not stored in the ODL repository, but somewhere else. On first access, the data is downloaded to a local folder in the home directory where it is cached.

So far I've only added two datasets (the walnut and lotus root from finland), but it should be relatively simple to add more.

Is this something we want?

Example results:

walnut_fbp

lotus_root_fbp

### DIFF
--- a/odl/__init__.py
+++ b/odl/__init__.py
@@ -16,7 +16,8 @@ from __future__ import absolute_import
 
 __version__ = '0.6.1.dev0'
 __all__ = ('diagnostics', 'discr', 'operator', 'set', 'space', 'solvers',
-           'tomo', 'trafos', 'util', 'phantom', 'deform', 'ufunc_ops')
+           'tomo', 'trafos', 'util', 'phantom', 'deform', 'ufunc_ops',
+           'datasets')
 
 # Propagate names defined in __all__ of all submodules into the top-level
 # module
@@ -41,6 +42,8 @@ from . import util
 from . import phantom
 from . import deform
 from . import ufunc_ops
+from . import datasets
+
 
 from .util import test
 __all__ += ('test',)
--- a/odl/datasets/__init__.py
+++ b/odl/datasets/__init__.py
@@ -6,11 +6,11 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
-"""Access to real or simulated datasets for use with ODL."""
+"""Access to real or simulated datasets."""
 
 from __future__ import absolute_import
 
 
-__all__ = ('ray_transform',)
+__all__ = ('tomo',)
 
-from . import ray_transform
+from . import tomo
--- a/odl/datasets/__init__.py
+++ b/odl/datasets/__init__.py
@@ -1,0 +1,16 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Access to real or simulated datasets for use with ODL."""
+
+from __future__ import absolute_import
+
+
+__all__ = ('ray_transform',)
+
+from . import ray_transform
--- a/odl/datasets/ray_transform.py
+++ b/odl/datasets/ray_transform.py
@@ -14,6 +14,10 @@ from odl.discr import uniform_partition
 from odl.tomo import FanFlatGeometry
 
 
+__all__ = ('walnut_data', 'walnut_geometry',
+           'lotus_root_data', 'lotus_root_geometry')
+
+
 DATA_SUBSET = 'ray_transform'
 
 def walnut_data():
@@ -140,22 +144,27 @@ if __name__ == '__main__':
     
     # Walnut example
     space = odl.uniform_discr([-20, -20], [20, 20], [2296, 2296])
-    geometry = walnut_geometry()
+    geometry = odl.datasets.ray_transform.walnut_geometry()
     
     ray_transform = odl.tomo.RayTransform(space, geometry)
     fbp_op = odl.tomo.fbp_op(ray_transform, filter_type='Hann')
     
-    data = walnut_data()
+    data = odl.datasets.ray_transform.walnut_data()
     fbp_op(data).show('walnut fbp', clim=[0, 0.05])
     
 
     # Lotus root example
     space = odl.uniform_discr([-50, -50], [50, 50], [2240, 2240])
-    geometry = lotus_root_geometry()
+    geometry = odl.datasets.ray_transform.lotus_root_geometry()
     
     ray_transform = odl.tomo.RayTransform(space, geometry)
     fbp_op = odl.tomo.fbp_op(ray_transform, filter_type='Hann')
     
-    data = lotus_root_data()
+    data = odl.datasets.ray_transform.lotus_root_data()
     fbp_op(data).show('lotus root fbp', clim=[0, 0.1])
+    
+    # Run doctests
+    # pylint: disable=wrong-import-position
+    from odl.util.testutils import run_doctests
+    run_doctests()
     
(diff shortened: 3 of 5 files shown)


odl/datasets/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Access to real or simulated datasets for use with ODL."""
+
+from __future__ import absolute_import
+
+
+__all__ = ('ray_transform',)
+
+from . import ray_transform
```
kohr-h: Add `ray_transform.__all__` to `__all__`

odl/datasets/ray_transform.py
```diff
@@ -0,0 +1,170 @@
...
+    
+    # Walnut example
+    space = odl.uniform_discr([-20, -20], [20, 20], [2296, 2296])
+    geometry = odl.datasets.ray_transform.walnut_geometry()
+    
+    ray_transform = odl.tomo.RayTransform(space, geometry)
+    fbp_op = odl.tomo.fbp_op(ray_transform, filter_type='Hann')
+    
+    data = odl.datasets.ray_transform.walnut_data()
+    fbp_op(data).show('walnut fbp', clim=[0, 0.05])
```
kohr-h: "Walnut FBP reconstruction"

odl/datasets/ray_transform.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Lotus root example
+    space = odl.uniform_discr([-50, -50], [50, 50], [2240, 2240])
+    geometry = odl.datasets.ray_transform.lotus_root_geometry()
+    
+    ray_transform = odl.tomo.RayTransform(space, geometry)
+    fbp_op = odl.tomo.fbp_op(ray_transform, filter_type='Hann')
+    
+    data = odl.datasets.ray_transform.lotus_root_data()
+    fbp_op(data).show('lotus root fbp', clim=[0, 0.1])
```
kohr-h: "Lotus root FBP reconstruction"

odl/__init__.py
```diff
@@ -41,6 +42,8 @@
 from . import phantom
 from . import deform
 from . import ufunc_ops
+from . import datasets
```
kohr-h: How about having only one subpackage for data, containing both real data from wherever, and phantom data?

kohr-h:
I was too quick. Please check also the now hidden comments.

odl/datasets/ray_transform.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Datasets for the ray transform."""
```
adler-j: Because tomography also means MRI, so I was planning to keep them separate. Is that a bad idea?

odl/datasets/util.py
```diff
@@ -0,0 +1,84 @@
...
+from shutil import copyfileobj
+from scipy import io
+
+
+__all__ = ('get_data_dir', 'get_data')
+
+
+def get_data_dir():
+    """Get the data directory."""
+    data_home = expanduser(join('~', 'odl_data'))
```
adler-j: I agree with all you said.  Regarding windows it works fine

odl/datasets/__init__.py
adler-j: I'm not sure here. Are we not afraid of a growth in phantoms here? I suspect we could long term end up with a very large number of data sets.

kohr-h: But it's still confined to the `datasets` package, so it won't pollute the top-level namespace. Anyway, I don't have a strong opinion here. Given that it's easier to later add it to `__all__` than to remove it again, we can keep it the way it is here.

odl/__init__.py
adler-j: Wouldn't we run into the risk of said package becomming overpopulated? With that said I agree that there is no real sharp limit, e.g. where would we put complicated humanoid phantoms?

I'll need to think more about it.

kohr-h: Actually, when I think about it, phantoms are a totally different thing (e.g. space-wise) than data, so it's only logical to keep them separate. So even humanoid phantoms are phantoms and go into `phantoms`, whereas `datasets` can also be simulated data.

So I'd say it's fine like this.

odl/datasets/ray_transform.py
adler-j: How about I name it `tomo`, like `odl.tomo`?

odl/datasets/ray_transform.py
```diff
@@ -0,0 +1,170 @@
...
+    See Also
+    --------
+    walnut_geometry
+    
+    References
+    ----------
+    .. Tomographic X-ray data of a walnut: https://arxiv.org/abs/1502.04064
+    """
+    
+    # TODO: Store data in some ODL controlled url
```
adler-j: Well it could also be temporary etc. I'd prefer if we owned the data. Anyway I'll leave it as is for now.

odl/datasets/util.py
adler-j: Went with:

```

    data_home = os.environ.get('ODL_DATA_DIR',
                               expanduser(join('~', 'odl', 'datasets')))
```

adler-j: Second thought, lets go with a single base directory for ODL 

```
    base_odl_dir = os.environ.get('ODL_DIR', expanduser(join('~', 'odl')))
    data_home = join(base_odl_dir, 'datasets')
```

odl/datasets/util.py
```diff
@@ -0,0 +1,84 @@
...
+
+    filename = join(data_dir, filename)
+
+    # if the file does not exist, download it
+    if not exists(filename):
+        print('data {}/{} missing, downloading from {}'
+              ''.format(subset, filename, url))
+        
+        # open the url of the data
+        with urlopen(url) as data_url:
```
adler-j: As an error written where?

odl/datasets/util.py
kohr-h: I'd find `ODL_DIR` confusing since it could also mean that ODL is installed there. Not sure what would be a better name, suggestions:
- `ODL_CONFIG_DIR`
- `ODL_DATA_DIR`
- `ODL_PREFIX`
- `ODL_HOME`
- `ODL_LOCAL_DIR`

Can you also add a dot before `odl`? It's not unlikely that people have the ODL repository in `~/odl` so we shouldn't pollute that by default.

odl/datasets/util.py
kohr-h: I meant [this comment](https://github.com/odlgroup/odl/pull/992#discussion_r115684409) but couldn't reference it during review.

adler-j:
This needs a re-review

odl/datasets/tomo.py
```diff
@@ -0,0 +1,169 @@
...
+
+    See Also
+    --------
+    walnut_geometry
+
+    References
+    ----------
+    .. _Tomographic X-ray data of a walnut: https://arxiv.org/abs/1502.04064
+    """
+
```
kohr-h: remove blank row

odl/datasets/util.py
```diff
@@ -0,0 +1,82 @@
...
+from shutil import copyfileobj
+from scipy import io
+
+
+__all__ = ('get_data_dir', 'get_data')
+
+
+def get_data_dir():
+    """Get the data directory."""
+    base_odl_dir = os.environ.get('ODL_DIR', expanduser(join('~', 'odl')))
```
kohr-h: See [my previous comment](https://github.com/odlgroup/odl/pull/992#discussion_r115858930)

odl/datasets/util.py
adler-j: ODL_HOME i guess? And I'll add the dot

adler-j:
Merging after CI, but this obviously has some long term stuff:

* Add all the walnut and lotus root datasets (e.g. subsampled)
* Add proper MRI datasets
* Add Elekta CBCT datasets

I'll make an issue.

PR #995, opened 2017-05-10
Issue 994  empty

kohr-h:

PR #996, opened 2017-05-10
Issue 671  mixed uni nonuni

kohr-h:
~~Based on #995, needs rebase after that one is merged.~~

Based on master now, ready for review.

### DIFF
--- a/odl/discr/discr_ops.py
+++ b/odl/discr/discr_ops.py
@@ -260,7 +260,7 @@ class ResizingOperatorBase(Operator):
          [5.0, 6.0, 7.0, 8.0],
          [1.0, 2.0, 3.0, 4.0]]
         """
-        from builtins import range as range_seq
+        from builtins import range as builtin_range
 
         if not isinstance(domain, DiscreteLp):
             raise TypeError('`domain` must be a `DiscreteLp` instance, '
@@ -285,9 +285,9 @@ class ResizingOperatorBase(Operator):
                 raise ValueError('`offset` can only be combined with '
                                  '`ran_shp`')
 
-            for i in range_seq(domain.ndim):
-                if ((range.is_uniform_byaxis[i] and
-                    domain.is_uniform_byaxis[i]) and
+            for i in builtin_range(domain.ndim):
+                if (range.is_uniform_byaxis[i] and
+                    domain.is_uniform_byaxis[i] and
                         not np.isclose(range.cell_sides[i],
                                        domain.cell_sides[i])):
                     raise ValueError(
@@ -485,9 +485,10 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
 
     affected = np.not_equal(newshp, discr.shape)
     ndim = discr.ndim
-    if any(affected[i] for i in range(ndim)
-           if not discr.is_uniform_byaxis[i]):
-        raise ValueError('cannot resize non-uniformly discretized axes')
+    for i in range(ndim):
+        if affected[i] and not discr.is_uniform_byaxis[i]:
+            raise ValueError('cannot resize in non-uniformly discretized '
+                             'axis {}'.format(i))
 
     grid_min, grid_max = discr.grid.min(), discr.grid.max()
     cell_size = discr.cell_sides
@@ -530,27 +531,22 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
     dspace = fn(np.prod(newshp), dtype=dtype, impl=impl, exponent=exponent,
                 weighting=weighting)
 
-    # This part is a bit messy since we need to stack together the (old)
-    # nonuniform axes and the (new) uniform axes in the right order
-    nonuni_axes = np.where(np.logical_not(discr.is_uniform_byaxis))[0]
-    if nonuni_axes.size == 0:
-        nonuni_part = []
-    else:
-        nonuni_part = uniform_partition([], [], ())
-        for axis in nonuni_axes:
-            nonuni_part = nonuni_part.append(discr.partition.byaxis[axis])
+    # Stack together the (unchanged) nonuniform axes and the (new) uniform
+    # axes in the right order
+    nonuni_axes = [i for i in range(discr.ndim)
+                   if not discr.is_uniform_byaxis[i]]
+    nonuni_part = discr.partition.byaxis[nonuni_axes]
 
     part = uniform_partition([], [], ())
-    i_uni = i_nonuni = 0
+    count_nonuni = 0
     for i in range(ndim):
         if discr.is_uniform_byaxis[i]:
             part = part.append(
                 uniform_partition(new_minpt[i], new_maxpt[i], newshp[i],
                                   nodes_on_bdry=nodes_on_bdry[i]))
-            i_uni += 1
         else:
-            part = part.append(nonuni_part.byaxis[i_nonuni])
-            i_nonuni += 1
+            part = part.append(nonuni_part.byaxis[count_nonuni])
+            count_nonuni += 1
 
     return DiscreteLp(fspace, part, dspace, exponent=exponent, interp=interp,
                       order=order)
--- a/odl/discr/discr_ops.py
+++ b/odl/discr/discr_ops.py
@@ -16,8 +16,10 @@ from builtins import super
 
 import numpy as np
 
-from odl.discr import DiscreteLp, uniform_discr
+from odl.discr import DiscreteLp, uniform_partition, nonuniform_partition
 from odl.operator import Operator
+from odl.set import IntervalProd
+from odl.space import FunctionSpace, fn
 from odl.util import (
     normalized_scalar_param_list, safe_int_conv, resize_array)
 from odl.util.numerics import _SUPPORTED_RESIZE_PAD_MODES
@@ -258,12 +260,11 @@ class ResizingOperatorBase(Operator):
          [5.0, 6.0, 7.0, 8.0],
          [1.0, 2.0, 3.0, 4.0]]
         """
+        from builtins import range as range_seq
+
         if not isinstance(domain, DiscreteLp):
             raise TypeError('`domain` must be a `DiscreteLp` instance, '
                             'got {!r}'.format(domain))
-
-        if not domain.is_uniform:
-            raise ValueError('`domain` is not uniformly discretized')
 
         offset = kwargs.pop('offset', None)
         discr_kwargs = kwargs.pop('discr_kwargs', {})
@@ -284,11 +285,16 @@ class ResizingOperatorBase(Operator):
                 raise ValueError('`offset` can only be combined with '
                                  '`ran_shp`')
 
-            if not np.allclose(range.cell_sides, domain.cell_sides):
-                raise ValueError(
-                    'cell sides of domain and range differ significantly '
-                    '(difference {})'
-                    ''.format(range.cell_sides - domain.cell_sides))
+            for i in range_seq(domain.ndim):
+                if ((range.is_uniform_byaxis[i] and
+                    domain.is_uniform_byaxis[i]) and
+                        not np.isclose(range.cell_sides[i],
+                                       domain.cell_sides[i])):
+                    raise ValueError(
+                        'in axis {}: cell sides of domain and range differ '
+                        'significantly: (difference {})'
+                        ''.format(i,
+                                  range.cell_sides[i] - domain.cell_sides[i]))
 
             self.__offset = _offset_from_spaces(domain, range)
 
@@ -325,6 +331,12 @@ class ResizingOperatorBase(Operator):
     def pad_const(self):
         """Constant used by this operator in case of constant padding."""
         return self.__pad_const
+
+    @property
+    def axes(self):
+        """Dimensions in which an actual resizing is performed."""
+        return tuple(i for i in range(self.domain.ndim)
+                     if self.domain.shape[i] != self.range.shape[i])
 
 
 class ResizingOperator(ResizingOperatorBase):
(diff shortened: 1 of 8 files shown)


odl/discr/discr_ops.py
```diff
@@ -284,11 +285,16 @@ def __init__(self, domain, range=None, ran_shp=None, **kwargs):
...
                                  '`ran_shp`')
 
-            if not np.allclose(range.cell_sides, domain.cell_sides):
-                raise ValueError(
-                    'cell sides of domain and range differ significantly '
-                    '(difference {})'
-                    ''.format(range.cell_sides - domain.cell_sides))
+            for i in range_seq(domain.ndim):
+                if ((range.is_uniform_byaxis[i] and
+                    domain.is_uniform_byaxis[i]) and
```
kohr-h: extra parens

odl/discr/discr_ops.py
```diff
@@ -499,9 +525,35 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
...
-                         exponent=exponent, interp=interp, order=order,
-                         weighting=weighting, nodes_on_bdry=nodes_on_bdry)
+    fspace = FunctionSpace(IntervalProd(new_minpt, new_maxpt),
+                           out_dtype=dtype)
+    dspace = fn(np.prod(newshp), dtype=dtype, impl=impl, exponent=exponent,
+                weighting=weighting)
+
+    # This part is a bit messy since we need to stack together the (old)
+    # nonuniform axes and the (new) uniform axes in the right order
+    nonuni_axes = np.where(np.logical_not(discr.is_uniform_byaxis))[0]
```
kohr-h: Maybe also here just
```python
[i for i in range(discr.ndim) if not discr.is_uniform_byaxis[i]]
```
Easier on the eye.

odl/discr/discr_ops.py
```diff
@@ -499,9 +525,35 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
...
+    if nonuni_axes.size == 0:
+        nonuni_part = []
+    else:
+        nonuni_part = uniform_partition([], [], ())
+        for axis in nonuni_axes:
+            nonuni_part = nonuni_part.append(discr.partition.byaxis[axis])
```
kohr-h: This should actually work with list indexing.

odl/test/discr/discr_ops_test.py
```diff
@@ -242,5 +235,27 @@ def test_resizing_op_adjoint(padding, fn_impl):
         assert almost_equal(inner1, inner2, places=dtype_places(dtype))
 
 
+def test_resizing_op_mixed_uni_nonuni():
+    """Check if resizing along uniform axes in mixed discretizations works."""
+    nonuni_part = odl.nonuniform_partition([0, 1, 4])
+    uni_part = odl.uniform_partition(-1, 1, 4)
+    part = uni_part.append(nonuni_part).append(uni_part).append(nonuni_part)
```
adler-j: Perhaps we should add an `extend` method that basically does

```
def extend(self, lst):
    result = self
    for el in lst:
        result = result.append(el)
    return result
```

Then you could do

```
part = uni_part.extend([nonuni_part, uni_part, nonuni_part])
```

which is much cleaner

odl/discr/grid.py
```diff
@@ -389,11 +389,10 @@ def stride(self):
...
         """
-        if not self.is_uniform:
-            raise NotImplementedError('`stride` not defined for non-uniform '
-                                      'grid')
         strd = np.zeros(self.ndim)
-        cond = np.array(self.nondegen_byaxis, dtype=bool)
+        uniform = np.array(self.is_uniform_byaxis, dtype=bool)
+        strd[~uniform] = float('nan')
+        cond = np.logical_and(uniform, self.nondegen_byaxis)
         strd[cond] = self.extent[cond] / (np.array(self.shape)[cond] - 1)
```
kohr-h: This is also just a complicated way of writing
```python
strd = [self.extent[i] / (self.shape[i] - 1)
        if self.is_uniform_byaxis[i] else float('nan')
        for i in range(self.ndim)]
```

odl/test/discr/discr_ops_test.py
```diff
@@ -242,5 +235,27 @@ def test_resizing_op_adjoint(padding, fn_impl):
...
+    fspace = odl.FunctionSpace(odl.IntervalProd(part.min_pt, part.max_pt))
+    dspace = odl.rn(part.size)
+    space = odl.DiscreteLp(fspace, part, dspace)
+
+    # Keep non-uniform axes fixed
```
adler-j: Do we get a nice error in the other case?

odl/test/discr/discr_ops_test.py
```diff
@@ -242,5 +235,27 @@ def test_resizing_op_adjoint(padding, fn_impl):
...
+    res_op = odl.ResizingOperator(space, ran_shp=(6, 3, 6, 3))
+
+    assert res_op.axes == (0, 2)
+    assert res_op.offset == (1, 0, 1, 0)
+
```
adler-j: No evaluation test at all? Only testing the adjoint is scary since this may as well be the zero operator and we'll never know.

odl/discr/discr_ops.py
kohr-h: It's actually a superfluous pair of parentheses

odl/discr/discr_ops.py
```diff
@@ -499,9 +525,35 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
...
+
+    part = uniform_partition([], [], ())
```
kohr-h: No, that's simple, just two lines in `uniform_partition`. But is it obvious what "no argument" means? This "explicitly empty" construction is perfectly clear, the other one somewhat less so IMO.

odl/discr/grid.py
```diff
@@ -392,13 +388,11 @@ def stride(self):
...
         """
-        if not self.is_uniform:
-            raise NotImplementedError('`stride` not defined for non-uniform '
-                                      'grid')
         strd = np.zeros(self.ndim)
-        strd[self.nondegen_byaxis] = (
-            self.extent[self.nondegen_byaxis] /
-            (np.array(self.shape)[self.nondegen_byaxis] - 1))
+        uniform = np.array(self.is_uniform_byaxis, dtype=bool)
+        strd[~uniform] = float('nan')
```
kohr-h: Well, for the `strides` property it's a way of signalizing "not defined in this axis". Later on the NaN values are not processed since the non-uniform axes are not considered. At least if there's no bug in the code. Hopefully the NaNs will mess up later code that tries to wrongly process the non-uniform axes.

odl/test/discr/discr_ops_test.py
kohr-h: Good suggestion!

odl/test/discr/discr_ops_test.py
kohr-h: We get
```python
  File "[...]/odl/discr/discr_ops.py", line 280, in __init__
    range = _resize_discr(domain, ran_shp, offset, discr_kwargs)
  File "[...]/odl/discr/discr_ops.py", line 490, in _resize_discr
    raise ValueError('cannot resize non-uniformly discretized axes')
ValueError: cannot resize non-uniformly discretized axes
```
I guess I should print the axis, too.

odl/test/discr/discr_ops_test.py
kohr-h: Okay okay :-)
I ran the code by hand before, and the result was not zero. But you're right, a bit more confidence on correctness cannot hurt.

odl/discr/discr_ops.py
```diff
@@ -499,9 +525,35 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
...
+    i_uni = i_nonuni = 0
```
kohr-h: Well, I do use them for indexing.. And `num_bla` is for me the *total* number of `bla`. Perhaps `count_bla` instead?

kohr-h: And `i_uni` isn't used, so that one has to go.

odl/discr/grid.py
```diff
@@ -392,13 +388,11 @@ def stride(self):
...
+        cond = np.logical_and(uniform, self.nondegen_byaxis)
+        strd[cond] = self.extent[cond] / (np.array(self.shape)[cond] - 1)
```
kohr-h: `extent` is a float array, but I guess it won't hurt to be cautious.

kohr-h:
Comments fixed, needs a green flag

odl/discr/discr_ops.py
```diff
@@ -499,9 +526,30 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
...
-                         exponent=exponent, interp=interp, order=order,
-                         weighting=weighting, nodes_on_bdry=nodes_on_bdry)
+    fspace = FunctionSpace(IntervalProd(new_minpt, new_maxpt),
+                           out_dtype=dtype)
+    dspace = fn(np.prod(newshp), dtype=dtype, impl=impl, exponent=exponent,
+                weighting=weighting)
+
+    # Stack together the (unchanged) nonuniform axes and the (new) uniform
+    # axes in the right order
+    nonuni_axes = [i for i in range(discr.ndim)
```
adler-j: I would prefer if this was moved down, to the `if` statemnt further down, that way we don't have conditionals in two places and the code becomes less complex.

odl/discr/grid.py
```diff
@@ -389,13 +389,15 @@ def stride(self):
...
-            raise NotImplementedError('`stride` not defined for non-uniform '
-                                      'grid')
-        strd = np.zeros(self.ndim)
-        cond = np.array(self.nondegen_byaxis, dtype=bool)
-        strd[cond] = self.extent[cond] / (np.array(self.shape)[cond] - 1)
-        return strd
+        strd = []
+        for i in range(self.ndim):
+            if not self.is_uniform_byaxis[i]:
+                strd.append(float('nan'))
```
adler-j: This has to be documented

odl/discr/grid.py
```diff
@@ -681,9 +684,46 @@ def append(self, other):
...
+        extend
         """
         return self.insert(self.ndim, other)
 
+    def extend(self, seq):
+        """Return a copy of ``self``, extended by all members of ``seq``.
+
+        Parameters
+        ----------
+        seq : sequence of `RectGrid`
```
adler-j: Second thought we should be really careful about this. Overall, we should do "what people expect", and for lists extend has the syntax `new_list = list.extend(something_that_is_like_a_list)`, implemented somewhat like this fella.

Anyway, wouldn't people expect this to work in a similar manner, e.g. `new_grid = grid.extend(something_that_is_like_a_grid)`, which would imply that calling this with a grid should work nicely. But I THINK that is not the case now, right?

We need to give this some more thought.

odl/discr/discr_ops.py
```diff
@@ -258,13 +260,12 @@ def __init__(self, domain, range=None, ran_shp=None, **kwargs):
          [5.0, 6.0, 7.0, 8.0],
          [1.0, 2.0, 3.0, 4.0]]
         """
+        from builtins import range as builtin_range
```
kohr-h: Hm, I rather see this as an "emergency import" and would prefer not polluting the module namespace with this. Where possible, only `range` should be available. 
That's not a strong preference, though, for me it's just a second case beside an expensive import where lazy imports are okay.

odl/discr/discr_ops.py
kohr-h: Oh god and it's so trivial that that removes the error-prone count thing. Too obvious :-)

odl/discr/grid.py
kohr-h: Will do

odl/discr/grid.py
kohr-h: ~~Well, to some degree it's already too late since we're already down that route with `insert` and `append`. For sure `RectGrid` and `RectPartition` are no container types, so there's a bit of tension with the container association.
The only way to solve this would be to allow these types to be constructed with one or more of their own type, e.g.,~~
```python
>>> grid = RectGrid(grid1, grid2, ..., gridN)
```
...
... 
Oh, this is really bad:
```python
>>> g = RectGrid([0, 1, 2], [0, 1, 4])
>>> g.extend(g)
RectGrid(
    [0.0, 1.0, 2.0],
    [0.0, 1.0, 4.0],
    [0.0],
    [0.0, 1.0, 4.0],
    [1.0],
    [0.0, 1.0, 4.0],
    [2.0],
    [0.0, 1.0, 4.0]
)
```
Clear oops, so I think `extend has to go away again.

adler-j: There really is no proper way out, right? An option would be to simply let append take several arguments.

```
def append(self, *stuff):
    If len(stuff) > 1:
        return self.append(stuff[0]).append(stuff [1:])
    else:
        #current code
```

Would be an implementation

kohr-h: Decent suggestion, also because it makes one out of two methods look less similar to the corresponding list method. It would then also make sense to do the same for `insert`, still with a single index, so one can add a bunch of stuff at a given position.

kohr-h:
Comments fixed, I guess a quick review on the `insert` and `append` stuff is required.

odl/discr/grid.py
```diff
@@ -638,33 +640,54 @@ def insert(self, index, other):
...
+
+        index, index_in = safe_int_conv(index), index
+        if not -self.ndim <= index <= self.ndim:
+            raise IndexError('index {0} outside the valid range -{1} ... {1}'
+                             ''.format(index_in, self.ndim))
+        if index < 0:
+            index += self.ndim
+
+        if len(grids) > 1:
+            return self.insert(index, grids[0]).insert(
```
adler-j: This is O(n^2) complexity with the input checks, I'm not sure if that is a problem or not?

kohr-h: Yes I realized that but put it off, thinking "nah, n is small". But we never know what people may come up with, so I'll fix it anyway.

PR #998, opened 2017-05-12
BUG: Fix bug with negative indices in ellipsoid_phantom

adler-j:
Negative indices seem to be interpreted as "from the end", giving bugs for phantoms that go outside of the volume. This PR fixes that.


adler-j:
No, they simply disappeared since the feasible region became essentially empty. E.g. for x in [0, 1], -0.5 < x < 0.2 was interpreted as 0.5 < x < 0.2, which has no solutions

kohr-h:
> No, they simply disappeared since the feasible region became essentially empty. E.g. for x in [0, 1], -0.5 < x < 0.2 was interpreted as 0.5 < x < 0.2, which has no solutions

Oh, that's funny, the condition got wrapped around :-)

PR #1003, opened 2017-05-22
Issue 999  random state

adler-j:


odl/util/utility.py
```diff
@@ -805,6 +805,43 @@ def pkg_supports(feature, pkg_version, pkg_feat_dict):
     return False
 
 
+class NumpyRandomSeed(object):
+    """Context manager for numpy randoms seeds."""
```
kohr-h: Numpy random seeds

odl/util/utility.py
```diff
@@ -805,6 +805,43 @@ def pkg_supports(feature, pkg_version, pkg_feat_dict):
...
+
+    def __init__(self, seed):
+        """
```
kohr-h: Missing

odl/util/utility.py
```diff
@@ -805,6 +805,43 @@ def pkg_supports(feature, pkg_version, pkg_feat_dict):
...
+
+        Parameters
+        ----------
+        seed : int or None
+            Seed value for the random number generator.
+            ``None`` is interpreted as keeping the current seed.
+
+        Examples
+        --------
+        Always returns same value with same seed:
```
kohr-h: Make text a bit nicer, e.g., "Use this to make drawing pseudo-random numbers repeatable:"

odl/phantom/noise.py
```diff
@@ -125,6 +140,9 @@ def salt_pepper_noise(vector, fraction=0.05, salt_vs_pepper=0.5,
         The "salt" value in the noise.
         Default: maximuim value of ``vector``. For product spaces the maximum
         value per subspace is taken.
+    seed : int, optional
+        Random seed to use for generating the noise. If ``None``, uses current
```
kohr-h: For \`\`None\`\`, use the current seed

odl/util/utility.py
adler-j: What do we write here? "Initialize an instance"?

kohr-h: Why not? You can as well create an instance of the context manager and then write `with instance:`.

PR #1004, opened 2017-05-22
ENH: Add as_scipy_functional

adler-j:


odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
                                               dtype=dtype)
 
 
+def as_scipy_functional(func, return_gradient=False):
+    """Wrap ``op`` as a function operating on linear arrays.
+
+    This is intended to be used with the scipy solvers.
```
kohr-h: This could link to a scipy doc page.

odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
...
+
+    Parameters
+    ----------
+    func : `Functional`
+        A functional that should be wrapped
```
kohr-h: dot

odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
...
+    return_gradient : bool, optional
+        If the gradient of the functional should also be returned
```
kohr-h: dot

We abandoned this style, didn't we? In favor of "\`\`True\`\` if \<condition fulfilled\>, \`\`False\`\` otherwise."

odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
...
+
+    Returns
+    -------
+    function : ``callable``
+        The wrapped functional.
+    gradient : ``callable``, optional
```
kohr-h: "optional" only makes sense with arguments, not return values. The doc already explains what happens.

odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
...
+
+    Examples
+    --------
+    Wrap functional and solve simple problem
+    (here toy problem ``min_x ||x||^2``):
+
+    >>> func = odl.solvers.L2NormSquared(odl.rn(3))
+    >>> scipy_func = as_scipy_functional(func)
+    >>> from scipy.optimize import minimize
+    >>> result = minimize(scipy_func, [0, 1, 0])
```
kohr-h: What is the second argument?

odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
...
+    >>> np.allclose(result.x, [0, 0, 0])
+    True
+
+    Can also provide gradient:
```
kohr-h: The gradient can also be provided:

odl/operator/oputils.py
```diff
@@ -319,6 +319,76 @@ def rmatvec(v):
...
+    >>> func = odl.solvers.L2NormSquared(odl.rn(3))
+    >>> scipy_func, scipy_grad = as_scipy_functional(func, True)
+    >>> from scipy.optimize import minimize
+    >>> result = minimize(scipy_func, [0, 1, 0], jac=scipy_grad)
+    >>> np.allclose(result.x, [0, 0, 0])
+    True
+
+    Notes
+    -----
+    If the data representation of ``op``'s domain and range is of type
```
kohr-h: update (only domain is relevant)

odl/operator/oputils.py
adler-j: Well it is optional in a sense, no? I.e. does not always happen

kohr-h: Yeah, but the user does not have this option, at least not in principle. Here it happens to be that case. At least potentially it's outside the control of the user, so it's rather a "maybe" or "may not be returned". 
But let's not be too orthodox. If you happen to come across a better formulation, take that, otherwise just go ahead.

adler-j: Well the user has an option (specifying a parameter), but i see what you mean.

Since I can't come up with anything smarter, I'll go ahead for now.

PR #1006, opened 2017-05-22
ENH: Make Fourier transforms select fastest available backend, closes #988

### LINKED ISSUE #988: Change FourierTransform to use fastest available backend
Similar to the change we did for `RayTransform` where the "fastest available" backend is picked, we should do the same in the fourier transform. The order should be pyfftw > numpy.

adler-j:


odl/trafos/fourier.py
```diff
@@ -77,9 +79,10 @@ def __init__(self, inverse, domain, range=None, axes=None, sign='-',
             arrays.
             Otherwise, calculate the full complex FFT. If ``dom_dtype``
             is a complex type, this option has no effect.
-        impl : {'numpy', 'pyfftw'}, optional
+        impl : {'numpy', 'pyfftw', ``None``}, optional
             Backend for the FFT implementation. The 'pyfftw' backend
             is faster but requires the ``pyfftw`` package.
+            ``None`` selects the fastest available packagend.
```
kohr-h: typo here

PR #1007, opened 2017-05-22
ENH: Add additional protection against returning wrong value in operator, closes #962

### LINKED ISSUE #962: "out" as non-optional argument?
This is not really an issue, rather an implementation question.

In many ODL functions "out" is not declared an optional argument but still behaves as it was. Why is this? When I tried to write my own ODL prox operator this gave me a lot of headache as for me it did not behave like this. But rather than complaining it gave me nonsense as output.

Here is an example where I run a function without the "out" argument but the actual code does not declare it as being optional. Further, I define my own operator in a similar fashion but it does not behave similar

```python
import odl
from builtins import super

space = odl.uniform_discr([0, 0], [1, 1], [1, 1])
f = odl.solvers.proximal_cconj_l2_squared(space)
sigma = 2
x = odl.phantom.white_noise(space)

print(f(1)(x))

class Prox(odl.operator.Operator):

    def __init__(self, sigma):
        self.sigma = float(sigma)
        super().__init__(domain=space, range=space,
                         linear=False)

    def _call(self, x, out):
        out = 77
        return out
        
print(Prox(1)(x))
```
The output on screen:

```
[[-0.538313231696]]
[[-0.538313231696]]
```

adler-j:
Finally getting this done, now less bug prone.

### DIFF
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -50,7 +50,12 @@ def _default_call_out_of_place(op, x, **kwargs):
         evaluation.
     """
     out = op.range.element()
-    op._call_in_place(x, out, **kwargs)
+    result = op._call_in_place(x, out, **kwargs)
+    if result is not None and result is not out:
+        raise ValueError('`op` returned a different value than `out`.'
+                         'With in place evaluation, the operator can '
+                         'only return nothing (`None`) or the `out` '
+                         'parameter.')
     return out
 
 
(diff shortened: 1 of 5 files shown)


odl/operator/operator.py
```diff
@@ -50,7 +50,12 @@ def _default_call_out_of_place(op, x, **kwargs):
         evaluation.
     """
     out = op.range.element()
-    op._call_in_place(x, out, **kwargs)
+    result = op._call_in_place(x, out, **kwargs)
+    if result is not None and result is not out:
+        raise ValueError('`op` returned a different value than `out`.'
+                         'With in place evaluation, the operator can '
```
kohr-h: in-place

adler-j: Fixed in like 20 places, merging.

PR #1008, opened 2017-05-23
STYLE: Minor improvement to simple_operator example

adler-j:
Very minor improvement proposal


adler-j:
Well, we obviously need more examples in this folder. This is in a sense the most simple example you can think about so it makes sense as a "first example".

Issue #15, opened 2015-11-11
We need a logo


adler-j:
All the cool kids seem to have a logo of some sort, we need one aswell. Any input on what would look cool would be cool.

We need one big logo (to use in presentations etc). And one small logo to use in webpage thumbnails and such.

kohr-h:
Sure thing. Don't we have a school of arts nearby?

adler-j:
Vetenskap och **konst**, its OK we can do this!

adler-j:
Examples:

STIR (currently down)
![STIR](http://stir.sourceforge.net/images/stir_logo_lr.jpg)

NiftyRec
![Nifty](http://cmic.cs.ucl.ac.uk/uploads/RTEmagicC_NiftyRec.png.png)

VTK
![VTK](http://www.vtk.org/doc/nightly/html/vtk-logo.gif)

RTK
![RTK](http://www.openrtk.org/opensourcelogos/rtk75.png)

ASTRA
![ASTRA](http://fastra.ua.ac.be/images/astra_logo.png)

CONRAD
unknown/none

RVL
unknown/none

adler-j:
This could use some creative more input, @ozanoktem perhaps? :)

adler-j:
Pushing this once more, it would be nice for this to be up and running before an article.

kohr-h:
Yeah, creativity has not struck yet, obviously... @ozanoktem?
Also #20 is important (although the discussion went into a different direction there)

ozanoktem:
Ok ok ok ... I will try to install ODL on my laptop. I got the message!

adler-j:
This can once more be given some thought.

adler-j:
Bumping it go make you guys feel the pressure :)

@kohr-h 
@ozanoktem 
@chongchenmath

adler-j:
Bumping for justice as usual, this is more important than you realize.

ozanoktem:
So here is a proposal for constructing a logo. I suggest we take the word "ODL" (binary function), generate a sinogram from it and use that as a background that fades out into white space. On this background, we write "ODL", or perhaps the reconstruction of ODL.

kohr-h:
Here's a TV reconstruction from 20 projections, using fan beam and some noise
![odl_reco](https://cloud.githubusercontent.com/assets/5030250/26365528/250207ac-3fe8-11e7-9ade-6430d82cb649.png)

ozanoktem:
[...]

PR #1009, opened 2017-05-23
ENH: Add text phantom

adler-j:
Rather simple addition of a phantom that is simply some text. Useful as an example for many cases.

Inspired by our discussion in #15.

Examples:

```python
space = odl.uniform_discr([-1, -1], [1, 1], [300, 300])
for font in ['arial', 'impact', 'calibri']: 
    odl.phantom.text(space, text='ODL', font=font).show(font)
```

arial
calibri
impact

### DIFF
--- a/odl/phantom/misc_phantoms.py
+++ b/odl/phantom/misc_phantoms.py
@@ -146,15 +146,15 @@ def _submarine_2d_nonsmooth(space):
     return out.ufuncs.minimum(1, out=out)
 
 
-def text(space, text, font='arial', scale=0.8, inverted=True):
-    """Create phantom of text.
+def text(space, text, font='arial', border=0.2, inverted=True):
+    """Create phantom from text.
 
-    The text is represented by an scalar image taking values in [0, 1].
+    The text is represented by a scalar image taking values in [0, 1].
     Depending on the choice of font, the text may or may not be anti-aliased.
     anti-aliased text can take any value between 0 and 1, while
-    non-anti-aliased text takes values exclusively in {0, 1}.
+    non-anti-aliased text produces a binary image.
 
-    This method requires the PIL package.
+    This method requires the ``pillow`` package.
 
     Parameters
     ----------
@@ -162,20 +162,36 @@ def text(space, text, font='arial', scale=0.8, inverted=True):
         Discretized space in which the phantom is supposed to be created.
         Must be two-dimensional.
     text : str
-        The text that should be written in the phantom.
+        The text that should be written onto the background.
     font : str, optional
         The font that should be used to write the text. Available options are
         platform dependent.
-    scale : float, optional
-        Scaling of the text. 1.0 indicates that the phantom should occupy all
-        of the space along its largest dimension.
+    border : float, optional
+        Padding added around the text. 0.0 indicates that the phantom should
+        occupy all of the space along its largest dimension while 1.0 gives a
+        maximally padded image (text not visible).
     inverted : bool, optional
-        If the phantoms should be given in inverted style, i.e. white on black.
+        If the phantom should be given in inverted style, i.e. white on black.
 
     Returns
     -------
     phantom : ``space`` element
         The text phantom in ``space``.
+
+    Notes
+    -----
+    The set of available fonts is highly platform dependent, and there is no
+    obvious way (except from trial and error) to find what fonts are supported
+    on an arbitrary platform.
+
+    In general, the fonts ``'arial'``, ``'calibri'`` and ``'impact'`` tend to
+    be available.
+
+    Platform dependent tricks:
+
+    **Linux**::
+
+        $ find /usr/share/fonts -name "*.[to]tf"
     """
     from PIL import Image, ImageDraw, ImageFont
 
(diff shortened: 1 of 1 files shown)


odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
 
+def text(space, text, font='arial', scale=0.8, inverted=True):
+    """Create phantom of text.
+
+    The text is represented by an scalar image taking values in [0, 1].
+    Depending on the choice of font, the text may or may not be anti-aliased.
+    anti-aliased text can take any value between 0 and 1, while
+    non-anti-aliased text takes values exclusively in {0, 1}.
+
+    This method requires the PIL package.
```
kohr-h: Recommend [`pillow`](https://python-pillow.org/) rather than the deprecated (AFAIK) PIL. (Import stays the same.)

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
     return out.ufuncs.minimum(1, out=out)
 
 
+def text(space, text, font='arial', scale=0.8, inverted=True):
+    """Create phantom of text.
```
kohr-h: from

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+
+    The text is represented by an scalar image taking values in [0, 1].
```
kohr-h: a scalar image

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+    Depending on the choice of font, the text may or may not be anti-aliased.
+    anti-aliased text can take any value between 0 and 1, while
+    non-anti-aliased text takes values exclusively in {0, 1}.
```
kohr-h: produces a binary image?

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+    Parameters
+    ----------
+    space : `DiscreteLp`
+        Discretized space in which the phantom is supposed to be created.
+        Must be two-dimensional.
+    text : str
+        The text that should be written in the phantom.
+    font : str, optional
+        The font that should be used to write the text. Available options are
+        platform dependent.
```
kohr-h: Any way to find out in pillow which ones there are? On a system level, the easiest way under Unix-like platforms is `find /usr/share/fonts -name "*.[to]tf"`, just to get some.

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+    scale : float, optional
+        Scaling of the text. 1.0 indicates that the phantom should occupy all
+        of the space along its largest dimension.
```
kohr-h: `scale` doesn't really convey anything. How about inverting the logic and use `border=0.2`?

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+    from PIL import Image, ImageDraw, ImageFont
+
+    if space.ndim != 2:
+        raise ValueError('`space` must be two-dimensional')
+
+    text = str(text)
+
+    # Figure out what font size we should use by creating a very fine font
+    # and calculating its size
+    init_size = 1000
```
kohr-h: No need to store this, given the text above

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+        The text phantom in ``space``.
+    """
+    from PIL import Image, ImageDraw, ImageFont
+
+    if space.ndim != 2:
+        raise ValueError('`space` must be two-dimensional')
+
+    text = str(text)
+
+    # Figure out what font size we should use by creating a very fine font
```
kohr-h: "fine" somewhat clashes with the number 1000 below. Do you mean a *huge* font?

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+
+    This method requires the PIL package.
+
+    Parameters
+    ----------
+    space : `DiscreteLp`
+        Discretized space in which the phantom is supposed to be created.
+        Must be two-dimensional.
+    text : str
+        The text that should be written in the phantom.
```
kohr-h: written onto the background?

odl/phantom/misc_phantoms.py
```diff
@@ -146,6 +146,82 @@ def rect(x):
...
+    font : str, optional
+        The font that should be used to write the text. Available options are
+        platform dependent.
+    scale : float, optional
+        Scaling of the text. 1.0 indicates that the phantom should occupy all
+        of the space along its largest dimension.
+    inverted : bool, optional
+        If the phantoms should be given in inverted style, i.e. white on black.
```
kohr-h: phantom

odl/phantom/misc_phantoms.py
adler-j: Perhaps we can add that to the docstring

adler-j: It seems there is no easy way, but I'll add notes with some suggestions

odl/phantom/misc_phantoms.py
adler-j: fine as in fine details. But I'll change to "high resolution"

odl/phantom/misc_phantoms.py
adler-j: It is re-used a few lines below, trying to not violate DRY and also avoid magic numbers, so this will stay.

adler-j:
Updated according to comments. Ready to merge?

kohr-h:
> I guess the failing test is since pillow is not installed on Travis. Needs fix, I guess a fenced call at the bottom.

Seems to be only pep8 stuff. Add the fence for the call anyway.

PR #1010, opened 2017-05-26
BUG: fix bad return of in-place ReductionOperator

kohr-h:
The added check that `out` is returned captured this bug (albeit somewhat late since there were no tests).


adler-j:
Could you try to get this out of the way?

kohr-h:
Fixed and rebased, will merge when CI goes green.

PR #1012, opened 2017-05-29
(Not so) minor optimizations

adler-j:
This started out small but in the end i achieved a x2 speedup for small tomographic inversion problems, clearly nice.

Improvements include:

* MInor optimization on hot paths
    * Remove unncessary "isinstance" call
    * Remove repeated "size" etc calls
* Cache some often re-used stuff (like proximals) using `functools.lrc_cache`
* Reduce number of "element" calls by using proper inplace arithmetic in douglas rachford solver
* Performed some (rudimentary) performance tests and we now have three "levels" of lincomb, each optimal for its own case:
    * size < 100: Callback to pure numpy
    * 100 <= size <= 50000: Use blas style arithmetic but without blas
    * 50000 < size: Use blas
* Renamed `_lincomb` to `_lincomb_impl` since the previous did not play well with the spyder profiler

### DIFF
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -535,7 +535,6 @@ class DiscretizedSpaceElement(DiscretizedSetElement, FnBaseVector):
 
     def __init__(self, space, data):
         """Initialize a new instance."""
-        assert isinstance(space, DiscretizedSpace)
         DiscretizedSetElement.__init__(self, space, data)
 
     def __ipow__(self, p):
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -185,6 +185,9 @@ class RectGrid(Set):
                 raise ValueError('vector {} contains duplicates'
                                  ''.format(i + 1))
 
+        # Lazily evaluates strides when needed but stores the result
+        self.__stride = None
+
         self.__coord_vectors = vecs
 
         # Non-degenerate axes
@@ -381,19 +384,33 @@ class RectGrid(Set):
         If the grid contains axes that are not uniform, ``stride`` has
         a ``NaN`` entry.
 
+        For degenerate (length 1) axes, ``stride`` has value ``0.0``.
+
+        Returns
+        -------
+        stride : numpy.array
+            Array of dtype ``float`` and length `ndim`.
+
         Examples
         --------
         >>> rg = uniform_grid([-1.5, -1], [-0.5, 3], (2, 3))
         >>> rg.stride
         array([ 1.,  2.])
+
+        NaN returned for non-uniform dimension:
+
         >>> g = RectGrid([0, 1, 2], [0, 1, 4])
         >>> g.stride
         array([  1.,  nan])
+
+        0.0 returned for degenerate dimension:
+
+        >>> g = RectGrid([0, 1, 2], [0])
+        >>> g.stride
+        array([  1.,  0.0])
         """
         # Cache for efficiency instead of re-computing
-        try:
-            strd = self.__stride
-        except AttributeError:
+        if self.__stride is None:
             strd = []
             for i in range(self.ndim):
                 if not self.is_uniform_byaxis[i]:
@@ -403,9 +420,8 @@ class RectGrid(Set):
                 else:
                     strd.append(0.0)
             self.__stride = np.array(strd)
-            return self.__stride.copy()
-        else:
-            return strd.copy()
+
+        return self.__stride.copy()
 
     @property
     def extent(self):
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -390,15 +390,22 @@ class RectGrid(Set):
         >>> g.stride
         array([  1.,  nan])
         """
-        strd = []
-        for i in range(self.ndim):
-            if not self.is_uniform_byaxis[i]:
-                strd.append(float('nan'))
-            elif self.nondegen_byaxis[i]:
-                strd.append(self.extent[i] / (self.shape[i] - 1.0))
-            else:
-                strd.append(0.0)
-        return np.array(strd)
+        # Cache for efficiency instead of re-computing
+        try:
+            strd = self.__stride
+        except AttributeError:
+            strd = []
+            for i in range(self.ndim):
+                if not self.is_uniform_byaxis[i]:
+                    strd.append(float('nan'))
+                elif self.nondegen_byaxis[i]:
+                    strd.append(self.extent[i] / (self.shape[i] - 1.0))
+                else:
+                    strd.append(0.0)
+            self.__stride = np.array(strd)
+            return self.__stride.copy()
+        else:
+            return strd.copy()
 
     @property
     def extent(self):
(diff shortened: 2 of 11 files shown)


adler-j:
Added several new optimizations, squeezing out about 20% more perfomance in the afforementioned small examples.

Some short guide on how I profiled to do this:

```
python -m cProfile -o profiling_data.pyprof profile_mri.py
pyprof2calltree -i profiling_data.pyprof
```

here, cprofile ships with python and `pyprof2calltree`  can be installed via `pip install pyprof2calltree`

I then visualized the result using [KCacheGrind](http://kcachegrind.sourceforge.net/html/Home.html)

Overall this was quite easy and I don't have to feel as ashamed of my reconstruction times when I present, we should do it more often.

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -160,32 +160,54 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
-            prox_cc_g[i](sigma[i])(v[i] + (sigma[i] / 2.0) * L[i](w1),
-                                   out=p2[i])
+            tmp = v[i] + (sigma[i] / 2.0) * L[i](w1)
+            prox_cc_g[i](sigma[i])(tmp, out=p2[i])
             w2[i].lincomb(2.0, p2[i], -1, v[i])
 
-        tmp_2 = sum(Li.adjoint(wi) for Li, wi in zip(L, w2))
-        z1.lincomb(1.0, w1, - (tau / 2.0), tmp_2)
-        x += lam(k) * (z1 - p1)
+        # Compute tmp_domain = sum(Li.adjoint(wi) for Li, w2i in zip(L, w2))
```
adler-j: wi -> w2i

odl/space/npy_ntuples.py
```diff
@@ -491,81 +491,84 @@ def _blas_is_applicable(*args):
...
-        return x
-
-    def fallback_copy(x1, x2, n):
-        """Fallback copy implementation."""
-        x2[...] = x1[...]
-        return x2
-
-    if _blas_is_applicable(x1, x2, out):
+    # If data is very big, use BLAS if possible
+    if size >= 50000 and _blas_is_applicable(x1, x2, out):
```
kohr-h: Yes, please

odl/discr/discretization.py
```diff
@@ -535,7 +535,6 @@ class DiscretizedSpaceElement(DiscretizedSetElement, FnBaseVector):
 
     def __init__(self, space, data):
         """Initialize a new instance."""
-        assert isinstance(space, DiscretizedSpace)
```
kohr-h: Fully agree, I think I remove all of these in the #861 PR

odl/discr/grid.py
```diff
@@ -390,15 +390,22 @@ def stride(self):
...
-                strd.append(float('nan'))
-            elif self.nondegen_byaxis[i]:
-                strd.append(self.extent[i] / (self.shape[i] - 1.0))
-            else:
-                strd.append(0.0)
-        return np.array(strd)
+        # Cache for efficiency instead of re-computing
+        try:
+            strd = self.__stride
+        except AttributeError:
```
kohr-h: I don't like this pattern of trying to access some attribute of `self` and react on failure. I would clearly prefer to initialize the attribute to some nonsense ("sentinel") value like `None` and check for that instead of checking if it's there at all. The check `a is None` should incur no cost whatsoever and make this more robust.

odl/discr/grid.py
```diff
@@ -390,15 +390,22 @@ def stride(self):
...
+                if not self.is_uniform_byaxis[i]:
+                    strd.append(float('nan'))
+                elif self.nondegen_byaxis[i]:
+                    strd.append(self.extent[i] / (self.shape[i] - 1.0))
+                else:
+                    strd.append(0.0)
+            self.__stride = np.array(strd)
+            return self.__stride.copy()
+        else:
+            return strd.copy()
```
kohr-h: Why copy? I can see this as a general pattern, but that would require changing similar code in lots of places.

odl/util/numerics.py
```diff
@@ -280,14 +279,14 @@ def fast_1d_tensor_mult(ndarr, onedim_arrs, axes=None, out=None):
...
-            slc[ax] = np.s_[:]
+            slc[ax] = slice(None)
             factor = factor * arr[slc]
 
         out *= factor
 
         # Finally multiply by the remaining 1d array
         slc = [None] * out.ndim
-        slc[last_ax] = np.s_[:]
+        slc[last_ax] = slice(None)
```
kohr-h: Good change, now that we understand what's going on :-)

odl/util/utility.py
```diff
@@ -223,31 +223,55 @@ def __new__(cls, name, this_bases, d):
     return metaclass('temporary_class', None, {})
 
 
+def cache_arguments(function):
+    """Decorate function to cache the result with given arguments.
+
+    This is equivalent to `functools.lru_cache` with Python 3, and currently
+    does nothing with Python 2 but this may change at some later point.
```
kohr-h: WTF?? Okay, I guess this is still better than depending on some backport package for Python 2. The "later point" at which this changes will be the point when we drop Python 2 compatibility.

In the `future` library, ~~this seems to be available anyway: check [here](http://python-future.org/standard_library_imports.html?highlight=lru_cache)~~ OK, not without an extra `pip install`.

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -160,32 +160,54 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+    tmp_domain = x.space.zero()
+
     for k in range(niter):
-        tmp_1 = sum(Li.adjoint(vi) for Li, vi in zip(L, v))
-        tmp_1.lincomb(1, x, -tau / 2, tmp_1)
-        f.proximal(tau)(tmp_1, out=p1)
+        lam_k = lam(k)
+
+        # Compute tmp_domain = sum(Li.adjoint(vi) for Li, vi in zip(L, v))
+        L[0].adjoint(v[0], out=tmp_domain)
```
kohr-h: Do we reach this point if `L` is empty?

odl/discr/grid.py
adler-j: Well because otherwise there is a severe risk of stuff like this:

```
strides = grid.strides()
strides += 2

# lots of code

strides = grid.strides()  # returns bullshit
```

This (with the copy) is equivalent to the old code

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: Do we allow `L` empty? we did not before anyway.

I guess we can add an if statement for it, but I'm not even sure if the algorithm is valid in that case.

odl/util/utility.py
adler-j: I'll leave it for now. Python 2 is mostly supported because we have to in a sense. Minor optimizations of 10% should not really be expected on legacy platforms.

adler-j:
Fixed the comments, will merge.

Issue #1017, opened 2017-06-01
Add Figure Of Merits


zickert:
Add implementation of figure of merits to quantitatively compare reconstructions with known ground truth. Separate branch with corresponding pull request in preparation.

adler-j:
A general comment is that if we make a `odl.contrib.fom` package (which i think we should) then there are some sub-modules that should be kept separate (in my opinion). Some suggested such sub-modules would be (names are obviously preliminary):

* `unsupervised`: FOMs of the form `fom(reconstruction)` attempting to estimate features from a single image.
* `supervised`: FOMs of the form `fom(reconstruction, true)` computing some generalized distance
* `ensamble`: FOMs of the form `fom(reconstruction1, reconstruction2, ..., reconstructionn, [true])` computing some kind of batch statistic (e.g. noise sensitivity, etc)

adler-j:
I was thinking about this when implementing some other stuff and I realized that there are some features that would be nice to have. These do NOT need to be implemented in the PR, but should be added shortly after.

* More FOM, in particular we want "signal to noise ratio" (SNR) and "peak signal to noise ratio" (PSNR)
* Utilities to normalize values, either as a parameter to the FOMs (e.g. normalization=...) or by external functions. Examples include [z-scores](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.stats.zscore.html) and rescaling into [0, 1].
* Unsupervised FOMs, in particular some type of noise estimation similar to [this one](https://github.com/odlgroup/odl/pull/972/files#diff-2802eb089e5c95578ebafe78e56ac907R107)

Anyway, would be great to get the main skeleton in so we can start building on top of it.

adler-j:
Closed by #1018

PR #1018, opened 2017-06-01
Issue 1017  fom

zickert:
This is related to issue #1017

### DIFF
--- a/odl/contrib/fom/__init__.py
+++ b/odl/contrib/fom/__init__.py
@@ -1,0 +1,14 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+from __future__ import absolute_import
+
+__all__ = ()
+
+from .supervised import *
+__all__ += supervised.__all__
--- a/odl/contrib/fom/examples/supervised_comparison.py
+++ b/odl/contrib/fom/examples/supervised_comparison.py
@@ -1,0 +1,88 @@
+"""Simple example of implemented figures-of-merit (FOMs).
+
+Numerical test of a few implemented FOMs (mean square error, mean
+absolute error, mean value difference, standard deviation difference,
+range difference, blurring, false structures and structural similarity)
+as a function of increasing noise level.
+
+"""
+
+import odl
+from odl.contrib import fom
+import numpy as np
+import matplotlib.pyplot as plt
+
+# Discrete space: discretized functions on the rectangle
+# [-20, 20]^2 with 100 samples per dimension.
+reco_space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[100, 100])
+
+# Create a discrete Shepp-Logan phantom (modified version)
+phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+mse = []
+mae = []
+mvd = []
+std_diff = []
+range_diff = []
+blur = []
+false_struct = []
+ssim = []
+
+# Create mask for ROI to evaluate blurring and false structures. Arbitrarily
+# chosen as bone in Shepp-Logan phantom.
+mask = (np.asarray(phantom) == 1)
+
+for stddev in np.linspace(0.1, 10, 100):
+    phantom_noisy = phantom + odl.phantom.white_noise(reco_space,
+                                                      stddev=stddev)
+    mse.append(fom.mean_squared_error(phantom_noisy,
+                                      phantom,
+                                      normalized=True))
+
+    mae.append(fom.mean_absolute_error(phantom_noisy,
+                                       phantom,
+                                       normalized=True))
+
+    mvd.append(fom.mean_value_difference(phantom_noisy,
+                                         phantom,
+                                         normalized=True))
+
+    std_diff.append(fom.standard_deviation_difference(phantom_noisy,
+                                                      phantom,
+                                                      normalized=True))
+
+    range_diff.append(fom.range_difference(phantom_noisy,
+                                           phantom,
+                                           normalized=True))
+
+    blur.append(fom.blurring(phantom_noisy,
+                             phantom,
+                             mask,
+                             normalized=True,
+                             smoothness_factor=30))
+
+    false_struct.append(fom.false_structures(phantom_noisy,
+                                             phantom,
+                                             mask,
+                                             normalized=True,
+                                             smoothness_factor=30))
+
+    ssim.append(fom.ssim(phantom_noisy,
+                         phantom,
+                         normalized=True))
+
+plt.figure('Figures of merit')
+plt.plot(mse, label='Mean squared error')
+plt.plot(mae, label='Mean absolute error')
+plt.plot(mvd, label='Mean value difference')
+plt.plot(std_diff, label='Standard deviation difference')
+plt.plot(range_diff, label='Range difference')
+plt.plot(blur, label='Blurring')
+plt.plot(false_struct, label='False structures')
+plt.plot(ssim, label='Structural similarity')
+
+plt.legend(loc='upper center', bbox_to_anchor=(0.5, -0.15),
+           fancybox=True, shadow=True, ncol=1)
+plt.xlabel('Noise level')
+plt.ylabel('FOM')
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -1,0 +1,575 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""
+Implementation of Figures of Merit (FOMs) for comparing reconstructions with
+a given reference.
+"""
+
+import odl
+import numpy as np
+
+__all__ = ('mean_squared_error', 'mean_absolute_error',
+           'mean_value_difference', 'standard_deviation_difference',
+           'range_difference', 'blurring', 'false_structures', 'ssim')
+
+
+def mean_squared_error(data, ground_truth, mask=None, normalized=False):
+    """Return L2-distance between ``data`` and ``ground_truth``.
+
+    Evaluates `mean squared error
+    <https://en.wikipedia.org/wiki/Mean_squared_error>`_ between
+    input (``data``) and reference (``ground_truth``) Allows for normalization
+    (``normalized``) and a masking of the two spaces (``mask``).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+        \| f - g \|^2_2,
+
+    or, in normalized form
+
+    .. math::
+        \\frac{\| f - g \|^2_2}{\| f \|^2_2 + \| g \|^2_2}.
+
+    The normalized FOM takes values in [0, 1].
+
+    Parameters
+    ----------
+    data : `FnBaseVector`
+        Input data or reconstruction.
+    ground_truth : `FnBaseVector`
+        Reference to compare ``data`` to.
+    mask : `FnBaseVector`, optional
+        Mask to define ROI in which FOM evaluation is performed. The mask is
+        allowed to be weighted (i.e. non-binary), see ``blurring`` and
+        ``false_structures.``
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+
+    Returns
+    -------
+    fom : float
+        Scalar (float) indicating mean squared error between ``data`` and
+        ``ground_truth``. In normalized form the FOM takes values in
+        [0, 1], with higher correspondance at lower FOM value.
+    """
+    l2_normSquared = odl.solvers.L2NormSquared(data.space)
+
+    if mask is not None:
+        data = data * mask
+        ground_truth = ground_truth * mask
+
+    diff = data - ground_truth
+    fom = l2_normSquared(diff)
+
+    if normalized:
+            fom /= (l2_normSquared(data) + l2_normSquared(ground_truth))
+
+    return fom
+
+
+def mean_absolute_error(data, ground_truth, mask=None, normalized=False):
+    """Return L1-distance between ``data`` and ``ground_truth``.
+
+    Evaluates `mean absolute error
+    <https://en.wikipedia.org/wiki/Mean_absolute_error>`_ between
+    input (``data``) and reference (``ground_truth``). Allows for normalization
+    (``normalized``) and a masking of the two spaces (``mask``).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+        \| f - g \|_1,
+
+    or, in normalized form
+
+    .. math::
+        \\frac{\| f - g \|_1}{\| f \|_1 + \| g \|_1}.
+
+    The normalized FOM takes values in [0, 1].
+
+    Parameters
+    ----------
+    data : `FnBaseVector`
+        Input data or reconstruction.
+    ground_truth : `FnBaseVector`
+        Reference to compare ``data`` to.
+    mask : `FnBaseVector`, optional
+        Binary mask to define ROI in which FOM evaluation is performed.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+
+    Returns
+    -------
+    fom : float
+        Scalar (float) indicating mean absolute error between ``data`` and
+        ``ground_truth``. In normalized form the FOM takes values in
+        [0, 1], with higher correspondance at lower FOM value.
+    """
+    l1_norm = odl.solvers.L1Norm(data.space)
+    if mask:
+        data = data * mask
+        ground_truth = ground_truth * mask
+    diff = data - ground_truth
+    fom = l1_norm(diff)
+
+    if normalized:
+        fom /= (l1_norm(data) + l1_norm(ground_truth))
+
+    return fom
+
+
+def mean_value_difference(data, ground_truth, mask=None, normalized=False):
+    """Return difference in mean value between ``data`` and ``ground_truth``.
+
+    Evaluates difference in `mean value
+    <https://en.wikipedia.org/wiki/Mean_of_a_function>`_ between input
+    (``data``) and reference (``ground_truth``). Allows for normalization
+    (``normalized``) and a masking of the two spaces (``mask``).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+         \\bigg \\lvert \\lvert  \\overline{f} \\rvert -
+                \\lvert \\overline{g} \\rvert \\bigg \\rvert,
+
+    or, in normalized form
+
+    .. math::
+         \\bigg \\lvert \\frac{\\lvert \\overline{f} \\rvert -
+                               \\lvert \\overline{g} \\rvert}
+                              {\\lvert \\overline{f} \\rvert +
+                               \\lvert \\overline{g} \\rvert} \\bigg \\rvert
+
+    where
+
+    .. math::
+        \\overline{f} := \\frac{1}{\|1_\Omega\|_1} \\int_\Omega f dx,
+
+    and
+
+    .. math::
+        \\overline{g} := \\frac{1}{\|1_\Omega\|_1} \\int_\Omega g dx.
+
+    The normalized FOM takes values in [0, 1], with higher correspondance
+    at lower FOM value.
+
+    Parameters
+    ----------
+    data : `FnBaseVector`
+        Input data or reconstruction.
+    ground_truth : `FnBaseVector`
+        Reference to compare ``data`` to.
+    mask : `FnBaseVector`, optional
+        Binary mask to define ROI in which FOM evaluation is performed.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+
+    Returns
+    -------
+    fom : float
+        Scalar (float) indicating difference in mean value between
+        ``data`` and ``ground_truth``. In normalized form the FOM takes
+        values in [0, 1], with higher correspondance at lower FOM value.
+    """
+    l1_norm = odl.solvers.L1Norm(data.space)
+    if mask:
+        data = data * mask
+        ground_truth = ground_truth * mask
+
+    # Volume of space
+    vol = l1_norm(data.space.one())
+
+    data_mean = data.inner(data.space.one()) / vol
+    ground_truth_mean = ground_truth.inner(ground_truth.space.one()) / vol
+
+    fom = np.abs(np.abs(data_mean) - np.abs(ground_truth_mean))
+
+    if normalized:
+        fom /= (np.abs(data_mean) + np.abs(ground_truth_mean))
+
+    return fom
+
+
+def standard_deviation_difference(data, ground_truth, mask=None,
+                                  normalized=False):
+    """Return absolute difference in std between ``data`` and ``ground_truth``.
+
+    Evaluates difference in standard deviation (std) between input (``data``)
+    and reference (``ground_truth``). Allows for normalization (``normalized``)
+    and a masking of the two spaces (``mask``).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+         \\lvert \| f - \\overline{f} \|_2 -
+                 \| g - \\overline{g} \|_2 \\rvert,
+
+    or, in normalized form
+
+    .. math::
+        \\bigg \\lvert \\frac{\| f - \\overline{f} \|_2 -
+                              \| g - \\overline{g} \|_2}
+                             {\| f - \\overline{f} \|_2 +
+                              \| g - \\overline{g} \|_2 } \\bigg \\rvert,
+
+    where
+
+    .. math::
+        \\overline{f} := \\frac{1}{\|1_\Omega\|_1} \\int_\Omega f dx,
+
+    and
+
+    .. math::
+        \\overline{g} := \\frac{1}{\|1_\Omega\|_1} \\int_\Omega g dx.
+
+    The normalized FOM takes values in [0, 1], with higher correspondance
+    at lower FOM value.
+
+    Parameters
+    ----------
+    data : `FnBaseVector`
+        Input data or reconstruction.
+    ground_truth : `FnBaseVector`
+        Reference to compare ``data`` to.
+    mask : `FnBaseVector`, optional
+        Binary mask to define ROI in which FOM evaluation is performed.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+
+    Returns
+    -------
+    fom : float
+        Scalar (float) indicating absolute difference in standard deviation
+        between ``data`` and ``ground_truth``. In normalized form the FOM
+        takes values in [0, 1], with higher correspondance at lower FOM value.
+    """
+    l1_norm = odl.solvers.L1Norm(data.space)
+    l2_norm = odl.solvers.L2Norm(data.space)
+
+    if mask:
+        data = data * mask
+        ground_truth = ground_truth * mask
+
+    # Volume of space
+    vol = l1_norm(data.space.one())
+
+    data_mean = data.inner(data.space.one()) / vol
+    ground_truth_mean = ground_truth.inner(ground_truth.space.one()) / vol
+
+    fom = np.abs((l2_norm(data - data_mean) -
+                  l2_norm(ground_truth - ground_truth_mean)))
+
+    if normalized:
+        fom /= (l2_norm(data - data_mean) +
+                l2_norm(ground_truth - ground_truth_mean))
+
+    return fom
+
+
+def range_difference(data, ground_truth, mask=None, normalized=False):
+    """Return difference in range between ``data`` and ``ground_truth``.
+
+    Evaluates difference in range between input (``data``) and reference
+    data (``ground_truth``). Allows for normalization (``normalized``) and a
+    masking of the two spaces (``mask``).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+        \\lvert \\left(\\max(f) - \\min(f) \\right) -
+                \\left(\\max(g) - \\min(g) \\right) \\rvert
+
+    or, in normalized form
+
+    .. math::
+        \\bigg \\lvert \\frac{\\left(\\max(f) - \\min(f) \\right) -
+                              \\left(\\max(g) - \\min(g)\\right)}
+                             {\\left(\\max(f) - \\min(f)\\right) +
+                              \\left(\\max(g) - \\min(g)\\right)}
+        \\bigg \\rvert
+
+    The normalized FOM takes values in [0, 1], with higher correspondance
+    at lower FOM value.
+
+    Parameters
+    ----------
+    data : `FnBaseVector`
+        Input data or reconstruction.
+    ground_truth : `FnBaseVector`
+        Reference to compare ``data`` to.
+    mask : `FnBaseVector`, optional
+        Binary mask to define ROI in which FOM evaluation is performed.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized
+        FOM.alse_structures.
+
+    Returns
+    -------
+    fom : float
+        Scalar (float) indicating absolute difference in range between
+        ``data`` and ``ground_truth``. In normalized form the FOM takes
+        values in [0, 1], with higher correspondance at lower FOM value.
+    """
+    if mask:
+        indices = np.where(mask is True)
+        data_range = (np.max(data.asarray()[indices]) -
+                      np.min(data.asarray()[indices]))
+        ground_truth_range = (np.max(ground_truth.asarray()[indices]) -
+                              np.min(ground_truth.asarray()[indices]))
+    else:
+        data_range = np.max(data) - np.min(data)
+        ground_truth_range = np.max(ground_truth) - np.min(ground_truth)
+
+    fom = np.abs(data_range - ground_truth_range)
+
+    if normalized:
+        fom /= np.abs(data_range + ground_truth_range)
+
+    return fom
+
+
+def blurring(data, ground_truth, mask=None, normalized=False,
+             smoothness_factor=None):
+    """Return weighted L2-distance, emphasizing regions defined by ``mask``.
+
+    Evaluates `mean squared error
+    <https://en.wikipedia.org/wiki/Mean_squared_error>`_ between input
+    (``data``) and reference data (``ground_truth``) using an added binary
+    mask (``mask``), such that the error is weighted with higher importance
+    given to the defined structure-of-interest. Allows for normalization
+    (``normalized``).
+
+    .. note:: If omitting the mask argument, the blurring FOM is equivalent
+              to the mean squared error FOM.
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+        \|\\alpha (f - g) \|^2_2,
+
+    or, in normalized form
+
+    .. math::
+        \\frac{\| \\alpha (f - g) \|^2_2}{\| \\alpha f \|^2_2 +
+                                  \| \\alpha  g \|^2_2},
+
+    where :math:`\\alpha` is a weighting function with higher values near a
+    structure of interest defined by ``mask``. The weighting function is given
+    as
+
+    .. math::
+        \\alpha = e^{-\\frac{1}{k} \\beta},
+
+    where :math:`\\beta(x)` is the Euclidian distance from :math:`x` to the
+    complement of the structure of interest, and :math:`k`
+    (``smoothness_factor``) is a positive real number that controls the
+    'smoothness' of the weighting function :math:`\\alpha`.
+
+    The normalized FOM takes values in [0, 1], with higher correspondance
+    at lower FOM value.
+
+    Parameters
+    ----------
+    data : `FnBaseVector`
+        Input data or reconstruction.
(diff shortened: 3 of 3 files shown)


odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
...
+
+Numerical test of a few implemented FOMs (mean square error, mean
+absolute error, blur, and false structures) as a function of
+increasing noise level.
+
+"""
+
+import odl
+import numpy as np
+#import fom
```
adler-j: Missing

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
+"""Simple test of implemented figure-of-merits (FOMs)
+
+Numerical test of a few implemented FOMs (mean square error, mean
+absolute error, blur, and false structures) as a function of
+increasing noise level.
+
```
adler-j: Extra newline

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
+"""Simple test of implemented figure-of-merits (FOMs)
```
adler-j: This file should not be under `odl.fom`, it should go under `examples/fom`

adler-j: Also it is not a test, so title should be "example(s)" or something similar

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
...
+phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+mse = []
+mae = []
+blur = []
+false_struct = []
+
+mask = (np.asarray(phantom) == 1)
+
+for i in np.linspace(0.1, 1, 10):
```
adler-j: `i` is not a good name for "noise level", try `stddev` instead.

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Create a discrete Shepp-Logan phantom (modified version)
+phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+mse = []
+mae = []
+blur = []
+false_struct = []
+
+mask = (np.asarray(phantom) == 1)
```
adler-j: Why this choice of mask?

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
...
+import numpy as np
+#import fom
+import matplotlib.pyplot as plt
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 300 samples per dimension.
+reco_space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[100, 100], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
```
adler-j: Looks like copy paste. Why does this example even need tomography in it? It is never used later

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
...
+
+"""
+
+import odl
+import numpy as np
+#import fom
+import matplotlib.pyplot as plt
+
+# Discrete reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 300 samples per dimension.
```
adler-j: This does not agree with the text below

odl/fom/fom_example.py
```diff
@@ -0,0 +1,45 @@
...
+reco_space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[100, 100], dtype='float32')
```
adler-j: No real reason to use `float32`, so can be removed.

odl/fom/fom.py
```diff
@@ -0,0 +1,250 @@
...
+    -------
+    fom: scalar
+        Scalar (float) indicating perfect correspondance (1) to no
+        correspondance (0)
+    """
+    if mask is None:
+        mask = np.ones(reco.shape, dtype=bool)
+    l2_normSquared = odl.solvers.L2NormSquared(reco.space)
+    mask = scimorph.distance_transform_edt(1-mask)
+    mask = np.exp(-mask/30)
```
adler-j: Where does 30 come from?

odl/fom/fom.py
```diff
@@ -0,0 +1,250 @@
...
+    if len(np.unique(mask)) != 1:
```
adler-j: Why do this?

kohr-h:
Nice to see your first PR @zickert - thanks for the effort!

Instead of fixing everything within this PR, you can also 
- fix the coarse things that make Travis fail (PEP8, imports etc.),
- make this PR into `odl.contrib` instead, so it can go in immediately,
- fix the remaining issues there

It's up to you.

Regarding final placement `odl.util` is certainly an option. The FOM are general-purpose enough to pass that filter.

zickert:
@adler-j and @kohr-h 
Thank you for your comments, I will take a look at them and discuss with the rest of the people who have been working on this.

@kohr-h 
Several people within the group at KTH have worked on the code for the FOM's (and evaluation of reco methods using these), so it was a joint effort and I certainly can't take all credit ;)

adler-j:
Hi! Are we making any progress in this? Would be nice to get it in.

zickert:
Hi! David and I will work on this tomorrow.

kohr-h:
> Hi! David and I will work on this tomorrow.

Okay great, I'll try to fix #1020 until then.

zickert:
@davlars and I have now done some modifications, event though it's still very much WIP. The main change is though that it's now been moved to ```odl.contrib``` instead of as a ```odl.fom``` (the example has also been moved to ```odl/examples/contrib/```). We're still working on some changes indicated above (such as e.g. improved documentation, testing, etc.)

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    Allows for a masking of the two spaces (mask).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+        1 - \\frac{\| f - g \|^2_2}{\| f \|^2_2 + \| g \|^2_2}.
+
+    The FOM takes its values in [0 , 1], with higher correspondance
```
adler-j: extra space before `,`

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+        Binary mask to define ROI in which FOM evaluation is performed.
+
+    Returns
+    -------
+    fom : scalar
+        Scalar (float) indicating perfect correspondance (1) to no
+        correspondance (0)
+    """
+
+    import odl
```
adler-j: This is a circular-ish import. Instead import it explicitly.

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    diff = test_data - ref_data
+    fom = 1.0 - l2_normSquared(diff) / (l2_normSquared(test_data) +
+                                        l2_normSquared(ref_data))
+    return fom
+
+
+def mean_absolute_error(test_data, ref_data, mask=None):
+    """Mean absolute error-based FOM between test_data and ref_data.
+
+    Evaluates mean absolute error FOM between input (test_data) and
```
adler-j: Once more this is not what happens, since you apply normalization.

While this new incarnation is well documented, I'm still very much in favor of adding a `normalized=False` parameter and only performing the normalization in case users explicitly ask for it, otherwise I'm sure almost all users would expect the function "`mean_absolute_error`" to compute "the mean of the absolute error", not "one minus the mean of the absolute error divided by the sum of the one norms".

With that said, I guess we can let it go for now and fix it in a later release, but it would be a breaking change.

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    reference data (ref_data), by measuring consistency using the L1-norm.
+    Allows for a masking of the two spaces (mask).
+
+    Notes
+    ----------
+    The FOM evaluates
+
+    .. math::
+        1 - \\frac{\| f - g \|_1}{\| f \|_1 + \| g \|_1}.
```
adler-j: Somewhat skeptical to this choice of normalization. A short example as to why:

fom(1, 1) = 1 - 0 / 2 = 1
fom(1, 0) = 1 - 1  / 1 = 0 
fom(1, -1) = 1 - 2  / 2 = 0

Not quite sure if this is what we want, but perhaps it is as intended? Is there some standard reference on these normalized FoMs?

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    at higher FOM value.
+
+    Parameters
+    ----------
+    test_data : `LinearSpace`
+        Input data or reconstruction.
+    ref_data : `LinearSpace`
+        Reference data to compare 'reco' to.
+    mask : `LinearSpace`
+        Binary mask to define ROI in which FOM evaluatoin is performed.
```
adler-j: evaluatoin

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+
+    The FOM takes its values in [0 , 1], with higher correspondance
+    at higher FOM value.
+
+    Parameters
+    ----------
+    test_data : `LinearSpace`
+        Input data or reconstruction.
+    ref_data : `LinearSpace`
+        Reference data to compare 'reco' to.
```
adler-j: ~~'reco'~~ \`\`reco\`\`

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+        1 - \\frac{\| f - g \|_1}{\| f \|_1 + \| g \|_1}.
+
+    The FOM takes its values in [0 , 1], with higher correspondance
+    at higher FOM value.
+
+    Parameters
+    ----------
+    test_data : `LinearSpace`
+        Input data or reconstruction.
+    ref_data : `LinearSpace`
```
adler-j: I guess we're looking for `LinearSpaceElement` here?

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+
+    .. math::
+        1 - \\bigg \\lvert \\frac{ \\overline{f} - \\overline{g} }
+                                 {\\overline{f} + \\overline{g}} \\bigg
+                                                                 \\rvert.
+
+    where
+
+    .. math::
+        \\overline{f} := \\frac{1}{\|1_\Omega\|_1} \\int_\Omega f dx,
```
adler-j: I guess the `\\frac{1}{\|1_\Omega\|_1}` part will cancel as long as we use normalization?

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+        Binary mask to define ROI in which FOM evaluatoin is performed.
+
+    Returns
+    -------
+    fom: scalar
+        Scalar (float) indicating perfect correspondance (1) to no
+        correspondance (0)
+    """
+
+    import numpy as np
```
adler-j: Numpy is already imported

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    mask : `LinearSpace`
+        Binary mask to define ROI in which FOM evaluatoin is performed.
+
+    Returns
+    -------
+    fom: scalar
+        Scalar (float) indicating perfect correspondance (1) to no
+        correspondance (0)
+    """
+
```
adler-j: Extra newline

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    import numpy as np
+    import odl
```
adler-j: Circular import

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+    ref_data_mean = (1 / vol) * (l1_norm(pos_ref_data) -
+                                 l1_norm(neg_ref_data))
+
+    fom = 1 - np.abs((l2_norm(test_data - test_data_mean) -
+                      l2_norm(ref_data - ref_data_mean)) /
+                     ((l2_norm(test_data - test_data_mean) +
+                       l2_norm(ref_data - ref_data_mean))))
+
+    return fom
+mdv = []
```
adler-j: leftover?

odl/contrib/fom/fom.py
```diff
@@ -0,0 +1,434 @@
...
+        correspondance (0)
+    """
+    import numpy as np
+    import odl
+    import scipy.ndimage.morphology as scimorph
+
+    l2_normSquared = odl.solvers.L2NormSquared(test_data.space)
+    if mask is not None:
+        mask = scimorph.distance_transform_edt(1 - mask)
+        mask = np.exp(-mask / 30)
```
adler-j: magic number 30?

davlars:
@zickert and I were trying to incorporate most of you (very good!) comments but it seems some still slipped through undealt with... Anyhow, I agree that for a potential wider use of any implemented "mean square error", "mean absolute error", etc. those should follow standard conventions (i.e. non-Normalized). I think we were simply quite focused on out specific FOM usage :-)  

Anyhow, adding a flag and modifying the formulation should be a tricky task. We'll have a look at it and update it as soon as possible.

adler-j:
No problem! Everything is looking good and the hardest PR is always the first since it kinda sets the standard for what is to come. I think once the normalization is fixed this is basically good to go.

davlars:
Hmmm. What I meant to say was that adding a flag and modifying the formulation should NOT be a tricky task. We'll let you know when we've had a look at it...

adler-j:
> Hmmm. What I meant to say was that adding a flag and modifying the formulation should NOT be a tricky task. We'll let you know when we've had a look at it...

I guessed as much from the rest of the comment :)

adler-j:
Are the final fixes on their way here?

zickert:
> Are the final fixes on their way here?

Yes, @davlars and I will work on this tomorrow.

zickert:
@davlars and I have now added normalization flags and folder structure for supervised and unsupervised FOM's.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+           'false_structures')
+
+
+def mean_square_error(test_data, ref_data, mask=None, normalized=False):
+    """Mean square error-based FOM between ``test_data`` and ``ref_data``.
+
+    Evaluates mean square error between input (``test_data``) and
+    reference data (``ref_data``), by measuring consistency using the L2 -norm.
+    Allows for normalization (``normalized``) and a masking of the two spaces
+    (``mask``).
```
kohr-h: This doesn't really add any information. If there's nothing more to say that's not apparent from signature, one-line summary and `Parameters`, there's no need to write anything.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+    ref_data : `LinearSpaceElement`
+        Reference data to compare ``test_data`` to.
+    mask : `LinearSpaceElement`
+        Binary mask to define ROI in which FOM evaluation is performed.
+    normalized  : `Boolean`
+        Boolean flag to switch between unormalized and normalized FOM.
+
+    Returns
+    -------
+    fom : `Scalar`
```
kohr-h: \`Scalar\` -> float

(The principle is: be generous in what you accept and strict in what you return. That means, allow *compatible* objects as inputs that can be converted to what you work with, but return one and only one type.)

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+
+    Parameters
+    ----------
+    test_data : `LinearSpaceElementElement`
+        Input data or reconstruction.
+    ref_data : `LinearSpaceElement`
+        Reference data to compare ``test_data`` to.
+    mask : `LinearSpaceElement`
+        Binary mask to define ROI in which FOM evaluation is performed.
+    normalized  : `Boolean`
```
kohr-h: \`Boolean\` -> bool

(Explanation: Single-tick keywords are tried to match to names defined by us, that can be classes, functions or glossary terms. That only applies to things like \`Operator\` or \`LinearSpace\` but not to built-in types like `bool`, `int`, `float`, `list`, `tuple` etc.. For those we don't make links because there would be just too many.)

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+
+    Returns
+    -------
+    fom : `Scalar`
+        Scalar (float) indicating mean square error between ``test_data`` and
+        ``ref_data``. In normalized form the FOM takes values in [0, 1], with
+        higher correspondance at lower FOM value.
+    """
+
+    import odl
```
kohr-h: I'd say do this at module level. It won't lead to circular imports because `import odl` does not import stuff from `odl.contrib`.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+    pos_ref_data = ref_data.space.element(pos_ref_data)
+
+    # Identify negative part of ref_data
+    neg_ref_data = ref_data.copy()
+    neg_ref_data = neg_ref_data.asarray()
+    neg_ref_data[neg_ref_data > 0] = 0
+    neg_ref_data = ref_data.space.element(neg_ref_data)
+
+    # Volume of space
+    vol = l1_norm(test_data.space.one())
```
kohr-h: Way too expensive. Do `np.prod(test_data.space.partition.extent)` or even simpler `test_data.space.domain.volume`.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+
+    .. math::
+         \\lvert  \\overline{f} - \\overline{g} \\rvert,
+
+    or, in normalized form
+
+    .. math::
+        \\bigg \\lvert \\frac{ \\overline{f} - \\overline{g} }
+                                 {\\overline{f} + \\overline{g}} \\bigg
+                                                                 \\rvert.
```
kohr-h: This seems wrong. What if `g = -f`? Shouldn't it be absolute values of both `f` and `g`.

More generally, why would a normalized form make sense here?

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+    # Mean value of ref_data
+    ref_data_mean = (1 / vol) * (l1_norm(pos_ref_data) -
+                                 l1_norm(neg_ref_data))
+
+    fom = np.abs((test_data_mean - ref_data_mean))
+
+    if normalized:
+        fom /= (np.abs(test_data_mean) + np.abs(ref_data_mean))
+
+    return fom
```
kohr-h: What's the difference between this and
```python
mean_x = x.inner(x.space.one() * mask) / x.space.domain.volume
mean_y = y.inner(y.space.one() * mask) / y.space.domain.volume
mvd = np.abs(mean_x - mean_y)
if normalized:
    return mvd / (np.abs(mean_x) + np.abs(mean_y))
else:
    return mvd
```

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,570 @@
...
+                  l2_norm(ref_data - ref_data_mean)))
+
+    if normalized:
+        fom /= (l2_norm(test_data - test_data_mean) +
+                l2_norm(ref_data - ref_data_mean))
+
+    return fom
+
+
+def range_difference(test_data, ref_data, mask=None, normalized=False):
```
kohr-h: Is that range as in "dynamic range"? Would be more clear.

kohr-h:
And before I only mention stuff that should be changed: It's great to have a common signature to the FOMs with optional normalization and masking. Good effort also on the documentation side.

Oh, also before I forget it: the plural of "Figure of Merit" is "Figures of Merit", not "Figure of Merits".

zickert:
@davlars have just fixed some naming and references following comments from @kohr-h. Some comments remain to be addressed.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+    import odl.contrib.fom
+    import scipy.ndimage.morphology as scimorph
+
+    if mask is not None:
+        mask = scimorph.distance_transform_edt(1 - mask)
+        mask = np.exp(-mask / weight_factor)
+
+    fom = odl.contrib.fom.mean_square_error(data,
+                                            ground_truth,
+                                            mask=mask,
```
adler-j: Not at all obvious that this is legal given that you say in the `mean_square_error` doc that `mask` is binary. I guess that doc has to change.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+        emphasizing complement of regions defined by ``mask``.
+
+    Evaluates `mean squared error
+    <https://en.wikipedia.org/wiki/Mean_squared_error>`_ between input
+    (``data``) and reference data (``ground_truth``) using an added binary
+    mask (``mask``), such that the error is weighted with lower importance
+    given to the defined structure-of-interest. Allows for normalization
+    (``normalized``).
+
+    NOTE: If omitting the mask argument, the blurring FOM is equivalent to the
```
adler-j: Copy paste error?

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+        Scalar (float) indicating weighted mean squared error between
+        ``data`` and ``ground_truth``. In normalized form the FOM takes
+        values in [0, 1], with higher correspondance at lower FOM value.
+    """
+
+    import numpy as np
+    import odl.contrib.fom
+    import scipy.ndimage.morphology as scimorph
+
+    if mask is not None:
```
adler-j: The more I see them, the more I wonder if these (the e.g. "blurring" fom) really warrant their own function. Why wouldn't users simply do these two lines and then call whatever FOM they want?

Less work for us to maintain at least.

odl/contrib/fom/unsupervised/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2014-2017 The ODL contributors
```
adler-j: I'd remove this since empty files laying around does not help anything.

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+
+    where :math:`\\alpha` is a weighting function with higher values near a
+    structure of interest defined by ``mask``. The weighting function is given
+    as
+
+    .. math::
+        \\alpha = e^{-\\frac{1}{k} \\beta},
+
+    where :math:`\\beta(x)` is the Euclidian distance transform from :math:`x`
+    to the complement of the structure of interest, and :math:`k` is a positive
```
adler-j: What is `k` and how do i set it? (I know it is `weight_factor`, but this is not obvious to readers.)

examples/contrib/fom/fom_example.py
```diff
@@ -0,0 +1,115 @@
...
+false_struct = []
+
+# Create mask for ROI to evaluate blurring and false structures. Arbitrarily
+# chosen as bone in Shepp-Logan phantom.
+mask = (np.asarray(phantom) == 1)
+
+for stddev in np.linspace(0.1, 10, 100):
+    phantom_noisy = phantom + odl.phantom.white_noise(reco_space,
+                                                      stddev=stddev)
+    mse.append(odl.contrib.fom.supervised.
```
kohr-h: We shouldn't do that if we don't have to. `import a as b` puts one extra layer of obscurity to names which should be avoided if possible. The `import numpy as np` is an exception since basically everybody does it (including the Numpy people themselves). But in general, renaming imports should be avoided.

In this case, `from odl.contrib import fom` should be sufficient (it's the recommended way of importing `contrib` packages anyway).

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+    fom = l1_norm(diff)
+
+    if normalized:
+        fom /= (l1_norm(data) + l1_norm(ground_truth))
+
+    return fom
+
+
+def mean_value_difference(data, ground_truth, mask=None, normalized=False):
+    """FOM returning the  difference in mean value between ``data``
```
kohr-h: Since it's a function returning something, we tend to use `Return ...`.
Regarding FOM, agree, it's part of the `fom` package so it should be obvious that it's a FOM.

examples/contrib/fom/fom_example.py
adler-j: Ok, agree with you there.

examples/contrib/fom/fom_example.py
```diff
@@ -0,0 +1,115 @@
+"""Simple example of implemented figure-of-merits (FOMs)
```
adler-j: Also we agreed on moving these guys into `odl/contrib/fom/examples`

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+
+    import numpy as np
+    import odl.contrib.fom
+    import scipy.ndimage.morphology as scimorph
+
+    if mask is not None:
+        mask = scimorph.distance_transform_edt(1 - mask)
+        mask = np.exp(-mask / weight_factor)
+
+    fom = odl.contrib.fom.mean_square_error(data,
```
adler-j: I noticed that this call does not work since the function name is `mean_squared_error`, so this needs a fix.

zickert:
Just FYI, I have been working on this today and hope to fix the final things tomorrow :)

adler-j:
See my PR into this branch, perhaps you could add that while at it

zickert:
> See my PR into this branch, perhaps you could add that while at it
Sure.

odl/contrib/fom/fom.py
zickert: I agree that this might be a somewhat weird. Perhaps one could do |f-g|_1/(2* max( |f|_1, |g|_1) ) instead, but I am not sure that this is better. Regarding references we have loosely been following section 2.4 in [this article](http://www.sciencedirect.com/science/article/pii/S1047847701943382). I have not found any reference for normalized FOMs unfortunately.

odl/contrib/fom/supervised/supervised_fom.py
zickert: Changed to absolute values now. I am not sure it makes sense with normalization here, but I guess we wanted to have normalized FOMs for easy comparison between them.

odl/contrib/fom/supervised/supervised_fom.py
zickert: There seems to be no difference, so I'll change to this.

odl/contrib/fom/supervised/supervised_fom.py
zickert: I am not sure about this, perhaps someone could help clear this out. Wikipedia says the following about "dynamic range":

> It is measured either as a ratio or as a base-10 (decibel) or base-2 (doublings, bits or stops) logarithmic value of the difference between the smallest and largest signal values

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+    pos_ground_truth = ground_truth.space.element(pos_ground_truth)
+
+    # Identify negative part of ground_truth
+    neg_ground_truth = ground_truth.copy()
+    neg_ground_truth = neg_ground_truth.asarray()
+    neg_ground_truth[neg_ground_truth > 0] = 0
+    neg_ground_truth = ground_truth.space.element(neg_ground_truth)
+
+    # Volume of space
+    vol = data.space.domain.volume
```
zickert: I changed this now to
`vol = data.space.size*data.space.weighting.const`

odl/contrib/fom/supervised/supervised_fom.py
```diff
@@ -0,0 +1,575 @@
...
+
+    if normalized:
+        fom /= (np.abs(data_mean) + np.abs(ground_truth_mean))
+
+    return fom
+
+
+def standard_deviation_difference(data, ground_truth, mask=None,
+                                  normalized=False):
+    """FOM returning absolute difference in standard deviation between
```
zickert: Changed this to 

> Return absolute difference in standard deviation between ``data`` and
>     ``ground_truth``.

Is it OK to use abbreviations here to make it fit on one line?

odl/contrib/fom/supervised/supervised_fom.py
zickert: Yes I agree that e.g. the blurring and false_structures FOMs might be unnecessary. If you think that they will never be used we can remove them. Do you have some opinion on this @davlars ?

davlars: For the (internal) usage we're planning with the FOMs I think there is a point to make in having a mask input for blurring/false structures, such that all users are using the same decay function (using the abritrarily defined weight factor). For that specific reason I think we should keep them. But you could argue that you could decay mask any FOM...

odl/contrib/fom/supervised/supervised_fom.py
kohr-h: That formula can be off for spaces where the boundary pixels are not complete (the `nodes_on_bdry` option). So I suggest that you do
```python
try:
    vol = data.space.domain.volume
except AttributeError:
    vol = data.space.weighting.constant
```

odl/contrib/fom/supervised/supervised_fom.py
kohr-h: You also don't *have to* refer to the input parameters if that helps. But abbreviations are OK I think, if they're obvious.

adler-j: I say yes as long as it is elaborated on further below.

odl/contrib/fom/supervised/supervised_fom.py
zickert: We keep them for now then I guess.

adler-j: > But you could argue that you could decay mask any FOM...

I guess this is my main point.

Another option would be to create a wrapper function that adds this, i.e. (freestyle coded)

```python
def add_blurring_to_mask(mask, weight_factor):
    mask = scimorph.distance_transform_edt(1 - mask)
    mask = np.exp(-mask / weight_factor)
    return mask
```

and then simply call

```python
mask = add_blurring_to_mask(mask)
fom = odl.contrib.fom.mean_square_error(data,
                                        ground_truth,
                                        mask=mask)
```

but if you were to do this, then the name "mask" is obviously missleading, and I'd change it to "weighting" to emphasise that it can be any type floating point values.

odl/contrib/fom/supervised/supervised_fom.py
adler-j: @kohr-h's suggestion does not really make sense given that they give wildly different values, I guess you meant

```python
try:
    vol = data.space.domain.volume
except AttributeError:
    vol = data.space.size * data.space.weighting.constant
```

With that said, a much more simple option with only a minor speed penalty would be

```python
vol = l1_norm(data.space.one())
```

works in all cases.

zickert: `l1_norm(data.space.one())`

is the one that that we used originally, but @kohr-h thought it was 

> Way too expensive.

;)

zickert: Since you know this much better I guess I will leave this decision to you (?)

kohr-h: Let's also remind ourselves that this is gonna end up in `odl.contrib` so according to [the guidelines](https://github.com/odlgroup/odl/blob/master/odl/contrib/README.md) it's up to you guys (@zickert @davlars) to make the final decision. So we shouldn't over-discuss minor points.

> @kohr-h's suggestion does not really make sense given that they give wildly different values, I guess you meant
> 
> try:
>     vol = data.space.domain.volume
> except AttributeError:
>     vol = data.space.size * data.space.weighting.constant

Yeah I just copied from your comment @adler-j ;-)

zickert:
So I am almost ready to push now, but I have (a probably trivial) git-related question: 
Would  `git push --force` be an OK way of getting around the problem that I can't simply `git push` now that I have merged Jonas PR into my branch?

adler-j:
> So I am almost ready to push now, but I have (a probably trivial) git-related question: 
Would  `git push --force` be an OK way of getting around the problem that I can't simply `git push` now that I have merged Jonas PR into my branch?

Have you done `git pull --rebase`? That should merge the remote changes with your local changes.

If you force push you would overwrite what is on the remote, which I guess you don't want to do

odl/contrib/fom/supervised/supervised_fom.py
davlars: I think that is pretty neat. Maybe we can call it ```add_weighting_to_mask```, remove ```blurring``` and ```false_structures``` from utilities, and simply give these two (using ```mean_squared_error```) in the given example? Or would you mind keeping it @zickert? In the example, and as long as we stick to one ```weight_factor``` we should be fine

(I now I'm basically restating what @adler-j said, but just trying to clear things up for myself :-)

zickert: Yes this seems like a good way of doing it. I will keep things as they are now however, perhaps we could discuss/work on this together later? :-)

davlars: If it's a matter of shifting things around, I guess this can be adjusted later (by a second PR). The current implementation works, but the @adler-j's suggestion would be more generic. I'm fine with doing it like that, if the rest are as well.

kohr-h: Just do it, your call :-)

zickert:
> Have you done git pull --rebase? That should merge the remote changes with your local changes.
> 
> If you force push you would overwrite what is on the remote, which I guess you don't want to do

No, but I did that now. Thanks :)

odl/contrib/fom/supervised/supervised_fom.py
zickert: To add in a later PR sounds good to me.

davlars: Decided :-) And thanks for your efforts today :+1:

zickert:
I have now pushed a new commit. One big thing that remains to be addressed is the lack of reference for some FOMs.  For the definitions of the FOMs @davlars and I we have very loosely followed section 2.4 in [this article](http://www.sciencedirect.com/science/article/pii/S1047847701943382), but we have not found any reference for normalized FOMs.

odl/__init__.py
```diff
@@ -43,7 +43,7 @@
 from . import deform
 from . import ufunc_ops
 from . import datasets
-
+from . import contrib
```
adler-j: This line should be removed

adler-j: Second comment after reading more:

It seems like you need to rebase this onto master since this is far behind the current master

odl/contrib/fom/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+from __future__ import absolute_import
+
+__all__ = ()
+
+from .supervised_fom import *
```
adler-j: supervised_fom -> supervised

odl/contrib/fom/examples/supervised_fom_comparison.py
```diff
@@ -0,0 +1,87 @@
...
+    false_struct.append(fom.false_structures(phantom_noisy,
+                                             phantom,
+                                             mask,
+                                             normalized=True,
+                                             smoothness_factor=30))
+
+    ssim.append(fom.ssim(phantom_noisy,
+                         phantom,
+                         normalized=True))
+
```
adler-j: plt.figure('Figures of merit')

odl/contrib/fom/__init__.py
zickert: This also means that supervised_fom.py should be renamed to supervised.py, right?

odl/__init__.py
zickert: I'm a bit confused now, the line `from . import contrib` is present in the odlgroup/odl/master version of  `odl/__init__.py ` (?)

kohr-h: Yeah, this is actually correct. It only makes `contrib` appear as a name on the top level, but it's not filled with anything.

odl/contrib/fom/__init__.py
kohr-h: Exactly

zickert:
Oh, I guess I did two `git rebase` in the wrong order before the last push. Can it go in like this or should I change it?

kohr-h:
> Oh, I guess I did two git rebase in the wrong order before the last push. Can it go in like this or should I change it?

No, this looks a bit botched unfortunately. But as I see it you have everything isolated in one commit, so the easiest thing is to just make a new branch off `master` and cherry-pick the commit:
```
git fetch origin
git checkout origin/master
git branch --move issue-1017__fom issue-1017__fom_old  # move old cruft out of the way
git checkout -b issue-1017__fom  # new branch even with master
git cherry-pick  ff428c6  # FOM commit on top
git push -fu zickert issue-1017__fom  # force-push and set to track zickert/issue-1017__fom
```

zickert:
> No, this looks a bit botched unfortunately. But as I see it you have everything isolated in one commit,

Yes I squashed it into one commit.

> so the easiest thing is to just make a new branch off master and cherry-pick the commit:
> 
> git fetch origin
> git checkout origin/master
> git branch --move issue-1017__fom issue-1017__fom_old  # move old cruft out of the way
> git checkout -b issue-1017__fom  # new branch even with master
> git cherry-pick  ff428c6  # FOM commit on top
> git push -fu zickert issue-1017__fom  # force-push and set to track zickert/issue-1017__fom

I will do this, thanks.

kohr-h:
Yes, why not.

Issue #894, opened 2017-02-07
Add MRC test with an extended header


adler-j:
As highlighted by #892, we don't have any tests of the MRC functions with an extended header. We should add such a test.

kohr-h:
There's now [mrcfile](https://github.com/ccpem/mrcfile) which looks very good to me. It support standard and extended header, as well as some other goodies like memory-mapped files. We should slim down our own MRC file handler and just do some internal bookkeeping.

PR #1020, opened 2017-06-02
MAINT: add skeleton contrib package

kohr-h:
TODO:
- [x] Add document explaining how to add stuff here (separate README.md, same license, no big stuff, keep commit history clean etc.)
- [x] Mention `odl/contrib` in the development docs
- [x] Write that issues on the tracker specific to a contributed module will be assigned to the author (responsibility remains with the author)
- [x] Write section on testing
- [x] Move MRC stuff


adler-j:
Do you need any input on the README or will you fix it?

kohr-h:
> Do you need any input on the README or will you fix it?

I'll go ahead with a suggestion that you can comment on.

kohr-h:
Readme added, comments welcome.

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+**No:**
+
+- Code for specialized algorithms written specifically for scientific papers
+- Large files like images, datasets etc.
+- Code under licenses other than MPL-2.0
+
+
+## Procedure
+
+Since you do not have write access to the target location in the ODL repository, do the initial work on the code on a personal fork. Within this fork, add your contribution to the `odl/contrib` directory. When the code is ready, the steps are as follows:
```
adler-j: Missing newline after dot.

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Submit a pull request. If you would like an in-depth review by a maintainer, please indicate that in writing or by choosing a reviewer on the Github pull request page.
+- A reviewer will check the pull request as a whole and the commit history (and optionally the code).
+- If no in-depth review was asked for, the reviewer will simply merge the PR if everything is okay. Otherwise, merging is delayed until you indicate that you have modified the code according to review.
```
adler-j: Missing newline after dot. Well, in a few more places...

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Code for specialized algorithms written specifically for scientific papers
+- Large files like images, datasets etc.
+- Code under licenses other than MPL-2.0
+
+
+## Procedure
+
+Since you do not have write access to the target location in the ODL repository, do the initial work on the code on a personal fork. Within this fork, add your contribution to the `odl/contrib` directory. When the code is ready, the steps are as follows:
+
+- Submit a pull request. If you would like an in-depth review by a maintainer, please indicate that in writing or by choosing a reviewer on the Github pull request page.
```
adler-j: Missing newline after dot.

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Generic code that would fit into main ODL but is still experimental
+- Special-purpose code that is not generic enough to be a fit for main ODL (e.g. [figures of merit](https://github.com/odlgroup/odl/pull/1018) or application-specific code)
+- Extended, more complex examples than [the ones in main ODL](https://github.com/odlgroup/odl/tree/master/examples)
+
+**No:**
+
+- Code for specialized algorithms written specifically for scientific papers
+- Large files like images, datasets etc.
+- Code under licenses other than MPL-2.0
```
adler-j: Missing dot

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Commit history
+
+Please keep the commit history as short as possible. Consider squashing commits of addition/removal cycles or work-in-progress incremental commits.
+
+
+## Package layout
+
+All code goes into the `odl/contrib` subdirectory. If a contribution consists in a single Python module `mymodule.py`, it can be added directly to that directory. Otherwise, a subdirectory `mylib` should be created that contains all code, i.e., *modules*, *examples* and *tests*.
```
adler-j: Disagree. I'd say they all should go into a folder, mostly for future extensibility and improved compartamentalization.

adler-j: Also make it clear that it is OK (even recommended) if existing modules are appropriately extended. E.g. if they want to add a new FoM they should do so by adding a new function/file under the `odl.contrib.fom` package.

adler-j: Also note that if something has a clear correspondence to things in core ODL, it should use the same naming if possible.

E.g. if you add a special function that creates the geometry for your scanner, it should go under `odl.contrib.tomo.geometry`.

odl/contrib/README.md
adler-j: Also remove "by a maintainer", it should be obvious

odl/contrib/README.md
kohr-h: This is Markdown, so newlines will be kept as-is. That would make the one-line-per-sentence policy produce bad-looking documents. Will keep it like this.

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Submit a pull request. If you would like an in-depth review by a maintainer, please indicate that in writing or by choosing a reviewer on the Github pull request page.
+- A reviewer will check the pull request as a whole and the commit history (and optionally the code).
+- If no in-depth review was asked for, the reviewer will simply merge the PR if everything is okay. Otherwise, merging is delayed until you indicate that you have modified the code according to review.
+
+That's all.
+
+
+## Commit history
+
+Please keep the commit history as short as possible. Consider squashing commits of addition/removal cycles or work-in-progress incremental commits.
```
kohr-h: Yeah, let's go with that, good suggestion. That also solves the issue of cluttered and bloated commit histories.

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+                |        |-- example2.py
+                |        ...
+                |
+                |-- tests
+                |      |
+                ...    |-- test_mymodule1.py
+                       |-- test_mymodule2.py
+                       ...
+```
+Your package will then be importable as `from odl.contrib import mylib`.
```
kohr-h: Yes

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Here are some guidelines for contributors regarding the scope and the rules of this platform.
+
+
+## Scope
+
+**Yes:**
+
+- Generic code that would fit into main ODL but is still experimental
+- Special-purpose code that is not generic enough to be a fit for main ODL (e.g. [figures of merit](https://github.com/odlgroup/odl/pull/1018) or application-specific code)
```
kohr-h: True

kohr-h: What's the bug?

odl/contrib/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Extended, more complex examples than [the ones in main ODL](https://github.com/odlgroup/odl/tree/master/examples)
```
kohr-h: I prefer a directory tree per contributed package, and I don't really care about internal organization there. Having `contrib` in 3 places (`odl/contrib`, `examples/contrib` and `test/contrib`) will be super confusing, and forcing users to adhere to this organization is not desirable IMO.

odl/contrib/README.md
kohr-h: I'd make these last two recommendations, not requirements. Basically users should be free to choose their own structure.

kohr-h:
> Also note that we need to add a pytest --contrib option to run the tests in here.
> 
> We also need to comment somewhere that we WILL require that contrib code tests pass and that the code follows pep8 like the rest of ODL.

I think that's too harsh. People can run their tests with `pytest [--pep8] odl/contrib/mylib` and everything is fine. I won't bother running those tests myself, and I also don't want to add that to Travis.

Regarding PEP8, I'm okay with requiring that code to adhere to it. Mostly because I assume that `pytest --pep8` will descend into `odl/contrib` by default.

I assume (and for myself that's the case) that code there will have doctests at best, so we can to some extent assume that it's broken. When others complain about the code, that's a good opportunity to fix it, and if it's complex we can recommend unit tests. But I don't see that happening by default.

kohr-h:
OK, I changed the layout from everything in `odl/contrib` to split into modules, examples and tests. 

- Examples cannot be part of the module directory since they cause CI to fail (matplotlib).
- Tests could be, and I'm considering putting them back there. The advantage would be that they wouldn't be run together with the other tests, which would put less harsh requirements on them (failure, run time, ...). Opinion?
- Probably the testing stuff requires that packages in `odl/contrib` may not be broken in the sense of non-importable.

odl/contrib/README.md
adler-j: Perhaps the bug was fixed, but there still is no proper test coverage, see https://github.com/odlgroup/odl/issues/894.

odl/contrib/README.md
adler-j: Second thought, I agree with you. Having each contrib package self contained is better.

odl/contrib/README.md
adler-j: Sure it is a recommendation, but we should still enforce some kind of structure here so it kinda depends on the case. For totally obvious stuff (like adding a new fom) we should enforce it but for corner cases (adding support for your scanner with some pre-processing scripts etc) we can be lenient.

odl/contrib/README.md
```diff
@@ -0,0 +1,98 @@
...
+- Extended, more complex examples than [the ones in main ODL](https://github.com/odlgroup/odl/tree/master/examples).
+
+**No:**
+
+- Code for specialized algorithms written specifically for your latest paper,
+- Large files like images, datasets etc.,
+- Huge amounts of code with questionable usefulness,
+- Code under licenses other than MPL-2.0.
+
+For code of the type "your latest paper", we recommend using a separate Github repository like [this one](https://github.com/kohr-h/variable_lp_paper) and making a pull request to add your project to the "Projects using ODL" section of the main README.
```
adler-j: Link to the readme. Specifically to the "Projects using ODL" section.

odl/contrib/README.md
```diff
@@ -0,0 +1,98 @@
...
+**Important:** Responsibility for code that you contribute will remain with you. This means in particular that issues on the tracker about your module will be assigned to you. Of course, others will try to help if necessary.
+
+**Note:** If you consider your code to be useful for the main library, please consider filing a pull request anyway. We try to be as helpful as possible in our reviews and make suggestions, so that getting the code ready for merging doesn't take too long.
+Note also that contributed modules that attract significant attention and user adoption will be considered for core ODL.
+
+Here are some guidelines for contributors regarding the scope and the rules of this platform.
+
+
+## Scope
+
```
adler-j: Feels very empty to start with a list, perhaps add something like:

"To determine if your code is suited for `odl.contrib`, consider the following guidelines:"

odl/contrib/README.md
adler-j: Noticed that it is not there yet, I'll fix.

odl/contrib/README.md
```diff
@@ -0,0 +1,98 @@
...
+For code of the type "your latest paper", we recommend using a separate Github repository like [this one](https://github.com/kohr-h/variable_lp_paper) and making a pull request to add your project to the "Projects using ODL" section of the main README.
+
+
+## Procedure
+
+Since you do not have write access to the target location in the ODL repository, do the initial work on the code on [your personal fork](https://odl.readthedocs.io/dev/gitwash/forking_hell.html#forking). Within this fork, add your contribution to the `odl/contrib` directory. When the code is ready, the steps are as follows:
+
+- Submit a pull request. If you would like an in-depth review, please indicate so in writing or by choosing a reviewer on the Github pull request page.
+- We will check the pull request as a whole and (optionally) the code. Note that your tests must pass and the code must adhere to [PEP8](https://www.python.org/dev/peps/pep-0008/) for Travis CI to go green.
```
adler-j: must adhere to **the python coding style guide**

adler-j: ~~for Travis CI to go green~~

since this is part of the automatic tests run on all code.

odl/contrib/README.md
```diff
@@ -0,0 +1,98 @@
...
+- To keep the history clean and to avoid accidental additions and deletions that bloat the repository, we will squash your pull request into a single commit that summarizes the whole addition. You can choose to do that yourself and write [a good commit message](https://odl.readthedocs.io/dev/gitwash/development_workflow.html#the-commit-message).
+- If no in-depth review was asked for, we will simply merge the PR if everything is okay. Otherwise, merging is delayed until you indicate that you have modified the code according to review.
+
+That's all.
+
+
+## Package layout
```
adler-j: A very important question here is how we handle imports in contrib packages. We need to somehow enforce lazy importing here. Is that done by default as currently written?

odl/contrib/README.md
```diff
@@ -0,0 +1,98 @@
...
+If your code extends existing ODL functionality, consider using the same directory structure. For instance, if you add a new geometry, you may want to put it into `odl/contrib/my_new_geometry/tomo/geometry/mygeometry.py` and add a modified ray transform to `odl/contrib/my_new_geometry/tomo/operators/ray_trafo.py`.
+
+**Important:** The top-level `odl/contrib/__init__.py` should *never* be modified. In particular, it should not import any modules by default. This ensures that code in a submodule does not break or slow down the import of ODL as a whole.
+
+
+## Tests
+
+We recommend that you add [doctests](https://odlgroup.github.io/odl/dev/testing.html#doctests) to your code from the beginning. They are a natural way to show users examples of your code in action, while being like small unit tests. For complex functionality, you should consider adding [unit tests](https://odlgroup.github.io/odl/dev/testing.html#unit-tests).
+
+**Note:** If you add unit tests, they will be run together with the rest of the ODL test suite. This means that they must pass and should not take too long time.
```
adler-j: As mentioned somewhere, this should (imo) not be the case and these should only be run when then `--contrib` flag is used.

odl/contrib/mrc/README.md
```diff
@@ -0,0 +1,21 @@
+# mrc
```
adler-j: capital letters, also this title should be larger than others

Also, we really should add readme's like this to different places. The prime examples would be the examples folders.

odl/contrib/mrc/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Data I/O functionality for the [MRC2014](http://www.ccpem.ac.uk/mrc_format/mrc2014.php) file format [Che+2015] including [FEI extended headers](http://www.2dx.unibas.ch/documentation/mrc-software/fei-extended-mrc-format-not-used-by-2dx)
+- Support for related uncompressed binary file formats with fixed-size header
```
adler-j: missing dot

odl/contrib/mrc/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Example usage
```
adler-j: This examples says absolutely nothing as is. The hard part is clearly how to extract the data you need from the header. Any chance you can do that?

Further I'd probably prefer if this was added to the examples folder so that it is run by `pytest --contrib --examples` which ensures that it at least does *something*.

odl/contrib/mrc/mrc.py
```diff
@@ -20,7 +20,7 @@
 import struct
 import warnings
 
-from odl.tomo.data.uncompr_bin import (
+from .uncompr_bin import (
```
adler-j: The recommendation is to always do `odl.contrib.mrc.uncompr_bin`, any reason you went absolute paths here?

odl/contrib/README.md
kohr-h: Yes, currently nothing in `contrib` is imported when you `import odl`. I tested that with putting a sleep in the module code. It only applies when you explicitly `from odl.contrib import mylib`.
So lazy importing happens at the level of `odl/contrib`, and I guess that's good enough. With `odl/contrib` ignored by `pytest` by default this shouldn't pose any problems whatsoever. But I'll test this.

odl/contrib/README.md
kohr-h: I'll change that.

odl/contrib/mrc/mrc.py
kohr-h: Oh no, nothing specific. Probably a work-in-progress change while moving stuff around.

odl/contrib/README.md
adler-j: Is `from odl.contrib import mylib` unaffected by slow imports in `odl.contrib.some_other_lib`?

odl/util/testutils.py
```diff
@@ -229,6 +229,11 @@ def is_subdict(subdict, dictionary):
         reason='Need --benchmark option to run'
     )
 
+    skip_if_no_contrib = pytest.mark.skipif(
```
adler-j: I guess we want this to be automatic for all tests in the folder and not something that users need to add themselves? Not sure if we can do that though

kohr-h: Yeah, I think the same. This was just a WIP push while googling for a better solution.

odl/contrib/README.md
kohr-h: For the same reason (limited usefulness / completeness / testing) we should also consider moving the STIR bindings and the Proximal bindings to `contrib`. Frankly, when has any of that been used the last time?

odl/contrib/README.md
kohr-h: Yes, no interference.

odl/contrib/README.md
kohr-h: I didn't introduce a `--contrib` flag. Instead users should manually run `pytest odl/contrib/mylib` which is ignored by default.

odl/contrib/README.md
adler-j: I guess we could do that on both of those, but lets do that after this one is in.

odl/contrib/README.md
```diff
@@ -0,0 +1,95 @@
...
+ |                     |      ...
+ |                     |
+ |                     |--- tests
+ |                     ...    |-- test_mymodule1.py
+ |                            |-- test_mymodule2.py
+ ...                          ...
+```
+Your package will then be importable as `from odl.contrib import mylib`.
+
+If your code extends existing ODL functionality, consider using the same directory structure. For instance, if you add a new geometry, you may want to put it into `odl/contrib/my_new_geometry/tomo/geometry/mygeometry.py` and add a modified ray transform to `odl/contrib/my_new_geometry/tomo/operators/ray_trafo.py`.
```
kohr-h: Ah so that's what you meant? So replicate the tree directly in `odl/contrib` and then add custom stuff?
Not sure about that since it works against the isolated packages principle. I think it would make contributors feel less responsible for added stuff if it just becomes part of a bigger mass of stuff. If it lives in a directory on its own it's clear that it's their package, their responsibility, their credit.

I'm actually inclined to removing this recommendation altogether. I'd rather simply let them do whatever they want and just give comments when the code comes in.

odl/contrib/mrc/README.md
```diff
@@ -0,0 +1,26 @@
+# MRC
```
kohr-h: Well, as I wrote somewhere else, MRC doesn't provide enough information in itself to create a geometry, so there's not really a good way to get there in a front-page example. People using MRC know that from own painful experience :-)

kohr-h:
> From another comment I got the fealing that you were going to remove --contrib for now?

This is a left-over, needs to be removed. Good catch.

odl/contrib/README.md
adler-j: Well if we have no general answer I agree with you, remove and give comments.

odl/contrib/mrc/README.md
adler-j: I guessed as much. Perhaps there is a open dataset somewhere and you could add it to datasets and make a proper example?

kohr-h: I'll see if I can find one, in particular one that can simply be downloaded by URL. Perhaps @ozanoktem has an idea?

kohr-h:
Last things fixed, history cleaned up, rebased. Will merge after CI.

odl/contrib/mrc/README.md
adler-j: We could re-host it ourselves I guess, if there is none

adler-j:
Go ahead

Issue #1014, opened 2017-05-31
Potential memory leak


chongchenmath:
There seems to be a potential memory leak somewhere in `odl`. Running the code from @davlars [repo](https://github.com/davlars/ad-skull-reconstruction) to make reconstructions on simulated 3D data, the used RAM memory is continuously increasing. Trying a TV-reconstruction with the Chambolle-Pock solver we (@aringh and I) get the following memory usage in the iterations:
```
iter = 0
CPU usage: [11.6, 12.5, 49.4, 11.5, 9.2, 9.2, 56.5, 8.8]
RAM usage: svmem(total=67514511360, available=27549761536, percent=59.2, used=39263768576, free=20183449600, active=46021120000, inactive=513998848, buffers=47173632, cached=8020119552, shared=38600704)
SWAP usage: sswap(total=4294963200, used=2080911360, free=2214051840, percent=48.5, sin=28048973824, sout=45718224896)

[...]

iter = 100
CPU usage: [19.8, 19.1, 17.1, 31.8, 13.6, 14.0, 12.8, 75.2]
RAM usage: svmem(total=67514511360, available=25842774016, percent=61.7, used=40964677632, free=18139529216, active=48008773632, inactive=556429312, buffers=60911616, cached=8349392896, shared=39292928)
SWAP usage: sswap(total=4294963200, used=2048077824, free=2246885376, percent=47.7, sin=28079247360, sout=45718224896)

[...]

iter = 500
CPU usage: [16.5, 19.1, 25.1, 45.3, 14.7, 14.1, 13.9, 61.1]
RAM usage: svmem(total=67514511360, available=18392444928, percent=72.8, used=48325439488, free=11604901888, active=50951569408, inactive=3880685568, buffers=491245568, cached=7092924416, shared=43077632)
SWAP usage: sswap(total=4294963200, used=1929170944, free=2365792256, percent=44.9, sin=28226437120, sout=45718732800)
```
As can be seen, the % of used RAM memory is increased from 59.2 to 72.8 over the iterations.

We do not yet know where the leak is, and it needs further investigation. Can anyone suggest a good way to do it?

The code used to get the above out-prints is (note however, that as is it will take several hours to run 500 iterations)
```python
import odl
import numpy as np
import os
import adutils
import sys
import psutil

saveReco = True
savePath = '/home/chchen/dev/ad-skull-reconstruction/reco/'


# Create a text file and write everything both in terminal and in file
if not os.path.exists(savePath):
[...]

PR #1024, opened 2017-06-08
ENH: add callback that shows memory and cpu usage.

aringh:
Add callback that prints memory and cpu usage. As in the code example in issue #1014

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -859,12 +859,10 @@ class CallbackPrintHardwareUsage(SolverCallback):
 
     """Callback for printing memory and CPU usage.
 
-    To use this callback, the user needs to have the package ``psutil``
-    installed.
+    This callback requires the ``psutil`` package.
     """
 
-    def __init__(self, step=1, print_cpu=True, print_mem=True,
-                 print_swap=True, fmt_cpu='CPU usage (% each core): {}',
+    def __init__(self, step=1, fmt_cpu='CPU usage (% each core): {}',
                  fmt_mem='RAM usage: {}', fmt_swap='SWAP usage: {}'):
         """Initialize a new instance.
 
@@ -872,37 +870,28 @@ class CallbackPrintHardwareUsage(SolverCallback):
         ----------
         step : positive int, optional
             Number of iterations between output. Default: 1
-        print_cpu : bool, optional
-            If ``True`` then the CPU usage is printed. Default: ``True``
-        print_mem : bool, optional
-            If ``True`` then the RAM memory usage is printed. Default: ``True``
-        print_swap : bool, optional
-            If ``True`` then the SWAP memory usage is printed.
-            Default: ``True``
         fmt_cpu : string, optional
             Formating that should be applied. The CPU usage is printed as ::
 
                 print(fmt_cpu.format(cpu))
 
             where ``cpu`` is a vector with the percentage of current CPU usaged
-            for each core.
+            for each core. An empty format string disables printing of CPU
+            usage.
         fmt_mem : string, optional
             Formating that should be applied. The RAM usage is printed as ::
 
                 print(fmt_mem.format(mem))
 
-            where ``mem`` is the current RAM memory usaged.
+            where ``mem`` is the current RAM memory usaged. An empty format
+            string disables printing of RAM memory usage.
         fmt_swap : string, optional
             Formating that should be applied. The SWAP usage is printed as ::
 
                 print(fmt_swap.format(swap))
 
-            where ``swap`` is the current SWAP memory usaged.
-
-        Raises
-        ------
-        ImportError
-            If the package ``psutil`` is not installed.
+            where ``swap`` is the current SWAP memory usaged. An empty format
+            string disables printing of SWAP memory usage.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


aringh:
Travis CI seems to have crashed for not finding the package `psutil`, which is used in the callback to get memory and cpu states. Could @adler-j  or @kohr-h have a look at that? Is it bad to build more dependencies into `odl`?

odl/solvers/util/callback.py
```diff
@@ -854,6 +856,52 @@ def __repr__(self):
...
+
+        >>> callback = CallbackPrintMemory(step=10)
+        """
+        self.step = step
+        self.iter = 0
+
+    def __call__(self, _):
+        """Print the memory and CPU usage"""
+        if self.iter % self.step == 0:
+            print('CPU usage (% each core): {}'.format(
```
adler-j: I would prefer an optional format string as is the case in some other callbacks. Also keeps it too one line by default.

odl/solvers/util/callback.py
```diff
@@ -854,6 +856,52 @@ def __repr__(self):
             self.logy)
 
 
+class CallbackPrintMemory(SolverCallback):
```
adler-j: It prints more than memory so perhaps some other name. `CallbackPrintHardwareUse` or something.

odl/solvers/util/callback.py
```diff
@@ -13,6 +13,7 @@
 from future import standard_library
 standard_library.install_aliases()
 
+import psutil
```
adler-j: As discussed, Lazily import inside the call.

odl/solvers/util/callback.py
kohr-h: Suggestion:
- remove swap (not really needed)
- use format strings for each resource separately, like `fmt_cpu`, `fmt_mem`, with sensible defaults
- add flags `print_cpu` and `print_mem` to switch on/off

odl/solvers/util/callback.py
kohr-h: True, `Usage` it would be in any case.

aringh:
I am working on the updates. Where in the doc should the dependency be mentioned? I tried to check how it was done for ray transform (with astra) and FFT (with pyfftw), but I could not see a clear pattern...

aringh:
I think that all comments are addressed. However, as my previous comment indicates I was not sure of where in the doc to write about the requirements on `psutil`.

odl/solvers/util/callback.py
```diff
@@ -854,6 +855,113 @@ def __repr__(self):
...
+    def __init__(self, step=1, print_cpu=True, print_mem=True,
+                 print_swap=True, fmt_cpu='CPU usage (% each core): {}',
+                 fmt_mem='RAM usage: {}', fmt_swap='SWAP usage: {}'):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        step : positive int, optional
+            Number of iterations between output. Default: 1
+        print_cpu : bool, optional
```
adler-j: Suggestion: scrap the "print_xyz" things and infer it from the fmt strings (perhaps with a rename), e.g.

`swap`, default is string but anything that evaluates to False (e.g. do a `if swap:`) means don't print.

Much more simple interface.

odl/solvers/util/callback.py
```diff
@@ -854,6 +855,113 @@ def __repr__(self):
...
+            Formating that should be applied. The SWAP usage is printed as ::
+
+                print(fmt_swap.format(swap))
+
+            where ``swap`` is the current SWAP memory usaged.
+
+        Raises
+        ------
+        ImportError
+            If the package ``psutil`` is not installed.
```
adler-j: This is enough imo, no need to spam doc with info in this dependency. In 98% of cases people will get the error and run pip install psutil without even reading the doc.

odl/solvers/util/callback.py
```diff
@@ -854,6 +855,113 @@ def __repr__(self):
...
+        self.fmt_cpu = str(fmt_cpu)
+        self.fmt_mem = str(fmt_mem)
+        self.fmt_swap = str(fmt_swap)
+
+        self.iter = 0
+
+    def __call__(self, _):
+        """Print the memory and CPU usage"""
+
+        try:
```
adler-j: I prefer propagating the error as is, I.e. no try catch

odl/solvers/util/callback.py
```diff
@@ -854,6 +855,113 @@ def __repr__(self):
...
+        self.iter += 1
+
+    def reset(self):
+        """Set `iter` to 0."""
+        self.iter = 0
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        optargs = [('step', self.step, 1)]
+        inner_str = signature_string([], optargs)
```
adler-j: Missing stuff here

odl/solvers/util/callback.py
kohr-h: I'd write a sentence
```
This callback requires the ``psutil`` package.
```
in the class doc and just let the error propagate.

odl/solvers/util/callback.py
kohr-h: But you can keep the old names for the format string parameters, they're good. Add a sentence in the format string docs then that "An empty format string disables printing of XYZ usage".

kohr-h: Otherwise I agree.

adler-j: I agree with this comment

aringh:
I hope this should do :-)

aringh:
Related to this, in the debugging in #1014 we also use ``pympler`` to print new objects in memory. I [created a callback](https://github.com/aringh/odl/compare/callback_print_memory...aringh:callback_print_changes_in_memory?expand=1) for this as well in the same spirit as the callback in this PR. Is this of interest?

kohr-h:
> Related to this, in the debugging in #1014 we also use pympler to print new objects in memory. I created a callback for this as well in the same spirit as the callback in this PR. Is this of interest?

That kind of functionality definitely useful, but I'm a bit skeptical as to whether that requires a separate callback. The `pympler` library has quite a bit more to offer, like tracking class instances, print all kinds of statistics, etc.. 
So there I think it would make more sense write an example that just uses `CallbackApply` with a user-defined function that prints the interesting stuff.

PR #1027, opened 2017-06-09
MAINT: Performance improvements to chambolle pock

adler-j:
Improves some minor stuff in the `chambolle_pock_solver`. In a simple benchmark (the TGV example) this improved run-time from 19.5 seconds down to 14.6, a ~25% improvement.


adler-j:
No, the bug is an actual bug and we should fix it promptly. It has to do with the `__array_ufunc__` stuff recently introduced which changes some lookup orders. I guess we need a separate PR to get that fixed.

PR #1029, opened 2017-06-10
Issue 1015  show indices none

kohr-h:


examples/solvers/functional_basic_example.py
```diff
@@ -4,11 +4,7 @@
 information on functionals, see `the ODL functional guide
 <http://odlgroup.github.io/odl/guide/in_depth/functional_guide.html>`_
 """
-
-# Imports for common Python 2/3 codebase
-from __future__ import print_function, division, absolute_import
-from future import standard_library
-standard_library.install_aliases()
+from __future__ import print_function, division
```
adler-j: Not needed?

odl/space/pspace.py
```diff
@@ -873,24 +873,42 @@ def show(self, title=None, indices=None, **kwargs):
             else:
                 indices = list(np.linspace(0, self.size - 1, 4, dtype=int))
         else:
-            if isinstance(indices, tuple):
+            if (isinstance(indices, tuple) or
+                    (isinstance(indices, list) and
+                     not all(isinstance(idx, Integral) for idx in indices))):
+                # Tuples or lists containing non-integers index by axis.
```
adler-j: This disagrees with classical np indexing wherein tuples and lists are strictly different in indexing. Do we really want to differ?

adler-j:
Also we clearly need to get `__array_ufunc__` fixed asap since it is screwing up our merges.

odl/space/pspace.py
kohr-h: No it doesn't. Lists that contain slice objects behave the same as tuples in Numpy indexing:
```py
>>> a = np.arange(12).reshape((3, 4))
>>> a[[0, 1, 2]]
array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11]])
>>> a[[0, slice(1, 3)]]
array([1, 2])
>>> a[(0, slice(1, 3))]
array([1, 2])
```

examples/solvers/functional_basic_example.py
kohr-h: There is some printing in the example so that's needed. The rest not (maybe also not division).

kohr-h:
Rebased on master with the temporary fix, ready to get in after that. How about the rest? To me it seems there is no issue to be solved.

examples/solvers/functional_basic_example.py
adler-j: Well print with parenthesis works as well in py2 as py3, what the function does is allow it to be used as a first class object, e.g. pass it as an argument to a function. But you don't do that here.

adler-j:
Made a minor comment, then merge

odl/discr/lp_discr.py
```diff
@@ -880,6 +878,10 @@ def show(self, title=None, method='', coords=None, indices=None,
             raise ValueError('too many axes ({} > {})'.format(len(indices),
                                                               self.ndim))
 
+        # Map `None` to `slice(None)` in indices for syntax like `coords`
```
adler-j: Have you added this to docs?

odl/space/pspace.py
```diff
@@ -873,24 +873,42 @@ def show(self, title=None, indices=None, **kwargs):
...
+                # We use the first index for the current pspace and pass
+                # on the rest.
                 indices, kwargs['indices'] = indices[0], indices[1:]
 
+            # Support `indices=[None, 0, None]` like syntax
```
adler-j: This comment does not seem to correspond to what is below?

odl/space/pspace.py
```diff
@@ -873,24 +873,42 @@ def show(self, title=None, indices=None, **kwargs):
...
+        parts = self[indices]
+        if len(parts) > 1:
+            # Give unique titles by indexed part
+            for i, part, fig in zip(indices, parts, in_figs):
+                fig = part.show(title='{}. Part {}'.format(title, i), fig=fig,
+                                **kwargs)
+                figs += [fig]
+        else:
+            # Don't extend the title if there is only one plot
+            for i, part, fig in zip(indices, parts, in_figs):
```
adler-j: Why a for loop here? It should be length one, right?

kohr-h: Could also be length 0 (not sure what happens then, will check), so I don't want to unconditionally pick out `parts[0]` and such.

examples/solvers/functional_basic_example.py
kohr-h: Right, but it also avoids prints with parens producing tuples in the output in Py2, which is not a catastrophe but a bit ugly.

adler-j: There are no parens when you do this in py2:

```python
>>> print(2)
2
```

It only occurs when you use commas:

```python
>>> print(2, 3)
(2, 3)
```

which differs from python 3:

```python
>>> print(2, 3)
2 3
```

odl/space/pspace.py
kohr-h: Actually yes, but it was obscured a bit by the code above it since it changed what `indices` is.

odl/space/pspace.py
kohr-h: Made this more explicit.

odl/discr/lp_discr.py
kohr-h: Good point, I updated the docs.

examples/solvers/functional_basic_example.py
kohr-h: Oh you're right, `print(2)` is the same as `print (2)`, which is of course not a tuple.

PR #1030, opened 2017-06-12
MAINT: remove examples/__init__.py

kohr-h:
Somehow left-over in the source tree. It messes up the package layout quite a bit since examples are installed directly under `site-packages`. This fixes the issue, but an intermediate release is needed so that users can get a fixed package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
     - conda info -a
 
     # Install dependencies and enter test environment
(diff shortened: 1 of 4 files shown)


kohr-h:
I think I'll temporarily add `numpy<1.13` to the dependencies until the issue is resolved. Will merge after that.

adler-j:
Go ahead

PR #1031, opened 2017-06-12
WIP: implement __array_ufunc__, closes #225

### LINKED ISSUE #225: Using __array_ufunc__
Since numpy 1.10 the function `__numpy_ufunc__` is available, this allows us to "steal" calls such as:

`np.multiply(x, y)` and use our own implementation. This could be useful for CUDA, and would allow users to use the more traditional `np.absolute(x)` syntax instead of only allowing `x.ufunc.absolute()`.

kohr-h:
This is a work-in-progress PR for the new interface. Would be nice to get some high-level input before I put a lot of work into fixing the details.

I chose to move most of the logic in `odl.util.ufuncs` to the `__array_ufunc__` method and retail the old interface as a wrapper around the new one, together with a deprecation warning. The warnings become very annoying over time, so I'll try to find a way to only print the warning once. Probably some module global flag.

The `ntuples_test` suite goes green, but probably we need more detailed tests now that we can use more parameters in the ufuncs (e.g. `dtype`) which require creating a new space.

### DIFF
--- a/odl/space/base_ntuples.py
+++ b/odl/space/base_ntuples.py
@@ -293,41 +293,6 @@ class NtuplesBaseVector(with_metaclass(ABCMeta, object)):
         """Number of bytes this vector uses in memory."""
         return self.size * self.itemsize
 
-    def __array__(self, dtype=None):
-        """Return a Numpy array containing this vector's data.
-
-        Parameters
-        ----------
-        dtype :
-            Specifier for the data type of the output array.
-
-        Returns
-        -------
-        array : `numpy.ndarray`
-        """
-        if dtype is None:
-            return self.asarray()
-        else:
-            return self.asarray().astype(dtype, copy=False)
-
-    def __array_wrap__(self, obj):
-        """Return a new vector from the data in ``obj``.
-
-        Parameters
-        ----------
-        obj : `numpy.ndarray`
-            Array that should be wrapped.
-
-        Returns
-        -------
-        vector : `NtuplesBaseVector`
-            Numpy array wrapped back into this vector's element type.
-        """
-        if obj.ndim == 0:
-            return self.space.field.element(obj)
-        else:
-            return self.space.element(obj)
-
     def __int__(self):
         """Return ``int(self)``.
 
@@ -394,14 +359,40 @@ class NtuplesBaseVector(with_metaclass(ABCMeta, object)):
             raise TypeError('only size 1 vectors can be converted to complex')
         return complex(self[0])
 
-    def __str__(self):
-        """Return ``str(self)``."""
-        return array1d_str(self)
+    def __array__(self, dtype=None):
+        """Return a Numpy array containing this vector's data.
 
-    def __repr__(self):
-        """Return ``repr(self)``."""
-        return '{!r}.element({})'.format(self.space,
-                                         array1d_repr(self))
+        Parameters
+        ----------
+        dtype :
+            Specifier for the data type of the output array.
+
+        Returns
+        -------
+        array : `numpy.ndarray`
+        """
+        if dtype is None:
+            return self.asarray()
+        else:
+            return self.asarray().astype(dtype, copy=False)
+
+    def __array_wrap__(self, obj):
+        """Return a new vector from the data in ``obj``.
+
+        Parameters
+        ----------
+        obj : `numpy.ndarray`
+            Array that should be wrapped.
+
+        Returns
+        -------
+        vector : `NtuplesBaseVector`
+            Numpy array wrapped back into this vector's element type.
+        """
+        if obj.ndim == 0:
+            return self.space.field.element(obj)
+        else:
+            return self.space.element(obj)
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         """Interface to Numpy's ufunc machinery.
@@ -452,8 +443,8 @@ class NtuplesBaseVector(with_metaclass(ABCMeta, object)):
             Method on ``ufunc`` that should be called on ``self``.
             Possible values:
 
-            ``'__call__'``, ``'accumulate'``, ``'at'``, ``'outer'``,
-            ``'reduce'``, ``'reduceat'``.
+            ``'__call__'``, ``'accumulate'``, ``'at'``, ``'reduce'``,
+            ``'reduceat'``
 
         input1, ..., inputN:
             Positional arguments to ``ufunc.method``.
@@ -603,7 +594,7 @@ numpy.ufunc.reduceat.html
             # We need to raise since returning `NotImplemented` will
             # fall back to native Numpy, resulting in a `numpy.ndarray`.
             # TODO: add when tensors are available
-            raise TypeError('`outer` not supported for 1-dimensional spaces')
+            raise NotImplementedError('`outer` currently not supported')
         else:
             if out is None:
                 result = getattr(ufunc, method)(*inputs, **kwargs)
@@ -627,16 +618,15 @@ numpy.ufunc.reduceat.html
 
     @property
     def ufuncs(self):
-        """Internal class for access to Numpy style universal functions.
+        """Access to Numpy style universal functions.
 
         These default ufuncs are always available, but may or may not be
         optimized for the specific space in use.
 
         .. note::
-            This interface is deprecated and will be removed as soon
-            as Numpy 1.13 becomes the minimum required version.
-            Use Numpy ufuncs directly, e.g., ``np.sqrt(x)`` instead of
-            ``x.ufuncs.sqrt()``.
+            This interface is will be deprecated when Numpy 1.13 becomes
+            the minimum required version. Use Numpy ufuncs directly, e.g.,
+            ``np.sqrt(x)`` instead of ``x.ufuncs.sqrt()``.
         """
         return NtuplesBaseUfuncs(self)
 
@@ -691,6 +681,15 @@ numpy.ufunc.reduceat.html
     def impl(self):
         """Implementation of this vector's space."""
         return self.space.impl
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return array1d_str(self)
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{!r}.element({})'.format(self.space,
+                                         array1d_repr(self))
 
 
 class FnBase(NtuplesBase, LinearSpace):
--- a/odl/space/base_ntuples.py
+++ b/odl/space/base_ntuples.py
@@ -23,7 +23,7 @@ from odl.util.ufuncs import NtuplesBaseUfuncs
 from odl.util import (
     array1d_repr, array1d_str, dtype_repr,
     is_scalar_dtype, is_real_dtype, is_floating_dtype,
-    complex_dtype, real_dtype)
+    complex_dtype, real_dtype, writable_array)
 from odl.util.utility import with_metaclass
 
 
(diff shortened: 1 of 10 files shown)


adler-j:
Build still failing? Will review anyway

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+                    kwargs['out'] = (out_arr,)
+                    ufunc(*inputs, **kwargs)
+                return out
+
+            elif ufunc.nout == 2:
+                if out1 is None:
+                    out1 = out_space.element()
+                if out2 is None:
+                    out2 = out_space.element()
+                with writable_array(out1, **array_kwargs) as out1_arr:
```
adler-j: Use a "multiple with" instead

https://stackoverflow.com/questions/893333/multiple-variables-in-python-with-statement

```python
with A() as a, B() as b, C() as c:
```

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+numpy.ufunc.reduce.html
+        """
+        return self.__array_ufunc__(
+            np.multiply, 'reduce', self,
+            axis=axis, dtype=dtype, out=(out,), keepdims=keepdims)
+
+    def min(self, axis=None, dtype=None, out=None, keepdims=False):
+        """Return the minimum of ``self``.
+
+        See the `ufunc.reduce documentation`_ for an explanation of the
```
adler-j: Why not something like `numpy.min` documentation?

odl/space/npy_ntuples.py
```diff
@@ -429,10 +429,18 @@ def __setitem__(self, indices, values):
         else:
             self.data[indices] = values
 
+    # Old ufuncs interface, deprecated
```
adler-j: deprecated since

odl/test/space/ntuples_test.py
```diff
@@ -1531,61 +1521,83 @@ def test_ufuncs(fn, ufunc):
...
+        out_arr_kwargs = {'out': out_arrays[0]}
+        out_vec_kwargs = {'out': out_vectors[0]}
+    elif nout == 2:
+        out_arr_kwargs = {'out1': out_arrays[0], 'out2': out_arrays[1]}
+        out_vec_kwargs = {'out1': out_vectors[0], 'out2': out_vectors[1]}
+    else:
+        assert False
+
+    # Get function to call, depending on the preferred interface
+    if USE_OLD_UFUNC_INTERFACE:
```
adler-j: Old should always be tested as long as we support it. Logic should be that we test the new interface only if available.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -199,6 +199,7 @@ def __repr__(self):
         """Return ``repr(self)``."""
         return '{}({!r})'.format(name, self.domain)
 
+    """
```
adler-j: Huh?

odl/util/ufuncs.py
```diff
@@ -68,63 +70,50 @@
...
-
-        elif n_out == 1:
+        if n_out == 1:
             def wrapper(self, out=None):
-                if out is None:
-                    out = self.vector.space.element()
-
-                out[:] = wrapped(self.vector)
-                return out
+                if parse_version(np.__version__) >= parse_version('1.13'):
```
adler-j: Perhaps we should not do this. What if someone (ourselves, for example) want to support old numpy and hence use the old interface?

odl/util/ufuncs.py
```diff
@@ -160,68 +136,24 @@ def __init__(self, vector):
...
+    def prod(self):
+        return self.vector.prod()
 
-    elif n_in == 2:
-        if n_out == 1:
-            def wrapper(self, x2, out=None):
-                if out is None:
-                    out = self.vector.space.element()
+    def min(self):
+        return self.vector.min()
```
adler-j: Do these work if users have numpy 1.12 or did we just break some users?

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
                                          array1d_repr(self))
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        """Interface to Numpy's ufunc machinery.
+
+        This method is called by Numpy version 1.13 and higher as a single
+        point for the ufunc dispatch logic. An object implementing
+        ``__array_ufunc__`` takes over control when a `numpy.ufunc` is
+        called on it, allowing it to use custom implementations and
+        output types.
```
kohr-h: OK

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+            When using operations that alter the shape (like ``reduce``),
+            or the data type (can be any of the methods),
+            the resulting array is wrapped in a space of the same
+            type as ``self.space``, however **only** using the minimal
+            set of parameters ``size`` and ``dtype``. If more properties
+            are supposed to be propagated, this method must be overridden.
+
+        Parameters
+        ----------
+        ufunc : `numpy.ufunc`
```
kohr-h: Yes, it is.

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+            ``'reduce'``, ``'reduceat'``.
+
+        input1, ..., inputN:
+            Positional arguments to ``ufunc.method``.
+        kwargs:
+            Keyword arguments to ``ufunc.method``.
+
+        Returns
+        -------
+        ufunc_result : # TODO fill in
```
kohr-h: Yeah sure, was too lazy to look up (or maybe the Numpy docs weren't very clear about this) what the function actually returns.

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+
+        The ``add.at`` method operates in-place. Here we add the second
+        operand ``[5, 10]`` to ``x`` at indices ``[0, 2]``:
+
+        >>> x = rn.element([1, 2, 3])
+        >>> np.add.at(x, [0, 2], [5, 10])
+        >>> x
+        rn(3).element([6.0, 2.0, 13.0])
+
+        The output of ``add.outer`` would have ``ndim = 2`` axes, so we
```
kohr-h: I'd say it's okay for now, in particular when using this in `DiscreteLp` and perhaps `ProductSpace` later. The order can be specified in the ufunc call, defaulting to `'C'`. In that sense, `ravel` is perfectly well-defined since the result of `outer` is contiguous either in C or F order, and `ravel` will use it to flatten by default.

odl/space/base_ntuples.py
kohr-h: Clearly better

odl/space/base_ntuples.py
kohr-h: Yeah, makes more sense.

odl/test/space/ntuples_test.py
kohr-h: Sure. I was getting annoyed by thousands of warnings from using the old interface :-)
Clearly I should find a way of printing the deprecation warning only once.

odl/ufunc_ops/ufunc_ops.py
kohr-h: That's just temporary since it made importing ODL fail as long as the interface wasn't working properly.

odl/util/ufuncs.py
kohr-h: Okay, maybe not. An alternative transition plan is to fully support this interface until Numpy 1.13 becomes minimum, then add a deprecation warning and remove it in the release after that. This also gives us some leeway to port our own code.

odl/util/ufuncs.py
kohr-h: I don't think so but I'll check. The point is that `__array_ufunc__` is always *available*, no matter which Numpy version is used. So the methods using it will always find it. The only difference is the behavior of `numpy.some_ufunc(odl_vec, ...)` which uses the new interface from 1.13 on, and doesn't for older versions.

odl/space/base_ntuples.py
adler-j: Well, since there is no acute use-case of this, why don't we simply leave it undefined for now and then later change it to work like it does for arrays. Having wildly different behavior is just weird imo.

Also we'll want stuff like `np.multiply.outer(odl.uniform_discr(0, 1, 10).one(), [1, 2])` to work decently, I'd personally expect it to return a vector valued space.

odl/test/space/ntuples_test.py
adler-j: I'm we wait with the warning for a while, as mentioned somewhere else.

odl/util/ufuncs.py
adler-j: I agree.

odl/space/base_ntuples.py
kohr-h: That's actually a tricky case. Should that be a product space like thing or a new `DiscreteLp` (but then on which interval?).
Okay, I'm convinced, I'll leave this case undefined for now and think about the `DiscreteLp` case some more.

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+
+        .. _reduceat documentation:
+           https://docs.scipy.org/doc/numpy/reference/generated/\
+numpy.ufunc.reduceat.html
+        """
+        # Unwrap out if provided. The output parameters are all wrapped
+        # in one tuple, also if there is only one.
+        out_tuple = kwargs.pop('out', ())
+
+        # We allow our own vectors and `numpy.ndarray` objects as `out`
```
kohr-h: Any comment on this? I thought it'd be nice to allow Numpy arrays as `out` but I'm not 100 % about it.

adler-j: I'd say go ahead with allowing numpy arrays.

odl/space/base_ntuples.py
```diff
@@ -403,6 +403,281 @@ def __repr__(self):
...
+                    # Wrap result in an appropriate space
+                    # TODO: use `shape` and `order` when tensors are available
+                    result_space = type(self.space)(result.size, result.dtype)
+                    return result_space.element(result.ravel())
+            else:
+                with writable_array(out) as out_arr:
+                    kwargs['out'] = (out_arr,)
+                getattr(ufunc, method)(*inputs, **kwargs)
+
+    # Some often used reductions (may not make sense in all cases)
```
kohr-h: Okay to put them into the class namespace?

adler-j: I'd actually say no since we want people to use the free functions whenever possible and we've seen some confusion regarding this before.

odl/util/ufuncs.py
kohr-h: Tested, runs fine with Numpy 1.12

odl/space/base_ntuples.py
kohr-h: Actually it makes sense also because it makes `arr += odl_vec` work immediately.

kohr-h:
Okay, I'm not putting any more effort into this PR. The design works, but taking it further would mean re-implementing in a quirky way stuff that is already done in #861 (space indexing mostly, plus order, vector-valued functions etc.). 

My suggestion is that we go with Numpy 1.12. for now and I'll implement the interface in #861 since it just comes much more naturally there.

adler-j:
Sure, so the conclusion is that we close this for now and instead focus on getting #861 done?

kohr-h:
Yes, it should be a matter of a few weeks. I'll port the code.

PR #1032, opened 2017-06-13
Fix release doc refs

kohr-h:
Bad references in docs. I temporarily changed the `.travis.yml` to ensure that the docs are built for this PR, will remove it again when it succeeds.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -78,7 +78,6 @@ def skip(app, what, name, obj, skip, options):
     if name in ['_multiply',
                 '_divide',
                 '_lincomb',
-                '_apply',
                 '_call']:
         return False
     return skip
@@ -90,12 +89,11 @@ def setup(app):
 # Autosummary
 autosummary_generate = glob.glob("./*.rst")
 
-# Default role for ambiguous ':any:' targets
-default_role = 'py:class'
-
 # Stops WARNING: toctree contains reference to nonexisting document
+# (not anymore since Sphinx 1.6)
 numpydoc_show_class_members = True
 numpydoc_show_inherited_class_members = True
+numpydoc_class_members_toctree = True
 
 # Set order to mirror source
 autodoc_member_order = 'bysource'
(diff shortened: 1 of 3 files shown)


kohr-h:
Previous builds failed because `sphinx` would output such a massive amount of warnings that the log would become too big (> 4 Mb) and Travis would bail out. This current change breaks the docs in bad ways, so that's not an option either. 

Piping in `.travis.yml` is not allowed, so we cannot filter the warnings on the console without wrapping it into some kind of custom script.
An alternative would be to simply disable all warnings in sphinx (or at least all that can be silenced) and check manually once in a while. Also kinda ugly.

Currently there are ~32000 warnings, we need to do something about it. For reference:

- A bunch of actual errors
- Several `WARNING: document isn't included in any toctree`
- Some `WARNING: Citation [XYZ] is not referenced`¨
- Many `WARNING: more than one target found for 'any' cross-reference`
- TONS of `<autosummary>:: WARNING: toctree contains reference to nonexisting document`
  Basically one such warning for each class and each attribute as it seems

doc/source/conf.py
```diff
@@ -91,11 +90,12 @@ def setup(app):
 autosummary_generate = glob.glob("./*.rst")
 
 # Default role for ambiguous ':any:' targets
-default_role = 'py:class'
+default_role = ':py:class:'
 
 # Stops WARNING: toctree contains reference to nonexisting document
 numpydoc_show_class_members = True
 numpydoc_show_inherited_class_members = True
+numpydoc_class_members_toctree = False  # this should avoid a bunch of warnings
```
kohr-h: but also breaks other things

doc/source/conf.py
```diff
@@ -134,7 +134,7 @@ def setup(app):
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['*.py', '*.pyc']
+exclude_patterns = ['*.py', '*.pyc', 'doc/source/generated']
```
kohr-h: this was just a desperate attempt, breaks stuff of course

doc/source/conf.py
```diff
@@ -91,11 +90,12 @@ def setup(app):
 autosummary_generate = glob.glob("./*.rst")
 
 # Default role for ambiguous ':any:' targets
-default_role = 'py:class'
+default_role = ':py:class:'
 
 # Stops WARNING: toctree contains reference to nonexisting document
```
kohr-h: No it doesn't

kohr-h:
For now the sphinx build redirects stderr to /dev/null so the log doesn't explode in size.

PR #1035, opened 2017-06-15
Issue 206  premade geometries

adler-j:
Planned to only add a contrib.tomo.elekta package but got some performance issues while doing it that I fixed.

With that said, this does not currently work on my machine, specifically the returned volume is all zeros. I've checked and it seems like we're feeding ASTRA non-zero stuff but ASTRA returns all zeros (I checked before and after the astra call) so that needs debugging.

Could use a review of the performance stuff specifically.

### DIFF
--- a/odl/contrib/tomo/elekta.py
+++ b/odl/contrib/tomo/elekta.py
@@ -1,0 +1,125 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Tomography helpers for Elekta systems."""
+
+import numpy as np
+import odl
+
+
+__all__ = ('elekta_icon_geometry', 'elekta_icon_space',
+           'elekta_icon_reconstruction')
+
+
+def elekta_icon_geometry(sad=780, sdd=1000,
+                         piercing_point=[390, 0],
+                         angles=np.linspace(1.2, 5.0, 332)):
+    """Tomographic geometry of the Elekta Icon CBCT system.
+
+    See the [whitepaper]_ for specific descriptions of each parameter.
+
+    All measurments are given in millimeters unless otherwise stated.
+
+    Parameters
+    ----------
+    sad : float, optional
+        Source to Axis distance.
+    sdd : float, optional
+        Source to Detector distance.
+    piercing_point : array-like, optional
+        Position in the detector (in pixel coordinates) that a beam from the
+        source, passing through the axis of rotation at a straight angle,
+        hits.
+    angles : array-like, optional
+        List of angles that the projection images were taken at.
+        Given in radians.
+
+    See Also
+    --------
+    elekta_icon_space : Default reconstruction space for the Elekta Icon CBCT.
+
+    References
+    ----------
+    .. [whitepaper] *Design and performance characteristics of a Cone Beam
+       CT system for Leksell Gamma Knife Icon*
+    """
+
+    sad = float(sad)
+    assert sad > 0
+    sdd = float(sdd)
+    assert sdd > sad
+    piercing_point = np.array(piercing_point, dtype=float)
+    assert piercing_point.size == 2
+    angles = np.array(angles, dtype=float)
+
+    # Constant system parameters
+    pixel_size = 0.368
+    det_shape_pixels = np.array([780, 720])
+
+    # Create a (possible non-uniform) partition given the angles
+    angle_partition = odl.nonuniform_partition(angles)
+
+    # Compute the detector partition
+    pp_mm = pixel_size * piercing_point
+    det_min_pt = -pp_mm
+    det_max_pt = det_min_pt + pixel_size * det_shape_pixels
+    detector_partition = odl.uniform_partition(min_pt=det_min_pt,
+                                               max_pt=det_max_pt,
+                                               shape=det_shape_pixels)
+
+    # Create the geometry
+    geometry = odl.tomo.CircularConeFlatGeometry(
+        angle_partition, detector_partition,
+        src_radius=sad, det_radius=sdd - sad,
+        axis=[0, 0, 1])
+
+    return geometry
+
+
+def elekta_icon_space(**kwargs):
+    """Default reconstruction space for the Elekta Icon CBCT.
+
+    See the [whitepaper]_ for further information.
+
+    Parameters
+    ----------
+    kwargs :
+        Keyword arguments to pass to `uniform_discr` to modify the space, e.g.
+        use another backend.
+
+    Returns
+    -------
+    elekta_icon_geometry : DiscreteLp
+
+    Examples
+    --------
+    >>> space = odl.contrib.tomo.elekta_icon_volume()
+    """
+    return odl.uniform_discr(min_pt=[-112.0, -112.0, 0.0],
+                             max_pt=[112.0, 112.0, 224.0],
+                             shape=(448, 448, 448),
+                             dtype='float32')
+
+
+def elekta_icon_reconstruction(ray_transform, parker_weighting=True):
+    """Approximation of the FDK reconstruction used in the Elekta Icon."""
+
+    fbp_op = odl.tomo.fbp_op(ray_transform, padding=False,
+                             filter_type='Hann', frequency_scaling=0.6)
+    if parker_weighting:
+        parker_weighting = odl.tomo.parker_weighting(ray_transform)
+        fbp_op = fbp_op * parker_weighting
+
+    return fbp_op
+
+
+if __name__ == '__main__':
+    # Run doctests
+    # pylint: disable=wrong-import-position
+    from odl.util.testutils import run_doctests
+    run_doctests()
--- a/odl/contrib/tomo/elekta.py
+++ b/odl/contrib/tomo/elekta.py
@@ -13,10 +13,10 @@ import odl
 
 
 __all__ = ('elekta_icon_geometry', 'elekta_icon_space',
-           'elekta_icon_reconstruction')
+           'elekta_icon_fbp')
 
 
-def elekta_icon_geometry(sad=780, sdd=1000,
+def elekta_icon_geometry(sad=780.0, sdd=1000.0,
                          piercing_point=(390.0, 0.0),
                          angles=np.linspace(1.2, 5.0, 332),
                          detector_shape=(780, 720)):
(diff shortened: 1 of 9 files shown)


odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,125 @@
...
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Tomography helpers for Elekta systems."""
+
+import numpy as np
+import odl
```
adler-j: Not quite sure about this but it works. Perhaps we should have a talk on it

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,125 @@
...
+        src_radius=sad, det_radius=sdd - sad,
+        axis=[0, 0, 1])
+
+    return geometry
+
+
+def elekta_icon_space(**kwargs):
+    """Default reconstruction space for the Elekta Icon CBCT.
+
+    See the [whitepaper]_ for further information.
```
adler-j: dead link

odl/tomo/backends/astra_cuda.py
```diff
@@ -39,12 +39,7 @@ class AstraCudaProjectorImpl(object):
 
     """Thin wrapper around ASTRA."""
 
-    algo_id = None
-    vol_id = None
-    sino_id = None
-    proj_id = None
-
-    def __init__(self, geometry, reco_space, proj_space, use_cache):
+    def __init__(self, geometry, reco_space, proj_space):
```
adler-j: Overall I have no idea why we did not do this before. Isn't this a strict improvement in readability and maintainability? For example I noticed a "small" memory "leak" (we forgot to remove the in and out arrays in the backprojection) which lost me 1gb of memory and this was trivially fixed by changing this.

adler-j:
It turns out that the "everything is zero" thing had to do with ASTRA not playing well with disk paging (i.e. I ran out of memory). This is a bit shameful that we do this for problems of this size so it is something we need to look into. I did some improvements here but I guess there is more we could do.

Anyway, now it works.

odl/contrib/tomo/elekta.py
kohr-h: As long as the `odl/contrib/__init__.py` does nothing there is no problem since `import odl` doesn't care about anything in `odl/contrib`. So I currently don't see a way this can go wrong.

adler-j:
Rebased to master with proper contrib package. needs review.

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+
+    See Also
+    --------
+    elekta_icon_space : Default reconstruction space for the Elekta Icon CBCT.
+    elekta_icon_fbp: Default reconstruction method for the Elekta Icon CBCT.
+
+    References
+    ----------
+    .. [whitepaper] *Design and performance characteristics of a Cone Beam
+       CT system for Leksell Gamma Knife Icon*
```
kohr-h: Some more info here? Where published or where available? One way to do this would be to make the text `[whitepaper]` (formatting also off here) a hyperlink and have two items for it in the `References` section, one for the link and one like now, just without the `.. ` so that the section is actually printed. Like so:
```
See the `[whitepaper]`_ for specific descriptions of each parameter.

...

References
----------
.. _[whitepaper]: <insert URL here>
[whitepaper] *Design and performance characteristics of a Cone Beam
    CT system for Leksell Gamma Knife Icon*
```

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    sad = float(sad)
+    assert sad > 0
+    sdd = float(sdd)
+    assert sdd > sad
+    piercing_point = np.array(piercing_point, dtype=float)
+    assert piercing_point.size == 2
```
kohr-h: Test `piercing_point.shape == (2,)`

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+    angles = np.array(angles, dtype=float)
+    detector_shape = np.array(detector_shape, dtype=int)
+
+    # Constant system parameters
+    pixel_size = 0.368
+    det_shape_mm = np.array([287.04, 264.96])
```
kohr-h: extent instead of shape?

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+
+    See Also
+    --------
+    elekta_icon_geometry: Geometry for the Elekta Icon CBCT.
+    elekta_icon_fbp: Default reconstruction method for the Elekta Icon CBCT.
+
+    References
+    ----------
+    .. [whitepaper] *Design and performance characteristics of a Cone Beam
+       CT system for Leksell Gamma Knife Icon*
```
kohr-h: See above

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+    return odl.uniform_discr(min_pt=[-112.0, -112.0, 0.0],
+                             max_pt=[112.0, 112.0, 224.0],
+                             shape=shape,
+                             dtype='float32')
+
+
+def elekta_icon_fbp(ray_transform,
+                    padding=False, filter_type='Hann', frequency_scaling=0.6,
+                    parker_weighting=True):
+    """Approximation of the FDK reconstruction used in the Elekta Icon.
```
kohr-h: Elekta Icon system?

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Parameters
+    ----------
+    ray_transform : `RayTransform`
+        The ray transform to be used, should have an Elekta Icon geometry.
+    padding : bool, optional
+        If the fbp filter should use padding, increases memory use
```
kohr-h: If -> Whether
fbp -> FBP

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+        significantly.
+    filter_type : str, optional
+        Type of filter to apply in the FBP filter.
+    frequency_scaling : float, optional
+        Frequency scaling for FBP filter
+    parker_weighting : bool, optional
+        True if parker weighting should be applied to compensate for finite
```
kohr-h: Parker
finite? I guess you mean "less than 360 degrees", maybe "incomplete" or "less than full-circle".

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+        angle scan.
+
+    Returns
+    -------
+    elekta_icon_fbp : `DiscreteLp`
+
+    Examples
+    --------
+    Create default fbp for default geometry:
```
kohr-h: FBP

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+
+    >>> from odl.contrib import tomo
+    >>> geometry = tomo.elekta_icon_geometry()
+    >>> space = tomo.elekta_icon_space()
+    >>> ray_transform = odl.tomo.RayTransform(space, geometry)
+    >>> fbp_op = tomo.elekta_icon_fbp(ray_transform)
+    """
+
```
kohr-h: remove blank

odl/contrib/tomo/elekta.py
```diff
@@ -0,0 +1,198 @@
...
+    if parker_weighting:
+        parker_weighting = odl.tomo.parker_weighting(ray_transform)
+        fbp_op = fbp_op * parker_weighting
+
+    return fbp_op
+
+
+if __name__ == '__main__':
+    # Run doctests
+    # pylint: disable=wrong-import-position
```
kohr-h: I'd say screw it and just let pylint complain.

odl/contrib/tomo/examples/elekta_icon_tomography.py
```diff
@@ -0,0 +1,42 @@
...
+space = tomo.elekta_icon_space(shape=(112, 112, 112))
+
+# Create ray transform
+ray_transform = odl.tomo.RayTransform(space, geometry,
+                                      use_cache=False)
+
+# Get default FDK reconstruction
+recon_op = tomo.elekta_icon_fbp(ray_transform)
+
+# Create simplified-phantom
```
kohr-h: remove dash

odl/contrib/tomo/examples/elekta_icon_tomography.py
```diff
@@ -0,0 +1,42 @@
+"""Example of using the Elekta Icon geometry.
+
+In this example we create an Elekta Icon geometry and use it to create some
+artificial data and reconstruct it using backprojection.
+
+Note that this is a fully 3d dataset and requires some memory to run.
```
kohr-h: ~~fully~~

odl/contrib/tomo/examples/elekta_icon_tomography.py
```diff
@@ -0,0 +1,42 @@
+"""Example of using the Elekta Icon geometry.
+
+In this example we create an Elekta Icon geometry and use it to create some
+artificial data and reconstruct it using backprojection.
```
kohr-h: filtered

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -250,8 +251,15 @@ def parker_weighting(ray_trafo, q=0.25):
     elif ndim == 3:
         # Find projection of axis on detector
         rot_dir = _rotation_direction_in_detector(ray_trafo.geometry)
-        dx = (rot_dir[0] * ray_trafo.range.meshgrid[1] +
-              rot_dir[1] * ray_trafo.range.meshgrid[2])
+        # If axis is axis oriented, save some memory and time by using
```
kohr-h: aligned to a coordinate axis?

odl/tomo/backends/astra_cuda.py
kohr-h: Yeah, kind of an obvious change. Why should the wrapper whose purpose is caching ask again if it should cache? Basically it was asking if it was allowed to stay alive after usage, weird :-)
Nice catch anyway.

odl/contrib/tomo/elekta.py
adler-j: There is no good permalink for this so I thought it better to just have people Google it. Or is a nonpermalink ok?

odl/contrib/tomo/elekta.py
adler-j: Partial scan is the correct term, will fix

odl/contrib/tomo/elekta.py
adler-j: That would be a global change

PR #1040, opened 2017-06-21
BUG: First solution to misdefined KL distance

MatthiasJE:
This commit contains a few changes:
- change name of variable from "prior" to "target". I am not sure if target is the best way to put it, but I find prior certainly misleading. In the context of PET imaging, it should be called "data" but this may be different for other applications and usages.
- "prior"/"target" should be non-optional. It does not make much sense to have it optional. If the "target" is constant 1, then this becomes x - log(x) which may be interesting in itself but not what one would be looking for under the term "Kullback Leibler".
- The actual important change in here is the redefinition of the KL divergence to include the definition 0 log(0) := 0 and NOT 0 log(0) := infty. This has implications also for the convex conjugate. The proximal operators don't need to be changed interestingly.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -1003,7 +1003,7 @@ class KullbackLeibler(Functional):
 
     Notes
     -----
-    The functional :math:`F` with prior :math:`g>0` is given by:
+    The functional :math:`F` with target :math:`g>0` is given by:
 
     .. math::
         F(x)
@@ -1023,7 +1023,7 @@ class KullbackLeibler(Functional):
     `KullbackLeiblerCrossEntropy`. The KL cross entropy is the one
     diescribed in `this Wikipedia article
     <https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence>`_, and
-    the functional :math:`F` is obtained by switching place of the prior and
+    the functional :math:`F` is obtained by switching place of the target and
     the varialbe in the KL cross entropy functional.
 
     For a theoretical exposition, see `Csiszar1991`_.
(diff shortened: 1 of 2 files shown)


MatthiasJE:
I found another "bug" in the KL distance. Unfortunately, I cannot produce a minimal example as this error only occurs for me on large data sets (variables of size 1 GB +)

The following two lines do NOT produce the same result on unweighted spaces:

```
Y1
Out[7]: uniform_discr([0.0, 0.0], [68516.0, 4084.0], (68516, 4084), dtype='float32', weighting='none')
```

```
            tmp = ((x - self.prior +
                   scipy.special.xlogy(self.prior, self.prior / x))
                   .inner(self.domain.one()))

```
```
            tmp = (x - self.prior +
                   scipy.special.xlogy(self.prior, self.prior / x)).ufuncs.sum()
```

One results in 

```
y2
Out[4]: 332767968.0
```
and the other in 
```
y1
Out[6]: 423516992.0

```

Any ideas?

adler-j:
The precision of float32 is about 7 digits, so if you compute the sum of more than 10^7 digits (you have 2.7 * 10^8) then you *will* run into numerical issues. I'm not even sure which one of those is correct.

You can try it with float64 to get the correct value I guess, and we can see if we can make the implementation more numerically stable.

MatthiasJE:
Of course. I did not expect to get the same result but these are off by 25 % !!! The underlying has to be float32 for me as I perform CUDA computations on them (even if not using odlcuda). However, the objective value computation could be performed using float64.

adler-j:
> Of course. I did not expect to get the same result but these are off by 25 %. The underlying has to be float32 for me as I perform CUDA computations on them (even if not using odlcuda). However, the objective value computation could be performed using float64.

Well floating point errors really tend to sneak on you. In general, the errors are proportional to the number of elements you sum, so you may be having 1% error with 10^7 elements and then you have 25% error with 2.5 * 10^8 and 100% with 10^9.

Could you please try this:

> You can try it with float64 to get the correct value I guess, and we can see if we can make the implementation more numerically stable.

MatthiasJE:
Here you are:

```
            x = Y1.one()
            d = Y1.element(data)
            b = Y1.element(background)
            % time tmp1 = (x + b - d + scipy.special.xlogy(d, d / (x + b))).inner(Y1.one())
            % time tmp2 = (x + b - d + scipy.special.xlogy(d, d / (x + b))).ufuncs.sum()

            Y2 = Y1.astype('float64')
            x = Y2.element(x)
            d = Y2.element(d)
            b = Y2.element(b)
            % time tmp3 = (x + b - d + scipy.special.xlogy(d, d / (x + b))).inner(Y2.one())
            % time tmp4 = (x + b - d + scipy.special.xlogy(d, d / (x + b))).ufuncs.sum()
```

```
CPU times: user 12.7 s, sys: 668 ms, total: 13.4 s
Wall time: 13.4 s
CPU times: user 12.3 s, sys: 512 ms, total: 12.8 s
Wall time: 12.9 s
CPU times: user 13.1 s, sys: 1.22 s, total: 14.3 s
Wall time: 14.3 s
CPU times: user 12.4 s, sys: 1.07 s, total: 13.5 s
Wall time: 13.5 s
```

```
tmp1 = 332767968.0
tmp2 = 423501120.0
tmp3 = 423502526.60952723
tmp4 = 423502526.6099115
(tmp1-tmp4)/tmp4 = -0.21424797470804033
(tmp2-tmp4)/tmp4 = -3.32137312795075e-06
```

PR #1042, opened 2017-06-26
BUG: fix 0 log(0) bug in KullbackLeibler

MatthiasJE:
I deleted the branch of the previous PR #1040 which permanently shut that one down :(

Therefore, here the second PR. This time I did not change the name of "prior" to "target" as a good variable name depends on the application. Sometimes "prior" might be better. Sometimes "target" or "data". Similar for the default values.

This PR changes the definition of the KL divergence to include the definition 0 log(0) := 0 and NOT 0 log(0) := infty. This has implications also for the convex conjugate. The proximal operators don't need to be changed interestingly.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -1003,18 +1003,19 @@ class KullbackLeibler(Functional):
 
     Notes
     -----
-    The functional :math:`F` with prior :math:`g>0` is given by:
+    The functional :math:`F` with prior :math:`g>=0` is given by:
 
     .. math::
         F(x)
         =
         \\begin{cases}
-            \\sum_{i} \left( x_i - g_i + g_i \log \left( \\frac{g_i}{ x_i }
+            \\sum_{i} \left( x_i - g_i + g_i \log \left( \\frac{g_i}{x_i}
             \\right) \\right) & \\text{if } x_i > 0 \\forall i
             \\\\
             +\\infty & \\text{else.}
         \\end{cases}
 
+    Note that we use the common definition 0 log(0) := 0.
     KL based objectives are common in MLEM optimization problems and are often
     used as data-matching term when data noise governed by a multivariate
     Poisson probability distribution is significant.
@@ -1046,8 +1047,23 @@ class KullbackLeibler(Functional):
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
         prior : ``space`` `element-like`, optional
-            Data term, positive.
+            Depending on the context, the prior, target or data
+            distribution. It is assumed to be nonnegative.
             Default: if None it is take as the one-element.
+
+        Examples
+        --------
+        >>> space = odl.rn(3)
+        >>> prior = 3*space.one()
+        >>> func = KullbackLeibler(space, prior=prior)
+        >>> func(prior)
+        0.0
+
+        >>> prior = space.zero()
+        >>> func = KullbackLeibler(space, prior=prior)
+        >>> x = space.one()
+        >>> func(x)
+        3.0
         """
         super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
 
@@ -1072,7 +1088,8 @@ class KullbackLeibler(Functional):
         if self.prior is None:
             tmp = ((x - 1 - np.log(x)).inner(self.domain.one()))
         else:
-            tmp = ((x - self.prior + self.prior * np.log(self.prior / x))
+            tmp = ((x - self.prior +
+                   scipy.special.xlogy(self.prior, self.prior / x))
                    .inner(self.domain.one()))
         if np.isnan(tmp):
             # In this case, some element was less than or equal to zero
@@ -1082,7 +1099,11 @@ class KullbackLeibler(Functional):
 
     @property
     def gradient(self):
-        """Gradient operator of the functional.
+        """The gradient of `KullbackLeibler` with ``prior`` :math:`g` is given as
+
+        .. math::
+
+        \\nabla F(x) = 1 - \frac{g}{x}.
 
         The gradient is not defined in points where one or more components
         are non-positive.
@@ -1141,7 +1162,7 @@ class KullbackLeiblerConvexConj(Functional):
 
     Notes
     -----
-    The functional :math:`F^*` with prior :math:`g>0` is given by:
+    The functional :math:`F^*` with prior :math:`g>=0` is given by:
 
     .. math::
         F^*(x)
@@ -1166,7 +1187,8 @@ class KullbackLeiblerConvexConj(Functional):
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
         prior : ``space`` `element-like`, optional
-            Data term, positive.
+            Depending on the context, the prior, target or data
+            distribution. It is assumed to be nonnegative.
             Default: if None it is take as the one-element.
         """
         super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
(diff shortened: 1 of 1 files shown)


odl/solvers/functional/default_functionals.py
```diff
@@ -1046,8 +1047,23 @@ def __init__(self, space, prior=None):
...
         prior : ``space`` `element-like`, optional
-            Data term, positive.
+            Depending on the context, the prior, target or data
+            distribution. It is assumed to be nonnegative.
             Default: if None it is take as the one-element.
+
+        Examples
+        --------
+        >>> space = odl.rn(3)
+        >>> prior = 3*space.one()
```
adler-j: 3 * space.one()

odl/solvers/functional/default_functionals.py
```diff
@@ -1082,7 +1099,11 @@ def _call(self, x):
 
     @property
     def gradient(self):
-        """Gradient operator of the functional.
+        """The gradient of `KullbackLeibler` with ``prior`` :math:`g` is given as
+
+        .. math::
+
```
adler-j: Not properly typeset, do

```
.. math::
    \\nabla F(x) = 1 - \\frac{g}{x}.
```

odl/solvers/functional/default_functionals.py
```diff
@@ -1046,8 +1047,23 @@ def __init__(self, space, prior=None):
...
-            Data term, positive.
+            Depending on the context, the prior, target or data
+            distribution. It is assumed to be nonnegative.
             Default: if None it is take as the one-element.
+
+        Examples
+        --------
+        >>> space = odl.rn(3)
+        >>> prior = 3*space.one()
+        >>> func = KullbackLeibler(space, prior=prior)
```
adler-j: Do `func = odl.solvers.KullbackLeibler(space, prior=prior)`, changes nothing here but makes the example more instructive and copy-pastable for others

odl/solvers/functional/default_functionals.py
```diff
@@ -1046,8 +1047,23 @@ def __init__(self, space, prior=None):
         space : `DiscreteLp` or `FnBase`
             Domain of the functional.
         prior : ``space`` `element-like`, optional
-            Data term, positive.
+            Depending on the context, the prior, target or data
+            distribution. It is assumed to be nonnegative.
             Default: if None it is take as the one-element.
+
+        Examples
+        --------
```
adler-j: If you could, adding a short sentence explaining what happens usually lightens up the mood here

ozanoktem:
# Regularizing the entropy at zero
Sometimes one encounters the case when :math:`x_i=0` and this can in fact be handled. The entropy term :math:`x \log x` is defined and continuous up to :math:`x=0` but its derivative is singular, so one needs to introduce a regularizing step. Below is a suggestion for how to do this. 

## Problem
Let us first consider the function :math:`S(x):= - x\log x` defined for all :math:`x \geq 0`. This function is positive on :math:`[0,1]` and continuous up to :math:`x=0`. However, its derivative is given as :math:`S'(x)=-1-\log x` which tends to :math:`\infty` as :math:`x \to 0`. Hence, one might experience numerical difficulties in situations where the derivative of :math:`S` is needed in a neighborhood of :math:`x=0`. This would especially be the case when one needs to minimize a function over :math:`\geq 0` that includes :math:`S` as a component.

## Proposed solution
One very crude way out is simply to choose a threshold :math:`\epsilon >0` and then simply modify :math:`S` with 
.. math::
  S_{\epsilon}(x) := 
       \begin{cases}
          S(x)                        & x>\epsilon, \\\\
          -\epsilon \log\epsilon      & x\leq \epsilon.
       \end{cases}
       
This new function (which in fact is defined for all :math:`x:math:`) has a derivative that is given as 
.. math::
  S_{\epsilon}'(x) = 
       \begin{cases}
          S'(x)  & x>\epsilon, \\\\
          0      &  x\leq \epsilon.
       \end{cases}
    
Observe that the problem with this approach is that the derivative :math:`S_{\epsilon}'(x)` makes a jump at :math:`x=\epsilon` that equals :math:`-1-\log \epsilon`. Since :math:`\epsilon` is typically a small number, this jump can be rather large. Also, this jump is rather sensitive for where the threshold :math:`\epsilon` is chosen. Hence, there is still a fair amount of instability in this way of handling the discontinuity of :math:`S'(x)` at :math:`x=0`.

The next choice is to replace the part of the function :math:`S(x)` with a simple line. In fact, choose :math:`\epsilon>0` as a threshold and now we define 
.. math::
  S_{\epsilon}(x) := 
       \begin{cases}
          S(x)                        & x>\epsilon, \\\\
          -x \log\epsilon             & x\leq \epsilon.
       \end{cases}

This new function (which in also defined for all :math:`x`) has a derivative that is given as  
.. math::
  S_{\epsilon}'(x) = 
       \begin{cases}
          S'(x)               & x>\epsilon, \\\\
          -\log \epsilon      & x\leq \epsilon.
       \end{cases}

Now we see that the derivative :math:`S_{\epsilon}'(x)` still makes a jump at :math:`x=\epsilon`, but this time the jump equals :math:`-1`, which is independent of :math:`\epsilon`. In fact, when :math:`\epsilon` is small, :math:`S'(\epsilon)` is quite large and the jump of :math:`-1` is relatively small. Hence, this latter way of replacing :math:`S` with a function that is differentiable at :math:`x=0` is much more robust.

One can of course go on and fit a spline of arbitrary smoothness that is flat at :math:`x=0` and the resulting function is smooth up to a rather high order. However, the above proposed scheme is very simple to implement and definitely worth trying.

adler-j:
Interesting writeup @ozanoktem, but I would stay clear of that issue in this PR, perhaps you could make a new issue w.r.t. this and we could look into fixing this in the future?

kohr-h:
@ozanoktem Do we know convex conjugate and prox for those guys?

ozanoktem:
Have not thought about the convex conjugate and prox for the regularized entropy. On the other hand, the whole idea with the regularization is to have differentiable expression that can be used in methods that rely on derivatives, so the need for a prox is perhaps less urgent.

kohr-h:
Of course, usually we use KL as a differentiable functional and are interested in the gradient. Still, sometimes one just happens to use a solver that doesn't make use of the gradient but just treats KL as a convex functional (e.g. Chambolle-Pock or our Douglas-Rachford solver).

odl/solvers/functional/default_functionals.py
MatthiasJE: Agreed, should be done for many other ODL examples, too!

odl/solvers/functional/default_functionals.py
MatthiasJE: I am surprised you say this. I was already cursing other ODL examples for not doing so :)

adler-j: Any improvements are welcome w.r.t. this :D Its just that I've realized this over time and now started doing it, never got around to fixing it globally (for obvious reasons)

odl/solvers/functional/default_functionals.py
kohr-h: My bad :-)

MatthiasJE:
I did something similar for the KL divergence to make it globally Lipschitz differentiable in the second argument, see https://arxiv.org/abs/1706.04957. The convex conjugate and prox operators are still closed form (though calculations get a little more nasty) so I expect it to be no problem here, too.

@ozanoktem: Do you have a use case for these smooth approximations?

ozanoktem:
> @ozanoktem: Do you have a use case for these smooth approximations?

@MatthiasJE , many years ago we used the KL as a regularizer (not as data discrepancy) when performing electron tomography. Then, it was quite important to use the regularized version. The source code for that is unfortunately not available, it was part of a company software portfolio.

odl/solvers/functional/default_functionals.py
```diff
@@ -1046,8 +1047,29 @@ def __init__(self, space, prior=None):
...
+
```
adler-j: Remove extra blankline

odl/solvers/functional/default_functionals.py
```diff
@@ -1046,8 +1047,29 @@ def __init__(self, space, prior=None):
...
+        Test that KullbackLeibler(x,x) = 0
```
adler-j: KullbackLeibler(x, x)

PR #1043, opened 2017-06-27
Optimizations to the case of small data

adler-j:
Even more boring performance hunting. Now made a particular (internal) script run about 30-40% faster by doing this. Additionally, improves the runtime of the test suite by 5% when compared to master.

Also fixed a bug in running pytest.

### DIFF
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -226,17 +226,22 @@ class RectGrid(Set):
     @property
     def ndim(self):
         """Number of dimensions of the grid."""
-        return len(self.coord_vectors)
+        try:
+            return self.__ndim
+        except AttributeError:
+            ndim = len(self.coord_vectors)
+            self.__ndim = ndim
+            return ndim
 
     @property
     def shape(self):
         """Number of grid points per axis."""
         try:
-            shape = self.__shape
+            return self.__shape
         except AttributeError:
             shape = tuple(len(vec) for vec in self.coord_vectors)
             self.__shape = shape
-        return shape
+            return shape
 
     @property
     def size(self):
--- a/odl/discr/grid.py
+++ b/odl/discr/grid.py
@@ -231,7 +231,12 @@ class RectGrid(Set):
     @property
     def shape(self):
         """Number of grid points per axis."""
-        return tuple(len(vec) for vec in self.coord_vectors)
+        try:
+            shape = self.__shape
+        except AttributeError:
+            shape = tuple(len(vec) for vec in self.coord_vectors)
+            self.__shape = shape
+        return shape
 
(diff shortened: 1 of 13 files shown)


odl/discr/grid.py
```diff
@@ -226,12 +226,22 @@ def coord_vectors(self):
     @property
     def ndim(self):
         """Number of dimensions of the grid."""
-        return len(self.coord_vectors)
+        try:
+            return self.__ndim
+        except AttributeError:
+            ndim = len(self.coord_vectors)
+            self.__ndim = ndim
```
kohr-h: As I had written somewhere else, I'm not very fond of the `try: self.bla except AttributeError: self.bla = ...` pattern. Setting the attributes in question to `None` in the initializer together with a comment makes the intent much clearer and makes cached attributes easier to find.

odl/discr/lp_discr.py
```diff
@@ -273,6 +273,22 @@ def tangent_bundle(self):
         else:
             return ProductSpace(self, self.ndim)
 
+    @property
+    def is_uniform_weighting(self):
+        """If the weighting of the space is the same for all pixels."""
```
kohr-h: pixels -> points

odl/discr/lp_discr.py
```diff
@@ -273,6 +273,22 @@ def tangent_bundle(self):
         else:
             return ProductSpace(self, self.ndim)
 
+    @property
+    def is_uniform_weighting(self):
```
kohr-h: `is_uniformly_weighted`

odl/discr/grid.py
kohr-h: I just stumbled over this while compiling the release notes.

What you wrote @adler-j may have been true for older versions of Python. Since 3.6, however, dictionaries have an [internal version](https://www.python.org/dev/peps/pep-0509/) (in the C implementation) that lets the interpreter see if the dictionary has changed since the last access. If not, a cached version is used, several times faster than the full lookup.

The effect of this is that classes that define all of their instance attributes in `__init__` get rewarded by faster lookups later on, since the instance `__dict__` stays constant.

Anyway, we should revise this pattern and do some benchmarking to get the numbers.

adler-j: I agree that we should try.

With that said, wouldnt the cache update after the first access anyway, evening out in the long term?

kohr-h: Yeah, I'm not entirely sure what the implications are. I just heard one of the core Python devs mention this detail as a reason to set up all attributes during `__init__`, but you're right, it might just play a role for the first lookup. Anyway, should be easy to test.

PR #1045, opened 2017-06-28
BUG: Fix clipped titles in show, closes #1044

### LINKED ISSUE #1044: titles not displaying correctly in the standard .show() function
Using the standard .show(title="") fucntion results in a figure where the title is cut off (see the figure below).
Note that this problem does not seem to occur when displaying the figure inline.

The following code illustrates the problem
```
import odl

space = odl.uniform_discr(min_pt=0, max_pt=20, shape=20, dtype='float32')

ramp = space.element(range(20))
ramp.show(title='Figure of a ramp with cut off title', method='plot')
```
This will result in this figure:

figure_of_a_ramp_with_cut_off_title

adler-j:
Trivial change to fix clipping of title, now sets tight layout **after** adding the title.

### DIFF
--- a/odl/util/graphics.py
+++ b/odl/util/graphics.py
@@ -416,14 +416,14 @@ def show_discrete_data(values, grid, title=None, method='',
                 csub.colorbar.draw_all()
 
     # Fixes overlapping stuff at the expense of potentially squashed subplots
-    if not update_in_place:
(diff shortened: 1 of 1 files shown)


adler-j:
Forced this in myself since it was so trivial.

PR #1046, opened 2017-06-28
ENH: Remove unncessary warning when calling matplotlib.pause

adler-j:
Cherry picked this so people no longer have to be annoyed. Please accept as is to avoid problems in TF branch :)


adler-j:
Fixing that, will merge after CI.

PR #1047, opened 2017-06-29
MAINT: Change references to be local in doc

adler-j:
Now fixed all of this for the code doc. I guess we maybe want to kill `refs.rst` totally and move all references into the specific files?

### DIFF
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -136,15 +136,19 @@ class Functional(Operator):
 
         The concept is also known as the Legendre transformation.
 
-        References
-        ----------
-        Wikipedia article on `Convex conjugate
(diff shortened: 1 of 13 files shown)


odl/solvers/functional/functional.py
```diff
@@ -136,15 +136,21 @@ def convex_conj(self):
...
         Wikipedia article on `Convex conjugate
         <https://en.wikipedia.org/wiki/Convex_conjugate>`_.
 
         Wikipedia article on `Legendre transformation
         <https://en.wikipedia.org/wiki/Legendre_transformation>`_.
 
         For literature references see, e.g., [Lue1969]_, [Roc1970]_.
+
+        References
+        ----------
```
kohr-h: This heading shouldn't have been moved. Also move the sentence "For literature ..." out of References.

PR #1049, opened 2017-06-29
DOC: Add coverage badge

adler-j:
I feel that we should add this badge, if nothing else because it conveys that this is a serious library with good coverage. Do we agree?


kohr-h:
Yeah, why not

PR #1050, opened 2017-06-29
MAINT: Update coveragerc

adler-j:
This file has obviously become way outdated

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,16 +9,19 @@ omit =
     */__init__.py
     odl/diagnostics/*
     odl/util/*
-    
-    # Omit until coveralls supports cuda.
-    odl/space/cu_ntuples.py
-    
+
     # Omit until coveralls supports pywt.
     odl/trafos/wavelet.py
 
     # Omit until coveralls supports astra.
-    odl/tomo/operators/xray_trafo.py
+    odl/tomo/operators/ray_trafo.py
     odl/tomo/backends/*
+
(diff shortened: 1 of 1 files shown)


kohr-h:
Hm, did you check where this decrease of coverage comes from? Seems like some stuff is omitted that shouldn't.

adler-j:
I agree, something definitely went wrong here. I'll look through it

kohr-h:
Status?

adler-j:
I'll merge this for now, but it seems our coverage is lower than we thought :/

kohr-h:
Please remove the "merge master into branch" commit.

.coveragerc
```diff
@@ -10,16 +10,19 @@ omit =
     odl/diagnostics/*
     odl/util/*
 
-    # Omit until coveralls supports cuda.
-    odl/space/cu_ntuples.py
-
     # Omit until coveralls supports pywt.
     odl/trafos/wavelet.py
```
kohr-h: We install pywavelets on Travis now, so this line can be removed.

.coveragerc
```diff
@@ -46,6 +49,13 @@ exclude_lines =
     # Decorators
```
kohr-h: Comment here: `return NotImplemented` should go off the list, since it's part of actual functionality.

kohr-h: Instead, we may think about adding the `RuntimeError` stuff that's just defensive coding.

PR #1056, opened 2017-07-04
TST: Improve angles test for divergent beam geometries, closes #1053

### LINKED ISSUE #1053: Use more robust angles unit test for ray transforms
The [current unit test for the angle definition in geometries](https://github.com/odlgroup/odl/blob/b2a7a616dc50e83ec3e3b3f2bd84edd7df1ca3fc/odl/test/tomo/operators/ray_trafo_test.py#L248-L254) is rather fragile since it takes a maximum where it could probably better take a center of mass or something.

adler-j:
Fixed scaling by magnification factor, test is now quite accurate for all geometries.


adler-j:
Merge after CI

PR #1059, opened 2017-07-04
Vectorize geometries

kohr-h:
~~Incomplete~~ PR for vectorization of geometry methods.

### TODO:
- [x] Propagate `check_bounds` to detectors
- [x] Update doc for detectors
- [x] Improve `repr` of detectors
- [x] Decide on what to do with nonsense parameters (e.g. `normalized` when only one choice is valid, but the parameter is there for API consistency)
      **Solution:** I removed `normalized` from `ParallelGeometry.det_to_src`, can't seem to remember what other cases I had in mind.
- [x] Replace `super()` calls
      **Solution:** I did like suggested n #1061 
- [x] Check if the custom `all_contained()` is really needed, seems fishy
      **Solution:** Turned out a simple transpose did the job
- [x] Fix methods depending on two variables (as in 56393e6)
- [x] ~~Add unit tests~~ Have already a bunch of doctests, and ASTRA vec geometry generation already uses vectorization, should be good enough?

### DIFF
--- a/odl/tomo/backends/astra_setup.py
+++ b/odl/tomo/backends/astra_setup.py
@@ -245,27 +245,25 @@ def astra_conebeam_3d_geom_to_vec(geometry):
     angles = geometry.angles
     vectors = np.zeros((angles.size, 12))
 
-    for ang_idx, angle in enumerate(angles):
-        # Source position
-        vectors[ang_idx, 0:3] = geometry.src_position(angle)
+    # Source position
+    vectors[:, 0:3] = geometry.src_position(angles)
 
-        # Center of detector in 3D space
-        mid_pt = geometry.det_params.mid_pt
-        vectors[ang_idx, 3:6] = geometry.det_point_position(angle, mid_pt)
+    # Center of detector in 3D space
+    mid_pt = geometry.det_params.mid_pt
+    vectors[:, 3:6] = geometry.det_point_position(angles, mid_pt)
 
-        # Vectors from detector pixel (0, 0) to (1, 0) and (0, 0) to (0, 1)
-        det_axes = geometry.det_axes(angle)
-        px_sizes = geometry.det_partition.cell_sides
-
-        # Swap detector axes to have better memory layout in  projection data.
-        # ASTRA produces `(v, theta, u)` layout, and to map to ODL layout
-        # `(theta, u, v)` a complete roll must be performed, which is the
-        # worst case (compeltely discontiguous).
-        # Instead we swap `u` and `v`, resulting in the effective ASTRA result
-        # `(u, theta, v)`. Here we only need to swap axes 0 and 1, which
-        # keeps at least contiguous blocks in `v`.
-        vectors[ang_idx, 9:12] = det_axes[0] * px_sizes[0]
-        vectors[ang_idx, 6:9] = det_axes[1] * px_sizes[1]
+    # Vectors from detector pixel (0, 0) to (1, 0) and (0, 0) to (0, 1)
+    det_axes = geometry.det_axes(angles)
+    px_sizes = geometry.det_partition.cell_sides
+    # Swap detector axes to have better memory layout in  projection data.
+    # ASTRA produces `(v, theta, u)` layout, and to map to ODL layout
+    # `(theta, u, v)` a complete roll must be performed, which is the
+    # worst case (compeltely discontiguous).
+    # Instead we swap `u` and `v`, resulting in the effective ASTRA result
+    # `(u, theta, v)`. Here we only need to swap axes 0 and 1, which
+    # keeps at least contiguous blocks in `v`.
+    vectors[:, 9:12] = det_axes[0] * px_sizes[0]
+    vectors[:, 6:9] = det_axes[1] * px_sizes[1]
 
     # ASTRA has (z, y, x) axis convention, in contrast to (x, y, z) in ODL,
     # so we need to adapt to this by changing the order.
@@ -317,19 +315,19 @@ def astra_conebeam_2d_geom_to_vec(geometry):
     angles = geometry.angles
     vectors = np.zeros((angles.size, 6))
 
-    for ang_idx, angle in enumerate(angles):
-        # Source position
-        vectors[ang_idx, 0:2] = rot_minus_90.dot(geometry.src_position(angle))
+    # Source position
+    src_pos = geometry.src_position(angles)
+    vectors[:, 0:2] = rot_minus_90.dot(src_pos.T).T  # dot along 2nd axis
 
-        # Center of detector
-        mid_pt = geometry.det_params.mid_pt
-        vectors[ang_idx, 2:4] = rot_minus_90.dot(
-            geometry.det_point_position(angle, mid_pt))
+    # Center of detector
+    mid_pt = geometry.det_params.mid_pt
+    centers = geometry.det_point_position(angles, mid_pt)
+    vectors[:, 2:4] = rot_minus_90.dot(centers.T).T
 
-        # Vector from detector pixel 0 to 1
-        det_axis = rot_minus_90.dot(geometry.det_axis(angle))
-        px_size = geometry.det_partition.cell_sides[0]
-        vectors[ang_idx, 4:6] = det_axis * px_size
+    # Vector from detector pixel 0 to 1
+    det_axis = rot_minus_90.dot(geometry.det_axis(angles).T).T
+    px_size = geometry.det_partition.cell_sides[0]
+    vectors[:, 4:6] = det_axis * px_size
 
     return vectors
 
@@ -371,28 +369,26 @@ def astra_parallel_3d_geom_to_vec(geometry):
     angles = geometry.angles
     vectors = np.zeros((angles.shape[0], 12))
 
-    for ang_idx, angle in enumerate(angles):
-        mid_pt = geometry.det_params.mid_pt
+    mid_pt = geometry.det_params.mid_pt
 
-        # Ray direction = -(detector-to-source normal vector)
-        vectors[ang_idx, 0:3] = -geometry.det_to_src(angle, mid_pt)
+    # Ray direction = -(detector-to-source normal vector)
+    vectors[:, 0:3] = -geometry.det_to_src(angles, mid_pt)
 
-        # Center of the detector in 3D space
-        vectors[ang_idx, 3:6] = geometry.det_point_position(angle, mid_pt)
+    # Center of the detector in 3D space
+    vectors[:, 3:6] = geometry.det_point_position(angles, mid_pt)
 
-        # Vectors from detector pixel (0, 0) to (1, 0) and (0, 0) to (0, 1)
-        det_axes = geometry.det_axes(angle)
-        px_sizes = geometry.det_partition.cell_sides
-
-        # Swap detector axes to have better memory layout in  projection data.
-        # ASTRA produces `(v, theta, u)` layout, and to map to ODL layout
-        # `(theta, u, v)` a complete roll must be performed, which is the
-        # worst case (compeltely discontiguous).
-        # Instead we swap `u` and `v`, resulting in the effective ASTRA result
-        # `(u, theta, v)`. Here we only need to swap axes 0 and 1, which
-        # keeps at least contiguous blocks in `v`.
-        vectors[ang_idx, 9:12] = det_axes[0] * px_sizes[0]
-        vectors[ang_idx, 6:9] = det_axes[1] * px_sizes[1]
+    # Vectors from detector pixel (0, 0) to (1, 0) and (0, 0) to (0, 1)
+    det_axes = geometry.det_axes(angles)
+    px_sizes = geometry.det_partition.cell_sides
+    # Swap detector axes to have better memory layout in  projection data.
+    # ASTRA produces `(v, theta, u)` layout, and to map to ODL layout
+    # `(theta, u, v)` a complete roll must be performed, which is the
+    # worst case (compeltely discontiguous).
+    # Instead we swap `u` and `v`, resulting in the effective ASTRA result
+    # `(u, theta, v)`. Here we only need to swap axes 0 and 1, which
+    # keeps at least contiguous blocks in `v`.
+    vectors[:, 9:12] = det_axes[0] * px_sizes[0]
+    vectors[:, 6:9] = det_axes[1] * px_sizes[1]
 
     # ASTRA has (z, y, x) axis convention, in contrast to (x, y, z) in ODL,
     # so we need to adapt to this by changing the order.
--- a/odl/tomo/backends/astra_setup.py
+++ b/odl/tomo/backends/astra_setup.py
@@ -322,7 +322,9 @@ def astra_conebeam_2d_geom_to_vec(geometry):
 
     # Center of detector
     mid_pt = geometry.det_params.mid_pt
-    centers = geometry.det_point_position(angles, mid_pt)
+    # Need to cast `mid_pt` to float since otherwise the empty axis is
+    # not removed
+    centers = geometry.det_point_position(angles, float(mid_pt))
     vectors[:, 2:4] = rot_minus_90.dot(centers.T).T
 
     # Vector from detector pixel 0 to 1
--- a/odl/tomo/backends/astra_setup.py
+++ b/odl/tomo/backends/astra_setup.py
@@ -254,7 +254,8 @@ def astra_conebeam_3d_geom_to_vec(geometry):
     vectors[:, 3:6] = geometry.det_point_position(angles, mid_pt)
 
     # Vectors from detector pixel (0, 0) to (1, 0) and (0, 0) to (0, 1)
-    det_axes = geometry.det_axes(angles)
+    # `det_axes` gives shape (N, 2, 3), swap to get (2, N, 3)
+    det_axes = np.moveaxis(geometry.det_axes(angles), -2, 0)
     px_sizes = geometry.det_partition.cell_sides
     # Swap detector axes to have better memory layout in  projection data.
     # ASTRA produces `(v, theta, u)` layout, and to map to ODL layout
@@ -381,7 +382,8 @@ def astra_parallel_3d_geom_to_vec(geometry):
     vectors[:, 3:6] = geometry.det_point_position(angles, mid_pt)
 
     # Vectors from detector pixel (0, 0) to (1, 0) and (0, 0) to (0, 1)
-    det_axes = geometry.det_axes(angles)
+    # `det_axes` gives shape (N, 2, 3), swap to get (2, N, 3)
+    det_axes = np.moveaxis(geometry.det_axes(angles), -2, 0)
     px_sizes = geometry.det_partition.cell_sides
     # Swap detector axes to have better memory layout in  projection data.
     # ASTRA produces `(v, theta, u)` layout, and to map to ODL layout
--- a/odl/tomo/backends/astra_setup.py
+++ b/odl/tomo/backends/astra_setup.py
@@ -50,7 +50,8 @@ import numpy as np
 
 from odl.discr import DiscreteLp, DiscreteLpElement
 from odl.tomo.geometry import (
-    Geometry, DivergentBeamGeometry, ParallelBeamGeometry, FlatDetector)
+    Geometry, DivergentBeamGeometry, ParallelBeamGeometry,
+    Flat1dDetector, Flat2dDetector)
 from odl.tomo.util.utility import euler_matrix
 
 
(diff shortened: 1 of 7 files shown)


kohr-h:
Ready for review. As often before when working on geometries, there's a lot of new documentation, including doctests, which accounts for the biggest part of the PR. Documentation is way improved now IMO.

odl/tomo/geometry/detector.py
```diff
@@ -125,356 +180,628 @@ def surface_measure(self, param):
...
+        The direction is chosen such that the circle is traversed clockwise
+        for growing angle parameter.
+        """
+        return perpendicular_vector(self.center_dir)
 
     def surface(self, param):
-        """Parametrization of the detector reference surface.
+        """Return the detector surface point corresponding to ``param``.
+
+        The surface point lies on a circle around ``radius * center_dir``
```
kohr-h: \`circle\`

odl/tomo/geometry/geometry.py
```diff
@@ -251,64 +375,139 @@ class DivergentBeamGeometry(Geometry):
...
         Returns
         -------
-        vec : `numpy.ndarray`, shape (`ndim`,)
-            (Unit) vector pointing from the detector to the source.
+        det_to_src : `numpy.ndarray`
+            Vector(s) pointing from a detector point to the source (at
+            infinity).
+            The shape of the returned array is as follows:
+
+            - ``mparam`` and ``dparam`` single: ``(ndim,)``
```
kohr-h: angle

odl/tomo/geometry/parallel.py
```diff
@@ -99,14 +98,17 @@ def det_refpoint(self, angle):
 
         Parameters
         ----------
-        angle : float
-            Parameter describing the detector rotation, must be
-            contained in `motion_params`.
+        angle : float or `array-like`
+            Angle(s) in radians describing the counter-clockwise
+            rotation of the detector.
```
kohr-h: Needs more general formulation that includes multiple angles.

odl/tomo/geometry/parallel.py
```diff
@@ -713,31 +843,80 @@ def det_axes_init(self):
...
+            If ``angles`` is a pair (or triplet) of Euler angles, a single
+            matrix is returned, otherwise a stack of matrices along axis 0.
         """
-        if angles not in self.motion_params:
+        squeeze_out = (np.shape(angles) == (self.motion_params.ndim,))
+        angles = np.array(angles, dtype=float, copy=False, ndmin=2)
+
+        # Need custom check here until #861 is in because arrays are
+        # assumed to be flat in the "point enumeration" axis
+        # TODO: replace when #861 is in
```
kohr-h: Missed this old check, needs to be replaced.

odl/tomo/geometry/geometry.py
```diff
@@ -45,18 +45,30 @@ class Geometry(object):
...
+            The detector of this geometry.
+        translation : `array-like`, optional
+            Global translation of the geometry. This is added last in any
+            method that computes an absolute vector, e.g., `det_refpoint`.
+            Default: zero vector of length ``ndim``
+
+        Other Parameters
+        ----------------
+        check_bounds : bool, optional
+            If ``True``, methods perform sanity checks on provided input
```
adler-j: Someone ambiguous what parameters are refered to here, I'd read it as the parameters to this method.

odl/tomo/geometry/geometry.py
```diff
@@ -85,6 +98,11 @@ def __init__(self, ndim, motion_part, detector, translation=None):
         # Cache geometry-related objects for backends that require computation
         self.__implementation_cache = {}
 
+        # Make sure there are no leftover kwargs
```
adler-j: This really should be added to more places, i've had errors N times because of stuff like this where I think i gave an argument but it was simply ignored.

odl/tomo/geometry/geometry.py
```diff
@@ -155,83 +173,189 @@ def translation(self):
         """Shift of the origin of this geometry."""
         return self.__translation
 
-    def det_refpoint(self, mpar):
+    @property
+    def check_bounds(self):
+        """Whether to check if method parameters are in the valid range."""
```
adler-j: Also be more clear as to what "methods" are refered to

odl/tomo/geometry/geometry.py
```diff
@@ -155,83 +173,189 @@ def translation(self):
...
+            If ``True``, normalize the resulting vector(s) to unit length.
 
         Returns
         -------
-        vec : `numpy.ndarray`, shape (`ndim`,)
-            (Unit) vector pointing from the detector to the source
+        vec : `numpy.ndarray`
+            (Unit) vector(s) pointing from the detector to the source.
+            If both ``mparam`` and ``dparam`` are single parameters, a single
+            vector is returned, otherwise a stack of vectors.
```
adler-j: Does it support broadcasting, i.e. outer products?

odl/tomo/geometry/geometry.py
```diff
@@ -251,64 +375,139 @@ class DivergentBeamGeometry(Geometry):
...
         Returns
         -------
-        vec : `numpy.ndarray`, shape (`ndim`,)
-            (Unit) vector pointing from the detector to the source.
+        det_to_src : `numpy.ndarray`
+            Vector(s) pointing from a detector point to the source (at
+            infinity).
+            The shape of the returned array is as follows:
+
+            - ``angle`` and ``dparam`` single: ``(ndim,)``
```
adler-j: Not quite sure I like these cases all over the place, feels error prone.

What about instead saying that the returned array always has shape

```
(angle + dparam).shape + (ndim,)
```

that should cover all special cases and also allows a use case that this does not (i.e. a list of angles and dparams, without outer product) and also does not enforce any order. Users would call it by simply appending an axis where they want to use broadcasting.

This can be generalized to all of these, and would make the doc somewhat easier to read.

adler-j: Util this is settled I'll not review in depth stuff related to this.

odl/tomo/geometry/parallel.py
```diff
@@ -713,31 +848,71 @@ def det_axes_init(self):
...
 
     def det_axes(self, angles):
-        """Return the detector axes tuple at ``angle``."""
+        """Return the detector axes tuple at ``angles``.
+
+        Parameters
+        ----------
+        angles : `array-like`
+            Euler angles in radians describing the rotation of the detector.
+            If multiple angle pairs (or triplets) are given, they must be
```
adler-j: The above proposed fix would solve this so users can give carthesian products, as they would likely expect that they could.

odl/tomo/geometry/parallel.py
```diff
@@ -713,31 +848,71 @@ def det_axes_init(self):
...
+            stacked along axis 0.
+
+        Returns
+        -------
+        axes : tuple of `numpy.ndarray`'s
```
adler-j: I think i mentioned this above, but if possible try to make this a large ndarray. Is there a reason we want a tuple?

odl/tomo/geometry/parallel.py
```diff
@@ -1014,9 +1196,9 @@ def frommatrix(cls, apart, dpart, init_matrix):
         # Use the standard constructor with these vectors
         axis, det_pos, det_axis_0, det_axis_1 = transformed_vecs
         if translation.size == 0:
-            kwargs = {}
+            pass
```
adler-j: just make an `if translation.size != 0:`

odl/tomo/util/utility.py
```diff
@@ -82,7 +82,11 @@ def euler_matrix(*angles):
              sth * cps,
              cth]])
 
-    return mat
+    if np.isscalar(angles[0]):
+        return mat.squeeze()
+    else:
+        # Move third axis to the beginning
```
adler-j: This is obvious from the code. But why do we do it?

odl/tomo/util/utility.py
```diff
@@ -525,37 +541,60 @@ def perpendicular_vector(vec):
...
+    squeeze_out = (np.ndim(vec) == 1)
+    vec = np.array(vec, dtype=float, copy=False, ndmin=2)
 
-    if np.all(vec == 0):
+    if np.any(np.all(vec == 0, axis=1)):
         raise ValueError('zero vector')
 
     result = np.zeros(vec.shape)
+    cond = np.any(vec[:, :2] != 0, axis=1)
+    if np.any(cond):
```
adler-j: does this "any" make a difference in any way? I guess performance but I'd not over-optimize this right now.

odl/tomo/util/utility.py
```diff
@@ -525,37 +541,60 @@ def perpendicular_vector(vec):
...
+        result[cond, 0] = -vec[cond, 1]
+        result[cond, 1] = vec[cond, 0]
+    if np.any(~cond):
+        result[~cond, 0] = 1
+    result / np.linalg.norm(result, axis=1, keepdims=True)
```
adler-j: We're not afraid of zero division here, right?

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -29,12 +29,18 @@
...
     import pyfftw
     PYFFTW_AVAILABLE = True
-    if parse_version(pyfftw.__version__) < parse_version('0.10.4'):
-        warnings.warn('PyFFTW < 0.10.4 is known to cause problems with some '
-                      'ODL functionality, see issue #1002.',
-                      RuntimeWarning)
 except ImportError:
     PYFFTW_AVAILABLE = False
+else:
+    import pip
```
adler-j: I don't like adding `pip` as a dependency to ODL, and further you mention this taking some performance.

I'd rather remove this, or as mentioned somewhere else explcitly check against something like

```python
if parse_version(pyfftw.__version__) < parse_version('0.10.3.devSOMETHING'):
    warnings.warn('PyFFTW < 0.10.4 is known to cause problems with some '
                  'ODL functionality, see issue #1002.',
                  RuntimeWarning)
```

odl/tomo/backends/astra_setup.py
```diff
@@ -431,7 +430,7 @@ def astra_projection_geometry(geometry):
         raise ValueError('non-uniform detector sampling is not supported')
 
     if (isinstance(geometry, ParallelBeamGeometry) and
-            isinstance(geometry.detector, FlatDetector) and
+            isinstance(geometry.detector, (Flat1dDetector, Flat2dDetector)) and
```
kohr-h: Well not really. It just says that the whole geometry is embedded in a 2D space, the detector can be anything. This was probably supposed to be the duck-typed version of `isinstance(geometry, Parallel2dGeometry)`.
But honestly if somebody decides to write a new geometry that doesn't subclass `Parallel2dGeometry` but still fulfills these conditions (whatever that may be), it would require a new conversion implementation anyway. So we might as well just do `isinstance(geometry, Parallel2dGeometry)`.

odl/tomo/geometry/geometry.py
kohr-h: Any suggestion on how to call these methods? I guess it would be a short version of "methods that compute vectors associated with the geometry"

odl/tomo/geometry/conebeam.py
```diff
@@ -362,17 +369,29 @@ def src_position(self, angle):
...
+        >>> points = geom.src_position([0, np.pi / 2])
+        >>> np.allclose(points[0], [0, -2])
+        True
+        >>> np.allclose(points[1], [2, 0])
+        True
         """
-        if angle not in self.motion_params:
-            raise ValueError('`angle` {} is not in the valid range {}'
-                             ''.format(angle, self.motion_params))
+        squeeze_out = np.isscalar(angle)
```
kohr-h: Good point. I guess I would like to change this to `np.shape(angle) == ()` which covers both cases.

odl/tomo/geometry/conebeam.py
```diff
@@ -665,23 +711,28 @@ def __init__(self, apart, dpart, src_radius, det_radius, pitch=0,
 
         # Check and normalize `src_to_det_init`. Detector axes are
         # normalized in the detector class.
-        if np.linalg.norm(src_to_det_init) <= 1e-10:
-            raise ValueError('`src_to_det_init` norm {} too close to 0'
-                             ''.format(np.linalg.norm(src_to_det_init)))
+        if np.linalg.norm(src_to_det_init) == 0:
```
kohr-h: I expect issues regarding scales. If somebody works on electron tomography where lengths are in the range of nanometers, and they decide to not (or forget to) normalize the vector, the old check will produce an unreasonable error.

With all these closeness checks on user input my attitude is to let the user figure out what to provide and how to deal with numerical accuracy issues, instead of us imposing some arbitrary bounds.

odl/tomo/geometry/conebeam.py
```diff
@@ -847,9 +890,34 @@ def angles(self):
...
+
+        Parameters
+        ----------
+        angles : float or `array-like`
+            Angle(s) in radians describing the counter-clockwise rotation
+            of the detector around `axis`.
+
+        Returns
+        -------
+        axes : tuple of `numpy.ndarray`'s
```
kohr-h: Makes sense, yes. I mainly did it in the original version to resemble the input for `axes` parameters, i.e., a sequence of vectors.
I'll change it to a big array with shape `(num_axes, num_params, ndim)`.

odl/tomo/geometry/geometry.py
kohr-h: I'll probably go for `broadcast(angle, dparam).shape + (ndim,)` since that's something you can literally do in Numpy (the addition looks somewhat weird in terms of parameters, although it's valid array-wise).
But anyway, good suggestion regarding simplification.

odl/tomo/geometry/geometry.py
kohr-h: Not quite, only the cases described in the doc, not something like a 2d array of parameters. But with some minor tweaks in the code it should be simple to allow that, too.

odl/tomo/geometry/geometry.py
kohr-h: Actually no, both versions are not quite valid, because they lack the dimension to broadcast over. Technically correct would be 
```python
broadcast(angle[..., None], dparam[None, ...]).shape + (ndim,)
```

odl/tomo/geometry/parallel.py
kohr-h: Sure, this was the work-in-progress version just in case I want to switch back. Forgot to change to final.

odl/tomo/util/utility.py
kohr-h: We want `mat[0]` to be the first matrix etc. But you're right, the comment should say that rather than state the obvious.

odl/tomo/util/utility.py
kohr-h: This cannot happen. The rows in which there are nonzero entries in the first two coordinates we switch them, so the norm cannot be zero. In the other rows we insert a `1` so no way to get zero norm either (including rows that are all zeros).

odl/tomo/util/utility.py
kohr-h: Well this `any` goes along the first axis whose size is arbitrary (and can be large) so I think using `np.any` makes sense there.

odl/trafos/backends/pyfftw_bindings.py
kohr-h: Yeah this was just a fix for myself to reduce annoyance. I'll just remove the commit.

odl/tomo/geometry/conebeam.py
adler-j: Agree. Good change.

odl/tomo/geometry/geometry.py
adler-j: I guess your suggestion is fine, I can't come up with a better one anyway.

odl/tomo/util/utility.py
adler-j: My point was if the `if` statement is needed, or if we can simply go to the next line, where we get an empty assignment. I guess performance wise it shouldn't matter?

odl/tomo/geometry/geometry.py
kohr-h: Yes, I'll go with that implementation. Never thought about the "pairs of parameters" case, thx for the heads-up. It's definitely best to just support Numpy broadcasting, not making up own broadcasting rules.

odl/tomo/util/utility.py
kohr-h: ~~Ah, no that actually doesn't work. If the condition is empty, the array cannot be used for indexing.~~
Hm, now it works. Must have been some issue during the implementation. Will remove the checks.

odl/tomo/geometry/conebeam.py
kohr-h: Do we want the same for the `axes` property of detectors? Single array instead of tuple?

adler-j: For consistency yeah i think so

kohr-h: OK I'll go with it

odl/tomo/geometry/geometry.py
kohr-h: Tried to come up with a better yet short explanation.

odl/tomo/geometry/geometry.py
kohr-h: Turned out that for the "minor tweaks" I had to rethink the whole vectorization implementation, but now I'm pretty happy with what it supports.
Now you can use broadcasting across all parameters (like 2 angles and 2 detector parameters), and for that the parameters must be provided one component at a time, not one parameter at a time (otherwise no broadcasting possible).

So now it is like this (in the extreme case):
```python
        >>> geom = odl.tomo.Parallel3dEulerGeometry(...)
        >>> # 2 values for each variable, resulting in 2 vectors
        >>> angles = ([0, np.pi / 2], [0, np.pi])
        >>> dparams = ([-1, 0], [-1, 0])
        >>> pts = geom.det_point_position(angles, dparams)
        >>> pts[0]  # Corresponds to angle = (0, 0), dparam = (-1, -1)
        array([-1.,  1., -1.])
        >>> pts.shape
        (2, 3)
        >>> # 4 x 5 parameters for both
        >>> angles = dparams = (np.zeros((4, 5)), np.zeros((4, 5)))
        >>> geom.det_point_position(angles, dparams).shape
        (4, 5, 3)
        >>> # Broadcasting angles to shape (4, 5, 1, 1)
        >>> angles = (np.zeros((4, 1, 1, 1)), np.zeros((1, 5, 1, 1)))
        >>> # Broadcasting dparams to shape (1, 1, 6, 7)
        >>> dparams = (np.zeros((1, 1, 6, 1)), np.zeros((1, 1, 1, 7)))
        >>> # Total broadcast parameter shape is (4, 5, 6, 7)
        >>> geom.det_point_position(angles, dparams).shape
        (4, 5, 6, 7, 3)
```

odl/tomo/geometry/parallel.py
```diff
@@ -713,31 +905,115 @@ def det_axes_init(self):
...
+            case of an array) must be equal to the number of Euler angles
+            in this geometry.
+
+        Returns
+        -------
+        axes : `numpy.ndarray`
+            Unit vector(s) along which the detector is aligned.
+            If ``angles`` is a single pair (or triplet) of Euler angles,
+            the returned array has shape ``(2, 3)``, otherwise
+            is ``(2,) + broadcast(*angles).shape + (3,)``.
```
kohr-h: Not quite sure about this ordering. Opinion? Is it better to just add the extra dimensions to the left?

kohr-h:
Rebased, ready for the final round.

odl/tomo/geometry/parallel.py
adler-j: I'd say that they should go to the left, more consistent with other stuff

odl/tomo/geometry/detector.py
```diff
@@ -98,383 +109,799 @@ def size(self):
...
+            # Allow `param` with ndim > 1 by checking the raveled array
             raise ValueError('`param` {} not in the valid range '
                              '{}'.format(param, self.params))
-        return self.axis
+        if squeeze_out:
+            return self.axis
+        else:
+            # Produce array of shape `param.shape + (ndim,)` by broadcasting
+            axis_slc = (None,) * param.ndim + (slice(None),)
+            # TODO: use broadcast_to from Numpy when v1.10 is required
```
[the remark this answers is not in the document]
kohr-h: No, it's currently buried somewhere internally. It only exists on the tensor branch.

kohr-h:
Added the `space_ndim` thing to the abstract `Detector` class and reordered the `det_axes` and `surface_deriv` things.

I'll rebase and if CI is happy, I'll merge.

adler-j:
Great work!

PR #1066, opened 2017-07-07
API: Changed FunctionalLinearPertub to quadratic

adler-j:
I guess we could use some more tests but this is very close to done.

### DIFF
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -846,12 +846,6 @@ class FunctionalQuadraticPerturb(Functional):
         super().__init__(space=func.domain,
                          linear=func.is_linear)
 
-        # Only compute the grad_lipschitz if it is not inf
-        if (not func.grad_lipschitz == np.inf and
-                not np.isnan(func.grad_lipschitz)):
-            self.__grad_lipschitz = (func.grad_lipschitz +
-                                     linear_term.norm())
-
         self.__functional = func
         self.__quadratic_term = func.domain.field.element(quadratic_term)
 
@@ -863,6 +857,16 @@ class FunctionalQuadraticPerturb(Functional):
             self.__linear_term = func.domain.element(linear_term)
         else:
             self.__linear_term = func.domain.zero()
+
+        # Only compute the grad_lipschitz if it is not inf
+        if (not func.grad_lipschitz == np.inf and
+                not np.isnan(func.grad_lipschitz) and
+                self.quadratic_term == 0):
+            if linear_term is None:
+                self.__grad_lipschitz = func.grad_lipschitz
+            else:
+                self.__grad_lipschitz = (func.grad_lipschitz +
+                                         self.linear_term.norm())
 
     @property
     def functional(self):
@@ -889,7 +893,7 @@ class FunctionalQuadraticPerturb(Functional):
     def gradient(self):
         """Gradient operator of the functional."""
         return (self.functional.gradient +
-                self.quadratic_term * IdentityOperator(self.domain) +
+                (2 * self.quadratic_term) * IdentityOperator(self.domain) +
                 ConstantOperator(self.linear_term))
 
     @property
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -26,7 +26,7 @@ from odl.solvers.nonsmooth import (proximal_arg_scaling, proximal_translation,
 __all__ = ('Functional', 'FunctionalLeftScalarMult',
            'FunctionalRightScalarMult', 'FunctionalComp',
            'FunctionalRightVectorMult', 'FunctionalSum', 'FunctionalScalarSum',
-           'FunctionalTranslation', 'FunctionalLinearPerturb',
+           'FunctionalTranslation', 'FunctionalQuadraticPerturb',
            'FunctionalProduct', 'FunctionalQuotient', 'simple_functional')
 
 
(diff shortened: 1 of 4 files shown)


odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
```diff
@@ -24,20 +24,21 @@
...
 stepsize = simple_fixture('stepsize', [0.1, 1.0, 10.0])
-offset = simple_fixture('offset', [False, True])
+linear_offset = simple_fixture('linear_offset', ['none', True])
+quadratic_offset = simple_fixture('quadratic_offset', [False, True])
 dual = simple_fixture('dual', [False, True])
 
 
 func_params = ['l1', 'l2', 'l2^2', 'kl', 'kl_cross_ent', 'const',
                'groupl1-1', 'groupl1-2',
-               'nuclearnorm-1-1', 'nuclearnorm-1-2', 'nuclearnorm-1-inf'
```
adler-j: That this bug has been there for quite some time is scary. How often do we run largescale tests?

adler-j:
This should now be feature complete, since @MatthiasJE suggested this change, perhaps you could help out with a quick review, especially of the features?

odl/solvers/functional/functional.py
```diff
@@ -905,18 +931,25 @@ def convex_conj(self):
         optimization problems*. IEEE Signal Processing Magazine, 32.6 (2015),
         pp 31--54.
         """
-        return self.functional.convex_conj.translated(
-            self.linear_term)
+        if self.quadratic_term == 0:
+            return self.functional.convex_conj.translated(
+                self.linear_term)
+        else:
+            return super().convex_conj
```
adler-j: I guess this should be changed to python 2 style super calls

odl/test/solvers/functional/functional_test.py
```diff
@@ -510,52 +496,67 @@ def test_multiplication_with_vector(space):
...
 
     # Create an element in the space, in which to evaluate
     x = noise_element(space)
 
     # Test for evaluation of the functional
-    assert all_almost_equal(functional(x), x.norm()**2 + x.inner(linear_term),
+    assert all_almost_equal(functional(x),
+                            (orig_func(x) +
+                             quadratic_term * x.inner(x) +
+                             x.inner(linear_term)),
```
kohr-h: I'm honestly not so sure about these kinds of tests. What I mean by "these tests" is tests of the *implementation* of something rather than the *intended functionality* of something.

What I think one should do instead (and this is not easy while implementing stuff) is to take a step back, ask oneself "What is this function supposed to do?" and then test that. Or test things that implicitly follow but aren't directly implemented.
In this case it's almost the same, but there are still some differences, for instance what happens with weighted spaces. Errors there won't be caught by this test.

Here I think it's better to test an explicit example, and when we feel we need to settle on some behavior regarding weighting (it will influence the gradient) we add a test for that.

odl/solvers/functional/functional.py
```diff
@@ -843,45 +846,68 @@ def __init__(self, func, linear_term):
...
+        self.__quadratic_term = func.domain.field.element(quadratic_term)
+
+        if self.quadratic_term < 0:
+            raise TypeError('`quadratic_term` {} must be non-negative'
+                            ''.format(self.quadratic_term))
+
+        if linear_term is not None:
+            self.__linear_term = func.domain.element(linear_term)
+        else:
+            self.__linear_term = func.domain.zero()
```
adler-j: Main idea here is that users should be able to simply call `FunctionalQuadraticPerturb(func, quadratic_term=3.2)` without mentioning the linear term

odl/test/solvers/functional/functional_test.py
adler-j: > I'm honestly not so sure about these kinds of tests. What I mean by "these tests" is tests of the implementation of something rather than the intended functionality of something.

Well (imo) these dont test anything right now, but are only there as regression tests to give us a huge red flag if something gets changed so we actually notice. I don't really see any issue with that. They also make sure the code is run at least once which stops the worst of bugs.

With that said, you are certainly correct in that this could be done in a better manner, but that would require quite some effort here, something I probably wouldn't feel is warranted for a function that is as specialized as this once.

In other places, like `Gradient` which is called all the time, I certainly agree and a proper look at those tests is sorely needed.

odl/solvers/functional/functional.py
```diff
@@ -822,19 +822,22 @@ def __str__(self):
...
-    def __init__(self, func, linear_term):
+    def __init__(self, func, quadratic_term=0, linear_term=None):
         """Initialize a new instance.
 
         Parameters
         ----------
         func : `Functional`
             Function corresponding to ``f``.
-        linear_term : `domain` element
+        quadratic_term : ``domain.field`` element, optional
```
adler-j: Any better suggestion? In math I'd just call this something like `a`, but that clearly won't work here.

kohr-h: How about `quadratic_coeff`? It's a polynomial after all.

odl/solvers/functional/functional.py
kohr-h: Fair enough, but there's still the issue that the functional doesn't work with e.g. `RealNumbers` like that.

odl/solvers/functional/functional.py
kohr-h: In some sense, the `linear_term` (name also not ideal) is also just a vector of coefficients, at least in the discrete case. Otherwise it's more of a factor than a term.

adler-j: quadratic_coeff is a great idea! I'll give the linear one some thought.

odl/solvers/functional/functional.py
```diff
@@ -822,19 +822,22 @@ def __str__(self):
...
+            Quadratic term.
```
adler-j: Second thought, why do we need that. Its only a requirement for the proximal. I'll remove the requirement.

odl/solvers/functional/functional.py
adler-j: RealNumbers does not impement `inner`, currently at least, so that ain't much of an issue for now.

kohr-h: But still weird to let this guy raise an error for the default in that case.

Issue #1028, opened 2017-06-09
Proper pickling of ODL objects


adler-j:
Currently we rely on the built in pickle interface when people want to use pickle to store ODL objects. This works fine as long as everyone agrees on a ODL version, but for long term stability this is problematic. In particular, any backend change tends to break the pickle, even if there is no surface level change.

To solve this, we need to implement proper pickling, that is, add the stuff under [Pickling Class Instances](https://docs.python.org/3/library/pickle.html#pickling-class-instances).

Initially the most commonly pickled ODL object is geometries so suggestion would be to start this effort there.

PR #1067, opened 2017-07-08
ENH: Add Operator.norm, closes #1065

### LINKED ISSUE #1065: Add Operator.norm
Computing operator norms happens all the time in code, and we mostly refer to the `odl.power_method_opnorm` method, which is fine but error prone and occationally slow.

I propose that we add a new method called `Operator.norm` with the following default implementation:

```python
def norm(self, estimate=False):
    if not estimate:
        raise NotImplementedError('Operator norm not implemented')
    else:
        return odl.power_method_opnorm(self)
```

Some classes could then over-ride this method when there are known formulae. For example, in `DiagonalOperator` we could implement

```python
def norm(self, estimate=False):
    # Pen-and-paper exercise shows that this holds
    norms = [op.norm(estimate=estimate) for op in self.operators]
    return max(norms)
```

for some other operators I'm sure we could obtain similar bounds.

One obvious issue with this is non-linear operators, so we'd need to solve #551.

adler-j:
Making a small initial PR to review the concept. If we go further we can start adding this to various classes.

### DIFF
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -707,6 +707,54 @@ class Operator(object):
                     raise_from(new_exc, err)
         return out
 
+    def norm(self, estimate=False, **kwargs):
+        """Return the operator norm of this operator.
+
+        If this operator is non-linear, this should be the Lipschitz constant.
+
+        Parameters
+        ----------
+        estimate : bool
+            If true, estimate the operator norm. By default, it is estimated
+            using `power_method_opnorm`, which is only applicable for linear
+            operators.
+            Subclasses are allowed to ignore this parameter if they can provide
+            an exact value.
+
+        Other Parameters
+        ----------------
+        kwargs :
+            If ``estimate`` is True, pass these arguments to the
+            `power_method_opnorm` call.
+
+        Returns
+        -------
+        norm : float
+
+        Examples
+        --------
+        Some operators know their own operator norm and do not need an estimate
+
+        >>> spc = odl.rn(3)
+        >>> id = odl.IdentityOperator(spc)
+        >>> id.norm(True)
+        1.0
+
+        For others, there is no closed form expression and an estimate is
+        needed:
+
+        >>> spc = odl.uniform_discr(0, 1, 3)
+        >>> grad = odl.Gradient(spc)
+        >>> opnorm = grad.norm(estimate=True)
+        """
+        if not estimate:
+            raise NotImplementedError('`Operator.norm()` not implemented, use '
+                                      '`Operator.norm(estimate=True)` to '
+                                      'obtain an estimate.')
+        else:
+            from odl.operator.oputils import power_method_opnorm
+            return power_method_opnorm(self, **kwargs)
+
     def __add__(self, other):
         """Return ``self + other``.
 
--- a/odl/operator/default_ops.py
+++ b/odl/operator/default_ops.py
@@ -15,6 +15,7 @@ standard_library.install_aliases()
 from builtins import super
 
 from copy import copy
+import numpy as np
 
 from odl.operator.operator import Operator
 from odl.space import ProductSpace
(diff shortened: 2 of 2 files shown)


adler-j:
Perhaps @MatthiasJE could give a review of this as well?

odl/operator/operator.py
```diff
@@ -707,6 +707,54 @@ def __call__(self, x, out=None, **kwargs):
...
+        >>> grad = odl.Gradient(spc)
+        >>> opnorm = grad.norm(estimate=True)
+        """
+        if not estimate:
+            raise NotImplementedError('`Operator.norm()` not implemented, use '
+                                      '`Operator.norm(estimate=True)` to '
+                                      'obtain an estimate.')
+        else:
+            from odl.operator.oputils import power_method_opnorm
+            return power_method_opnorm(self)
```
adler-j: missing **kwargs

odl/operator/operator.py
```diff
@@ -707,6 +707,54 @@ def __call__(self, x, out=None, **kwargs):
                     raise_from(new_exc, err)
         return out
 
+    def norm(self, estimate=False, **kwargs):
+        """Return the operator norm of this operator.
+
+        If this operator is non-linera, this should be the Lipschitz constant.
+
```
MatthiasJE: typo "lineara"

odl/operator/operator.py
```diff
@@ -707,6 +707,54 @@ def __call__(self, x, out=None, **kwargs):
                     raise_from(new_exc, err)
         return out
 
+    def norm(self, estimate=False, **kwargs):
```
MatthiasJE: I think it would be good if it was possible to set the operator norm manually. The current framework would not allow for that.

odl/operator/operator.py
```diff
@@ -707,6 +707,54 @@ def __call__(self, x, out=None, **kwargs):
...
+
+        For others, there is no closed form expression and an estimate is
+        needed:
+
+        >>> spc = odl.uniform_discr(0, 1, 3)
+        >>> grad = odl.Gradient(spc)
+        >>> opnorm = grad.norm(estimate=True)
+        """
+        if not estimate:
+            raise NotImplementedError('`Operator.norm()` not implemented, use '
```
MatthiasJE: Most gradient operators have closed form norms that depend on the discretization. It may be useful to utilize this here.

odl/operator/operator.py
adler-j: Quite skeptical to a global way of doing this. The main reason is that `Operator`s are immutable through-out ODL, which means that if I am given an operator, I know it will always return the same thing at all times, and I don't have to be afraid of it changing values due to some change. While we really only depend on this property for `__call__`, it is health to try to keep it that way (simply because code becomes so much easier to debug).

Even so, this framework allows you to manually set the norm in sub-classes, its just that we don't provide a global way of doing it.

Is there any specific use-case where you would want to set it manually? Perhaps that would give some more insight into how to design this.

odl/operator/operator.py
adler-j: I agree, but that will be added in another PR, in order to keep this simple. Other operators that are prime candidates include the Fourier transform and \*caugh @kohr-h\* convolution operator, as well as `ProductSpaceOperator`.

odl/operator/operator.py
MatthiasJE: At the moment I am looking at X-ray transform examples (e.g. PET, CT). Here you need to estimate the operator norm but depending on the computational cost, you may want to do it differently for each case. For example, on a small scanner you may want to run 100 power iterations, on a big one you may want to do it for 20 iterations and add a 5 % margin to it or to load stored values for it. Of course, this can be done for subclasses that really need it.

kohr-h: For that case I think you can just steer via `kwargs` how many iterations you run and then add the safety margin yourself.
And yes, keeping things immutable as far as possible is favorable, also in the light of serialization (#1028).

kohr-h:
> Also the changes do not include automatic derivation of norms for composition of operators.

That would be useful, agreed, but then you get two different variants of "estimate" namely "estimate numerically" or "estimate from the norms of the constituents". If we default to the latter, I think it wouldn't be right to call it `norm` anymore since it's only an upper bound.

> reuse computed results.

One way of doing that, and I suggest that we think about that separately from this PR, is to add a cache to this method. A possible implementation of that would be to store a tuple `(num_powmethod_iter, norm_approx)`, and when you call the function with `num_iter <= num_powmethod_iter` the method would return the cached `norm_approx`, and otherwise rerun the iteration and store the new value. A kind of stripped-down version of an LRU cache.

PR #1070, opened 2017-07-10
BUG: fix gradient of GroupL1Norm. Also safeguard derivative of PointwiseNorm.

### LINKED ISSUE #950: GroupL1Norm gradient not implemented for product spaces
I get the following error when I want to take the gradient of the `GroupL1Norm`.
```
>>> import numpy as np
>>> import odl
>>> reco_space = odl.uniform_discr(0,1,10)
>>> grad = odl.Gradient(reco_space)
>>> l1 = odl.solvers.GroupL1Norm(grad.range)
>>> x = reco_space.element(lambda x: x**2)
>>> (l1 * grad).gradient(x)
  File "/home/aringh/git/odl/odl/operator/operator.py", line 695, in __call__
    out = self._call_out_of_place(x, **kwargs)
  File "/home/aringh/git/odl/odl/solvers/functional/functional.py", line 577, in _call
    return op.derivative(x).adjoint(func.gradient(op(x)))
  File "/home/aringh/git/odl/odl/operator/operator.py", line 695, in __call__
    out = self._call_out_of_place(x, **kwargs)
  File "/home/aringh/git/odl/odl/solvers/functional/default_functionals.py", line 305, in _call
    np.divide(x, result, out=result, where=result != 0)
TypeError: return arrays must be of ArrayType
```
Is this a bug, or am I making a logical mistake here?

Trying to make the example somewhat smaller still produces the error:
```
>>> y = grad(x)
>>> l1.gradient(y)
  File "/home/aringh/git/odl/odl/operator/operator.py", line 695, in __call__
    out = self._call_out_of_place(x, **kwargs)
  File "/home/aringh/git/odl/odl/solvers/functional/default_functionals.py", line 305, in _call
    np.divide(x, result, out=result, where=result != 0)
TypeError: return arrays must be of ArrayType
```

aringh:
This should fix #950. Also added the group L1 norm to the test suit of functionals.

The commit also contains a safeguarding of the derivative of the `PointwiseNorm`. If I understand correctly, for Lp-spaces with exponent >= 2 the derivative of the pointwise norm exists everywhere and the only issue should be when one gets 0/0 in some points. Ergo the current fix.

### DIFF
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -340,14 +340,15 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
+            gi *= np.abs(gi) ** (self.exponent - 2)
             if self.exponent >= 2:
-                tmp = vf_pwnorm_fac * np.abs(gi) ** (2 - self.exponent)
                 # Any component that is zero is not divided with
-                nz = (tmp.asarray() != 0)
-                gi[nz] /= tmp[nz]
-
+                nz = (vf_pwnorm_fac.asarray() != 0)
+                gi[nz] /= vf_pwnorm_fac[nz]
             else:
-                gi *= np.abs(gi) ** (self.exponent - 2) / vf_pwnorm_fac
+                # For exponents < 2, there will be a singularity if any
+                # component is zero
+                gi /= vf_pwnorm_fac
 
         return PointwiseInner(self.domain, inner_vf, weighting=self.weights)
 
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -341,10 +341,11 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
 
         for gi in inner_vf:
             if self.exponent >= 2:
-                with np.errstate(invalid='ignore', divide='ignore'):
-                    tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
-                    # Any component that is zero is not divided with
-                    gi /= tmp if not tmp else 1
+                tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
+                # Any component that is zero is not divided with
+                for i, val in zip(np.arange(len(tmp)), tmp):
+                    if not val == 0.0:
+                        gi[i] /= val
             else:
                 gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
 
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -341,13 +341,13 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
 
         for gi in inner_vf:
             if self.exponent >= 2:
-                tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
+                tmp = vf_pwnorm_fac * np.abs(gi) ** (2 - self.exponent)
                 # Any component that is zero is not divided with
-                for i, val in zip(np.arange(len(tmp)), tmp):
-                    if not val == 0.0:
-                        gi[i] /= val
+                nz = (tmp.asarray() != 0)
+                gi[nz] /= tmp[nz]
+
             else:
-                gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+                gi *= np.abs(gi) ** (self.exponent - 2) / vf_pwnorm_fac
 
         return PointwiseInner(self.domain, inner_vf, weighting=self.weights)
 
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -340,14 +340,15 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
-            gi *= np.abs(gi) ** (self.exponent - 2)
+            gi *= gi.ufuncs.absolute().ufuncs.power(self.exponent - 2)
             if self.exponent >= 2:
                 # Any component that is zero is not divided with
                 nz = (vf_pwnorm_fac.asarray() != 0)
                 gi[nz] /= vf_pwnorm_fac[nz]
             else:
-                # For exponents < 2, there will be a singularity if any
-                # component is zero
+                # For exponents < 2 there will be a singularity if any
+                # component is zero. This reullts in inf or nan. See the
+                # documentation for further details.
                 gi /= vf_pwnorm_fac
 
         return PointwiseInner(self.domain, inner_vf, weighting=self.weights)
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -340,7 +340,12 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
-            gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+            if self.exponent >= 2:
+                with np.errstate(invalid='ignore', divide='ignore'):
+                    gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+                    gi = np.nan_to_num(x=gi)
+            else:
+                gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
 
         return PointwiseInner(self.domain, inner_vf, weighting=self.weights)
 
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -347,7 +347,7 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
                 gi[nz] /= vf_pwnorm_fac[nz]
             else:
                 # For exponents < 2 there will be a singularity if any
-                # component is zero. This reullts in inf or nan. See the
+                # component is zero. This results in inf or nan. See the
                 # documentation for further details.
                 gi /= vf_pwnorm_fac
 
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -342,8 +342,9 @@ class PointwiseNorm(PointwiseTensorFieldOperator):
         for gi in inner_vf:
             if self.exponent >= 2:
                 with np.errstate(invalid='ignore', divide='ignore'):
-                    gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
-                    gi = np.nan_to_num(x=gi)
+                    tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
+                    # Any component that is zero is not divided with
+                    gi /= tmp if not tmp else 1
             else:
                 gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
 
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -291,8 +291,11 @@ class GroupL1Norm(Functional):
 
             def _call(self, x):
                 """Return ``self(x)``."""
-                return functional.pointwise_norm.derivative(x).adjoint(
-                    np.sign(functional.pointwise_norm(x)))
+                to_return = functional.pointwise_norm(x)
+                to_return.ufuncs.sign(out=to_return)
+                functional.pointwise_norm.derivative(x).adjoint(to_return,
+                                                                out=to_return)
+                return to_return
 
         return GroupL1Gradient()
 
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -289,13 +289,14 @@ class GroupL1Norm(Functional):
                 super(GroupL1Gradient, self).__init__(
                     functional.domain, functional.domain, linear=False)
 
-            def _call(self, x):
+            def _call(self, x, out):
                 """Return ``self(x)``."""
-                to_return = functional.pointwise_norm(x)
-                to_return.ufuncs.sign(out=to_return)
-                functional.pointwise_norm.derivative(x).adjoint(to_return,
-                                                                out=to_return)
-                return to_return
+                pwnorm_x = functional.pointwise_norm(x)
+                pwnorm_x.ufuncs.sign(out=pwnorm_x)
+                functional.pointwise_norm.derivative(x).adjoint(pwnorm_x,
+                                                                out=out)
+
+                return out
 
         return GroupL1Gradient()
 
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -291,22 +291,8 @@ class GroupL1Norm(Functional):
 
             def _call(self, x):
                 """Return ``self(x)``."""
-                p = functional.pointwise_norm.exponent
-
-                if functional.pointwise_norm.exponent == 1:
-                    result = np.abs(x)
-                    np.divide(x, result, out=result, where=result != 0)
-                    return result
-                elif functional.pointwise_norm.exponent == 2:
-                    result = functional.pointwise_norm(x)
-                    np.divide(x, result, out=result, where=result != 0)
-                    return result
-                else:
-                    dividend = np.power(np.abs(x), p - 2) * x
-                    divisor = np.power(functional.pointwise_norm(x), p - 1)
-                    np.divide(dividend, divisor, out=divisor,
-                              where=divisor != 0)
-                    return divisor
+                return functional.pointwise_norm.derivative(x).adjoint(
+                    np.sign(functional.pointwise_norm(x)))
 
         return GroupL1Gradient()
 
--- a/odl/test/operator/tensor_ops_test.py
+++ b/odl/test/operator/tensor_ops_test.py
@@ -215,7 +215,7 @@ def test_pointwise_norm_gradient_real(exponent):
     tmp = pwnorm(point).ufuncs.power(1 - exponent)
     v_field = vfspace.element()
     for i in range(len(v_field)):
-        v_field[i] = tmp * point[i] * np.abs(point[i])**(exponent - 2)
+        v_field[i] = tmp * point[i] * np.abs(point[i]) ** (exponent - 2)
     pwinner = odl.PointwiseInner(vfspace, v_field)
     expected_result = pwinner(direction)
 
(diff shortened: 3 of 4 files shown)


odl/operator/tensor_ops.py
```diff
@@ -340,7 +340,12 @@ def derivative(self, vf):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
-            gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+            if self.exponent >= 2:
+                with np.errstate(invalid='ignore', divide='ignore'):
+                    gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+                    gi = np.nan_to_num(x=gi)
```
adler-j: Second thought, this does not do anything, it simply modifies `gi` in here, not `inner_vf`, so it does not change the final result.

Could you add a test for this to make sure it really works?

adler-j:
Any chance this can get fixed quickly?

adler-j:
Bumping this again

aringh:
So, finally I hope that this is done. @adler-j you were, as always ;-), right - my previous change in the derivative did nothing. I have now fixed it, and also added a (very small) test to check that it indeed does not produce `nan`s for exponents >= 2.

More test for the derivative should be added. The question is if this should be done in this pull request, or if we should let this one go in, close #950, and open a new issue for adding more test?

odl/operator/tensor_ops.py
```diff
@@ -337,7 +337,14 @@ def derivative(self, vf):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
-            gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+            if self.exponent >= 2:
+                tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
+                # Any component that is zero is not divided with
+                for i, val in zip(np.arange(len(tmp)), tmp):
```
aringh: My previous attempt on this looked like
```python
tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
# Any component that is zero is not divided with
gi /= tmp if not tmp else 1.0
```
in which I hoped that the ``if``-``else`` statement would make a point-wise comparison and thus create a mask. However, when travis faild on this one i checked and that was not the case. @adler-j or @kohr-h : is there a better way to do this, instead of looping over the components?

odl/operator/tensor_ops.py
```diff
@@ -337,7 +337,14 @@ def derivative(self, vf):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
-            gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+            if self.exponent >= 2:
```
kohr-h: Should be `< 2`, no? The code handles the division-by-zero case, which does not correspond to this condition.

odl/operator/tensor_ops.py
kohr-h: No, `if numpy_array` will fail because an array cannot be unambiguously treated as a boolean.

Use a boolean mask:
```python
nz = (tmp != 0)
gi[nz] /= tmp[nz]
```

odl/operator/tensor_ops.py
```diff
@@ -337,7 +337,14 @@ def derivative(self, vf):
...
+                tmp = vf_pwnorm_fac * gi ** (self.exponent - 2)
+                # Any component that is zero is not divided with
+                for i, val in zip(np.arange(len(tmp)), tmp):
+                    if not val == 0.0:
```
kohr-h: This is gonna go away anyway, but just a small comment: If there is a comparison operator for the opposite, use it, instead of prepending `not`: `val != other_val` rather than `not val == other_val`. More readable.

odl/test/operator/tensor_ops_test.py
```diff
@@ -200,6 +200,65 @@ def test_pointwise_norm_weighted(exponent):
     assert all_almost_equal(out, true_norm.reshape(-1))
 
 
+def test_pointwise_norm_gradient_real(exponent):
```
kohr-h: Short docstring

odl/operator/tensor_ops.py
kohr-h: And cast to Numpy arrays before doing that, of course. (We really need to get the tensor branch in, then all this casting back and forth has an end...)

odl/operator/tensor_ops.py
aringh: I understood the problem as being that for any ``exponent``, if the element contains a zero component the code will divide by 0. For ``exponent < 2`` the derivative is actually undefined in this point (and hence we let it divide by zero...), but for ``exponent >= 2`` the derivative is actually zero in that direction. Thus the safe-guarding is only in the case of ``exponent >= 2``.

Have I misunderstood the derivative?

kohr-h: Ah, I was under the assumption that you solve the `0 ** (exponent - 2)` issue. Now the distinction makes sense.

Anyway, I would still suggest a different implementation for exponent < 2, namely
```python
gi *= gi ** (2 - self.exponent) / vf_pwnorm_fac
```
It also seems to me that we should have `np.abs(gi) ** (2 - self.exponent)` and `np.abs(gi) ** (self.exponent - 2)`, respectively. Compare with the formula.

kohr-h:
Any chance this can go in soon?

aringh:
I think this should be the last round on this. However, locally I get the following test failures. But I do not think that I have made any changes in these files here...? 

I rebased on master earlier today, and I have the same test failures on there.

```
=================================== FAILURES ===================================
_____________________________ test_init_npy_tspace _____________________________

    def test_init_npy_tspace():
        """Test initialization patterns and options for ``NumpyTensorSpace``."""
        # Basic class constructor
        NumpyTensorSpace((3, 4))
        NumpyTensorSpace((3, 4), dtype=int)
        NumpyTensorSpace((3, 4), dtype=float)
        NumpyTensorSpace((3, 4), dtype=complex)
        NumpyTensorSpace((3, 4), dtype=complex, exponent=1.0)
        NumpyTensorSpace((3, 4), dtype=complex, exponent=float('inf'))
>       NumpyTensorSpace((3, 4), dtype='S1')

odl/test/space/tensors_test.py:120: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'NumpyTensorSpace' object has no attribute '_NumpyTensorSpace__weighting'") raised in repr()] NumpyTensorSpace object at 0x7fd812f7fe48>
shape = (3, 4), dtype = 'S1', kwargs = {}

    def __init__(self, shape, dtype=None, **kwargs):
        """Initialize a new instance.
    
            Parameters
            ----------
            shape : positive int or sequence of positive ints
                Number of entries per axis for elements in this space. A
                single integer results in a space with rank 1, i.e., 1 axis.
            dtype :
                Data type of each element. Can be provided in any
                way the `numpy.dtype` function understands, e.g.
                as built-in type or as a string. For ``None``,
                the `default_dtype` of this space (``float64``) is used.
            exponent : positive float, optional
                Exponent of the norm. For values other than 2.0, no
                inner product is defined.
    
                This option has no impact if either ``dist``, ``norm`` or
                ``inner`` is given, or if ``dtype`` is non-numeric.
    
                Default: 2.0
    
            Other Parameters
            ----------------
            weighting : optional
                Use weighted inner product, norm, and dist. The following
                types are supported as ``weighting``:
    
                ``None``: no weighting, i.e. weighting with ``1.0`` (default).
    
                `Weighting`: Use this weighting as-is. Compatibility
                with this space's elements is not checked during init.
    
                ``float``: Weighting by a constant.
    
                array-like: Pointwise weighting by an array.
    
                This option cannot be combined with ``dist``,
                ``norm`` or ``inner``. It also cannot be used in case of
                non-numeric ``dtype``.
    
            dist : callable, optional
                Distance function defining a metric on the space.
                It must accept two `NumpyTensor` arguments and return
                a non-negative real number. See ``Notes`` for
                mathematical requirements.
    
                By default, ``dist(x, y)`` is calculated as ``norm(x - y)``.
    
                This option cannot be combined with ``weight``,
                ``norm`` or ``inner``. It also cannot be used in case of
                non-numeric ``dtype``.
    
            norm : callable, optional
                The norm implementation. It must accept a
                `NumpyTensor` argument, return a non-negative real number.
                See ``Notes`` for mathematical requirements.
    
                By default, ``norm(x)`` is calculated as ``inner(x, x)``.
    
                This option cannot be combined with ``weight``,
                ``dist`` or ``inner``. It also cannot be used in case of
                non-numeric ``dtype``.
    
            inner : callable, optional
                The inner product implementation. It must accept two
                `NumpyTensor` arguments and return an element of the field
                of the space (usually real or complex number).
                See ``Notes`` for mathematical requirements.
    
                This option cannot be combined with ``weight``,
                ``dist`` or ``norm``. It also cannot be used in case of
                non-numeric ``dtype``.
    
            kwargs :
                Further keyword arguments are passed to the weighting
                classes.
    
            See Also
            --------
            rn : constructor for real tensor spaces
            cn : constructor for complex tensor spaces
            tensor_space :
                constructor for tensor spaces of arbitrary scalar
                data type
    
            Notes
            -----
            - A distance function or metric on a space :math:`\mathcal{X}`
              is a mapping
              :math:`d:\mathcal{X} \\times \mathcal{X} \\to \mathbb{R}`
              satisfying the following conditions for all space elements
              :math:`x, y, z`:
    
              * :math:`d(x, y) \geq 0`,
              * :math:`d(x, y) = 0 \Leftrightarrow x = y`,
              * :math:`d(x, y) = d(y, x)`,
              * :math:`d(x, y) \\leq d(x, z) + d(z, y)`.
    
            - A norm on a space :math:`\mathcal{X}` is a mapping
              :math:`\| \cdot \|:\mathcal{X} \\to \mathbb{R}`
              satisfying the following conditions for all
              space elements :math:`x, y`: and scalars :math:`s`:
    
              * :math:`\| x\| \geq 0`,
              * :math:`\| x\| = 0 \Leftrightarrow x = 0`,
              * :math:`\| sx\| = |s| \cdot \| x \|`,
              * :math:`\| x+y\| \\leq \| x\| +
                \| y\|`.
    
            - An inner product on a space :math:`\mathcal{X}` over a field
              :math:`\mathbb{F} = \mathbb{R}` or :math:`\mathbb{C}` is a
              mapping
              :math:`\\langle\cdot, \cdot\\rangle: \mathcal{X} \\times
              \mathcal{X} \\to \mathbb{F}`
              satisfying the following conditions for all
              space elements :math:`x, y, z`: and scalars :math:`s`:
    
              * :math:`\\langle x, y\\rangle =
                \overline{\\langle y, x\\rangle}`,
              * :math:`\\langle sx + y, z\\rangle = s \\langle x, z\\rangle +
                \\langle y, z\\rangle`,
              * :math:`\\langle x, x\\rangle = 0 \Leftrightarrow x = 0`.
    
            Examples
            --------
            Explicit initialization with the class constructor:
    
            >>> space = NumpyTensorSpace(3, float)
            >>> space
            rn(3)
            >>> space.shape
            (3,)
            >>> space.dtype
            dtype('float64')
    
            A more convenient way is to use factory functions:
    
            >>> space = odl.rn(3, weighting=[1, 2, 3])
            >>> space
            rn(3, weighting=[1, 2, 3])
            >>> space = odl.tensor_space((2, 3), dtype=int)
            >>> space
            tensor_space((2, 3), dtype=int)
            """
        super(NumpyTensorSpace, self).__init__(shape, dtype)
        if self.dtype.char not in self.available_dtypes():
            raise ValueError('`dtype` {!r} not supported'
>                            ''.format(dtype_str(dtype)))
E           ValueError: `dtype` '|S1' not supported

odl/space/npy_tensors.py:224: ValueError
__________________________ PEP8-check(ignoring E402) ___________________________
/home/aringh/git/odl/odl/tomo/backends/astra_cuda.py:38:1: W293 blank line contains whitespace

^
/home/aringh/git/odl/odl/tomo/backends/astra_cuda.py:43:1: W293 blank line contains whitespace

^
/home/aringh/git/odl/odl/tomo/backends/astra_cuda.py:193:1: W293 blank line contains whitespace

^
/home/aringh/git/odl/odl/tomo/backends/astra_cuda.py:198:1: W293 blank line contains whitespace

^

======= 2 failed, 3843 passed, 1074 skipped, 13 xfailed in 25.75 seconds =======
```

adler-j:
We just merged the Tensor branch and expect quite a bit of breakage going forward, it seems this is one of those.

odl/test/operator/tensor_ops_test.py
```diff
@@ -192,6 +192,65 @@ def test_pointwise_norm_weighted(exponent):
...
+    test_point = np.array([[[0, 0],  # This makes the point singular for p < 2
+                            [1, 2]]])
+    test_direction = np.array([[[1, 2],
+                                [4, 5]]])
+
+    point = vfspace.element(test_point)
+    direction = vfspace.element(test_direction)
+    func_pwnorm = pwnorm.derivative(point)
+
+    assert not np.isnan(func_pwnorm(direction)).asarray().any()
```
aringh: These tests were changed in the last commit due to my confusion reported in #1234, but before I realized that one could also use `np.any`. Should they be changed or is it fine?

kohr-h:
The issue with `dtype='S1'` is fixed now by #1236

aringh:
Is this one good to go?

odl/solvers/functional/default_functionals.py
```diff
@@ -291,22 +291,8 @@ def __init__(self):
...
-                    np.divide(x, result, out=result, where=result != 0)
-                    return result
-                else:
-                    dividend = np.power(np.abs(x), p - 2) * x
-                    divisor = np.power(functional.pointwise_norm(x), p - 1)
-                    np.divide(dividend, divisor, out=divisor,
-                              where=divisor != 0)
-                    return divisor
+                return functional.pointwise_norm.derivative(x).adjoint(
+                    np.sign(functional.pointwise_norm(x)))
```
kohr-h: Nice reduction! Can we also try to optimize things a bit here? Not much, only add `out` to the signature and do
```python
pwnorm_x = functional.pointwise_norm(x)
pwnorm_x.ufuncs.sign(out=pwnorm_x)
functional.pointwise_norm.derivative(x).adjoint(pwnorm_x, out=out)
return out
```

odl/test/operator/tensor_ops_test.py
kohr-h: I like `np.any` better because it comes first. It reads more like normal English, i.e., `assert not any <lengthy expression>` vs. `assert not <lengthy expression> any`.

odl/test/operator/tensor_ops_test.py
```diff
@@ -192,6 +192,65 @@ def test_pointwise_norm_weighted(exponent):
...
+
+    # 3d
+    fspace = odl.uniform_discr([0, 0], [1, 1], (2, 2))
+    vfspace = ProductSpace(fspace, 3)
+    pwnorm = PointwiseNorm(vfspace, exponent)
+
+    test_point = np.array([[[0, 0],  # This makes the point singular for p < 2
```
kohr-h: Currently this is skipped, right?

odl/test/operator/tensor_ops_test.py
```diff
@@ -192,6 +192,65 @@ def test_pointwise_norm_weighted(exponent):
...
+    if exponent == float('inf'):
+        fspace = odl.uniform_discr([0, 0], [1, 1], (2, 2))
+        vfspace = ProductSpace(fspace, 1)
+        pwnorm = PointwiseNorm(vfspace, exponent)
+        point = vfspace.one()
+        with pytest.raises(NotImplementedError):
+            pwnorm.derivative(point)
+        return
+
+    # TODO: implement good tests also for the 'normal behaviour'
```
kohr-h: Would be nice.

kohr-h:
That said, these are minor changes, overall the PR looks very good.

odl/operator/tensor_ops.py
```diff
@@ -340,7 +340,15 @@ def derivative(self, vf):
         inner_vf = vf.copy()
 
         for gi in inner_vf:
-            gi /= vf_pwnorm_fac * gi ** (self.exponent - 2)
+            gi *= np.abs(gi) ** (self.exponent - 2)
```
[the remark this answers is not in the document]
aringh: Took me a while to figure out that this is in fact wrong ;-) We want the result of line 343 to be ```gi * |gi|^(p - 2)```, at least according to the doc-string (which is what I have been following). Luckily I also followed the advise an implemented a test for the calculations, which caught it.

aringh:
Good to go now? :-)

odl/operator/tensor_ops.py
```diff
@@ -340,14 +340,15 @@ def derivative(self, vf):
...
+            gi *= gi.ufuncs.absolute().ufuncs.power(self.exponent - 2)
             if self.exponent >= 2:
                 # Any component that is zero is not divided with
                 nz = (vf_pwnorm_fac.asarray() != 0)
                 gi[nz] /= vf_pwnorm_fac[nz]
             else:
-                # For exponents < 2, there will be a singularity if any
-                # component is zero
+                # For exponents < 2 there will be a singularity if any
+                # component is zero. This reullts in inf or nan. See the
```
kohr-h: typo "resullts"

odl/test/operator/tensor_ops_test.py
```diff
@@ -203,10 +203,50 @@ def test_pointwise_norm_gradient_real(exponent):
...
+    pwnorm = PointwiseNorm(vfspace, exponent)
+
+    point = noise_element(vfspace)
+    direction = noise_element(vfspace)
+
+    # Computing expected result
+    tmp = pwnorm(point).ufuncs.power(1 - exponent)
+    v_field = vfspace.element()
+    for i in range(len(v_field)):
+        v_field[i] = tmp * point[i] * np.abs(point[i])**(exponent - 2)
```
kohr-h: spaces around `**`

odl/test/operator/tensor_ops_test.py
```diff
@@ -203,10 +203,50 @@ def test_pointwise_norm_gradient_real(exponent):
...
+    point = noise_element(vfspace)
+    direction = noise_element(vfspace)
 
-    # The gradient is only well-defined in all points if the exponent is >= 2
-    if exponent < 2:
+    # Computing expected result
+    tmp = pwnorm(point).ufuncs.power(1 - exponent)
+    v_field = vfspace.element()
+    for i in range(len(v_field)):
+        v_field[i] = tmp * point[i] * np.abs(point[i])**(exponent - 2)
```
kohr-h: same here

aringh:
Last changes done. Can someone with the authority to merge do it after travis is done?

PR #1075, opened 2017-07-10
add infimal convolution as an abstract functional

MatthiasJE:
This is a tiny pull request. It only implements the infimal convolution as an abstract functional. The only easy implementation is the convex conjugate.

### DIFF
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -26,8 +26,9 @@ from odl.solvers.nonsmooth import (proximal_arg_scaling, proximal_translation,
 __all__ = ('Functional', 'FunctionalLeftScalarMult',
            'FunctionalRightScalarMult', 'FunctionalComp',
            'FunctionalRightVectorMult', 'FunctionalSum', 'FunctionalScalarSum',
-           'FunctionalTranslation', 'FunctionalQuadraticPerturb',
-           'FunctionalProduct', 'FunctionalQuotient', 'simple_functional')
+           'FunctionalTranslation', 'InfimalConvolution',
+           'FunctionalQuadraticPerturb', 'FunctionalProduct',
+           'FunctionalQuotient', 'simple_functional')
 
 
 class Functional(Operator):
@@ -820,6 +821,83 @@ class FunctionalTranslation(Functional):
         """Return ``str(self)``."""
         return '{}.translated({})'.format(self.functional,
                                           self.translation)
+
+
+class InfimalConvolution(Functional):
+
+    """The functional representing ``h(x) = inf_y f(x-y) + g(y)``."""
+
+    def __init__(self, funcA, funcB):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        funcA : `Functional`
+            Function corresponding to ``f``.
+        funcB : `Functional`
+            Function corresponding to ``g``.
+
+        Example
+        -------
+        >>> space = odl.rn(3)
+        >>> l1 = odl.solvers.L1Norm(space)
+        >>> l2 = odl.solvers.L2Norm(space)
+        >>> f = odl.solvers.InfimalConvolution(l1, l2)
+        >>> f.convex_conj(0.5*f.domain.one())
+        >>> 0
+        """
+        if not isinstance(funcA, Functional):
+            raise TypeError('`func` {} is not a `Functional` instance'
+                            ''.format(funcA))
+
+        if not isinstance(funcB, Functional):
+            raise TypeError('`func` {} is not a `Functional` instance'
+                            ''.format(funcB))
+
+        self.__functional_left = funcA
+        self.__functional_right = funcB
+
+        super().__init__(space=funcA.domain,
+                         linear=False,
+                         grad_lipschitz=np.nan)
+
+    @property
+    def functional_left(self):
+        """Left functional."""
+        return self.__functional_left
+
+    @property
+    def functional_right(self):
+        """Right functional."""
+        return self.__functional_right
+
+    @property
+    def convex_conj(self):
+        """Convex conjugate functional of the functional.
+
+        Notes
+        -----
+        The convex conjugate of the infimal convolution
+        .. math::
+            h(x) = inf_y f(x-y) + g(y)
+        is the sum of it:
+        .. math::
+            h^*(x) = f^*(x) + g(x)
+        """
+        return self.functional_left.convex_conj + \
+            self.functional_right.convex_conj
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.functional_left,
+                                       self.functional_right)
+
+    def __str__(self):
+        """Return ``str(self)``."""
+        return '{}({}, {})'.format(self.__class__.__name__,
+                                   self.functional_left,
+                                   self.functional_right)
 
 
 class FunctionalQuadraticPerturb(Functional):
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -21,6 +21,7 @@ from odl.operator.default_ops import (IdentityOperator, ConstantOperator)
 from odl.solvers.nonsmooth import (proximal_arg_scaling, proximal_translation,
                                    proximal_quadratic_perturbation,
                                    proximal_const_func, proximal_convex_conj)
+from odl.util import (signature_string, indent_rows)
 
 
 __all__ = ('Functional', 'FunctionalLeftScalarMult',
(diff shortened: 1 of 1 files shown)


odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
                                           self.translation)
 
 
+class InfimalConvolution(Functional):
+
+    """The functional representing ``h(x) = inf_y f(x-y) + g(y)``."""
```
adler-j: ~~The ~~

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+
+        self.__functional_left = funcA
+        self.__functional_right = funcB
+
+        super().__init__(space=funcA.domain,
+                         linear=False,
+                         grad_lipschitz=np.nan)
+
+    @property
+    def functional_left(self):
```
adler-j: We call these `left`/`right` in other places. See

https://github.com/odlgroup/odl/blob/20f0ce38909f4b8ae77761ceefa353be2c8b1d6c/odl/operator/operator.py#L1146-L1154

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+        """Right functional."""
+        return self.__functional_right
+
+    @property
+    def convex_conj(self):
+        """Convex conjugate functional of the functional.
+
+        Notes
+        -----
+        The convex conjugate of the infimal convolution
```
adler-j: Missing blanklines before and after `.. math::`. I think the current version has problems displaying properly

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+        .. math::
+            h(x) = inf_y f(x-y) + g(y)
+        is the sum of it:
+        .. math::
+            h^*(x) = f^*(x) + g(x)
```
adler-j: Missing ^* on g

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+        """
+        return self.functional_left.convex_conj + \
```
adler-j: Prefer using parentheses

```python
return (self.functional_left.convex_conj +
        self.functional_right.convex_conj)
```

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+
+        Parameters
+        ----------
+        funcA : `Functional`
+            Function corresponding to ``f``.
+        funcB : `Functional`
+            Function corresponding to ``g``.
+
+        Example
+        -------
```
adler-j: A short line describing what's happening here would be great.

A tip to make it easier to understand would be to create functionals that are themselves convex conjugates, that way evaluating the convex conjugate of this will give an "obvious" answer.

odl/solvers/functional/functional.py
kohr-h: I think before `.. math::` a blank line is necessary, after not. Just check with Spyder.

odl/solvers/functional/functional.py
kohr-h: 👍 And use the same names for the parameters, makes it much more obvious.

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+            h^*(x) = f^*(x) + g(x)
+        """
+        return self.functional_left.convex_conj + \
+            self.functional_right.convex_conj
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r}, {!r})'.format(self.__class__.__name__,
+                                       self.functional_left,
+                                       self.functional_right)
```
kohr-h: We have a utility `signature_string` (in `odl.util`) for `repr`, I'd suggest
```python
posargs = [self.left, self.right]
inner_str = signature_string(posargs, [], sep=',\n')
return '{}(\n{}\n)'.format(self.__class__.__name__,
                           indent_rows(inner_str))
```
`indent_rows` is also in `odl.util`.
The advantage is that each argument starts on a new line, which makes it more readable in cases like this where the `repr` of these guys tend to be rather long themselves.

odl/solvers/functional/functional.py
```diff
@@ -822,6 +823,83 @@ def __str__(self):
...
+
+    def __str__(self):
+        """Return ``str(self)``."""
```
kohr-h: We tend to just `return repr(self)` nowadays. (Yes, much old code doesn't do that yet.)

odl/solvers/functional/functional.py
MatthiasJE: Done

odl/solvers/functional/functional.py
MatthiasJE: Done

odl/solvers/functional/functional.py
MatthiasJE: Where would I see how this looks like. I tried "Inspect current object" in Spyder's ipython which gives me some text but no nice mathematics:

> ==================
> InfimalConvolution
> ==================
> 
> Definition: InfimalConvolution(x) = inf_y f(x-y)
> Type: Function of odl.solvers.functional.functional module
> 
> ----
> 
> The functional representing ``h(x) = inf_y f(x-y) + g(y)``.

It also looks quite poor and the definition is even wrong: 

> Definition: InfimalConvolution(x) = inf_y f(x-y)

What am I doing wrong here?

MatthiasJE: OK, this is not how it looks for me ... GIThub did some conversion here ...

odl/solvers/functional/functional.py
MatthiasJE: Why is that?

MatthiasJE: In the current version, it is only one line. Do you still prefer parentheses?

odl/solvers/functional/functional.py
MatthiasJE: Makes sense. I was just following other ODL examples.

odl/solvers/functional/functional.py
MatthiasJE: Thanks

odl/solvers/functional/functional.py
MatthiasJE: Done. What is 

> Edit @kohr-h: fixed formatting

 supposed to mean?

MatthiasJE: My "the" was actually copy pasting from "FunctionalQuadraticPerturb". How would one correct these typos? Probably not with a pull request for every typo?

adler-j: One huge pull request fixing documentation would be something like christmas eve for us ;)

adler-j: > Done. What is 
> Edit @kohr-h: fixed formatting
> supposed to mean?

It means he edited my comment and by courtesy he also mentioned that he did it (so I won't wake up and think I'm crazy and that my comments magically change).

odl/solvers/functional/functional.py
adler-j: I guess the reason is that it is more readable (to me at least) and does not depend on "text level" tricks. But frankly the main reason is that "we've always done it like that", and stylistic consistency is quite important in keeping the library readable.

odl/solvers/functional/functional.py
adler-j: > I was just following other ODL examples.

Big mistake there :O

odl/solvers/functional/functional.py
MatthiasJE: OK, and 

> In the current version, it is only one line. Do you still prefer parentheses?
?

adler-j: No, if it fits on one line you should avoid them!

adler-j:
Improvements for later would perhaps be to add some more doc, i.e. improve the `Examples` doc and and a literature reference, no crisis there tho.

MatthiasJE:
If I want to make further changes, is it still possible with this pull request or would I need to make another one?

odl/solvers/functional/functional.py
kohr-h: > Done. What is
> 
>     Edit @kohr-h: fixed formatting
> 
> supposed to mean?

That was just regarding @adler-j's comment where the strikethrough was botched.

odl/solvers/functional/functional.py
kohr-h: The guideline here is [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) (emphasis added by me):

The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by **wrapping expressions in parentheses**. These should be used **in preference to
using a backslash for line continuation**.

Backslashes may still be appropriate at times. For example, long, multiple with-statements cannot use implicit continuation, so backslashes are acceptable: 

[...]

adler-j:
Maybe i was a bit rushed but yeah you need a new one for that!

odl/solvers/functional/functional.py
adler-j: Ok, so we actually have an authoritative source on this!

PR #1082, opened 2017-07-14
Issue 1079  update tgv

adler-j:


examples/solvers/chambolle_pock_tomography_tgv.py
```diff
@@ -47,12 +47,7 @@
 # --- Generate artificial data --- #
 
 # Create phantom
-def phantom_func(pts):
-    x, y = pts
-    sign = ((x < -10) | (x > 10) | (y < -10) | (y > 10)) * 3 - 2
-    return 0.5 * ((x / 20.0) * sign + 1)
-
-phantom = U.element(phantom_func)
+phantom = odl.phantom.tgv_phantom(U)
```
kohr-h: Can we go with the usual `reco_space` here? Calling `A` something else like `fwd_op` would also be nice.

odl/phantom/geometric.py
```diff
@@ -687,6 +687,9 @@ def tgv_phantom(space):
     space : `DiscreteLp`, 2 dimensional
         Discretized space in which the phantom is supposed to be created.
         Needs to be two-dimensional.
+    edge_smoothness : float, optional
+        Smoothing of the edges of the phantom.
+        Given in the width of the smoothing in units of pixels.
```
kohr-h: Looking at the implementation, perhaps "[...] phantom, given as smoothing width in units of minimum pixel size."

odl/phantom/geometric.py
```diff
@@ -687,6 +687,9 @@ def tgv_phantom(space):
     space : `DiscreteLp`, 2 dimensional
         Discretized space in which the phantom is supposed to be created.
         Needs to be two-dimensional.
+    edge_smoothness : float, optional
```
kohr-h: nonnegative

examples/solvers/chambolle_pock_tomography_tgv.py
adler-j: @MatthiasJE used this notation so i followed him. I say we leave this for now.

PR #1090, opened 2017-08-21
MAINT: Change entry_points to be lazy

adler-j:
This PR does two things:

* Changes `entry_points` to being lazy, meaning that we don't go looking for backends before it is really needed. This reduces import time noticably.

* Fixes the circular import problem in https://github.com/odlgroup/odlcuda/issues/19

### DIFF
--- a/odl/discr/discretization.py
+++ b/odl/discr/discretization.py
@@ -565,10 +565,6 @@ def dspace_type(space, impl, dtype=None):
         Space type selected after the space's field, the backend and
         the data type
     """
-    spacetype_map = {RealNumbers: fn_impl,
-                     ComplexNumbers: fn_impl,
-                     type(None): ntuples_impl}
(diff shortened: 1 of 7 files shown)


odl/discr/discretization.py
```diff
@@ -592,9 +592,9 @@ def dspace_type(space, impl, dtype=None):
         raise TypeError('non-scalar data type {!r} cannot be combined with '
                         'a `LinearSpace`'.format(dtype))
 
-    stype = spacetype_map[field_type].get(impl, None)
-
-    if stype is None:
+    try:
+        stype = spacetype_map[field_type](impl)
```
adler-j: In this case it is actually intentional, perhaps a comment would help.

adler-j:
Fixed comments. Merge after CI.

PR #1092, opened 2017-08-22
Issue1077 chambolle pock

MatthiasJE:
I renamed Chambolle-Pock to PDHG / primal dual hybrid gradient in many files. I hope I have not overlooked one.

Moreover, I changed the meaning of "gamma" in PDHG. This is now replaced with gamma_primal. In addition, one can do dual acceleration with gamma_dual.

There are still 3 small issues that should be solved before any merging.
- The documentation is updated, but I think there are two links that still contain the name "Chambolle-Pock" and I didnt know how to change that as I didnt know where exactly they are pointing to.
- The gamma_primal part is tested in a new example of L2-TV denoising (ROF). Here I would like to compute the objective function values over the iterations to show that it gets faster with gamma_primal > 0. How would I do that? Moreover, I wanted to compare different versions (assignments to functionals f and g). Somehow they converge to different objective function values (which they should not ...).
- A simple example where gamma_dual > 0 makes sense would be something with Huberized TV. Is this or something similar available in ODL?


adler-j:
Try having a look at the failing tests, I'll review.

doc/source/guide/primal_dual_hybrid_gradient_guide.rst
```diff
@@ -1,10 +1,10 @@
-.. _chambolle_pock_guide:
+.. _primal_dual_hybrid_gradient_guide:
 
 #####################
-Chambolle-Pock solver
+Primal-Dual Hybrid Gradient (PDHG) solver
 #####################
 
-The `chambolle_pock_solver` was introduced in 2011 by Chambolle and Pock in the paper `A first-order primal-dual algorithm for convex problems with applications to imaging
+The `primal_dual_hybrid_gradient_solver` was studied in 2011 by Chambolle and Pock in the paper `A first-order primal-dual algorithm for convex problems with applications to imaging
```
adler-j: I'd scrap `_solver` from the function name now. Too long as is.

Perhaps we could even go with `pdhg`?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,124 @@
+"""Total variation denoising using the primal-dual hybrid gradient algorithm.
```
adler-j: Not quite clear what this function adds? I thought we had something similar before.

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
```diff
@@ -0,0 +1,81 @@
...
+# Set up the functionals
+
+# Isotropic TV-regularization: l1-norm of grad(x)
+reg_param = 0.15
+l1_norm = reg_param * odl.solvers.L1Norm(gradient.range)
+
+# Make separable sum of functionals, order must correspond to the operator K
+f = l1_norm
+
+# Data fit with non-negativity constraint
```
adler-j: This could use a word on what is happening here.

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
```diff
@@ -0,0 +1,81 @@
...
+# Add noise
+image += np.random.normal(0, 0.1, shape)
+
+# Data of noisy image
+noisy = space.element(image)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
+
+# Matrix of operators
```
adler-j: No matrix to be seen here?

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
```diff
@@ -0,0 +1,81 @@
...
+# Rescale max to 1
+image /= image.max()
+
+# Discretized spaces
+space = odl.uniform_discr([0, 0], shape, shape)
+
+# Original image
+orig = space.element(image)
+
+# Add noise
```
adler-j: Comments like this add very little. What you do is obvious from the code here

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -196,14 +203,24 @@ def chambolle_pock_solver(x, f, g, L, tau, sigma, niter, **kwargs):
...
+
+    gamma_dual = kwargs.pop('gamma_dual', None)
+    if gamma_dual is not None:
+        gamma_dual, gamma_dual_in = float(gamma_dual), gamma_dual
+        if gamma_dual < 0:
+            raise ValueError('`gamma_dual` must be non-negative, got {}'
+                             ''.format(gamma_dual_in))
+
+    if gamma_primal is not None and gamma_dual is not None:
+        raise ValueError('Only one acceleration parameter can be used')
```
adler-j: Any reason for this? Also this should be mentioned in the function definition.

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
MatthiasJE: This is your or Holger's comment. I am happy to remove it :)

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
MatthiasJE: The current version of this file runs 3 algorithms for the ROF (L2^2 + TV) problem. 1) the algorithm of the old example 2) a variant that most people would use for this and 3) the 1/k^2 convergence version that makes use of strong convexity.

MatthiasJE: It also shows how to compare algorithms with ODL. Thereby this add some features that other examples don't have.

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
MatthiasJE: removed

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
MatthiasJE: This file is removed and the primal acceleration bit is on the algorithmic comparison

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
MatthiasJE: Done. Depending on which part is strongly convex, the step sizes get either multiplied or divided by theta. Both doesn't make sense.

examples/solvers/pdhg_denoising_ROF_primal_acceleration.py
kohr-h: Feel free to remove old cruft :-)

MatthiasJE:
All changes are done and Travis is happy. @adler-j, do you want to have another look? Maybe @kohr-h wants to add anything?

MatthiasJE:
Before, I forget. The third point I raised initially about dual acceleration is still not met. I think it would be great if one looked at Huberized ROF and compares two algorithms: 1) PDHG with dual acceleration and 2) linearly convergent PDHG.

I won't be able to do that this week as this needs some non-trivial coding I suppose.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+odl.solvers.pdhg(x_alg3, f, g, op, tau=tau, sigma=sigma, niter=niter,
+                 gamma_primal=strong_convexity, callback=callback)
+obj_alg3 = callback.callbacks[1].obj_function_values
+obj_ergodic_alg3 = callback.callbacks[1].obj_function_values_ergodic
+
+# %% Display results
+# show images
+plt.figure(0)
+ax1 = plt.subplot(231)
+ax1.imshow(orig, clim=[0, 1])
```
kohr-h: Usually one needs a conversion from ODL's `'xy'` axes to matplotlib's `'ij'` axes. Probably the images end up correctly oriented like this, but if you do, e.g., `orig.show()` the image will be rotated by 90 degrees.

When wrapping a raw array representing an image in `'ij'` convention, the conversion is `np.rot90(img, -1)`, and the other way around it is `np.rot90(odl_elem, 1)`.

It's up to you if you consider that useful here.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+plt.title('Function values')
+plt.legend()
+
+# show convergence rates
+plt.figure(2)
+plt.clf()
+obj_opt = min(obj_alg1 + obj_alg2 + obj_alg3)
+
+
+def rel_fun(x): return (np.array(x) - obj_opt)/(x[0] - obj_opt)
```
kohr-h: Function body on a new line.

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -22,14 +22,15 @@
...
 
 
 # TODO: add dual gap as convergence measure
 # TODO: diagonal preconditioning
 
-def chambolle_pock_solver(x, f, g, L, tau, sigma, niter, **kwargs):
-    """Chambolle-Pock algorithm for non-smooth convex optimization problems.
+def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
+    """Primal-dual hybrid gradient algorithm for non-smooth convex optimization
+    problems.
```
kohr-h: This should fit on one line (perhaps simply remove "problems")

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -39,12 +40,12 @@ def chambolle_pock_solver(x, f, g, L, tau, sigma, niter, **kwargs):
 
     where ``L`` is an operator and ``F`` and ``G`` are functionals.
```
kohr-h: The `F` and `G` references should all be lowercase, could you please fix that while at it?

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -74,12 +75,20 @@ def chambolle_pock_solver(x, f, g, L, tau, sigma, niter, **kwargs):
...
         Default: 1
-    gamma : non-negative float, optional
+    gamma_primal : non-negative float, optional
         Acceleration parameter. If not ``None``, it overrides ``theta`` and
         causes variable relaxation parameter and step sizes to be used,
-        with ``tau`` and ``sigma`` as initial values. Requires ``G`` or
-        ``F^*`` to be uniformly convex.
+        with ``tau`` and ``sigma`` as initial values. Requires ``G`` to be
+        strongly convex. ``gamma_primal`` is the strong convexity constant of
+        ``G``. Acceleration can either be done on the primal part or the dual
```
kohr-h: I don't quite understand this. It seems impossible that `gamma_primal` is both a derived property of `G` and a user choice.
Is the strong convexity constant maybe an upper bound for the choice of acceleration parameter?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+noisy = space.element(image)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
+
+# regularization parameter
+reg_param = 0.3
+
+# l2-squared data matching
+factr = 1./reg_param * 0.5
```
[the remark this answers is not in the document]
MatthiasJE: Neither Travis nor Spyder complained about pep8 here. Can you be more specific what you think is violating pep8?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+l1_norm = reg_param * odl.solvers.L1Norm(gradient.range)
+
+# characteristic function
+char_fun = odl.solvers.IndicatorNonnegativity(space)
+
+# define objective
+obj = l2_norm + l1_norm * gradient + char_fun
+
+# strong convexity of "g"
+strong_convexity = 1./reg_param
```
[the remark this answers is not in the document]
MatthiasJE: I suppose the same as above?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+  (ROF)    min_{x >= 0}  1/2 ||x - d||_2^2 + lam || |grad(x)| ||_1
+
+Where ``grad`` the spatial gradient and ``d`` is given noisy data.
+
+Algorithms 1 and 2 are two different assignments of the functional parts of ROF
+to the functions f and g of PDHG. Algorithm 3 improves upon algorithm 2 by
+making use of the strong convexity of the problem.
+
+For further details and a description of the solution method used, see
+:ref:`PDHG` in the ODL documentation.
```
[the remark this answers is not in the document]
MatthiasJE: Better now?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+class CallbackStore(odl.solvers.util.callback.SolverCallback):
+    def __init__(self):
+        self.iteration_count = 0
+        self.iteration_counts = []
+        self.ergodic_iterate = 0
+        self.obj_function_values = []
+        self.obj_function_values_ergodic = []
+
+    def __call__(self, x):
+        # Fill in proper calls to functionals here
```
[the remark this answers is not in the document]
MatthiasJE: Done

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
MatthiasJE: Done

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
MatthiasJE: done

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
MatthiasJE: True. In practice, one wants to have gamma as large as possible. Also the use of "strong convexity constant" is often not unique. Some authors refer to it as the largest constant such that property X holds, others as any constant such that X holds.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+
+# Discretized spaces
+space = odl.uniform_discr([0, 0], shape, shape)
+
+# Create space element of ground truth
+orig = space.element(image.copy())
+
+# Add noise and convert to space element
+image += np.random.normal(0, 0.1, shape)
+noisy = space.element(image)
```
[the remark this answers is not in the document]
MatthiasJE: Done

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
```diff
@@ -0,0 +1,221 @@
...
+# The operator norm of the gradient with forward differences is well-known
+op_norm = np.sqrt(8) + 1e-4
+
+tau = 1.0 / op_norm  # Step size for the primal variable
+sigma = 1.0 / op_norm  # Step size for the dual variable
+
+# Run algorithms 2 and 3
+x_alg2 = x_start.copy()
+callback.reset()
+callback.callbacks[1].reset()
```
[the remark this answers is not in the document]
MatthiasJE: Done.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
MatthiasJE: I would like to use img.show() but for this example, it is most important to create plot that are next to each other. Of course the user can drag them there but I really would not recommend that to anyone. As far as I know img.show() will always open a new figure, so it cannot be used here. Any ideas how to combine these two ideas?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
adler-j: Sounds good!

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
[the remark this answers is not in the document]
adler-j: I guess he's comming at `1./reg_param` missing spaces.

With that said, why not `0.5/reg_param`?

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
adler-j: Well you have a point there, and no, I don't have any idea on how to solve it. I guess this way is fine.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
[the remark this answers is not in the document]
kohr-h: Travis doesn't run the examples and thus also no PEP8 check. And indeed, Spyder doesn't complain.

`autopep8` seems to do a better (more complete job) here. It's available in `pip`, and if you run it with the `-d` flag it gives you a nice diff with suggested changes. You can also run with `-i` to get the changes in-place.

MatthiasJE:
Did the corrections. I would propose to merge it for now and we add the dual acceleration example at some other point. For instance when implementing Stochastic PDHG.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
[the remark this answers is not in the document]
kohr-h: Looking good to me.

examples/solvers/pdhg_denoising_ROF_algorithm_comparison.py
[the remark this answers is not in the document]
kohr-h: yes

kohr-h:
> Did the corrections. I would propose to merge it for now and we add the dual acceleration example at some other point. For instance when implementing Stochastic PDHG.

I agree. Could you open a new issue for this so we have it on the agenda?

MatthiasJE:
Done, see #1101

kohr-h:
I ran the examples and found some issues, PR coming up.

kohr-h:
See https://github.com/MatthiasJE/odl/pull/1

There's one issue left, and I don't know the source: The TGV tomography example craps out after 1 iteration. If you set `niter = 1` the result is still OK (not NaN), but the next iteration makes everything NaN.
Any idea where this comes from? I haven't checked against the old version of the example, so I can't say if this PR introduces the issue.

kohr-h:
> Interestingly, if I run the example the nan values sometimes occur and sometimes not. @kohr-h, can you confirm this or do you always get nan?

I'll check again, but if I remember correctly I ran it about 5 times (also varying `tau and `sigma`) and got NaN in the second iteration every time.

kohr-h:
For me it seems to reliably fail. The reason are apparently the derivatives: after 1 iteration I get this (the white parts are NaN values):
derivatives _part_0
derivatives _part_1

kohr-h:
But actually, on the `master` branch we have the same issue, so this PR is not the cause.

So IMO it's fine that we just merge this and make a new issue for the TGV example.

MatthiasJE:
So, what needs to be done here? There seems to be a conflict in two files. How do we resolve this? I am happy to do it if you tell me how.

adler-j:
Basically what you need to do is:

```
git checkout master
git pull
git checkout Issue1077_ChambollePock
git rebase master
```

Then you'll get some conflicts that you need to solve, and once that is done we can merge this.

adler-j:
So something went very wrong in the rebase here :O

MatthiasJE:
Can you be more specific? What is it what you don't like?

MatthiasJE:
What I did was pretty much what you said:

```
git checkout upstream/master
git pull
git checkout Issue1077_ChambollePock
git rebase upstream/master

```
I resolved the conflicts by choosing my files and deleting some others. Then continued with the rebase and pushed the final result.

kohr-h:
Seems to be somehow rebased on the wrong commit. There are old commits apparently added as new ones, which can be seen from the numerous items in history on this page.
You can undo the whole thing by running `git reflog`, look for the last thing before the `git rebase`, copy the hash of that commit and `git checkout <last_good_commit_hash>`. Then you can delete the botched  branch `git branch -D Issue1077_ChambollePock` and `git checkout -b Issue1077_ChambollePock` to make a good version again.
Then start from that commit again.

It might be that the first part
```
git checkout upstream/master
git pull
```
didn't do what you expected it to do. It's good that you rebase on `upstream/master`, not your local one, since that's also a source of error. My recommended workflow is like this:
```
git fetch upstream  # updates the remote branches
git rebase upstream/master
```
Simpler, less error-prone.

MatthiasJE:
I did a bit different strategy:
```
	$ git reset --hard 053f4b251fb408d1944acf97c107b1b0fe6cb78d
	$ git rebase upstream/master
```
which seems to have done exactly what was needed. Is that OK? Otherwise I can always undo this since I have not pushed these changes. While my branch seems to be exactly what I wanted, I am a bit puzzled that I did not receive conflicts.

I would now push my changes to the remote branch and thereby getting rid of the noise above. Is that fine?

kohr-h:
> I did a bit different strategy:
> 
> 	$ git reset --hard 053f4b251fb408d1944acf97c107b1b0fe6cb78d
> 	$ git rebase upstream/master
> 
> which seems to have done exactly what was needed. Is that OK? Otherwise I can always undo this since I have not pushed these changes. While my branch seems to be exactly what I wanted, I am a bit puzzled that I did not receive conflicts.
> 
> I would now push my changes to the remote branch and thereby getting rid of the noise above. Is that fine?

If that works for you, all good. It's possible that you didn't get conflicts because they could be auto-merged. If you check the messages during rebase and see something like `auto-merge` then that's it. Usually this happens when the same file has been edited upstream, but in a different place.

Just go ahead and push, there's nothing that can't be fixed 😉

MatthiasJE:
Yes, there were auto-merges. I double checked the changes of the current version of this pull request and it looks good to me. I could not find anything that was missing or incorrect.

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -92,38 +101,40 @@ def chambolle_pock_solver(x, f, g, L, tau, sigma, niter, **kwargs):
...
+    Convergence is only guaranteed if :math:`L` is linear, :math:`X, Y`
     are finite dimensional and the step lengths :math:`\\sigma` and
     :math:`\\tau` satisfy
 
     .. math::
-       \\tau \\sigma \|K\| < 1
+       \\tau \\sigma \|K\|^2 < 1
 
-    where :math:`\|K\|` is the operator norm of :math:`K`.
+    where :math:`\|K\|` is the operator norm of :math:`L`.
```
kohr-h: Should be `L` in the first one, too

MatthiasJE:
Done. I found even another "K" that should have been an "L". :)

kohr-h:
> Done. I found even another "K" that should have been an "L". :)

Perfect. Just that nasty PEP8 check blew up :-)

kohr-h:
Ok, I'll merge like this, no need to rebase on master again. Thanks @MatthiasJE!

adler-j:
Now this is breaking changes for some downstream stuff (e.g. all my machine learning) that needs to be fixed.

We need to have a look at that.

kohr-h:
Maybe you should use a [pinned git version of ODL in the requirements.txt](https://pip.pypa.io/en/stable/reference/pip_install/#git)? 😛

PR #1094, opened 2017-08-23
BUG: Fix bug with CallbackStore on objects without copy, closes #1093

### LINKED ISSUE #1093: callback store
There is an issue when storing scalar values with CallbackStore. I followed the example in CallbackStore(SolverCallback). I suppose the example would have given the same error if the callback was applied!
 
```python
import odl
norm_function = lambda x: x.norm()
callback = odl.solvers.CallbackStore() * norm_function
space = odl.rn(3)
x = space.one()
callback(x)
```

>   File "<ipython-input-1-f49024937eaf>", line 8, in <module>
>     callback(x)
> 
>   File "/home/ehrhardt/repositories/git_ODL/odl/solvers/util/callback.py", line 162, in __call__
>     self.callback(self.operator(result))
> 
>   File "/home/ehrhardt/repositories/git_ODL/odl/solvers/util/callback.py", line 240, in __call__
>     self.results.append(result.copy())
> 
> AttributeError: 'float' object has no attribute 'copy'

adler-j:


adler-j:
Well implementing `__copy__` is the pythonic way. Doing it your way would be to do check for `obj.add` every time before we do `a + b`. And there's nothing that guarantees that users will do `.copy()`, they might as well do `.clone()`, `.replicate()` or w/e. I'd say we support the "official" interface and go with it.

kohr-h:
You have a point there, although `__copy__` feels a bit borderline since the core Python language does not "understand" it, unlike `__add__`. It still requires importing a module, albeit from the standard library.

Anyway, let's go with it then, and let's never forget to implement `__copy__` 😃

PR #1096, opened 2017-08-23
BUG: fix nonuniform inner, norm and add tests, closes #1091

### LINKED ISSUE #1091: Inner product on a space with nonuniform discretization is not defined correctly
As the title reads: The inner product on spaces with nonuniform discretizations are not defined correctly.

The following code:
```python
import odl
import numpy as np

part = odl.nonuniform_partition([0,1,4], min_pt=0, max_pt=4)
fspace = odl.FunctionSpace(part.set)
discr = odl.DiscreteLp(fspace, part, odl.rn(part.size), interp='linear')
space = odl.DiscreteLp(fspace, part, discr)

x1 = space.element(np.ones(3))
x2 = space.element(np.ones(3))

x3 = space.element(np.ones(3))
x4 = space.element(np.arange(3))
inner1 = space.inner(x1, x2)
inner2 = space.inner(x3, x4)
```
Returns:
inner1 = 3, inner2 = 3
Where I would expect:
inner1 = 4, inner2 = 5

I suspect that the inner product on this space is simply:
```python
np.sum(x1*x2)
```

EDIT: @adler-j updated typesetting

kohr-h:


odl/test/discr/lp_discr_test.py
```diff
@@ -934,6 +934,40 @@ def test_power(fn_impl, power):
...
+    linear = discr.element(lambda x: x)
+
+    # Exact inner product is the integral from 0 to 5 of x, which is 12.5
+    exact_inner = 12.5
+    inner = one.inner(linear)
+    assert inner == pytest.approx(exact_inner)
+
+
+def test_norm_nonuniform():
+    """Check if inner products are correct in non-uniform discretizations."""
```
kohr-h: Will update this, then merge.

PR #1099, opened 2017-08-29
ENH/BUG: update of unweighted ray transform.

aringh:
If the domain of the ray transform is unweighted, the weighting in the range should not be `1.0` but it should return an unweighted space, i.e., the argument passed should be `None`.


aringh:
@adler-j commented that there is probably an adjoint which needs a similar update.

adler-j:
I checked: nothing with adjoint since this was in the base class

PR #1102, opened 2017-08-30
ENH/BUG: update in fbp to handle unweighted range of ray transform.

aringh:
The `range` of the ray transform "inherits" the weighting (i.e., it is unweighted) from the `domain` if this is unweighted (see commit 12344455 and df30402f4). However, in this case creating an FBP operator will not give an analytic inverse. It needs to be scaled. This is a first draft on this.

Thank you @adler-j for pointing it out.

### DIFF
--- a/odl/tomo/analytic/filtered_back_projection.py
+++ b/odl/tomo/analytic/filtered_back_projection.py
@@ -15,6 +15,7 @@ import numpy as np
 import scipy as sp
 from odl.discr import ResizingOperator
 from odl.trafos import FourierTransform, PYFFTW_AVAILABLE
+from odl.space.weighting import NoWeighting
 
 
 __all__ = ('fbp_op', 'fbp_filter_op', 'tam_danielson_window',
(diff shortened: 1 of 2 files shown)


adler-j:
This could really use a test, specifically here:

https://github.com/odlgroup/odl/blob/5c2130bab592cdc355f8f152746454a3f8b9a551/odl/test/largescale/tomo/analytic_slow_test.py

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -440,7 +441,12 @@ def fourier_filter(x):
     ramp_function = fourier.range.element(fourier_filter)
 
     # Create ramp filter via the convolution formula with fourier transforms
-    return fourier.inverse * ramp_function * fourier
+    if isinstance(ray_trafo.range.weighting, NoWeighting):
+        # Compensate for potentially unweighted range of the ray transform
+        weights = ray_trafo.range.cell_volume
```
adler-j: Would prefer if the ramp itself was weighted by this, that would improve performance

Issue #1098, opened 2017-08-25
Add as_theano_layer


adler-j:
Similar to the new functionality `as_tensorflow_layer` we should do the same for at least theano.

Some short tests show that this works:

```python

class ODLTheanoOp(theano.Op):
    __props__ = ()

    def __init__(self, operator):
        self.operator = operator

    def make_node(self, x):
        x = theano.tensor.as_tensor_variable(x)
        out_type = theano.tensor.TensorType(
            self.operator.range.dtype,
            [False] * len(self.operator.range.shape))
        return theano.Apply(self, [x], [out_type()])

    def perform(self, node, inputs, output_storage):
        x = inputs[0]
        z = output_storage[0]
        z[0] = np.asarray(self.operator(x))

    def infer_shape(self, node, i0_shapes):
        return [self.operator.range.shape]

    def grad(self, inputs, output_grads):
        try:
            dom_weight = self.operator.domain.weighting.const
        except AttributeError:
            dom_weight = 1.0

        try:
            ran_weight = self.operator.range.weighting.const
        except AttributeError:
            ran_weight = 1.0

        scale = dom_weight / ran_weight
        return [ODLTheanoOp(scale * self.operator.adjoint)(output_grads[0])]

    def R_op(self, inputs, eval_points):
        if eval_points[0] is None:
            return eval_points
        return self.grad(inputs, eval_points)
```

EDIT: Updated code

kohr-h:
Agreed. I have this lying around [here](https://github.com/kohr-h/misc/blob/master/dl/odl_op_as_theano_op.py). This works, similarly to your code.

The main problem is that this thing is of limited use, at least as far as I can see currently, in the context of Deep Learning. Theano itself is only responsible for the low-level symbolic/numeric computation stuff and the graph building and optimization. It has no high-level notions and relies on another library [blocks](https://github.com/mila-udem/blocks) that builds on top to deal with layers, inputs, outputs, and so on.
Perhaps there's a simpler way to do it, but it seems you need to wrap the Theano op again so `blocks` understands what it does.

adler-j:
While I agree that we need to support "more stuff", simply doing this helps for lots of stuff.

Myself, I'm currently looking into Bayesian inverse problems and pymc3 was recommended to me by @olivierverdier. There, in order to do hybrid monte-carlo we need full integration with theano. 

With that, we can do bayesian inverse problems quite easily. For example, this "snippet" performs fullybayesian inversion with a TV regularizer.

```python
[...]

PR #1103, opened 2017-08-30
ENH: add extraglobs option to run_doctests

kohr-h:
Tiny addition to doctests for the case extra modules are required, as discussed [here](https://github.com/adler-j/odl/pull/1#discussion_r136057920).


odl/util/testutils.py
```diff
@@ -686,8 +693,14 @@ def run_doctests(skip_if=False):
                               'interpreter if the doctests do not work.',
                               RuntimeWarning)
 
-    testmod(optionflags=optionflags,
-            extraglobs={'odl': odl, 'np': np})
+    if extraglobs is None:
```
adler-j: Weird logic here, why not:

```python
globs = {'odl': odl, 'np': np}
if extraglobs:
    globs.update(extraglobs_in)

testmod(optionflags=optionflags, extraglobs=globs)
```

kohr-h: Of course 😲

kohr-h:
> How would this work with pytest?

How wouldn't it? Apparently Travis has no issue with it, and I can't see where it should come from.

If you mean modules that have an extra module required for tests, okay, sure, that may not work (will check), but those tend to be in `contrib` anyway.
Anyhow, I'll see if there's a solution for that.

adler-j:
Someone screwed up a rebase :D

kohr-h:
> Someone screwed up a rebase :D

😇

kohr-h:
Turns out that both running doctests directly and through pytest doesn't require amending the namespace. The import of `theano` or whatever in the module is sufficient (at least for me).

I'll close this and change the stuff in #1098

adler-j:
Even if they are in contrib we even recommend in the doc:

> To test all of your code with pytest, run `pytest odl/contrib/mylib/`

PR #1109, opened 2017-09-01
ENH: pytorch integration

kohr-h:
~~Almost done, there's just an issue with stride 0 arrays breaking some stuff, the example is not quite complete, and tests are missing.~~ See TODO list

But this is the deep learning framework with the slimmest wrapper so far. Almost trivial.

TODOs:
- [x] Fix stride-0 issue
- [x] Optimize for linear operators
- [x] Finish example
- [x] Adjust for weighted spaces
- [x] Add a test

### DIFF
--- a/odl/contrib/pytorch/examples/basic.py
+++ b/odl/contrib/pytorch/examples/basic.py
@@ -1,0 +1,76 @@
+"""Demonstration of basic ODL->pytorch integration functionality."""
+
+import numpy as np
+import torch
+import odl
+from odl.contrib.pytorch import TorchOperator
+
+# --- Forward --- #
+
+# Define ODL operator
+matrix = np.array([[1, 2, 3],
+                   [4, 5, 6]], dtype=float)
+odl_op = odl.MatrixOperator(matrix)
+
+# Wrap as torch operator
+torch_op = TorchOperator(odl_op)
+
+# Define evaluation point and wrap into a variable. Mark as
+# `requires_gradient`, otherwise `backward()` doesn't do anything.
+# This is supported by the ODL wrapper.
+x = torch.DoubleTensor([1, 1, 1])
+x_var = torch.autograd.Variable(x, requires_grad=True)
+
+# Evaluate torch operator
+res_var = torch_op(x_var)
+
+# ODL result
+odl_res = odl_op(x.numpy())
+
+print('pytorch result: ', res_var.data.numpy())
+print('ODL result    : ', odl_res.asarray())
+
+# --- Gradient (backward) --- #
+
+# Define ODL cost functional
+odl_cost = odl.solvers.L2NormSquared(odl_op.range)
+
+# Wrap with torch
+torch_cost = TorchOperator(odl_cost)
+
+# Compute forward pass
+res_var = torch_cost(torch_op(x_var))
+
+# Populate gradients by backwards pass
+res_var.backward()
+
+# ODL result
+odl_grad = (odl_cost * odl_op).gradient(x.numpy())
+
+print('pytorch gradient: ', x_var.grad.data.numpy())
+print('ODL gradient    : ', odl_grad.asarray())
+
+# --- Gradient in general spaces --- #
+
+# Same steps as above
+
+# ODL part
+dom = odl.uniform_discr(0, 6, 3)  # weight 2.0
+odl_op = odl.MatrixOperator(matrix, domain=dom)
+odl_cost = odl.solvers.L2NormSquared(odl_op.range)
+odl_functional = odl_cost * odl_op
+
+# Torch part
+torch_op = TorchOperator(odl_op)
+torch_cost = TorchOperator(odl_cost)
+
+x = torch.ones((3,)).type(torch.DoubleTensor)
+x_var = torch.autograd.Variable(x, requires_grad=True)
+y_var = torch_op(x_var)
+res_var = torch_cost(y_var)
+res_var.backward()
+
+# These may be different, see https://github.com/odlgroup/odl/issues/1068
+print('pytorch gradient (weighted): ', x_var.grad.data.numpy())
+print('ODL gradient (weighted)    : ',
+      odl_functional.gradient(x.numpy()).asarray())
--- a/odl/contrib/pytorch/layer.py
+++ b/odl/contrib/pytorch/layer.py
@@ -1,0 +1,232 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Utilities for converting ODL operators to pytorch layers.
+
+This requires the ``torch`` module to be available via the ``pytorch``
+package, which can be installed through ``pip`` or ``conda``
+
+See `the pytorch installation guide
+<https://github.com/pytorch/pytorch#installation>`_.
+"""
+
+import numpy as np
+import torch
+
+__all__ = ('TorchOperator',)
+
+# TODO: ProductSpaceOperator as multi-input, multi-output construction?
+
+
+class TorchOperator(torch.autograd.Function):
+
+    """Wrap an ODL operator as a Torch autograd Function."""
+
+    def __init__(self, operator):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        operator : `Operator`
+            The ODL operator to be wrapped. For gradient computations to
+            work, ``operator.derivative(x).adjoint`` must be implemented.
+
+        Examples
+        --------
+        >>> matrix = np.array([[1, 0, 1],
+        ...                    [0, 1, 1]], dtype=float)
+        >>> odl_op = odl.MatrixOperator(matrix)
+        >>> torch_op = TorchOperator(odl_op)
+        >>> torch_op.operator is odl_op
+        True
+        """
+        self.operator = operator
+
+    def forward(self, input):
+        """Evaluate forward pass on the input.
+
+        Parameters
+        ----------
+        input : `torch.tensor._TensorBase`
+            Point at which to evaluate the operator.
+
+        Returns
+        -------
+        result : `torch.autograd.variable.Variable`
+            Variable holding the result of the evaluation.
+
+        Examples
+        --------
+        Perform a matrix multiplication:
+
+        >>> matrix = np.array([[1, 0, 1],
+        ...                    [0, 1, 1]], dtype=float)
+        >>> odl_op = odl.MatrixOperator(matrix)
+        >>> torch_op = TorchOperator(odl_op)
+        >>> x = torch.DoubleTensor([1, 2, 3])
+        >>> x_var = torch.autograd.Variable(x)
+        >>> torch_op(x_var)
+        Variable containing:
+         4
+         5
+        [torch.DoubleTensor of size 2]
+
+        Evaluate a functional, i.e., an operator with scalar output:
+
+        >>> odl_func = odl.solvers.L2NormSquared(odl.rn(3))
+        >>> torch_func = TorchOperator(odl_func)
+        >>> x = torch.DoubleTensor([1, 2, 3])
+        >>> x_var = torch.autograd.Variable(x)
+        >>> torch_func(x_var)
+        Variable containing:
+         14
+        [torch.DoubleTensor of size 1]
+        """
+        if not self.operator.is_linear:
+            # Only needed for nonlinear operators
+            self.save_for_backward(input)
+
+        input_arr = input.numpy()
+        if any(s == 0 for s in input_arr.strides):
+            # TODO: remove when Numpy issue #9165 is fixed
+            # https://github.com/numpy/numpy/pull/9177
+            input_arr = input_arr.copy()
+
+        op_result = self.operator(input_arr)
+        return torch.from_numpy(np.array(op_result, copy=False, ndmin=1))
+
+    def backward(self, grad_output):
+        """Apply the adjoint of the derivative at ``grad_output``.
+
+        This method is usually not called explicitly but as a part of the
+        ``cost.backward()`` pass of a backpropagation step.
+
+        Parameters
+        ----------
+        grad_output : `torch.tensor._TensorBase`
+            Tensor to which the Jacobian should be applied. See Notes
+            for details.
+
+        Returns
+        -------
+        result : `torch.autograd.variable.Variable`
+            Variable holding the result of applying the Jacobian to
+            ``grad_output``. See Notes for details.
+
+        Examples
+        --------
+        Compute the Jacobian adjoint of the matrix operator, which is the
+        operator of the transposed matrix. We compose with the ``sum``
+        functional to be able to evaluate ``grad``:
+
+        >>> matrix = np.array([[1, 0, 1],
+        ...                    [0, 1, 1]], dtype=float)
+        >>> odl_op = odl.MatrixOperator(matrix)
+        >>> torch_op = TorchOperator(odl_op)
+        >>> x = torch.DoubleTensor([1, 2, 3])
+        >>> x_var = torch.autograd.Variable(x, requires_grad=True)
+        >>> op_x_var = torch_op(x_var)
+        >>> cost = op_x_var.sum()
+        >>> cost.backward()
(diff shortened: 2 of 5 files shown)


odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,45 @@
...
+# Define ODL operator
+matrix = np.array([[1, 2, 3],
+                   [4, 5, 6]], dtype='float32')
+matrix_op = odl.MatrixOperator(matrix)
+
+# Wrap as torch operator
+torch_op = TorchOperator(matrix_op)
+
+# Define evaluation point and wrap into a variable, mark as requiring gradient
+x = torch.ones((3,)).type(torch.FloatTensor)
```
adler-j: There's no `dtype` argument we can use?

odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Define ODL operator
+matrix = np.array([[1, 2, 3],
+                   [4, 5, 6]], dtype='float32')
+matrix_op = odl.MatrixOperator(matrix)
+
+# Wrap as torch operator
+torch_op = TorchOperator(matrix_op)
+
+# Define evaluation point and wrap into a variable, mark as requiring gradient
```
adler-j: Mention that ODL "works well" with `requires_grad`

odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Populate gradients by backwards pass
+res_var.backward()
+
+# ODL result
+odl_functional = odl_cost * matrix_op
+odl_grad = odl_functional.gradient(x.numpy())
+
+
+# TODO: print stuff
```
adler-j: yupp

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,201 @@
...
+        ``backward`` boils down to computing ``[f'(x)^*(y)]``, where
+        ``grad_output == [y]`` and ``inputs == [x]``.
+        """
+        # TODO: scaling
+        input, = self.saved_variables
+        grad = None
+
+        if self.needs_input_grad[0]:
+            input_arr = input.data.numpy()
+            # TODO: something wrong here, getting strides=(0,) sometimes
```
adler-j: In what cases are you getting strides=(0,)?

odl/contrib/pytorch/examples/basic.py
kohr-h: Haven't seen one so far. They have some work to do on the convenience side. This is the way to produce a certain dtype that I've seen everywhere in their docs. Seems wasteful, but maybe the JIT will take care of that.

odl/contrib/pytorch/layer.py
kohr-h: When using the torch builtin `sum()` in particular, I guess. It may be a very memory-friendly way of storing an array of all same values, but it's not compatible with `matrix.dot()` using BLAS. Perhaps `MatrixOperator` needs a check for that.

kohr-h: In particular, it may be a feature, not a bug, and we have to deal with it.

kohr-h: It's a Numpy issue, and the fix is [here](https://github.com/numpy/numpy/pull/9177). In the meanwhile I guess we have to copy in that case. Or maybe always, since in-place mutation (like `arr *= 2`) yields wrong results.

kohr-h: For the records, the `broadcast_to` function also creates a zero-stride array:
```python
>>> np.broadcast_to(np.ones((3, 1)), (3, 4)).strides
(8, 0)
```

adler-j: Overall 0 stride arrays are not really valid datacontainers for ODL since they don't work with in place operations. Perhaps we should disallow them.

kohr-h: True. In the cases when Numpy produces them (like `broadcast_to`), the arrays are usually read-only, for a good reason. So we should stay away from them.

odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,75 @@
...
+odl_cost = odl.solvers.L2NormSquared(odl_op.range)
+odl_functional = odl_cost * odl_op
+torch_cost = TorchOperator(odl_cost)
+
+x = torch.ones((3,)).type(torch.DoubleTensor)
+x_var = torch.autograd.Variable(x, requires_grad=True)
+y_var = torch_op(x_var)
+res_var = torch_cost(y_var)
+res_var.backward()
+grad = x_var.grad
```
adler-j: Move this into the print statement, e.g. `x_var.grad.data.numpy()` (this stuff is dense as is)

odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,75 @@
...
+
+print('pytorch gradient: ', grad.data.numpy())
+print('ODL gradient    : ', odl_grad.asarray())
+
+# --- Gradient in weighted space --- #
+
+# Same steps as above
+dom = odl.uniform_discr(0, 6, 3)  # weight 2.0
+odl_op = odl.MatrixOperator(matrix, domain=dom)
+torch_op = TorchOperator(odl_op)
```
adler-j: Try not to mix the ODL and torch code, e.g. first do odl stuff, then torch

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,245 @@
...
+``pytorch`` package available throught ``pip`` or ``conda``.
+"""
+
+import numpy as np
+import torch
+
+__all__ = ('TorchOperator',)
+
+# TODO: ProductSpaceOperator as multi-input, multi-output construction?
+# TODO: fix issue with stride-0 array
```
adler-j: Solve this or add a check

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,245 @@
...
+        Variable containing:
+         14
+        [torch.DoubleTensor of size 1]
+        """
+        if not self.operator.is_linear:
+            # Only needed for nonlinear operators
+            self.save_for_backward(input)
+
+        input_arr = input.numpy()
+        if all(s == 0 for s in input_arr.strides):
```
adler-j: I guess this will be a problem if there are at least 2 zero strides?

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,242 @@
...
+            # TODO: remove when Numpy issue #9165 is fixed
+            # https://github.com/numpy/numpy/pull/9177
+            input_arr = input_arr.copy()
+
+        op_result = self.operator(input_arr)
+        return torch.from_numpy(np.array(op_result, copy=False, ndmin=1))
+
+    def backward(self, grad_output):
+        """Apply the adjoint of the Jacobian at ``grad_output``.
```
adler-j: dislike use of Jacobian, go with derivative?

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,242 @@
...
+
+        This method is usually not called directly but as a part of the
```
adler-j: directly -> explicitly

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,242 @@
...
+            Variable holding the result of applying the Jacobian to
+            ``grad_output``. See Notes for details.
+
+        Examples
+        --------
+        Compute the Jacobian adjoint of the matrix operator, which is the
+        operator of the transposed matrix. We compose with the ``sum``
+        functional to be able to evaluate ``grad``:
+
+        >>> space = odl.rn(3)
```
adler-j: This stuff is DENSE for a doctest. Is there no way to make a more simple example?

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,242 @@
...
+        >>> matrix = np.array([[1, 0, 1],
+        ...                    [0, 1, 1]], dtype=float)
+        >>> op = odl.MatrixOperator(matrix, domain=space)
+        >>> matrix_op = TorchOperator(op)
+        >>> x_elem = space.element([1, 2, 3])
+        >>> x = torch.from_numpy(x_elem.asarray())
```
adler-j: is `as_array` needed here?

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,241 @@
...
+        Notes
+        -----
+        This method applies the contribution of this node, i.e., the
+        transpose of the Jacobian of its outputs with respect to its inputs,
+        to the gradients of some cost function with respect to the outputs
+        of this node.
+
+        Example: Assume that this node computes :math:`x \mapsto f(x)`, and
+        its output is connected to a cost function that computes
+        :math:`y --> C(y)`. Here, :math:`x`, :math:`y` and :math:`f(x)` are
```
adler-j: just go for x -> C(f(x)), easier to follow

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,241 @@
...
+
+        Then, the parameter ``grad_output`` contains a single tensor
+        variable ``y`` that stands for :math:`\\nabla C(f(x))`. Thus,
+        ``backward`` boils down to computing ``[f'(x)^*(y)]``, where
+        ``grad_output == [y]`` and ``inputs == [x]``.
+        """
+        if not self.operator.is_linear:
+            input, = self.saved_variables
+            input_arr = input.data.numpy()
+            if all(s == 0 for s in input_arr.strides):
```
adler-j: See above comment on at least 2 zero strides being an issue

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,241 @@
...
+            f.derivative(x).adjoint(C.gradient(f(x))).
+
+        Then, the parameter ``grad_output`` contains a single tensor
+        variable ``y`` that stands for :math:`\\nabla C(f(x))`. Thus,
+        ``backward`` boils down to computing ``[f'(x)^*(y)]``, where
+        ``grad_output == [y]`` and ``inputs == [x]``.
+        """
+        if not self.operator.is_linear:
+            input, = self.saved_variables
+            input_arr = input.data.numpy()
```
adler-j:
```python
input_arr = self.saved_variables[0].data.numpy()
```

is perfectly OK imo. The magic `,` on the line above is weird and trivial to miss

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,241 @@
...
+            if all(s == 0 for s in input_arr.strides):
+                # TODO: remove when Numpy issue #9165 is fixed
+                # https://github.com/numpy/numpy/pull/9177
+                input_arr = input_arr.copy()
+
+        grad = None
```
adler-j: ?

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,241 @@
...
+        try:
+            ran_weight = self.operator.range.weighting.const
+        except AttributeError:
+            ran_weight = 1.0
+
+        scaling = dom_weight / ran_weight
+
+        if self.needs_input_grad[0]:
+            grad_output_arr = grad_output.numpy()
+            if all(s == 0 for s in grad_output_arr.strides):
```
adler-j: At this point we should perhaps add a utility for this

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,241 @@
...
+                # TODO: remove when Numpy issue #9165 is fixed
+                # https://github.com/numpy/numpy/pull/9177
+                grad_output_arr = grad_output_arr.copy()
+
+            if self.operator.is_linear:
```
adler-j: This if statement does nothing. There is already a shortcut in `operator.derivative`.

If you truly want to have a performance gain, you would put this if statement around the creation of the array `input_arr`, which is not needed.

odl/contrib/pytorch/examples/basic.py
kohr-h: Will untangle

odl/contrib/pytorch/layer.py
kohr-h: Obviously, it should be `any`

odl/contrib/pytorch/layer.py
kohr-h: Well technically it's correct, the question is more whether we use ODL or torch lingo. No strong opinion here.

odl/contrib/pytorch/layer.py
kohr-h: I'll see what I can do.

odl/contrib/pytorch/layer.py
kohr-h: Good question, haven't tried.

odl/contrib/pytorch/layer.py
kohr-h: OK, I guess the check that there is only one element is silly anyway.

odl/contrib/pytorch/layer.py
kohr-h: Initialization for the case that `self.needs_input_grad[0]` is `False`. It's like that in their guidelines.

odl/contrib/pytorch/layer.py
kohr-h: That's what I'm doing. The input is saved for usage here only if the operator is non-linear, so I have it only in that case and need the distinction also here.

odl/contrib/pytorch/layer.py
kohr-h: One shortcut: `x = torch.DoubleTensor([1, 2, 3])` instead of the two current lines.

odl/contrib/pytorch/layer.py
kohr-h: Doesn't work. But I'll remove the line anyway.

kohr-h:
But I'll squash the commits into one before doing that.

odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Define ODL cost functional
+odl_cost = odl.solvers.L2NormSquared(odl_op.range)
+
+# Wrap with torch
+torch_cost = TorchOperator(odl_cost)
+
+# Compute forward pass
+y_var = torch_op(x_var)
+res_var = torch_cost(y_var)
```
adler-j: `res_var = torch_cost(torch_op(x_var))` is perfectly fine

odl/contrib/pytorch/examples/basic.py
```diff
@@ -0,0 +1,77 @@
...
+                   [4, 5, 6]], dtype=float)
+odl_op = odl.MatrixOperator(matrix)
+
+# Wrap as torch operator
+torch_op = TorchOperator(odl_op)
+
+# Define evaluation point and wrap into a variable. Mark as
+# `requires_gradient`, otherwise `backward()` doesn't do anything.
+# This is supported by the ODL wrapper.
+x = torch.ones((3,)).type(torch.DoubleTensor)
```
adler-j: `x = torch.DoubleTensor([1, 1, 1])` should work, no?

odl/contrib/pytorch/layer.py
```diff
@@ -0,0 +1,232 @@
...
+            f.derivative(x).adjoint(C.gradient(f(x))).
+
+        Hence, the parameter ``grad_output`` is a tensor variable containing
+        :math:`y = \\nabla C(f(x))`. Then, ``backward`` boils down to
+        computing ``[f'(x)^*(y)]`` using the input ``x`` stored during
+        the previous `forward` pass.
+        """
+        if not self.operator.is_linear:
+            input_arr = self.saved_variables[0].data.numpy()
+            if any(s == 0 for s in input_arr.strides):
```
adler-j: Does it not suffice that 2 of them are zero for us to get issues?

odl/contrib/pytorch/examples/basic.py
kohr-h: Of course

odl/contrib/pytorch/layer.py
kohr-h: Already 1 zero is too much, at least potentially. I'd rather stay away from those guys completely (note the `any`).

adler-j: Ah, you are correct.

PR #1110, opened 2017-09-04
WIP: Add indexing to geometries

adler-j:
Before I spend too long on this, I could use some feedback on the interface. Does it look OK?

@kohr-h and @MatthiasJE

### DIFF
--- a/examples/solvers/kaczmarz_tomography.py
+++ b/examples/solvers/kaczmarz_tomography.py
@@ -1,0 +1,66 @@
+"""Tomography using the `kaczmarz` solver.
+
+Solves the inverse problem
+
+    A(x) = g
+
+Where ``A`` is a parallel beam forward projector, ``x`` the result and
+ ``g`` is given noisy data.
+
+In order to solve this using `kaczmarz`s method, the operator is split into
+several sub-operators (each representing a subset of the angles).
+"""
+
+import numpy as np
+import odl
+
+
+# --- Set up the forward operator (ray transform) --- #
+
+
+# Reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 300 samples per dimension.
+space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+# Angles: uniformly spaced, n = 360, min = 0, max = pi
+angle_partition = odl.uniform_partition(0, np.pi, 360)
+
+# Detector: uniformly sampled, n = 300, min = -30, max = 30
+detector_partition = odl.uniform_partition(-30, 30, 300)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# Create the forward operators
+n = 10  # number of sub-operators
+ray_trafos = [odl.tomo.RayTransform(space, geometry[i::n]) for i in range(n)]
+
+# --- Generate artificial data --- #
+
+
+# Create phantom
+phantom = odl.phantom.shepp_logan(space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+data = [ray_trafo(phantom) for ray_trafo in ray_trafos]
+noisy_data = [d + odl.phantom.white_noise(d.space) * np.mean(d) * 0.1
+              for d in data]
+
+omega = [odl.power_method_opnorm(ray_trafo) ** (-2)
+         for ray_trafo in ray_trafos]
+
+# Optionally pass callback to the solver to display intermediate results
+callback = (odl.solvers.CallbackPrintIteration() &
+            odl.solvers.CallbackShow())
+
+# Choose a starting point
+x = ray_trafo.domain.zero()
+
+# Run the algorithm
+odl.solvers.kaczmarz(
+    ray_trafos, x, noisy_data, niter=20, omega=omega, callback=callback)
+
+# Display images
+phantom.show(title='original image')
+data.show(title='sinogram')
+x.show(title='reconstructed image', force_show=True)
--- a/examples/solvers/kaczmarz_tomography.py
+++ b/examples/solvers/kaczmarz_tomography.py
@@ -66,7 +66,7 @@ x = space.zero()
 # Note that using only 2 iterations still gives a decent reconstruction.
 odl.solvers.kaczmarz(
     ray_trafos, x, data, niter=2, omega=omega,
-    callback=callback, callback_call='inner')
+    callback=callback, callback_loop='inner')
 
 # Display images
 phantom.show(title='original image')
--- a/examples/solvers/kaczmarz_tomography.py
+++ b/examples/solvers/kaczmarz_tomography.py
@@ -5,13 +5,13 @@ Solves the inverse problem
     A(x) = g
 
 Where ``A`` is a parallel beam forward projector, ``x`` the result and
- ``g`` is given noisy data.
+ ``g`` is given data.
 
 In order to solve this using `kaczmarz`s method, the operator is split into
-several sub-operators (each representing a subset of the angles).
+several sub-operators (each representing a subset of the angles and detector
+points). This allows a faster solution.
 """
 
-import numpy as np
 import odl
 
 
@@ -24,16 +24,24 @@ space = odl.uniform_discr(
     min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='float32')
 
 # Make a parallel beam geometry with flat detector
-# Angles: uniformly spaced, n = 360, min = 0, max = pi
-angle_partition = odl.uniform_partition(0, np.pi, 360)
+geometry = odl.tomo.cone_beam_geometry(space,
+                                       src_radius=40, det_radius=40,
+                                       num_angles=360, det_shape=360)
 
-# Detector: uniformly sampled, n = 300, min = -30, max = 30
-detector_partition = odl.uniform_partition(-30, 30, 300)
-geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+# Here we split the geometry according to both angular subsets and
+# detector subsets.
 
-# Create the forward operators
-n = 10  # number of sub-operators
-ray_trafos = [odl.tomo.RayTransform(space, geometry[i::n]) for i in range(n)]
+# For practical applications these choices should be fine tuned,
+# these values are selected to give an illustrative visualization.
+n = 20
+m = 1
+ns = 360 // n
+ms = 360 // m
+
+# Create the forward operators.
+ray_trafos = [odl.tomo.RayTransform(space, geometry[i * ns:(i + 1) * ns,
+                                                    j * ms:(j + 1) * ms])
+              for i in range(n) for j in range(m)]
 
 # --- Generate artificial data --- #
 
@@ -43,8 +51,6 @@ phantom = odl.phantom.shepp_logan(space, modified=True)
 
 # Create sinogram of forward projected phantom with noise
 data = [ray_trafo(phantom) for ray_trafo in ray_trafos]
-noisy_data = [d + odl.phantom.white_noise(d.space) * np.mean(d) * 0.1
-              for d in data]
 
 omega = [odl.power_method_opnorm(ray_trafo) ** (-2)
          for ray_trafo in ray_trafos]
@@ -54,13 +60,14 @@ callback = (odl.solvers.CallbackPrintIteration() &
             odl.solvers.CallbackShow())
 
 # Choose a starting point
-x = ray_trafo.domain.zero()
+x = space.zero()
 
-# Run the algorithm
+# Run the algorithm, call the callback in each iteration for visualization.
+# Note that using only 2 iterations still gives a decent reconstruction.
 odl.solvers.kaczmarz(
-    ray_trafos, x, noisy_data, niter=20, omega=omega, callback=callback)
+    ray_trafos, x, data, niter=2, omega=omega,
+    callback=callback, callback_call='inner')
 
 # Display images
 phantom.show(title='original image')
-data.show(title='sinogram')
 x.show(title='reconstructed image', force_show=True)
--- a/examples/solvers/kaczmarz_tomography.py
+++ b/examples/solvers/kaczmarz_tomography.py
@@ -4,10 +4,10 @@ Solves the inverse problem
 
     A(x) = g
 
-Where ``A`` is a parallel beam forward projector, ``x`` the result and
+Where ``A`` is a fan (cone) beam forward projector, ``x`` the result and
  ``g`` is given data.
 
-In order to solve this using `kaczmarz`s method, the operator is split into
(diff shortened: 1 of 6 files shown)


examples/solvers/kaczmarz_tomography.py
```diff
@@ -0,0 +1,63 @@
...
+# Angles: uniformly spaced, n = 360, min = 0, max = pi
+angle_partition = odl.uniform_partition(0, np.pi, 360)
+
+# Detector: uniformly sampled, n = 300, min = -30, max = 30
+detector_partition = odl.uniform_partition(-30, 30, 300)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# Create the forward operators
+n = 10  # number of sub-operators
+ray_trafos = [odl.tomo.RayTransform(space, geometry[i::n]) for i in range(n)]
```
adler-j: Basically the magic happens in this line

odl/tomo/geometry/parallel.py
```diff
@@ -472,6 +472,42 @@ def __repr__(self):
...
+        """Return self[slc]
+
+        Examples
+        --------
+        Extract sub-geometry:
+
+        >>> apart = odl.uniform_partition(0, np.pi, 4)
+        >>> dpart = odl.uniform_partition(-1, 1, 20)
+        >>> geom = Parallel2dGeometry(apart, dpart)
+        >>> geom[::2, :]
```
adler-j: We need a discussion on what the expected behavior is here. I'm not totally convinced what we see here is correct.

Specifically I would also be OK with the `min_pt` and `max_pt` staying un-changed.

With that said, i guess any such change should go into `Partition.__getitem__`?

odl/tomo/geometry/parallel.py
```diff
@@ -1094,7 +1130,7 @@ def parallel_beam_geometry(space, num_angles=None, det_shape=None):
         If ``space`` is 3d, return a `Parallel3dAxisGeometry`.
 
     Examples
-    --------
+    --------testmod
```
adler-j: Obvious misstake

odl/tomo/geometry/parallel.py
kohr-h: Without reading any details, the basic expectation I have is that `geom[i]` gives the geometry for the i-th angle, leaving the rest unchanged. The handling of weightings is a discussion point, with three possibilities, namely (1) take cell volume, (2) take angle interval length, or (3) take 1.0.
I have a preference for (1) because it makes the product space operator with 1 angle per operator equivalent to the original operator.
Then, the remaining indices should index the detector, as defined by the partition.

kohr-h: Looking at the example, I think that one should raise an error because we cannot "rip holes" into intervals. For a ray transform of each second angle, I think one should use `geoms = list(geometry)` and then `ops = [RayTransform(geom, ...) for geom in geoms]` and finally `op = BroadcastOperator(*ops)`.

I think there are some utilities like that in [this commit](https://github.com/kohr-h/odl/commit/fea1f92bd0bb7b2b0e6c279242f5ef95e9048e97).

adler-j: While i totally agree that `geometry[i]` should return the i:th angle so to say, weighting would then implicitly be (1).

Anyway regarding your approach of creating a single projector for each geometry, it simply is not performance-wise a feasible solution. For example, doing all at once (`'all'`) vs doing them all at once (`'single'`) for some typical geometry gives this timing:

                           all :      1.238 
                        single :     30.094
                        
So proper vectorization is a must, and that approach is not feasible.

Anyway, for me the only reasonable thing here is that

```python
geometry[...].partition == geometry.partition[...]
```

so I guess this discussion might as well be about what `partition[...]` should mean.

With that said, there are basically some options for what this code should do

```python
>>> partition = odl.uniform_partition(0, 10, 10)
>>> partition[::2]
```

* Current:
  ```python
  nonuniform_partition(
      [0.5, 2.5, 4.5, 6.5, 8.5],
      min_pt=0.0, max_pt=9.0
  )
  ```
  
* Keep full domain:
  ```python
  nonuniform_partition(
      [0.5, 2.5, 4.5, 6.5, 8.5],
      min_pt=0.0, max_pt=10.0
  )
  ```
  
* Centered:  
  ```python
  uniform_partition(0, 10, 5)
  ```
  
* Throw exception
  ```python
  IndexException(...)
  ```

I think I'd personally be in favor of the second option, e.g. keeping the far edges and selecting the inner points according to the indices.

adler-j: Perhaps @ozanoktem has some input on this?

kohr-h: Well, the inefficiency is not completely our fault. Have you tried profiling the angle-by-angle code vs. bulk? I think I remember @wjp mention possible speedups in ASTRA for this case.

Anyway, assuming that we need a way of doing this in bulk, I'd definitely vote against option 3 since it moves grid points. That's a no-go IMO.

Regarding 1 vs. 2, I think we should look at the weights that the two suggest (keeping also `[0, 2*pi]` with periodicity in mind.
Let's check a 1D case where we partition an interval `[a, b]` with nodes `a <= x_1 < ... < x_n <= b`. When using nearest interpolation, the "inner" weights (`i = 2, ..., n-1`) are computed as 

    w_i = (x_{i+1} - x_{i-1}) / 2

For the left boundary `x_1` we get

    w_1 = (x_1 + x_2) / 2 - a

for the right boundary

    w_{n} = b - (x_{n-1} + x_n) / 2

If `a = x_1`, the left weight becomes `w_1 = (x_2 - x_1) / 2`, i.e., half the weight if the spacing in uniform.
Also, for periodic functions (`f_1 = f_n` implicitly, and `x_1 = a, x_n = b`) the combined boundary weight is

    w_1^* = (x_n - x_{n-1}) / 2 + (x_2 - x_1) / 2,

which is just `h/2 + h/2 = h` for uniform sampling.

Let's check that with the example above: 
```python
>>> partition = odl.uniform_partition(0, 10, 10)
>>> partition.grid.coord_vectors[0]  # grid points
array([ 0.5,  1.5,  2.5,  3.5,  4.5,  5.5,  6.5,  7.5,  8.5,  9.5])
>>> partition.cell_sizes_vecs[0]  # the w_i
array([ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.])
```

Option 1:
```python
>>> subpart = odl.nonuniform_partition(
...     [0.5, 2.5, 4.5, 6.5, 8.5],
...     min_pt=0.0, max_pt=9.0
... )
>>> subpart.grid.coord_vectors[0]  # all good
array([ 0.5,  2.5,  4.5,  6.5,  8.5])
>>> subpart.cell_sizes_vecs[0]  # not quite as good
array([ 1.5,  2. ,  2. ,  2. ,  1.5])
```

Option 2:
```python
>>> subpart = odl.nonuniform_partition(
...     [0.5, 2.5, 4.5, 6.5, 8.5],
...     min_pt=0.0, max_pt=10.0
... )
>>> subpart.grid.coord_vectors[0]  # all good
array([ 0.5,  2.5,  4.5,  6.5,  8.5])
>>> subpart.cell_sizes_vecs[0]  # better
array([ 1.5,  2. ,  2. ,  2. ,  2.5])
```

Just for the records, also Option 3:
```python
>>> subpart = odl.uniform_partition(0, 10, 5)
>>> subpart.grid.coord_vectors[0]  # Nooooooo!
array([ 1.,  3.,  5.,  7.,  9.])
>>> subpart.cell_sizes_vecs[0]  # best, but worthless
array([ 2.,  2.,  2.,  2.,  2.])
```

Why is option 2 better? The boundary weights are not perfect, but at least they *sum up to the right value*, meaning that periodic functions will be weighted correctly.
There's little to be done about the doubling of the weights overall since the subintervals actually double in size. So a convenience function that, e.g., splits the `RayTransform` should take that weighting change into account. The goal should be that stacking together angles in an interleaved way should produce the same result after backprojecting as the original set of angles.

adler-j: I agree with what you write here and I think the conculsion (as per this PR) is that we go for option 2 for now (I'll update `partition.__getitem__`), specifically this means that we use the implementation as is for the geometries.

adler-j:
Implemented according to comments. Still not 100% happy about cell sizes being off, but this can be fixed later (and is "isolated" to partitions) further, given current backends this does exactly what we'd expect.

kohr-h:
> Implemented according to comments. Still not 100% happy about cell sizes being off, but this can be fixed later (and is "isolated" to partitions) further, given current backends this does exactly what we'd expect.

How about doing the "obviously right" thing for indexing without "holes", and using the discussed thing only if the index is a slice with stride > 1 or an index list?

adler-j:
Not quite sure how that differs from the current implementation?

kohr-h:
> Not quite sure how that differs from the current implementation?

Probably not at all. I was simply wondering what happens if you do `geometry[:2, ...]`, would the angle max be the old one or would you get the right boundary of the second cell.

adler-j:
It would be the right boundary of the second cell

adler-j:
The change here is only for cases like `geometry[:2:2]` which would previously give the right boundary of the first cell (since the stride "skips" the second cell).

kohr-h:
> It would be the right boundary of the second cell

Then there is no issue.

adler-j:
Ready for final review.

odl/discr/partition.py
```diff
@@ -561,6 +572,12 @@ def __getitem__(self, indices):
         for cvec, idx in zip(self.cell_boundary_vecs, indices):
             # Determine the subinterval min_pt and max_pt vectors. Take the
             # first min_pt as new min_pt and the last max_pt as new max_pt.
+            if isinstance(idx, slice):
+                # Only use the slice to extract min and max without using
+                # the step size. This is in order for expressions like
+                # self[::2] to not change the maximum.
```
kohr-h: Good comment

odl/solvers/iterative/iterative.py
```diff
@@ -385,8 +386,8 @@ def gauss_newton(op, x, rhs, niter, zero_seq=exp_zero_seq(2.0),
             callback(x)
 
 
-def kaczmarz(ops, x, rhs, niter, omega=1, projection=None,
-             callback=None):
+def kaczmarz(ops, x, rhs, niter, omega=1, projection=None, random=False,
+             callback=None, callback_call='outer'):
```
kohr-h: `callback_call` doesn't really make clear what `'outer'` would mean, how about `callback_loop`?

odl/solvers/iterative/iterative.py
```diff
@@ -416,8 +417,13 @@ def kaczmarz(ops, x, rhs, niter, omega=1, projection=None,
         Function that can be used to modify the iterates in each iteration,
         for example enforcing positivity. The function should take one
         argument and modify it in-place.
+    random : bool, optional
+        If `True`, the order of the operators is randomized in each iteration.
     callback : callable, optional
         Object executing code per iteration, e.g. plotting each iterate.
+    callback_call : {'inner', 'outer'}
+        If the callback should be called in the inner or outer loop.
```
kohr-h: If -> Whether

odl/tomo/geometry/conebeam.py
```diff
@@ -474,6 +474,44 @@ def __repr__(self):
...
+
+        This is defined by::
+
+            self[indices].partition == self.partition[indices]
+
+        where all other parameters are the same.
+
+        Examples
+        --------
+        Extract sub-geometry:
```
kohr-h: More specific, using something like "every second angle".

odl/tomo/geometry/conebeam.py
```diff
@@ -1002,6 +1040,46 @@ def __repr__(self):
...
```
kohr-h: See above

odl/tomo/geometry/parallel.py
```diff
@@ -472,6 +472,40 @@ def __repr__(self):
...
+
+        This is defined by::
+
+            self[indices].partition == self.partition[indices]
+
+        where all other parameters are the same.
+
+        Examples
+        --------
+        Extract sub-geometry:
```
kohr-h: Above

odl/tomo/geometry/parallel.py
```diff
@@ -1059,6 +1093,41 @@ def __repr__(self):
...
```
kohr-h: Above

examples/solvers/kaczmarz_tomography.py
```diff
@@ -0,0 +1,73 @@
+"""Tomography using the `kaczmarz` solver.
+
+Solves the inverse problem
+
+    A(x) = g
+
+Where ``A`` is a parallel beam forward projector, ``x`` the result and
+ ``g`` is given data.
+
+In order to solve this using `kaczmarz`s method, the operator is split into
```
kohr-h: I think this link won't work, you need a `'` after the closing tick.

examples/solvers/kaczmarz_tomography.py
```diff
@@ -0,0 +1,73 @@
...
+# --- Set up the forward operator (ray transform) --- #
+
+
+# Reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 300 samples per dimension.
+space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+geometry = odl.tomo.cone_beam_geometry(space,
```
kohr-h: Update comment above, and docstring at the top (they say parallel).

adler-j:
Fixed all comments, merge after CI

adler-j:
Frankly very happy about this syntax, easy to use and very intuitive. Some minor kinks to fix ofc, but still.

I'd love some comments from @MatthiasJE on it.

MatthiasJE:
I tried to follow the development but quickly lost track of what is happening. I guess most of this is not very intuitive to non-ODL people. Maybe it is a good idea if I have a look at the examples that you provided to see how I like the usage of the indexed geometries. Do you @adler-j think this makes sense?

adler-j:
Sure! The interface is what if like feedback on

examples/solvers/kaczmarz_tomography.py
```diff
@@ -0,0 +1,74 @@
...
+# Reconstruction space: discretized functions on the rectangle
+# [-20, 20]^2 with 300 samples per dimension.
+space = odl.uniform_discr(
+    min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='float32')
+
+# Make a fan (cone) beam geometry with flat detector
+geometry = odl.tomo.cone_beam_geometry(space,
+                                       src_radius=40, det_radius=40,
+                                       num_angles=360, det_shape=360)
+
```
MatthiasJE: This only works with astra? I receive the error message

`TypeError: 'skimage' backend only supports 2d parallel geometries
`

adler-j: I made some modifications so we should be able to use this with parallel beam instead. I'll change that.

PR #1116, opened 2017-09-07
BUG: fix bug in mse fom (contrib)

aringh:


odl/contrib/fom/supervised.py
```diff
@@ -69,6 +69,7 @@ def mean_squared_error(data, ground_truth, mask=None, normalized=False):
 
     diff = data - ground_truth
     fom = l2_normSquared(diff)
+    fom /= data.space.one().norm()**2
```
adler-j: l2_normSquared(data.space.one()) at least looks better for me?

aringh:
It was handled in `mean_value_difference`. I updated it to the same in `mean_absolute_error` and `mean_squared_error`

PR #1119, opened 2017-09-08
minor fixes for the examples

grlee77:
The first commit here fixes outdated calls to `odl.FN_IMPL` in a couple of the examples so that they run successfully.

The third commit adds a legend to the plot produced by the Rosenbrock minimization example, but this turns out to be slightly complicated to do.  I can drop that commit if you think it complicates things too much.  I initially tried just naively passing `label='CG'`, etc. to `CallbackShowConvergence`, but this resulted in a legend with labels for every single point in the scatter plot!

The rest is mainly just adding `force_show=True` to examples where it wasn't already present so that the plots stick around when the examples are run from the command line.


examples/solvers/rosenbrock_minimization.py
```diff
@@ -38,24 +39,35 @@
 x = space.zero()
 odl.solvers.steepest_descent(f, x, line_search=line_search,
                              callback=callback)
+legend_artists = [callback.ax.collections[-1], ]
```
adler-j: This is an OK change for now, but I'll see if we can add this functionality to `CallbackShowConvergence` in the long term, see #1121

grlee77:
Thanks for the quick review and merge!

Issue #590, opened 2016-09-20
Functionals over complex spaces


adler-j:
For some applications (notably MRI, but possibly also phase contrast) it would be interesting to look at data spaces that are complex. In that case, we need to support this for the functionals as well.

##### Convex conjugate

To my knowledge, the issue with this is mostly that we cannot define the convex conjugate properly. The definition is usually:

```
f^*(x^*) = sup_x { <x^*, x> - f(x) }
```

This obviously becomes problematic in case `X` is complex, since then `<x^*, x>` is complex valued and this doesn't make any sense. One way to solve this, used in [this article](http://download.springer.com/static/pdf/491/art%253A10.1007%252Fs001860050004.pdf?originUrl=http%3A%2F%2Flink.springer.com%2Farticle%2F10.1007%2Fs001860050004&token2=exp=1474387469~acl=%2Fstatic%2Fpdf%2F491%2Fart%25253A10.1007%25252Fs001860050004.pdf%3ForiginUrl%3Dhttp%253A%252F%252Flink.springer.com%252Farticle%252F10.1007%252Fs001860050004*~hmac=be918533964713cf8a2eeaa48a299b9d16ab906b51a252af327b85f7aebacdc6) is to simply take the real part:

```
f^*(x^*) = sup_x { Re <x^*, x> - f(x) }
```

At least for the `|| . ||_2^2` functional, this gives the usual result, and I'm quite sure we get nice results for the norms as well.

adler-j:
Bumping this. I'll go ahead with the #587 PR merge, but it causes the `douglas_rachford_primal_dual_mri` example to fail, since we need to define a L2 functional with a complex domain.

We could use @aringh input here.

adler-j:
Bumping this issue again, we really need to fix this since this change introduced problems down the line.

kohr-h:
Yes. Mathematically I think we should go for the "real part" solution since it effectively identifies C with R^2, which is probably what you usually want. In many cases we can remap the convex conjugate to well-known ones for real spaces, for example the 1-norm would correspond to the group norm 1,2 (or 2,1?).

PR #1120, opened 2017-09-08
Fix: avoid improper casts from complex to float when solving complex-valued problems

grlee77:
I am a new user to ODL and wanted to try a simple example with complex-valued data.  As a first step, I tried creating a complex-valued image equivalent of  `chambolle_pock_denoising.py`.   The modified example is included here as 019bf0d.  It may not be worth keeping this example as it is very similar to the real-valued case, but I have included it here for now so you can see the problem that inspired the fixes in this PR.

Commits 1 and 3 are changes that were necessary to avoid errors from calling `float()` on arguments that may be complex-valued (i.e.  `TypeError: can't convert complex to float`)

The `gradient` operator was apparently passing a complex-valued `pad_const` to `finite_diff`.  I have changed the cast from `float()` to `complex()` for the case where the data input to `finite_diff` is complex-valued.

A minimal example that fails prior to this PR, but succeeds afterward is:
```
import numpy as np
import odl
odl.discr.diff_ops.finite_diff(np.arange(8., dtype=np.complex64), axis=0, pad_const=0+0j)
```

28de98a ensures that `quadratic_coeff` ends up real-valued (the users can pass a complex-valued input as long as the imaginary part is 0).  The error with `quadratic_coeff` being complex also occurs for the existing example at `examples/solvers/douglas_rachford_pd_mri.py`.  That example also runs successfully when using the changes proposed in this PR.

Another TypeError in the L1 proximal operator where the step size, `sigma` also had complex type.  I propose taking the `abs` of the step size so that it is guaranteed to be real-valued.  I have attempted to replicate this behaviour across the various other proximal operators as well.

Being new to ODL, there very well may be better alternative solutions to each of these, but hopefully this is helpful!

### DIFF
--- a/examples/solvers/chambolle_pock_denoising_complex.py
+++ b/examples/solvers/chambolle_pock_denoising_complex.py
@@ -11,13 +11,12 @@ For further details and a description of the solution method used, see
 """
 
 import numpy as np
-import scipy
+import scipy.misc
 import odl
 
 # Read test image: use only every second pixel, convert integer to float,
 # and rotate to get the image upright
-dtype = np.complex64
-image = np.rot90(scipy.misc.ascent()[::1, ::1], 3).astype(dtype)
+image = np.rot90(scipy.misc.ascent()[::1, ::1], 3).astype('float32')
 image = image + 1j*image.T
 shape = image.shape
 
@@ -25,20 +24,16 @@ shape = image.shape
 image /= image.real.max()
 
 # Discretized spaces
-space = odl.uniform_discr([0, 0], shape, shape, dtype=dtype)
+space = odl.uniform_discr([0, 0], shape, shape, dtype='complex64')
 
 # Original image
 orig = space.element(image)
 
 # Add noise
-image += (np.random.normal(0, 0.05, shape) +
-          1j*np.random.normal(0, 0.05, shape))
-
-# Data of noisy image
-noisy = space.element(image)
+noisy = image + 0.05 * odl.phantom.white_noise(orig.space)
 
 # Gradient operator
-gradient = odl.Gradient(space, method='forward')
+gradient = odl.Gradient(space)
 
 # Matrix of operators
 op = odl.BroadcastOperator(odl.IdentityOperator(space), gradient)
--- a/examples/solvers/chambolle_pock_denoising_complex.py
+++ b/examples/solvers/chambolle_pock_denoising_complex.py
@@ -1,0 +1,83 @@
+"""Total variation denoising of complex data using the Chambolle-Pock solver.
+
+Solves the optimization problem
+
+    min_{x}  1/2 ||x - g||_2^2 + lam || |grad(x)| ||_1
+
+Where ``grad`` the spatial gradient and ``g`` is given noisy data.
+
+For further details and a description of the solution method used, see
+:ref:`chambolle_pock` in the ODL documentation.
+"""
+
+import numpy as np
+import scipy
+import odl
+
+# Read test image: use only every second pixel, convert integer to float,
+# and rotate to get the image upright
+dtype = np.complex64
+image = np.rot90(scipy.misc.ascent()[::1, ::1], 3).astype(dtype)
+image = image + 1j*image.T
+shape = image.shape
+
+# Rescale max to 1
+image /= image.real.max()
+
+# Discretized spaces
+space = odl.uniform_discr([0, 0], shape, shape, dtype=dtype)
+
+# Original image
+orig = space.element(image)
+
+# Add noise
+image += (np.random.normal(0, 0.05, shape) +
+          1j*np.random.normal(0, 0.05, shape))
+
+# Data of noisy image
+noisy = space.element(image)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
+
+# Matrix of operators
+op = odl.BroadcastOperator(odl.IdentityOperator(space), gradient)
+
+# Set up the functionals
+
+# l2-squared data matching
+l2_norm = odl.solvers.L2NormSquared(space).translated(noisy)
+
+# Isotropic TV-regularization: l1-norm of grad(x)
+l1_norm = 0.15 * odl.solvers.L1Norm(gradient.range)
+
+# Make separable sum of functionals, order must correspond to the operator K
+f = odl.solvers.SeparableSum(l2_norm, l1_norm)
+
+g = odl.solvers.ZeroFunctional(op.domain)
+
+# --- Select solver parameters and solve using Chambolle-Pock --- #
+
+
+# Estimated operator norm, add 10 percent to ensure ||K||_2^2 * sigma * tau < 1
+op_norm = 1.1 * odl.power_method_opnorm(op, xstart=noisy)
+
+niter = 200  # Number of iterations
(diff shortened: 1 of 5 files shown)


odl/discr/diff_ops.py
```diff
@@ -802,7 +802,12 @@ def finite_diff(f, axis, dx=1.0, method='forward', out=None, **kwargs):
     if pad_mode not in _SUPPORTED_PAD_MODES:
         raise ValueError('`pad_mode` {} not understood'
                          ''.format(pad_mode))
-    pad_const = float(kwargs.pop('pad_const', 0))
+
+    pad_const = kwargs.pop('pad_const', 0)
+    if np.iscomplexobj(f):
```
adler-j: This should be written `pad_const = f.dtype.type(pad_const)` in order to get the correct datatype (e.g. `float32` vs `float64`).

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -146,7 +146,7 @@ def convex_conj_prox_factory(sigma):
         proximal : `Operator`
             The proximal operator of ``s * F^*`` where ``s`` is the step size
         """
-        sigma = float(sigma)
+        sigma = float(abs(sigma))
```
adler-j: Are we sure that this is the correct formula in general? I'm frankly not sure but we could use input from @aringh and Sebastian who know some optimization theory.

Also note the issue  #590 where we discuss this in depth.

adler-j:
Overall a very nice pull request and I truly low the dedication you are putting into this. I'll try to give some further thought to the exact mathematics behind this to make sure we're doing the right thing.

odl/discr/diff_ops.py
grlee77: Sure, that is a better approach.

odl/solvers/nonsmooth/proximal_operators.py
grlee77: I'm not entirely sure on that either and am not an expert on the mathematics of optimization (I am a biomedical engineer by training with expertise primarily in MRI acquisition methods and medical image reconstruction).  It seems to work okay for the denoising example provided here, but I agree that we should wait for expert feedback on the correct approach.

kohr-h:
Hi @grlee77, and thanks so much for the effort to make a PR!

Regarding the issue, as @adler-j wrote we had a long and not 100 % conclusive discussion about how to handle complex spaces. At some point we need an expert to sort this out, or one of us needs to read up on it.
The proximal `sigma`, however, should never be affected by this issue. It's a step-size (an implicit one), and by definition real and positive. I also don't see where a `proximal_*` function should ever get a complex value from, except from wrong code.
In fact, when I run your example @grlee77, it works perfectly fine when I go back from `float(abs(sigma))` to `float(sigma)`. So I don't think we need to do anything here. I'd rather let it fail for complex `sigma`, since that means there is an error somewhere.

Regarding your example, I agree it's very close to the existing denoising example, but on the other hand we also have a complex FBP example which is largely a clone of the real one. And since we don't have a unit test for complex optimization yet, I think it's okay to add the example. That way it gets tested at least with each new release.

odl/solvers/functional/functional.py
```diff
@@ -923,7 +923,13 @@ def __init__(self, func, quadratic_coeff=0, linear_term=None):
                             ''.format(func))
 
         self.__functional = func
-        self.__quadratic_coeff = func.domain.field.element(quadratic_coeff)
+        quadratic_coeff = func.domain.field.element(quadratic_coeff)
+        if np.iscomplexobj(quadratic_coeff):
```
adler-j: This line can actually be ommited since implicitly if `imag != 0` then the object is complex.

```python
quadratic_coeff = func.domain.field.element(quadratic_coeff)
if quadratic_coeff.imag != 0:
    raise ...
self.__quadratic_coeff = quadratic_coeff.real
```

examples/solvers/chambolle_pock_denoising_complex.py
```diff
@@ -0,0 +1,83 @@
...
+image /= image.real.max()
+
+# Discretized spaces
+space = odl.uniform_discr([0, 0], shape, shape, dtype=dtype)
+
+# Original image
+orig = space.element(image)
+
+# Add noise
+image += (np.random.normal(0, 0.05, shape) +
```
adler-j: If you instead use

```python
noisy = image + 0.05 * odl.phantom.white_noise(image.space)
```

the noise is implicitly complex valued if the space is so.

examples/solvers/chambolle_pock_denoising_complex.py
```diff
@@ -0,0 +1,83 @@
...
+dtype = np.complex64
+image = np.rot90(scipy.misc.ascent()[::1, ::1], 3).astype(dtype)
+image = image + 1j*image.T
+shape = image.shape
+
+# Rescale max to 1
+image /= image.real.max()
+
+# Discretized spaces
+space = odl.uniform_discr([0, 0], shape, shape, dtype=dtype)
```
adler-j: Just do `dtype='complex64'` here

examples/solvers/chambolle_pock_denoising_complex.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+import numpy as np
+import scipy
+import odl
+
+# Read test image: use only every second pixel, convert integer to float,
+# and rotate to get the image upright
+dtype = np.complex64
+image = np.rot90(scipy.misc.ascent()[::1, ::1], 3).astype(dtype)
```
adler-j: This `astype` is unnecessary since the next line will implicitly cast to complex

examples/solvers/chambolle_pock_denoising_complex.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Add noise
+image += (np.random.normal(0, 0.05, shape) +
+          1j*np.random.normal(0, 0.05, shape))
+
+# Data of noisy image
+noisy = space.element(image)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
```
adler-j: `method='forward'` is the default and can be omited.

examples/solvers/chambolle_pock_denoising_complex.py
grlee77: I can remove the `dtype` variable, but .astype('float32') will be needed if we want to end up as 'complex64' and not 'complex128'

grlee77:
> In fact, when I run your example @grlee77, it works perfectly fine when I go back from float(abs(sigma)) to float(sigma). So I don't think we need to do anything here. I'd rather let it fail for complex sigma, since that means there is an error somewhere.

This sounds reasonable.  I am not sure why I thought I needed that for the sigmas.  I will remove that commit, replacing with one that only enforces real input for the the `scaling` argument to `proximal_arg_scaling`.  That one was still required in order for the example to run.  Do you consider that to be a bug upstream of `proximal_arg_scaling`?

grlee77:
@adler-j .  I think I have addressed the comments raised during review. All float(abs(sigma)) calls have been removed.  There is a check on `scaling.imag != 0` in 6b44861 that mirrors the behaviour for `quadratic_coeff` in the functionals.

adler-j:
Agree that all comments are fixed. I'll merge this.

PR #1126, opened 2017-09-14
ENH: Add some tests for supervised FoM

adler-j:
Adds tests for the most used supervised FoM.

### DIFF
--- a/odl/contrib/fom/test/test_supervised.py
+++ b/odl/contrib/fom/test/test_supervised.py
@@ -3,13 +3,17 @@ import pytest
 import scipy.signal
 import scipy.misc
 import odl
+import odl.contrib.fom
 from odl.contrib.fom.util import filter_image_sep2d
 from odl.util.testutils import simple_fixture
 
-fft_impl_params = [odl.util.testutils.never_skip('numpy'),
-                   odl.util.testutils.skip_if_no_pyfftw('pyfftw')]
-fft_impl = simple_fixture('fft_impl', fft_impl_params,
+fft_impl = simple_fixture('fft_impl',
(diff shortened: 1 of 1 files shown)


adler-j:
Could use a review

odl/contrib/fom/test/test_supervised.py
```diff
@@ -0,0 +1,54 @@
...
+                                 odl.uniform_discr(0, 1, 10)])
+
+def test_mean_squared_error(space):
+    true = odl.phantom.white_noise(space)
+    data = odl.phantom.white_noise(space)
+
+    result = odl.contrib.fom.mean_squared_error(data, true)
+    expected = np.mean((true - data) ** 2)
+
+    assert pytest.approx(result) == expected
```
kohr-h: Slightly weird to have it this way, I prefer `result == pytest.approx(expected)`

odl/contrib/fom/test/test_supervised.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+import pytest
+import odl
+import odl.contrib.fom
+from odl.util.testutils import simple_fixture
+
+space = simple_fixture('space', [odl.rn(3),
+                                 odl.uniform_discr(0, 1, 10)])
+
+def test_mean_squared_error(space):
```
kohr-h: Given the recent issue #1136 an additional check that the normalized variants do the right thing would be good. That means checking `0 <= result_normalized <= 1` for random inputs, `(x, x)` and `(x, -x)` which are often best and worst case, resp.

adler-j:
Merging after CI

Issue #944, opened 2017-02-28
access/domain(?) of functionals


MatthiasJE:
Why are functionals part of solvers, e.g. odl.solvers.L1Norm? To me functionals should probably have their own category (or at least some non-contradicting category).

adler-j:
Well the rationale was that we use them with solvers, but your point is actually quite valid, it would feel more natural to access them as `odl.functionals.L1Norm`.

What do we feel about such a change @aringh, @kohr-h?

aringh:
The rationale is what @adler-j wrote. I have no strong opinion in the question, but moving sounds like a good idea :-)

kohr-h:
Part of the rationale was to not clutter the top-level namespace more than necessary. Let me give you some details about this.

The overall structure of ODL currently looks like this:
```
odl
 |
 |-- set  *
 |-- space  *
 |-- operator  *
 |-- discr  *
 |-- trafos
 |-- tomo
 |-- solvers
 |-- phantom
 |-- util
 |-- deform
 |-- ufunc_ops
```
Everything with a `*` at the end is considered part of the "core" in the sense that all names that are exposed in any of those modules (including submodules) in `__all__` are available **at top level**. Therefore you can write `odl.IdentityOperator` instead of going down to `odl.operator.operator`.

However, one namespace can only hold a certain amount of names before it gets hopelessly cluttered. For that reason, all the subpackages have their own namespaces that are *not* added to the top level. For example, you need to type `odl.trafos.FourierTransform`, it's not available at top level.

Now, in the case of functionals, there are so many that they would lead to a significant amount of clutter if we would simply dump them to top level. Therefore I'm not in favor of making them part of the core in that  sense.
The alternative would be to put them into their own subpackage `functional` (without "s") together with the prox stuff and all that. I'm not so sure about that either, much of it is strongly related to convex optimization, and taking it out of that context feels wrong. There is also the silent assumption that all functionals are convex which clearly makes more sense if you put them together with other convex optimization stuff.

What we could alternatively think about would be a rename of `solvers` to `optim` or something like that. That would at least remove the association "everything in here is a solver", which is not true anyway.

kohr-h:
Any new insights here? I just want to mention that there's a related unsolved issue #637, namely where to put concrete operators that depend on substantial parts of the library. For example operators that use the Fourier transform (not mentioning names here) which itself depends on a fair amount of stuff.

[...]

PR #1130, opened 2017-09-18
WIP: implement convolution

kohr-h:
Going a different route this time. First discrete convolution (`R^n -> R^n`) using direct or FFT methods, then discretized continuous convolution using the discrete one internally.

I chose ``issue-342__tensor`` as base rather than ``master``, since implementation was a bit easier. So this branch goes in after the tensor one.

**TODO:**
- [x] Implement  `DiscreteConvolution`
- [x] ~~Add option to provide kernel in Fourier space~~ This is too much of a hassle, makes the initializer way too complicated. It's better to expose helpers to do this manually.
- [x] Turn assertions into proper exceptions with error messages
- [x] Do all input checks according to the documentation
- [x] Write unit tests for `DiscreteConvolution`
- [x] Profile the code to see whether it has avoidable bottlenecks (no obvious ones)
- [x] Implement `DiscreteConvolution.adjoint`
- [ ] Optimize case when padding is 0, since `out` can be used then
- [ ] Write example for `DiscreteConvolution`
- [ ] Implement `Convolution`
- [ ] Write unit tests for `Convolution`
- [ ] Write example for `Convolution`


kohr-h:
Note: I made `oplib` a place for "advanced" operators that cannot be part of any other subpackage due to circular dependencies. Currently in there are `ufunc_ops.py` and `convolution.py`. As discussed in another thread (#944), we can think about moving functionals to the same place.

odl/oplib/convolution.py
```diff
@@ -0,0 +1,528 @@
...
+            Space of output elements of the convolution. Must be of the
+            same shape as ``domain``. If not given, the range is equal to
+            ``domain.astype(result_dtype)``, where ``result_dtype`` is
+            the data type of the convolution result. If ``impl='real'``,
+            integer dtypes are preserved, while for ``impl='fft'``,
+            the smallest possible floating-point type is chosen.
+        axis : int or sequence of ints, optional
+            Coordinate axis or axes in which to take the convolution.
+            ``None`` means all input axes.
+        impl : {'fft', 'real'}
```
kohr-h: Not sure about naming, maybe `'direct'` would be better.

odl/oplib/convolution.py
```diff
@@ -0,0 +1,528 @@
...
+            axis, with padding values corresponding to ``axis`` entries
+            as provided.
+            Default: ``min(kernel.shape - 1, 64)``
+        padded_shape : sequence of ints, optional
+            Apply zero-padding with this target shape. Cannot be used
+            together with ``padding``.
+        cache_kernel_ft : bool, optional
+            If ``True``, store the Fourier transform of the kernel for
+            later reuse.
+            Default: ``False``
```
kohr-h: There should also be an option to provide the kernel directly in Fourier space.

odl/oplib/convolution.py
```diff
@@ -0,0 +1,528 @@
...
+        ...                    [-3, -6, -9]])
+        >>> conv(x)
+        rn((3, 3)).element(
+            [[  1.,   3.,   5.],
+             [  3.,   9.,  15.],
+             [ -1.,  -3.,  -5.]]
+        )
+
+        Convolution in selected axes can be done either with broadcasting
+        or with "stacked kernels":
```
kohr-h: Mention that this is not possible with `impl='real'`

adler-j:
Why close this?

kohr-h:
Got auto-closed when the base branch disappeared, will make a new PR into master.

PR #1132, opened 2017-09-19
DOC: fix source installation instructions

kohr-h:
The TL;DR instructions for installation from source were wrong for conda, this fixes them.


kohr-h:
> We really gotta release a new version btw :D

One before 1.0rc1?

PR #1137, opened 2017-09-20
ENH: Change kaczmarz to parallel beam

adler-j:
Fix according to @MatthiasJE

### DIFF
--- a/examples/solvers/kaczmarz_tomography.py
+++ b/examples/solvers/kaczmarz_tomography.py
@@ -49,6 +49,9 @@ elif split == 'interlaced':
     ray_trafos = [odl.tomo.RayTransform(space, geometry[i::n])
                   for i in range(n)]
 
+# Create one large ray transform from components
+ray_trafo = odl.BroadcastOperator(*ray_trafos)
+
 # --- Generate artificial data --- #
 
 
@@ -56,11 +59,10 @@ elif split == 'interlaced':
 phantom = odl.phantom.shepp_logan(space, modified=True)
 
 # Create sinogram of forward projected phantom with noise
-data = [ray_trafo(phantom) for ray_trafo in ray_trafos]
+data = ray_trafo(phantom)
 
-# Compute steplengths
-omega = [odl.power_method_opnorm(ray_trafo) ** (-2)
-         for ray_trafo in ray_trafos]
+# Compute steplength
+omega = n * odl.power_method_opnorm(ray_trafo) ** (-2)
 
 # Optionally pass callback to the solver to display intermediate results
 callback = (odl.solvers.CallbackPrintIteration() &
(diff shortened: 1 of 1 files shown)


MatthiasJE:
I recommend to make this example smaller scale or move it to "contrib", e.g. shape=[100, 100], num_angles=160.

In addition, I was thinking that it may be better to make these subset operators a "Broadcasting operator".

Otherwise it works really well.

MatthiasJE:
A positive comment on speed that may be included in the example:

```
ray_trafos = odl.BroadcastOperator(*ray_trafos_list)

import src

x1 = ray_trafos.range.zero()
src.tic()
ray_trafos(phantom, out=x1)
src.toc()

x2 = ray_trafos.range.zero()
src.tic()
for i in range(ray_trafos.size):
    ray_trafos[i](phantom, out=x2[i])
src.toc()

```

> Elapsed time is 0.166328 seconds.
> Elapsed time is 0.167632 seconds.

This is really good!

MatthiasJE:
Here is a negative result about adjointness:

```
A = ray_trafos[0]
X = A.domain
Y = A.range

x = X.one()
y = Y.one()

print(y.inner(A(x)))
print(x.inner(A.adjoint(y)))
print(y.inner(A(x)) - x.inner(A.adjoint(y)))
```

> 249.86969972769987
> 251.32757812500006
> -1.4578783973001919

MatthiasJE:
Ups, this is about #1110.

adler-j:
Regarding the size, even with the current parameters this takes like one second with astra_gpu, making it smaller would basically make it instant and ruins the visualization, but perhaps I can add a pause to the callback instead.

The adjoint-ness is a "well known" problem in that our backends do not have exactly matched adjoints (unless you use `impl='astra-cpu'`).

MatthiasJE:
Sure, but I would prefer if "standard" examples would not require backends like astra and hardware like gpus. In the end, this example is about the functionality of ODL. That being said, it would also be good to have more complicated examples that require certain backends and hardware (and show how to use them).

adler-j:
I agree. I'll try to downscale it and add some artificial pauses to improve the visualization.

adler-j:
Fixed the scaling

MatthiasJE:
But you did not commit this yet, right?

MatthiasJE:
I installed astra with

`conda install -c astra-toolbox astra-toolbox`

and indeed, the scaling issue is gone. So easy :)

MatthiasJE:
If we introduce the BroadcastOperator

`ray_trafo = odl.BroadcastOperator(ray_trafos)`

e.g. in line 51

then we can write line 59 as 

`data = ray_trafo(phantom)`

and line 74 as

`odl.solvers.kaczmarz(ray_trafo, x, data, niter=5, omega=omega, callback=callback, callback_loop='inner')
`
both of which I find more intuitive.

Otherwise, I like the example and it could be merged.

adler-j:
Agree with your comments, I'll do that!

adler-j:
Considering this reviewed. Will merge after CI.

PR #1138, opened 2017-09-20
MAINT: Improve handling of angles in elekta

adler-j:
Very minor fixes to the elekta stuff

### DIFF
--- a/odl/contrib/tomo/elekta.py
+++ b/odl/contrib/tomo/elekta.py
@@ -85,21 +85,18 @@ def elekta_icon_geometry(sad=780.0, sdd=1000.0,
     if angles is not None and num_angles is not None:
         raise ValueError('cannot provide both `angles` and `num_angles`')
     elif angles is not None:
-        angles = np.array(angles, dtype=float)
+        angles = odl.nonuniform_partition(angles)
         assert angles.ndim == 1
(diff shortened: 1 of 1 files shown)


odl/contrib/tomo/elekta.py
```diff
@@ -85,22 +85,19 @@ def elekta_icon_geometry(sad=780.0, sdd=1000.0,
     if angles is not None and num_angles is not None:
         raise ValueError('cannot provide both `angles` and `num_angles`')
     elif angles is not None:
-        angles = np.array(angles, dtype=float)
+        angles = odl.nonuniform_partition(angles)
         assert angles.ndim == 1
     elif num_angles is not None:
-        angles = np.linspace(1.2, 5.0, num_angles)
+        angles = odl.uniform_partition(1.2, 5.0, num_angles)
```
kohr-h: Not including 1.2 and 5.0? What are they anyway?

adler-j: Approximately the start and stop angles. Mentioned in the whitepaper (indirectly, I think).

adler-j: Further we want to be able to handle `num_angles=1`, in which case this is needed.

PR #1142, opened 2017-09-20
Issue 1117  lincomboperator adjoint

adler-j:


odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
+    def __init__(self, space, *coeffs):
         """Initialize a new instance.
 
         Parameters
         ----------
         space : `LinearSpace`
             Space of elements which the operator is acting on.
-        a, b : ``space.field`` elements
-            Scalars to multiply ``x[0]`` and ``x[1]`` with, respectively.
+        coeffs : ``space.field`` elements
```
kohr-h: We usually write this as `coeff1, ..., coeffN`

odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
-        rn(3).element([2.0, 4.0, 6.0])
+        >>> op([[1, 2, 3],
+        ...     [2, 3, 4]],
+        ...    out=z)  # Returns z
+        rn(3).element([3.0, 5.0, 7.0])
         >>> z
-        rn(3).element([2.0, 4.0, 6.0])
+        rn(3).element([3.0, 5.0, 7.0])
+
+        Can also compute arbitrary weighted sums:
```
kohr-h: It is possible to

odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
+        else:
+            out.lincomb(self.coeffs[0], x[0], self.coeffs[1], x[1])
+            for i in range(2, n):
+                out.lincomb(1, out, self.coeffs[i], x[i])
+
         return out
 
+    @property
+    def adjoint(self):
+        """Return the adjoint.
```
kohr-h: ~~Return~~

odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
+        return '{}({!r}, {})'.format(self.__class__.__name__,
+                                     self.range, coeff_str)
 
-    def __str__(self):
-        """Return ``str(self)``."""
-        return "{}*x + {}*y".format(self.a, self.b)
+
+class ProductSpaceEmbedding(Operator):
+
+    """Operator mapping a vector to several copies of it::
```
kohr-h: Same here, don't continue the summary line.

odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
+
+        ProductSpaceEmbedding(spc, a, b)(x) == [a * x, b * x]
```
kohr-h: This is just `BroadcastOperator(*[ScalingOperator(ci) for ci in coeffs])`, no need for a new operator.

odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
+        """Return the adjoint.
+
+        Examples
+        --------
+        >>> r3 = odl.rn(3)
+        >>> op = LinCombOperator(r3, 1.0, 1.0)
+        >>> op.adjoint
+        ProductSpaceEmbedding(ProductSpace(rn(3), 2), 1.0, 1.0)
+        """
+        return ProductSpaceEmbedding(self.domain, *self.coeffs)
```
kohr-h: Should be `self.range`.

odl/operator/default_ops.py
```diff
@@ -171,53 +171,157 @@ def __str__(self):
...
+    @property
+    def adjoint(self):
+        """Return the adjoint.
+
+        Examples
+        --------
+        >>> r3 = odl.rn(3)
+        >>> op = LinCombOperator(r3, 1.0, 1.0)
+        >>> op.adjoint
+        ProductSpaceEmbedding(ProductSpace(rn(3), 2), 1.0, 1.0)
```
kohr-h: Wrong result

kohr-h:
> Good initiative, but too much code :-P

By that I mean that the adjoint doesn't need a new operator implementation.

odl/operator/default_ops.py
adler-j: Well the other fella is `ReductionOperator(*[ScalingOperator(ci) for ci in coeffs])` not convinced we should remove only one of them.

adler-j: But now that you mention it there is a very clear "duality" going on here. Let's think about it for a while

kohr-h: You're right. Maybe just make it a function instead?

adler-j: Yes that may actually be the best option. It would also preserve `op.adjoint.adjoint == op` in some sense

kohr-h:
Status?

adler-j:
Basically this can be closed, If we do this, it would be according to the discussion in the comments which would be orthogonal to what has been done here.

PR #1144, opened 2017-09-21
MAINT: various fixes to supervised FOMs, closes #1136

### LINKED ISSUE #1136: contrib.fom.mean_squared_error logical error in range
As noted by @MJLagerwerf the documentation of `mean_squared_error` is wrong about the range of the normalized FOM. It claims that
```
MSE(f, g) = ||f - g||^2 / (||f||^2 + ||g||^2)
``` 
lies in `[0, 1]`. However, for `g = -f` the result is 2. We should change the doc or the formula.

kohr-h:
- fix bad normalization in MSE
- fix broken mask input checks
- reduce redundancy in docs
- improve math formulas a bit
- clarify usage of masks
- move Notes below Parameters
- fix wrong mean error implementations for `normalized=True`
  (don't normalize by space volume in this case)
- fix broken masking in MVD
- rename `normalize` to `normalized` in PSNR
- remove `normalized` option from SSIM, makes no sense

### DIFF
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -37,7 +37,7 @@ def mean_squared_error(data, ground_truth, mask=None, normalized=False):
     Returns
     -------
     mse : float
-        FOM value, with higher value meaning lower correspondence.
+        FOM value, where a lower value means a better match.
 
     Notes
     -----
@@ -95,7 +95,7 @@ def mean_absolute_error(data, ground_truth, mask=None, normalized=False):
     Returns
     -------
     mae : float
-        FOM value, with higher value meaning lower correspondence.
+        FOM value, where a lower value means a better match.
 
     Notes
     -----
@@ -108,10 +108,12 @@ def mean_absolute_error(data, ground_truth, mask=None, normalized=False):
     of the functions. For :math:`\mathbb{R}^n` type spaces, this is equal
     to the number of elements :math:`n`.
 
+    The normalized form is
+
     .. math::
         \mathrm{MAE_N}(f, g) = \\frac{\| f - g \|_1}{\| f \|_1 + \| g \|_1}.
 
-    The normalized FOM takes values in [0, 1].
+    The normalized variant takes values in :math:`[0, 1]`.
     """
     l1_norm = odl.solvers.L1Norm(data.space)
     if mask is not None:
@@ -145,7 +147,7 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False):
     Returns
     -------
     mvd : float
-        FOM value, with higher value meaning lower correspondence.
+        FOM value, where a lower value means a better match.
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


adler-j:
This should be done with #1126 in one way or another

kohr-h:
You mean as a PR into your branch?

adler-j:
I'll quickly get that one merged. Point is that these should have tests too

odl/contrib/fom/supervised.py
```diff
@@ -18,180 +18,160 @@
...
+    -----
+    The FOM evaluates
+
+    .. math::
+        \mathrm{MAE}(f, g) = \\frac{\| f - g \|_1}{\| 1 \|_1},
+
+    where :math:`\| 1 \|_1` is the volume of the domain of definition
+    of the functions. For :math:`\mathbb{R}^n` type spaces, this is equal
+    to the number of elements :math:`n`.
+
```
adler-j: Missing "the normalized form is"

odl/contrib/fom/supervised.py
```diff
@@ -213,61 +193,51 @@ def standard_deviation_difference(data, ground_truth, mask=None,
...
+        Reference to compare ``data`` to.
+    mask : `array-like`, optional
+        If given, ``data * mask`` is compared to ``ground_truth * mask``.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+
+    Returns
+    -------
+    sdd : float
+        FOM value, with higher value meaning lower correspondence.
```
adler-j: I would overall change "correspondence" to something like "meaning a better fit", hard to read as is

odl/contrib/fom/supervised.py
```diff
@@ -353,189 +327,161 @@ def range_difference(data, ground_truth, mask=None, normalized=False):
...
     ----------
     data : `FnBaseVector`
-        Input data or reconstruction.
+        Input data to compare to the ground truth.
     ground_truth : `FnBaseVector`
         Reference to compare ``data`` to.
-    normalized : bool
-        TEXT
+    size :
+        TODO...
```
adler-j: Yupp

kohr-h: Well, I already did more than necessary, but this would require me to check and see what the parameters do. Nope. This is on the original authors. Ping @davlars @zickert

adler-j: Leave it here then

kohr-h: I added the two obvious ones, but the `K1`, `K2` and `dynamic_range` params are for later.

kohr-h:
Fixes are in, merging after CI.

odl/contrib/fom/supervised.py
zickert: To finish this could perhaps be a topic for the upcoming team workshop day.

PR #1148, opened 2017-09-21
ENH: Add uniform_noise, closes #1143

### LINKED ISSUE #1143: uniform noise
Not very urgent, but it would be good if ODL supports also uniform noise. This is particularly useful when dealing with ProductSpaces. As they do not have a "shape" the numpy workaround would be a couple of lines which is to be avoided.

adler-j:


odl/phantom/noise.py
```diff
@@ -67,6 +67,58 @@ def white_noise(space, mean=0, stddev=1, seed=None):
...
+
+    Parameters
+    ----------
+    space : `FnBase` or `ProductSpace`
+        The space in which the noise is created.
+    low : ``space.field`` element or ``space`` `element-like`, optional
+        The lower bound of the uniform noise. If a scalar, it is interpreted as
+        ``low * space.one()``.
+        If ``space`` is complex, the real and imaginary parts are interpreted
+        as the mean of their respective part of the noise.
```
kohr-h: Hm, I don't really understand this. From the code I think it's just taking real and imaginary parts separately, no?

odl/phantom/noise.py
```diff
@@ -67,6 +67,58 @@ def white_noise(space, mean=0, stddev=1, seed=None):
...
+    low : ``space.field`` element or ``space`` `element-like`, optional
+        The lower bound of the uniform noise. If a scalar, it is interpreted as
+        ``low * space.one()``.
+        If ``space`` is complex, the real and imaginary parts are interpreted
+        as the mean of their respective part of the noise.
```
kohr-h: Copy-paste from above

odl/phantom/noise.py
adler-j: "mean" is a doc copy paste from white noise.

adler-j:
Merge after CI

PR #1153, opened 2017-09-22
Add readmes

adler-j:
As suggested by @MatthiasJE, this should make the ODL code much more easy to navigate.


kohr-h:
> all good

I mean, up to the current import failure.

Issue #1026, opened 2017-06-09
"saved filedescriptor not valid anymore" when running individual tests


adler-j:
#### Steps to reproduce:

Simply run any individual test file (e.g. not `pytest` from console, but simply run the test file) and I get the error:

```

  File "D:/Programming/ODL/odl/test/solvers/nonsmooth/chambolle_pock_test.py", line 145, in <module>
    pytest.main([str(__file__.replace('\\', '/')), '-v'])

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\config.py", line 49, in main
    config = _prepareconfig(args, plugins)

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\config.py", line 158, in _prepareconfig
    pluginmanager=pluginmanager, args=args)

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 745, in __call__
    return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 339, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 334, in <lambda>
    _MultiCall(methods, kwargs, hook.spec_opts).execute()

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 613, in execute
    return _wrapped_call(hook_impl.function(*args), self.execute)

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 250, in _wrapped_call
    wrap_controller.send(call_outcome)

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\helpconfig.py", line 34, in pytest_cmdline_parse
    config = outcome.get_result()

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 279, in get_result
    raise ex[1].with_traceback(ex[2])

  File "D:\Anaconda\envs\tensorflow\lib\site-packages\_pytest\vendored_packages\pluggy.py", line 265, in __init__
    self.result = func()

[...]

PR #1154, opened 2017-09-22
MAINT: add test_file solves #1026

adler-j:
Finally getting this one out of the way.

### DIFF
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -26,7 +26,7 @@ __all__ = ('almost_equal', 'all_equal', 'all_almost_equal', 'never_skip',
            'skip_if_no_pyfftw', 'skip_if_no_largescale',
            'noise_array', 'noise_element', 'noise_elements',
            'Timer', 'timeit', 'ProgressBar', 'ProgressRange',
-           'test', 'run_doctests')
+           'test', 'run_doctests', 'test_file')
 
 
 def _places(a, b, default=None):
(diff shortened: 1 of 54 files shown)


odl/util/testutils.py
```diff
@@ -700,5 +700,22 @@ def run_doctests(skip_if=False, **kwargs):
     testmod(optionflags=optionflags, extraglobs=extraglobs, **kwargs)
 
 
+def test_file(file, args=None):
+    """Run tests in file with proper default arguments."""
+    try:
+        import pytest
+    except ImportError:
+        raise ImportError('ODL tests cannot be run without `pytest` installed.'
+                          '\nRun `$ pip install [--user] odl[testing]` in '
```
kohr-h: Sure we want to always recommend the `pip install` option? People will blindly follow instructions and potentially mess up their environments. I would rather add "or `conda install pytest`" to the message.

And, why so complicated? This utility just needs plain `pytest`, no other plugins.

adler-j: Well pup is always present, competent users will know what to do in other cases.

W.r.t complications, there is simply some stuff we need to do, perhaps I should document them.

Specifically we do a fix to string formatting for Windows, add verbose output and ensure stdout is correctly captured.

kohr-h: > Well pup is always present, competent users will know what to do in other cases.

Beginners in the field increasingly use conda as go-to solution for virtual environments. I feel we should help them by not favoring the slightly wrong solution for that scenario.



> W.r.t complications, there is simply some stuff we need to do, perhaps I should document them.
> 
> Specifically we do a fix to string formatting for Windows, add verbose output and ensure stdout is correctly captured.

That's not what I meant. It was about the `$ pip install [--user] odl[testing]` thing, which installs also `pytest-pep8` and `pytest-cov` which are not needed. So a simple `pip install pytest` will do.

adler-j: > Beginners in the field increasingly use conda as go-to solution for virtual environments. I feel we should help them by not favoring the slightly wrong solution for that scenario.

Well what issues could they expect from that? I've never noticed any at all myself (except for stuff that is compiled ofc)

kohr-h: The most significant one is that pip packages won't be updated unless explicitly updated with `pip -U`. But who does that?
And, if a package is already installed by conda and then a user runs `pip install` with the same package, pip will overwrite the old one but not delete the conda package list entry. Then you have to force-uninstall them later.

So it's not a major thing, but can be an annoyance.

adler-j: Frankly, I feel that stuff like this (while important) should be considered more a "bug" with the ecosystem, and not something we should explicitly handle unless it really creates lots of issues.

My proposal is that we always propose pip unless the packages are compiled, in which case conda should at least be mentioned.

kohr-h: Well, alternatively we could simply say "pleas install package X", not mentioning any tool. At least in those error messages, where we cannot be super-detailed. The docs can clarify.

PR #1157, opened 2017-09-24
Add helical_geometry

adler-j:
This is work in progress but it seems like a nice feature.

### DIFF
--- a/examples/tomo/filtered_backprojection_helical_3d.py
+++ b/examples/tomo/filtered_backprojection_helical_3d.py
@@ -17,29 +17,21 @@ import odl
 
 
 # Reconstruction space: discretized functions on the cube
-# [-20, 20]^2 x [0, 40] with 300 samples per dimension.
-reco_space = odl.uniform_discr(
-    min_pt=[-20, -20, 0], max_pt=[20, 20, 40], shape=[300, 300, 300],
+# [-20, 20]^3  with 200 samples per dimension.
+space = odl.uniform_discr(
+    min_pt=[-20, -20, -20], max_pt=[20, 20, 20], shape=[200, 200, 200],
     dtype='float32')
 
-# Make a helical cone beam geometry with flat detector
-# Angles: uniformly spaced, n = 2000, min = 0, max = 8 * 2 * pi
-# This gives 8 full turns of the helix.
-angle_partition = odl.uniform_partition(0, 8 * 2 * np.pi, 2000)
-# Detector: uniformly sampled with a small height,
-# n = (512, 64), min = (-30, -4), max = (30, 4)
-detector_partition = odl.uniform_partition([-40, -4], [40, 4], [512, 64])
-# Create geometry
-geometry = odl.tomo.ConeFlatGeometry(
-    angle_partition, detector_partition, src_radius=100, det_radius=100,
-    pitch=5.0)
-
+# Create helical geometry
+geometry = odl.tomo.helical_geometry(space,
+                                     src_radius=100, det_radius=100,
+                                     num_turns=7.5, num_angles=1000)
 
 # --- Create Filtered Back-Projection (FBP) operator --- #
 
 
 # Ray transform (= forward projection).
-ray_trafo = odl.tomo.RayTransform(reco_space, geometry)
+ray_trafo = odl.tomo.RayTransform(space, geometry)
 
 # Unwindowed fbp
 # We select a Hamming filter, and only use the lowest 80% of frequencies to
@@ -54,7 +46,7 @@ windowed_fbp = fbp * odl.tomo.tam_danielson_window(ray_trafo)
 
 
 # Create a discrete Shepp-Logan phantom (modified version)
-phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+phantom = odl.phantom.shepp_logan(space, modified=True)
 
 # Create projection data by calling the ray transform on the phantom
 proj_data = ray_trafo(phantom)
--- a/odl/test/tomo/geometry/geometry_test.py
+++ b/odl/test/tomo/geometry/geometry_test.py
@@ -710,5 +710,41 @@ def test_cone_beam_geometry_helper():
     assert geometry.det_partition.extent == pytest.approx(det_width)
 
 
+def test_helical_geometry_helper():
+    """Test that helical_geometry satisfies the sampling conditions.
+
+    See the `helical_geometry` documentation for the exact conditions.
+    """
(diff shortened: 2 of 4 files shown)


examples/tomo/filtered_backprojection_helical_3d.py
```diff
@@ -17,29 +17,21 @@
 
 
 # Reconstruction space: discretized functions on the cube
```
kohr-h: Actually, we could almost remove this whole line. Better version would be
`# Reconstruction space: functions on [-20, 20]^3, uniformly discretized with 200 samples per dimension`

adler-j:
Now ready for review.

Also note that two rather major bug-fixes are part of this PR, first the parker weighting bug and second the miss-use of pytest.approx that seems to have been quite common.

adler-j:
Bump

odl/tomo/geometry/conebeam.py
```diff
@@ -1435,6 +1435,138 @@ def cone_beam_geometry(space, src_radius, det_radius, num_angles=None,
...
+    >>> geometry.angles.size
+    234
+    >>> geometry.detector.shape
+    (57, 10)
+
+    Notes
+    -----
+    See `cone_beam_geometry` for a description on the mathematics in 2d.
+
+    This geometry adds the Tuy condition.
```
kohr-h: Some details?

odl/tomo/geometry/conebeam.py
```diff
@@ -1435,6 +1435,138 @@ def cone_beam_geometry(space, src_radius, det_radius, num_angles=None,
...
+    w = 2 * rho * (rs + rd) / rs
+
+    # Compute minimum number of pixels given the constraint on the
+    # sampling interval and the computed width
+    rb = np.hypot(r, w / 2)  # length of the boundary ray to the flat detector
+    num_px_horiz = 2 * int(np.ceil(w * omega * r / (2 * np.pi * rb))) + 1
+
+    # Compute the vertical size needed in order to get a full sampling
+    # according to the Tuy condition
+    theta = 2 * np.arctan((w / 2) / (src_radius + det_radius))
```
kohr-h: Use `rs` and `rd` for consistency

odl/tomo/geometry/conebeam.py
```diff
@@ -1435,6 +1435,138 @@ def cone_beam_geometry(space, src_radius, det_radius, num_angles=None,
...
+
+    # Compute lower and upper bound
+    source_to_line_distance = src_radius * (1 + np.cos(theta))
+    scale = (src_radius + det_radius) / source_to_line_distance
+
+    source_to_line_lower = (pitch *
+                            (num_turns * np.pi / 2 - theta) / (2 * np.pi))
+    h = 2 * source_to_line_lower * scale
```
kohr-h: Perhaps 2 more words on the geometric relations used here

adler-j:
Possibility for review on the new version?

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -700,15 +700,51 @@ def test_cone_beam_geometry_helper():
...
+
+    # Create object
+    space = odl.uniform_discr([-1, -1, -2], [1, 1, 2], [20, 20, 40])
+    geometry = odl.tomo.helical_geometry(space, src_radius, det_radius,
+                                         num_turns=num_turns)
+
+    # Validate angles
+    assert geometry.motion_partition.is_uniform
+    assert (geometry.motion_partition.extent ==
+            pytest.approx(num_turns * 2 * np.pi))
```
olivierverdier: Same comment: some duplication with `test_cone_beam_geometry_helper` (which is perhaps ok for testing).

odl/tomo/geometry/conebeam.py
```diff
@@ -1435,6 +1435,136 @@ def cone_beam_geometry(space, src_radius, det_radius, num_angles=None,
...
+                         'infinite detector for full coverage')
+    rd = float(det_radius)
+    r = src_radius + det_radius
+    w = 2 * rho * (rs + rd) / rs
+
+    # Compute minimum number of pixels given the constraint on the
+    # sampling interval and the computed width
+    rb = np.hypot(r, w / 2)  # length of the boundary ray to the flat detector
+    num_px_horiz = 2 * int(np.ceil(w * omega * r / (2 * np.pi * rb))) + 1
+
```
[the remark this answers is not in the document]
adler-j: I guess it is also shared with `fan_beam_geometry`, I'll see what I can do. Perhaps I'll simply call `fan_beam_geometr` and pick the relevant parameters from it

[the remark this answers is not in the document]
adler-j: Ok, I've now read this through and I'm not sure if this is a good idea, it seems I would have to move the 2d code out separately and that part is barely ~10 lines or so. I might actually leave it as is for now

odl/test/tomo/geometry/geometry_test.py
adler-j: There is quite some duplication but it seems there is also quite some non-duplication :/ separating them in a natural way might be hard so I guess I'll just leave it for now

adler-j:
had a thorough look at the comments and fixed rebasing from master. Can we go ahead with this now?

odl/test/tomo/geometry/geometry_test.py
kohr-h: Agreed, also with @olivierverdier's comment in parens. Code duplication in tests is more forgivable as long as it's not easy to factor it out into a fixture. The one thing we certainly don't want is a bug in a testing helper 😉

odl/tomo/geometry/conebeam.py
```diff
@@ -1433,6 +1433,158 @@ def cone_beam_geometry(space, src_radius, det_radius, num_angles=None,
...
+    Notes
+    -----
+    In the "fan beam direction", the sampling exactly follows the
+    two-dimensional case see `cone_beam_geometry` for a description.
+
+    In the "axial direction", e.g. along the [0, 0, 1] axis, the geometry is
+    sampled according to two criteria. First, the bounds of the detector
+    are chosen to satisfy the tuy condition.
+    See TSS1998_ for a full description.
```
kohr-h: Can you make that a working link? Like
```
`[TSS1998] <link to online resource>`_
```

kohr-h:
Just fix the citation link, other than that I'm good with this version.

kohr-h:
After rebasing and fixing the link this should go in.

PR #1158, opened 2017-09-25
BUG: fix broken colorbar ticks

kohr-h:
An earlier fix broke the ticks in the updated figures.

### DIFF
--- a/odl/util/graphics.py
+++ b/odl/util/graphics.py
@@ -428,7 +428,7 @@ def show_discrete_data(values, grid, title=None, method='',
                 # If it exists and we should update it
                 csub.colorbar.set_clim(minval, maxval)
                 csub.colorbar.set_ticks(ticks)
-                csub.colorbar.set_ticklabels([fmt.format(*ticks)])
+                csub.colorbar.set_ticklabels([fmt % t for t in ticks])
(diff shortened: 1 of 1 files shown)


odl/util/graphics.py
```diff
@@ -428,7 +428,7 @@ def show_discrete_data(values, grid, title=None, method='',
                 # If it exists and we should update it
                 csub.colorbar.set_clim(minval, maxval)
                 csub.colorbar.set_ticks(ticks)
-                csub.colorbar.set_ticklabels([fmt.format(*ticks)])
+                csub.colorbar.set_ticklabels([fmt % t for t in ticks])
```
adler-j: Why would we want to use old style formatting here?

kohr-h: The `_colorbar_format` helper returns an old-style format string because `plt.colorbar(..., format=fmt)` expects it. So either we make a separate helper or just also use the old-style format here.

adler-j: damnit, I guess we'll have to live with it then

kohr-h: Sad but true.

Is that "Approve"?

adler-j: Sure!

PR #1159, opened 2017-09-25
DOC: Correct typos in the documentation and error messages.

sbanert:
While reading through a part ODL I noticed some typos which should be corrected.


odl/solvers/iterative/iterative.py
```diff
@@ -471,7 +471,7 @@ def kaczmarz(ops, x, rhs, niter, omega=1, projection=None, random=False,
     """
     domain = ops[0].domain
     if any(domain != opi.domain for opi in ops):
-        raise ValueError('`opi[i].domain` are not all equal')
+        raise ValueError('`ops[i].domain` are not all equal')
```
adler-j: It seems this sentence made no sense before either and could use an improvement, e.g. "domain of `ops` not all equal"

PR #1161, opened 2017-09-26
MAINT: unify usage of calls to parent classes, closes #1061

### LINKED ISSUE #1061: Use super(cls, self) instead of super() or hard-coded class?
So far we always use e.g. `super().__init__(...)` in subclasses and get Python 2 compatibility by importing a `super` backport from the `future` library. That is, however, slow, and thus a bunch of speed-critical classes already replace this by hard-coded `ParentClass.__init__(self, ...)` calls.

This is not very pretty and particularly unfriendly towards downstream subclassers. Thus, an alternative could be to use the built-in `super` in both Python 2 and 3 and stick to the (required in Py2, redundant but valid in Py3) syntax `super(cls, self).__init__(...)`.

@adler-j do you have any good test lying around to check the speed of this compared to `super` from `future` and hard-coded parents?

kohr-h:
Getting this out of the way.

## Caveat
There are some cases when `super()` is the wrong choice, most notably with multiple inheritance and "conflicting" methods. Everything is good as long as a certain method only exists once or things stay nicely in order. However, consider the case of `ScalingFunctional` which I stumbled over while doing this change.

We have `class ScalingFunctional(Functional, ScalingOperator)`, with MRO
```python
[odl.solvers.functional.default_functionals.ScalingFunctional,
 odl.solvers.functional.functional.Functional,
 odl.operator.default_ops.ScalingOperator,
 odl.operator.operator.Operator,
 object]
```
So far, so good. The `__init__` method of `ScalingFunctional` even hardcodes the `__init__` calls to both parent classes, not using `super`. Everything should be fine, or?

**Turns out, no**. The crucial part is the common misconception (also on my part) that in an instance method of a class, `self` always refers to an instance of that class. That's **not** true. In this case, when `Functional.__init__` uses `self`, then `self` is an instance of `ScalingFunctional`. In particular, when using `super` to resolve the MRO, it is still the one of `ScalingFunctional`. 

Hence, when `Functional.__init__` calls `super(Functional, self).__init__(...)`, then the `__init__` method that is called is **not** the one of `Operator` (the only parent of `Functional`), but the one of `ScalingOperator`, because it comes first in the MRO.
We can confirm this by printing `self.__class__.mro()` in  `Functional.__init__`:
```python
[<class 'odl.solvers.functional.default_functionals.ScalingFunctional'>,
 <class 'odl.solvers.functional.functional.Functional'>,
 <class 'odl.operator.default_ops.ScalingOperator'>,
 <class 'odl.operator.operator.Operator'>,
 <class 'object'>]
```

### Conclusion

We have to be careful with multiple inheritance the MRO contains potentially conflicting methods. As [the documentation](https://docs.python.org/3/library/functions.html#super) says, `super` is intended for **collaborative** classes, and this situation is not like that.
Sometimes it's hard to tell who will call what. Two ways to mitigate this are:

1. Don't use `super` in classes that **may** participate in non-collaborative multiple inheritance. That applies mostly to core classes. It may be hard to know in advance which classes will be affected.
2. Avoid non-collaborative multiple inheritance as far as possible. In particular, try to keep the class hierarchy linear if possible.

### DIFF
--- a/odl/test/operator/oputils_test.py
+++ b/odl/test/operator/oputils_test.py
@@ -94,7 +94,7 @@ def test_matrix_representation_product_to_product():
     ran_and_dom = ProductSpace(rn, rm)
 
     AB_matrix = np.vstack([np.hstack([A, np.zeros((n, m))]),
-                          np.hstack([np.zeros((m, n)), B])])
+                           np.hstack([np.zeros((m, n)), B])])
     ABop = ProductSpaceOperator([[Aop, 0],
                                  [0, Bop]],
                                 ran_and_dom, ran_and_dom)
@@ -117,7 +117,7 @@ def test_matrix_representation_product_to_product_two():
     ran_and_dom = ProductSpace(rn, 2)
 
     AB_matrix = np.vstack([np.hstack([A, np.zeros((n, n))]),
-                          np.hstack([np.zeros((n, n)), B])])
+                           np.hstack([np.zeros((n, n)), B])])
     ABop = ProductSpaceOperator([[Aop, 0],
                                  [0, Bop]],
                                 ran_and_dom, ran_and_dom)
@@ -148,8 +148,7 @@ def test_matrix_representation_wrong_domain():
         """Small test operator."""
         def __init__(self):
             super(MyOp, self).__init__(
-                domain=ProductSpace(odl.rn(3),
-                                    ProductSpace(odl.rn(3), odl.rn(3))),
+                domain=odl.rn(3) * odl.rn(3) ** 2,
                 range=odl.rn(4),
                 linear=True)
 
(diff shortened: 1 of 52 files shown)


odl/test/operator/oputils_test.py
```diff
@@ -129,52 +128,54 @@ def test_matrix_representation_product_to_product_two():
...
+    class MyOp(odl.Operator):
+        """Small test operator."""
         def __init__(self):
-            super().__init__(domain=ProductSpace(odl.rn(3),
-                                                 ProductSpace(odl.rn(3),
-                                                              odl.rn(3))),
-                             range=odl.rn(4), linear=True)
+            super(MyOp, self).__init__(
+                domain=ProductSpace(odl.rn(3),
+                                    ProductSpace(odl.rn(3), odl.rn(3))),
```
adler-j: odl.rn(3) * (odl.rn(3) * odl.rn(3))

odl/test/operator/oputils_test.py
```diff
@@ -129,52 +128,54 @@ def test_matrix_representation_product_to_product_two():
...
+        """Small test operator."""
         def __init__(self):
-            super().__init__(domain=odl.rn(3),
-                             range=ProductSpace(odl.rn(3),
-                                                ProductSpace(odl.rn(3),
-                                                             odl.rn(3))),
-                             linear=True)
+            super(MyOp, self).__init__(
+                domain=odl.rn(3),
+                range=ProductSpace(odl.rn(3),
```
adler-j: odl.rn(3) * (odl.rn(3) * odl.rn(3))

PR #1163, opened 2017-09-27
Add pointwise estimate noise

adler-j:
It's sometimes nice to get a pointwise noise estimate (e.g. with poisson noise in CT). This allows just that.


adler-j:
It seems this "shrinks" the output by 2 pixels, that should be fixed

adler-j:
Fixed the mentioned bug, ready to review and go in.

PR #1164, opened 2017-09-27
BLD: enhance .travis.yml with more builds and faster failing

### LINKED ISSUE #1162: travis should exit on "first fail"
Specifically we run it with both `pytest` and `odl.test()`. Imo, if the first fails there is no need to run the second and we could go right onto reporting an error. This should speed up the CI cycle somewhat.

kohr-h:
Testing PR.

Closes #1162 ~~when finished.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,13 @@
 language: python
 
-python:
-  - "2.7"
-  - "3.6"
+matrix:
+    include:
+        - python: 2.7
+          env: NUMPY_VERSION=1.12
+
+        - python: 3.6
+          env: NUMPY_VERSION=1.12 COVERALLS=true BUILD_DOCS=true
+
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


kohr-h:
Works, see [this log](https://travis-ci.org/odlgroup/odl/jobs/280634153) for example. Stops after the first failure.

.travis.yml
```diff
@@ -1,8 +1,16 @@
...
-python:
-  - "2.7"
-  - "3.6"
+matrix:
+    include:
+        - python: 2.7
+          env: NUMPY_VERSION=1.12
+
+        - python: 3.6
+          env: NUMPY_VERSION=1.12 NODEPS=true
```
adler-j: Why do we have two 3.6 tests now?

kohr-h: This one tests with only the minimum dependencies installed. I thought it couldn't hurt since we tend to test with our development environment where everything is installed, so errors like unfenced imports aren't caught.

adler-j: I'd say that is overly ambitious, especially as we want these to run asap and also to give easy to read errors, adding two slightly different will confuse most people

kohr-h: Runtime isn't an issue since they run in parallel and the first failure determines the overall result. I can also make the variable more self-explanatory, like `MINIMAL_DEPENDENCIES` or so.

That said, I don't really mind removing the setting. But then I'll probably add an option `COVERALLS` so that only one of the builds runs `coveralls`.

adler-j: Agree that only one needs to run coveralls. Overall I think "micro-managing" like this is somewhat risky since it makes this stuff harder to follow

.travis.yml
```diff
@@ -36,36 +41,39 @@ install:
     - conda info -a
 
     # Install dependencies and enter test environment
-    - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION "numpy<1.13" scipy future nomkl sphinx sphinx_rtd_theme pytest pytest-pep8 pytest-cov
-    - source activate test-environment
-
-    # Some packages which are only on PyPI, not on conda
-    - pip install pyfftw pywavelets travis-sphinx
+    - conda create -n testenv python=$TRAVIS_PYTHON_VERSION nomkl "numpy=$NUMPY_VERSION" scipy future pywavelets scikit-image sphinx sphinx_rtd_theme pytest pytest-pep8 pytest-cov
```
adler-j: Perhaps only install sphinx when building docs? e.g. after the tests

kohr-h: True.

For some odd reason the creation of the initial testenv takes forever (like 2 minutes). Also weird that it downloads mkl, but that doesn't really add anything. Perhaps it's just conda that's slow.

kohr-h: I'll try using pip for the dependencies that do not require compile steps. Maybe that's faster.

kohr-h:
Build runs faster by 1 minute with pip, and the conditionals are in place. Merge?

PR #1165, opened 2017-09-29
DOC: Add The MAX IV imaging concept to the readme

adler-j:
Article that mentions ODL!


kohr-h:
The failures on Travis are caused by pyfftw failing to compile. I suggest we just get the conda package from `conda-forge`: `conda install -c conda-forge pyfftw`.

adler-j:
Agree we should fix conda. Can you do that int he travis branch?

kohr-h:
I already did it there, will get the stuff there fixed and merged.

Issue #1107, opened 2017-08-31
Rename LinearSpace to Space


kohr-h:
As [suggested by @adler-j](https://github.com/odlgroup/odl/pull/1088/files/02109904333d30265d1156ee83108833b4085a4a#diff-f6ddef3abbfa8c9763285b791e19d177).

adler-j:
For obvious reasons I'm in favor of this change

PR #1166, opened 2017-09-29
ENH: Add support for vector-valued stepsizes for SeparableSum proximal

sbanert:
This commit changes the generation of the proximal factory for the class
SeparableSum so that it is possible to choose a stepsize per summand
instead of one stepsize for all. A simple test checks the code.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -651,8 +651,8 @@ class L2NormSquared(Functional):
     >>> space = odl.rn(3)
     >>> f = odl.solvers.L2NormSquared(space)
     >>> x = space.one()
-    >>> f.proximal([0.5, 1.0, 1.5])(x)
-    rn(3).element([ 0.5       ,  0.33333333,  0.25      ])
+    >>> f.proximal([0.5, 1.5, 2.0])(x)
+    rn(3).element([ 0.5 ,  0.25,  0.2 ])
     """
 
     def __init__(self, space):
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -561,6 +561,14 @@ class L1Norm(LpNorm):
 
     .. math::
         \| x \|_1 = \\int_\Omega |x(t)| dt.
+
+    The `proximal` factory allows using vector-valued stepsizes:
+
+    >>> space = odl.rn(3)
+    >>> f = odl.solvers.L1Norm(space)
+    >>> x = space.one()
+    >>> f.proximal([0.5, 1.0, 1.5])(x)
+    rn(3).element([ 0.5,  0. ,  0. ])
     """
 
     def __init__(self, space):
@@ -637,6 +645,14 @@ class L2NormSquared(Functional):
 
     .. math::
         \| x \|_2^2 = \\int_\Omega |x(t)|^2 dt.
+
+    The `proximal` factory allows using vector-valued stepsizes:
+
+    >>> space = odl.rn(3)
+    >>> f = odl.solvers.L2NormSquared(space)
+    >>> x = space.one()
+    >>> f.proximal([0.5, 1.0, 1.5])(x)
+    rn(3).element([ 0.5       ,  0.33333333,  0.25      ])
     """
 
     def __init__(self, space):
@@ -1594,8 +1610,17 @@ class SeparableSum(Functional):
         [\mathrm{prox}_{\\sigma f_1}(x_1),
          \mathrm{prox}_{\\sigma f_2}(x_2),
          ...,
-         \mathrm{prox}_{\\sigma f_n}(x_n)]
+         \mathrm{prox}_{\\sigma f_n}(x_n)].
 
+    If :math:`\\sigma = (\\sigma_1, \\sigma_2, \\ldots, \\sigma_n)` is a list
+    of positive `float`s, then it distributes, too:
+
+    .. math::
+        \mathrm{prox}_{\\sigma h}(x_1, x_2, ..., x_n) =
+        [\mathrm{prox}_{\\sigma_1 f_1}(x_1),
+         \mathrm{prox}_{\\sigma_2 f_2}(x_2),
+         ...,
+         \mathrm{prox}_{\\sigma_n f_n}(x_n)].
     """
 
     def __init__(self, *functionals):
@@ -1616,6 +1641,15 @@ class SeparableSum(Functional):
         >>> l1 = odl.solvers.L1Norm(space)
         >>> l2 = odl.solvers.L2Norm(space)
         >>> f_sum = odl.solvers.SeparableSum(l1, l2)
+
+        The `proximal` factory allows using vector-valued stepsizes:
+
+        >>> x = f_sum.domain.one()
+        >>> f_sum.proximal([0.5, 2.0])(x)
+        ProductSpace(rn(3), 2).element([
+            [ 0.5,  0.5,  0.5],
+            [ 0.,  0.,  0.]
+        ])
 
         Create functional ``f([x1, ... ,xn]) = \sum_i ||xi||_1``:
 
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -159,9 +159,8 @@ def proximal_convex_conj(prox_factory):
 
         mult_inner = MultiplyOperator(1.0 / sigma, domain=space, range=space)
         mult_outer = MultiplyOperator(sigma, domain=space, range=space)
-        result = IdentityOperator(space) \
-            - OperatorComp(OperatorComp(
-                mult_outer, prox_factory(1.0 / sigma)), mult_inner)
+        result = IdentityOperator(space) - OperatorComp(OperatorComp(
+            mult_outer, prox_factory(1.0 / sigma)), mult_inner)
         return result
 
     return convex_conj_prox_factory
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -77,15 +77,19 @@ def combine_proximals(*factory_list):
 
         Parameters
         ----------
-        sigma : positive float
-            Step size parameter
+        sigma : positive float or list of positive floats
+            Step size parameter(s), if a list, the length must match the length of the `factory_list`
+        
 
         Returns
         -------
         diag_op : `DiagonalOperator`
         """
+        if np.isscalar(sigma):
+            sigma = [sigma]*len(factory_list)
+
         return DiagonalOperator(
-            *[factory(sigma) for factory in factory_list])
+            *[factory(sigmai) for (sigmai,factory) in zip(sigma, factory_list)])
 
     return diag_op_factory
 
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -232,7 +232,10 @@ def proximal_arg_scaling(prox_factory, scaling):
         proximal operator of ``F``
     scaling : float or sequence of floats or space element
         Scaling parameter. The permissible types depent on the stepsizes
-        accepted by prox_factory.
+        accepted by prox_factory. It may not contain any nonzero imaginary
+        parts. If it is a scalar, it may be zero, in which case the
+        resulting proxmial operator is the identity. If not a scalar,
+        it may not contain any nonzero components.
 
     Returns
     -------
@@ -262,9 +265,18 @@ def proximal_arg_scaling(prox_factory, scaling):
     2011.
     """
 
-    if scaling.imag != 0 * scaling:
-        raise ValueError("Complex scaling not supported.")
-    if scaling == 0 * scaling:
+    # To begin, we could check for two things:
+    # * Currently, we do not support complex scaling. We could therefore catch
+    #   nonempty imaginary parts.
+    # * If some components of scaling are zero, then the following routine will
+    #   crash with a division-by-zero error. The correct solution would be to
+    #   just keep these components and do the following computations only for
+    #   the others.
+    # Since these checks are computationally expensive, we do not execute them
+    # unconditionally, but only if the scaling factor is a scalar:
+    if np.isscalar(scaling) and scaling.imag != 0:
+        raise ValueError("complex scaling not supported.")
+    if np.isscalar(scaling) and scaling == 0:
         return proximal_const_func(prox_factory(1.0).domain)
 
     def arg_scaling_prox_factory(sigma):
@@ -282,11 +294,6 @@ def proximal_arg_scaling(prox_factory, scaling):
             the step size
         """
         scaling_square = scaling * scaling
-#       TODO Catch division by zero errors. The following doesn't work:
-#       if not scaling_square > 0 * scaling:
-#           raise NotImplementedError("Proximal scaling: "
-#                                     "Vector-valued scaling "
-#                                     "with zeros not yet supported.")
         prox = prox_factory(sigma * scaling_square)
         space = prox.domain
         mult_inner = MultiplyOperator(scaling, domain=space, range=space)
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -896,6 +896,9 @@ def proximal_convex_conj_l2_squared(space, lam=1, g=None):
                     sig.multiply(g, out=out)
                     out.lincomb(1.0, x, -1.0, out=out)
                     out.divide(1 + 0.5 / lam * sig, out=out)
+            else:
+                raise RuntimeError('Error in ProximalConvexConjL2Squared: sig '
+                                   'is neither a scalar nor a space element.')
 
     return ProximalConvexConjL2Squared
 
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -158,8 +158,8 @@ def proximal_convex_conj(prox_factory):
 
         mult_inner = MultiplyOperator(1.0 / sigma, domain=space, range=space)
         mult_outer = MultiplyOperator(sigma, domain=space, range=space)
-        result = IdentityOperator(space) \
-            - mult_outer * prox_factory(1.0 / sigma) * mult_inner
+        result = (IdentityOperator(space) -
+                  mult_outer * prox_factory(1.0 / sigma) * mult_inner)
         return result
 
     return convex_conj_prox_factory
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -887,7 +887,8 @@ def proximal_convex_conj_l2_squared(space, lam=1, g=None):
                 if g is None:
                     x.divide(1 + 0.5 / lam * sig, out)
                 else:
-                    sigma.multiply(g, out)
+                    sig.multiply(g, out)
+                    out.lincomb(1.0, x, -1.0, out)
                     out.divide(1 + 0.5 / lam * sig, out)
 
     return ProximalConvexConjL2Squared
@@ -952,17 +953,28 @@ def proximal_l2_squared(space, lam=1, g=None):
             """
             super(ProximalL2Squared, self).__init__(
                 domain=space, range=space, linear=g is None)
-            self.sigma = float(sigma)
+            if np.isscalar(sigma):
+                self.sigma = float(sigma)
+            else:
+                self.sigma = space.element(sigma)
 
         def _call(self, x, out):
             """Apply the operator to ``x`` and store the result in ``out``"""
             # (x + 2*sig*lam*g) / (1 + 2*sig*lam))
             sig = self.sigma
-            if g is None:
-                out.lincomb(1.0 / (1 + 2 * sig * lam), x)
-            else:
-                out.lincomb(1.0 / (1 + 2 * sig * lam), x,
-                            2 * sig * lam / (1 + 2 * sig * lam), g)
+            if np.isscalar(sig):
+                if g is None:
+                    out.lincomb(1.0 / (1 + 2 * sig * lam), x)
+                else:
+                    out.lincomb(1.0 / (1 + 2 * sig * lam), x,
+                                2 * sig * lam / (1 + 2 * sig * lam), g)
+            else:   # sig in space
+                if g is None:
+                    x.divide(1.0 + 2.0 * sig * lam, out=out)
+                else:
+                    sig.multiply(2.0 * lam * g, out=out)
+                    out.lincomb(1.0, x, 1.0, out)
+                    out.divide(1.0 + 2 * sig * lam, out)
 
     return ProximalL2Squared
 
@@ -1276,7 +1288,10 @@ def proximal_l1(space, lam=1, g=None):
             """
             super(ProximalL1, self).__init__(
                 domain=space, range=space, linear=False)
-            self.sigma = float(sigma)
+            if np.isscalar(sigma):
+                self.sigma = float(sigma)
+            else:
+                self.sigma = space.element(sigma)
 
         def _call(self, x, out):
             """Return ``self(x, out=out)``."""
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -231,8 +231,9 @@ def proximal_arg_scaling(prox_factory, scaling):
     prox_factory : callable
         A factory function that, when called with a step size, returns the
         proximal operator of ``F``
-    scaling : float
-        Scaling parameter
+    scaling : float or list of floats or space element
+        Scaling parameter. The permissible types depent on the stepsizes
+        accepted by prox_factory.
 
     Returns
     -------
@@ -262,11 +263,9 @@ def proximal_arg_scaling(prox_factory, scaling):
     2011.
     """
 
-    if scaling.imag != 0:
-        raise ValueError("complex scaling not supported.")
-    else:
-        scaling = float(scaling.real)
-    if scaling == 0:
+    if scaling.imag != 0 * scaling:
+        raise ValueError("Complex scaling not supported.")
+    if scaling == 0 * scaling:
         return proximal_const_func(prox_factory(1.0).domain)
 
     def arg_scaling_prox_factory(sigma):
@@ -283,8 +282,17 @@ def proximal_arg_scaling(prox_factory, scaling):
             The proximal operator of ``sigma * F( . * a)`` where ``sigma`` is
             the step size
         """
-        prox = prox_factory(sigma * scaling ** 2)
-        return (1 / scaling) * prox * scaling
+        scaling_square = scaling * scaling
+# TODO Catch division by zero errors. The following doesn't work:
+#        if not scaling_square > 0 * scaling:
+#            raise NotImplementedError("Proximal scaling: "
+#                                      "Vector-valued scaling "
+#                                      "with zeros not yet supported.")
+        prox = prox_factory(sigma * scaling_square)
+        space = prox.domain
+        mult_inner = MultiplyOperator(scaling, domain=space, range=space)
+        mult_outer = MultiplyOperator(1 / scaling, domain=space, range=space)
+        return OperatorComp(OperatorComp(mult_outer, prox), mult_inner)
 
     return arg_scaling_prox_factory
 
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -885,11 +885,11 @@ def proximal_convex_conj_l2_squared(space, lam=1, g=None):
                                 -sig / (1 + 0.5 * sig / lam), g)
             elif sig in space:
                 if g is None:
-                    x.divide(1 + 0.5 / lam * sig, out)
+                    x.divide(1 + 0.5 / lam * sig, out=out)
                 else:
-                    sig.multiply(g, out)
-                    out.lincomb(1.0, x, -1.0, out)
-                    out.divide(1 + 0.5 / lam * sig, out)
+                    sig.multiply(g, out=out)
+                    out.lincomb(1.0, x, -1.0, out=out)
+                    out.divide(1 + 0.5 / lam * sig, out=out)
 
     return ProximalConvexConjL2Squared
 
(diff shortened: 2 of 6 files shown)


adler-j:
To run tests locally:

    pytest --pep8

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -77,15 +77,19 @@ def diag_op_factory(sigma):
...
         Returns
         -------
         diag_op : `DiagonalOperator`
         """
+        if np.isscalar(sigma):
+            sigma = [sigma]*len(factory_list)
+
         return DiagonalOperator(
-            *[factory(sigma) for factory in factory_list])
+            *[factory(sigmai) for (sigmai,factory) in zip(sigma, factory_list)])
```
adler-j: heres the bug

adler-j:
This looks OK in isolation, but I'd very much like to have an example where this is used in a solver.

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -505,6 +505,23 @@ def test_moreau_envelope_l2_sq(space, sigma):
     assert all_almost_equal(smoothed_l2_sq.gradient(x),
                             x * 2 / (1 + 2 * sigma))
 
+def test_weighted_proximal():
```
adler-j: Use a fixture here, e.g.

```
def test_weighted_proximal(space):
```

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -505,6 +505,23 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    """Test for the weighted proximal of a SeparableSum functional."""
+
+    space = odl.space.rn(2)
```
adler-j: `rn` is exposed to the top level, so we should use `odl.rn`

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -505,6 +505,23 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+
+    l1 = odl.solvers.L1Norm(space)
+    l2 = odl.solvers.L2Norm(space)
+
+    x = space.one()
```
adler-j: Weird ordering, try to do it "in sequence" so to say:

* define space
* define operators/functionals
* define values
* call operator on values

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -505,6 +505,23 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    y = space.one()
+
+    func = odl.solvers.SeparableSum(l1, l2)
+
+    sigma = [0.5, 1.0]
+
+    prox = func.proximal(sigma)([x, y])
+    assert all_almost_equal(prox, [l1.proximal(0.5)(x), l2.proximal(1.0)(y)])
```
adler-j: this violates [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  Either do simply

```python
prox = func.proximal([0.5, 1.0])([x, y])
expected = [l1.proximal(0.5)(x), l2.proximal(1.0)(y)])
assert all_almost_equal(prox, expected)       
```

or go for

```python
x = func.domain.one()
sigma = [0.5, 1.0]

prox = func.proximal(sigma)(x)
expected = [l1.proximal(sigma[0])(x[0]), l2.proximal(sigma[1])(x[1])])
assert all_almost_equal(prox, expected)       
```

sbanert:
Support of (diagonally) weighted proximals for the L2_norm_squared functional is implemented together with a simple test. I am not completely satisfied with the code (e.g. I had to remove the array(0.5) from the fixture of sigma, because numpy does not see this as a scalar). Please review the code so that I can start to adapt the L1_norm as well.

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -544,7 +544,7 @@ def test_weighted_proximal_L2_norms_squared(space):
     assert all_almost_equal(p1, p2)
 
     # Check if the subdifferential inequalities are satisfied.
-    # p = prox_{σf}(x) ⇔ (x - p)/σ = ∇f(p)
+    # p = prox_{sigma * f}(x) iff (x - p)/sigma = grad f(p)
```
kohr-h: Agreed that we should use ASCII when possible, to avoid hassle. But when we e.g. cite authors with non-ASCII characters in their names, I don't think the best solution would be to misspell their names. 
Instead, according to [PEP 263](https://www.python.org/dev/peps/pep-0263/) one can put a comment `# coding: utf-8` (I'd prefer this style) into the *first or second line* of the file so that at least parsing the source works for Python 2.7 and 3. 
The characters may still come out wrong in strings, but that can be fixed with `from __future__ import unicode_literals`. This option is a bit dangerous, though, especially when passing strings to other libraries.

adler-j: I strongly oppose using unicode in code as long as we support python 2. We're bound to get problems at some point from doing this, not to mention the irritating bugs one can get from unicode characters that look EXACTLY the same

kohr-h: Docstrings are not code, at least not Python code. And PEP 263 is implemented since Python 2.3.

adler-j: Still sceptical. Editor support for Unicode is sketchy at best.

Either way, if we're going to allow Unicode that should have its own issue and discussion and not be added as an afterthought in some PR.

If it can safely be done, I'm all for it. But we need to check all major editors etc. We also need rules on where we allow it (i.e. only in docstrings?)

kohr-h: Sure, this is not the right place for discussion. 

But anyway, what I wrote is not about editors at all. It is about the Python interpreter that has to process a code file, and that can be reliably steered with the comment according to the PEP.

See #1207

sbanert:
Now the fancy stepsizes are implemented and tested for L2 norm squared, L1 norm and their respective conjugates as well as for `SeparableSum`s. The only thing I am currently unsure about is the `isotropic` flag of the L1 norm.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -138,17 +142,27 @@ def convex_conj_prox_factory(sigma):
...
-        prox_other = sigma * prox_factory(1.0 / sigma) * (1.0 / sigma)
-        return IdentityOperator(prox_other.domain) - prox_other
+
+        # Get the underlying space. At the same time, check if the given
+        # prox_factory accepts stepsize objects of the type given by sigma.
+        space = prox_factory(sigma).domain
+
+        mult_inner = MultiplyOperator(1.0 / sigma, domain=space, range=space)
+        mult_outer = MultiplyOperator(sigma, domain=space, range=space)
+        result = IdentityOperator(space) \
```
adler-j: Overall we prefer to use parentheses instead of \, this is mostly since it makes the code easier to read.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -270,8 +282,17 @@ def arg_scaling_prox_factory(sigma):
             The proximal operator of ``sigma * F( . * a)`` where ``sigma`` is
             the step size
         """
-        prox = prox_factory(sigma * scaling ** 2)
-        return (1 / scaling) * prox * scaling
+        scaling_square = scaling * scaling
+# TODO Catch division by zero errors. The following doesn't work:
```
adler-j: Comments should be on the same indent level as other code

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -839,22 +862,34 @@ def __init__(self, sigma):
...
-                            -sig / (1 + 0.5 * sig / lam), g)
+            if np.isscalar(sig):
+                if g is None:
+                    out.lincomb(1.0 / (1 + 0.5 * sig / lam), x)
+                else:
+                    out.lincomb(1.0 / (1 + 0.5 * sig / lam), x,
+                                -sig / (1 + 0.5 * sig / lam), g)
+            elif sig in space:
+                if g is None:
+                    x.divide(1 + 0.5 / lam * sig, out)
```
adler-j: prefer to use `out=out`, easier to read

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+
+def test_weighted_proximal_L2_norm_squared(space):
+    """Test for the weighted proximal of the squared L2 norm"""
+
+    # Define the functional on the space.
+    func = odl.solvers.L2NormSquared(space)
+
+    # Set the stepsize as a random element of the spaces
+    # with elements between 1 and 10.
+    sigma = odl.phantom.noise.uniform_noise(space, 1, 10)
```
adler-j: Obviously a bug here!

`uniform_noise` should be added here:

https://github.com/odlgroup/odl/blob/6b73551e2287dc80315d455eb14a53c88c2addbe/odl/phantom/noise.py#L17

and you should simply use `odl.phantom.uniform_noise`

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+
+    # Start at the one vector.
+    x = space.one()
+
+    # Calculate the proximal point inline and non-inline
```
adler-j: inline means "in the same line", the word you are looking for here is in-place, and the other case would be out-of-place

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    p1 = space.element()
+    func.proximal(sigma)(x, out=p1)
+    p2 = func.proximal(sigma)(x)
+
+    # Both should contain the same vector now.
+    assert all_almost_equal(p1, p2)
+
+    # Check if the subdifferential inequalities are satisfied.
+    # p = prox_{sigma * f}(x) iff (x - p)/sigma = grad f(p)
+    assert all_almost_equal(func.gradient(p1), space.divide(x - p1, sigma))
```
adler-j: Tests can be written without using divide etc, just do

```python
assert all_almost_equal(func.gradient(p1),
                        (x - p1) / sigma)
```

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    """Test for the weighted proximal of the L1 norm"""
+
+    # Define the functional on the space.
+    func = odl.solvers.L1Norm(space)
+
+    # Set the stepsize as a random element of the spaces
+    # with elements between 1 and 10.
+    sigma = odl.phantom.noise.uniform_noise(space, 1, 10)
+
+    # Start for away from zero so that the L1 norm will be differentiable
```
adler-j: for -> far

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    p1 = space.element()
+    func.proximal(sigma)(x, out=p1)
+    p2 = func.proximal(sigma)(x)
+
+    # Both should contain the same vector now.
+    assert all_almost_equal(p1, p2)
+
+    # Check if the subdifferential inequalities are satisfied.
+    # p = prox_{sigma * f}(x) iff (x - p)/sigma = grad f(p)
+    print(func.gradient(p1))
```
adler-j: leftover print

adler-j:
While its fresh: Always avoid comments like "MAINT: Fix the easy issues of Jonas' review."

A commit comment should detail what happens inside the commit, not how it came about. They are mainly intended for yourself in 2 years, and at that point you will not know what Jonas wanted 2 years ago (especially since there is not even a reference to go find what I wanted).

Instead go for commits like "MAINT: Add uniform_noise to __all__" which describes what happens.

sbanert:
Corrected everything except for the doctests. Notice that in one of the references the name "Lemaréchal" appeared. I could choose between making a spelling mistake and adding the utf-8 encoding in the beginning of the corresponding file. I decided for the latter alternative, see #1207.

sbanert:
Doctests are now added, too. I think that everything is done now.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -24,12 +24,11 @@
...
-from odl.operator import (
-    Operator, IdentityOperator, ScalingOperator, ConstantOperator,
-    DiagonalOperator, PointwiseNorm)
-from odl.set.space import LinearSpaceElement
+from odl.operator import (Operator, IdentityOperator, ScalingOperator,
+                          ConstantOperator, DiagonalOperator, PointwiseNorm,
+                          OperatorComp, MultiplyOperator)
 from odl.space import ProductSpace
-from odl.util import cache_arguments
+from odl.set.space import LinearSpaceElement
```
adler-j: This is all using the old master, you need to `git rebase master` to use the latest version

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -77,15 +76,20 @@ def diag_op_factory(sigma):
...
         -------
         diag_op : `DiagonalOperator`
         """
+        if np.isscalar(sigma):
+            sigma = [sigma] * len(factory_list)
+
         return DiagonalOperator(
-            *[factory(sigma) for factory in factory_list])
+            *[factory(sigmai)
+              for (sigmai, factory) in zip(sigma, factory_list)])
```
adler-j: parentheses around `(sigmai, factory)` not needed

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -248,14 +262,11 @@ def proximal_arg_scaling(prox_factory, scaling):
     2011.
     """
 
-    if scaling.imag != 0:
-        raise ValueError("complex scaling not supported.")
-    else:
-        scaling = float(scaling.real)
-    if scaling == 0:
+    if scaling.imag != 0 * scaling:
```
adler-j: If scaling is a list, this gives

```
0 * [1, 2, 3] = []
```

etc. So the list has to be converted to a vector I guess?

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -217,8 +230,9 @@ def proximal_arg_scaling(prox_factory, scaling):
     prox_factory : callable
         A factory function that, when called with a step size, returns the
         proximal operator of ``F``
-    scaling : float
-        Scaling parameter
+    scaling : float or list of floats or space element
```
adler-j: sequence of floats

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -77,15 +76,20 @@ def diag_op_factory(sigma):
 
         Parameters
         ----------
-        sigma : positive float
-            Step size parameter
+        sigma : positive float or list of positive floats
```
adler-j: sequence of positive floats

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -270,8 +281,17 @@ def arg_scaling_prox_factory(sigma):
             The proximal operator of ``sigma * F( . * a)`` where ``sigma`` is
             the step size
         """
-        prox = prox_factory(sigma * scaling ** 2)
-        return (1 / scaling) * prox * scaling
+        scaling_square = scaling * scaling
+#       TODO Catch division by zero errors. The following doesn't work:
```
adler-j: Why not add some sort of check for that? It would at least improve the error message!

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -839,22 +861,34 @@ def __init__(self, sigma):
 
             Parameters
             ----------
-            sigma : positive float
-                Step size parameter
+            sigma : positive float and componentwise positive space.element
```
adler-j: componentwise -> pointwise?

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -839,22 +861,34 @@ def __init__(self, sigma):
...
+                else:
+                    out.lincomb(1.0 / (1 + 0.5 * sig / lam), x,
+                                -sig / (1 + 0.5 * sig / lam), g)
+            elif sig in space:
+                if g is None:
+                    x.divide(1 + 0.5 / lam * sig, out=out)
+                else:
+                    sig.multiply(g, out=out)
+                    out.lincomb(1.0, x, -1.0, out=out)
+                    out.divide(1 + 0.5 / lam * sig, out=out)
```
adler-j: Else? The rule is in general that if you have an "else" case that should never happen, you add `RuntimeException('This should not occur')` or something like it, otherwise this can give silent errors

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
                             x * 2 / (1 + 2 * sigma))
 
 
+def test_weighted_proximal(space):
```
adler-j: test_weighted_proximal  -> test_weighted_separablesum

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+
+    # Set the stepsize as a random element of the spaces
+    # with elements between 1 and 10.
+    sigma = odl.phantom.uniform_noise(space, 1, 10)
+
+    # Start at the one vector.
+    x = space.one()
+
+    # Calculate the proximal point in-place and out-of-place
+    p1 = space.element()
```
adler-j: Prefer `p_ip` for in place and `p_oop` for out of place (or something like it), 1 and 2 does not convey much information

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,75 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    assert all_almost_equal(p1, p2)
+
+    # Check if the subdifferential inequalities are satisfied.
+    # p = prox_{sigma * f}(x) iff (x - p)/sigma = grad f(p)
+    assert all_almost_equal(func.gradient(p1),
+                            (x - p1) / sigma)
+
+
+def test_weighted_proximal_L1_norm(space):
+    """Test for the weighted proximal of the L1 norm"""
```
adler-j: This (and the above) function only tests in the "high x domain", e.g. it only validates that the correct behaviour occurs in the linear domain of the l1-proximal. However, we would also need to know that the scaling works in the non-linear domain (basically in [-sigma, sigma]), so the test has to somehow include that. The easiest way is to compare against a manually computed result I guess.

odl/test/solvers/nonsmooth/proximal_operator_test.py
```diff
@@ -220,22 +220,28 @@ def test_proximal_convconj_l2_sq_wo_data():
...
     prox = prox_factory(sigma)
+    proxv = prox_factory(sigmav)
 
     assert isinstance(prox, odl.Operator)
+    assert isinstance(proxv, odl.Operator)
 
-    # Allocate output element
+    # Allocate output elements
     x_out = space.element()
+    x_out2 = space.element()
```
adler-j: stay consistent in naming, e.g. if you use sigmav above, go for x_outv here

odl/test/solvers/functional/default_functionals_test.py
kohr-h: Although `p_oop` looks funny if you leave out the underscore 😉

adler-j: woop_s

odl/solvers/nonsmooth/proximal_operators.py
sbanert: What do you mean? I rebased on master and after removing this line, I get 29 pytest errors, e.g., NameError: name 'LinearSpaceElement' is not defined.

By the way, this line is not new in the code, see
https://github.com/odlgroup/odl/blob/master/odl/solvers/nonsmooth/proximal_operators.py

adler-j: `LinearSpaceElement` was replaced by `Tensor`, so the code needs to be updated wr.t. that.

kohr-h: > LinearSpaceElement was replaced by Tensor, so the code needs to be updated wr.t. that.

Err, no. `FnBaseVector` was replaced by `Tensor`.

adler-j: Edit: it should be `SpaceElement`

kohr-h: We didn't make that change yet, #1107 is still open.

odl/solvers/nonsmooth/proximal_operators.py
sbanert: Hmmm, the problem here is that for a SeparableSum functional, the “canonical” choice for a stepsize would be a list, which cannot be converted to a vector.

odl/solvers/functional/default_functionals.py
```diff
@@ -637,6 +645,14 @@ class L2NormSquared(Functional):
...
     .. math::
         \| x \|_2^2 = \\int_\Omega |x(t)|^2 dt.
+
+    The `proximal` factory allows using vector-valued stepsizes:
+
+    >>> space = odl.rn(3)
+    >>> f = odl.solvers.L2NormSquared(space)
+    >>> x = space.one()
+    >>> f.proximal([0.5, 1.0, 1.5])(x)
+    rn(3).element([ 0.5       ,  0.33333333,  0.25      ])
```
kohr-h: Please use numbers that can be computed and represented exactly, e.g. `[0.5, 1.5, 2.0]` I guess. Printing approximate numbers may depend on Python version and Numpy version and ruin the doctest.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -77,15 +76,20 @@ def diag_op_factory(sigma):
 
         Parameters
         ----------
-        sigma : positive float
-            Step size parameter
+        sigma : positive float or sequence of positive floats
+            Step size parameter(s), if a sequence, the length must match
+            the length of the `factory_list`
```
kohr-h: dot at the end

kohr-h: \`factory_list\` is a dead link, use double ticks

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -138,17 +142,26 @@ def convex_conj_prox_factory(sigma):
 
         Parameters
         ----------
-        sigma : positive float
-            Step size parameter
+        sigma : positive float or componentwise positive space element
```
kohr-h: pointwise

This should fit on one line. How about just speaking of `positive float or array-like` and describe what cases of arrays are allowed?

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -138,17 +142,26 @@ def convex_conj_prox_factory(sigma):
...
-        return IdentityOperator(prox_other.domain) - prox_other
+
+        # Get the underlying space. At the same time, check if the given
+        # prox_factory accepts stepsize objects of the type given by sigma.
+        space = prox_factory(sigma).domain
+
+        mult_inner = MultiplyOperator(1.0 / sigma, domain=space, range=space)
+        mult_outer = MultiplyOperator(sigma, domain=space, range=space)
+        result = IdentityOperator(space) - OperatorComp(OperatorComp(
+            mult_outer, prox_factory(1.0 / sigma)), mult_inner)
```
kohr-h: `OperatorComp` should never be necessary, just use the `*` operator.

odl/solvers/nonsmooth/proximal_operators.py
kohr-h: I would not check this at all. You also allow `scaling` to be a space element, and checking this can be quite expensive. Also this check will break down for arrays with the message
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -248,14 +262,11 @@ def proximal_arg_scaling(prox_factory, scaling):
...
     """
 
-    if scaling.imag != 0:
-        raise ValueError("complex scaling not supported.")
-    else:
-        scaling = float(scaling.real)
-    if scaling == 0:
+    if scaling.imag != 0 * scaling:
+        raise ValueError("Complex scaling not supported.")
+    if scaling == 0 * scaling:
```
kohr-h: Same applies here. And why would this be equivalent to scaling with 1.0 anyway?

odl/solvers/nonsmooth/proximal_operators.py
kohr-h: As long as space elements are allowed as inputs, I am against doing this kind of (potentially expensive) checks unconditionally.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -839,22 +861,34 @@ def __init__(self, sigma):
 
             Parameters
             ----------
-            sigma : positive float
-                Step size parameter
+            sigma : positive float and pointwise positive space.element
```
kohr-h: and -> or
Also style is different from other places

odl/solvers/nonsmooth/proximal_operators.py
kohr-h: But if you like you can also do a check only if you have "small stuff", like a scalar or a short list of scalings. In that case the documentation should mention this "inconsistency for performance reasons".

sbanert: I removed all the checks, except for those stuff which satisfies np.isscalar. The latter is necessary to avoid regressions, because in the master branch, scaling with zero returned the identity operator.

odl/solvers/nonsmooth/proximal_operators.py
sbanert: Nothing changed here.

sbanert:
I addressed all the requested changes.

odl/solvers/nonsmooth/proximal_operators.py
kohr-h: bump

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -217,8 +229,12 @@ def proximal_arg_scaling(prox_factory, scaling):
...
         A factory function that, when called with a step size, returns the
         proximal operator of ``F``
-    scaling : float
-        Scaling parameter
+    scaling : float or sequence of floats or space element
+        Scaling parameter. The permissible types depent on the stepsizes
+        accepted by prox_factory. It may not contain any nonzero imaginary
+        parts. If it is a scalar, it may be zero, in which case the
+        resulting proxmial operator is the identity. If not a scalar,
+        it may not contain any nonzero components.
```
kohr-h: zero?

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -248,14 +264,20 @@ def proximal_arg_scaling(prox_factory, scaling):
...
+    #   the others.
+    # Since these checks are computationally expensive, we do not execute them
+    # unconditionally, but only if the scaling factor is a scalar:
+    if np.isscalar(scaling) and scaling.imag != 0:
         raise ValueError("complex scaling not supported.")
-    else:
-        scaling = float(scaling.real)
-    if scaling == 0:
+    if np.isscalar(scaling) and scaling == 0:
         return proximal_const_func(prox_factory(1.0).domain)
```
kohr-h: You need an `else` with conversion to an array here, otherwise [this line](https://github.com/odlgroup/odl/pull/1166/files#diff-96685646d224192b17008ecba5b0eab1R295) will crash.

kohr-h: Suggestion:
```python
if np.isscalar(scaling):
    if scaling == 0:
        return proximal_const[...]
    elif scaling.imag != 0:
        raise ...
    else:
        scaling = float(scaling)
else:
    scaling = np.asarray(scaling)
```

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -248,14 +264,20 @@ def proximal_arg_scaling(prox_factory, scaling):
...
 
-    @cache_arguments
```
kohr-h: Instead of removing, can we make this more specific so it stores only "simple" types? If not now, then making a new issue would be the way to go.

odl/solvers/nonsmooth/proximal_operators.py
sbanert: To be honest, I am not sure what @adler-j means. Is it the line break? Should I put the product in parentheses and break the line in between?

adler-j: Yeah, we typically use

```
(looooooooooooooooooooooooooooooooooooooong + 
 liiiiiiiiiiiiiiiiiiinnnnnnnnnnnnnnneeeeeeeeeeeeeeee)
```

over

```
(looooooooooooooooooooooooooooooooooooooong \
 liiiiiiiiiiiiiiiiiiinnnnnnnnnnnnnnneeeeeeeeeeeeeeee)
```

odl/solvers/nonsmooth/proximal_operators.py
sbanert: Done, see #1266.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -839,22 +872,37 @@ def __init__(self, sigma):
...
+                if g is None:
+                    x.divide(1 + 0.5 / lam * sig, out=out)
+                else:
+                    sig.multiply(g, out=out)
+                    out.lincomb(1.0, x, -1.0, out=out)
+                    out.divide(1 + 0.5 / lam * sig, out=out)
             else:
-                out.lincomb(1.0 / (1 + 0.5 * sig / lam), x,
-                            -sig / (1 + 0.5 * sig / lam), g)
+                raise RuntimeError('Error in ProximalConvexConjL2Squared: sig '
```
adler-j: "Error in" is redundant, an error is telling by itself.

PR #1167, opened 2017-10-01
WIP: Torch tensor space

kohr-h:
Since I'm going to work more with torch, I put some time into wrapping torch tensors into an ODL space. The module itself is largely finished, see TODOs.

With release 0.2, pytorch got broadcasting semantics and (almost) fancy indexing, so it can almost completely replace a Numpy array for the basic stuff. Most ufuncs are also available, although the syntax is different, and the accumulation and reduction methods need small custom wrapper to make for example `np.add.accumulate` use `torch.cumsum`.

What is a bit unfortunate is that the interoperability with Numpy is a bit limited. For instance, assigning a torch tensor with a Numpy array doesn't work: `tensor[:] = npy_arr` will fail. The same is true BTW for sequences.
I have improved the situation internally by simply monkey-patching the `__setitem__` method into all tensor classes 😇, and I've also added the methods `__array__` and `__array_wrap__` which are not there either.
(The array methods may actually be worth a PR in pytorch, but I'm not so sure about the setitem thing.)

A quite cool thing about torch is that you can transparently use pinned memory and asynchronous data transfer. With that we could potentially implement some Fessler-style super-fast TV reconstruction (or even more general).

Anyway, **TODOs**:
- [ ] Fix documentation
- [ ] Write minimal ufunc wrapper for reductions and accumulations
- [x] Remove remaining pygpu stuff
- [ ] Fix tests (only 18 failures on first attempt after fixing all the doctests!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ addons:
             texlive-latex-extra
             dvipng
 
-# Enable cache for $HOME/.cache/pip (wheels and such)
-cache: pip
 
 # Setup numpy + scipy using miniconda
 # See http://conda.pydata.org/docs/travis.html
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,13 @@ language: python
 matrix:
     include:
         - python: 2.7
+          env: NUMPY_VERSION=1.13
+
+        - python: 3.6
           env: NUMPY_VERSION=1.12
 
         - python: 3.6
-          env: NUMPY_VERSION=1.12 COVERALLS=true BUILD_DOCS=true
-
+          env: NUMPY_VERSION=1.13 COVERALLS=true BUILD_DOCS=true
 
 sudo: false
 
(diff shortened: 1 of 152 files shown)


adler-j:
Looking great, but: get the tensor branch in! :D we all want it :)

kohr-h:
> Looking great, but: get the tensor branch in! :D we all want it :)

You mean before CUDA replacement is ready? Okay, I can basically finish it off.

adler-j:
Basically yes. There are several feature that will build on that one and we need to get it battle tested so to say.

kohr-h:
So now I'm down to 1 failure for the unit tests, but they don't test GPU stuff so far. Getting to that. But I needed to increase the evilness and also monkey-patch `__getitem__`, although only slightly. There are these irritating little differences to Numpy which makes life a bit harder than need be.

I think some of these patches will have to stay in place one way or another since I imagine that not all of them would be welcome upstream. But I'm going to open an issue there.

adler-j:
Why close this?

kohr-h:
It got automatically closed as the base branch went away. I'll open a new PR into master.

Issue #1147, opened 2017-09-21
ODL pytest plugin interferes with foreign packages


kohr-h:
Unfortunately we may need to roll back some of the changes in #692. The `entry_points` setuptools machinery for the custom pytest plugin is installed as permanent hook that is used even when one runs tests in a different repository.

This is not quite the intention of these plugins, and it gets specifically hairy when some experimental package (like `dipp`) registers its broken plugin and makes testing ODL impossible.

Let me again write down the required and optional features of our unit test framework:

### Required
1. Distributed together with the code, since otherwise `odl.test()` won't work.
2. Skips tests of optional backends if they're not installed (`astra`, `pyfftw`, etc.).
3. Runs doctests, but only those where the backends are available.
4. All supported tests are run when invoking `pytest odl` from top level.

### Optional
5. All supported tests are run when invoking `pytest` from top level, skipping obvious non-test files like `setup.py` and everything in `examples`.

### Consequences
1. All test files and configuration must be inside the `odl` package folder.
2. The custom fixtures and `collect_ignore` things must be loaded at startup. This requires that [one of the available mechanisms](https://docs.pytest.org/en/latest/writing_plugins.html#plugin-discovery-order-at-tool-startup) must be used.
3. Since configurability for doctests is a bit limited, unsupported parts of the library must be excluded from collection at startup, otherwise the test command fails. Also the `doctest_namespace` fixture must be loaded.
4. This works when there is a file `odl/conftest.py` file and `odl/pytest.ini`.
5. This works only with the `entry_points` solution as it seems. For the other possibility, `conftest.py` and `pytest.ini` at top level, the command `pytest` still goes into the `examples` directory and clearly fails to respect the `testpaths` variable in `pytest.ini` and the `collect_ignore` thing in `conftest.py`.

### Suggestion
There must be a better way to do this, but for the short term, my suggestion is that we move `odl/util/pytest_plugins.py` to `odl/conftest.py`, remove the `entry_points` thing in `setup.py` and deal with the fact that we have to run `pytest odl` instead of just `pytest`.

adler-j:
Short comment: How hard would it be to detect inside `pytest_plugins.py` if it is getting called from an  ODL folder? I.e. only install the odl pytest plugins when pytest is run with ODL?

kohr-h:
That's too late, the plugin is installed when running `pip install`. It's a `setuptools` entry point.

adler-j:
[...]

PR #1170, opened 2017-10-02
WIP: change pytest setup

kohr-h:
This is a "discussion PR" for issue #1147 regarding conflicts of pytest plugins.

**Current status:** Requires `pytest odl` to run tests. Just `pytest` fails.

What I think we should do irrespective of the overall issue is to move the `test_examples.py` and `test_doc.py` out of `odl/test` to the `examples/` and `doc/`directories, respectively, and supply them with a custom `conftest.py`. That way, testing only the examples or only the docs becomes easier, namely one only needs `pytest examples` and `pytest doc`, respectively, without the extra `--examples` or `--doctest-doc` command line options.


doc/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+
+from odl.trafos.backends import PYFFTW_AVAILABLE, PYWT_AVAILABLE
+from odl.tomo.backends import (
+    ASTRA_AVAILABLE, ASTRA_CUDA_AVAILABLE, SKIMAGE_AVAILABLE)
+
+# We don't bother excluding stuff depending on available backends. We just
+# require them all (except stir)
+if not all([PYFFTW_AVAILABLE, PYWT_AVAILABLE,
+            ASTRA_AVAILABLE, ASTRA_CUDA_AVAILABLE, SKIMAGE_AVAILABLE]):
+    print('to be able to test the docs, you need to install all backends, '
```
adler-j: Why not just raise an exception?

doc/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+
+import os
+import sys
+
+from odl.trafos.backends import PYFFTW_AVAILABLE, PYWT_AVAILABLE
+from odl.tomo.backends import (
+    ASTRA_AVAILABLE, ASTRA_CUDA_AVAILABLE, SKIMAGE_AVAILABLE)
+
+# We don't bother excluding stuff depending on available backends. We just
+# require them all (except stir)
```
adler-j: We should just move stir to contrib asap :/

doc/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+    print('to be able to test the docs, you need to install all backends, '
+          'currently: '
+          '`astra`, `pyfftw`, `pywt`, `skimage`; '
+          'please consult the installation documentation for details',
+          file=sys.stderr)
+    sys.exit(-1)
+
+here = os.path.normpath(os.path.dirname(__file__))
+collect_ignore = [os.path.join(here, 'sphinxext')]
+collect_ignore = [os.path.normcase(ignored) for ignored in collect_ignore]
```
adler-j: Dislike overwriting names like this.

doc/test_doc.py
```diff
@@ -39,9 +39,8 @@
 doctest_extraglobs = {'odl': odl, 'np': numpy, 'scipy': scipy,
```
adler-j: Is this file still needed if we move to `pytest doc`?

examples/conftest.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Test configuration file."""
```
adler-j: This needs to be improved. Also add a mention of this file to the readme

doc/conftest.py
kohr-h: What's wrong with it? When it's 100 lines apart, okay, but two consecutive lines? Why use a new name?

adler-j:
```
collect_ignore = [os.path.normcase(os.path.join(here, 'sphinxext'))]
```

whats wrong with that?

kohr-h: Nothing, but how about other files with more entries in this list? Use `os.path.normcase(os.path.join())` for each entry? Seems like a lot of noise to me.

PR #1172, opened 2017-10-04
MAINT: add pep8speaks config file

kohr-h:
Let's see what happens :-)

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,20 @@
+message:  # Customize the comment made by the bot
+    opened:  # Messages when a new PR is submitted
+        # Just show something if there are violations
+        header: ""
+        footer: ""
+    updated:  # Messages when new commits are added to the PR
+        header: ""
+        footer: ""
+    no_errors: ""
+
+scanner:
+    diff_only: False  # If True, errors caused by only the patch are shown
+
+pycodestyle:
+    max-line-length: 79  # Default is 79 in PEP8
+    ignore:  # Errors and warnings to ignore
+        - E402
+
+only_mention_files_with_errors: True  # If False, a separate status comment for each file is made.
+descending_issues_order: False # If True, PEP8 issues in message will be displayed in descending order of line numbers in the file
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -6,7 +6,7 @@ message:  # Customize the comment made by the bot
     updated:  # Messages when new commits are added to the PR
         header: ""
         footer: ""
-    no_errors: ""
+    no_errors: "No PEP8 issues."
 
 scanner:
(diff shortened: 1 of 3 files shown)


kohr-h:
Hm, nothing.

.pep8speaks.yml
```diff
@@ -0,0 +1,22 @@
+# File : .pep8speaks.yml
+
+message:  # Customize the comment made by the bot
+    opened:  # Messages when a new PR is submitted
```
adler-j: Strongly against spamming all PRs like this. Better just leave a comment when something goes wrong.

.pep8speaks.yml
```diff
@@ -0,0 +1,22 @@
...
+        header: "Hi @{name}, thanks for updating the PR!"
+        footer: ""  # Why to comment the link to the style guide everytime? :)
+    no_errors: "There are no PEP8 issues."
+
+scanner:
+    diff_only: False  # If True, errors caused by only the patch are shown
+
+pycodestyle:
+    max-line-length: 79  # Default is 79 in PEP8
+    ignore:  # Errors and warnings to ignore
```
adler-j: I guess there are some in our settings?

kohr-h: We have one in our settings, but it doesn't seem to apply anymore. It's E402:
```
relative-beyond-top-level (E0402): *Attempted relative import beyond top-level package*
```
It's not triggered anymore as it seems, so I'd just leave this section empty for now.

.pep8speaks.yml
kohr-h: Yeah, this is sort of the default, I'll replace it with an empty string.

kohr-h:
Okay now it seems to work. The used config seems to be the last before pushing. I'll test again with the updated config.

kohr-h:
Travis is also happy, no longer reporting PEP8 violations.
I'll check what happens if I update with a new PEP8 violation. The doc of pep8speaks says that there should be only one comment at a time, not a new one every time.

kohr-h:
Testing now what happens when issues are fixed.

kohr-h:
Ah, E402 is triggered in `__init__.py` files. I'll add it again.

adler-j:
Looking good now I'd say!

kohr-h:
This last push updated the comment above. So perhaps it's a good idea anyway to write *something* regardless of issues or not to make sure that the bot message is always close to the top.

I'll try pushing an update that fixes pep8, and see what happens to the comment.
The I'm going to rebase after that and see if it also works.

kohr-h:
Okay, now it's updated obviously it uses the default config.

kohr-h:
Okay, also the rebased push triggered an update. I think we can merge this now, perhaps fix small annoyances later.

kohr-h:
Approve?

Issue #1174, opened 2017-10-05
Improved callbacks


MatthiasJE:
I would like to propose two small improvements for callbacks. First, an optional "end" argument should be given which allows elegant concatenation of print callbacks such as

Iter = 9, time = 10 s, total time = 20 s

based on three different callbacks in one line rather than 3.

Second, as already hinted in the previous example, a new callback that measures the total time. To me, this is often more interesting than the time per iteration.

If you are happy with the two proposed changes, then I would prepare a pull request for this.

MatthiasJE:
It is probably better to allow any kind of keyword arguments to be passed to the print function.

adler-j:
Overall this sounds like a good addition, and for callbacks we can't really have to many I guess. If you could make a PR we could help distill it!

adler-j:
I guess this is merged now, closing!

Thank you for the input!

PR #1176, opened 2017-10-05
Issue1174 improved callbacks

MatthiasJE:
This is the pull request for #1174. It does exactly what I proposed in that issue.

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -423,6 +423,8 @@ class CallbackPrintTiming(Callback):
             where ``runtime`` is the runtime since the last iterate.
         step : positive int, optional
             Number of iterations between prints.
+        cumulative : boolean, optional
+            Print the time since the initialization instead of the last call.
 
         Other Parameters
         ----------------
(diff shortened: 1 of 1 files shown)


odl/solvers/util/callback.py
```diff
@@ -445,11 +457,64 @@ def __repr__(self):
         return '{}({})'.format(self.__class__.__name__, inner_str)
 
 
+class CallbackPrintTotalTime(Callback):
```
kohr-h: Why do we need a new class for this? 
IMO we should instead just add a `cumulative=False` option to the existing `CallbackPrintTiming` and massage the code a bit to cover both cases.

adler-j: Actually this is a great suggestion. Perhaps that should be done instead.

MatthiasJE: Agree. I will make those changes.

odl/solvers/util/callback.py
```diff
@@ -416,25 +423,37 @@ def __init__(self, fmt='Time elapsed = {:<5.03f} s', step=1):
...
+        self.time = 0
+        self.kwargs = kwargs
 
     def __call__(self, _):
         """Print time elapsed from the previous iteration."""
         if self.iter % self.step == 0:
             t = time.time()
-            print(self.fmt.format(t - self.time))
-            self.time = t
+            self.time = t - self.base_time
```
adler-j: Is there any reason why we need to save this as `self...`?

MatthiasJE: I was thinking that it was good to have the cumulative time as an output attribute but actually it is cleaner for this not to be. This is now fixed.

Issue #1062, opened 2017-07-06
Product space broadcasting logic differs from Numpy's


kohr-h:
When discussing broadcasting in the context of `MultiplyOperator` in #1060 I realized that our current [broadcasting logic](https://github.com/odlgroup/odl/blob/9b401dbb2b1bf41b5de783f0ef1ed433b6a82487/odl/space/pspace.py#L963) is different from Numpy's logic.

What we do is broadcast arithmetic ops between elements of `space` and `ProductSpace(space, 3)`, say.

Translated to Numpy, this would mean broadcasting of `np.ones(5)` against `np.ones((3, 5))`. ~~But Numpy doesn't do that. To trigger broadcasting, the first array needs to get an empty first axis (of length 1).~~

~~Translated back to ODL, that means in the above example that we should only broadcast between elements of `ProductSpace(space, 1)` and `ProductSpace(space, 3)`. To keep things simple it should remain the duty of the user to wrap the "raw" `space` element into a product space of size 1.~~

kohr-h:
Seems I got the facts wrong, closing.

PR #1177, opened 2017-10-05
WIP: fix adjoint weightings

kohr-h:
This is a discussion PR. I think we could go somewhere along these lines to make adjoints "automatically correct" with respect to weighted spaces, in particular when the weights differ.

This solution should always be correct where it applies, however it wouldn't yet help in special cases like `RayTransform` where the weights need to be computed per axis.
The larger question is what to do instead of the monkey-patching here. I can imagine a bunch of scenarios:

1. Do this automatically in `Operator`. This means that users implement the unweighted adjoint, and the weighted one is implemented with a helper function similar to this.
   This would require us to communicate this "magic" clearly and prominently.
2. As in 1., but depending on some flag or whatever that is provided by the user. It could be a class attribute like `_auto_weighting` or so. Don't do anything by default.
   This option would still need good communication but leads to less unexpected behavior on the user side.
3. Just offer a tool (e.g. a decorator) for `adjoint`, like `@auto_weighting` that users can choose to use.
   This option has the least surprise and the simplest migration path, we would just have to use it a lot ourselves so users pick it up.

Opinions?

For the records: What I **don't** want is to let `adjoint` return an operator composition. That would be ugly and a usability nightmare, since operator properties will be buried deep down.


adler-j:
In my opinion, adding a utility is the only option that is not outrageously surprising to users. This could be, as you suggest, in the form of a decorator, but I'd simply make it a free function that is either called on the adjoint:

```python
@property
def adjoint(self):
    ...
    return auto_weighting(adjoint)
```

or in the operator itself 

```python
class AdjointyOperator:
    def __call__(self, x):
        return auto_weighting(adjoint)
```

kohr-h:
> In my opinion, adding a utility is the only option that is not outrageously surprising to users.

I think the same. Only for the sake of not introducing bias in the discussion I left it kind of open.

Note that the option of decorating the adjoint, which is equivalent to your first code box, would still require some monkey-patching. However, in this case I would find it more acceptable since it happens "almost" during object creation.

The second option is cleaner in that respect, but has the disadvantage that it would make it harder to use existing (unweighted) operators off the shelf. Things like `return ThisOtherOperator()` in `adjoint` will not be "auto-weightable" since that would require a `_call` method to decorate.

So I'm leaning towards the operator decoration option.

And BTW, this has interesting implications for `OperatorComp` where you could start pushing around weighting constants all the way to the smallest space, for instance.

I'll think a bit more about this.

kohr-h:
> , which is equivalent to your first code box,

*roughly* :-)

kohr-h:
Here's a more serious suggestion. Obviously needs more tests, but does the job.

kohr-h:
Also obviously, the implementation does not make use of the `optimize` flag yet and just does it unconditionally. Further, it does it badly by computing those `1 / something` arrays. That needs to change.

kohr-h:
So this is now a serious PR. I've added tests and an example, and I think this concept would actually work. Comments welcome.

kohr-h:
Once we settle on a solution, I think this PR should also go through the "standard library" and fix all instances of operators that need fixing. I'll make a TODO list at the top.

kohr-h:
Open question: what to do with spaces that have no `weighting` attribute? (See TODOs)

kohr-h:
Ping @MatthiasJE this may be interesting to you, too. You also had trouble with unscaled adjoints, right?

MatthiasJE:
Yes, in the most elegant definition of TGV I would like to use weighted spaces for the second derivative. It then needs proper adjoints and prox operators. This should be used in 
`examples/solvers/pdhg_denoising_tgv.py`. What would be an easy way for me to test this? Can I checkout your branch?

kohr-h:
Sure, you can do
```
git remote add kohr-h https://github.com/odlgroup/odl.git
git fetch kohr-h
git checkout issue-1068__ajoint_weighting  # sorry about the typo :-)
```

mehrhardt:
@kohr-h, I tried the TGV example again with weighted spaces. It does not work as weighting is not implemented for product spaces?
`NotImplementedError: weighted spaces not supported`

kohr-h:
What happens if you remove that check and add the `@auto_weighting` decorator to `ProductSpaceOperator.adjoint`?

mehrhardt:
I tried the following but it did not help significantly. It does then result in other errors about things not being well-defined (see below).

```
+++ b/odl/operator/pspace_ops.py
@@ -18,6 +18,7 @@ from odl.operator.operator import Operator
 from odl.operator.default_ops import ZeroOperator
 
+from odl.space.space_utils import auto_weighting
 
 __all__ = ('ProductSpaceOperator',
            'ComponentProjection', 'ComponentProjectionAdjoint',
@@ -148,15 +149,11 @@ class ProductSpaceOperator(Operator):
             if not isinstance(domain, ProductSpace):
                 raise TypeError('`domain` {!r} not a ProductSpace instance'
                                 ''.format(domain))
-            if domain.is_weighted:
-                raise NotImplementedError('weighted spaces not supported')
 
         if range is not None:
             if not isinstance(range, ProductSpace):
                 raise TypeError('`range` {!r} not a ProductSpace instance'
                                 ''.format(range))
-            if range.is_weighted:
-                raise NotImplementedError('weighted spaces not supported')
 
         # Convert ops to sparse representation
         self.ops = scipy.sparse.coo_matrix(operators)
@@ -312,6 +309,7 @@ class ProductSpaceOperator(Operator):
         return ProductSpaceOperator(deriv_matrix, self.domain, self.range)
 
     @property
+    @auto_weighting
     def adjoint(self):
         """Adjoint of this operator.
```

```
normE = E.norm(estimate=True)
normE2 = E2.norm(estimate=True)


  File "<ipython-input-2-b66a9577355f>", line 1, in <module>
    normE = E.norm(estimate=True)

  File "/home/ehrhardt/repositories/git_myODL/odl/operator/operator.py", line 751, in norm
    return power_method_opnorm(self, **kwargs)

  File "/home/ehrhardt/repositories/git_myODL/odl/operator/oputils.py", line 229, in power_method_opnorm
    op.adjoint(tmp, out=x)

  File "/home/ehrhardt/repositories/git_myODL/odl/operator/operator.py", line 686, in __call__
    result = self._call_in_place(x, out=out, **kwargs)

  File "/home/ehrhardt/repositories/git_myODL/odl/space/space_utils.py", line 421, in _call
    x = new_ran_w * x

TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'ProductSpaceElement'
```

mehrhardt:
Maybe it would be the best if you have a look yourself. I tried to run `pdhg_denoising_tgv.py` with the following modification after the definition of derivates:

```
# Create symmetrized operator and weighted space.
# TODO: As the weighted space is currently not supported in ODL we find a
# workaround.
W = odl.ProductSpace(U, 3, weighting=[1, 1, 2])
E = odl.operator.ProductSpaceOperator(
    [[Dx, 0], [0, Dy], [0.5*Dy, 0.5*Dx]], range=W)

E2 = odl.operator.ProductSpaceOperator(
    [[Dx, 0], [0, Dy], [0.5 * Dy, 0.5 * Dx], [0.5 * Dy, 0.5 * Dx]])
W2 = E.range

normE = E.norm(estimate=True)
normE2 = E2.norm(estimate=True)
```

kohr-h:
Without actually looking, this has to do with #1062 in some sense. What happens here is that the code tries to multiply the result, a product space element, with a Numpy array that has the weights, one for each product space component. That doesn't work, unfortunately, which is part of a bigger issue.

I can't find the discussion we had before, but let me make a suggestion how to solve this and #1062, and probably some other issues, too. I would suggest this:

```python
>>> pspace = odl.rn(2) * odl.rn(3)
>>> z = pspace.one()
>>> z * [1, 2]  # should work, doesn't currently
ProductSpace(rn(2), rn(3)).element([
    [1.0, 1.0],
    [2.0, 2.0, 2.0]
])
>>> z * np.array([1, 2])  # same as above
ProductSpace(rn(2), rn(3)).element([
    [1.0, 1.0],
    [2.0, 2.0, 2.0]
])
>>> pspace = odl.rn(2) ** 2
>>> z = pspace.one()
>>> z * np.array([1, 2])[None, :]  # broadcast along pspace axis, now produces Numpy array
ProductSpace(rn(2), rn(3)).element([
    [1.0, 2.0],
    [1.0, 2.0]
])
```
The principle would be as follows:
- For non-power product spaces, only 1D arrays of the same length as the space make sense. So the only sensible thing is to do the equivalent of `z[i] * arr[i] for i in range(length)`.
- Since power spaces are just a special case, the same thing must work for them.
- To broadcast in a different manner, one has to explicitly add empty axes into the array, and the same broadcasting rules as in Numpy apply.

That would make it unambiguous as to how to interpret
```python
>>> pspace = odl.rn(2) ** 2
>>> z = pspace.one()
>>> z * [1, 2]  # along axis 0 or axis 1??
```
It would always be axis 0, and for axis 1 one would have to do `z * [[1, 2]]` or with Numpy, `z * np.array([1, 2])[None, :]`. And it would solve the current issue that the latter variant produces a Numpy array, which I think is a bug.

kohr-h:
In view of the current issue with product spaces, I think I'll implement a workaround for that case to move things forward.

kohr-h:
> In view of the current issue with product spaces, I think I'll implement a workaround for that case to move things forward.

Will do that now.

kohr-h:
Now the decorator works also for operators on product spaces. Needs a proper review now. @mehrhardt if you like, you can test your code again.

Anyway, I think this is a rather important issue since many adjoints are wrong as of now, we should give it some priority.

mehrhardt:
I have almost good news. After removing a few lines in `pspace_ops.py`

```
        # Validate input data
        if domain is not None:
            if not isinstance(domain, ProductSpace):
                raise TypeError('`domain` {!r} not a ProductSpace instance'
                                ''.format(domain))
#            if domain.is_weighted:
#                raise NotImplementedError('weighted spaces not supported')

        if range is not None:
            if not isinstance(range, ProductSpace):
                raise TypeError('`range` {!r} not a ProductSpace instance'
                                ''.format(range))
#            if range.is_weighted:
#                raise NotImplementedError('weighted spaces not supported')
```

the following example

```
# two models of the 'same' operator

import odl

U = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300],
                      dtype='float32')
G = odl.Gradient(U, method='forward', pad_mode='symmetric')
V = G.range

Dx = odl.PartialDerivative(U, 0, method='backward', pad_mode='symmetric')
Dy = odl.PartialDerivative(U, 1, method='backward', pad_mode='symmetric')

E = odl.operator.ProductSpaceOperator(
    [[Dx, 0], [0, Dy], [0.5 * Dy, 0.5 * Dx], [0.5 * Dy, 0.5 * Dx]])
W = E.range

W2 = odl.ProductSpace(U, 3, weighting=[1, 1, 2])
E2 = odl.operator.ProductSpaceOperator(
        [[Dx, 0], [0, Dy], [0.5*Dy, 0.5*Dx]], range=W2)

l1 = odl.solvers.L1Norm(W)
l12 = odl.solvers.L1Norm(W2)


w = W.one()
w2 = W2.one()

v = odl.phantom.white_noise(V)

print(w2.inner(E2(v)))
print(w.inner(E(v)))

print(l1(w))
print(l12(w2))

print(E.norm(estimate=True, xstart=v))
print(E2.norm(estimate=True, xstart=v))
```

gives the output:

> 10.128983855247498
> 10.128984451293945
> 6400.0
> 6400.0
> 21.1059752145
> 18.2776606142

Only the norm estimate is wrong. Maybe you have an idea what is wrong here? Also it might be the easiest if you commit the small changes in `pspace_ops.py` yourself rather me making a PR.

kohr-h:
Nice! The norm estimation will have to use the same fix, I'll look into it.

kohr-h:
I was also thinking of a way to enable or disable the auto-weighting. I tend to think it should go as an extra `auto_weighting=True` parameter of `Operator`, stored as `_auto_weighting` flag, and the decorator would check that flag and act accordingly.

examples/space/auto_weighting.py
```diff
@@ -0,0 +1,111 @@
+"""Example demonstrating the usage of the ``auto_weighting`` decorator."""
```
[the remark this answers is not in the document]
adler-j: Also, this reads more like a test than an example. Could you not reduce it significantly and then move most of this into a test?

odl/space/space_utils.py
```diff
@@ -203,6 +206,233 @@ def rn(size, dtype=None, impl='numpy', **kwargs):
     return rn
 
 
+class auto_weighting(OptionalArgDecorator):
+
+    """Make an unweighted adjoint automatically account for weightings.
```
adler-j: We need to clearly define what "unweighted" means here. For discrete lp spaces, is it the "unit weighting", or is it the "weight 1 for each pixel" weighting so to say?

odl/space/space_utils.py
```diff
@@ -203,6 +206,233 @@ def rn(size, dtype=None, impl='numpy', **kwargs):
...
+    To avoid the inconvenience of dealing with `OperatorComp` objects,
+    the given operator is monkey-patched instead of composed.
+
+    Parameters
+    ----------
+    unweighted_adjoint : `Operator`
+        Unweighted variant of the adjoint. It will be patched with a
+        new ``_call()`` method.
+        The weightings of ``domain`` and ``range`` of the operator
+        must be `ArrayWeighting` or `ConstWeighting`.
```
adler-j: and we still need to somehow deal with `NoWeighting`

odl/space/space_utils.py
```diff
@@ -203,6 +206,233 @@ def rn(size, dtype=None, impl='numpy', **kwargs):
...
+        \\langle Af, g \\rangle_{Y_v} =
+        \\langle Bf, v \cdot g \\rangle_Y =
+        \\langle f, B^*(v \cdot g) \\rangle_X =
+        \\langle f, w^{-1}\, B^*(v \cdot g) \\rangle_{X_w}.
+
+    Thus, from the existing unweighted adjoint :math:`B^*` one can compute
+    the weighted one as :math:`A^* = w^{-1}\, B^*(v\, \cdot)`.
+    Depending on the types of weighting, this expression can be simplified
+    further, e.g., a constant weight can be absorbed into the other weight.
+    """
```
adler-j: This could use a minimal example

odl/space/space_utils.py
```diff
@@ -203,6 +206,233 @@ def rn(size, dtype=None, impl='numpy', **kwargs):
...
+            raise TypeError(
+                'weighting of `unweighted_adjoint.domain` must be of '
+                'type `ArrayWeighting` or `ConstWeighting`, got {}'
+                ''.format(type(ran_weighting)))
+
+        # Compute the effective weights and mark constants 1.0 as to be
+        # skipped
+        if not optimize:
+            new_dom_w, new_ran_w = dom_w, ran_w
+            skip_dom = dom_w_type == 'const' and dom_w == 1.0
```
adler-j: Add parentheses to make this easier to read

odl/space/space_utils.py
```diff
@@ -203,6 +206,233 @@ def rn(size, dtype=None, impl='numpy', **kwargs):
...
+
+        # Define the new `_call` depending on original signature
+        self = unweighted_adjoint
+
+        def mul_weight(x, w):
+            """Multiplication with weights that works in product spaces."""
+            if not isinstance(x, ProductSpaceElement) or np.isscalar(w):
+                return w * x
+            else:
+                # Product space, array weight
```
adler-j: Product of product etc?

odl/space/space_utils.py
```diff
@@ -203,6 +206,233 @@ def rn(size, dtype=None, impl='numpy', **kwargs):
...
+            new_dom_w, new_ran_w = dom_w, ran_w
+            skip_dom = skip_ran = False
+        elif dom_w_type == 'array' and ran_w_type == 'const':
+            new_dom_w = dom_w / ran_w
+            new_ran_w = 1.0
+            skip_dom = False
+            skip_ran = True
+        elif dom_w_type == 'const' and ran_w_type == 'array':
+            new_dom_w = 1.0
+            new_ran_w = ran_w / dom_w
```
adler-j: Is this permanently stored (e.g. could we be looking at 1gb of memory for this?)

kohr-h:
> This looks quite functional (with some minor hickups), but the main issue I have is this:
> 
> ```python
   >>> rn = odl.rn(2)
   >>> rn_w = odl.rn(2, weighting=2)
   >>> op = ScalingOp(rn, rn_w, 1.5)
   >>> x = space.one()
   >>> op(x) == op.adjoint(x)
> False  # as expected
   >>> op == op.adjoint
> True  # Or why would this not happen?
   >>> repr(op) == repr(op.adjoint)
> True  # wat
>  ```

> In short, the monkey patching is extremely non-transparent and makes things that are strictly different seem like they are the same.
> 
> Is there some way to handle this?

Nothing like that happens. In both cases the result is `False`, and I have no idea why it should be `True`.
In particular, the code makes sure that the `adjoint` method *never* returns the same operator. And since `Operator` does not implement `__eq__`, only the identical operator would return `True` when comparing `op == op.adjoint`. So that comparison evaluates to `False` by construction.
And for the `repr` part you get
```python
>>> repr(op)
'ScalingOp: rn(2) -> rn(2, weighting=2.0)'
>>> repr(op.adjoint)
'ScalingOp: rn(2, weighting=2.0) -> rn(2)'
```
which is fine I guess?

examples/space/auto_weighting.py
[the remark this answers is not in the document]
kohr-h: I'll reduce the example to a single case and less dense output. The unit test already does all of the stuff in here.

odl/space/space_utils.py
kohr-h: I don't see that. Unweighted is whatever comes in as adjoint operator, and all the other weightings are defined relative to that.

odl/space/space_utils.py
kohr-h: If you refer to the merging rules, they are all explained in detail. As for the general idea, I agree.

odl/space/space_utils.py
kohr-h: Yes, it will be part of the closure of the new `_call` method and thus have the same lifetime as the operator.
I added a note in the documentation about this effect.

Obviously, we also need to mention this in the online documentation, especially in the "How to write an operator" doc.

odl/space/space_utils.py
kohr-h: Fixed

odl/space/space_utils.py
kohr-h: Done that, please check the updated docs.

kohr-h:
OK I've addressed all the comments I think. The only TODO (if the design is ok) is to properly document it in the online doc.

kohr-h:
Bump for review.

See also the weight handling in #1238 with per-axis weighting. The weight merging there is done in the [`adjoint_weightings`](https://github.com/kohr-h/odl/blob/bb31079f7c7d6fcd897ba9e206242dba41dbe845/odl/space/weighting.py#L790-L988) function.

examples/operator/auto_weighting.py
```diff
@@ -0,0 +1,58 @@
+"""Example demonstrating the usage of the ``auto_weighting`` decorator."""
```
adler-j: This needs to be somewhat fleshed out, especially given that new users come here

odl/space/space_utils.py
```diff
@@ -281,6 +284,303 @@ def rn(shape, dtype=None, impl='numpy', **kwargs):
     return rn
 
 
+class auto_weighting(OptionalArgDecorator):
```
adler-j: Suggest rename to `auto_adjoint_weighting` to make this more clear

odl/space/space_utils.py
```diff
@@ -281,6 +284,303 @@ def rn(shape, dtype=None, impl='numpy', **kwargs):
...
+    In this context, the term "unweighted adjoint" of an operator ``A``
+    refers to the the adjoint of ``A`` defined between **unweighted
+    Euclidean spaces**. In particular, discretized ``L^p`` type spaces are
+    interpreted as weighted variants of ``R^n`` or ``C^n`` spaces.
+
+    For instance, an (m x n) real matrix ``M`` is also an operator
+    ``M: R^n -> R^m``, and its unweighted adjoint is defined by the
+    transposed matrix ``M^T: R^m -> R^n``.
+
+    Similarly, when an integration operator ``A: L^2(I) -> R`` on an
```
adler-j: Wouldnt always using the R^n -> R style be better?

odl/space/space_utils.py
```diff
@@ -281,6 +284,303 @@ def rn(shape, dtype=None, impl='numpy', **kwargs):
...
+    as a summation operator ``S: R^n -> R, S(x) = dx * sum(x)`` scaled
+    by the subinterval size ``dx``.
+    Here, the unweighted adjoint is the adjoint ``S^*: R -> R^n`` of ``S``,
+    given by ``S^*(c) = dx * (c, ..., c)``.
+    The weighting to correctly define the *weighted* adjoint
+    ``A^*: R -> L^2(I)`` is taken care of by this decorator.
+
+    Parameters
+    ----------
+    unweighted_adjoint : `Operator`
```
adler-j: Shouldn't the type of this be "method" or something like that?

odl/space/space_utils.py
```diff
@@ -281,6 +284,303 @@ def rn(shape, dtype=None, impl='numpy', **kwargs):
...
+            skip_ran = True
+        elif dom_w_type == 'const' and ran_w_type == 'array':
+            new_dom_w = 1.0
+            new_ran_w = ran_w / dom_w
+            skip_dom = True
+            skip_ran = False
+        elif dom_w_type == 'const' and ran_w_type == 'const':
+            if unweighted_adjoint.domain.size < unweighted_adjoint.range.size:
+                new_dom_w = 1.0
+                new_ran_w = ran_w / dom_w
```
adler-j: Are we sure this always makes sense? What about operators `C -> R` etc?

odl/test/space/space_utils_test.py
```diff
@@ -79,5 +86,196 @@ def test_vector_numpy():
...
+        def _call(self, x):
+            return self.c * x
+
+        @property
+        @auto_weighting
+        def adjoint(self):
+            return ScalingOp(self.range, self.domain, self.c)
+
+    op1 = ScalingOp(rn, rn, 1.5)
+    op2 = ScalingOp(rn_w, rn_w, 1.5)
```
adler-j: Prefer not using 1.5 everywhere

odl/test/space/space_utils_test.py
```diff
@@ -79,5 +86,196 @@ def test_vector_numpy():
...
+        dom_el = noise_element(op.domain)
+        ran_el = noise_element(op.range)
+        assert pytest.approx(op(dom_el).inner(ran_el),
+                             dom_el.inner(op.adjoint(ran_el)))
+
+
+def test_auto_weighting_cached_adjoint():
+    """Check if auto_weighting plays well with adjoint caching."""
+    rn = odl.rn(2)
+    rn_w = odl.rn(2, weighting=2)
```
adler-j: Use a "random" value here

odl/test/space/space_utils_test.py
```diff
@@ -79,5 +86,196 @@ def test_vector_numpy():
...
+def test_auto_weighting_raise_on_return_self():
+    """Check that auto_weighting raises when adjoint returns self."""
+    rn = odl.rn(2)
+
+    class InvalidScalingOp(odl.Operator):
+
+        def __init__(self, dom, ran, c):
+            super(InvalidScalingOp, self).__init__(dom, ran, linear=True)
+            self.c = c
+            self._adjoint = None
```
adler-j: What does this line do?

kohr-h:
> Could this somehow become buggy if adjoint is an attribute instead of a method?

That cannot happen since `adjoint` is a property on `Operator` and cannot be overwritten by an attribute:
```python
>>> class A(odl.Operator):
...     def __init__(self):
...         super(A, self).__init__(odl.rn(2), odl.rn(2))
...         self.adjoint = None
>>>
>>> a = A()
    [...]

    self.adjoint = None

AttributeError: can't set attribute
```

kohr-h:
> Does this commute with @property?

No. I tried to hack it in, but the wrapping does not seem to work as expected. I'll leave it as it is, but I can improve the error message to give users a hint.

kohr-h:
> How does this handle adjoints that are defined in terms of other operators?

Not sure what you mean. Like compositions and such?

kohr-h:
> Shouldn't this be a operator util instead of a space util?

Yes I guess so.

examples/operator/auto_weighting.py
kohr-h: Can do

odl/space/space_utils.py
kohr-h: What do you mean by "using"? In this example?

odl/space/space_utils.py
kohr-h: Technically yes. I'll change it.

odl/space/space_utils.py
kohr-h: Yes, because weightings can never be complex. (Note to self: make sure that this is checked.)

odl/test/space/space_utils_test.py
kohr-h: That must be copy-paste from the caching test, should be removed.

kohr-h:
Current status here is that there's some mysterious issue when `adjoint` defers to `inverse`.

I'll think a bit about better ways to do this, without doing brain surgery on the operator returned by `adjoint`. Maybe a metaclass after all.

PR #1183, opened 2017-10-10
TST: Add doctest to RayTransform

adler-j:


adler-j:
Then why did the test pass?

kohr-h:
Because we have no Travis test scenario for minimal dependencies :-)

adler-j:
Well that is a bit worrying. With that said, don't we ignore certain files in the tests?

kohr-h:
Let me just in a non-smart-ass way recall [this conversation](https://github.com/odlgroup/odl/pull/1164#discussion_r141533959), maybe it wasn't such a bad idea after all.

We don't ignore anything in `tomo`, that would also help, of course.

adler-j:
*whistles* anyway I guess you have quite a point there :/ With that said there are also weird combinations of dependencies that we technically should test as well (ASTRA but not PYFFTW?).

Anyway, would a solution not be to simply ignore tomo depending on the existence of *any* backend?

kohr-h:
> whistles anyway I guess you have quite a point there :/ With that said there are also weird combinations of dependencies that we technically should test as well (ASTRA but not PYFFTW?).

Well, of course, we cannot test *all* scenarios, and maybe not all are independent. But I think one test for the minimal installation scenario would already catch the majority of bugs, which are unfenced imports of optional packages and tests that should be excluded but aren't.



> Anyway, would a solution not be to simply ignore tomo depending on the existence of any backend?

That would make sense. Of course, we can use a finer granularity if we want to (e.g. geometries don't have extra deps), but that seems like too much extra maintenance required.

adler-j:
Since this cannot go in as is, I'll close this and perhaps re-open it at some later point, but that would be some effort in doing it properly.

PR #1184, opened 2017-10-10
TST: Add general tests for supervised FOMs

adler-j:


odl/contrib/fom/supervised.py
```diff
@@ -394,67 +394,54 @@ def blurring(data, ground_truth, mask=None, normalized=False,
...
 
-def false_structures(data, ground_truth, mask=None, normalized=False,
-                     smoothness_factor=None):
-    """Return weighted L2 distance, de-emphasizing regions defined by ``mask``.
-
-    .. note::
-        If the mask argument is omitted, this FOM is equivalent to the
-        mean squared error.
+def false_structures_mask(foreground, smoothness_factor=None):
+    """Return mask empasizing areas outside ``foreground``.
```
kohr-h: emphasizing ("h" missing)
Same in other places

odl/contrib/fom/supervised.py
```diff
@@ -394,67 +394,54 @@ def blurring(data, ground_truth, mask=None, normalized=False,
...
+    Raises
+    ------
+    ValueError
+        If foreground is all zero or contains values not in {0, 1}.
 
-    .. math::
-        \\alpha(x) = e^{-\\frac{1}{k} \\beta_m(x)},
+    Notes
+    -----
+    This helper function computes the euclidean distance transform from each
```
kohr-h: Euclidean

odl/contrib/fom/supervised.py
```diff
@@ -394,67 +394,54 @@ def blurring(data, ground_truth, mask=None, normalized=False,
...
-    data : `FnBaseVector`
-        Input data to compare to the ground truth.
-    ground_truth : `FnBaseVector`
-        Reference to compare ``data`` to.
-    mask : `array-like`, optional
-        Binary mask to define ROI in which FOM evaluation is performed.
-    normalized  : bool, optional
-        Boolean flag to switch between unormalized and normalized FOM.
+    foreground : `FnBaseVector`
+        The region that should be de-emphasised
```
kohr-h: dot

odl/contrib/fom/supervised.py
```diff
@@ -394,67 +394,54 @@ def blurring(data, ground_truth, mask=None, normalized=False,
...
 
-    if smoothness_factor is None:
-        smoothness_factor = np.mean(data.space.shape) / 10
-
-    if mask is not None:
-        mask = distance_transform_edt(1 - mask)
-        mask = np.exp(mask / smoothness_factor)
+    unique = np.unique(foreground)
+    if not np.array_equiv(unique, [0., 1.]):
+        raise ValueError('`foreground` contains values not in [0, 1] or has '
```
kohr-h: I read `[0, 1]` as an interval. Also this restriction is nowhere mentioned. I'd use the term "binary mask".

odl/contrib/fom/supervised.py
```diff
@@ -538,9 +525,12 @@ def psnr(data, ground_truth, normalized=False):
...
     ground_truth : `FnBaseVector`
         Reference to compare ``data`` to.
-    normalized : bool
-        If true, normalize ``data`` and ``ground_truth`` to have the
-        same mean and variance before comparison.
+    use_zscore : bool
+        If true, normalize ``data`` and ``ground_truth`` to have zero mean and
+        unit variance before comparison.
+    force_lower_is_better : bool
+        If true, then lower value indicates better fit. In this case the output
```
kohr-h: \`\`True\`\`
Harmonize with formulations elsewhere (other branches/PRs)

odl/contrib/fom/supervised.py
```diff
@@ -538,9 +525,12 @@ def psnr(data, ground_truth, normalized=False):
         Input data to compare to the ground truth.
     ground_truth : `FnBaseVector`
         Reference to compare ``data`` to.
-    normalized : bool
-        If true, normalize ``data`` and ``ground_truth`` to have the
-        same mean and variance before comparison.
+    use_zscore : bool
```
kohr-h: Why this change? It breaks uniformity of the API.

odl/contrib/fom/test/test_supervised.py
```diff
@@ -56,17 +87,43 @@ def test_mean_absolute_error(space):
...
+    # Test psnr with both constants is infinity
+    assert odl.contrib.fom.psnr(zero, zero) == np.inf
 
+    # Test psnr with ground truth constant is negative infinity
+    assert odl.contrib.fom.psnr(data, zero) == -np.inf
+
+    # Compute the true value
     mse = np.mean((true - data) ** 2)
     maxi = np.max(np.abs(true))
     expected = 10 * np.log10(maxi**2 / mse)
```
kohr-h: spaces around `**`

odl/contrib/fom/supervised.py
adler-j: Basically the normalization here is of the *input* while the others normalize their *output*. Having the same name was quite bad programming (in my opinion)

kohr-h: Good point. So the `normalized` option is simply missing.

adler-j: PSNR cannot in any "nice" way be normalized. E.g. it can go from -infty to infty, and irritatingly, infty is the value we get from `psnr(x, x)` so normalizing it is "weird". Sure, we could apply something like e.g. a sigmoid, but well, what would the end really mean then? But perhaps its better than nothing.

kohr-h: Hm, so how about this: If `M` is the squared max of the ground truth, and `E` the mean squared error, then
```
PSNR = log_10( (M / E)^(10) ) = ln( (M / E)^(10) ) / ln(10)
```
So putting that into a sigmoid `s(x) = 1 / (1 + e^(-a x))` for `a = ln(10)` means
```
s(PSNR) = 1 / (1 + e^(-ln( (M/E)^(10) )) = 1 / (1 + (M/E)^(-10))
```
which would be very close to 1 for anything relevant. But by choosing a smaller `a` one could find something that yields 0.5 for "moderate" PSNR and close to 1 for "high" PSNR, where those are experience values. Maybe moderate would be 20 or so, and high 40?

adler-j: Well we could let users simply select the parameter. I'd probably write the normalized formula

```
nPSNR(x) = 1 / (1 + e^(-x/a))
```

and then users can easily state a "standard error", e.g. 20 decibels.

kohr-h: Sure, why not. It needs a bit of experimenting I think, but some normalization for a user-specified db value would make sense.

kohr-h:
Stupid fails because of `scipy` wheels. It's probably better to add `scipy` to the list of packages to be installed with `conda`.

adler-j:
Totally agree that conda is the way to go for scipy, other than that this should be mergable?

kohr-h:
Yes

PR #1185, opened 2017-10-10
ENH: Pull request of group 1 (Workshop day, work on FOMs)

sbanert:
Add flags to mean_squared_error, mean_absolute_error, SSIM, improve the documentation and add tests (the latter applies only to SSIM).

### DIFF
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -34,12 +34,13 @@ def mean_squared_error(data, ground_truth, mask=None,
         If given, ``data * mask`` is compared to ``ground_truth * mask``,
         where multiplication is pointwise.
     normalized  : bool, optional
-        Boolean flag to switch between unormalized and normalized FOM.
-        See `Notes` for details.
+        If ``True``, the output values are mapped to the interval
+        :math:`[0, 1]` (see `Notes` for details), otherwise return the
+        original mean squared error.
     force_lower_is_better : bool, optional
-        Boolean flag which enforces that lower values correspond to better
-        matches.
-
+        If ``True``, it is ensured that lower values correspond to better
+        matches. For the mean squared error, this is already the case, and
+        the flag is only present for compatibility to other figures of merit.
     Returns
     -------
     mse : float
@@ -79,8 +80,8 @@ def mean_squared_error(data, ground_truth, mask=None,
         fom /= l2norm_squared(data.space.one())
 
     # The mean squared error is already in the desired order.
-    if force_lower_is_better:
-        pass
+    # The output does not need to be changed depending on the
+    # force_lower_is_better flag.
 
     return fom
 
@@ -102,11 +103,13 @@ def mean_absolute_error(data, ground_truth, mask=None,
         If given, ``data * mask`` is compared to ``ground_truth * mask``,
         where multiplication is pointwise.
     normalized  : bool, optional
-        Boolean flag to switch between unormalized and normalized FOM.
-        See `Notes` for details.
+        If ``True``, the output values are mapped to the interval
+        :math:`[0, 1]` (see `Notes` for details), otherwise return the
+        original mean absolute error.
     force_lower_is_better : bool, optional
-        Boolean flag which enforces that lower values correspond to better
-        matches.
+        If ``True``, it is ensured that lower values correspond to better
+        matches. For the mean absolute error, this is already the case, and
+        the flag is only present for compatibility to other figures of merit.
 
     Returns
     -------
@@ -144,8 +147,8 @@ def mean_absolute_error(data, ground_truth, mask=None,
         fom /= l1_norm(data.space.one())
 
     # The mean absolute error is already in the desired order.
-    if force_lower_is_better:
-        pass
+    # The output does not need to be changed depending on the
+    # force_lower_is_better flag.
 
     return fom
 
@@ -494,16 +497,22 @@ def ssim(data, ground_truth, size=11, sigma=1.5, K1=0.01, K2=0.03,
         it is :math:`[0, 1]`. Default: `None`, obtain maximum and minimum
         from the ground truth.
     normalized  : bool, optional
-        Boolean flag to switch between unormalized and normalized FOM.
-        See `Notes` for details.
+        If ``True``, the output values are mapped to the interval
+        :math:`[0, 1]` (see `Notes` for details), otherwise return the
+        original SSIM.
     force_lower_is_better : bool, optional
-        Boolean flag which enforces that lower values correspond to better
-        matches by returning the negative of the (possibly normalized) SSIM.
+        If ``True``, it is ensured that lower values correspond to better
+        matches by returning the negative of the SSIM, otherwise the (possibly
+        normalized) SSIM is returned. If both `normalized` and
+        `force_lower_is_better` are ``True``, then the order is reversed before
+        mapping the outputs, so that the latter are still in the interval
+        :math:`[0, 1]`.
 
     Returns
     -------
     ssim : float
-        FOM value, where a higher value means a better match.
+        FOM value, where a higher value means a better match
+        if `force_lower_is_better` is ``False``.
 
     Notes
     -----
@@ -556,7 +565,7 @@ def ssim(data, ground_truth, size=11, sigma=1.5, K1=0.01, K2=0.03,
     result = np.mean(pointwise_ssim)
 
     if normalized:
-        result = (result + 1.0)/2.0
+        result = (result + 1.0) / 2.0
 
     if force_lower_is_better:
         result = -result
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -564,11 +564,11 @@ def ssim(data, ground_truth, size=11, sigma=1.5, K1=0.01, K2=0.03,
 
     result = np.mean(pointwise_ssim)
 
-    if normalized:
-        result = (result + 1.0) / 2.0
-
     if force_lower_is_better:
         result = -result
+
+    if normalized:
+        result = (result + 1.0) / 2.0
 
     return result
 
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -508,6 +508,7 @@ def ssim(data, ground_truth,
     window /= np.sum(window)
 
     def smoothen(img):
+        """Smoothes an image by convolving with a window function."""
         return fftconvolve(window, img, mode='valid')
 
     if dynamic_range is None:
@@ -526,9 +527,9 @@ def ssim(data, ground_truth,
     sigma2_sq = smoothen(ground_truth * ground_truth) - mu2_sq
     sigma12 = smoothen(data * ground_truth) - mu1_mu2
 
-    nom = (2 * mu1_mu2 + C1) * (2 * sigma12 + C2)
+    num = (2 * mu1_mu2 + C1) * (2 * sigma12 + C2)
     denom = (mu1_sq + mu2_sq + C1) * (sigma1_sq + sigma2_sq + C2)
-    pointwise_ssim = nom / denom
+    pointwise_ssim = num / denom
 
     return np.mean(pointwise_ssim)
 
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -17,7 +17,8 @@ __all__ = ('mean_squared_error', 'mean_absolute_error',
            'psnr', 'haarpsi', 'noise_power_spectrum')
 
 
-def mean_squared_error(data, ground_truth, mask=None, normalized=False):
+def mean_squared_error(data, ground_truth, mask=None,
+                       normalized=False, force_lower_is_better=False):
     """Return mean squared L2 distance between ``data`` and ``ground_truth``.
 
     See also `this Wikipedia article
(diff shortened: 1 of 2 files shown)


odl/contrib/fom/supervised.py
```diff
@@ -473,20 +494,41 @@ def ssim(data, ground_truth,
...
+        Difference between the maximum and minimum value that the pixels
+        can attain. Use 255 if pixel range is :math:`[0, 255]` and 1 if
+        it is :math:`[0, 1]`. Default: `None`, obtain maximum and minimum
+        from the ground truth.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+        See `Notes` for details.
+    force_lower_is_better : bool, optional
+        Boolean flag which enforces that lower values correspond to better
+        matches by returning the negative of the (possibly normalized) SSIM.
```
kohr-h: We usually write this as
```
If ``True``, [insert "True" case here], otherwise [what happens otherwise].
```
Applies to `normalized` as well.

odl/contrib/fom/supervised.py
```diff
@@ -72,10 +78,15 @@ def mean_squared_error(data, ground_truth, mask=None, normalized=False):
     else:
         fom /= l2norm_squared(data.space.one())
 
+    # The mean squared error is already in the desired order.
+    if force_lower_is_better:
+        pass
```
kohr-h: Just a comment suffices, the code really does nothing.

odl/contrib/fom/supervised.py
```diff
@@ -473,20 +494,41 @@ def ssim(data, ground_truth,
...
+        Small constants to stabilize the result. See [Wan+2004] for details.
+    dynamic_range : nonnegative float, optional
+        Difference between the maximum and minimum value that the pixels
+        can attain. Use 255 if pixel range is :math:`[0, 255]` and 1 if
+        it is :math:`[0, 1]`. Default: `None`, obtain maximum and minimum
+        from the ground truth.
+    normalized  : bool, optional
+        Boolean flag to switch between unormalized and normalized FOM.
+        See `Notes` for details.
+    force_lower_is_better : bool, optional
```
kohr-h: I'm not sure I like this flag. What's the reasoning behind it? I assume the goal of this and `normalized` is to have a set of FOMs that are immediately comparable.

I'm a bit worried about adding a parameter that does nothing in a lot of FOMs (all that end in `error` basically). For those we should state the purpose of the flag, namely that it does nothing in this case and exists to make all FOMs directly comparable, in particular the "error" type with the "goodness-of-fit" type.

odl/contrib/fom/supervised.py
```diff
@@ -519,11 +562,19 @@ def smoothen(img):
...
-    nom = (2 * mu1_mu2 + C1) * (2 * sigma12 + C2)
+    num = (2 * mu1_mu2 + C1) * (2 * sigma12 + C2)
     denom = (mu1_sq + mu2_sq + C1) * (sigma1_sq + sigma2_sq + C2)
-    pointwise_ssim = nom / denom
+    pointwise_ssim = num / denom
+
+    result = np.mean(pointwise_ssim)
+
+    if normalized:
+        result = (result + 1.0)/2.0
```
kohr-h: pep8: spaces around "/"

odl/contrib/fom/supervised.py
sbanert: Our motivation for adding this was that any compiler would complain about the unused variable, and so did the syntax checker of my Emacs environment. If this warning is not a problem for ODL, we can of course leave this out.

kohr-h: Yes, pylint would probably complain, too (maybe that's what Emacs uses). But in general I'm against adding code just to silence style checkers.

sbanert:
I addressed all the objections. Maybe we should discuss the force_lower_is_better flag again before merging the reconstruction_pipeline branch into master, since all pull requests are using it?

kohr-h:
> I addressed all the objections. Maybe we should discuss the force_lower_is_better flag again before merging the reconstruction_pipeline branch into master, since all pull requests are using it?

Right, I overlooked at first that this is not a PR into master. Then I think we'll just go ahead and look more closely when the final PR comes up.
What's the plan with this? Should it be merged to the branch now?

aringh:
I think the idea was to collect the common effort we did on Tuesday (10/10) on one branch and then merge it into ``master``. But I think that, at least from @adler-j perspective, before merging all these pull requests into ``reconstruction_pipeline`` us mere mortals to give it a go and try some code reviewing ;-)

adler-j:
We'll I could do the reviewing, but I thought this was an excellent occation for people other than me and @kohr-h to do some of the more "administrative" work so to say. Nice way to get going and no need to feel that you can do something wrong, since we have another branch before the main merge (where you could break "production" code).

Anyway, as soon as someone else gives these an "approve" I'll merge. Until then I'll patiently wait :)

odl/contrib/fom/supervised.py
```diff
@@ -457,8 +480,9 @@ def false_structures(data, ground_truth, mask=None, normalized=False,
     return mean_squared_error(data, ground_truth, mask, normalized)
 
 
-def ssim(data, ground_truth,
-         size=11, sigma=1.5, K1=0.01, K2=0.03, dynamic_range=None):
+# TODO Add details on the computation in the docstring.
```
adler-j: Please make an issue about this

odl/contrib/fom/supervised.py
```diff
@@ -473,20 +497,47 @@ def ssim(data, ground_truth,
...
     -------
     ssim : float
-        FOM value, where a higher value means a better match.
+        FOM value, where a higher value means a better match
+        if `force_lower_is_better` is ``False``.
+
+    Notes
+    -----
+    The unnormalized values are in the interval :math:`[-1, 1]`, where 1
+    corresponds to a perfect match. The normalized values are obtained by
```
adler-j: what does -1 correspond to?

odl/contrib/fom/supervised.py
```diff
@@ -519,11 +571,19 @@ def smoothen(img):
...
-    pointwise_ssim = nom / denom
+    pointwise_ssim = num / denom
+
+    result = np.mean(pointwise_ssim)
+
+    if normalized:
+        result = (result + 1.0) / 2.0
+
+    if force_lower_is_better:
+        result = -result
```
adler-j: This looks like it would return the wrong value if both `force_lower_is_better=True` and `normalize=True`. Reordering these if statements fixes that.

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,23 @@ def test_psnr(space):
...
+    # SSIM with ground truth zero should always give zero.
+    data = odl.phantom.white_noise(space)
+    result = odl.contrib.fom.ssim(data, space.zero())
+    assert pytest.approx(result) == 0
+
+    # SSIM should be symmetric if the dynamic range is set explicitly.
+    result1 = odl.contrib.fom.ssim(data, ground_truth, dynamic_range=1)
+    result2 = odl.contrib.fom.ssim(ground_truth, data, dynamic_range=1)
+    assert pytest.approx(result1) == result2
+
```
adler-j: also needs tests for the flags.

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,23 @@ def test_psnr(space):
...
+    # SSIM of true image should be one.
+    result = odl.contrib.fom.ssim(ground_truth, ground_truth)
+    assert pytest.approx(result) == 1
+
+    # SSIM with ground truth zero should always give zero.
+    data = odl.phantom.white_noise(space)
+    result = odl.contrib.fom.ssim(data, space.zero())
+    assert pytest.approx(result) == 0
+
```
adler-j: What should give -1?

adler-j:
Minor comments could use a fix before we merge

odl/contrib/fom/supervised.py
sbanert: Done.

odl/contrib/fom/supervised.py
sbanert: Does not seem to be so easy … first, the constants K1 and K2 (which prevent from dividing by zero) have to be set to zero, and even then the smallest possible value can only be achieved in special situations.

adler-j: Well in general there are basically two cases

either

* d(x, 0)

or 

* d(x, -x)

is minimal. In this case I guess that -1 corresponds to the second case?

sbanert: Checked this with Spyder. Some examples resulting from our implementation:

odl.contrib.fom.ssim([1], [-1])
Out[3]: 1.0

odl.contrib.fom.ssim([1], [1])
Out[4]: 1.0

odl.contrib.fom.ssim([1, -1], [-1, 1])
Out[5]: 0.57007198720979002

odl.contrib.fom.ssim([1, 0], [0, 1])
Out[6]: -0.080178871240947275

odl.contrib.fom.ssim([1, 0, 1], [0, 1, 0])
Out[7]: -0.16322592107522449

odl.contrib.fom.ssim([1, -1, 1], [-1, 1, -1])
Out[8]: 0.87354822079002448

odl.contrib.fom.ssim([1, -1, 1], [0, 0, 0])
Out[9]: 0.0

odl/contrib/fom/test/test_supervised.py
sbanert: Wouldn’t it be easier to add the corresponding fixtures to all FOM tests after merging?

sbanert: Tests for the flags are now added.

sbanert:
A comment on something which might be completely unrelated to this pull request, but which I noted by executing the tests: There are some random failures, most often in test_unsupervised.py, but also in test_supervised.py, e.g. the following:

=================================================== FAILURES ====================================================
 test_general[ space=uniform_discr(0.0, 1.0, 10) - FOM=<function standard_deviation_difference at 0x7f535f8f6c80> ] 

space = uniform_discr(0.0, 1.0, 10), scalar_fom = <function standard_deviation_difference at 0x7f535f8f6c80>

    def test_general(space, scalar_fom):
        """Test general properties of FOMs."""
        for name, ground_truth in space.examples:
            ground_truth = np.abs(ground_truth)
            scale = np.max(ground_truth) - np.min(ground_truth)
            noise = scale * np.abs(noise_element(space))
            data = ground_truth + noise
    
            # Check that range is a real number
            assert np.isscalar(scalar_fom(data, ground_truth))
    
            # Check that FOM is minimal when ground truth is compared with itself
            assert (scalar_fom(ground_truth, ground_truth) <=
                    scalar_fom(data, ground_truth))
    
            # Check that FOM is monotonic wrt noise level
>           assert (scalar_fom(ground_truth + noise, ground_truth) <=
                    scalar_fom(ground_truth + 2 * noise, ground_truth))
E           assert 0.19965137644341813 <= 0.0010310703059935866
E            +  where 0.19965137644341813 = <function standard_deviation_difference at 0x7f535f8f6c80>((uniform_discr(0.0, 1.0, 10).element([0.0, 0.0, 0.0, ..., 1.0, 1.0, 1.0]) + uniform_discr(0.0, 1.0, 10).element([1.1443617334697693, 1.4944476697569422, 0.59670962891352408, ..., 0.40638468796214089, 0.19865931921456095, 0.49401345831140742])), uniform_discr(0.0, 1.0, 10).element([0.0, 0.0, 0.0, ..., 1.0, 1.0, 1.0]))
E            +  and   0.0010310703059935866 = <function standard_deviation_difference at 0x7f535f8f6c80>((uniform_discr(0.0, 1.0, 10).element([0.0, 0.0, 0.0, ..., 1.0, 1.0, 1.0]) + (2 * uniform_discr(0.0, 1.0, 10).element([1.1443617334697693, 1.4944476697569422, 0.59670962891352408, ..., 0.40638468796214089, 0.19865931921456095, 0.49401345831140742]))), uniform_discr(0.0, 1.0, 10).element([0.0, 0.0, 0.0, ..., 1.0, 1.0, 1.0]))

test_supervised.py:45: AssertionError

adler-j:
Nice catch, I guess we need to loosen some boundaries for those tests, I'll have a look!

odl/contrib/fom/supervised.py
adler-j: I guess we'll leave this for now then since this seems complicated

odl/contrib/fom/test/test_supervised.py
```diff
@@ -124,6 +124,53 @@ def test_psnr(space):
...
+    # * 0 with force_lower_is_better == True and normalized == True.
+    result = odl.contrib.fom.ssim(ground_truth, ground_truth)
+    result_normalized = odl.contrib.fom.ssim(ground_truth, ground_truth,
+                                             normalized=True)
+    result_flib = odl.contrib.fom.ssim(ground_truth, ground_truth,
+                                       force_lower_is_better=True)
+    result_nf = odl.contrib.fom.ssim(ground_truth, ground_truth,
+                                     normalized=True,
+                                     force_lower_is_better=True)
+    assert pytest.approx(result) == 1
```
adler-j: Usually we write these the other way around (no need to fix now), but language wise it reads nicer as "result is approximately 1", instead of "approximately result is 1"

sbanert:
Or we need to check if all FOMs are really monotonic wrt the noise level “almost surely”.
The assertion 0.19965137644341813 <= 0.0010310703059935866 does not look like a mere inaccuracy.

odl/contrib/fom/test/test_supervised.py
```diff
@@ -124,6 +124,53 @@ def test_psnr(space):
     assert result == pytest.approx(expected)
 
 
+def test_ssim(space):
```
adler-j: This is fine for now, but style wise (for next time), it could probably be written using fixtures for `force_lower_is_better` and `normalized`, which would significantly reduce duplication.

adler-j:
> Or we need to check if all FOMs are really monotonic wrt the noise level “almost surely”.
The assertion 0.19965137644341813 <= 0.0010310703059935866 does not look like a mere inaccuracy

I guess in this case the problem is that "standard deviation difference" is somewhat of a vague fom, I'll see what goes wrong here. Is it always that fom that fails?

sbanert:
> I guess in this case the problem is that "standard deviation difference" is somewhat of a vague fom, I'll see what goes wrong here. Is it always that fom that fails?

I have not tested extensively, only about 10 times. I got a failure for the SSIM symmetry test resulting from an inaccuracy of order 10^(-7). Unfortunately, I have not logged this.

PR #1187, opened 2017-10-10
Optimize parameters

sbanert:
Move the example regarding parameter optimization to a contrib module. Separate function from example. Add tests.


odl/contrib/param_opt/param_opt.py
```diff
@@ -0,0 +1,94 @@
...
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Method(s) to find optimal reconstruction/regularization parameters with
+respect to given figures of merit."""
+
+# import odl
```
aringh: Should be removed.

odl/contrib/param_opt/param_opt.py
```diff
@@ -0,0 +1,94 @@
...
+import numpy as np
+import scipy
+
+__all__ = ('optimal_parameters', )
+
+
+def optimal_parameters(reconstruction, fom, phantoms, data,
+                       initial_param=0):
```
aringh: Is the initial value ``0`` of the form ``array-like``, as the doc specifics? This is an open question, I do not know how python handles it.

aringh: On line 77 it seems to be cast to an array though...

sbanert: I googled it and found the following answer (yes, since the code works with `np.array` instead of `np.asarray` on line 77):
https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects
The hint ‘optional’ and the default values in the doc are missing, though.

adler-j:
I'll go ahead and merge this.

PR #1188, opened 2017-10-10
Reco pipeline mob

bgris:
team day work : tests on FOMS and evaluation

### DIFF
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -178,13 +178,13 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False):
 
     .. math::
          \mathrm{MVD}(f, g) =
-         \\Big| |\\overline{f}| - |\\overline{g}| \\Big|,
+         \\Big| \\overline{f} - \\overline{g} \\Big|,
 
     or, in normalized form
 
     .. math::
          \mathrm{MVD_N}(f, g) =
-         \\frac{\\Big| |\\overline{f}| - |\\overline{g}| \\Big|}
+         \\frac{\\Big| \\overline{f} - \\overline{g} \\Big|}
                {\\Big| |\\overline{f}| + |\\overline{g}| \\Big|}
 
     where :math:`\\overline{f}` is the mean value of :math:`f`,
@@ -205,7 +205,7 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False):
     data_mean = data.inner(data.space.one()) / vol
     ground_truth_mean = ground_truth.inner(ground_truth.space.one()) / vol
 
-    fom = np.abs(np.abs(data_mean) - np.abs(ground_truth_mean))
+    fom = np.abs(data_mean - ground_truth_mean)
 
     if normalized:
         fom /= (np.abs(data_mean) + np.abs(ground_truth_mean))
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -272,12 +272,17 @@ def standard_deviation_difference(data, ground_truth, mask=None,
     data_mean = data.inner(data.space.one()) / vol
     ground_truth_mean = ground_truth.inner(ground_truth.space.one()) / vol
 
-    fom = np.abs((l2_norm(data - data_mean) -
-                  l2_norm(ground_truth - ground_truth_mean)))
+    deviation_data = l2_norm(data - data_mean)
+    deviation_ground_truth = l2_norm(ground_truth - ground_truth_mean)
+    fom = np.abs(deviation_data - deviation_ground_truth)
 
     if normalized:
-        fom /= (l2_norm(data - data_mean) +
-                l2_norm(ground_truth - ground_truth_mean))
+        sum_deviation = deviation_data + deviation_ground_truth
+        fom /=  sum_deviation
+        if np.isnan(fom):
+            fom = 0
+
+
 
     return fom
 
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -154,7 +154,7 @@ def mean_absolute_error(data, ground_truth, mask=None,
 
 
 def mean_value_difference(data, ground_truth, mask=None, normalized=False,
-                          force_lower_is_better=False):
+                          force_lower_is_better=True):
     """Return difference in mean value between ``data`` and ``ground_truth``.
 
     Parameters
@@ -167,6 +167,11 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False,
         If given, ``data * mask`` is compared to ``ground_truth * mask``.
     normalized  : bool, optional
         Boolean flag to switch between unormalized and normalized FOM.
+    force_lower_is_better : bool, optional
+        If ``True``, it is ensured that lower values correspond to better
+        matches. For the mean value difference, this is already the case, and
+        the flag is only present for compatibility to other figures of merit.
+
 
     Returns
     -------
@@ -216,7 +221,7 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False,
 
 
 def standard_deviation_difference(data, ground_truth, mask=None,
-                                  normalized=False, force_lower_is_better=False):
+                                  normalized=False, force_lower_is_better=True):
     """Return absolute difference in std between ``data`` and ``ground_truth``.
 
     Parameters
@@ -229,6 +234,11 @@ def standard_deviation_difference(data, ground_truth, mask=None,
         If given, ``data * mask`` is compared to ``ground_truth * mask``.
     normalized  : bool, optional
         Boolean flag to switch between unormalized and normalized FOM.
+    force_lower_is_better : bool, optional
+        If ``True``, it is ensured that lower values correspond to better
+        matches. For the standard deviation difference, this is already the case, and
+        the flag is only present for compatibility to other figures of merit.
+
 
     Returns
     -------
@@ -287,7 +297,7 @@ def standard_deviation_difference(data, ground_truth, mask=None,
     return fom
 
 
-def range_difference(data, ground_truth, mask=None, normalized=False):
+def range_difference(data, ground_truth, mask=None, normalized=False, force_lower_is_better=True):
     """Return dynamic range difference between ``data`` and ``ground_truth``.
 
     Evaluates difference in range between input (``data``) and reference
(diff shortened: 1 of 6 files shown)


odl/contrib/fom/supervised.py
```diff
@@ -130,7 +130,8 @@ def mean_absolute_error(data, ground_truth, mask=None, normalized=False):
     return fom
 
 
-def mean_value_difference(data, ground_truth, mask=None, normalized=False):
+def mean_value_difference(data, ground_truth, mask=None, normalized=False,
+                          force_lower_is_better=False):
```
aringh: The flag should be added to the doc as well, describing what it does.

odl/contrib/fom/supervised.py
```diff
@@ -182,16 +183,17 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False):
...
     if normalized:
         fom /= (np.abs(data_mean) + np.abs(ground_truth_mean))
 
+
     return fom
 
 
 def standard_deviation_difference(data, ground_truth, mask=None,
-                                  normalized=False):
+                                  normalized=False, force_lower_is_better=False):
```
aringh: Same here: the flag should be added to the doc, and maybe a comment in the code

odl/contrib/fom/supervised.py
aringh: Also, even if the flag does not change anything I think it is good to have a comment in the code that states this. See, .e.g., how @sbanert did [here](https://github.com/sbanert/odl/blob/952db84246ac131648aa09523239160e5a874110/odl/contrib/fom/supervised.py#L82-L84).

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
     assert result == pytest.approx(expected)
 
 
+def test_mean_value_difference_sign():
```
aringh: Use a fixture for space, as in the ``test_mean_value_difference_range_value``, or is this not necessary? Could tolerance be calibrated give what space one tests against, or is this over-kill?

odl/contrib/fom/supervised.py
```diff
@@ -247,12 +249,17 @@ def standard_deviation_difference(data, ground_truth, mask=None,
...
+    deviation_data = l2_norm(data - data_mean)
+    deviation_ground_truth = l2_norm(ground_truth - ground_truth_mean)
+    fom = np.abs(deviation_data - deviation_ground_truth)
 
     if normalized:
-        fom /= (l2_norm(data - data_mean) +
-                l2_norm(ground_truth - ground_truth_mean))
+        sum_deviation = deviation_data + deviation_ground_truth
+        fom /=  sum_deviation
+        if np.isnan(fom):
```
adler-j: Prefer checking this before:

```
if fom == sum_deviation == 0:
    fom = 0.0
else:
    fom /= sum_deviation
```

this should also be documented

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+    space = odl.uniform_discr(0, 1, 10)
+    I0 = space.one()
+    I1 = -I0.copy()
```
adler-j: `-` creates a copy (where else would the result be stored?), no need to do it twice

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+    assert np.abs(odl.contrib.fom.mean_value_difference(I0, I1)) > 0.1
```
adler-j: we should do `from odl.contrib import fom` up top to make these lines much more readable.

adler-j: Further, in this case we know the exact answer, right? It should be 2.0, so test against that.

```
assert fom.mean_value_difference(I0, -I0) == pytest.approx(2.0)
```

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+    assert np.abs(odl.contrib.fom.mean_value_difference(I0, I1,
+                  normalized=True)) > 0.1
+
+
+def test_mean_value_difference_range_value(space):
+    I0 = space.element(np.random.normal(0, 1, size=space.shape))
+    I1 = space.element(np.random.normal(0, 1, size=space.shape))
+
+    max0 = np.max(I0.asarray())
```
adler-j: no need for `.asarray`, simply `max0 = np.max(I0)` works fine.

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+    space = odl.uniform_discr(0, 1, 10)
+    I0 = space.one()
+    I1 = -I0.copy()
+    assert np.abs(odl.contrib.fom.mean_value_difference(I0, I1)) > 0.1
+    assert np.abs(odl.contrib.fom.mean_value_difference(I0, I1,
+                  normalized=True)) > 0.1
+
+
+def test_mean_value_difference_range_value(space):
+    I0 = space.element(np.random.normal(0, 1, size=space.shape))
```
adler-j: there is a utility `odl.util.testutils.noise_element` which does exactly this (but works with e.g. complex numbers etc equally well).

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+
+    max0 = np.max(I0.asarray())
+    max1 = np.max(I1.asarray())
+    min0 = np.min(I0.asarray())
+    min1 = np.min(I1.asarray())
+
+
+    assert odl.contrib.fom.mean_value_difference(I0, I1) <= max(max0 - min1,
+                                                max1 - min0)
+    assert pytest.approx(odl.contrib.fom.mean_value_difference(I0, I0)) == 0
```
adler-j: usually we write this the other way around

value is approximately true value

```
assert fom.mean_value_difference(I0, I0) == pytest.approx(0)
```

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+    I1 = space.element(np.random.normal(0, 1, size=space.shape))
+
+    max0 = np.max(I0.asarray())
+    max1 = np.max(I1.asarray())
+    min0 = np.min(I0.asarray())
+    min1 = np.min(I1.asarray())
+
+
+    assert odl.contrib.fom.mean_value_difference(I0, I1) <= max(max0 - min1,
+                                                max1 - min0)
```
adler-j: Indentation is off

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+
+
+def test_standard_deviation_difference_range_value(space):
+    I0 = space.element(np.random.normal(0, 1, size=space.shape))
+    const = np.random.normal(0, 10)
+
+    assert pytest.approx(odl.contrib.fom.standard_deviation_difference(
+            I0, I0)) == 0
+    assert odl.contrib.fom.standard_deviation_difference(10*I0, I0,
+            normalized=True) <= 1.0
```
adler-j: why should this hold?

odl/contrib/fom/test/test_supervised.py
```diff
@@ -68,5 +68,62 @@ def test_psnr(space):
...
+def test_range_difference(space):
+    I0 = space.element(np.random.normal(0, 1, size=space.shape))
+    I1 = space.element(np.random.normal(0, 1, size=space.shape))
+    const = np.random.normal(0, 10)
+
+    assert pytest.approx(odl.contrib.fom.range_difference(
+            I0, I0)) == 0
+    assert pytest.approx(odl.contrib.fom.range_difference(
+            I0 + const, I0)) == 0
+    aconst=np.abs(const)
```
adler-j: PEP8, spaces round =

adler-j:
Would be nice to get this updated so we can merge

adler-j:
I guess this could use some fixes so we can merge it, waste if it does not get finished.

adler-j:
This is the last remaining PR before #1280, lets get it out of the way

odl/contrib/fom/supervised.py
```diff
@@ -130,7 +130,8 @@ def mean_absolute_error(data, ground_truth, mask=None, normalized=False):
     return fom
 
 
-def mean_value_difference(data, ground_truth, mask=None, normalized=False):
+def mean_value_difference(data, ground_truth, mask=None, normalized=False,
+                          force_lower_is_better=False):
     """Return difference in mean value between ``data`` and ``ground_truth``.
```
sbanert: Can you change this to a raw string with r"""string""" and remove the double backslashes in the docstring? Otherwise, I think that you should change \mathrm to \\mathrm.

odl/contrib/fom/supervised.py
```diff
@@ -182,16 +183,17 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False):
...
         fom /= (np.abs(data_mean) + np.abs(ground_truth_mean))
 
+
     return fom
 
 
 def standard_deviation_difference(data, ground_truth, mask=None,
-                                  normalized=False):
+                                  normalized=False, force_lower_is_better=False):
     """Return absolute difference in std between ``data`` and ``ground_truth``.
```
sbanert: Same here: either raw string or inserting double backslashes

odl/contrib/fom/test/test_supervised.py
bgris: This is in case we want to normalize.

kohr-h:
Something I've always wondered: what does "mob" stand for here? Does it refer to ["A large or disorderly group of people; especially one bent on riotous or destructive action"?](https://en.wiktionary.org/wiki/mob) 😉

sbanert:
To warm this up: I think that there is something wrong with the rebase/merge: The commits of @adler-j and me should not appear here because they are already in the branch you want to merge in. It might be possible to resolve this problem and also failing Travis by a rebase on master.

adler-j:
Final bump here. I'll simply close this and go ahead with the rest of the PR's before the end of April.

aringh:
New PR for this in #1353

PR #1190, opened 2017-10-11
BUG: disable BLAS for too large arrays (int overflow), closes #1189

### LINKED ISSUE #1189: Unary minus broken
```python
>>> np.sum(-odl.rn(2**31).one())
0.0
>>> np.sum(-odl.rn(2**31-1).one())
-2147483647.0
```

kohr-h:


adler-j:
Blas only works for int32? We truly need to check if this is really the case with the underlying implementation, sounds outrageous. Are we sure it is not our `native` call?

adler-j:
Anyway, I'd say add a largescale test for this before we merge.

kohr-h:
There are weird things going on in BLAS, see also [this issue](https://github.com/numpy/numpy/issues/9852). And yes, you can trigger the error in #1189 completely independently of ODL, no `native` issue whatsoever.

adler-j:
Now I never thought we'd find BLAS bugs by running ODL, we truly are pushing some limits here.

My suggestion is we add a largescale test for a very large space and test all arithmetic on it. I think all that is needed is basically adding another space to a fixture

kohr-h:
Note that there is another [issue](https://github.com/xianyi/OpenBLAS/issues/1326) with using BLAS for `float` arrays that falls into the category "wrong, but expected". The nastiness of this one is that it depends on the size of the final result as well as on the implementation of the summation itself (size of partial sums etc.).

kohr-h:
> Approving this. Long term we could do sub-sets ourselves.

I had the same thought lately. Made a new issue: #1200

Issue #597, opened 2016-09-22
Implement smoothed L1 style "norms"


adler-j:
We should add the option of smoothing L1 norms so that they can be used with smooth solvers.
### Smoothed L1

The most simple way, which i say we go for, is to add a parameter `smoothing` to `L1Norm` so users can call it as such

```
>>> space = rn(2)
>>> l1 = L1Norm(space, smoothing=0.01)
>>> l1([1, 2]) == (sqrt(1**2 + 0.01**2) + sqrt(2**2 + 0.01**2))
```

Precisely, with smoothness `eps`

```
smoothed L1 = sum_i sqrt(eps^2 + x_i^2)
```

This has the nice property of having a easy to compute gradient `L1Norm(space, smoothing).gradient(p) == p / np.sqrt(p**2 + smoothing**2)`
### Huber

[Huber regularization](https://en.wikipedia.org/wiki/Huber_loss) is similar, but instead of a smooth transition it agrees almost exactly (down to a offset) with the l1 norm in case `x > smoothing` for some (small) scale `smoothing`, and agrees exactly with the l2 norm for `x < smoothing`

Precisely, with smoothness `eps`

```
Huber L1 = sum_i { x_i^2 / (2 eps)   if  |x_i| \leq eps
                   |x_i| - eps / 2   else

```
### The design question

To me, there are a few options in how to implement this
- Create separate classes for the above cases
- Let the above cases all be options to `L1Norm`
- Create a new `HuberFunctional` class, but let the smoothed case be an option to `L1Norm`
### Notes
- @ozanoktem has expressed interest in this so I'll assign to him.
- Along with implementing this we could use an example of using this with a smooth solver to solve TV regularized problems.

aringh:
Working on other things, I needed a Huber norm. So I made a crude implementation of the Huber norm (it only has call and gradient and does NOT work for all kind of linear spaces) and then I cherry-picket that commit and put it in another branch (called issue-597__smooth_l1_and_huber). If @ozanoktem is still interested in implementing this, he could start from there :)

adler-j:
Could you make a separate branch with only the huber norm @aringh?

aringh:
I created a new branch (issue-597__huber_norm), cherry-picked the "huber-norm commit", pushed the branch, and removed the other branch. That was the easiest way I found :)

kohr-h:
Well done :-)

adler-j:
Pull request incomming?

adler-j:
[...]

PR #1191, opened 2017-10-11
Issue1101 pdhg examples

mehrhardt:
Here are the two examples I promised a few weeks ago. I suppose the documentation can be improved but first I would like to have some more general feedback. For these, I also needed the Huber L1 norm for which I can compute the prox of the convex conjugate and its function evaluation.

Please let me know what you think about these.

### DIFF
--- a/examples/solvers/pdhg_denoising_L1_HuberTV.py
+++ b/examples/solvers/pdhg_denoising_L1_HuberTV.py
@@ -5,8 +5,9 @@ This exhaustive example solves the L1-HuberTV problem
         min_{x >= 0} ||x - d||_1
             + lam * sum_i eta_gamma(||grad(x)_i||_2)
 
-where grad the spatial gradient and d is given noisy data. Here eta_gamma
-denotes the Huber function. For more details, see the Huber documentation.
+where ``grad`` is the spatial gradient and ``d`` is given noisy data. Here
+``eta_gamma`` denotes the Huber function. For more details, see the Huber
+documentation.
 
 For further details and a description of the solution method used, see
 https://odlgroup.github.io/odl/guide/pdhg_guide.html in the ODL documentation.
@@ -29,7 +30,7 @@ d = odl.phantom.salt_pepper_noise(orig)
 
 # Define objective functional
 op = odl.Gradient(space)  # operator
-op.norm = np.sqrt(8) + 1e-4  # norm with forward differences is well-known
+norm_op = np.sqrt(8) + 1e-4  # norm with forward differences is well-known
 lam = 1  # Regularization parameter
 g = 1 / lam * odl.solvers.L1Norm(space).translated(d)  # data fit
 f = odl.solvers.Huber(op.range, gamma=.1)  # regularization
@@ -56,10 +57,10 @@ class CallbackStore(odl.solvers.Callback):  # Callback to store function values
         self.obj_function_values = []
 
 
-callback = (odl.solvers.CallbackPrintIteration(step=10) & CallbackStore())
+callback = odl.solvers.CallbackPrintIteration(step=10) & CallbackStore()
 niter = 500  # Number of iterations
-tau = 1.0 / op.norm  # Step size for primal variable
-sigma = 1.0 / op.norm  # Step size for dual variable
+tau = 1.0 / norm_op  # Step size for primal variable
+sigma = 1.0 / norm_op  # Step size for dual variable
 
 # Run algorithm
 x = space.zero()
--- a/examples/solvers/pdhg_denoising_L1_HuberTV.py
+++ b/examples/solvers/pdhg_denoising_L1_HuberTV.py
@@ -58,13 +58,13 @@ reg_param = 1
 l1_norm = 1 / reg_param * odl.solvers.L1Norm(space).translated(noisy)
 
 # HuberTV-regularization
-huber_l1_norm = odl.solvers.HuberL1L2(gradient.range, gamma=.1)
+huber = odl.solvers.Huber(gradient.range, gamma=.1)
 
 # Define objective
-obj_fun = l1_norm + huber_l1_norm * gradient
+obj_fun = l1_norm + huber * gradient
 
 # Strong convexity of "f*"
-strong_convexity = 1 / huber_l1_norm.grad_lipschitz
+strong_convexity = 1 / huber.grad_lipschitz
 
 
 # Define callback to store function values
@@ -92,7 +92,7 @@ niter = 500
 
 # Assign operator and functionals
 op = gradient
-f = huber_l1_norm
+f = huber
 g = l1_norm
 
 tau = 1.0 / gradient.norm  # Step size for primal variable
--- a/examples/solvers/pdhg_denoising_L1_HuberTV.py
+++ b/examples/solvers/pdhg_denoising_L1_HuberTV.py
@@ -1,22 +1,12 @@
 """Total variation denoising using PDHG.
 
-This exhaustive example solve the L1-HuberTV problem
+This exhaustive example solves the L1-HuberTV problem
 
-    .. math::
-        \\min_{x >= 0} ||x - d||_1
-            + \\lambda \\sum_i \\eta_\\gamma(||grad(x)_i||_2)
+        min_{x >= 0} ||x - d||_1
+            + lam * sum_i eta_gamma(||grad(x)_i||_2)
 
-where ``grad`` the spatial gradient and ``d`` is given noisy data. Here
-``\\eta_\\gamma`` denotes the Huber function defined as
-
-    .. math::
-        \\eta_\\gamma(x) =
-        \\begin{cases}
-            \\frac{1}{2 \\gamma} x^2 + \\frac{gamma}{2}
-                & \\text{if } |x| \leq \\gamma \\\\
-            |x|
-                & \\text{if } |x| > \\gamma,
-        \\end{cases}.
+where grad the spatial gradient and d is given noisy data. Here eta_gamma
+denotes the Huber function. For more details, see the Huber documentation.
 
 For further details and a description of the solution method used, see
 https://odlgroup.github.io/odl/guide/pdhg_guide.html in the ODL documentation.
@@ -29,46 +19,27 @@ import matplotlib.pyplot as plt
 
 # --- define setting --- #
 
-# Read test image: use only every second pixel, convert integer to float
+# Define ground truth, space and noisy data
 image = np.rot90(scipy.misc.ascent()[::2, ::2].astype('float'), 3)
 shape = image.shape
-
-# Rescale max to 1
 image /= image.max()
-
-# Discretized space
 space = odl.uniform_discr([0, 0], shape, shape)
-
-# Create space element of ground truth
 orig = space.element(image.copy())
+d = odl.phantom.salt_pepper_noise(orig)
 
-# Create noisy image
-noisy = odl.phantom.salt_pepper_noise(orig)
+# Define objective functional
+op = odl.Gradient(space)  # operator
+op.norm = np.sqrt(8) + 1e-4  # norm with forward differences is well-known
+lam = 1  # Regularization parameter
+g = 1 / lam * odl.solvers.L1Norm(space).translated(d)  # data fit
+f = odl.solvers.Huber(op.range, gamma=.1)  # regularization
+obj_fun = f * op + g  # combined functional
+mu_f = 1 / f.grad_lipschitz  # Strong convexity of "f*"
 
-# Gradient operator
-gradient = odl.Gradient(space)
-
-# The operator norm of the gradient with forward differences is well-known
-gradient.norm = np.sqrt(8) + 1e-4
-
-# Regularization parameter
-reg_param = 1
-
-# l1 data matching
-l1_norm = 1 / reg_param * odl.solvers.L1Norm(space).translated(noisy)
-
-# HuberTV-regularization
-huber = odl.solvers.Huber(gradient.range, gamma=.1)
-
-# Define objective
-obj_fun = l1_norm + huber * gradient
-
-# Strong convexity of "f*"
-strong_convexity = 1 / huber.grad_lipschitz
+# Define algorithm parameters
 
 
-# Define callback to store function values
-class CallbackStore(odl.solvers.Callback):
+class CallbackStore(odl.solvers.Callback):  # Callback to store function values
     def __init__(self):
         self.iteration_count = 0
         self.iteration_counts = []
@@ -85,23 +56,15 @@ class CallbackStore(odl.solvers.Callback):
         self.obj_function_values = []
 
 
-callback = (odl.solvers.CallbackPrintIteration() & CallbackStore())
+callback = (odl.solvers.CallbackPrintIteration(step=10) & CallbackStore())
+niter = 500  # Number of iterations
+tau = 1.0 / op.norm  # Step size for primal variable
+sigma = 1.0 / op.norm  # Step size for dual variable
 
-# Number of iterations
-niter = 500
-
-# Assign operator and functionals
-op = gradient
-f = huber
-g = l1_norm
-
-tau = 1.0 / gradient.norm  # Step size for primal variable
-sigma = 1.0 / gradient.norm  # Step size for dual variable
-
-# Run algorithms 2 and 3
+# Run algorithm
 x = space.zero()
-callback(x)
-odl.solvers.pdhg(x, f, g, op, tau, sigma, niter, gamma_dual=strong_convexity,
+callback(x)  # store values for initialization
+odl.solvers.pdhg(x, f, g, op, tau, sigma, niter, gamma_dual=mu_f,
                  callback=callback)
 obj = callback.callbacks[1].obj_function_values
 
@@ -111,7 +74,7 @@ clim = [0, 1]
 cmap = 'gray'
 
 orig.show('original', clim=clim, cmap=cmap)
-noisy.show('noisy', clim=clim, cmap=cmap)
+d.show('noisy', clim=clim, cmap=cmap)
 x.show('denoised', clim=clim, cmap=cmap)
 
 
@@ -123,13 +86,12 @@ def rel_fun(x):
 
 i = np.array(callback.callbacks[1].iteration_counts)
 
-plt.figure(1)
-plt.clf()
-plt.loglog(i, rel_fun(obj), label='alg')
+plt.figure()
+plt.loglog(i, rel_fun(obj), label='pdhg')
 plt.loglog(i[1:], 1. / i[1:], '--', label='$O(1/k)$')
-plt.loglog(i[1:], 4. / i[1:]**2, ':', label='$O(1/k^2)$')
+plt.loglog(i[1:], 4. / i[1:] ** 2, ':', label='$O(1/k^2)$')
 rho = 0.97
-plt.loglog(i[1:], rho**i[1:], '-',
+plt.loglog(i[1:], rho ** i[1:], '-',
            label='$O(\\rho^k), \\rho={:3.2f}$'.format(rho))
 plt.title('Function values')
 plt.legend()
--- a/examples/solvers/pdhg_denoising_L1_HuberTV.py
+++ b/examples/solvers/pdhg_denoising_L1_HuberTV.py
@@ -1,10 +1,22 @@
 """Total variation denoising using PDHG.
 
-Solve the L1-HuberTV problem
+This exhaustive example solve the L1-HuberTV problem
 
-        min_{x >= 0} ||x - d||_1 + lam TV_gamma(x)
+    .. math::
+        \\min_{x >= 0} ||x - d||_1
+            + \\lambda \\sum_i \\eta_\\gamma(||grad(x)_i||_2)
 
-where ``grad`` the spatial gradient and ``d`` is given noisy data.
+where ``grad`` the spatial gradient and ``d`` is given noisy data. Here
+``\\eta_\\gamma`` denotes the Huber function defined as
+
+    .. math::
+        \\eta_\\gamma(x) =
+        \\begin{cases}
+            \\frac{1}{2 \\gamma} x^2 + \\frac{gamma}{2}
+                & \\text{if } |x| \leq \\gamma \\\\
+            |x|
+                & \\text{if } |x| > \\gamma,
+        \\end{cases}.
 
 For further details and a description of the solution method used, see
 https://odlgroup.github.io/odl/guide/pdhg_guide.html in the ODL documentation.
@@ -24,19 +36,22 @@ shape = image.shape
 # Rescale max to 1
 image /= image.max()
 
-# Discretized spaces
+# Discretized space
 space = odl.uniform_discr([0, 0], shape, shape)
 
 # Create space element of ground truth
 orig = space.element(image.copy())
 
-# Add noise and convert to space element
+# Create noisy image
 noisy = odl.phantom.salt_pepper_noise(orig)
 
 # Gradient operator
-gradient = odl.Gradient(space, method='forward')
+gradient = odl.Gradient(space)
 
-# regularization parameter
+# The operator norm of the gradient with forward differences is well-known
+gradient.norm = np.sqrt(8) + 1e-4
+
+# Regularization parameter
 reg_param = 1
 
 # l1 data matching
@@ -45,15 +60,15 @@ l1_norm = 1 / reg_param * odl.solvers.L1Norm(space).translated(noisy)
 # HuberTV-regularization
 huber_l1_norm = odl.solvers.HuberL1L2(gradient.range, gamma=.1)
 
-# define objective
+# Define objective
 obj_fun = l1_norm + huber_l1_norm * gradient
 
-# strong convexity of "f*"
+# Strong convexity of "f*"
 strong_convexity = 1 / huber_l1_norm.grad_lipschitz
 
 
-# define callback to store function values
-class CallbackStore(odl.solvers.util.callback.Callback):
+# Define callback to store function values
+class CallbackStore(odl.solvers.Callback):
     def __init__(self):
         self.iteration_count = 0
         self.iteration_counts = []
@@ -72,7 +87,7 @@ class CallbackStore(odl.solvers.util.callback.Callback):
 
 callback = (odl.solvers.CallbackPrintIteration() & CallbackStore())
 
-# number of iterations
+# Number of iterations
 niter = 500
 
 # Assign operator and functionals
@@ -80,11 +95,8 @@ op = gradient
 f = huber_l1_norm
 g = l1_norm
 
-# The operator norm of the gradient with forward differences is well-known
-op_norm = np.sqrt(8) + 1e-4
-
-tau = 1.0 / op_norm  # Step size for the primal variable
-sigma = 1.0 / op_norm  # Step size for the dual variable
+tau = 1.0 / gradient.norm  # Step size for primal variable
+sigma = 1.0 / gradient.norm  # Step size for dual variable
 
 # Run algorithms 2 and 3
 x = space.zero()
@@ -94,7 +106,7 @@ odl.solvers.pdhg(x, f, g, op, tau, sigma, niter, gamma_dual=strong_convexity,
 obj = callback.callbacks[1].obj_function_values
 
 # %% Display results
-# show images
+# Show images
 clim = [0, 1]
 cmap = 'gray'
 
@@ -103,7 +115,7 @@ noisy.show('noisy', clim=clim, cmap=cmap)
 x.show('denoised', clim=clim, cmap=cmap)
 
 
-# show convergence rate
+# Show convergence rate
 def rel_fun(x):
     x = np.array(x)
     return (x - min(x)) / (x[0] - min(x))
--- a/examples/solvers/pdhg_denoising_L1_HuberTV.py
+++ b/examples/solvers/pdhg_denoising_L1_HuberTV.py
@@ -1,0 +1,123 @@
+"""Total variation denoising using PDHG.
+
+Solve the L1-HuberTV problem
+
+        min_{x >= 0} ||x - d||_1 + lam TV_gamma(x)
+
+where ``grad`` the spatial gradient and ``d`` is given noisy data.
+
+For further details and a description of the solution method used, see
+https://odlgroup.github.io/odl/guide/pdhg_guide.html in the ODL documentation.
+"""
+
+import numpy as np
+import scipy
+import odl
+import matplotlib.pyplot as plt
+
+# --- define setting --- #
+
+# Read test image: use only every second pixel, convert integer to float
+image = np.rot90(scipy.misc.ascent()[::2, ::2].astype('float'), 3)
+shape = image.shape
+
+# Rescale max to 1
+image /= image.max()
+
+# Discretized spaces
+space = odl.uniform_discr([0, 0], shape, shape)
+
+# Create space element of ground truth
+orig = space.element(image.copy())
+
+# Add noise and convert to space element
+noisy = odl.phantom.salt_pepper_noise(orig)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
+
+# regularization parameter
+reg_param = 1
+
+# l1 data matching
+l1_norm = 1 / reg_param * odl.solvers.L1Norm(space).translated(noisy)
+
+# HuberTV-regularization
+huber_l1_norm = odl.solvers.HuberL1L2(gradient.range, gamma=.1)
+
+# define objective
+obj_fun = l1_norm + huber_l1_norm * gradient
+
+# strong convexity of "f*"
+strong_convexity = 1 / huber_l1_norm.grad_lipschitz
+
+
+# define callback to store function values
+class CallbackStore(odl.solvers.util.callback.Callback):
+    def __init__(self):
+        self.iteration_count = 0
+        self.iteration_counts = []
+        self.obj_function_values = []
+
+    def __call__(self, x):
+        self.iteration_count += 1
+        self.iteration_counts.append(self.iteration_count)
+        self.obj_function_values.append(obj_fun(x))
+
+    def reset(self):
+        self.iteration_count = 0
+        self.iteration_counts = []
+        self.obj_function_values = []
+
+
+callback = (odl.solvers.CallbackPrintIteration() & CallbackStore())
+
+# number of iterations
+niter = 500
+
+# Assign operator and functionals
+op = gradient
+f = huber_l1_norm
+g = l1_norm
+
+# The operator norm of the gradient with forward differences is well-known
+op_norm = np.sqrt(8) + 1e-4
+
+tau = 1.0 / op_norm  # Step size for the primal variable
+sigma = 1.0 / op_norm  # Step size for the dual variable
+
+# Run algorithms 2 and 3
+x = space.zero()
+callback(x)
+odl.solvers.pdhg(x, f, g, op, tau, sigma, niter, gamma_dual=strong_convexity,
+                 callback=callback)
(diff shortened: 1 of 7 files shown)


examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
+"""Total variation denoising using PDHG.
+
+Solve the L1-HuberTV problem
+
+        min_{x >= 0} ||x - d||_1 + lam TV_gamma(x)
+
+where ``grad`` the spatial gradient and ``d`` is given noisy data.
```
adler-j: there is no "grad" above, but i personally like the wrong version which is `||grad(x)||_1`

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
...
+# --- define setting --- #
+
+# Read test image: use only every second pixel, convert integer to float
+image = np.rot90(scipy.misc.ascent()[::2, ::2].astype('float'), 3)
+shape = image.shape
+
+# Rescale max to 1
+image /= image.max()
+
+# Discretized spaces
```
adler-j: space~s~

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
...
+space = odl.uniform_discr([0, 0], shape, shape)
+
+# Create space element of ground truth
+orig = space.element(image.copy())
+
+# Add noise and convert to space element
```
adler-j: there is no conversion in the line below

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
...
+noisy = odl.phantom.salt_pepper_noise(orig)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
```
adler-j: `method='forward'` is the default, so no need to add it

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
...
+
+# regularization parameter
+reg_param = 1
+
+# l1 data matching
+l1_norm = 1 / reg_param * odl.solvers.L1Norm(space).translated(noisy)
```
adler-j: why `1 / reg_param`

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
...
+
+# define objective
+obj_fun = l1_norm + huber_l1_norm * gradient
+
+# strong convexity of "f*"
+strong_convexity = 1 / huber_l1_norm.grad_lipschitz
+
+
+# define callback to store function values
+class CallbackStore(odl.solvers.util.callback.Callback):
```
adler-j: This adds *a lot* of complexity, are we truly sure we need it? I see why we do it, but I'm afraid it will scare off new users.

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,123 @@
...
+
+# number of iterations
+niter = 500
+
+# Assign operator and functionals
+op = gradient
+f = huber_l1_norm
+g = l1_norm
+
+# The operator norm of the gradient with forward differences is well-known
```
adler-j: feel free to add this to `Gradient.norm` and instead call `op_norm = gradient.norm()` here, that would be much better!

examples/solvers/pdhg_denoising_L2_HuberTV.py
```diff
@@ -0,0 +1,165 @@
...
+space = odl.uniform_discr([0, 0], shape, shape)
+
+# Create space element of ground truth
+orig = space.element(image.copy())
+
+# Create noisy observation
+noisy = odl.phantom.white_noise(space, orig, 0.1)
+
+# Gradient operator
+gradient = odl.Gradient(space, method='forward')
```
adler-j: See earlier comment on `method='forward'` and other minor comments

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2293,82 @@ def gradient(self):
                 (1 / self.sigma) * self.functional.proximal(self.sigma))
 
 
+class HuberL1L2(Functional):
```
adler-j: Adding this partially solves #597 which should be mentioned i guess

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2293,82 @@ def gradient(self):
...
+    """The functional corresponding to the Huberized L1-norm."""
+
+    def __init__(self, space, gamma):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
```
adler-j: DiscreteLp is a subclass of FnBase so no need to mention

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2293,82 @@ def gradient(self):
...
+            Domain of the functional.
+        gamma : float
+            Smoothing parameter of Huberization. If gamma = 0, then functional
+            is non-smooth corresponds to the usual L1 norm. For gamma > 0, it
+            has a 1/gamma-Lipschitz gradient so that its convex conjugate is
+            gamma-strongly convex.
+
+        Examples
+        --------
+
```
adler-j: Remove extra blankline

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2293,82 @@ def gradient(self):
...
+        >>> import odl
+        >>> X = odl.uniform_discr([0, 0], [1, 1], [5, 5])
+        >>> x = odl.phantom.white_noise(X)
+        >>> alpha = 2
+        >>> gamma = 0
+        >>> H = alpha * odl.solvers.HuberL1L2(X, gamma)
+        >>> L1 = alpha * odl.solvers.GroupL1Norm(X, 2)
+        >>> abs(H(x) - L1(x)) < 1e-10
+        """
```
adler-j: It would be tremendous with a `Notes` section here where you give the definition of the functional

odl/solvers/functional/default_functionals.py
adler-j: Also not 100% sure about the naming, is there no way to create something like `Huber` and control the `L1L2` part via parameters?

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2293,82 @@ def gradient(self):
...
+            i = n.ufuncs.less(self.gamma)
+            n = (i * (1 / (2 * self.gamma) * n**2 + self.gamma / 2) +
+                 i.ufuncs.logical_not() * n)
+
+        return n.inner(n.space.one())
+
+    @property
+    def convex_conj(self):
+        '''The convex conjugate'''
+        return FunctionalQuadraticPerturb(GroupL1Norm(self.domain,
```
adler-j: split this on two lines for readability

examples/solvers/pdhg_denoising_L1_HuberTV.py
MatthiasJE: True, modified it similar to your suggestion.

examples/solvers/pdhg_denoising_L1_HuberTV.py
MatthiasJE: The regularization parameter changes the strong convexity of f*. Thus, it is sometimes better to have it as part of g. Here it is 1, so it doesn't matter.

examples/solvers/pdhg_denoising_L1_HuberTV.py
MatthiasJE: I see your point. This is also related with your top comment about "advanced" examples. On the one hand, this may be too advanced for some beginners, but then this is something many people need in practice which would justify it being part of the examples. I would suggest to keep it for now, but to move this example somewhere else in the future.

examples/solvers/pdhg_denoising_L1_HuberTV.py
MatthiasJE: I agree. However, the generality of the gradient operator makes this rather difficult. For forward and backward differences on some regular space we can do it. Not sure how to implement it such that only the "known" cases are covered. Maybe you can make a start with the case differentiation and I add the formulas?

odl/solvers/functional/default_functionals.py
MatthiasJE: Good point, maybe this is possible. So far, I only came across the L1L2 version. Maybe we can add a TODO note and implement this when there is need for this (which may be never?!?)?

examples/solvers/pdhg_denoising_L1_HuberTV.py
adler-j: Sounds like a reasonable solution, perhaps also write something in the top level comment that this example is "exhaustive" and "not for beginners" so to say.

odl/solvers/functional/default_functionals.py
MatthiasJE: Done, please have a look.

mehrhardt:
Anything left to do here before merging?

odl/solvers/functional/default_functionals.py
adler-j: I still feel that we need to improve the naming here before we go ahead. Is there any "reasonable" way to make this simply `Huber`?

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,112 @@ def gradient(self):
...
+            Smoothing parameter of Huberization. If ``gamma = 0``, then
+            functional is non-smooth corresponds to the usual L1 norm. For
+            ``gamma > 0``, it has a ``1/gamma``-Lipschitz gradient so that its
+            convex conjugate is ``gamma``-strongly convex.
+
+        Examples
+        --------
+        Compare HuberL1L2 and L1 for vanishing smoothing ``\\gamma=0``
+
+        >>> import odl
```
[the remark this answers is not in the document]
mehrhardt: I know. I thought we discussed this a few months ago and you preferred this so that one can copy-paste the examples into an ipython shell?

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,112 @@ def gradient(self):
...
+        >>> X = odl.uniform_discr([0, 0], [1, 1], [5, 5])
+        >>> x = odl.phantom.white_noise(X)
+        >>> alpha = 2
+        >>> gamma = 0
```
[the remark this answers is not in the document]
mehrhardt: OK, done.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,112 @@ def gradient(self):
...
+            convex conjugate is ``gamma``-strongly convex.
+
+        Examples
+        --------
+        Compare HuberL1L2 and L1 for vanishing smoothing ``\\gamma=0``
+
+        >>> import odl
+        >>> X = odl.uniform_discr([0, 0], [1, 1], [5, 5])
+        >>> x = odl.phantom.white_noise(X)
+        >>> alpha = 2
```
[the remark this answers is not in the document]
mehrhardt: OK, done.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,112 @@ def gradient(self):
...
+            functional is non-smooth corresponds to the usual L1 norm. For
+            ``gamma > 0``, it has a ``1/gamma``-Lipschitz gradient so that its
+            convex conjugate is ``gamma``-strongly convex.
+
+        Examples
+        --------
+        Compare HuberL1L2 and L1 for vanishing smoothing ``\\gamma=0``
+
+        >>> import odl
+        >>> X = odl.uniform_discr([0, 0], [1, 1], [5, 5])
```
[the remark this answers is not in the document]
mehrhardt: OK, done.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,112 @@ def gradient(self):
...
+            grad_lipschitz = 1 / self.gamma
+        else:
+            grad_lipschitz = np.inf
+
+        super(HuberL1L2, self).__init__(space=space, linear=False,
+                                        grad_lipschitz=grad_lipschitz)
+
+    def _call(self, x):
+        '''Return the HuberL1L2-norm of ``x``.'''
+        if isinstance(self.domain, ProductSpace):
```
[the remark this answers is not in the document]
mehrhardt: Good point. I don't know how to handle this. You are the expert here I guess.

mehrhardt:
I made the changes you requested. The product space issue is hard to handle I suppose. It needs to be clear what "norm" is locally defined and not for all norms does this construction make sense. Probably it is best to check for these things and to throw an error if one tries something strange?

Regarding the naming, I changed it to "Huber" as this is the only HuberNorm I know of that is being used. One could then generalize this later if need be without breaking backward compatibility by ensuring that the default parameter will refer to this version.

It would also be good to merge this version with the one of Axel #1195 to get the best of two worlds.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,219 @@ def gradient(self):
...
+        --------
+        odl.solvers.proximal_huber : `proximal factory` for the Huber
+            norm.
+        """
+        return proximal_huber(space=self.domain, gamma=self.gamma)
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional.
+
```
kohr-h: A math formula would be great.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,219 @@ def gradient(self):
...
+        n = self.local_norm(x)
+
+        if self.gamma > 0:
+            i = n.ufuncs.less(self.gamma)
+            out = i * 1 / (2 * self.gamma) * n**2
+
+            i.ufuncs.logical_not(out=i)
+            out += i * (n - self.gamma / 2)
+        else:
+            out = n
```
kohr-h: This is a matter of taste, but a good principle for local variable names is "the larger the scope, the more verbose the name". 
In practice that means that names like `i` and `n` are fine for short loops or list comprehensions. As soon as the scope spans more than, say, 5 lines of code, though, one-letter names make code less readable simply because they are harder to distinguish from numbers, e.g. `i` or `l` from `1`.

Here, `n` could easily be `norm` without making lines annoyingly long. Also, the `i` multiplication can be replaced by boolean array indexing by going through Numpy:
```python
    norm = self.local_norm(x).asarray()
    with writable_array(out) as out_arr:
        if self.gamma > 0:
            # Quadratic part
            i = (norm < gamma)
            out_arr[i] = norm[i] ** 2 / (2 * self.gamma)
            # Absolute value part
            np.logical_not(i, out=i)
            out_arr[i] = norm[i] - self.gamma / 2
```
This is slightly slower due to boolean indexing, but more memory-friendly. You can scrape off even more copies by changing `norm` in-place before assigning to `out`, at the expense of an additional (slow) boolean indexing each time.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,219 @@ def gradient(self):
...
+        return out.inner(out.space.one())
+
+    @property
+    def convex_conj(self):
+        '''The convex conjugate'''
+        n = self.local_norm_operator()
+        f = FunctionalQuadraticPerturb(n.convex_conj,
+                                       quadratic_coeff=self.gamma / 2)
+
+        f.strong_convexity = 1 / self.grad_lipschitz
```
kohr-h: See above. Assigning attributes to foreign objects is not pretty (although I'm guilty of this, see #1177, but that's an exception 😇 ). We should aim for a solution that takes this as an optional initalization argument.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1406,6 +1407,77 @@ def _call(self, x, out):
     return ProximalConvexConjKLCrossEntropy
 
 
+def proximal_huber(space, gamma):
+    """Proximal factory of the Huber norm.
+
+    Parameters
+    ----------
+    space : `FnBase`
+        Space X which is the domain of the functional F
```
kohr-h: X and F don't refer to anything. I'd go for "Domain of the functional."

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1406,6 +1407,77 @@ def _call(self, x, out):
...
+
+    See Also
+    --------
+    odl.solvers.Huber : the Huber norm functional
+
+    Notes
+    -----
+    The proximal operator is given by given by the proximal operator of
+    1/(2*gamma) * L2 norm in points that are <= gamma, and by the
+    proximal operator of the l1 norm in points that are > gamma.
```
kohr-h: Can you use backticks for things like `gamma` and formula-like stuff?

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1406,6 +1407,77 @@ def _call(self, x, out):
...
+            """
+            self.sigma = float(sigma)
+            super(ProximalHuber, self).__init__(domain=space, range=space,
+                                                linear=False)
+
+        def __local_norm(self, x):
+            if isinstance(self.domain, ProductSpace):
+                return PointwiseNorm(self.domain, 2)(x)
+            else:
+                return x.ufuncs.absolute()
```
kohr-h: As above. Used only once, so just inline the code.

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,135 @@
+"""Total variation denoising using PDHG.
```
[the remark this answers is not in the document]
mehrhardt: Well, yes and no. While the application is almost the same, the choice of parameters for the algorithms are very different. The acceleration when parts of the objective are smooth is not widely used so I would like to keep this example.

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,135 @@
...
+``\\eta_\\gamma`` denotes the Huber function defined as
+
+    .. math::
+        \\eta_\\gamma(x) =
+        \\begin{cases}
+            \\frac{1}{2 \\gamma} x^2 + \\frac{gamma}{2}
+                & \\text{if } |x| \leq \\gamma \\\\
+            |x|
+                & \\text{if } |x| > \\gamma,
+        \\end{cases}.
```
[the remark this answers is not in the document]
mehrhardt: I removed the formulas for now. How would I point to the definition?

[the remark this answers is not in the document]
kohr-h: Well, "pointing" in the sense of writing something like "For more details, see the `Huber` documentation." or so.

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,135 @@
...
+l1_norm = 1 / reg_param * odl.solvers.L1Norm(space).translated(noisy)
+
+# HuberTV-regularization
+huber = odl.solvers.Huber(gradient.range, gamma=.1)
+
+# Define objective
+obj_fun = l1_norm + huber * gradient
+
+# Strong convexity of "f*"
+strong_convexity = 1 / huber.grad_lipschitz
```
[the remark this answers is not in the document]
mehrhardt: Agreed. In particular as this example is not very introductory.

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,135 @@
...
+
+callback = (odl.solvers.CallbackPrintIteration() & CallbackStore())
+
+# Number of iterations
+niter = 500
+
+# Assign operator and functionals
+op = gradient
+f = huber
+g = l1_norm
```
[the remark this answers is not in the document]
mehrhardt: Agreed

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
kohr-h: I see your point. But given that it's a lot of extra code just for that, how about having the alternative together with the parameter choice, commented out and next to the used definition? Then users can simply switch between the two by commenting in and out a couple of lines.

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,135 @@
...
+
+tau = 1.0 / gradient.norm  # Step size for primal variable
+sigma = 1.0 / gradient.norm  # Step size for dual variable
+
+# Run algorithms 2 and 3
+x = space.zero()
+callback(x)
```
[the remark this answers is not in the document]
mehrhardt: You mean the callback before the algorithm? I would like to store values for the initialization. By default ODL does not do that (which is OK!).

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,135 @@
...
+# Show convergence rate
+def rel_fun(x):
+    x = np.array(x)
+    return (x - min(x)) / (x[0] - min(x))
+
+
+i = np.array(callback.callbacks[1].iteration_counts)
+
+plt.figure(1)
+plt.clf()
```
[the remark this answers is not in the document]
mehrhardt: With the figure number like this, it would rewrite the figure if it already exists. Thus, the need for clf. I can of course delete these.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
kohr-h: That makes total sense. To make it obvious, a short comment like that would be nice.

examples/solvers/pdhg_denoising_L2_HuberTV.py
```diff
@@ -0,0 +1,178 @@
...
+
+# The operator norm of the gradient with forward differences is well-known
+gradient.norm = np.sqrt(8) + 1e-4
+
+# regularization parameter
+reg_param = 0.1
+
+# l2 data matching
+l2_norm = (1 / (2 * reg_param) *
+           odl.solvers.L2NormSquared(space).translated(noisy))
```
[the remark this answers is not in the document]
mehrhardt: That is the beauty of this example, as the performance is independent of this. It only depends on the condition number which is invariant to scaling.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,219 @@ def gradient(self):
...
+        >>> space = odl.ProductSpace(domain, 2)
+        >>> x = odl.phantom.white_noise(space)
+        >>> huber_norm = odl.solvers.Huber(space, gamma=0)
+        >>> l1_norm = odl.solvers.GroupL1Norm(space, 2)
+        >>> abs(huber_norm(x) - l1_norm(x)) < tol
+        True
+        """
+
+        self.__gamma = float(gamma)
+        self.strong_convexity = 0
```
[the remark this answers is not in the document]
mehrhardt: True, but I could (and should!) have used it. However, I see your point and kick it out.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,219 @@ def gradient(self):
...
+        if isinstance(self.domain, ProductSpace):
+            return PointwiseNorm(self.domain, 2)(x)
+        else:
+            return x.ufuncs.absolute()
+
+    def local_norm_operator(self):
+        if isinstance(self.domain, ProductSpace):
+            return GroupL1Norm(self.domain, 2)
+        else:
+            return L1Norm(self.domain)
```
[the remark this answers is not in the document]
mehrhardt: Sure. I find it a bit puzzling that this is needed in ODL. Why would GroupL1Norm(self.domain, 2) not work for scalar functions?

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
kohr-h: Yeah, just go for the default "next available" by not giving an argument.

odl/solvers/functional/default_functionals.py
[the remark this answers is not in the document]
kohr-h: That's a good question. Without looking at the internals, I assume it will work for product spaces with 1 component. We should think about making `GroupL1Norm` work for scalar functions as well.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
mehrhardt: To me there are too many changes the user would need to make. Changing the noise model, changing L2 to L1, changing the parameters, changing the input of PDHG, changing the plotting. Maybe we can combine this example with something else that ODL wants to showcase? I don't mind too much which data fit, prior or image we use as long as we have an example for the 1/k^2 case. Maybe something with colour and vectorial TV or nuclear norms?

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,219 @@ def gradient(self):
...
+    and the Huber norm is the integral of this functional over the domain,
+    i.e., the Huber norm is given by
+
+    .. math::
+        \\int_\Omega f_\\gamma(||x||_2) dx.
+
+    In the discrete case, this becomes
+
+    .. math::
+        \\sum_{i=1}^n f_\\gamma(||x_i||_2).
```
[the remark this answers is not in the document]
mehrhardt: Agreed, done.

odl/solvers/functional/default_functionals.py
mehrhardt: I agree with the variable names. When you write this code for yourself I guess one is just too lazy :)
 
About the computations, I am not so sure. I would like to keep the code numpy independent so that you can also use it on the gpu which I believe should be the case at the moment.

odl/solvers/nonsmooth/proximal_operators.py
mehrhardt: Too much copy-pasting :)

odl/solvers/functional/default_functionals.py
kohr-h: Fair enough. Just be aware that most of these ufuncs are not implemented in the CUDA backend and fall back to "copy to CPU, then use Numpy, then copy back to GPU". So you will probably end up being slower AND use more memory.

That said, Jonas is doing the (I hope) final round of review of #1088 which will make all this work much nicer.

odl/solvers/nonsmooth/proximal_operators.py
kohr-h: Yeah we sure have some cleanup to do in the older parts.

mehrhardt:
All implemented. The only thing that is left is the actual computation of the function call, gradient, prox etc where I would like not to involve numpy. In the end @adler-j, @kohr-h you are the experts how to implement this most efficiently.

mehrhardt:
OK, this looks weird. In the log of the files, these are not "new" commits.

What I did was to rebase the current branch onto upstream/master as

git fetch upstream
git rebase upstream/master

It then complained about the branches having diverged. I force-pushed my local version into the repo which "looks" good but github can't handle it (see fake commits above). What is the proper way to handle this situation?

kohr-h:
> It then complained about the branches having diverged. I force-pushed my local version into the repo which "looks" good but github can't handle it (see fake commits above). What is the proper way to handle this situation?

Which ones do you mean? Isn't 23fe237 the first commit in this PR?

mehrhardt:
It is, but my github reminds me again that I "added some commits 9 days ago". Is that different for you?

kohr-h:
> It is, but my github reminds me again that I "added some commits 9 days ago". Is that different for you?

Ah, I think I know what you mean. When you rebase, GitHub adds the commits as new ones below the last comment. If you reload the page, the old ones go away.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
kohr-h: Another possibility is to make a big
```python
if data_term == 'L1':
    # Do the L1 case, including plotting and whatnot
elif data_term == 'L2':
    # Other case
```
Since it's a rather elaborate example anyway, making one large example from two slightly-less-than-large would still be good IMO.
What's your opinion @adler-j?

examples/solvers/pdhg_denoising_L2_HuberTV.py
```diff
@@ -0,0 +1,139 @@
...
+image = np.rot90(scipy.misc.ascent()[::2, ::2].astype('float'), 3)
+shape = image.shape
+image /= image.max()
+space = odl.uniform_discr([0, 0], shape, shape)
+orig = space.element(image.copy())
+d = odl.phantom.white_noise(space, orig, 0.1)
+
+# Define objective functional
+op = odl.Gradient(space)  # operator
+op.norm = np.sqrt(8) + 1e-4  # norm with forward differences is well-known
```
kohr-h: We shouldn't be setting attributes on instances like this. In fact, this overrides the *method* `Operator.norm` with a constant.
Until `Gradient.norm` is implemented please make this a free variable instead. Do we have an open issue on that somewhere? I can't find one.

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,216 @@ def gradient(self):
...
+
+    def __init__(self, space, gamma):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `FnBase`
+            Domain of the functional.
+        gamma : float
+            Smoothing parameter of Huberization. If ``gamma = 0``, then
```
kohr-h: the

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,216 @@ def gradient(self):
...
+            functional is non-smooth corresponds to the usual L1 norm. For
```
kohr-h: and

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,216 @@ def gradient(self):
...
+            ``gamma > 0``, it has a ``1/gamma``-Lipschitz gradient so that
+            its convex conjugate is ``gamma``-strongly convex.
```
kohr-h: You surely prepare for this addition 😛

odl/solvers/functional/default_functionals.py
```diff
@@ -2292,6 +2294,216 @@ def gradient(self):
...
+
+        Compare Huber- and L1-norm for vanishing smoothing ``gamma=0``.
+
+        >>> x = odl.phantom.white_noise(space)
+        >>> huber_norm = odl.solvers.Huber(space, gamma=0)
+        >>> l1_norm = odl.solvers.L1Norm(space)
+        >>> abs(huber_norm(x) - l1_norm(x)) < tol
+        True
+
+        Redo previous example for a product space in two dimensions.
```
kohr-h: Colons at the end of those intermediate texts.

mehrhardt: done

odl/solvers/functional/default_functionals.py
mehrhardt: done

examples/solvers/pdhg_denoising_L2_HuberTV.py
mehrhardt: OK, changed.

odl/solvers/functional/default_functionals.py
mehrhardt: done

examples/solvers/pdhg_denoising_L2_HuberTV.py
```diff
@@ -0,0 +1,139 @@
+"""Linearly convergent total variation denoising using PDHG.
+
+This exhaustive example solves the L2-HuberTV problem
+
+        min_{x >= 0}  1/2 ||x - d||_2^2
+            + lam * sum_i eta_gamma(||grad(x)_i||_2)
+
+where grad the spatial gradient and d is given noisy data. Here eta_gamma
```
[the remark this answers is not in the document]
mehrhardt: done

examples/solvers/pdhg_denoising_L1_HuberTV.py
```diff
@@ -0,0 +1,97 @@
...
+        self.iteration_counts.append(self.iteration_count)
+        self.obj_function_values.append(obj_fun(x))
+
+    def reset(self):
+        self.iteration_count = 0
+        self.iteration_counts = []
+        self.obj_function_values = []
+
+
+callback = (odl.solvers.CallbackPrintIteration(step=10) & CallbackStore())
```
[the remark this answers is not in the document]
mehrhardt: I am OK with removing these but the two of you should settle on this. A while ago @adler-j told me to put parentheses for readability somewhere.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
mehrhardt: At first glance, I really like your idea as it reduces the code we have. When I thought about how to implement it, I reversed my opinion as this makes the example code even longer and more complicated. 

I think I am still in favour of keeping the two examples but to change one of them to show case more of ODLs tools. We could also rename them to "pdhg_linear_convergence_rate" and "pdhg_primal_1k2_rate" to highlight why both of them are actually there. In any case, I would like to settle this sooner than later. It keeps dragging on and on.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
kohr-h: For stuff on one line? I'd be surprised. It makes sense when you have stuff that spans 2 or more lines.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
kohr-h: I don't mind having both examples as such. My worries are mainly that the `examples/solvers` folder becomes really crowded and hard to navigate. We have a top-level README to help (which reminds me that these examples should be mentioned there 😄), but still.

A way of organizing things better would be to keep the "basic" examples where they are now, in `examples/solvers`, and to create a subfolder `advanced` or so for stuff like this.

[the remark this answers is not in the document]
mehrhardt: I would like an advanced examples folder. However, I feel that this discussion has been there a few times. Thus, I propose to merge things the way they are and to open a new issue with this.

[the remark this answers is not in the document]
kohr-h: Could be. You're right, let's not keep this from going in any longer.

The fact that this seems to be a recurring discussion indicates that it lacks a place for discussion. Would you mind opening an issue on that, i.e., finding a place for advanced examples?

kohr-h:
I made a small PR due to some things I noticed when doing the final skim. When that's in (you can also pick only the stuff that you like) I'll hit the merge button.

examples/solvers/pdhg_denoising_L1_HuberTV.py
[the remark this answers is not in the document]
mehrhardt: Done, see #1209

mehrhardt:
I liked all your changes @kohr-h so I merged it without cherry picking.

kohr-h:
Thanks for the work @mehrhardt!

PR #1192, opened 2017-10-12
ENH: Add support for complex vectors in adjoint of operator-vector mult.

zickert:


odl/operator/operator.py
```diff
@@ -2055,12 +2055,19 @@ def derivative(self, x):
...
 
-        The adjoint of the operator vector multiplication is the
-        vector multiplication of the operator adjoint:
+        For real spaces, the adjoint of the operator left vector multiplication
+        is the right multiplication of the given operator by the given vector.
 
             ``OperatorLeftVectorMult(op, y).adjoint ==
             OperatorRightVectorMult(op.adjoint, y)``
 
+        For complex spaces, the adjoint of the operator left vector
```
adler-j: Not quite sure we need two different comments here, this reduces to the above case. The only reason we have an if statement is that conj() produces a copy, which is bad for performance.

I'd try to "merge" these so to say, only leaving the code for the complex case but perhaps mentioning that the conj does nothing for real vectors.

zickert:
I can add test for this. (But not sure if I will find time to do it before the week after the next.)

odl/operator/operator.py
```diff
@@ -2073,8 +2074,11 @@ def adjoint(self):
         if not self.is_linear:
             raise OpNotImplementedError('nonlinear operators have no adjoint')
 
-        # TODO: handle complex vectors
-        return self.operator.adjoint * self.vector
+        if self.vector.space.is_rn:
+            '''The complex conjugate of a real vector is the vector itself.'''
```
adler-j: Always use # for inline comments. 

What happens here is technically that you create a string with the text, and then never assign it to anything. This is a slight performance problem.

Comments "at the top" are special since we actually *want* to assign them to something, but that is done implicitly by python. You can check `obj.__doc__` to see where it was assigned, and find more information here https://www.python.org/dev/peps/pep-0257.

Anyway, fix that and then we go!

adler-j:
Test can come later, but would indeed be a nice addition. Overall *all* of these operator utilities are very low in tests, and that should be improved.

Issue #1197, opened 2017-10-18
Solvers API slightly inconsistent


kohr-h:
Of the convex solvers we have so far, most have the structure that `f` is not composed, and the compositions with operators happen in the `g` functional(s). Only PDHG does it the other way around.

Should we change it?

I'm asking since I'm gonna add linearized ADMM which solves the same problems as PDHG, and I want this to be settled before.

kohr-h:
Heads-up to @mehrhardt and @adler-j for comments.

aringh:
This falls down to what I always ask: consistency with the paper or not? :-p As long as the documentation is clear I don't think it really matters.

adler-j:
My suggestion is either rename to `primal_functional`, which would solve the naming, or leave names as is and simply keep the argument order the same.

mehrhardt:
I think it would be good if our algorithms are consistent. That is more important than the consistency with the papers. To me, it is more natural that f is composed with an operator as the data term often comes first and f < g in a lexiographic sense. I suppose some of you feel the other way :)

kohr-h:
> My suggestion is either rename to primal_functional, which would solve the naming, or leave names as is and simply keep the argument order the same.

Not so sure about that. For some algorithms that are not really of primal-dual type (like ADMM?) the names "primal" and "dual" would be quite a stretch.

> I think it would be good if our algorithms are consistent. That is more important than the consistency with the papers. 

I fully agree with this. In papers, authors use whatever they like best, and if we mirror every paper notation we end up with a hopeless mess. 

There's also a practical side to it: when you want to switch between two solvers that handle exactly the same type of problem (like PDHG and ADMM), you shouldn't have to switch the roles of the functionals every time. The same holds when using Forward-Backward or Douglas-Rachford with just one `g` functional.

> To me, it is more natural that f is composed with an operator as the data term often comes first and f < g in a lexiographic sense. I suppose some of you feel the other way :)

This point is perhaps more a matter of taste and practicality. Going with `f` composed and `g` not would mean that we need to change everything except PDHG, the other possibility means only changing PDHG.

adler-j:
Did we arrive at a decision here, I'm currently doing this at @kohr-h's suggestion, but I'm not quite sure what to do. For sure, only changing pdhg is the easier route.

mehrhardt:
[...]

Issue #1181, opened 2017-10-09
Bug in prox of l1 norm


mehrhardt:
This example shows a bug in the prox operator of the l1 norm:
```
import odl

X = odl.rn(1)
f = odl.solvers.L1Norm(X)
x = 0.1 * X.one()

y1 = x.copy()
f.convex_conj.proximal(3)(y1, out=y1)

y2 = x.copy()
y2 = f.convex_conj.proximal(1)(y2)

(y1-y2).norm()
```
results in 0.9. Similar effects can be observed in `KullbackLeibler` but not in `L2Norm`.

adler-j:
What would be the expected result and why?

MatthiasJE:
As mentioned above already, replacing f by `f = odl.solvers.KullbackLeibler(X)` yields similar issues. Who knows which other functionals have the same problem.

adler-j:
Darn it, you are right. I'll look into it.

mehrhardt:
Any news here? Is there a way to write a test that gets run on all functionals (without us writing a new one every time)?

adler-j:
No update sadly, havn't had time. With that said there is a rather enormous large-scale test here:

https://github.com/odlgroup/odl/blob/baecb8ada64995b7437cd21590f5a005f5695e9e/odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py#L114

it should catch stuff like this, but apparently does not.

kohr-h:
> No update sadly, havn't had time. With that said there is a rather enormous large-scale test here:
> 
> odl/odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
> 
> Line 114 in baecb8a
>  def test_proximal_defintion(functional, stepsize): 
> 
> it should catch stuff like this, but apparently does not.

Does it test with aligned input & output?

adler-j:
> Does it test with aligned input & output?

Very much looks like it does not.

mehrhardt:
Well done! This indeed fixed the bug I was observing.

mehrhardt:
I just did a similar test with the l1-norm itself and not its conjugate. While the latter is indeed fixed, the former still has the same problem:

```
import odl

X = odl.rn(1)
f = odl.solvers.L1Norm(X)
x = 0.1 * X.one()

y1 = x.copy()
f.proximal(1e-2)(y1, out=y1)

y2 = x.copy()
y2 = f.proximal(1e-2)(y2)

(y1-y2).norm()
```

PR #1198, opened 2017-10-18
Add ADMM

kohr-h:
Depending on the outcome of #1197, `f` and `g` may change roles.

### DIFF
--- a/examples/solvers/admm_tomography.py
+++ b/examples/solvers/admm_tomography.py
@@ -1,11 +1,11 @@
 """Total variation tomography using linearized ADMM.
 
-Solves the optimization problem
+In this example we solve the optimization problem
 
     min_x  ||A(x) - y||_2^2 + lam * ||grad(x)||_1
 
 Where ``A`` is a parallel beam ray transform, ``grad`` the spatial
-gradient and ``y`` is given noisy data.
+gradient and ``y`` given noisy data.
 
 The problem is rewritten in decoupled form as
 
@@ -18,7 +18,7 @@ with a separable sum ``g`` of functionals and the stacked operator ``L``:
                ( A(x)    )
     z = L(x) = ( grad(x) ).
 
-See the documentation of the `admm_linearized` method for further details.
+See the documentation of the `admm_linearized` solver for further details.
 """
 
 import numpy as np
--- a/examples/solvers/admm_tomography.py
+++ b/examples/solvers/admm_tomography.py
@@ -71,8 +71,8 @@ sigma = 2.0  # Step size for the dual variable
 tau = sigma / op_norm ** 2  # Step size for the primal variable
 
 # Optionally pass callback to the solver to display intermediate results
-callback = (odl.solvers.CallbackPrintIteration() &
-            odl.solvers.CallbackShow())
+callback = (odl.solvers.CallbackPrintIteration(step=10) &
+            odl.solvers.CallbackShow(step=10))
 
 # Choose a starting point
 x = op.domain.zero()
--- a/examples/solvers/admm_tomography.py
+++ b/examples/solvers/admm_tomography.py
@@ -1,13 +1,24 @@
-"""Total variation tomography using ADMM.
+"""Total variation tomography using linearized ADMM.
 
 Solves the optimization problem
 
-    min_x  1/2 ||A(x) - g||_2^2 + lam || |grad(x)| ||_1
+    min_x  ||A(x) - y||_2^2 + lam * ||grad(x)||_1
 
 Where ``A`` is a parallel beam ray transform, ``grad`` the spatial
-gradient and ``g`` is given noisy data.
+gradient and ``y`` is given noisy data.
 
-See the documentation of the `admm` method for further details.
+The problem is rewritten in decoupled form as
+
+    min_x g(L(x))
+
+with a separable sum ``g`` of functionals and the stacked operator ``L``:
+
+    g(z) = ||z_1 - g||_2^2 + lam * ||z_2||_1,
+
+               ( A(x)    )
+    z = L(x) = ( grad(x) ).
+
+See the documentation of the `admm_linearized` method for further details.
 """
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


odl/solvers/nonsmooth/admm.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Parameters
+    ----------
+    x : ``L.domain`` element
+        Starting point of the iteration, updated in-place.
+    f, g : `Functional`
+        The functions ``f`` and ``g`` in the problem definition. They
+        need to implement the ``proximal`` method.
+    L : linear `Operator`
+        The linear operator that should be applied before ``f``. Its range
```
kohr-h: f -> g

odl/solvers/nonsmooth/admm.py
```diff
@@ -0,0 +1,143 @@
...
+    if not isinstance(niter, int) or niter < 0:
+        raise ValueError('`niter` {} not understood'
+                         ''.format(niter))
+
+    # Callback object
+    callback = kwargs.pop('callback', None)
+    if callback is not None and not callable(callback):
+        raise TypeError('`callback` {} is not callable'.format(callback))
+
+    # Initialize dual variables
```
kohr-h: Dual is probably the wrong name since we never look at convex conjugates, Just `u` and `z` should do.

examples/solvers/admm_tomography.py
```diff
@@ -0,0 +1,87 @@
...
+# l2-squared data matching
+l2_norm = odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+
+# Isotropic TV-regularization i.e. the l1-norm
+l1_norm = 0.015 * odl.solvers.L1Norm(gradient.range)
+
+# Combine functionals, order must correspond to the operator K
+g = odl.solvers.SeparableSum(l2_norm, l1_norm)
+
+# --- Select solver parameters and solve using PDHG --- #
```
kohr-h: update

odl/solvers/nonsmooth/admm.py
```diff
@@ -0,0 +1,143 @@
...
+        # L^*(L x^k + u^k - z^k)
+        L.adjoint(tmp_ran, out=tmp_dom)
+
+        # x^k - (tau/sigma) L^*(Lx^k + u^k - z^k)
+        x.lincomb(1, x, -tau / sigma, tmp_dom)
+        # x^(k+1) = prox[tau*f](_)
+        f.proximal(tau)(x, out=x)
+
+        # L x^(k+1) + u^k
+        L(x, out=tmp_ran)
```
mehrhardt: What is the difference between linearized ADMM and PDHG? In any case, I am pretty sure it is possible to rewrite your code calling "L" only once per iteration.

odl/solvers/nonsmooth/admm.py
```diff
@@ -0,0 +1,143 @@
...
+        tmp_ran += dual_u
+        tmp_ran -= dual_z
+        # L^*(L x^k + u^k - z^k)
+        L.adjoint(tmp_ran, out=tmp_dom)
+
+        # x^k - (tau/sigma) L^*(Lx^k + u^k - z^k)
+        x.lincomb(1, x, -tau / sigma, tmp_dom)
+        # x^(k+1) = prox[tau*f](_)
+        f.proximal(tau)(x, out=x)
```
mehrhardt: I like this line as it is very elegant but some of ODL's functionals do not support this at the moment, see #1181.

odl/solvers/nonsmooth/admm.py
kohr-h: The main difference is that ADMM uses the proximal of `f` in contrast to the proximal of `g^*` as in PDHG, so it's not a primal-dual method in that sense. But it's possible that one can still interpret it that way, dunno.

Anyway, you're right, I should be able to scrap one `L` evaluation, potentially at the cost of a new temporary.

odl/solvers/nonsmooth/admm.py
kohr-h: True, but this is not the place to implement workarounds. This code has to be able to assume that operators work with aliased input and output.

examples/solvers/admm_tomography.py
```diff
@@ -0,0 +1,87 @@
+"""Total variation tomography using ADMM.
```
adler-j: linearized?

examples/solvers/admm_tomography.py
```diff
@@ -0,0 +1,87 @@
...
+
+Solves the optimization problem
+
+    min_x  1/2 ||A(x) - g||_2^2 + lam || |grad(x)| ||_1
+
+Where ``A`` is a parallel beam ray transform, ``grad`` the spatial
+gradient and ``g`` is given noisy data.
+
+See the documentation of the `admm` method for further details.
```
adler-j: admm_linearized?

examples/solvers/admm_tomography.py
```diff
@@ -0,0 +1,87 @@
...
+# Isotropic TV-regularization i.e. the l1-norm
+l1_norm = 0.015 * odl.solvers.L1Norm(gradient.range)
+
+# Combine functionals, order must correspond to the operator K
+g = odl.solvers.SeparableSum(l2_norm, l1_norm)
+
+# --- Select solver parameters and solve using PDHG --- #
+
+# Estimated operator norm, add 10 percent to ensure ||K||_2^2 * sigma * tau < 1
+op_norm = 1.1 * odl.power_method_opnorm(op)
```
adler-j: Overall we should move towards `op.norm()` in these cases now that it's implemented.

odl/solvers/nonsmooth/admm.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # x <- x^k - (tau/sigma) L^*(Lx^k + u^k - z^k)
+        x.lincomb(1, x, -tau / sigma, tmp_dom)
+        # x^(k+1) <- prox[tau*f](x)
+        f.proximal(tau)(x, out=x)
+
+        # tmp_ran <- L x^(k+1)
+        L(x, out=tmp_ran)
+        # z^(k+1) <- prox[sigma*g](L x^(k+1) + u^k)
+        g.proximal(sigma)(tmp_ran + u, out=z)  # 1 copy here
```
adler-j: why not `tmp_ran += u` on the line above?

odl/test/solvers/nonsmooth/admm_test.py
```diff
@@ -0,0 +1,77 @@
...
+    x0 = noise_element(space)
+    x = x0.copy()
+    niter = 3
+
+    admm_linearized(x, f, g, L, tau=1.0, sigma=1.0, niter=niter)
+
+    assert x == x0
+
+    # Check that a provided callback is actually called
+    class CallbackTest(Callback):
```
adler-j:
```python
class CallbackTest(Callback):
    was_called = False
    def __call__(self, *args, **kwargs):
        self.was_called = True
```

odl/solvers/nonsmooth/admm.py
kohr-h: I need `L(x)` later to add the new `u`. The alternative would be to use another temporary, but that seems overkill to me.

examples/solvers/admm_tomography.py
kohr-h: Don't blame me for all the above, it's literally copied from the PDHG example (except for the admm typos).

adler-j: I guess we could make a "mother of all PRs" to fix that

odl/solvers/nonsmooth/admm.py
adler-j: Go with this for now

kohr-h:
Done fixing, will merge after CI

kohr-h:
In particular, I'll keep the roles of `f` and `g` as they are now, and we can decide simultaneously what we're gonna do regarding #1197.

PR #1202, opened 2017-10-21
ENH: add nonlinear version of ADMM, closes #1201

### LINKED ISSUE #1201: Nonlinear ADMM
As pointed out by @ozanoktem there's a nonlinear version of ADMM as detailed in the paper [Preconditioned ADMM with Nonlinear Operator Constraint](https://doi.org/10.1007/978-3-319-55795-3_10). The method is formulated to solve
```
min_(x, y) [ f(x) + g(y) ] s.t. F(x, y) = 0
```
Since we don't have solvers of that form my suggestion is that we go with the less general variant where `F(x, y) = L(x) - y`, i.e., where one can solve for `y` in the nonlinear constraint and rewrite the problem as
```
min_x [ f(x) + g(L(x)) ]
```
more in line with what we already have.

kohr-h:
**TODO:**
- [x] Implement algorithm
- [ ] Unit test
- [ ] Example


kohr-h:
Ready for review. I also found the bug that was causing headaches for aliased input and output, see #1181. It was in `OperatorSum`! Incredible that this hasn't turned up earlier.

examples/solvers/admm_spectral_tomography.py
```diff
@@ -0,0 +1,169 @@
+"""Total variation spectral tomography using preconditioned nonlinear ADMM.
```
adler-j: Examples like this truly need to go into a "complicated" folder so to say, this is getting out of hand.

Nothing against the example per se, just that we want beginners to find them "in the right order".

examples/solvers/admm_spectral_tomography.py
```diff
@@ -0,0 +1,169 @@
...
+
+# Define number of energy bins on both sides as well as the source spectrum
+num_bins_reco = 4
+num_bins_data = 2
+assert num_bins_reco >= num_bins_data
+assert num_bins_reco % num_bins_data == 0
+group_size = num_bins_reco // num_bins_data
+# Note: the spectrum (i.e. intensity per channel) is additive on the data
+# side, i.e., adding more entries adds to the total signal in the data bins
+source_spectrum = [8e4, 1e5, 8e4, 6e4]
```
adler-j: why soo high values?

odl/solvers/nonsmooth/admm.py
```diff
@@ -159,3 +170,212 @@ def admm_linearized_simple(x, f, g, L, tau, sigma, niter, **kwargs):
         u = L(x) + u - z
         if callback is not None:
             callback(x)
+
+
+def admm_precon_nonlinear(x, f, g, L, delta, niter, sigma=None, **kwargs):
```
adler-j: Not sure that "precon" should have its own function, IMO it should just be optional and this can be called "admm_nonlinear"

adler-j: Also, are we sure that this is *the* admm_nonlinear method, or can we get a naming collision here in the future?

odl/solvers/nonsmooth/admm.py
```diff
@@ -159,3 +170,212 @@ def admm_linearized_simple(x, f, g, L, tau, sigma, niter, **kwargs):
...
+                         ''.format(niter_in))
+
+    if sigma is None:
+        sigma = 0.5 / delta
+    else:
+        sigma, sigma_in = float(sigma), sigma
+        if not 0 < sigma < 1 / delta:
+            raise ValueError(
+                '`sigma` must lie strictly between 0 and 1/delta`, got '
+                'sigma={:.4} and (1/delta)={:.4}'.format(sigma_in, 1 / delta))
```
adler-j: 4 digits is bad practice for errors, just print all of them.

odl/solvers/nonsmooth/admm.py
```diff
@@ -159,3 +170,212 @@ def admm_linearized_simple(x, f, g, L, tau, sigma, niter, **kwargs):
...
+
+    callback = kwargs.pop('callback', None)
+    if callback is not None and not callable(callback):
+        raise TypeError('`callback` {} is not callable'.format(callback))
+
+    # Initialize range variables (quantities in [] are zero initially)
+    y = L.range.zero()
+    # mu = [mu_old +] delta * (L(x) [- y])
+    mu = delta * L(x)
+    # mubar = 2 * mu [- mu_old]
```
adler-j: leftover?

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -910,9 +904,34 @@ def proximal_l2_squared(space, lam=1, g=None):
...
+
+        def _call(self, x, out):
+            """Apply the operator to ``x`` and store the result in ``out``"""
+            # (x + 2*sig*lam*g) / (1 + 2*sig*lam))
+            sig = self.sigma
+            if g is None:
+                out.lincomb(1.0 / (1 + 2 * sig * lam), x)
+            else:
+                out.lincomb(1.0 / (1 + 2 * sig * lam), x,
+                            2 * sig * lam / (1 + 2 * sig * lam), g)
```
adler-j: mixing 1.0 and 1 here, lets go for one of them

odl/test/solvers/nonsmooth/admm_test.py
```diff
@@ -72,5 +74,65 @@ def test_admm_lin_l1():
     assert all_almost_equal(x, data_1, places=2)
 
 
+def test_admm_lin_vs_simple():
```
adler-j: I guess these are old, or do we intend for them to stay? If so, I'd move the "simple" guy into this file

odl/solvers/nonsmooth/admm.py
kohr-h: As I understand it, the preconditioning is part of the method, and the way it is used makes the method explicit, i.e., everything can be expressed in terms of proximals. *The* ADMM method has implicit steps in the form of some `argmin` which is not a proximal.

examples/solvers/admm_spectral_tomography.py
kohr-h: I'll have to work a bit on this example anyway. Currently it's a bit silly since for `n x n` problems, you can actually linearize.

PR #1210, opened 2017-10-26
Minor bugfixes

adler-j:
I had about 100 test failures from master due to this, minor fixes.

Anyway, how did `np.broadcast.ndim` ever work? It doesnt seem to exist:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast.html

### DIFF
--- a/odl/tomo/backends/astra_cuda.py
+++ b/odl/tomo/backends/astra_cuda.py
@@ -36,7 +36,12 @@ __all__ = ('ASTRA_CUDA_AVAILABLE',
 class AstraCudaProjectorImpl(object):
 
     """Thin wrapper around ASTRA."""
-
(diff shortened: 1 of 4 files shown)


kohr-h:
> Anyway, how did np.broadcast.ndim ever work? It doesnt seem to exist:

For me it exists:
```python
>>> a = np.ones((3, 1))
>>> b = np.ones((1, 2))
>>> bcast = np.broadcast(a, b)
>>> bcast.ndim
2
```
It seems to be undocumented, though. The documentation mentions an `nd` attribute which is the name for the number of dimensions of an `nditer` object I think. So we could use that instead to be on the safe side.
```python
>>> bcast.nd
2
```

PR #1211, opened 2017-10-26
Add images to odl

mehrhardt:
Add the four images I used for the SPDHG examples to ODL.

### DIFF
--- a/odl/contrib/datasets/README.md
+++ b/odl/contrib/datasets/README.md
@@ -21,3 +21,14 @@ Reference datasets with accompanying ODL geometries etc.
     CT data as provided by FIPS. The data is non-human and high resolution.
     * `walnut_data`
     * `lotus_root_data`
+* `images`
+
+  Two dimensional images.
+  * `cambridge`
+
+    Various images as provided by the University of Cambridge, see http://store.maths.cam.ac.uk/DAMTP/me404/data_sets/
+    * `brain_phantom`
+    * `resolution_phantom`
+    * `building`
+    * `rings`
+    * `blurring_kernel`
--- a/odl/contrib/datasets/images/__init__.py
+++ b/odl/contrib/datasets/images/__init__.py
@@ -1,0 +1,16 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Access to a set of test images."""
+
+from __future__ import absolute_import
+
+__all__ = ()
+
+from .cambridge import *
+__all__ += cambridge.__all__
--- a/odl/contrib/datasets/images/cambridge.py
+++ b/odl/contrib/datasets/images/cambridge.py
@@ -20,7 +20,7 @@ DATA_SUBSET = 'images_cambridge'
 URL_CAM = 'http://store.maths.cam.ac.uk/DAMTP/me404/data_sets/'
 
 
-def convert(image, shape, gray=None, dtype='float64', normalize='max'):
+def convert(image, shape, gray=False, dtype='float64', normalize='max'):
     """Convert image to standardized format.
 
     Several properties of the input image may be changed including the shape,
@@ -30,17 +30,20 @@ def convert(image, shape, gray=None, dtype='float64', normalize='max'):
 
     image.astype(dtype)
 
-    if gray is not None and gray:
-        image = np.mean(image, 2)
+    if gray:
+        image[..., 0] *= 0.2126
+        image[..., 1] *= 0.7152
+        image[..., 2] *= 0.0722
+        image = np.sum(image, axis=2)
 
     if shape is not None:
         image = imresize(image, shape)
 
     if normalize == 'max':
-        image = image / image.max()
+        image /= image.max()
 
     elif normalize == 'sum':
-        image = image / image.sum()
+        image /= image.sum()
 
     return image
 
@@ -48,7 +51,9 @@ def convert(image, shape, gray=None, dtype='float64', normalize='max'):
 def brain_phantom(shape=None):
     """Brain phantom for FDG PET simulations.
 
-    Output:
+    Returns
+    -------
+    An image with the following properties:
         image type: gray scales
         shape: [1024, 1024] (if not specified by `size`)
         scale: [0, 1]
@@ -66,7 +71,9 @@ def brain_phantom(shape=None):
 def resolution_phantom(shape=None):
     """Resolution phantom for tomographic simulations.
 
-    Output:
+    Returns
+    -------
+    An image with the following properties:
         image type: gray scales
         shape: [1024, 1024] (if not specified by `size`)
         scale: [0, 1]
@@ -85,7 +92,9 @@ def resolution_phantom(shape=None):
 def building(shape=None, gray=False):
     """Photo of the Centre for Mathematical Sciences in Cambridge.
 
-    Output:
+    Returns
+    -------
+    An image with the following properties:
         image type: color (or gray scales if `gray=True`)
         size: [442, 331] (if not specified by `size`)
         scale: [0, 1]
@@ -103,7 +112,9 @@ def building(shape=None, gray=False):
 def rings(shape=None, gray=False):
     """Photo of married couple holding hands.
 
-    Output:
+    Returns
+    -------
+    An image with the following properties:
         image type: color (or gray scales if `gray=True`)
         size: [3264, 2448] (if not specified by `size`)
         scale: [0, 1]
@@ -123,7 +134,9 @@ def blurring_kernel(shape=None):
 
     The kernel is scaled to sum to one.
 
-    Output:
+    Returns
+    -------
+    An image with the following properties:
         image type: gray scales
         size: [100, 100] (if not specified by `size`)
         scale: [0, 1]
--- a/odl/contrib/datasets/images/cambridge.py
+++ b/odl/contrib/datasets/images/cambridge.py
@@ -1,0 +1,160 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Images provided by the University of Cambridge."""
+
+import numpy as np
+import odl
+from odl.contrib.datasets.util import get_data
+from scipy.misc import imresize
+
+__all__ = ('brain_phantom', 'resolution_phantom', 'building', 'rings',
+           'blurring_kernel')
+
+
+DATA_SUBSET = 'images_cambridge'
+URL_CAM = 'http://store.maths.cam.ac.uk/DAMTP/me404/data_sets/'
+
+
+def convert(image, shape, to_odl, gray=None, dtype='float64', normalize='max'):
+    """Convert image an ODL object in standardized format.
+
+    In addition several properties of the input image may be changed including
+    the shape, data type and maximal value of the image. In addition, this
+    function may convert the image into an ODL object and/or a gray scale
+    image.
+    """
+
+    image.astype(dtype)
+
+    if gray is not None and gray:
+        image = np.mean(image, 2)
+
+    if shape is not None:
+        image = imresize(image, shape)
+
+    if normalize == 'max':
+        image = image / image.max()
+
+    elif normalize == 'sum':
+        image = image / image.sum()
+
+    if to_odl:
+        shape = image.shape
+
+        if len(shape) == 2:
+            space = odl.uniform_discr([0, 0], shape, shape)
+        elif len(shape) == 3:
+            d = shape[2]
+            shape = shape[:2]
+            image = np.transpose(image, (2, 0, 1))
+            space = odl.uniform_discr([0, 0], shape, shape) ** d
+
+        image = space.element(image)
+
+    return image
+
+
+def brain_phantom(shape=None, to_odl=True):
+    """Brain phantom for FDG PET simulations.
+
+    Output:
+        image type: gray scales
+        shape: [1024, 1024] (if not specified by `size`)
+        scale: [0, 1]
+        type: float64
+    """
+    # TODO: Store data in some ODL controlled url
+    name = 'PET_phantom.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset=DATA_SUBSET, url=url)
+    im = np.rot90(dct['im'], k=3)
+
+    return convert(im, shape, to_odl)
+
+
+def resolution_phantom(shape=None, to_odl=True):
+    """Resolution phantom for tomographic simulations.
+
+    Output:
+        image type: gray scales
+        shape: [1024, 1024] (if not specified by `size`)
+        scale: [0, 1]
+        type: float64
+    """
+    # TODO: Store data in some ODL controlled url
+    # TODO: This can be also done with ODL's ellipse_phantom
+    name = 'phantom_resolution.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset=DATA_SUBSET, url=url)
+    im = np.rot90(dct['im'], k=3)
+
+    return convert(im, shape, to_odl)
+
+
+def building(shape=None, gray=False, to_odl=True):
+    """Photo of the Centre for Mathematical Sciences in Cambridge.
+
+    Output:
+        image type: color (or gray scales if `gray=True`)
+        size: [442, 331] (if not specified by `size`)
+        scale: [0, 1]
+        type: float64
+    """
+    # TODO: Store data in some ODL controlled url
+    name = 'cms.mat'
+    url = URL_CAM + name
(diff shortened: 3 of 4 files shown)


odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,111 @@
...
+
+__all__ = ('brain_phantom', 'resolution_phantom', 'cms', 'cms_gray', 'rings',
+           'blurring_kernel')
+
+
+DATA_SUBSET = 'images'
+URL_CAM = 'http://store.maths.cam.ac.uk/DAMTP/me404/data_sets/'
+
+
+def brain_phantom():
```
adler-j: Would it be possible to add a flag to this to return this phantom in different forms, e.g.

- `'materials'` Materials by index (e.g. 0, 1, 2, 3).
- `'PET'` The pet phantom you have right now I guess
- `'density'`/`'CT'` Corresponding HU/density values
- `'MRI'` Fake proton densities

That would make this much more applicable!

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,111 @@
...
+    name = 'PET_phantom.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset='images', url=url)
+    im = np.rot90(dct['im'].astype('float64'), k=3)
+    im /= im.max()
+
+    return im
+
+
+def resolution_phantom():
```
adler-j: This could actually be implemented using the `ellipse_phantom` stuff in ODL, which should give a variable resolution. I'd personally prefer doing ti that way.

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,111 @@
...
+    name = 'cms.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset='images', url=url)
+    im = np.rot90(dct['im'].astype('float64'), k=3)
+    im /= im.max()
+
+    return im
+
+
+def cms_gray():
```
adler-j: I would prefer adding a `gray=False` flag to `cms like in

https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.misc.face.html

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,111 @@
...
+    name = 'phantom_resolution.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset='images', url=url)
+    im = np.rot90(dct['im'].astype('float64'), k=3)
+    im /= im.max()
+
+    return im
+
+
+def cms():
```
adler-j: Any chance you could rename this to e.g. "building" or something? Easier to find for users

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,111 @@
...
+    name = 'rings.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset='images', url=url)
+    im = np.rot90(dct['im'].astype('float64'), k=3)
+    im /= im.max()
+
+    return im
+
+
+def blurring_kernel():
```
adler-j: Unless this is some very specific kernel, it should perhaps be generated on the fly which would make it more general.

odl/contrib/datasets/images/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Access to a couple of images."""
```
adler-j: Somewhat less formal would be better i guess :D

odl/contrib/datasets/README.md
```diff
@@ -21,3 +21,6 @@ Reference datasets with accompanying ODL geometries etc.
     CT data as provided by FIPS. The data is non-human and high resolution.
     * `walnut_data`
     * `lotus_root_data`
+* `images`
+
+  miscellaneous images.
```
adler-j: I guess it would be nice if you added a sub-section for `cambridge`, and in there you can also add a direct link to `http://store.maths.cam.ac.uk/DAMTP/me404/data_sets/` so users can go there from github

odl/contrib/datasets/images/cambridge.py
adler-j: Also an overall comment: Try to document the size (in pixels) and output type of all the images

mehrhardt: Well, I have the code somewhere in MATLAB where you can define intensity levels for meaningful regions. If you are interested I can share this with you but I think I won't be interested in making this available to ODL myself.

mehrhardt: The main reason why I am not interested in this any more is because it is intrinsically 2D. To have a meaningful 3D version of this would be good.

odl/contrib/datasets/images/cambridge.py
mehrhardt: This image I have drawn in gimp. Not sure how to create a meaningful kernel on the fly.

odl/contrib/datasets/images/cambridge.py
mehrhardt: Agreed. How long would it take to do this? Maybe I can write a todo note there about this and we can implement this at some other point.

odl/contrib/datasets/README.md
mehrhardt: What do you mean here? How would I add a link?

adler-j: You can add it in cleartext, e.g.

```
* `images`

  Two dimensional images.
  * `cambridge`
  
    Various images as provided by the University of Cambridge, see http://store.maths.cam.ac.uk/DAMTP/me404/data_sets/
    * `brain_phantom`
    * `resolution_phantom`
    * `building`
    * `rings`
    * `blurring_kernel`    
```

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,151 @@
...
+import numpy as np
+import odl
+from odl.contrib.datasets.util import get_data
+from scipy.misc import imresize
+
+__all__ = ('brain_phantom', 'resolution_phantom', 'building', 'rings',
+           'blurring_kernel')
+
+
+DATA_SUBSET = 'images'
```
adler-j: this needs to be more specific, e.g. `images_cambridge` to avoid collisions with other image libraries in the future

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    # TODO: Store data in some ODL controlled url
+    name = 'motionblur.mat'
+    url = URL_CAM + name
+    dct = get_data(name, subset=DATA_SUBSET, url=url)
+
+    return convert(255 - dct['im'], size, to_odl, normalize='sum')
+
+
+def convert(image, size, to_odl, gray=None, dtype='float64', normalize='max'):
```
adler-j: We usually put utilities on top of the file (easier to see what is referenced when it is later used in functions).

Also, please add a one line doc-string.

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,151 @@
...
+    if size is not None:
+        image = imresize(image, size)
+
+    if normalize == 'max':
+        image = image / image.max()
+
+    elif normalize == 'sum':
+        image = image / image.sum()
+
+    if to_odl:
```
adler-j: I would personally vote against a `to_odl` function since it can be a bit confusing for users. Explicitly writing

```
space.element(cambridge.building())
```

is (imo) more clear.

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,151 @@
...
+
+def convert(image, size, to_odl, gray=None, dtype='float64', normalize='max'):
+
+    image.astype(dtype)
+
+    if gray is not None and gray:
+        image = np.mean(image, 2)
+
+    if size is not None:
+        image = imresize(image, size)
```
adler-j: Typically `size` is a number (e.g. the number of pixels in the image) while `shape` is the number of pixels per axis, I guess we should change `size` -> `shape` here

odl/contrib/datasets/images/examples/cambridge_images.py
```diff
@@ -0,0 +1,16 @@
+"""Example of the Cambridge images."""
+
+import odl.contrib.datasets.images as images
+
+images.brain_phantom(size=[200, 200]).show()
```
adler-j: Adding titles to these makes it easier to view later, e.g.

```
images.brain_phantom('brain_phantom', size=[200, 200]).show()
```

odl/contrib/datasets/images/examples/cambridge_images.py
```diff
@@ -0,0 +1,16 @@
...
+images.building().show()  # this is shown as gray scales ATM
+images.building(gray=True).show()
+images.rings(gray=True).show()
+images.rings(size=[100, 100]).show()  # this is shown as gray scales ATM
+images.blurring_kernel().show()
+
+# Run doctests
+# pylint: disable=wrong-import-position
+from odl.util.testutils import run_doctests
+run_doctests()
```
adler-j: This run_doctests in examples was a misstake by me in some other file and should not be there. Sorry about that :/

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,151 @@
...
+
+    if to_odl:
+        shape = image.shape
+
+        if len(shape) == 2:
+            space = odl.uniform_discr([0, 0], shape, shape)
+            image = space.element(image)
+        elif len(shape) == 3:
+            shape = shape[:2]
+            space = odl.uniform_discr([0, 0], shape, shape) ** 3
```
adler-j: ** shape[0]  just to be safe

odl/contrib/datasets/images/cambridge.py
mehrhardt: You mean that all images are converted to ODL without the flag or that no images are converted to ODL?

mehrhardt:
I made all changes as requested, except for the `to_odl` flag where I need more clarification, see above.

mehrhardt:
We also still have the issue with visualization of colour images. The current code produces three output images of the three channels.

mehrhardt:
I manually mashed the 7 commits into 1 to not spoil the history too much.

odl/contrib/datasets/images/cambridge.py
adler-j: I would suggest that no images are converted, e.g. that this simply returns numpy arrays.

mehrhardt: OK, done.

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,145 @@
...
+    """Convert image to standardized format.
+
+    Several properties of the input image may be changed including the shape,
+    data type and maximal value of the image. In addition, this function may
+    convert the image into an ODL object and/or a gray scale image.
+    """
+
+    image.astype(dtype)
+
+    if gray is not None and gray:
```
kohr-h: Weird construct, why not `gray=False` in the function signature and `if gray:` here?

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def brain_phantom(shape=None):
+    """Brain phantom for FDG PET simulations.
+
+    Output:
+        image type: gray scales
+        shape: [1024, 1024] (if not specified by `size`)
+        scale: [0, 1]
+        type: float64
```
kohr-h: Can you convert that to our Numpydoc style? Same for the other functions

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Several properties of the input image may be changed including the shape,
+    data type and maximal value of the image. In addition, this function may
+    convert the image into an ODL object and/or a gray scale image.
+    """
+
+    image.astype(dtype)
+
+    if gray is not None and gray:
+        image = np.mean(image, 2)
```
kohr-h: I don't think `mean` is the best way to convert color images to grayscale. I'd suggest you try to compare with
```python
image[..., 0] *= 0.2126
image[..., 1] *= 0.7152
image[..., 2] *= 0.0722
image = np.sum(image, axis=2)
```
see [here](https://en.wikipedia.org/wiki/Grayscale#Colorimetric_.28perceptual_luminance-preserving.29_conversion_to_grayscale).

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if shape is not None:
+        image = imresize(image, shape)
+
+    if normalize == 'max':
+        image = image / image.max()
```
kohr-h: Do this in place

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,145 @@
...
+
+    elif normalize == 'sum':
+        image = image / image.sum()
```
kohr-h: In place

odl/contrib/datasets/images/examples/cambridge_images.py
```diff
@@ -0,0 +1,45 @@
...
+        space = odl.uniform_discr([0, 0], shape, shape)
+    elif len(shape) == 3:
+        d = shape[2]
+        shape = shape[:2]
+        image = np.transpose(image, (2, 0, 1))
+        space = odl.uniform_discr([0, 0], shape, shape) ** d
+
+    image = space.element(image)
+
+    return image
```
kohr-h: I'd expect this to get the image axes a bit wrong. The conversion must be from `ij` to `xy`, which means that

- `shape` must be `reversed(image.shape[:2])`
- the image must (in each channel) be rotated by 90 degrees clockwise by `np.rot90(image, k=-1)`

odl/contrib/datasets/images/cambridge.py
mehrhardt: Sure, we can do this.

odl/contrib/datasets/images/cambridge.py
mehrhardt: Sure, I can change this. I don't think it is very important for these images though.

odl/contrib/datasets/images/cambridge.py
mehrhardt: What is your Numpydoc style?

odl/contrib/datasets/images/examples/cambridge_images.py
mehrhardt: Many of the images get rotated: `im = np.rot90(dct['im'], k=3)` The example file plot all images and they look the way they should do, so I think it is fine the way it is?

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,158 @@
...
+
+def convert(image, shape, gray=False, dtype='float64', normalize='max'):
+    """Convert image to standardized format.
+
+    Several properties of the input image may be changed including the shape,
+    data type and maximal value of the image. In addition, this function may
+    convert the image into an ODL object and/or a gray scale image.
+    """
+
+    image.astype(dtype)
```
adler-j: `image.astype(dtype)` returns a copy, so `image = image.astype(dtype)` would be correct

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if shape is not None:
+        image = imresize(image, shape)
+
+    if normalize == 'max':
+        image /= image.max()
+
+    elif normalize == 'sum':
+        image /= image.sum()
+
```
adler-j:
```
else:
    assert False
```

catch wrong inputs is nice :)

odl/contrib/datasets/images/cambridge.py
```diff
@@ -0,0 +1,158 @@
...
+        image[..., 1] *= 0.7152
+        image[..., 2] *= 0.0722
+        image = np.sum(image, axis=2)
+
+    if shape is not None:
+        image = imresize(image, shape)
+
+    if normalize == 'max':
+        image /= image.max()
+
```
adler-j: remove blankline here, makes some (including me) think that the if statement is over

odl/contrib/datasets/images/examples/cambridge_images.py
kohr-h: But then I understand this function wrong. I read this as "take image in Numpy array, return ODL space element", where the array hasn't been manipulated in any way. Expecting an already rotated image is surprising here.

Actually this kind of functionality would be useful in core ODL, too. Converting an image to ODL representation, including rotation and specifying a data type.

mehrhardt: I don't mind where to do the conversion but I see that this would be very useful for any kind of image and not just here in this example. However, I propose this can be discussed somewhere else and we adjust this example accordingly once we have figured this out.

mehrhardt:
Yes, I would merge it the way it is.

adler-j:
Thanks, great contribution!

PR #1223, opened 2017-11-09
ENH: add min/max_pt to ellipsoid_phantom

kohr-h:
All ellipsoid phantoms can now be created with provided `min_pt` and `max_pt`, for shifted or scaled versions.

### DIFF
--- a/odl/phantom/geometric.py
+++ b/odl/phantom/geometric.py
@@ -11,6 +11,9 @@
 from __future__ import print_function, division, absolute_import
 import numpy as np
 
+from odl.discr.lp_discr import uniform_discr_fromdiscr
+from odl.util.numerics import resize_array
+
 __all__ = ('cuboid', 'defrise', 'ellipsoid_phantom', 'indicate_proj_axis',
            'smooth_cuboid', 'tgv_phantom')
 
@@ -20,8 +23,8 @@ def cuboid(space, min_pt=None, max_pt=None):
 
     Parameters
     ----------
-    space : `DiscretizedSpace`
-        Discretized space in which the phantom is supposed to be created.
+    space : `DiscreteLp`
+        Space in which the phantom should be created.
     min_pt : array-like of shape ``(space.ndim,)``, optional
         Lower left corner of the cuboid. If ``None`` is given, a quarter
         of the extent from ``space.min_pt`` towards the inside is chosen.
@@ -88,22 +91,35 @@ def cuboid(space, min_pt=None, max_pt=None):
     return space.element(phantom)
 
 
-def defrise(space, nellipses=8, alternating=False):
+def defrise(space, nellipses=8, alternating=False, min_pt=None, max_pt=None):
     """Phantom with regularily spaced ellipses.
 
     This phantom is often used to verify cone-beam algorithms.
 
     Parameters
     ----------
-    space : `DiscretizedSpace`
-        Discretized space in which the phantom is supposed to be created.
-        Needs to be 2d or 3d.
+    space : `DiscreteLp`
+        Space in which the phantom should be created, must be 2- or
+        3-dimensional.
     nellipses : int, optional
         Number of ellipses. If more ellipses are used, each ellipse becomes
         thinner.
     alternating : bool, optional
         True if the ellipses should have alternating densities (+1, -1),
         otherwise all ellipses have value +1.
+    min_pt, max_pt : array-like, optional
+        If provided, use these vectors to determine the bounding box of the
+        phantom instead of ``space.min_pt`` and ``space.max_pt``.
+        It is currently required that ``min_pt >= space.min_pt`` and
+        ``max_pt <= space.max_pt``, i.e., shifting or scaling outside the
+        original space is not allowed.
+
+        Providing one of them results in a shift, e.g., for ``min_pt``::
+
+            new_min_pt = min_pt
+            new_max_pt = space.max_pt + (min_pt - space.min_pt)
+
+        Providing both results in a scaled version of the phantom.
 
     Returns
     -------
@@ -116,8 +132,7 @@ def defrise(space, nellipses=8, alternating=False):
     """
     ellipses = defrise_ellipses(space.ndim, nellipses=nellipses,
                                 alternating=alternating)
-
-    return ellipsoid_phantom(space, ellipses)
+    return ellipsoid_phantom(space, ellipses, min_pt, max_pt)
 
 
 def defrise_ellipses(ndim, nellipses=8, alternating=False):
@@ -126,7 +141,7 @@ def defrise_ellipses(ndim, nellipses=8, alternating=False):
     Parameters
     ----------
     ndim : {2, 3}
-        Dimension of the space the ellipses should be in.
+        Dimension of the space for the ellipses/ellipsoids.
     nellipses : int, optional
         Number of ellipses. If more ellipses are used, each ellipse becomes
         thinner.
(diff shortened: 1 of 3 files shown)


odl/phantom/geometric.py
```diff
@@ -449,24 +477,24 @@ def _ellipsoid_phantom_3d(space, ellipsoids):
     --------
     shepp_logan : The typical use-case for this function.
     """
+    # Blank volume
+    p = np.zeros(space.shape, dtype=space.dtype)
 
-    # Blank image
-    p = np.zeros(space.shape)
+    minp = space.grid.min_pt
```
adler-j: Needs an if statement here?

odl/phantom/geometric.py
```diff
@@ -330,24 +348,24 @@ def _ellipse_phantom_2d(space, ellipses):
     --------
     shepp_logan : The typical use-case for this function.
     """
-
     # Blank image
     p = np.zeros(space.shape, dtype=space.dtype)
 
+    minp = space.grid.min_pt
```
adler-j: If statement?

odl/phantom/geometric.py
```diff
@@ -90,22 +91,26 @@ def phantom(x):
...
     nellipses : int, optional
         Number of ellipses. If more ellipses are used, each ellipse becomes
         thinner.
     alternating : bool, optional
         True if the ellipses should have alternating densities (+1, -1),
         otherwise all ellipses have value +1.
+    min_pt, max_pt : array-like, optional
+        If provided, use these vectors to determine the bounding box of the
+        phantom instead of ``space.min_pt`` and ``space.max_pt``. Providing
+        one of them results in a shift, and providing both in a scaling.
```
adler-j: This needs further explanation

odl/phantom/geometric.py
```diff
@@ -611,14 +650,62 @@ def ellipsoid_phantom(space, ellipsoids):
...
-        return _ellipsoid_phantom_3d(space, ellipsoids)
+        _phantom = _ellipsoid_phantom_3d
     else:
         raise ValueError('dimension not 2 or 3, no phantom available')
 
+    if min_pt is None and max_pt is None:
+        return _phantom(space, ellipsoids)
+
+    else:
+        # Generate a temporary space with given `min_pt` and `max_pt`
```
adler-j: Is this truly the best way to implement this? Can't we change the function to do this properly?

kohr-h: Yes, I tried internal first, but the ellipses are all cropped because the bounding box calculation is wrong. I don't have an overview about what exactly it does, so I don't know (and don't want to invest loads of time into finding out) what exactly happens. This implementation is way simpler.

odl/phantom/geometric.py
kohr-h: What is unclear?

odl/phantom/geometric.py
kohr-h: No, this is exactly the old function, except for stuff moved around.

kohr-h: That is, the `min_pt` and `max_pt` arguments need to be removed.

odl/phantom/geometric.py
kohr-h: Added a bit of text.

adler-j: What do we mean by " a shift", where does it go? In what units? Perhaps by example is best here.

odl/phantom/geometric.py
adler-j: Ah I see, removing the arguments solves it.

odl/phantom/geometric.py
kohr-h: Example added

kohr-h:
Changes done. I think implementation-wise this is the simplest way of going forward. If we feel that this is clumsy or inefficient we can still go and change it.

odl/phantom/geometric.py
```diff
@@ -609,14 +658,62 @@ def ellipsoid_phantom(space, ellipsoids):
...
+        # (snapped to the cell grid), create the phantom in that space and
+        # resize to the target size for `space`.
+        if min_pt is not None:
+            if np.any(np.less(min_pt, space.min_pt)):
```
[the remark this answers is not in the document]
kohr-h: Not quite, rather something like 
```python
snapped_min_pt = space.partition[space.partition.index(min_pt)]
```
but that gives a `(1, 1)` partition and not the points.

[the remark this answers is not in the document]
adler-j: What would the difference of what I did and what you want to do here be?

odl/phantom/geometric.py
kohr-h: > Well, are we 100% sure we prefer this?

Not necessarily, but I also don't want to go into the low-level code that takes care of computing the bounding boxes for the ellipses and find out what part assumes that we look at the whole space all the time, and how to change it.

I honestly think that this solution is very acceptable and not a hack, just think about the logic of it: you create the fully sized phantom in the restricted (or shifted) space, and then you embed it into the original space. The only reason I'm not using a `ResizingOperator` or similar is that you don't want to make users hit the cell boundaries exactly with their `min_pt` and `max_pt`.

odl/phantom/geometric.py
```diff
@@ -542,36 +570,57 @@ def _ellipsoid_phantom_3d(space, ellipsoids):
...
+            'center_x', 'center_y', 'center_z',
+            'rotation_phi', 'rotation_theta', 'rotation_psi'
 
-        'value', 'axis_1', 'axis_2', 'center_x', 'center_y', 'rotation'
+        The provided ellipsoids need to be specified relative to the
+        reference rectangle ``[-1, -1] x [1, 1]``, or analogously in 3d.
+        The angles are to be given in radians.
 
-        If ``space`` is 3-dimensional each row should contain:
+    min_pt, max_pt : array-like, optional
```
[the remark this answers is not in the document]
kohr-h: > This is rather weird since all other parameters work in the set [-1, 1]^n, and here we suddenly work in "real" coordinates so to say. This should at least be mentioned.

I don't agree. The `space` is also not relative to [-1, 1]^n, and `min_pt` and `max_pt` explicitly relate to that.

odl/phantom/transmission.py
```diff
@@ -124,6 +124,19 @@ def shepp_logan(space, modified=False):
...
         The modified phantom has greatly amplified contrast to aid
         visualization.
+    min_pt, max_pt : array-like, optional
+        If provided, use these vectors to determine the bounding box of the
+        phantom instead of ``space.min_pt`` and ``space.max_pt``.
+        It is currently required that ``min_pt >= space.min_pt`` and
+        ``max_pt <= space.max_pt``, i.e., shifting or scaling outside the
+        original space is not allowed.
+
+        Providing one of them results in a shift, e.g., for ``max_pt``::
```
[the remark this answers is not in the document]
kohr-h: ok

odl/phantom/geometric.py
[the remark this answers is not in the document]
kohr-h: You retrieve the index only, I want the spatial location.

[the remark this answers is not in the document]
adler-j: But then simply do something like

```
snapped_min_pt = np.asarray(space.partition[space.partition.index(min_pt)])
```

or something like that.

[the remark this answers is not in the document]
kohr-h: ~~That doesn't work unfortunately, and also note that for min I want the lower left and for max the upper right corner of the indexed cell.~~ Oh wait, I think I could do
```python
space.partition[space.partition.index(min_pt)].min_pt
```
The max may need some hackery, but I'll check. Thx for the hint.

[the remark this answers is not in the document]
kohr-h: Much simpler implementation now.

kohr-h:
Anything more to do?

PR #1230, opened 2017-11-13
Discrete convolution

### LINKED ISSUE #209: Add convolution
We currently have convolution available in odl-examples, and a primitive cuda convolution available in this package. These should be merged into a well ordered convolution function made available similarly to the current derivative.

kohr-h:
Closes #209 

Note: this does not include the discretized continuous convolution, only the fully discrete one. But the former can be expressed in terms of the latter.


kohr-h:
@pep8speaks check this

adler-j:
Review ready?

kohr-h:
I'll get the auto-weighting thing out of the way, then yes.

kohr-h:
> I'll get the auto-weighting thing out of the way, then yes.

Actually I would prefer you have a look at #1177 first, since I don't want to add code here just to fix that missing part.

examples/space/auto_weighting.py
```diff
@@ -0,0 +1,57 @@
+"""Example demonstrating the usage of the ``auto_weighting`` decorator."""
```
adler-j: This reads more like a test than an example, is it possible to make it easier on the eyes?

odl/oplib/convolution.py
```diff
@@ -0,0 +1,736 @@
...
+            return x_ift[slc]
+
+    @property
+    @auto_weighting
+    def adjoint(self):
+        """Adjoint of the convolution operator.
+
+        The adjoint convolution is a convolution with the adjoint
+        kernel, which is the (complex conjugate of the) original kernel,
+        (roughly) flipped in the convolution axes. See Notes.
```
adler-j: notes where?

adler-j:
@ozanoktem walked by my desk and asked how this is going :smile:

kohr-h:
> @ozanoktem walked by my desk and asked how this is going 😄

😄 Well some stuff in the review that needs to be addressed. I also realized (by using the code) that a functional interface would be hugely useful for prototyping. Currently the functions `convolve` and `correlate` work by creating an operator and then calling it, but I think I'll do it the other way around, have the operator use the functional interface internally.

adler-j:
Well hello! How are we doing today?

kohr-h:
I did a rebase, plus some initial work on the more detailed interface for the implementation. WIP

mehrhardt:
@kohr-h , maybe this motivates you: I am looking forward to have this in ODL :-)

kohr-h:
> @kohr-h , maybe this motivates you: I am looking forward to have this in ODL :-)

^^ It does!

kohr-h:
Hold your horses, I only rebased!

adler-j:
horses

kohr-h:
:running_man:

Issue #909, opened 2017-02-15
Weighting with list of 1d vectors


kohr-h:
Between constant and "full array" weighting, there's weighting by an outer product of vectors that can be efficiently stored. Question: do we need this? It's a TODO in #861 anyway but won't be resolved there.

adler-j:
Main use case would be the weighting with uniformly sampled ray-transform data, for example.

It may be worth it, depending on the effort needed.

PR #1231, opened 2017-11-13
WIP: add cupy tensor space

kohr-h:


adler-j:
Looks like a trivial review, I'll try it tomorrow. Does this work with astra data containers?

kohr-h:
Sure, go ahead. It's work in progress though.

kohr-h:
I haven't tried ASTRA interoperability yet.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+            # matter anymore.
+            lico(a, x1.data, b, x2.data, out.data)
+
+
+# --- Space and element classes --- #
+
+
+class CupyTensorSpace(TensorSpace):
+
+    """Tensor space implemented with GPU arrays.
```
adler-j: This should mention what backend is used.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+                                     ''.format(self.shape,
+                                               self.weighting.array.shape))
+
+        elif dist is not None:
+            self.__weighting = CupyTensorSpaceCustomDist(dist)
+        elif norm is not None:
+            self.__weighting = CupyTensorSpaceCustomNorm(norm)
+        elif inner is not None:
+            self.__weighting = CupyTensorSpaceCustomInner(inner)
+        else:  # all None -> no weighing
```
adler-j: What happened to our no weighting vs const weighting debate? I.e. in several cases we expect stuff like this to be weighted:

```
space(weighting=n/10.0)  # "randomly" gives no weighting for n=10
```

I guess here it matters less than for the discretized spaces, but in them it really maters.

```
spc = odl.uniform_discr(0, 5, n)  # 'randomly' not weighted for n=5
```

this causes problems downstream for e.g. `RayTransform` which has special behaviour for non-weighted spaces.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+                                     ''.format(self.shape, inp.shape))
+
+                if isinstance(inp, cupy.ndarray):
+                    # Workaround for https://github.com/cupy/cupy/issues/590
+                    # TODO: remove when solved
+                    if (inp.dtype == self.dtype and
+                            inp.device.id == self.device):
+                        arr = inp
+                    else:
+                        arr = inp.astype(self.dtype)
```
adler-j: What if the device is wrong?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+        return '{}({})'.format(constructor_name, inner_str)
+
+    @property
+    def element_type(self):
+        """`CupyTensor`"""
+        return CupyTensor
+
+    @staticmethod
+    def available_dtypes():
+        """Return the data types available for this space."""
```
adler-j: But in lincomb we only support a very small subset of these?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+        -------
+        dtype : `numpy.dtype`
+            Numpy data type specifier. The returned defaults are:
+
+            ``RealNumbers()`` : ``np.dtype('float64')``
+
+            ``ComplexNumbers()`` : not supported
+        """
+        if field is None or field == RealNumbers():
+            return np.dtype('float64')
```
adler-j: Should be float32 on GPU systems for sure.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+                return float(np.asarray(arr))
+            elif arr.dtype.kind == 'c':
+                return complex(np.asarray(arr))
+            elif arr.dtype.kind in ('u', 'i'):
+                return int(np.asarray(arr))
+            else:
+                raise RuntimeError("no conversion for dtype {}"
+                                   "".format(arr.dtype))
+        else:
+            space = type(self.space)(arr.shape, dtype=self.dtype,
```
adler-j: weighting?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+        True
+
+        Notes
+        -----
+        Those ufuncs which are implemented natively on the GPU incur no
+        significant overhead. However, for missing functions, a fallback
+        Numpy implementation is used which causes significant overhead
+        due to data copies between host and device.
+        """
+        # TODO: Test with some native ufuncs, then remove this attribute
```
adler-j: fix this todo

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+        """Setter for the real part.
+
+        This method is invoked by ``tensor.real = other``.
+
+        Parameters
+        ----------
+        newreal : `array-like` or scalar
+            The new real part for this tensor.
+        """
+        self.real.data[:] = newreal
```
adler-j: I think `self.data.real[:]` would be slightly more efficient?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+    @property
+    def real(self):
+        """Real part of this tensor.
+
+        Returns
+        -------
+        real : `CupyTensor` view with real dtype
+            The real part of this tensor as an element of an `rn` space.
+        """
+        # Only real dtypes currently
```
adler-j: Quite sure there were complex ones above?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+
+# --- Weightings --- #
+
+
+def _weighting(weights, exponent):
+    """Return a weighting whose type is inferred from the arguments."""
+    if np.isscalar(weights):
+        weighting = CupyTensorSpaceConstWeighting(weights, exponent=exponent)
+    else:
+        # TODO: sequence of 1D array-likes
```
adler-j: Cite the issue number

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+from odl.util import dtype_str, is_floating_dtype, signature_string
+
+try:
+    import cupy
+except ImportError:
+    CUPY_AVAILABLE = False
+else:
+    from pkg_resources import parse_version
+    if parse_version(cupy.__version__) < parse_version('2.0.0rc1'):
+        raise ImportError('cupy <2.0.0rc1 not supported')
```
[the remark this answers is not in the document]
kohr-h: The thing is that version 2.0.0rc1 adds support for complex arrays, I really don't want anything below. Actually we should go for 2.0.0, the latest one on PyPI. It wasn't out at the time of writing this.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+    CUPY_AVAILABLE = True
+
+
+__all__ = ('CupyTensorSpace',)
+
+
+# --- Space method implementations --- #
+
+
+lico = cupy.ElementwiseKernel(in_params='T a, T x, T b, T y',
```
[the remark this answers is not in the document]
kohr-h: They should be, since `T` is a type template, so the only alternative would be to compile for all types, which I cannot imagine. The good thing is that the kernels are cached on disk, so after the first call things run way faster.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+    ------
+    ValueError :
+        If the data type is not supported by cuBLAS.
+    """
+    if np.dtype(dtype) == 'float32':
+        prefix = 's'
+    elif np.dtype(dtype) == 'float64':
+        prefix = 'd'
+    else:
+        raise ValueError('dtype {!r} not supported by cuBLAS'.format(dtype))
```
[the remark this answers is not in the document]
kohr-h: cuBLAS itself supports it, maybe the bindings are not implemented yet in cupy. I'll check.

odl/space/cupy_tensors.py
kohr-h: I'd still need a convincing argument why weighting with constant 1 would be fundamentally different from no weighting. I currently see it just as an optimization that scraps a multiplication with 1.

odl/space/cupy_tensors.py
kohr-h: Shouldn't we go with the default of the backend?

odl/space/cupy_tensors.py
kohr-h: `lincomb` should also support all the others, just many of them without cuBLAS acceleration.

odl/space/cupy_tensors.py
kohr-h: No idea 😇 I'll check

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Implementation of tensor spaces using ``pygpu``."""
```
kohr-h: Obviously the docs need update :-)

odl/space/cupy_tensors.py
kohr-h: Or rather: I don't have a way to check right now, so I'll have to check the documentation. I *guess* what happens is that the `.astype()` method creates a copy, and the `cupy.cuda.Device()` context ensures that the new array is created on the given device. But I'll make sure that is true.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2248 @@
...
+        False
+
+        A `CupyTensorSpace` with the same properties is considered
+        equal:
+
+        >>> same_space = odl.CupyTensorSpace(2, dtype='float64')
+        >>> same_space == space
+        True
+        """
+        return (super().__eq__(other) and
```
kohr-h: Update

odl/space/cupy_tensors.py
adler-j: Agree, I just expected it to be float32

kohr-h: Well the question was open. If we find single precision more reasonable we can go for that as well.

odl/space/cupy_tensors.py
kohr-h: Needs update obviously.

odl/space/cupy_tensors.py
kohr-h: Very possible.

odl/space/cupy_tensors.py
adler-j: Is this not a very weird behaviour:

```
>>> spc = odl.uniform_discr([0, 0], [9.999, 9.999], [10, 10])
>>> rt = odl.tomo.RayTransform(spc, odl.tomo.parallel_beam_geometry(spc))
>>> rt.range
uniform_discr([  0.    , -14.1407], [  3.1416,  14.1407], (45, 31))
```

vs

```
>>> spc = odl.uniform_discr([0, 0], [9.999, 9.999], [10, 10])
>>> rt = odl.tomo.RayTransform(spc, odl.tomo.parallel_beam_geometry(spc))
>>> rt.range
uniform_discr([  0.    , -14.1421], [  3.1416,  14.1421], (45, 31), weighting=1.0)
```

I mean, wat? this *has to* be considered a rather severe bug, no?

kohr-h: That seems quite wrong. 

But part of the problem seems to me that we need to jump through hoops currently to make the adjoint correct for weighted vs. unweighted spaces. Doing it via something like #1177 will solve the issue very elegantly by keeping all weighting stuff completely outside the operator definitions.
And when such a system is in place, no weighting and weighting with 1 will be *exactly* equivalent.

adler-j: Well this has nothing to do with the adjoint (in fact we don't even need the adjoint to have this problem), the problem here is that we want "the same weighting scheme" on the range as on the domain, but if we dont distinguish no weighting and 1.0 weight, we can't do that.

kohr-h: Oh right, I was totally forgetting about the fact that we infer the range. Okay now I got it. Hm. We do have an optional `range` parameter for the ray transform, don't we? I tend to think that the behavior should be 

- as currently when we have a `TensorSpace` with `cell_volume` weighting,
- no weighting in any other case.

To make it easier to change this, we could have a `asweighted()` method on `TensorSpace` where a new space of the same type but with new weighting can be created.

adler-j:
Ok, so odlcuda is officially broken. I'll try to fix it, but getting this in would be a preferable solution. What is the ETA here, code looks quite good.

kohr-h:
I'll give it a couple of days of focused work, but those days will be distributed a bit. I give it high prio at least.

adler-j:
I'm fixing odlcuda anyway

kohr-h:
I'll look into this today. Also in view of #1246.

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: Okay, complex stuff is available in 2.0.0, I'll emit a warning for earlier versions.

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: I just checked. When I removed `pkg_resources` (bloody slow, we should avoid it whenever possible, at least during `import odl`), the import time of `cupy_tensors.py` is practically identical to the time of `import cupy` (~40 ms).

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: Supported in cupy 2.0.0

odl/space/cupy_tensors.py
adler-j: Well it's not quite obvious to me how we would figure out if we're using cell_volume weighting 1.0 or no weighting unless we expose this flag somehow. We previously did it by exposing thetype of the weighting, but I guess we could fix this by exposing the argument used to create the weighting (somehow), i.e. `space.weighting_type` returns a string `'const'` or w/e. Does that sound reasonable?

I feel adding yet another `as_...` should hopefully not be needed right now, but I guess in the long run.

odl/space/cupy_tensors.py
adler-j: You can use jupyterhub on radon which has 8 cores to check this

odl/space/cupy_tensors.py
adler-j: Well what does the backend use? This http://docs.chainer.org/en/v1.21.0/_modules/cupy/creation/basic.html tells it it's float64, and then we should follow that I guess.

kohr-h: Okay. They try to stick as closely to Numpy, which is good for us. I'll go with their default then.

odl/space/cupy_tensors.py
kohr-h: Good call!

odl/space/cupy_tensors.py
adler-j: Please add some comment on this in this function (i.e. "we use the same as...") for future readers

odl/space/cupy_tensors.py
kohr-h: I'll port the current implementation in `NumpyTensorSpace`, the proper solution will come with #1238 
Just in general: if you squash axes, there's no obvious way to propagate weightings.

odl/space/cupy_tensors.py
kohr-h: This raises a question: If we have a cupy space element and run some numpy code on it that requires a cast to numpy array, should we just go ahead and do it, at the cost of hidden performance loss, or should we raise?
I have a slight preference but want to make sure we're on the same page.

adler-j: Well as always im in the "at least it works" boat. I guess it comes down to how feature complete cupy is, if it has almost everything it is fine, but if you stumble every thing you do, it's gonna be irritating to work with

kohr-h: I'm in the same boat, so that's fine then. As far as I can see, they're pretty complete on the ufunc front, but methods like `at` and `accumulated` are largely not implemented. So we need some workaround to use the CUDA code for the obvious ones (`sum`, `prod`, `cumsum`, `cumprod` for `add` and `multiply`), but otherwise drop out to Numpy.

The cupy folks are a bit more worried about efficiency, probably since in neural nets, things are already buried deep down and hard to debug, so this kind of bottleneck danger would be pretty bad for them.

kohr-h:
Getting there. I fixed the complex stuff, and all the doctests go green. Unit tests next.

odl/space/cupy_tensors.py
adler-j: Go ahead with that then!

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+
+        A more convenient way is to use the factory functions with the
+        ``impl='cupy'`` option:
+
+        >>> space = odl.rn(3, impl='cupy', weighting=[1, 2, 3])
+        >>> space
+        rn(3, impl='cupy', weighting=[1, 2, 3])
+        >>> space = odl.tensor_space((2, 3), impl='cupy', dtype=int)
+        >>> space
+        tensor_space((2, 3), 'int', impl='cupy')
```
adler-j: Won't these doctests fail misserably for users without cupy, or do we exclude them somehow?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+            self.__weighting = CupyTensorSpaceCustomDist(dist)
+        elif norm is not None:
+            self.__weighting = CupyTensorSpaceCustomNorm(norm)
+        elif inner is not None:
+            self.__weighting = CupyTensorSpaceCustomInner(inner)
+        else:  # all None -> no weighing
+            self.__weighting = CupyTensorSpaceConstWeighting(1.0, exponent)
+
+    @property
+    def device(self):
```
adler-j: Are we sure we want to make device a space attribute, is it not very similar to order in a sense?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+                    # Short-circuit for space elements and no enforced ordering
+                    return inp
+
+                if hasattr(inp, 'shape') and inp.shape != self.shape:
+                    raise ValueError('`inp` must have shape {}, got shape {}'
+                                     ''.format(self.shape, inp.shape))
+
+                if isinstance(inp, cupy.ndarray):
+                    # Workaround for https://github.com/cupy/cupy/issues/590
+                    # TODO: remove when solved
```
adler-j: Do we have an issue number for this locally? (if not, create and add comment)

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+        >>> out
+        array([[ 1.,  2.,  3.],
+               [ 4.,  5.,  6.]], dtype=float32)
+        >>> result is out
+        True
+        """
+        if out is None:
+            return cupy.asnumpy(self.data)
+        else:
+            if out.shape != self.shape:
```
adler-j: If out is not a numpy array, this line will fail in weird manners

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+                                 ''.format(self.dtype, out.dtype))
+            self.data.data.copy_to_host(
+                out.ctypes.data_as(np.ctypeslib.ctypes.c_void_p),
+                self.size * self.itemsize)
+            return out
+
+    @property
+    def data_ptr(self):
+        """A raw pointer to the data container.
+
```
adler-j:
```
Returns
-------
data_ptr : int
```

to show the type of the returned value

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+        Returns
+        -------
+        equals : bool
+            ``True`` if all entries of ``other`` are equal to this
+            tensor's entries, ``False`` otherwise.
+
+        Notes
+        -----
+        The element-by-element comparison is performed on the CPU,
```
adler-j: Really? Why not write a `ReductionKernel` for this? Should be rather simple

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+        as supported:
+
+        >>> r5 = odl.rn(5, impl='cupy')
+        >>> x = r5.element([1, 2, 3, 4, 5])
+        >>> x[1:4]
+        rn(3, impl='cupy').element([ 2.,  3.,  4.])
+        >>> x[::2]
+        rn(3, impl='cupy').element([ 1.,  3.,  5.])
+
+        The returned views are writable, so modificatons alter the
```
adler-j: Use "view"

adler-j: Also mention that this is not always the case (i guess?)

odl/space/cupy_tensors.py
adler-j: We need to settle what squashed axes mean. Imo it should set that dimension to "1" so to say, i.e. we keep `space[0:1].weighting` and `space[0].weighting` different.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+# R = real floating point type, usually for norm or dist output
+# W = real type for weights, can be floating point or integer
+#
+# Note: the kernels with an output type that does not also occur as an input
+# type must be called with output argument since the output type cannot be
+# inferred. The ouptut array must have shape `()` for full reduction.
+
+
+if CUPY_AVAILABLE:
+    dotw = cupy.ReductionKernel(in_params='T x, T y, W w',
```
adler-j: Are these compiled on the fly later? We don't want to create startup latency

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+from odl.space.base_tensors import TensorSpace, Tensor
+from odl.space.weighting import (
+    Weighting, ArrayWeighting, ConstWeighting,
+    CustomInner, CustomNorm, CustomDist)
+from odl.util import (
+    array_str, dtype_str, is_floating_dtype, real_dtype,
+    signature_string, indent)
+
+try:
+    import cupy
```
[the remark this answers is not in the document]
kohr-h: 40 ms, it's really fast.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+def fallback_scal(a, x):
+    x *= a
+    return x
+
+
+def fallback_axpy(a, x, y):
+    return lico(a, x, 1, y, y)
+
+
+def _flat_inc(arr):
```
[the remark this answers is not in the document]
kohr-h: It's the `incx`, `incy` stuff needed for the cuBLAS functions.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2372 @@
...
+    """Compute the flat element stride for cuBLAS if possible, else raise."""
```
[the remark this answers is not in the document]
kohr-h: Was working on that this morning. The implementation was also not correct :-)

odl/space/cupy_tensors.py
kohr-h: At least not when running the file. I'll have to check how to configure `pytest` to exclude them as well.

odl/space/cupy_tensors.py
kohr-h: Fair question. I think it's more clear-cut with `device` in the sense that keeping data on a device is much easier than keeping the order of an array. Also, I would not like to have a situation where you add two elements of the same space but require copies between devices. Here `space.element()` will make the copy transparent.
Another nice side effect of having `device` as a space property is that you can in fact implement multi-GPU parallelism with product spaces :-).

odl/space/cupy_tensors.py
kohr-h: I prefer not to add `Returns` to `@property` docstrings since they do not "feel" like functions that return something. I'll add it somewhere else.

odl/space/cupy_tensors.py
kohr-h: It will say `[...] has no attribute 'shape'`, is that so weird?

odl/space/cupy_tensors.py
kohr-h: Not for index arrays and boolean arrays, right.

odl/space/cupy_tensors.py
kohr-h: Makes sense.

odl/space/cupy_tensors.py
kohr-h: Yes, as I wrote somewhere else, the import time of this module is practically identical to the import time of `cupy`, around 40 ms. So these guys don't add overhead at creation time.

odl/space/cupy_tensors.py
kohr-h: Well, as of now, i.e., before #909 is done (part of #1238) we can't know how to handle weighting constants when removing axes. If all axes stay intact, we can keep the constant. If an axis is removed, we have to fall back to default. Not so great, but yeah.
For array weighting, i.e., the weights have the same shape as the space, we can just index them the same way, so that's a bit easier.

odl/space/cupy_tensors.py
adler-j: Anything goes as long as the info is there!

odl/space/cupy_tensors.py
adler-j: Well I guess that's OK actually, but what about other stuff (i.e. if you feed this a `NumpyTensor`?)

odl/space/cupy_tensors.py
adler-j: Well, how about a `default_device` construction? Just trying to make sure we do the right thing here.

odl/space/cupy_tensors.py
adler-j: Well if you explicitly run this file, you should get errors IMO, its harder with the global tests

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
adler-j: Great! No worries then

odl/space/cupy_tensors.py
kohr-h: Well yeah, I give up 😬 I'll do the instance check. After all, the docstring says a Numpy array is required.

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: I redid the cuBLAS handling, works as expected now. Also the `lico` fallback kernel takes arrays of different dtypes.

odl/space/cupy_tensors.py
kohr-h: Done

odl/space/cupy_tensors.py
kohr-h: Done.

odl/space/cupy_tensors.py
kohr-h: I've implemented the manual fiddling.

odl/space/cupy_tensors.py
kohr-h: Well for me, device 1 vs. device 2 is not much different from CPU vs GPU. And we decided to make those spaces different, for good reasons I think. 

The biggest reason is that is allows you to  to always know where the data is and not have any unwanted and costly transfers between physically separate memory. 
And the same reason applies IMO for the multi-device case. Transfers between them may be less expensive (I don't know exactly), but I'm sure implicit transfers would still be bad.
Another aspect of this is distribution work load. If the space guarantees its data to be on a specific device, you can do things like
```python
>>> rn_cpu = odl.rn((2, 1000000))
>>> rns_gpu = [odl.rn(1000000, impl='cupy', device=i) for i in (0, 1)]
>>> multi_gpu_space = odl.ProductSpace(*rns_gpu)
>>> chunk_op = ChunkingOperator(rn_cpu, multi_gpu_space)  # hypothetical
>>> busy_ops = [VeryExpensiveOperator(rn_gpu) for rn_gpu in rns_gpu]
>>> busy_op = odl.DiagonalOperator(*busy_ops)
>>> gather_op = GatherOperator(multi_gpu_space, rn_cpu)
>>> dist_gpu_op = gather_op * busy_op * chunk_op
```
I find this version of multi-GPU parallelism really appealing.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2743 @@
...
+            if (ufunc in (np.add, np.multiply) and
+                    method in ('reduce', 'accumulate')):
+                use_native = native_ufunc is not None
+            else:
+                use_native = (native_ufunc is not None and
+                              hasattr(native_ufunc, method))
+
+        force_native = kwargs.pop('force_native', False)
+        if force_native and not use_native:
+            raise ValueError(
```
kohr-h: I added this to give the user a way to ensure that no implicit dropdown to Numpy happens. You can do
```python
>>> rn = odl.rn((3, 4), impl='cupy')
>>> x = rn.one()
>>> np.add.reduce(x, force_native=True)
rn(4, impl='cupy').element([ 3.,  3.,  3.,  3.])
>>> np.divide.reduce(x, force_native=True)
  ...
ValueError: no native function available to evaluate `true_divide.reduce` [...]
```
It's really useful for debugging. Any objections?

adler-j: Functionality is great, but we may want to follow naming from other places, e.g. numba uses `nopython`:

http://numba.pydata.org/numba-doc/dev/user/jit.html

So here i guess `nofallback`?

adler-j:
How close is this to final review, e.g. whats missing?

odl/space/cupy_tensors.py
kohr-h: Well yeah, I kinda dislike the negative flags, in particular when the default is `False` like here. It's just unnecessarily hard to make out what it means. Compare `nofallback=False` with `force_native=False`. IMO the second is much clearer.
I'll look around for some other comparable flag name, `nofallback` is really not good.

kohr-h:
> How close is this to final review, e.g. whats missing?

Still got a bunch of tests to fix, but it feels like the big stuff is out of the way. The remaining things to fix (in the tensor tests) are mostly `__array__` issues and tests I haven't ported yet.
I'll give it another morning, and it should be close to done.

kohr-h:
Update: all tensor tests pass now, except for skips (mostly `float128` and `complex256`) and xfails (mainly ufuncs with sketchy complex support, like `greater_equal`, but honestly nothing serious). Moving on to other code and tests using `tspace_impl`.

kohr-h:
SO, for me all tests pass now, and I thing they're quite harsh. That means this is ready for review!

The details turned out to require some extra coding, wrapping etc. But once the tensor tests passed, the rest was not THAT big of a change. On the contrary, 95 % of the library just worked immediately with the new backend.

That said, most of the code is not optimized for the cupy backend. I don't expect that to be a big effort, at least not in places where we already used the `x.ufuncs` interface for everything. That also makes me think that we should keep that interface and not deprecate it, given that it made the transition was so smooth.
As a small proof of concept, I ported the `diff_ops` to work with native cupy arrays, and that turned out to be super-easy.

Another thing I need to mention: the unit tests with both backends (not including largescale) take REALLY long time. On my laptop 10 minutes. I suspect that part of it is `pytest` being slow, since also the Travis tests take almost 3 times as long as they used to. Maybe something to investigate.
But in any case, I think it makes sense to add an option flag for running the GPU tests, and set it to `False` by default.

odl/discr/diff_ops.py
```diff
@@ -559,11 +559,13 @@ def _call(self, x, out=None):
 
         ndim = self.range.ndim
         dx = self.range.cell_sides
+        impl = self.range.impl
 
-        tmp = np.empty(out.shape, out.dtype, order=out.space.default_order)
-        with writable_array(out) as out_arr:
+        tmp = array_module(impl).empty(
+            out.shape, out.dtype, order=out.space.default_order)
```
kohr-h: This of course assumes that `impl` has the same interface as numpy. Perhaps it's better to implement wrappers for the most common functions, like `asarray` (already done), `empty`, and whatever else is needed. That would be a bit more extensible.

odl/discr/discretization.py
```diff
@@ -331,16 +331,18 @@ def copy(self):
...
         Parameters
         ----------
-        out : `numpy.ndarray`, optional
-            Array in which the result should be written in-place.
-            Has to be contiguous and of the correct dtype.
+        out : ndarray, optional
+            Array into which the result should be written. Must be contiguous
+            and of the correct dtype.
+        impl : str, optional
+            Array backend for the output, used when ``out`` is not given.
```
kohr-h: Should probably mention where the valid choices come from.

odl/set/space.py
```diff
@@ -387,7 +387,7 @@ def __pow__(self, shape):
             shape = tuple(shape)
 
         pspace = self
-        for n in shape:
+        for n in reversed(shape):
```
kohr-h: This was an actual bug.

odl/space/npy_tensors.py
```diff
@@ -864,24 +864,24 @@ def data(self):
         """The `numpy.ndarray` representing the data of ``self``."""
         return self.__data
 
-    def asarray(self, out=None):
-        """Extract the data of this array as a ``numpy.ndarray``.
+    def asarray(self, out=None, impl='numpy'):
+        """Extract the data of this array as an ndarray.
```
kohr-h: Perhaps make a glossary term `ndarray` that explains the different flavors.

odl/space/npy_tensors.py
```diff
@@ -2266,7 +2288,7 @@ def norm(self, x):
         if self.exponent == 2.0:
             return float(np.sqrt(self.const) * _norm_default(x))
         elif self.exponent == float('inf'):
-            return float(self.const * _pnorm_default(x, self.exponent))
+            return float(_pnorm_default(x, self.exponent))
```
kohr-h: I've changed this because there are two reasons for it:

1. The mathematical one: `||x||_{p, w} --> ||x||_{inf, w}` should hold for `p --> inf` (we knew that).
2. Also structure-wise, all unweighted p-norms follow the pattern
   ```
   norm = post_map( reduce( map(x) ) )
   ```
   e.g. 

   - `1 <= p < inf`: `map(x) = abs(x) ** p`, `reduce(x) = sum(x)`, `post_map(x) = x ** (1/p)`
   - `p = inf`: `map(x) = abs(x)`, `reduce(x) = max(x)`, `post_map(x) = x`

   Weighted norms have the structure
   ```
   norm = post_map( reduce( weight * map(x) ) )
   ```
   and here it follows naturally that the weights play no role for `p = inf` since they do not change the `max`. Why treat this case completely differently?

odl/space/weighting.py
```diff
@@ -500,7 +500,7 @@ def array(self):
 
     def is_valid(self):
         """Return True if the array is a valid weight, i.e. positive."""
-        return np.all(np.greater(self.array, 0))
+        return (self.array > 0).all()
```
kohr-h: Again assuming a certain API of the arrays. Perhaps better to have wrapper `all`, `any` and such. Dunno.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -380,21 +384,28 @@ def __repr__(self):
...
+# Avoid tons of warnings on the console when testing
+npy_err_old = np.geterr()
+np.seterr(all='ignore')
 
 # Create an operator for each ufunc
-for name, nargin, nargout, docstring in UFUNCS:
+for name, nin, nout, docstring in UFUNCS:
+    if nin == 2:
+        # Currently not supported
+        continue
```
kohr-h: We had a bunch of dysfunctional ufunc ops created here.

odl/util/testutils.py
```diff
@@ -94,6 +94,19 @@ def almost_equal(a, b, places=None):
...
+    if CUPY_AVAILABLE:
+        if isinstance(iter1, CupyTensor):
+            iter1 = iter1.asarray()
+        elif isinstance(iter1, cupy.ndarray):
+            iter1 = cupy.asnumpy(iter1)
+
+        if isinstance(iter2, CupyTensor):
+            iter2 = iter2.asarray()
+        elif isinstance(iter2, cupy.ndarray):
+            iter2 = cupy.asnumpy(iter2)
```
kohr-h: Should use `utility.asarray` here I guess.

odl/util/testutils.py
```diff
@@ -181,6 +203,14 @@ def is_subdict(subdict, dictionary):
     return all(item in dictionary.items() for item in subdict.items())
 
 
+def xfail_if(condition, reason=''):
+    """Return a ``pytest.xfail`` object if ``condition`` is ``True``."""
```
kohr-h: Explain usage.

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -225,21 +225,23 @@ def __init__(self, space):
...
+        if nin == 1:
             domain = space0 = space
-            dtypes = [space.dtype]
-        elif nargin == len(space) == 2 and isinstance(space, ProductSpace):
+            dtypes_in = [space.dtype]
+        elif nin == 2:
+            if not (isinstance(space, ProductSpace) and len(space) == 2):
+                raise TypeError('`space` must be a `ProductSpace` of length '
+                                '{} for ufunc {!r}, got {!r}'
+                                ''.format(nin, name, space))
```
adler-j: `nin` can have only one value here, why use format for it?

odl/ufunc_ops/ufunc_ops.py
adler-j: Good catch

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -436,6 +447,7 @@ def ufunc_factory(domain=RealNumbers()):
     globals()[name] = ufunc_factory
     __all__ += (name,)
 
+np.seterr(**npy_err_old)
```
adler-j: We should honestly add some "we're so sorry" line above this :D

odl/util/testutils.py
adler-j: yes

odl/util/testutils.py
```diff
@@ -433,13 +481,21 @@ def noise_elements(space, n=1):
     noise_array
     noise_element
     """
-    arrs = tuple(noise_array(space) for _ in range(n))
+    from odl.space.pspace import ProductSpace
+
+    if isinstance(space, ProductSpace) and not space.is_power_space:
+        raise ValueError('`space` cannot be a non-power product space')
```
adler-j: Why not? Part of the reason we use this function is that it supports these things (as compared to e.g. `space.element(np.random.rand)`)

odl/util/ufuncs.py
```diff
@@ -183,28 +224,176 @@ def max(self, axis=None, dtype=None, out=None, keepdims=False):
...
+        ]
+    ])
+    """
+    from odl.space.pspace import ProductSpace, ProductSpaceElement
+    from odl.space.base_tensors import Tensor
+
+    # Workaround for `shape` not using the base space shape of a
+    # power space
+    # TODO: remove when fixed, see
+    # https://github.com/odlgroup/odl/pull/1152
```
adler-j: I'll get this in ASAP

odl/set/space.py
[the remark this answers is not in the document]
kohr-h: I don't think so. When doing `space[i]`, the index slices along the *outermost* power, i.e., for `space ** (3, 4)` you expect the valid indices `i` to be `0, 1, 2`, no? It's the same logic as `(R^m)^n == R^(n x m)`.
In your example, I would expect that `r2 ** (3, 4)` has shape `(3, 4, 2)`.

odl/util/ufuncs.py
adler-j: This is now in, yay!

adler-j:
So, I'm officially announcing that I'll "take over" this PR and make a new PR for it.

odl/discr/diff_ops.py
adler-j: That would be how e.g. Keras does it, but I say we leave this for now. Changing should be rather easy (just search replace basically).

odl/set/space.py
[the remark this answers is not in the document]
adler-j: I concur with your statement. It certainly makes sense.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+        _cublas_scal = _cublas_func('scal', arr.dtype)
+    except (ValueError, AttributeError):
+        return _fallback_scal
+
+    def scal(a, x):
+        """Implement ``x <- a * x`` with constant ``a``."""
+        return _cublas_scal(
+            x.data.device.cublas_handle, x.size, a, x.data.ptr, incx)
+
+    scal.__name__ = scal.__qualname__ = '_cublas_scal'
```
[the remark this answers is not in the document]
adler-j: Renamed this.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+    return axpy
+
+
+def _lincomb_impl(a, x1, b, x2, out):
+    """Linear combination implementation, assuming types have been checked.
+
+    This implementation is a highly optimized, considering all special
+    cases of array alignment and special scalar values 0 and 1 separately.
+    """
+    scal1 = _get_scal(x1.data)
```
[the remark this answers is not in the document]
adler-j: Cache is an optimization that will have to come later. I'll focus on getting this in first.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+
+    See https://github.com/cupy/cupy for details on the backend.
+    """
+
+    def __init__(self, shape, dtype='float64', device=None, **kwargs):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        shape : sequence of non-negative ints
```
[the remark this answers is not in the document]
adler-j: Did some improvements to this, standardized doc with `NumpyTensorSpace`

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+            Number entries per dimension.
+        dtype :
+            Data type for each tuple entry. Can be provided in any
+            way the `numpy.dtype` function understands, e.g.,
+            as built-in type, as one of NumPy's internal datatype
+            objects or as string.
+            See `available_dtypes` for the list of supported data types.
```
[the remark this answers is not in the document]
adler-j: Clarified this by adding "scalar"

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+            `shape` as the space.
+
+            sequence of 1D array-likes: Per-axis (tensor product) weighting
+            using broadcasting multiplication in each axis. ``None``
+            entries cause the corresponding axis to be skipped.
+
+            This option cannot be combined with ``dist``,
+            ``norm`` or ``inner``.
+
+            Default: no weighting
```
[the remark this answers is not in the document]
adler-j: Clarified this

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+        A `CupyTensorSpace` with the same properties is considered
+        equal:
+
+        >>> same_space = odl.CupyTensorSpace(2, dtype='float64')
+        >>> same_space == space
+        True
+        """
+        return (super(CupyTensorSpace, self).__eq__(other) and
+                self.device == other.device and
+                self.weighting == other.weighting)
```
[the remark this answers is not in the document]
adler-j: It seems `TensorSpace` does not have a weighting, so no.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+            use_native = False
+        elif not all(isinstance(inp, type(self.data)) or np.isscalar(inp)
+                     for inp in inputs):
+            native_ufunc = None
+            use_native = False
+        else:
+            native_ufunc = getattr(cupy, ufunc.__name__, None)
+
+            # Manually implemented cases
+            if (
```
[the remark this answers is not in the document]
adler-j: I reduced the "spam" a bit here, nothing major

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+
+        # --- Process `inputs` --- #
+
+        # Pull out the data container of the inputs if necessary
+        inputs = [inp.data if isinstance(inp, type(self)) else inp
+                  for inp in inputs]
+
+        # Determine native ufunc vs. Numpy ufunc
+        if any(isinstance(o, np.ndarray) for o in out_tuple):
+            native_ufunc = None
```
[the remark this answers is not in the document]
adler-j: I need some input here @kohr-h, i should be able to merge these if statements, right?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+            # https://github.com/odlgroup/odl/issues/1248
+            # We must exclude 'at' since it's in-place, thus we're not allowed
+            # to change the input.
+            for i in range(len(inputs)):
+                if isinstance(inputs[i], cupy.ndarray):
+                    inputs[i] = cupy.asnumpy(inputs[i])
+                elif isinstance(inputs[i], CupyTensor):
+                    inputs[i] = cupy.asnumpy(inputs[i].data)
+
+        # For debugging
```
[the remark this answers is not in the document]
adler-j: Removed these

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+        if use_native:
+            assert all(isinstance(i, cupy.ndarray) or
+                       np.isscalar(i) or
+
```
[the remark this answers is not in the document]
adler-j: Changed this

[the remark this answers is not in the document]
adler-j: This ended up in the wrong place, but the question remains. Shouldn't we use `conjugate` or something like it here? I'll add a test for this.

kohr-h:
> So, I'm officially announcing that I'll "take over" this PR and make a new PR for it.

Go for it! Maybe close this one for difficulty to open the page?

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+    1
+    >>> _get_flat_inc(arr_f, arr_f)
+    (1, 1)
+    >>> _get_flat_inc(arr_c, arr_f)
+    Traceback (most recent call last):
+        ...
+    ValueError
+
+    Slicing in the **fastest** axis is allowed as it results in a constant
+    stride in the flat memory. Slicing with stride in any other axis is
```
kohr-h: Oops, "is" shouldn't be there

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: Agreed, we have the same thing on the Numpy spaces, we can add caches for both in one go.

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: That's right, the abstract spaces don't define weighting, and in principle it's not required.

odl/space/cupy_tensors.py
kohr-h: BTW, I did some profiling lately, and this function turned out to be a bottleneck. In some PR (I think #1238) I changed this to use a `set` instead of a tuple, which cut down the time to create a new space by quite a bit.
If you chose to do so here as well, be aware that `set` compares by hashing, not by `__eq__`, so you have to use `np.dtype(dtype)` as comparison, otherwise `float` won't be considered valid, for instance.

odl/space/cupy_tensors.py
```diff
@@ -0,0 +1,2819 @@
...
+        .. warning::
+            Apart from ``'__call__'`` (invoked by, e.g., ``np.add(x, y))``,
+            CuPy has no native implementation of ufunc methods like
+            ``'reduce'`` or ``'accumulate'``. We manually implement the
+            mappings (covering most use cases)
+
+            - ``np.add.reduce`` -> ``cupy.sum``
+            - ``np.add.accumulate`` -> ``cupy.cumsum``
+            - ``np.multiply.reduce`` -> ``cupy.prod``
+            - ``np.multiply.reduce`` -> ``cupy.cumprod``.
```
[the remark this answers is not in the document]
kohr-h: Yep

odl/space/cupy_tensors.py
kohr-h: Maybe `allow_fallback=True` would be an alternative?

adler-j:
I'll close this once I've fixed the outstanding issues

odl/space/cupy_tensors.py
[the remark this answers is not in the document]
kohr-h: So by "native" I mean "CuPy" here, which could easily replace "native" in the names as well.

The reasoning for this distinction here is like this:
- If the output is (or contains) a Numpy array, use Numpy (=fallback). This is essential to support `numpy_array += odl_elem_using_cupy` for example, since it will use `out=(numpy_array,)`.
- If any of the operands is not either a scalar or a CuPy array, use Numpy. This is debatable, but any version will involve copies.
- If all inputs and all outputs are scalars or CuPy arrays, try to use the CuPy ufunc.

The success of "try to" depends on whether the function is implemented, which is true for almost all basic ufuncs, some reductions and accumulations, but nothing else. So there's another possibility for fallback here.

I guess you could merge the two first fallback cases logically.

odl/space/npy_tensors.py
```diff
@@ -503,9 +502,10 @@ def default_dtype(field=None):
         dtype : `numpy.dtype`
             Numpy data type specifier. The returned defaults are:
 
-                ``RealNumbers()`` : ``np.dtype('float64')``
+            - ``RealNumbers()`` : ``np.dtype('float64')``
```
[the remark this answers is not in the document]
adler-j: Ok, so this says returns so i guess we should be rigorous

odl/space/npy_tensors.py
```diff
@@ -928,7 +963,7 @@ def astype(self, dtype):
 
     @property
     def data_ptr(self):
-        """A raw pointer to the data container of ``self``.
+        """Memory address of the data container as 64-bit integer.
```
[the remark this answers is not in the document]
adler-j: Updated to document this

odl/test/discr/diff_ops_test.py
```diff
@@ -347,7 +347,7 @@ def test_gradient(space, method, padding):
     # Check not to use trivial data
     assert lhs != 0
     assert rhs != 0
-    assert almost_equal(lhs, rhs, places=places)
```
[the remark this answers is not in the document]
adler-j: That won't happen here though

odl/test/discr/lp_discr_test.py
```diff
@@ -731,12 +724,16 @@ def test_cell_volume():
...
 
 
-def test_astype():
+def test_astype(tspace_impl):
 
-    rdiscr = odl.uniform_discr([0, 0], [1, 1], [2, 2], dtype='float64')
-    cdiscr = odl.uniform_discr([0, 0], [1, 1], [2, 2], dtype='complex128')
-    rdiscr_s = odl.uniform_discr([0, 0], [1, 1], [2, 2], dtype='float32')
-    cdiscr_s = odl.uniform_discr([0, 0], [1, 1], [2, 2], dtype='complex64')
+    rdiscr = odl.uniform_discr([0, 0], [1, 1], [2, 2], dtype='float64',
```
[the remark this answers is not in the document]
adler-j: Fixed

odl/test/discr/lp_discr_test.py
```diff
@@ -760,174 +757,202 @@ def test_ufuncs(tspace_impl, ufunc):
...
+            assert False
+
+    def _check_result_is_out(result, out_seq):
+        if nout == 1:
+            assert result is out_seq[0]
+        elif nout > 1:
+            for i in range(nout):
+                assert result[i] is out_seq[i]
+        else:
+            assert False
```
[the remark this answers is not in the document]
adler-j: It should not

odl/test/discr/lp_discr_test.py
```diff
@@ -760,174 +757,202 @@ def test_ufuncs(tspace_impl, ufunc):
...
+
+    # See https://github.com/cupy/cupy/issues/794
+    cupy_ufuncs_broken_complex = [
```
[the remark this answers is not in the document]
adler-j: Used only once, I'll leave it in here.

odl/space/npy_tensors.py
```diff
@@ -902,10 +902,45 @@ def asarray(self, out=None):
         array([[ 1.,  1.,  1.],
                [ 1.,  1.,  1.]])
         """
+        from odl.space.cupy_tensors import cupy, CUPY_AVAILABLE
```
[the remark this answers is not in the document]
adler-j: Aah, its None otherwise, added a comment on this

odl/test/discr/lp_discr_test.py
```diff
@@ -760,174 +757,202 @@ def test_ufuncs(tspace_impl, ufunc):
...
+
+    # In-place -- in = elements -- out = numpy or own arrays -- ufunc = numpy
+    # This case is only supported with the new interface
     if USE_ARRAY_UFUNCS_INTERFACE:
-        out_tensors_new = tuple(space.tspace.element(np.empty_like(arr))
-                                for arr in out_arrays)
+        # Fresh arrays for output
+        out_arrays_npy = [np.empty_like(as_numpy(arr))
+                          for arr in arrays[nin:]]
+        out_arrays_own = [array_module(space.impl).empty_like(arr)
```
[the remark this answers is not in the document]
adler-j: As per above. We'll leave this for now and see when it fails. Changing it is not too hard.

odl/test/discr/lp_discr_test.py
```diff
@@ -760,174 +757,202 @@ def test_ufuncs(tspace_impl, ufunc):
...
+
+        try:
+            result_npy = ufunc_npy.reduce(in_array)
+        except TypeError:
+            pytest.xfail('numpy ufunc.reduce not valid for inputs')
+
+        # Out-of-place -- in = element
+        result = ufunc_npy.reduce(in_elem)
+        assert all_almost_equal(result, result_npy)
+        # keepdims not allowed
```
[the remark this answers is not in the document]
adler-j: I guess for lazy reasons right now. lets keep until someone needs it, it is quite rare.

odl/test/space/pspace_test.py
```diff
@@ -192,24 +192,27 @@ def test_metric():
     HxH = odl.ProductSpace(H, H, exponent=1.0)
     w1 = HxH.element([v11, v12])
     w2 = HxH.element([v21, v22])
-    assert almost_equal(HxH.dist(w1, w2),
-                        H.dist(v11, v21) + H.dist(v12, v22))
+    dist = HxH.dist(w1, w2)
+    expected_dist = sum([H.dist(v11, v21), H.dist(v12, v22)])
```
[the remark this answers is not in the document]
adler-j: Fixed

odl/test/space/pspace_test.py
```diff
@@ -192,24 +192,27 @@ def test_metric():
...
     # 2-norm
     HxH = odl.ProductSpace(H, H, exponent=2.0)
     w1 = HxH.element([v11, v12])
     w2 = HxH.element([v21, v22])
-    assert almost_equal(
-        HxH.dist(w1, w2),
-        (H.dist(v11, v21) ** 2 + H.dist(v12, v22) ** 2) ** (1 / 2.0))
+    dist = HxH.dist(w1, w2)
+    expected_dist = np.sqrt(
+        sum(d ** 2 for d in (H.dist(v11, v21), H.dist(v12, v22)))
```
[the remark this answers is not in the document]
adler-j: Changed all these to use this style, easier to read and less error prone.

n1kt0:
How far is the integration process?

Thanks in advance and best regards,

Nikita

n1kt0:
How far is the integration process?

Thanks in advance and best regards,
Nikita

n1kt0:
How far is the integration process?



Thanks in advance and best regards,

Nikita

adler-j:
This has sadly stalled quite badly and I will not have time to focus on it in the coming months. If anyway wants to pick up I'd be more than happy.

n1kt0:
Are there some tests or do you have a summary what needs to be done or
where one could start?

Am Do., 16. Mai 2019 um 17:13 Uhr schrieb Jonas Adler <
notifications@github.com>:

adler-j:
Sadly there is no summary but the main issue is to get the tests running. With that said this PR has become so old that it might be worth re-starting it with the newly updated spaces, perhaps @kohr-h knows how much has changes that touches upon this?

kohr-h:
I'm hesitating to re-activate this one before we have settled on #1459 and #1458. Depending on the outcome of the latter in particular, getting in support for CuPy will be massively simpler (or not).

PR #1235, opened 2017-11-14
Issue 1234  disable bool

adler-j:


odl/test/space/tensors_test.py
```diff
@@ -903,6 +903,27 @@ def test_conversion_to_scalar(tspace_impl):
             long(element)
 
 
+def test_bool_conversion(tspace_impl):
+    """Verify that the __array__ interface for NumPy works."""
```
kohr-h: Likely copy-paste

PR #1240, opened 2017-11-21
Issue 571  callback primal dual

adler-j:
Could use some feedback on this before I do it in general, but I think the overall idea should be quite applicable, and we keep the user-interface very clean.

### DIFF
--- a/odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
+++ b/odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
@@ -16,6 +16,7 @@ from __future__ import print_function, division, absolute_import
 import numpy as np
 
 from odl.operator import Operator
+from odl.solvers.util.callback import call_callback
 
 
 __all__ = ('pdhg',)
@@ -297,8 +298,7 @@ def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
         # Over-relaxation in the primal variable x
         x_relax.lincomb(1 + theta, x, -theta, x_old)
 
-        if callback is not None:
-            callback(x)
+        call_callback(x, dual_iterate=y)
 
 
 if __name__ == '__main__':
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -32,8 +32,8 @@ def call_callback(callback, x, **kwargs):
 
     Parameters
     ----------
-    callback : callable
-        Function taking at least one argument.
+    callback : callable or None
+        Function taking at least one argument. If `None`, does nothing.
     x :
         The main object to call the callback with, typically the current
         iterate.
@@ -59,6 +59,9 @@ def call_callback(callback, x, **kwargs):
     >>> call_callback(other_callback, 'arg', b='other_arg')
     arg other_arg
     """
+    if callback is None:
+        return
+
     if sys.version_info.major > 2:
         sig = inspect.signature(callback)
         filter_keys = [param.name for param in sig.parameters.values()
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -13,8 +13,10 @@ from builtins import object
 import copy
 import numpy as np
 import os
+import sys
 import time
 import warnings
+import inspect
 
 from odl.util import signature_string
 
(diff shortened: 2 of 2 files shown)


odl/solvers/util/callback.py
```diff
@@ -25,6 +27,62 @@
...
+    arg
+
+    Optionally ignore one argument:
+
+    >>> call_callback(callback, 'arg', b='other_arg')
+    arg
+    >>> def other_callback(x, b=None):
+    ...     print(x, b)
+    >>> call_callback(other_callback, 'arg', b='other_arg')
+    arg other_arg
```
kohr-h: The example is  really confusing. Above it says "Optionally ignore one argument", but the example doesn't do that, it just happens to print nothing for the default. Nowhere is `None` ignored. I was looking in the code below for something that ignores `None`, without success.

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -297,8 +298,7 @@ def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
         # Over-relaxation in the primal variable x
         x_relax.lincomb(1 + theta, x, -theta, x_old)
 
-        if callback is not None:
-            callback(x)
+        call_callback(x, dual_iterate=y)
```
mehrhardt: would this not be `call_callback(callback, x, dual_iterate=y)`?

odl/solvers/util/callback.py
```diff
@@ -25,6 +27,62 @@
            'CallbackProgressBar')
 
 
+def call_callback(callback, x, **kwargs):
```
mehrhardt: Why do we need this function? Why not just `callback(x, **kwargs)`? I.e. why do we need to filter the `kwargs`?

odl/solvers/util/callback.py
kohr-h: I also find using strings and the name `'arg'` confusing, since it looks as if `'arg'` was the name of a parameter, where it's actually just some value. How about just using numbers in the example?

odl/solvers/util/callback.py
```diff
@@ -25,6 +27,62 @@
...
+        spec = inspect.getargspec(callback)
+        func_kwargs = spec.keywords
+
+        if func_kwargs:
+            callback(x, **kwargs)
+        else:
+            filtered_kwargs = {filter_key: kwargs[filter_key]
+                               for filter_key in filter_keys
+                               if filter_key in kwargs}
+            callback(x, **filtered_kwargs)
```
mehrhardt: If we do it like this, then many callbacks have to be adjusted for this structure, like the concatenation of callbacks.

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
adler-j: You are totally correct!

odl/solvers/util/callback.py
kohr-h: This will only work at the top level, i.e. when passing in Python functions or callbacks with a well-defined signature (not using `*args` or `**kwargs`). Anything else will either break due to invalid args being passed in, or it will have to silently ignore arguments, which I'm against. `CallbackAnd` is a good example for that.

adler-j:
Without going into the details, is there some better way to do this? We have (imo) some hard requirements

* The method must not break the current flexibility of specifying callbacks, e.g. giving a function or lambda as a callback should work as it does now.
* Some optimizer should be able to give specific information to the callback, which the callback can *optionally* handle
* We need extensibility, e.g. we cannot define say 3 keywords now and expect them to last forever.

I guess the other way would be to add a line in the beginning of the code that does basically

```
if not isinstance(callback, Callback):
    callback = CallbackWrapper(callback)
```

and then hide the logic in there, but the internal logic would still be basically exactly this?

kohr-h:
> Without going into the details, is there some better way to do this? We have (imo) some hard requirements
> 
>     The method must not break the current flexibility of specifying callbacks, e.g. giving a function or lambda as a callback should work as it does now.
>     Some optimizer should be able to give specific information to the callback, which the callback can optionally handle
>     We need extensibility, e.g. we cannot define say 3 keywords now and expect them to last forever.

Fully agree on the list of requirements, and nobody said it was an easy problem :-)

I'll give it some thought..

mehrhardt:
Any news on this front? 
I now have also other use cases where `**kwargs` would be really good like providing `Ax` to an objective function evaluation if it has been calculated already before.

adler-j:
Closing in favor of #1252 solution.

PR #1242, opened 2017-11-22
STY: Fix PEP8 whitespace issue in astra_cuda.py.

sbanert:


kohr-h:
I'm wondering why PEP8speaks doesn't say anything. Is there an error in the config?

PR #1243, opened 2017-11-22
ENH: Add the alternating dual update method

sbanert:
This pull request adds a new (nonsmooth) solver, which is, according to the authors of the corresponding paper, especially good in doing tomography problems on a graphics card.

It contains the changes of #1166, and the corresponding test uses the functionality provided by #1166 as requested by Jonas in the discussion there.

### DIFF
--- a/odl/solvers/nonsmooth/adupdates.py
+++ b/odl/solvers/nonsmooth/adupdates.py
@@ -21,24 +21,24 @@ __all__ = ('adupdates_method',)
 
 def adupdates_method(x, g, L, stepsize, majs, niter, random=False,
                      callback=None, callback_loop='outer'):
-    """Alternating Dual updates method.
+    r"""Alternating Dual updates method.
 
     The Alternating Dual (AD) updates method of McGaffin and Fessler `[MF2015]
     <http://ieeexplore.ieee.org/document/7271047/>`_ is designed to solve an
-    optimization problem of the form
+    optimization problem of the form ::
 
-        min_x sum_i g_i(L_i x)
+        min_x [ sum_i g_i(L_i x) ]
 
     where ``g_i`` are proper, convex and lower semicontinuous functions and
-    ``L_i`` are linear `Operator`s.
+    ``L_i`` are linear `Operator` s.
 
     Parameters
     ----------
-    g : sequence of `Functional`s
-        All functions need to provide a `convex_conj` with a
-        `proximal` factory.
-    L : sequence of `Operator`s
-        Needs to provide as many elements as `g`.
+    g : sequence of `Functional` s
+        All functions need to provide a `Functional.convex_conj` with a
+        `Functional.proximal` factory.
+    L : sequence of `Operator` s
+        Needs to provide as many elements as ``g``.
     x : `LinearSpaceElement`
         Initial point, updated in-place.
     stepsize : positive float
@@ -47,19 +47,19 @@ def adupdates_method(x, g, L, stepsize, majs, niter, random=False,
         performance might depend on the choice of a good stepsize.
     majs : sequence of majorizers
         Parameters determining the stepsizes for the inner iterations. Must be
-        matched with the norms of `L`, and convergence is guaranteed if
-        `majs` are large enough. See the Notes section for details.
+        matched with the norms of ``L``, and convergence is guaranteed if
+        ``majs`` are large enough. See the Notes section for details.
     niter : int
         Number of (outer) iterations.
     random : bool, optional
         If `True`, the order of the dual upgdates is chosen randomly,
-        otherwise the order provided by the lists `g`, `L` and `majs`
+        otherwise the order provided by the lists ``g``, ``L`` and ``majs``
         is used.
     callback : callable, optional
         Function called with the current iterate after each iteration.
     callback_loop : {'inner', 'outer'}, optional
-       If 'inner', the `callback` function is called after each inner
-       iteration, i.e., after each dual update. If 'outer', the `callback`
+       If 'inner', the ``callback`` function is called after each inner
+       iteration, i.e., after each dual update. If 'outer', the ``callback``
        function is called after each outer iteration, i.e., after each primal
        update.
 
@@ -103,20 +103,19 @@ def adupdates_method(x, g, L, stepsize, majs, niter, random=False,
 
     where :math:`M_j` is a diagonal matrix with positive diagonal entries such
     that :math:`M_j - L_j L_j^*` is positive semidefinite. The matrix
-    :math:`M_j` is incorporated into this function via the variable `majs` in
+    :math:`M_j` is incorporated into this function via the variable ``majs`` in
     one of the following ways:
 
-    * Setting `majs[i]` a positive float :math:`\gamma` corresponds to the
-    choice :math:`M_i = \gamma \mathrm{Id}`.
-
-    * Assume that `g_j` is a `SeparableSum`, then setting `majs[i]` a list
-    :math:`(\gamma_1, \ldots, \gamma_k)`of positive floats corresponds to the
-    choice of a block-diagonal matrix :math:`M_j`, where each block corresponds
-    to one of the space components and equals :math:`\gamma_i \mathrm{Id}`.
-
-    * Assume that `g_j` is a `L1_norm` or a `L2_norm_squared`, then setting
-    `majs[i]` a `g_j.domain.element` :math:`z` corresponds to the choice
-    :math:`\mathrm{diag}(z)`.
+    * Setting ``majs[i]`` a positive float :math:`\gamma` corresponds to the
+      choice :math:`M_i = \gamma \mathrm{Id}`.
+    * Assume that ``g_j`` is a `SeparableSum`, then setting ``majs[i]`` a
+      list :math:`(\gamma_1, \ldots, \gamma_k)` of positive floats corresponds
+      to the choice of a block-diagonal matrix :math:`M_j`, where each block
+      corresponds to one of the space components and
+      equals :math:`\gamma_i \mathrm{Id}`.
+    * Assume that ``g_j`` is an `L1Norm` or an `L2NormSquared`, then setting
+      ``majs[i]`` a ``g_j.domain.element`` :math:`z` corresponds to the choice
+      :math:`\mathrm{diag}(z)`.
 
     References
     ----------
--- a/odl/solvers/nonsmooth/adupdates.py
+++ b/odl/solvers/nonsmooth/adupdates.py
@@ -16,7 +16,7 @@ from __future__ import print_function, division, absolute_import
 
 import numpy as np
 
-__all__ = ('adupdates', 'adupdates_simple')
+__all__ = ('adupdates',)
 
 
 def adupdates(x, g, L, stepsize, majs, niter, random=False,
--- a/odl/solvers/nonsmooth/adupdates.py
+++ b/odl/solvers/nonsmooth/adupdates.py
@@ -12,7 +12,6 @@ The alternating dual upgrade method is reported to be a flexible method which
 allows an acceleration by permitting efficient calculations on a graphics card.
 """
 
-# Imports for common Python 2/3 codebase
 from __future__ import print_function, division, absolute_import
 
 import numpy as np
--- a/odl/solvers/nonsmooth/adupdates.py
+++ b/odl/solvers/nonsmooth/adupdates.py
@@ -1,0 +1,177 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Alternating dual (AD) upgrade algorithm studied by McGaffin and Fessler.
+
+The alternating dual upgrade method is reported to be a flexible method which
+allows an acceleration by permitting efficient calculations on a graphics card.
+"""
+
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+
+import numpy as np
+
+__all__ = ('adupdates',)
+
+
+def adupdates(funcs, ops, x, stepsize, majs, niter, random=False,
+              callback=None, callback_loop='outer'):
+    """Alternating Dual updates method.
+
+    The Alternating Dual (AD) updates method of McGaffin and Fessler is
+    designed to solve an optimization problem of the form
+
+        min_x sum_i f_i(L_i x)
+
+    where ``f_i`` are proper, convex and lower semicontinuous functions and
+    ``L_i`` are linear `Operator`s
+
+    Parameters
+    ----------
+    funcs : sequence of `Functional`s
+        All functions need to provide a `convex_conj` with a
+        `proximal` factory.
+    ops : sequence of `Operator`s
+        Needs to provide as many elements as `funcs`.
+    x : `LinearSpaceElement`
+        Initial point, updated in-place.
+    stepsize : positive float
+        The stepsize for the outer (proximal point) iteration.
+    majs : sequence of majorizers
+        See the Notes section for details.
+    niter : int
+        Number of (outer) iterations.
+    random : bool, optional
+        If `True`, the order of the dual upgdates is chosen arbitrarily,
+        otherwise the order provided by the lists `funcs`, `ops` and `majs`
+        is used.
+    callback : callable, optional
+        Function called with the current iterate after each iteration.
+    callback_loop : string, optional
+       If 'inner', the `callback` function is called after each inner
+       iteration, i.e., after each dual update. If 'outer', the `callback`
+       function is called after each outer iteration, i.e., after each primal
+       update.
+
+    Notes
+    -----
+    The algorithm as implemented here is described in the article [MF2015],
+    where it is applied to a tomography problem. It solves the problem
+
+    .. math::
+        \min_x \sum_{i=1}^m f_i(L_i x),
+
+    where :math:`f_i` are proper, convex and lower semicontinuous functions
+    and :math:`L_i` are linear, continuous operators for
+    :math:`i = 1, \ldots, m`. In an outer iteration, the solution is found
+    iteratively by an iteration
+
+    .. math::
+        x_{n+1} = \mathrm{arg\,min}_x \sum_{i=1}^m f_i(L_i x)
+            + \frac{\mu}{2} \|x - x_n\|^2
+
+    according to the proximal point algorithm. In the inner iteration, dual
+    variables are introduces for each of the components of the sum. The
+    Lagrangian of the problem is given by
+
+    .. math::
+        S_n(x; v_1, \ldots, v_m) = \sum_{i=1}^m (\langle v_i, L_i x \rangle
+            - f_i^*(v_i)) + \frac{\mu}{2} \|x - x_n\|^2.
+
+    Given the dual variables, the new primal variable :math:`\tilde x_{n+1}`
+    can be calculated by directly minimizing :math:`S_n` with respect to
+    :math:`x`. This corresponds to the formula
+
+    .. math::
+         \tilde x_{n+1} = x_n - \frac{1}{\mu} \sum_{i=1}^m L_i^* v_i.
+
+    The dual updates are executed according to the following rule:
+
+    .. math::
+        v_j^+ = \mathrm{Prox}_{\mu M_j^{-1}}
+            (v_j + \mu M_j^{-1} L_j \tilde x_{n+1}),
+
+    where :math:`M_j` is a diagonal matrix with positive diagonal entries such
+    that :math:`M_j - L_j L_j^*` is positive semidefinite. The matrix
+    :math:`M_j` is incorporated into this function via the variable `majs` in
+    one of the following ways:
+
+    * Setting `majs[i]` a positive float :math:`\gamma` corresponds to the
+    choice :math:`M_i = \gamma \mathrm{Id}`.
+
+    * Assume that `f_j` is a `SeparableSum`, then setting `majs[i]` a list
+    :math:`(\gamma_1, \ldots, \gamma_k)`of positive floats corresponds to the
+    choice of a block-diagonal matrix :math:`M_j`, where each block corresponds
+    to one of the space components and equals :math:`\gamma_i \mathrm{Id}`.
+
+    * Assume that `f_j` is a `L1_norm` or a `L2_norm_squared`, then setting
+    `majs[i]` a `f_j.domain.element` :math:`z` corresponds to the choice
+    :math:`\mathrm{diag}(z)`.
+
+    References
+    ----------
+    [MF2015] McGaffin, M G, and Fessler, J A. *Alternating dual updates
+    algorithm for X-ray CT reconstruction on the GPU*. IEEE Transactions
+    on Computational Imaging, 1.3 (2015), pp 186--199.
+    """
+    # Check the lenghts of the lists (= number of dual variables)
+    length = len(funcs)
+    if not len(funcs) == len(ops):
+        raise ValueError('Number of `ops` {} and number '
+                         'of `funcs` {} do not agree.'
+                         ''.format(len(ops), len(funcs)))
+
+    # Check if operators have a common domain
+    # (the space of the primal variable):
+    domain = ops[0].domain
+    if any(domain != opi.domain for opi in ops):
+        raise ValueError('Domains of `ops` are not all equal.')
+
+    # Check if range of the operators equals domain of the functionals
+    ranges = [opi.range for opi in ops]
+    if any(ops[i].range != funcs[i].domain for i in range(length)):
+        raise ValueError('Ranges of operators and domains'
+                         'of functionals do not agree.')
+
+    # Initialization of the dual variables
+    duals = [space.zero() for space in ranges]
+
+    # Reusable elements in the ranges, one per type of space
+    unique_ranges = set(ranges)
+    tmp_rans = {ran: ran.element() for ran in unique_ranges}
+
+    # Prepare the proximal operators. Since the stepsize does not vary over
+    # the iterations, we always use the same proximal operator.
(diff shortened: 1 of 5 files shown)


odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+        else:
+            rng = range(len(ops))
+
+        for i in rng:
+            step = stepsize / majs[i]
+            arg = duals[i] + step * ops[i](x)
+            tmp_ran = tmp_rans[ops[i].range]
+            proxs[i](arg, out=tmp_ran)
+            x -= 1.0 / stepsize * ops[i].adjoint(tmp_ran - duals[i])
+            duals[i] = tmp_ran.copy()
```
adler-j: duals[i].assign(tmp_ran)

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+        # for i in range(length)])
+        for i in range(length):
+            x -= 1.0 / stepsize * ops[i].adjoint(duals[i])
+
+        if random:
+            rng = np.random.permutation(range(len(ops)))
+        else:
+            rng = range(len(ops))
+
+        for i in rng:
```
adler-j: Both loops use `i` as iteration variable, this is confusing. Prefer using different names, e.g. i and j

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+            step = stepsize / majs[i]
+            arg = duals[i] + step * ops[i](x)
+            tmp_ran = tmp_rans[ops[i].range]
```
adler-j: This whole code creates copies all over the place, I think it would be better to simply remove this temporary for now (alternatively fix all the copies, but that would be some work and I'm not sure how much it will help atm).

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Alternating dual (AD) upgrade algorithm studied by McGaffin and Fessler.
+
+The alternating dual upgrade method is reported to be a flexible method which
+allows an acceleration by permitting efficient calculations on a graphics card.
+"""
+
+# Imports for common Python 2/3 codebase
```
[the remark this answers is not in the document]
kohr-h: Yes we removed those comments.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Alternating dual (AD) upgrade algorithm studied by McGaffin and Fessler.
```
kohr-h: below it says "update" not "upgrade"

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+# Imports for common Python 2/3 codebase
+from __future__ import print_function, division, absolute_import
+
+import numpy as np
+
+__all__ = ('adupdates',)
+
+
+def adupdates(funcs, ops, x, stepsize, majs, niter, random=False,
+              callback=None, callback_loop='outer'):
```
kohr-h: I suggest that we use the same signature (where possible) as for the Douglas-Rachford method, i.e. `g` for functionals that are composed, `L` for the operators, etc.
See also #1197

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+
+    * Assume that `f_j` is a `L1_norm` or a `L2_norm_squared`, then setting
+    `majs[i]` a `f_j.domain.element` :math:`z` corresponds to the choice
+    :math:`\mathrm{diag}(z)`.
+
+    References
+    ----------
+    [MF2015] McGaffin, M G, and Fessler, J A. *Alternating dual updates
+    algorithm for X-ray CT reconstruction on the GPU*. IEEE Transactions
+    on Computational Imaging, 1.3 (2015), pp 186--199.
```
kohr-h: Reference looks perfect. If you want to use a citation with link, which would be something for the short summary before `Parameters`, you can use a link like
```
`[MF2015] <URL for DOI, arXiv or whatever>`_
```
This is at least my current favorite way of doing it, because it provides a clickable link *and* a text reference in the `References` section.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,177 @@
...
+            arg = duals[i] + step * ops[i](x)
+            tmp_ran = tmp_rans[ops[i].range]
+            proxs[i](arg, out=tmp_ran)
+            x -= 1.0 / stepsize * ops[i].adjoint(tmp_ran - duals[i])
+            duals[i] = tmp_ran.copy()
+
+            if callback is not None and callback_loop == 'inner':
+                callback(x)
+        if callback is not None and callback_loop == 'outer':
+            callback(x)
```
kohr-h: This is a pretty complex optimization method. I'm sure it's a good one.
What I can recommend is to make a "dead simple" implementation of the same optimizer that does not do any fancy stuff like caching, in-place operation of optimized arithmetic. Just with as little code as possible. That's really good for debugging. See for example [here](https://github.com/odlgroup/odl/blob/87e04941f4337de190654897691327f8964c0f03/odl/solvers/nonsmooth/admm.py#L149-L163).

odl/solvers/nonsmooth/adupdates.py
sbanert: I am currently doing this and I noticed that there is the parameter kwargs, which is not documented in the docstring of the Douglas–Rachford method. In the iterative methods, this parameter is not present at all. Which variant shall I implement here?

kohr-h: We don't really have a rule on that, but if you feel that the signature gets a bit too long with all the optional parameters, stuff the less common, less important or more advanced ones into `kwargs`. In this case, I'd say it's fine without `kwargs` since you have only one optional parameter `random` (I don't count `callback`s).

sbanert:
Should be ready from my side if Travis does not complain.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+
+        min_x sum_i g_i(L_i x)
+
+    where ``g_i`` are proper, convex and lower semicontinuous functions and
+    ``L_i`` are linear `Operator`s.
+
+    Parameters
+    ----------
+    g : sequence of `Functional`s
+        All functions need to provide a `convex_conj` with a
```
adler-j: dead link here, you need to write `Functional.convex_conj`

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+        `proximal` factory.
```
adler-j: Same as above, `Functional.proximal`

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+    L : sequence of `Operator`s
+        Needs to provide as many elements as `g`.
```
adler-j: This sentence could use some clarification. What is provided? Typically we say that the length of the sequence should be something

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+    x : `LinearSpaceElement`
+        Initial point, updated in-place.
+    stepsize : positive float
+        The stepsize for the outer (proximal point) iteration. The theory
+        guarantees convergence for any positive real number, but the
+        performance might depend on the choice of a good stepsize.
+    majs : sequence of majorizers
+        Parameters determining the stepsizes for the inner iterations. Must be
+        matched with the norms of `L`, and convergence is guaranteed if
+        `majs` are large enough. See the Notes section for details.
```
adler-j: majs is a dead link, use double \`\` for non-link

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+        `proximal` factory.
+    L : sequence of `Operator`s
+        Needs to provide as many elements as `g`.
+    x : `LinearSpaceElement`
+        Initial point, updated in-place.
+    stepsize : positive float
+        The stepsize for the outer (proximal point) iteration. The theory
+        guarantees convergence for any positive real number, but the
+        performance might depend on the choice of a good stepsize.
+    majs : sequence of majorizers
```
adler-j: majorizers is still not very obvious. What does this mean? If it is functions, just write "sequence of callables"

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+        Parameters determining the stepsizes for the inner iterations. Must be
+        matched with the norms of `L`, and convergence is guaranteed if
+        `majs` are large enough. See the Notes section for details.
+    niter : int
+        Number of (outer) iterations.
+    random : bool, optional
+        If `True`, the order of the dual upgdates is chosen randomly,
+        otherwise the order provided by the lists `g`, `L` and `majs`
```
adler-j: dead links

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+        is used.
+    callback : callable, optional
+        Function called with the current iterate after each iteration.
+    callback_loop : {'inner', 'outer'}, optional
+       If 'inner', the `callback` function is called after each inner
+       iteration, i.e., after each dual update. If 'outer', the `callback`
```
adler-j: dead link

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+    .. math::
+        S_n(x; v_1, \ldots, v_m) = \sum_{i=1}^m (\langle v_i, L_i x \rangle
+            - g_i^*(v_i)) + \frac{\mu}{2} \|x - x_n\|^2.
+
+    Given the dual variables, the new primal variable :math:`\tilde x_{n+1}`
+    can be calculated by directly minimizing :math:`S_n` with respect to
+    :math:`x`. This corresponds to the formula
+
+    .. math::
+         \tilde x_{n+1} = x_n - \frac{1}{\mu} \sum_{i=1}^m L_i^* v_i.
```
adler-j: What does tilde mean here?

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+
+    References
+    ----------
+    [MF2015] McGaffin, M G, and Fessler, J A. *Alternating dual updates
+    algorithm for X-ray CT reconstruction on the GPU*. IEEE Transactions
+    on Computational Imaging, 1.3 (2015), pp 186--199.
+    """
+    # Check the lenghts of the lists (= number of dual variables)
+    length = len(g)
+    if not len(L) == length:
```
adler-j:
```python
if len(L) != length:
```

is the preferable style

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+                         ''.format(len(L), length))
+
+    if not len(majs) == length:
+        raise ValueError('Number of `majs` {} and number of `g` {} do not'
+                         ' agree.'.format(len(majs), length))
+
+    # Check if operators have a common domain
+    # (the space of the primal variable):
+    domain = L[0].domain
+    if any(domain != opi.domain for opi in L):
```
adler-j: prefer swapping the order here to follow natural language, e..g. you would typically not say "if the expected domain is not equal to the domain", rather you would say "if the domain is not equal to the expected domain".

In code,

```python
if any(opi.domain != domain for opi in L):
```

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    # Check the lenghts of the lists (= number of dual variables)
+    length = len(g)
+    if not len(L) == length:
+        raise ValueError('Number of `L` {} and number of `g` {} do not agree.'
+                         ''.format(len(L), length))
+
+    if not len(majs) == length:
+        raise ValueError('Number of `majs` {} and number of `g` {} do not'
+                         ' agree.'.format(len(majs), length))
```
adler-j: We typically write the exception messages in lower case characters and without a dot, e.g.

```python
'number of `majs` {} and number of `g` {} do not agree'
```

See https://github.com/odlgroup/odl/pull/427 if you want a rationalle for this. Basically we follow the python core standard.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+
+The alternating dual upgrade method is reported to be a flexible method which
+allows an acceleration by permitting efficient calculations on a GPU.
+"""
+
+from __future__ import print_function, division, absolute_import
+
+import numpy as np
+
+__all__ = ('adupdates', 'adupdates_simple')
```
adler-j: The function `adupdates_simple` should not be exported.

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2014-2017 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Test for the Fessler alternating dual updates solver."""
```
adler-j: Since "Fessler" is not part of the ODL name, it should not be mentioned here either

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+
+from __future__ import division
+# import pytest
```
adler-j: Remove this line

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+import odl
+# import numpy as np
```
adler-j: Remove this line

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+    can be given in closed form. If we set
+
+    (a)   (25/12)                    (x_1) = (1)
+    (b)   (77/60)                    (x_2) = (1)
+    (c) = (19/20)         then       (x_3) = (1)
+    (d)   (319/420),                 (x_4) = (1).
+
+    We therefore solve the problem
+
+    min ||Ax - b||^2 + TV(x) s.t. x >= 0
```
adler-j: I was fine until this point. Why should this have the same solution as above? What does TV even mean here?

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+    mat1 = [[1, 1 / 2, 1 / 3, 1 / 4],
+            [1 / 2, 1 / 3, 1 / 4, 1 / 5]]
+    mat2 = [[1 / 3, 1 / 4, 1 / 5, 1 / 6],
+            [1 / 4, 1 / 5, 1 / 6, 1 / 7]]
+
+    # Create the linear operators
+    mat1op = odl.MatrixOperator(mat1)
+    mat2op = odl.MatrixOperator(mat2)
+    domain = mat1op.domain
+    tv1 = odl.MatrixOperator([[1.0, -1.0, 0.0, 0.0], [0.0, 0.0, 1.0, -1.0]])
```
adler-j: break line here for better readability

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+    nnegm = nneg.range.element([1.0, 1.0, 1.0, 1.0])
+    majs = [mat1m, mat2m, tv1m, tv2m, nnegm]
+
+    expected_solution = domain.element([1, 1, 1, 1])
+    # Create right-hand-sides of the equation
+    rhs1 = mat1op.range.element(mat1op(expected_solution))
+    rhs2 = mat2op.range.element(mat2op(expected_solution))
+
+    # Create the functionals
+    fid1 = odl.solvers.functional.functional.FunctionalTranslation(
```
adler-j: No need to use `odl.solvers.functional.functional.FunctionalTranslation`, simply use `.translated(...)` on the functional

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+    mat1m = mat1op(mat1op.adjoint(mat1op.range.one()))
+    mat2m = mat2op(mat2op.adjoint(mat2op.range.one()))
+    tv1m = tv1.range.element([2.0, 2.0])
+    tv2m = 2.0
+    nnegm = nneg.range.element([1.0, 1.0, 1.0, 1.0])
+    majs = [mat1m, mat2m, tv1m, tv2m, nnegm]
+
+    expected_solution = domain.element([1, 1, 1, 1])
+    # Create right-hand-sides of the equation
+    rhs1 = mat1op.range.element(mat1op(expected_solution))
```
adler-j: `mat1op.range.element` should not be needed here

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+    rhs2 = mat2op.range.element(mat2op(expected_solution))
+
+    # Create the functionals
+    fid1 = odl.solvers.functional.functional.FunctionalTranslation(
+        odl.solvers.functional.default_functionals.L2NormSquared(
+            mat1op.range), rhs1)
+    fid2 = odl.solvers.functional.functional.FunctionalTranslation(
+        odl.solvers.functional.default_functionals.L2NormSquared(
+            mat2op.range), rhs2)
+    reg1 = odl.solvers.functional.default_functionals.L1Norm(tv1.range)
```
adler-j: odl.solvers.L1Norm(tv1.range) should suffice

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,106 @@
...
+    reg2 = odl.solvers.functional.default_functionals.L1Norm(tv2.range)
+    ind = odl.solvers.functional.default_functionals.IndicatorNonnegativity(
+        nneg.range)
+    funcs = [fid1, fid2, reg1, reg2, ind]
+
+    # Start from zero
+    x = tv1.domain.zero()
+    x_simple = tv1.domain.zero()
+
+    # Select stepsize one
```
adler-j: This line is somewhat redundant, it follows from the variable name. Documentation should not document "obvious" stuff. In general, the rule is that documentation should be about "why" something is done. "What" is done should follow from the code.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
+# Copyright 2014-2017 The ODL contributors
```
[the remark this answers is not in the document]
kohr-h: I prefer one sweeping update for just that.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,210 @@
...
+__all__ = ('adupdates', 'adupdates_simple')
+
+
+def adupdates(x, g, L, stepsize, majs, niter, random=False,
+              callback=None, callback_loop='outer'):
+    """Alternating Dual updates method.
+
+    The Alternating Dual (AD) updates method of McGaffin and Fessler `[MF2015]
+    <http://ieeexplore.ieee.org/document/7271047/>`_ is designed to solve an
+    optimization problem of the form
```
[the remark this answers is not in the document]
kohr-h: We need something here, yes. If the text ends like this
```
text text text::
    code code code
```
the colon will be rendered by Sphinx. If the text should not end with a colon (like here), place a whitespace before the double colon, like this:
```
text text text ::
    code code code
```

odl/solvers/nonsmooth/adupdates.py
kohr-h: Remark here that applies to a lot of places in our docstrings:

If we use [raw string literals](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) whenever there's math in it, we can ditch all those double backslash ugliness.

Here, some math won't work (e.g. `\frac`), so the simplest fix is to just prepend the whole docstring with an `r`.

However, we should explore what happens with unicode stuff, in cases where names require non-ASCII characters.

odl/solvers/nonsmooth/adupdates.py
kohr-h: Small addition to this: it's more helpful for the user if the expected behavior is explicitly stated, e.g.,
```
'`len(majs)` should be equal `len(g)`, but {} != {}'
```
The library could definitely use a refreshment in that department, but new code should act as a good example 😄

odl/solvers/nonsmooth/adupdates.py
sbanert: But if I do not export this function, how can I use it in tests or for debugging?

adler-j: It is still available under the full path `odl.solvers.nonsmooth.adaupdates.adaupdates_simple`.

odl/solvers/nonsmooth/adupdates.py
sbanert: I tried this by building the documentation. `Functional.convex_conj` is a dead link, too.

odl/solvers/nonsmooth/adupdates.py
sbanert: Unicode seems to work fine at least in my environment with Python 3.6.

odl/solvers/nonsmooth/adupdates.py
kohr-h: Technically, the `__all__` thing is for defining explicitly what should be exposed to wildcard imports, i.e. `from odl.solvers.nonsmooth.adaupdates import *`. If it's undefined, the wildcard import just imports everything not starting with an underscore (roughly).
For us, defining stuff in `__all__` is also a mechanism to make things available on a higher level, since the `__init__.py` files on higher levels do the wildcard import for all submodules (with exceptions).

So it's always possible to import names by going all the way down to where the name is actually defined (= bound to a value).

odl/solvers/nonsmooth/adupdates.py
kohr-h: I've never really understood the rules as to when you can use the name without full path. @adler-j can you nail it down?

odl/solvers/nonsmooth/adupdates.py
kohr-h: "equal to" that is

odl/solvers/nonsmooth/adupdates.py
sbanert: Maybe it can’t find a destination. Links would have to go to entries of the attributes table in this file: https://odlgroup.github.io/odl/generated/odl.solvers.functional.functional.Functional.html

Actually, it would be nice to have the complete docstring of Functional.proximal and Functional.convex_conj printed somewhere in the docs.

sbanert:
I did everything except

* the dead links Functional.convex_conj and Functional.proximal
* upgrading to 2018.

odl/solvers/nonsmooth/adupdates.py
```diff
@@ -0,0 +1,217 @@
...
+The alternating dual upgrade method solves structured convex optimization
+problems by successively updating dual variables which are associated with
+each of the components.
+"""
+
+from __future__ import print_function, division, absolute_import
+
+import numpy as np
+
+__all__ = ('adupdates_method',)
```
adler-j: Why is it called "method" btw, wouldn't `adaupdates` look nicer?

odl/test/solvers/nonsmooth/adupdates_test.py
```diff
@@ -0,0 +1,102 @@
...
+    (c) = (19/20)         then       (x_3) = (1)
+    (d)   (319/420),                 (x_4) = (1).
+
+    We solve the problem
+
+    min ||Ax - b||^2 + TV(x) s.t. x >= 0
+
+    for the matrix A, the r.h.s. b as above and the total variation TV, which
+    is given as the sum of the absoulute values of consecutive entries of the
+    solution. The solution of this problem is clearly x = (1, 1, 1, 1), since
```
adler-j: Well this is assuming that we're using periodic boundary conditions for TV (which is not default in ODL), but I guess we can let that pass for now :O

sbanert: Not really, this total variation takes three differences for a four-component vector. This is still not the standard behaviour, though, because in ODL the absolute value of the last element would be added to the functional (which was very irritating to most of my students).

sbanert:
I fixed the issues and corrected a bug. Please do *not* merge, though, because this is a rebase on the branch of #1283 (otherwise #1279 would have been triggered).

sbanert:
Now, I rebased on master, but I think I will have to rebase once more because Travis will probably fail.

kohr-h:
Seems so, but after that it should all be fine.

aringh:
The fix for Travis doc build is now in master, so if you rebase again all test should hopefully pass :-)

adler-j:
Please perform that rebase!

sbanert:
Succeeded.

adler-j:
Merged, well done!

PR #1244, opened 2017-11-22
MAINT: attempt to fix pep8speaks config

kohr-h:
Possible that the heading comment in the `opened` section broke the config file parsing. Let's try.


kohr-h:
I'll merge this and see where it goes.

PR #1245, opened 2017-11-22
BUG: fix sign flip in 2D geometries

kohr-h:
Serious stuff. The detector axis had a sign flip in the first component, due to me misinterpreting the ASTRA axis convention. ~~The fix needs vec geometries, so until then parallel 2D is broken in that sense.~~

~~This is WIP, some tests fail.~~

Turned out that we only need to map `xmin, xmax -> -xmax, -xmin` in the first component of the *volume* geometry, I don't quite know why. The fix works with the release version of ASTRA. I added a test that catches this stuff.
I'll keep in the premature parallel_vec stuff anyway. If people want to play with a dev version of ASTRA, it will use that feature, and it seems to work fine.

### DIFF
--- a/odl/tomo/backends/astra_setup.py
+++ b/odl/tomo/backends/astra_setup.py
@@ -25,32 +25,35 @@ ODL geometry representation to ASTRA's data structures, including:
 
 from __future__ import print_function, division, absolute_import
 import numpy as np
+import warnings
 try:
     import astra
-    ASTRA_AVAILABLE = True
-    try:
-        # Available from 1.8 on
-        ASTRA_VERSION = astra.__version__
-        # 1.8.3 and later exposes also the patch version, need only first two
-        _maj, _min = [int(n) for n in ASTRA_VERSION.split('.')][:2]
-    except AttributeError:
-        # Below version 1.8
-        _maj = astra.astra.version() // 100
-        _min = astra.astra.version() % 100
-        ASTRA_VERSION = '.'.join([str(_maj), str(_min)])
-
-    # Don't import pkg_resources only for this, it's slow
-    if (_maj, _min) < (1, 7):
-        raise RuntimeError('ASTRA version < 1.7 not supported, please update')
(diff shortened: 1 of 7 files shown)


adler-j:
Could you separate the "urgent" stuff and we merge only that?

kohr-h:
Okay, parallel_vec stuff is in a separate branch now. This is only the bugfix (more or less). I did a wrong push that closed the PR, back to life now.

kohr-h:
Travis got confused as well, build is not failing.

kohr-h:
Well I don't expect more to come, but I'm honestly a bit puzzled by this. The error only occurs when the space is not centered, and only in 2D. For a shift along the x axis, the projection moves into the opposite direction. Kind of hard to spot.
I'm not sure the error is on our side, will check.

odl/tomo/backends/astra_setup.py
```diff
@@ -27,31 +27,32 @@
...
+if ASTRA_AVAILABLE:
+    try:
+        # Available from 1.8 on
+        ASTRA_VERSION = astra.__version__
+    except AttributeError:
+        # Below version 1.8
+        _maj = astra.astra.version() // 100
+        _min = astra.astra.version() % 100
+        if (_maj, _min) < (1, 7):
+            raise RuntimeError(
```
adler-j: Should be  a warning, we don't want peoples whole installation broken because one part is bad.

kohr-h: Well ok, I don't even know how to install that version of ASTRA.

adler-j: Well it's just a general philosophy thing I guess

kohr-h: Hm, apparently there is 1.6 in conda, so better do the warning...

kohr-h:
Ok, warning now instead of raising. Merging after CI.

PR #1255, opened 2017-12-09
BLD: make extra Travis build for the docs only

kohr-h:


.travis.yml
```diff
@@ -11,6 +11,9 @@ matrix:
         - python: 3.6
           env: NUMPY_VERSION=1.13 COVERALLS=true BUILD_DOCS=true
```
adler-j: No need to build them here then?

kohr-h: Of course not, :headbang:

kohr-h:
Just for the records, I let the new build always build the docs, then we see these kinds of issues earlier. Only the deploying step is left until the PR goes in.

PR #1258, opened 2017-12-14
DOC: Add link to odlworkshop in README.md

adler-j:


sbanert:
Maybe even a tag in the repository to a version which guarantees all the examples to work?

adler-j:
Preferably they should always work, that is what master is for :) Anything failing atm?

sbanert:
No, that is just for the long run. Since the examples are not automatically tested, they might break for future versions of master, and we wouldn't notice it.

PR #1265, opened 2018-01-13
MAINT: disable Numpy 1.14.0, closes #1264

### LINKED ISSUE #1264: New-style array printing of Numpy 1.14 breaks all doctests
Numpy 1.14 significantly changes the way arrays are printed, most notably making the string more compact by removing whitespace.

Well, tough for us since we support 1.9 and above. Fortunately there's a legacy switch that we can use as explained in the [release notes](https://docs.scipy.org/doc/numpy-1.14.0/release.html#many-changes-to-array-printing-disableable-with-the-new-legacy-printing-mode) to get the old behavior.

Well, not quite the old behavior. Turns out that [complex arrays are not printed correctly in legacy mode](https://github.com/numpy/numpy/issues/10356). So unfortunately we need to disable 1.14.0 for now. I'd expect a fix for that in the first patch release 1.14.1.

My suggestion is to only explicitly exclude 1.14.0 so we don't miss the update that fixes it (or not, in which case the issue is obvious and we can react).

kohr-h:


adler-j:
Does a printing error really warrant disabling it?

kohr-h:
> Does a printing error really warrant disabling it?

Well, since printing error means "lots of doctests fail", yes.

Issue #918, opened 2017-02-16
Bregman distances for functionals


mehrhardt:
It would be great if ODL was supporting Bregman distances. An easy way would be to define it as a function of x as (here without ODL):

```
given y and p
def BregmanG(x): return g(x) - g(y) - p.inner(x-y)
```
Is there a way to do this nicely in ODL? One would need to define a functional that does x -> p.inner(x-y).

adler-j:
Well, you can implement it as thus:

```python
>>> r3 = odl.rn(3)
>>> g = odl.solvers.L2Norm(r3)
>>> p = r3.element([1, 2, 3])
>>> y = r3.element([4, 5, 6])
>>> bregman = odl.solvers.FunctionalLinearPerturb(g, -p) - (g(y) + p.inner(y))
>>> bregman([5, 6, 7])
-68.286875905690607
```

which has all of the properties you are looking for i guess

```
>>> bregman.proximal(1)([5, 6, 7])
rn(3).element([5.5757359312880759, 7.4343145750507675, 9.2928932188134592])
>>> bregman.gradient([5, 6, 7])
rn(3).element([-0.52326870537720382, -1.4279224464526445, -2.3325761875280855])
>>> bregman.convex_conj([5, 6, 7])
inf
```

with that said, adding a `BregmanDistance` functional would make this more clear and less prone to misstakes.

aringh:
I have started working on Bregman functional on  my private fork some time ago. I just implemented and example running the iterations for Brgeman-TV, but did not have time to encapsulate it in a Bregman functional. If you are interested you could have a look at it: [the branch](https://github.com/aringh/odl/tree/bregman_iterations) and specifically [the commit](https://github.com/aringh/odl/commit/9c3643d76474d4728102d8b2f911c6f27e4d7453).

adler-j:
@aringh, any chance that you can make your implementation of the bregman functional public?

ozanoktem:
Somewhat off-topic, but I still want to mention the following publications that contain a nice overview of common functionals and their corresponding Bregman distances:
[http://www.jmlr.org/papers/volume6/banerjee05b/banerjee05b.pdf](http://www.jmlr.org/papers/volume6/banerjee05b/banerjee05b.pdf) (table 1 on p. 1709)
[...]

PR #1267, opened 2018-01-14
Issue 918  bregman functional

aringh:
Finally found some time to implement a Bregman distance functional, see #918.

- [X] Add Bregman distance functional.
- [X] Add it to the general test suit for functionals.
- [x] Add more specific tests, checking e.g. the use of a separate subgradient operator
- [ ] Add example of using Bregman functional, e.g. an example of Bregman-TV

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -25,7 +25,7 @@ from odl.solvers.nonsmooth.proximal_operators import (
     proximal_const_func, proximal_box_constraint,
     proximal_convex_conj_kl, proximal_convex_conj_kl_cross_entropy,
     combine_proximals, proximal_convex_conj)
-from odl.util import conj_exponent, moveaxis, signature_string, indent
+from odl.util import conj_exponent, moveaxis
 
 
 __all__ = ('ZeroFunctional', 'ConstantFunctional', 'ScalingFunctional',
@@ -37,7 +37,7 @@ __all__ = ('ZeroFunctional', 'ConstantFunctional', 'ScalingFunctional',
            'IndicatorNuclearNormUnitBall',
            'KullbackLeibler', 'KullbackLeiblerCrossEntropy',
            'QuadraticForm',
-           'SeparableSum', 'MoreauEnvelope', 'BregmanDistance')
+           'SeparableSum', 'MoreauEnvelope')
 
 
 class LpNorm(Functional):
@@ -2492,181 +2492,6 @@ class Huber(Functional):
         '''Return ``repr(self)``.'''
         return '{}({!r}, {!r})'.format(self.__class__.__name__, self.domain,
                                        self.gamma)
-
-
-class BregmanDistance(Functional):
-    """The Bregman distance functional.
-
-    Notes
-    -----
-    Given a functional :math:`f`, which has a (sub)gradient :math:`\partial f`,
-    and given a point :math:`y`, the Bregman distance functional
-    :math:`D_f(\cdot, y)` in a point :math:`x` is given by
-
-    .. math::
-        D_f(x, y) = f(x) - f(y) - \\langle \partial f(y), x - y \\rangle.
-
-
-    For mathematical details, see `[Bur2016]`_.
-
-    References
-    ----------
-    Wikipedia article: https://en.wikipedia.org/wiki/Bregman_divergence
-
-    [Bur2016] Burger, M. *Bregman Distances in Inverse Problems and Partial
-    Differential Equation*. In: Advances in Mathematical Modeling, Optimization
-    and Optimal Control, 2016. p. 3-33.
-
-    .. _[Bur2016]:  https://arxiv.org/abs/1505.05191
-    """
-
-    def __init__(self, functional, point, subgrad=None):
-        """Initialize a new instance.
-
-        Parameters
-        ----------
-        functional : `Functional`
-            Functional on which to base the Bregman distance. If the
-            optional argument `subgrad` is not given, the functional
-            needs to implement `functional.gradient`.
-        point : element of ``functional.domain``
-            Point from which to define the Bregman distance.
-        subgrad : `Operator`, optional
-            A subgradient operator. Operator which that takes an element in
-            `functional.domain` and returns a subdifferential of the functional
-            in that point.
-            Default: ``functional.gradient``.
-
-        Examples
-        --------
-        Example of initializing the Bregman distance functional:
-
-        >>> space = odl.uniform_discr(0, 1, 10)
-        >>> l2_squared = odl.solvers.L2NormSquared(space)
-        >>> point = space.one()
-        >>> Bregman_dist = odl.solvers.BregmanDistance(l2_squared, point)
-
-        This is gives squared L2 distance to the given point, ||x - 1||^2:
-
-        >>> expected_functional = L2NormSquared(space).translated(point)
-        >>> Bregman_dist(space.zero()) == expected_functional(space.zero())
-        True
-        """
-        if not isinstance(functional, Functional):
-            raise TypeError('`functional` {} not an instance of ``Functional``'
-                            ''.format(functional))
-        self.__functional = functional
-
-        if point not in functional.domain:
-            raise ValueError('`point` {} is not in `functional.domain` {}'
-                             ''.format(point, functional.domain))
-        self.__point = point
-
-        if subgrad is None:
-            # If the subgradient is not given, assign the gradient of the
-            # functional
-            try:
-                self.__subgrad = self.__functional.gradient
-            except NotImplementedError:
-                raise NotImplementedError(
-                    '`subgradient` not given, and the `functional` {} does '
-                    'not implement `functional.gradient`'.format(functional))
-        else:
-            # Check that given subgradient is an operator that maps from the
-            # domain of the functional to itself
-            if not isinstance(subgrad, Operator):
-                raise TypeError(
-                    '`subgrad` must be an instance of ``Operator``, got {}'
-                    ''.format(subgrad))
-            if subgrad.domain != self.__functional.domain:
-                raise ValueError('`functional.domain` {} is not the same as '
-                                 '`subgrad.domain` {}'
-                                 ''.format(self.__functional.domain,
-                                           subgrad.domain))
-            if subgrad.range != self.__functional.domain:
-                raise ValueError('`functional.domain` {} is not the same as '
-                                 '`subgrad.range` {}'
-                                 ''.format(self.__functional.domain,
-                                           subgrad.range))
-            self.__subgrad = subgrad
-
-        self.__initialized = False
-        super(BregmanDistance, self).__init__(space=functional.domain,
-                                              linear=False)
-
-    @property
-    def functional(self):
-        """The functional for the Bregman distance."""
-        return self.__functional
-
-    @property
-    def point(self):
-        """The point for the Bregman distance."""
-        return self.__point
-
-    @property
-    def subgrad(self):
-        """The subgradient operator for the Bregman distance."""
-        return self.__subgrad
-
-    @property
-    def initialized(self):
-        """Flag for if the Bregman distance functional is initialized."""
-        return self.__initialized
-
-    def _initialize(self):
-        """Helper function to initialize the functional."""
-
-        self.__subgrad_eval = self.subgrad(self.point)
-        self.__constant = (-self.functional(self.point) +
-                           self.subgrad_eval.inner(self.point))
-        self.__bregman_dist = FunctionalQuadraticPerturb(
-            self.functional, linear_term=-self.__subgrad_eval,
-            constant=self.__constant)
-
-        self.grad_lipschitz = (self.functional.grad_lipschitz +
-                               self.__subgrad_eval.norm())
-
-        self.__initialized = True
-
-    def _call(self, x):
-        """Return ``self(x)``."""
-        if not self.initialized:
-            self._initialize()
-
-        return self.__bregman_dist(x)
-
-    @property
-    def convex_conj(self):
-        """The convex conjugate"""
-        if not self.initialized:
-            self._initialize()
-
-        return self.__bregman_dist.convex_conj
-
-    @property
-    def proximal(self):
-        """Return the ``proximal factory`` of the functional."""
-        if not self.initialized:
-            self._initialize()
-
-        return self.__bregman_dist.proximal
-
-    @property
-    def gradient(self):
-        """Gradient operator of the functional."""
-        if not self.initialized:
-            self._initialize()
-
-        return self.subgrad - ConstantOperator(self.__subgrad_eval)
-
-    def __repr__(self):
-        '''Return ``repr(self)``.'''
-        posargs = [self.functional, self.point]
-        subgrad_op_default = getattr(self.functional, 'gradient', None)
-        optargs = [('subgrad', self.subgrad, subgrad_op_default)]
-        inner_str = signature_string(posargs, optargs, sep=',\n')
-        return '{}(\n{}\n)'.format(self.__class__.__name__, indent(inner_str))
 
 
 if __name__ == '__main__':
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -2587,9 +2587,11 @@ class BregmanDistance(Functional):
             self.__subgradient_op = subgradient_op
 
         self.__subgrad_eval = self.__subgradient_op(self.__point)
-        self.__bregman_dist = (FunctionalQuadraticPerturb(
-            self.__functional, linear_term=-self.__subgrad_eval) -
-            self.__functional(point) + self.__subgrad_eval.inner(point))
+        self.__constant = (-self.__functional(point) +
+                           self.__subgrad_eval.inner(point))
+        self.__bregman_dist = FunctionalQuadraticPerturb(
+            self.__functional, linear_term=-self.__subgrad_eval,
+            constant_coeff=self.__constant)
 
         super(BregmanDistance, self).__init__(
             space=functional.domain, linear=False,
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -2575,14 +2575,14 @@ class BregmanDistance(Functional):
                                 '``Operator``'.format(subgradient_op))
             if not self.__functional.domain == subgradient_op.domain:
                 raise ValueError('`functional.domain` {} is not the same as '
-                                 '`subgradient_op.domain` {}'.format(
-                                        self.__functional.domain,
-                                        subgradient_op.domain))
+                                 '`subgradient_op.domain` {}'
+                                 ''.format(self.__functional.domain,
+                                           subgradient_op.domain))
             if not self.__functional.domain == subgradient_op.range:
                 raise ValueError('`functional.domain` {} is not the same as '
-                                 '`subgradient_op.range` {}'.format(
-                                        self.__functional.domain,
-                                        subgradient_op.range))
+                                 '`subgradient_op.range` {}'
+                                 ''.format(self.__functional.domain,
+                                           subgradient_op.range))
             self.__subgradient_op = subgradient_op
 
         self.__subgrad_eval = self.__subgradient_op(self.__point)
@@ -2626,7 +2626,7 @@ class BregmanDistance(Functional):
     @property
     def gradient(self):
         """Gradient operator of the functional."""
-        return self.__bregman_dist.gradient
+        return self.subgradient_op - ConstantOperator(self.__subgrad_eval)
 
     def __repr__(self):
         '''Return ``repr(self)``.'''
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -2548,11 +2548,13 @@ class BregmanDistance(Functional):
         2.0
         """
         if not isinstance(functional, Functional):
-            raise TypeError('...')
+            raise TypeError('`functional` {} not an instance of ``Functional``'
+                            ''.format(functional))
         self.__functional = functional
 
         if point not in functional.domain:
-            raise TypeError('...')
+            raise ValueError('`point` {} is not in `functional.domain` {}'
+                             ''.format(point, functional.domain))
         self.__point = point
 
         if subgradient_op is None:
@@ -2569,11 +2571,18 @@ class BregmanDistance(Functional):
             # Check that given subgradient is an operator that maps from the
             # domain of the functional to the domain of the functional
             if not isinstance(subgradient_op, Operator):
-                raise TypeError('...')
+                raise TypeError('`subgradient_op` {} is not an instance of '
+                                '``Operator``'.format(subgradient_op))
             if not self.__functional.domain == subgradient_op.domain:
-                raise TypeError('...')
+                raise ValueError('`functional.domain` {} is not the same as '
+                                 '`subgradient_op.domain` {}'.format(
+                                        self.__functional.domain,
+                                        subgradient_op.domain))
             if not self.__functional.domain == subgradient_op.range:
-                raise TypeError('...')
+                raise ValueError('`functional.domain` {} is not the same as '
+                                 '`subgradient_op.range` {}'.format(
+                                        self.__functional.domain,
+                                        subgradient_op.range))
             self.__subgradient_op = subgradient_op
 
         self.__subgrad_eval = self.__subgradient_op(self.__point)
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -25,7 +25,7 @@ from odl.solvers.nonsmooth.proximal_operators import (
     proximal_const_func, proximal_box_constraint,
     proximal_convex_conj_kl, proximal_convex_conj_kl_cross_entropy,
     combine_proximals, proximal_convex_conj)
-from odl.util import conj_exponent, moveaxis
+from odl.util import conj_exponent, moveaxis, signature_string, indent
 
 
 __all__ = ('ZeroFunctional', 'ConstantFunctional', 'ScalingFunctional',
@@ -2574,12 +2574,12 @@ class BregmanDistance(Functional):
             if not isinstance(subgradient_op, Operator):
                 raise TypeError('`subgradient_op` {} is not an instance of '
                                 '``Operator``'.format(subgradient_op))
-            if not subgradient_op.domain == self.__functional.domain:
+            if subgradient_op.domain != self.__functional.domain:
                 raise ValueError('`functional.domain` {} is not the same as '
                                  '`subgradient_op.domain` {}'
                                  ''.format(self.__functional.domain,
                                            subgradient_op.domain))
-            if not subgradient_op.range == self.__functional.domain:
+            if subgradient_op.range != self.__functional.domain:
                 raise ValueError('`functional.domain` {} is not the same as '
                                  '`subgradient_op.range` {}'
                                  ''.format(self.__functional.domain,
@@ -2634,9 +2634,11 @@ class BregmanDistance(Functional):
 
     def __repr__(self):
         '''Return ``repr(self)``.'''
-        return '{}({!r}, {!r}, {!r})'.format(self.__class__.__name__,
-                                             self.functional, self.point,
-                                             self.subgradient_op)
+        posargs = [self.functional, self.point]
+        subgrad_op_default = getattr(self.functional, 'gradient', None)
+        optargs = [('subgradient_op', self.subgradient_op, subgrad_op_default)]
+        inner_str = signature_string(posargs, optargs, sep=',\n')
+        return '{}(\n{}\n)'.format(self.__class__.__name__, indent(inner_str))
 
 
 if __name__ == '__main__':
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -2504,8 +2504,10 @@ class BregmanDistance(Functional):
     :math:`D_f(\cdot, y)` in a point :math:`x` is given by
 
     .. math::
-        D_f(x, y) = f(x) - f(y) - \langle \partial f(y), x - y \\rangle.
+        D_f(x, y) = f(x) - f(y) - \\langle \partial f(y), x - y \\rangle.
 
+
+    For mathematical details, see `[Bur2016]`_.
 
     References
     ----------
@@ -2513,24 +2515,26 @@ class BregmanDistance(Functional):
 
     [Bur2016] Burger, M. *Bregman Distances in Inverse Problems and Partial
     Differential Equation*. In: Advances in Mathematical Modeling, Optimization
-    and Optimal Control, 2016. p. 3-33. Arxiv version online at
-    https://arxiv.org/abs/1505.05191
+    and Optimal Control, 2016. p. 3-33.
+
+    .. _[Bur2016]:  https://arxiv.org/abs/1505.05191
     """
 
-    def __init__(self, functional, point, subgradient_op=None):
+    def __init__(self, functional, point, subgrad=None):
         """Initialize a new instance.
 
         Parameters
         ----------
         functional : `Functional`
-            The functional on which to base the Bregman distance. If the
-            optional argument `subgradient_op` is not given, the functional
+            Functional on which to base the Bregman distance. If the
+            optional argument `subgrad` is not given, the functional
             needs to implement `functional.gradient`.
         point : element of ``functional.domain``
-            The point from which to define the Bregman distance.
-        subgradient_op : `Operator`, optional
-            The operator that takes an element in `functional.domain` and
-            returns a subgradient of the functional in that point.
+            Point from which to define the Bregman distance.
+        subgrad : `Operator`, optional
+            A subgradient operator. Operator which that takes an element in
+            `functional.domain` and returns a subdifferential of the functional
+            in that point.
             Default: ``functional.gradient``.
 
         Examples
@@ -2542,10 +2546,10 @@ class BregmanDistance(Functional):
         >>> point = space.one()
         >>> Bregman_dist = odl.solvers.BregmanDistance(l2_squared, point)
 
-        This is gives the shifted L2 norm squared ||x - 1||^2:
+        This is gives squared L2 distance to the given point, ||x - 1||^2:
 
-        >>> expected_value = l2_squared(space.one())
-        >>> Bregman_dist(space.zero()) == expected_value
+        >>> expected_functional = L2NormSquared(space).translated(point)
+        >>> Bregman_dist(space.zero()) == expected_functional(space.zero())
         True
         """
         if not isinstance(functional, Functional):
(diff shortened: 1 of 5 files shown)


aringh:
This is good to go for review. An example can come later, it doesn't even have to be in this PR.

In #918, @ozanoktem suggested having something like `functional.bregman`. Is this something we want to pursue?

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+    """The Bregman distance functional.
+
+    Notes
+    -----
+    Given a functional :math:`f`, which has a (sub)gradient :math:`\partial f`,
+    and given a point :math:`y`, the Bregman distance functional :math:`D_f`
+    in a point :math:`x` is given by
+
+    .. math::
+        D_f(x) = f(x) - f(y) - \langle \partial f(y), x - y \\rangle.
```
adler-j: Missing y on L.H.S.?

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        functional : `Functional`
+            The functional on which to base the Bregman distance. If the
+            optional argument `subgradient_op` is not given, the functional
+            needs to implement `functional.gradient`.
+        point : element of ``functional.domain``
+            The point from which to define the Bregman distance
```
adler-j: Missing .

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+        subgradient_op : `Operator`, optional
+            The operator that takes an element in `functional.domain` and
+            returns a subgradient of the functional in that point.
+            Default: ``functional.gradient``.
+
+        Examples
+        --------
+        Example of initializing the Bregman distance functional:
+
+        >>> space = odl.uniform_discr(0, 2, 14)
```
adler-j: 14 is rather arbitrary, prefer 10 or so

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+        >>> l2_squared = odl.solvers.L2NormSquared(space)
+        >>> point = space.one()
+        >>> Bregman_dist = odl.solvers.BregmanDistance(l2_squared, point)
+
+        This is gives the shifted L2 norm squared ||x - 1||:
```
adler-j: missing ^2

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+            try:
+                self.__subgradient_op = self.__functional.gradient
+            except NotImplementedError:
+                raise NotImplementedError('`subgradient` not given, and the '
+                                          '`functional` {} does not implement '
+                                          '`functional.gradient`'
+                                          ''.format(functional))
+        else:
+            # Check that given subgradient is an operator that maps from the
+            # domain of the functional to the domain of the functional
```
adler-j: to itself

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+            if not isinstance(subgradient_op, Operator):
+                raise TypeError('`subgradient_op` {} is not an instance of '
+                                '``Operator``'.format(subgradient_op))
+            if not self.__functional.domain == subgradient_op.domain:
```
adler-j: prefer

```
if (thing to be tested) == (expected value)
```

it more closely follows human language so it's easier to read

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+                                           subgradient_op.domain))
+            if not self.__functional.domain == subgradient_op.range:
+                raise ValueError('`functional.domain` {} is not the same as '
+                                 '`subgradient_op.range` {}'
+                                 ''.format(self.__functional.domain,
+                                           subgradient_op.range))
+            self.__subgradient_op = subgradient_op
+
+        self.__subgrad_eval = self.__subgradient_op(self.__point)
+        self.__bregman_dist = (FunctionalQuadraticPerturb(
```
adler-j: FunctionalQuadraticPerturb should also support a constant term, so no need to add that "outside" so to say

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+            self.__functional, linear_term=-self.__subgrad_eval) -
+            self.__functional(point) + self.__subgrad_eval.inner(point))
+
+        super(BregmanDistance, self).__init__(
+            space=functional.domain, linear=False,
+            grad_lipschitz=self.__functional.grad_lipschitz)
```
adler-j: Is this really correct? Shouldn't the linear term add something

odl/solvers/functional/default_functionals.py
```diff
@@ -2470,6 +2470,149 @@ def __repr__(self):
...
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional."""
+        return self.subgradient_op - ConstantOperator(self.__subgrad_eval)
+
+    def __repr__(self):
+        '''Return ``repr(self)``.'''
+        return '{}({!r}, {!r}, {!r}, {!r})'.format(self.__class__.__name__,
+                                                   self.domain,
```
adler-j: This argument is not in __init__

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,61 @@ def test_moreau_envelope_l2_sq(space, sigma):
                             x * 2 / (1 + 2 * sigma))
 
 
+def test_bregman_functional_no_gradient(space):
+    """Test that the Bregman distance functional fails if the underlying
+    functional does not have a gradient and no subgradient operator is
+    given."""
```
adler-j: Move """ to newline

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,61 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+    # In this case we should be able to call the gradient of the bregman
+    # distance, which would give us a subgradient
+    x = np.abs(noise_element(space))
+    expected_result = subgrad_op(x) - subgrad_op(point)
+    assert all_almost_equal(bregman_dist.gradient(x), expected_result)
+
+
+def test_bregman_functional_l2_squared(space, sigma):
+    """Test for the Bregman distance functional, using l2 norm squared as
+    underlying functional."""
```
adler-j: """ on newline

odl/test/solvers/functional/default_functionals_test.py
adler-j: Also first line should be "sufficient" and fit in one line, the rest should elaborate

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,61 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+
+    expected_func = odl.solvers.L2NormSquared(space).translated(point)
+
+    x = noise_element(space)
+
+    # Function evaluation
+    assert all_almost_equal(bregman_dist(x), expected_func(x))
+
+    # Gradient evaluation
+    assert all_almost_equal(bregman_dist(x), expected_func(x))
```
adler-j: Copy-paste-error here? No gradient

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -516,5 +516,61 @@ def test_moreau_envelope_l2_sq(space, sigma):
...
+
+
+def test_bregman_functional_l2_squared(space, sigma):
+    """Test for the Bregman distance functional, using l2 norm squared as
+    underlying functional."""
+    sigma = float(sigma)
+
+    l2_sq = odl.solvers.L2NormSquared(space)
+    point = noise_element(space)
+    subgrad_op = odl.ScalingOperator(space, 2.0)
```
adler-j: Why does this need to be provided?

odl/solvers/functional/default_functionals.py
aringh: It works with 8, 9, 11, 12, 13, and 14 points... but with 10 points apparently `Bregman_dist(space.zero())` returns 1.9999999999999998 and therefore fails... is there a way to make "almost equal" in doc-tests?

odl/solvers/functional/default_functionals.py
aringh: Currently it seems not to actually:
```python
class FunctionalQuadraticPerturb(Functional):

    """The functional representing ``F(.) + a * <., .> + <., u>``."""

    def __init__(self, func, quadratic_coeff=0, linear_term=None):
```
But maybe we should update this? If so, in this PR or save if for a separate PR?

odl/solvers/functional/default_functionals.py
aringh: You are absolutely right! But `grad D_f(x, y) = grad f(x) - grad f(y)`, so it is to me not obvious how much it adds... however by the triangle inequality it should be bounded by the `grad_lipschitz` + the norm of `grad f(y)`, right? Maybe that is ok as an estimate?

odl/test/solvers/functional/default_functionals_test.py
aringh: It does not need to be provided. It is a relic since I wrote this test before the other test, and wanted to check that it in fact return the expected gradient also when a subdifferential operator is given. But given the test above that now tests this, I will remove it.

odl/solvers/functional/default_functionals.py
adler-j: Tbh update it here

odl/solvers/functional/default_functionals.py
kohr-h: And generally, it is preferable to write `(something) != (something else)` instead of `not (something) == (something else)`. For example pylint will complain about this.

odl/solvers/functional/default_functionals.py
kohr-h: It's more maintainable to use the helper functions for this:
```python
posargs = [functional, point]
subgrad_op_default = getattr(self.functional, 'gradient', None)
optargs = [('subgradient_op', self.subgradient_op, subgrad_op_default)]
inner_str = signature_string(posargs, optargs, sep=',\n')
return '{}(\n{}\n)'.format(self.__class__.__name__, indent(inner_str))
```
The utilities `signature_string` and `indent` are in `odl.util`.
(BTW, I'm working on making this MUCH better).

aringh:
Question: do we want to be able to update the point of the Bregman distance? In iterative reconstructions using the Bregman distance, as far as I know one normally updates this point between iterations. This could of course be achieved by simply creating a new functional each time one needs to update the point, but it should also be possible to do something like
```python
    @point.setter
    def point(self, new_point):
        self.__point = new_point
        self.__subgrad_eval = self.subgradient_op(self.__point)
        self.__constant = (-self.__functional(self.__point) +
                           self.__subgrad_eval.inner(self.__point))
        self.__bregman_dist = FunctionalQuadraticPerturb(
            self.__functional, linear_term=-self.__subgrad_eval,
            constant_coeff=self.__constant)
```
Is this to much, or do we want to go for that?

Apart from this, I think that the other issues have been addressed in the last commits.

aringh:
I think that the overhead is approximately the same between this way and if the user creates a completely new Bregman distance functional. At least with this code since I create a new ``FunctionalQuadraticPerturb`` when the point is updated. For me it is more a question of user-friendliness vs. to much built-in intelligence. I think it looks nice from a user-persective, but I would like to hear what @kohr-h and @adler-j has to say about it.

A "counter example", maybe advocating that we should not do this, is that we for example create a new proximal operator each time the step size ``sigma`` is updated.

ozanoktem:
After thinking some on how ODL should handle notion of Bregman distance, and despite the arguments provided earlier by @adler-j, I think the best option would be to introduce it as a property to the functional. An advantage is that the Bregman distance respects functional calculus much like the derivative. Ideally, if a developer does not explicitly write a routine for computing the Bregman distance for a given functional, then ODL should automatically resort to the definition for computing it.

odl/solvers/functional/default_functionals.py
```diff
@@ -2494,6 +2494,153 @@ def __repr__(self):
...
+    """The Bregman distance functional.
+
+    Notes
+    -----
+    Given a functional :math:`f`, which has a (sub)gradient :math:`\partial f`,
+    and given a point :math:`y`, the Bregman distance functional
+    :math:`D_f(\cdot, y)` in a point :math:`x` is given by
+
+    .. math::
+        D_f(x, y) = f(x) - f(y) - \langle \partial f(y), x - y \\rangle.
```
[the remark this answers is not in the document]
kohr-h: See my comment on #1243: If you prefix the docstring with an `r` (for "raw string") you can write normal LaTeX without using double-backslashes all the time.

aringh:
Then I will just move the ``BregmanDistance`` class to ``functional.py``, not expose it in ``__all__``, and add a ``bregman``-property to the ``Functional`` class (that works much like ``proximal``). Sounds good?

Also: should I leave the option of supplying a subgradient?

kohr-h:
> Also: should I leave the option of supplying a subgradient?

That's a very good question. One way to handle that would be to do it in a similar way as in `derivative`, where you supply the point and get back an operator for that point. For such a construction, it would be easy to add an option for a custom subgradient.
If we do it the `proximal` way, i.e., as a property that returns a factory function, it's also possible if that function accepts an optional `subgrad` argument.

I actually prefer the more lightweight solution (the `derivative` way), since I think in hindsight that the proximal factory stuff is too complicated for what it does. We could actually have something like in mathematics: `proximal` is the prox operator for step size `1.0`, and the user (or optimization method) simply does `tau * func.proximal`. We know how to construct that guy as a `proximal_scaling`.

> Sounds good?

To me it does, but maybe we should have more votes on this. I'm thinking of @adler-j and @mehrhardt in particular.

adler-j:
> That's a very good question. One way to handle that would be to do it in a similar way as in derivative, where you supply the point and get back an operator for that point. For such a construction, it would be easy to add an option for a custom subgradient.
If we do it the proximal way, i.e., as a property that returns a factory function, it's also possible if that function accepts an optional subgrad argument.

Lets go for the `functional.bregman` implementation, people seem to be in favor of it and it looks more clean.

> I actually prefer the more lightweight solution (the derivative way), since I think in hindsight that the proximal factory stuff is too complicated for what it does. We could actually have something like in mathematics: proximal is the prox operator for step size 1.0, and the user (or optimization method) simply does tau * func.proximal. We know how to construct that guy as a proximal_scaling.

I actually tried implementing this at some point, wasting quite a bit of time, and it turns out that it is non-trivial to do it as you say. Basically this line:

https://github.com/odlgroup/odl/blob/fa6e8625e9a79ac8c77a5a388b2923bf3a05de15/odl/solvers/functional/functional.py#L490

is what screws with us. In this one, `FunctionalLeftScalarMult.proximal` calls `Functional.proximal` but with a scaled argument. Basically one cannot compute the proximal of `a * F` in closed form given only `a` and the proximal of `F`.

> To me it does, but maybe we should have more votes on this. I'm thinking of @adler-j and @mehrhardt in particular.

Go ahead

kohr-h:
> Basically one cannot compute the proximal of a * F in closed form given only a and the proximal of F.

Ah crap, you need the prox of the right-scaled functional which you cannot derive from the prox with scalar 1. Alright, never mind.

aringh:
I have updated the PR. Good to go for a new review round :-)

I also tried to add an example for Bregman-TV in a separate branch [here](https://github.com/aringh/odl/pull/2). If we want such an example, I would appreciate if someone could take a look at it. Maybe @ozanoktem that knows how it is supposed to work?

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
         return '{}.convex_conj'.format(self.convex_conj)
 
 
+class BregmanDistance(Functional):
```
adler-j: Wouldn't "`BregmanDivergence`" be a more correct term?

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+    """The Bregman distance functional.
```
adler-j: Perhaps a few non-technical words on what this does below this would help, e.g. (copy-paste from wikipedia) "The Bregman divergence or Bregman distance is similar to a metric, but satisfies neither the triangle inequality nor symmetry."

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+        Example of initializing the Bregman distance functional:
+
+        >>> space = odl.uniform_discr(0, 1, 10)
+        >>> l2_squared = odl.solvers.L2NormSquared(space)
+        >>> point = space.one()
+        >>> Bregman_dist = odl.solvers.BregmanDistance(l2_squared, point)
+
+        This is gives squared L2 distance to the given point, ||x - 1||^2:
+
+        >>> expected_functional = odl.solvers.L2NormSquared(space).translated(
```
adler-j: `expected_functional = l2_squared.translated(...`

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+                raise TypeError(
+                    '`subgrad` must be an instance of ``Operator``, got {}'
+                    ''.format(subgrad))
+            if subgrad.domain != self.__functional.domain:
+                raise ValueError('`functional.domain` {} is not the same as '
+                                 '`subgrad.domain` {}'
+                                 ''.format(self.__functional.domain,
+                                           subgrad.domain))
+            if subgrad.range != self.__functional.domain:
+                raise ValueError('`functional.domain` {} is not the same as '
```
adler-j: I recommend changing the order of this exception text to follow the if statement, easier to read.

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+                                 '`subgrad.range` {}'
+                                 ''.format(self.__functional.domain,
+                                           subgrad.range))
+            self.__subgrad = subgrad
+
+        self.__initialized = False
+        self.__subgrad_eval = None
```
adler-j: I don't think there is any reason to set these to None?

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+        return self.__subgrad
+
+    @property
+    def initialized(self):
+        """Flag for if the Bregman distance functional is initialized."""
+        return self.__initialized
+
+    def _initialize(self):
+        """Helper function to initialize the functional."""
+
```
adler-j: Remove extra blankline

adler-j: Also there is a very nice pattern that you can use to save some lines. Rename this to `initialize_if_needed` and then move the `if not self.initialized` line inside here. That way you can save some duplication.

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+        """Gradient operator of the functional."""
+        if not self.initialized:
+            self._initialize()
+
+        return self.subgrad - ConstantOperator(self.__subgrad_eval)
+
+    def __repr__(self):
+        '''Return ``repr(self)``.'''
+        posargs = [self.functional, self.point]
+        subgrad_op_default = getattr(self.functional, 'gradient', None)
```
adler-j: Functional always implements `gradient` (since it is an `Functional`), so the `getattr` here is redundant

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -614,5 +614,73 @@ def test_weighted_proximal_L1_norm_close(space):
...
+
+    Test that the Bregman distance functional fails if the underlying
+    functional does not have a gradient and no subgradient operator is
+    given. Also test giving the subgradient operator separately.
+    """
+
+    ind_func = odl.solvers.IndicatorNonnegativity(space)
+    point = noise_element(space)
+
+    # Indicator function has no gradient, hence one cannot create a bregman
```
adler-j: Frankly we should just add gradient=0 for the indicator function. It makes sense, no?

This would be for later though

odl/test/solvers/functional/default_functionals_test.py
```diff
@@ -614,5 +614,73 @@ def test_weighted_proximal_L1_norm_close(space):
...
+    # distance functional
+    with pytest.raises(NotImplementedError):
+        odl.solvers.BregmanDistance(ind_func, point)
```
adler-j: We need to do tests for both `functiona.bregman` and the pure functional `odl.solvers.BregmanDistance`, at least one that verifies that it points to the later I'd say.

adler-j:
W.r.t. the bregman TV example, I think it looks quite good but I'd make a separate PR after this one for it. Overall, if it works I'd like it in!

odl/solvers/functional/functional.py
```diff
@@ -1025,36 +1081,45 @@ def convex_conj(self):
...
         else:
             return super(FunctionalQuadraticPerturb, self).convex_conj
 
     def __repr__(self):
         """Return ``repr(self)``."""
-        return '{}({!r}, {!r}, {!r})'.format(self.__class__.__name__,
-                                             self.functional,
-                                             self.quadratic_coeff,
-                                             self.linear_term)
+        return '{}({!r}, {!r}, {!r}, {!r})'.format(self.__class__.__name__,
```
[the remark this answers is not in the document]
kohr-h: Nah, don't bother for now. I'm refurbishing this stuff anyway in #1276

odl/solvers/functional/functional.py
kohr-h: I had the same thought initially, but the main reference (Burger et al (?)) calls it Bregman distance. Divergence leans more on the statistical side I assume, so IMO let's go with the terms the inverse problems and imaging people use.

odl/solvers/functional/functional.py
```diff
@@ -225,6 +225,48 @@ def translated(self, shift):
...
         return FunctionalTranslation(self, shift)
 
+    def bregman(self, point, subgrad=None):
+        """Return the Bregman distance functional.
+
+        Parameters
+        ----------
+        point : element of ``functional.domain``
+            Point from which to define the Bregman distance.
+        subgrad : `Operator`, optional
```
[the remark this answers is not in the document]
aringh: We can go for gradient, no problem with me :-)

odl/test/solvers/functional/default_functionals_test.py
aringh: The indicator function on a set has gradient=0 on that set. But outside the set, where it takes value `inf`, I think that the subgradient is empty. So we could return `EmptySet` in these cases, but I don't know if that would make anyone happier or if it would just confuse people...?

odl/test/solvers/functional/default_functionals_test.py
aringh: My idea, when keeping these test after implementing the `functional.bregman`, was that they still test the `BregmanDistance` class. Not for all functionals, but  some kind of test for the class itself.

I also added a test for `functional.bregman` to the suite of tests for functionals, [here](https://github.com/odlgroup/odl/pull/1267/files/3f2a8c2dfb158a059a214858b3d5ecce8bb640a9#diff-18b72305369f0cf5b0f7408c5211a60dR581). My idea was that this should cover all available functionals and see that there `.bregman` works as expected. By that we also cover the case if one does a special optimized implementation of the Bregman distance for some functionals later (e.g., the squared L2 norm should just be a translated functional).

Is that good enough, or would you like to add some other tests or change some of the current ones?

odl/solvers/functional/functional.py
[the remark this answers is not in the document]
aringh: Actually, trying to go through and making this change it turns out that we would probably like to distinguish it from the `gradient`. Because this is the (sub)gradient of the "underlying" functional. Meanwhile, the `BregmanDistance` is going to have a `gradient` of its own (which is a translated version of what is now called `subgrad`, see [line 1505](https://github.com/odlgroup/odl/pull/1267/files/3f2a8c2dfb158a059a214858b3d5ecce8bb640a9#diff-4086f82398c755f2cf1063ced9ec1c93R1505)), and we don't want to confusing these two.

Any other naming suggestions? `gradient_op`? Or stick with `subgrad`?

odl/solvers/functional/functional.py
```diff
@@ -1268,6 +1333,186 @@ def __str__(self):
...
+        return self.__functional
+
+    @property
+    def point(self):
+        """The point for the Bregman distance."""
+        return self.__point
+
+    @property
+    def subgrad(self):
+        """The subgradient operator for the Bregman distance."""
```
aringh: Related to a comment on the (sub)gradient of the underlying functional and the gradient of the Bregman distance: from this description it is not obvious what is meant. It is suppose to be the gradient of the underlying functional, Suggestion: change to `The ... used in the Bregman distance` for all three above. Ok?

odl/solvers/functional/functional.py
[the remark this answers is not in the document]
kohr-h: I'm not in favor of changing `subgrad` to `gradient` anyway, it's less precise IMO. To make clear that it's a property of the underlying functional, we could call it `func_subgrad` or so.

odl/test/solvers/functional/default_functionals_test.py
kohr-h: I think what you're talking about @aringh is the sub*differential*, not sub*gradient*. The subgradient is not a set, it's an (arbitrary) element from the subdifferential. So rather than returning an empty set, the `subgrad` operator should in principle fail if we're outside the set of the indicator function, since logically, we're trying to return an element from the empty set.

I don't know if that trouble is worth going through. Is there a case where we would need this?

kohr-h: > Frankly we should just add gradient=0 for the indicator function. It makes sense, no?

I think this is mathematically wrong since 0 is not a subgradient in points outside the set of the indicator function. So IMO we shouldn't do this.

odl/solvers/functional/functional.py
kohr-h: Maybe "used to define the Bregman distance"? Generally a good suggestion.

odl/solvers/functional/functional.py
aringh: I do not think I fully understand how `signature_string` works... 

But even the current implementation can get into trouble. If I uses, e.g., an `IndicatorBox` which I supplied with an appropriate subgradient operator in the creation of the `BregmanDistance` as follows 

```python
>>> import odl
>>> space = odl.uniform_discr(0,1,10)
>>> indic_func = odl.solvers.IndicatorBox(space, 0.2, 0.8)
>>> point = 0.5 * space.one()
>>> subgrad = odl.ZeroOperator(space)
>>> breg = odl.solvers.BregmanDistance(indic_func, point, subgrad)
>>> breg.__repr__()
```

 I get the following output

```

  File "<ipython-input-1-dcb69fc015ba>", line 7, in <module>
    breg.__repr__()

  File "[...]/odl/odl/solvers/functional/functional.py", line 1503, in __repr__
    subgrad_op_default = getattr(self.functional, 'gradient', None)

  File "[...]/odl/odl/solvers/functional/functional.py", line 97, in gradient
    ''.format(self))

NotImplementedError: no gradient implemented for functional IndicatorBox(uniform_discr(0.0, 1.0, 10), 0.2, 0.8)
```

However `self.func_subgrad` will, by the way that `__init__` works, always have a value. Either it is supplied separately or it is set to be the gradient. Thus one could just change to

```python
        optargs = [('func_subgrad', self.func_subgrad, None)]
```

where `None` will never be used, right?

aringh:
The question about tests is not addressed, but for that I would need some more exact pointers on what needs to be changed. Otherwise, good to go in?

adler-j:
> The question about tests is not addressed, but for that I would need some more exact pointers on what needs to be changed. Otherwise, good to go in?

Excuse my bad memory, but what pointers would you like here?

aringh:
You made a comment about the tests just above
> We need to do tests for both `functiona.bregman` and the pure functional `odl.solvers.BregmanDistance`, at least one that verifies that it points to the later I'd say.

Summarizing my answer from above: we have two tests for `BregmanDistance` in `default_functionals_test.py`, and for all functionals in the test suite in `functional_test.py` we test the `.bregman` functionality. I just wander if that is sufficient, or if you want changes in the test code? If so, what kind of changes?

odl/solvers/functional/functional.py
kohr-h: This solution should be fine. It will make `repr` always print out the `func_subgrad` even if the user didn't specify it, but the logic is complex enough that you may want to skip it.
If you don't want this behavior, you probably need to do the `gettattr` thing to get the default and use that instead of `None` in `optargs`.

aringh:
I will poke this one: is it good to go in? Other changes?

Also: @adler-j or @kohr-h will have to explicitly merge it.

adler-j:
Could use a rebase on master, other than that I'll have a quick look!

aringh:
Rebased!

adler-j:
Merge after tests! Well done!

aringh:
Same here - doc build seems to fail. Is there a bug in Travis, or am I doing something wrong in the code?

adler-j:
I can't see whats wrong. @kohr-h?

kohr-h:
> I can't see whats wrong. @kohr-h?

No idea either. Can you remove the `2>/dev/null` in `.travis.yml` where the docs are built? Just push it and remove the commit later when we're done. We need some output to see what's wrong.

kohr-h:
I suspect that it has to do with the major version change of `travis-sphinx` from `1.x` to `2.x`. Perhaps they dropped some option or whatever.

aringh:
Let's see what happens in this build.

kohr-h:
It fails to import `build_main` from `sphinx`. I checked, it's a Sphinx 1.7 feature, so `travis-sphinx` has an implicit dependency on that.

aringh:
Here is the output from the failing doc-build:
```
[...]
ufuncs                    : generated odl.util.ufuncs.rst
utility                   : generated odl.util.utility.rst
vectorization             : generated odl.util.vectorization.rst
/home/travis/build/odlgroup/odl
  File "/home/travis/miniconda/envs/testenv/bin/travis-sphinx", line 11, in <module>
    sys.exit(main())
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/travis_sphinx/build.py", line 39, in build
    if sphinx.build_main(args + [source, outdir]):
AttributeError: module 'sphinx' has no attribute 'build_main'
```

kohr-h:
Weird enough, Travis does install sphinx 1.7.0 so that should be fine (still a bummer on the side of `travis-sphinx`).
Can you try excluding newer versions of travis-sphinx? Using `travis-sphinx<2.0` in the `pip install` command. While editing the Travis config, can you also remove the conda version restriction? The current version should be fine.

kohr-h:
Ah no, it's the other way around. The `build_main` is gone from sphinx 1.7 on, so just use `sphinx<1.7` in the pip install.

kohr-h:
See [here](https://github.com/Syntaf/travis-sphinx/issues/34) as well

kohr-h:
Ah shoot, you need to put it in quotes since the < sign is interpreted by the shell: `"sphinx<1.7"`

kohr-h:
Looks like this one will work, except that it may complain about too much console output. Get ready for re-inserting the `2>/dev/null`. And take out the conda restriction when you make the final version please.

aringh:
If this works I will take out the conda restriction and squash all the commits into one. But I just wanted to do one thing at a time, to see what works and what doesn't ;-)

aringh:
It passed! I will update as asked by @kohr-h and clean up last part of the commit history

aringh:
Could @kohr-h or @adler-j merge this? :-)

adler-j:
Done! Well done

ozanoktem:
It seems judging from the checkmarks that we still lack an example of a reconstruction algorithm that uses Bregma distances. 

In https://link.springer.com/chapter/10.1007/978-3-319-01712-9_2 there is a formulation, see eq. (100), of a Bregman-TV reconstruction methods for problems with Poisson noise that contains many state-of-the-art schemes as special cases. Perhaps we could implement that in ODL and use it to generate an example?

The same paper also lists closed form expressions for the Bregma distance for a number of functionals, see table 1 on p. 121.

adler-j:
Imo that would be far too complicated for the standard examples. A more simple example is more urgently needed.

With that said, adding that example as well would be nice I guess!

aringh:
As mentioned higher up in this conversation I have a Bregman-TV example for tomography lying around. But I am not to convinced by the actual performance of the algorithm. I will create a new PR, to make it easier to find, and if @ozanoktem would like to have a look at it that would be great :-)

aringh:
See PR #1305

odl/solvers/functional/functional.py
mehrhardt: When I define the Bregman distance, I normally would like to give a specific subgradient and not a subgradient operator that would compute one.

odl/solvers/functional/functional.py
```diff
@@ -225,6 +225,48 @@ def translated(self, shift):
...
+            The Bregman distance functional.
+
+        Notes
+        -----
+        Given a functional :math:`f`, which has a (sub)gradient
+        :math:`\partial f`, and given a point :math:`y`, the Bregman distance
+        functional :math:`D_f(\cdot, y)` in a point :math:`x` is given by
+
+        .. math::
+            D_f(x, y) = f(x) - f(y) - \\langle \partial f(y), x - y \\rangle.
```
mehrhardt: The notation here is non-standard. Usually, one would pick a single subgradient p from the subdifferential \partial f(y) and define D_f^p(x, y) = f(x) - f(y) - \\langle p, x - y \\rangle.

adler-j:
Thank you for the comments @mehrhardt, I agree with both of those.

kohr-h:
Without looking into it, I'll agree with the general idea that we should try to follow standard notation and make an implementation in accordance to that.

I made a new issue #1312

PR #1282, opened 2018-02-07
MAINT: update README with a DOI.

aringh:
Or should I write the full journal name? (SIAM Journal on Imaging Sciences)


ozanoktem:
Yes, write out the full journal name. There is no need for using abbreviated version since we have no page limitations.

aringh:
Updated according to the comment from @ozanoktem. Good to go?

adler-j:
Congratulations indeed!

Issue #1279, opened 2018-02-01
Functional translation does not convert stepsizes given as lists or array


sbanert:
The following code crashes:
```
import odl
import numpy as np

space = odl.rn(2)
functional = odl.solvers.L2NormSquared(space)
x = space.one()
stepsize = space.element([0.5, 2.0])
stepsize_list = [0.5, 2.0]
stepsize_array = np.asarray([0.5, 2.0])
y = functional.translated([0.5, 0.5]).convex_conj.proximal(stepsize)(x)
y_list = functional.translated([0.5, 0.5]).convex_conj.proximal(stepsize_list)(x)
y_array = functional.translated([0.5, 0.5]).convex_conj.proximal(stepsize_array)(x)
```
both for the calculation of y_list and y_array. The code works fine without ".translated".

adler-j:
The error messages are:

```
Traceback (most recent call last):

  File "<ipython-input-1-6777289be4b3>", line 11, in <module>
    y_list = functional.translated([0.5, 0.5]).convex_conj.proximal(stepsize_list)(x)

  File "e:\github\odl\odl\solvers\nonsmooth\proximal_operators.py", line 383, in quadratic_perturbation_prox_factory
    const = 1.0 / np.sqrt(sigma * 2.0 * a + 1)

TypeError: can't multiply sequence by non-int of type 'float'

Traceback (most recent call last):

  File "<ipython-input-1-6777289be4b3>", line 11, in <module>
    y_list = functional.translated([0.5, 0.5]).convex_conj.proximal(stepsize_list)(x)

  File "e:\github\odl\odl\solvers\nonsmooth\proximal_operators.py", line 383, in quadratic_perturbation_prox_factory
    const = 1.0 / np.sqrt(sigma * 2.0 * a + 1)

TypeError: can't multiply sequence by non-int of type 'float'
```

sbanert:
If both "list" lines are deleted, the error message is different:

```python
Traceback (most recent call last):

  File "<ipython-input-1-a405cae3e522>", line 1, in <module>
    runfile('[...]/.config/spyder-py3/temp.py', wdir='[...]/.config/spyder-py3')

  File "[...]/miniconda3/lib/python3.6/site-packages/spyder/utils/site/sitecustomize.py", line 705, in runfile
    execfile(filename, namespace)

  File "[...]/miniconda3/lib/python3.6/site-packages/spyder/utils/site/sitecustomize.py", line 102, in execfile
[...]

PR #1283, opened 2018-02-07
BUG: Add typecheck in quadratic perturbation proximal, solves #1279.

sbanert:
This seems to solve the problem, but maybe it should be checked if more of these problems exist, which could mess up proximals.


kohr-h:
Could you add a test, maybe something simple like [your minimal example](https://github.com/odlgroup/odl/issues/1279#issue-293557204)?
Also, if you use one of "closes" or "fixes" instead of "solves" in the commit message, the issue will be automatically closed when the PR is merged. That would be nice 😄

sbanert:
Done.

kohr-h:
> Done.

Thanks, looks good. You'll need to rebase on master anyway, can you do the "solves" -> "closes" thing? You can do that with `rebase -i <commit before first in PR>`, and then just reword the commit message.

sbanert:
Done.

odl/test/solvers/functional/functional_test.py
```diff
@@ -449,6 +449,31 @@ def test_translation_of_functional(space):
                         places=places)
 
 
+def test_translation_proximal():
```
adler-j: There is already such a test here:

https://github.com/odlgroup/odl/blob/cfcd823c20b7747da8a4121fd2b5d588a0db285b/odl/test/solvers/functional/functional_test.py#L394

I feel like the naming of this should perhaps be more specific to distinguish them

sbanert: First, I tried to integrate the test there, which was not really elegant because of the space fixture. With the name I wanted to make clear that it is a test specifically for the proxmal of this operator. Would “test_translation_proximal_stepsizes” be clearer?

adler-j: Sounds mouch more clear to me!

sbanert: Okay, I changed the name (with git amend).

sbanert:
Should I do anything more here?

PR #1285, opened 2018-02-08
BUG: fix DiscreteLp.astype from bool

kohr-h:
The issue was that the real/complex space caching
mechanism would look up space.real_space, which
failed with a TypeError.

### DIFF
--- a/odl/space/fspace.py
+++ b/odl/space/fspace.py
@@ -702,17 +702,24 @@ class FunctionSpace(LinearSpace):
         if out_dtype == self.out_dtype:
             return self
 
-        # Caching for real and complex versions (exact dtyoe mappings)
-        if out_dtype == self.real_out_dtype:
-            if self.__real_space is None:
-                self.__real_space = self._astype(out_dtype)
-            return self.__real_space
-        elif out_dtype == self.complex_out_dtype:
-            if self.__complex_space is None:
-                self.__complex_space = self._astype(out_dtype)
-            return self.__complex_space
-        else:
+        # Try to use caching for real and complex versions (exact dtype
+        # mappings). This may fail for certain dtype, in which case we
+        # just go to `_astype` directly.
+        try:
+            real_dtype = self.real_out_dtype
+        except TypeError:
             return self._astype(out_dtype)
+        else:
+            if out_dtype == real_dtype:
+                if self.__real_space is None:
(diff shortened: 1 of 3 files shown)


odl/space/fspace.py
```diff
@@ -702,17 +702,24 @@ def astype(self, out_dtype):
...
-            if self.__complex_space is None:
-                self.__complex_space = self._astype(out_dtype)
-            return self.__complex_space
-        else:
+        # Try to use caching for real and complex versions (exact dtype
+        # mappings). This may fail for certain dtype, in which case we
+        # just go to `_astype` directly.
+        try:
+            real_dtype = self.real_out_dtype
+        except TypeError:
```
adler-j: Why typeerror here? Shouldn't this be an attributeerror?

kohr-h: I'm reluctant to use `AttributeError` for this kind of stuff. To me it feels like `attr = getattr(obj, 'attr', None)` should be equivalent to
```python
try:
    attr = obj.attr
except AttributeError:
    attr = None
```
which it wouldn't if I raised an `AttributeError` for an existing attribute.

kohr-h:
Updated

odl/test/discr/lp_discr_test.py
```diff
@@ -754,6 +754,15 @@ def test_astype():
...
 
+    # More exotic dtype
+    discr = odl.uniform_discr([0, 0], [1, 1], [2, 2], dtype=bool)
+    as_float = discr.astype(float)
+    assert as_float.dtype == float
+    assert not as_float.is_weighted
+    as_complex = discr.astype(complex)
+    assert as_complex.dtype == complex
+    assert not as_complex.is_weighted
+
```
adler-j: Do we have any "raises" test that validates that it actually gives `AttributeError`? It seems not?

kohr-h: There's a test for that in `fspace_test.py`.

adler-j: Shouldn't it need to be updated then?

adler-j: I see it has been :)

kohr-h: Yep!

PR #1286, opened 2018-02-09
Default stepsize selection for optimizers

adler-j:
How does the ODL community feel about this change?

### DIFF
--- a/odl/solvers/nonsmooth/douglas_rachford.py
+++ b/odl/solvers/nonsmooth/douglas_rachford.py
@@ -16,9 +16,9 @@ from odl.operator import Operator
 __all__ = ('douglas_rachford_pd',)
 
 
-def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
+def douglas_rachford_pd(x, f, g, L, niter, tau=None, sigma=None,
                         callback=None, **kwargs):
-    """Douglas-Rachford primal-dual splitting algorithm.
+    r"""Douglas-Rachford primal-dual splitting algorithm.
 
     Minimizes the sum of several convex functions composed with linear
     operators::
@@ -48,12 +48,14 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
         ``g[i].convex_conj.proximal``.
     L : sequence of `Operator`'s
         Sequence of `Opeartor`'s with as many elements as ``g``.
-    tau : float
-        Step size parameter for ``f``.
-    sigma : sequence of floats
-        Step size parameters for the ``g_i``'s.
     niter : int
         Number of iterations.
+    tau : float, optional
+        Step size parameter for ``f``.
+        Default: Sufficient for convergence, see Notes.
+    sigma : sequence of floats, optional
+        Step size parameters for the ``g_i``'s.
+        Default: Sufficient for convergence, see Notes.
     callback : callable, optional
         Function called with the current iterate after each iteration.
 
@@ -89,19 +91,28 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
     can be obtained by setting
 
     .. math::
-        l(x) = 0 \\text{ if } x = 0, \infty \\text{ else.}
+        l(x) = 0 \text{ if } x = 0, \infty \text{ else.}
 
-    To guarantee convergence, the parameters :math:`\\tau`, :math:`\\sigma_i`
+    To guarantee convergence, the parameters :math:`\tau`, :math:`\sigma_i`
     and :math:`L_i` need to satisfy
 
     .. math::
-       \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2 < 4
+       \tau \sum_{i=1}^n \sigma_i ||L_i||^2 < 4
 
-    The parameter :math:`\\lambda` needs to satisfy :math:`0 < \\lambda < 2`
+    An example of such a choice is
+
+    .. math::
+        \tau = \frac{1}{\sum_{i=1}^n ||L_i||},
+        \quad
+        \sigma = \frac{2}{n \tau \|L_i\|^2}
+
+    which are the defaults for this implementation.
+
+    The parameter :math:`\lambda` needs to satisfy :math:`0 < \lambda < 2`
     and if it is given as a function it needs to satisfy
 
     .. math::
-        \\sum_{n=1}^\infty \\lambda_n (2 - \\lambda_n) = +\infty.
+        \sum_{n=1}^\infty \lambda_n (2 - \lambda_n) = +\infty.
 
     See Also
     --------
@@ -128,10 +139,16 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
         raise ValueError('not all operators in `L` are linear')
     if not all(x in op.domain for op in L):
         raise ValueError('`x` not in the domain of all operators')
-    if len(sigma) != m:
-        raise ValueError('len(sigma) != len(L)')
     if len(g) != m:
         raise ValueError('len(prox_cc_g) != len(L)')
+
+    if tau is None:
+        tau = 1 / sum(l.norm(estimate=True) for l in L)
+
+    if sigma is None:
+        sigma = [(2.0 / m) / (tau * l.norm(estimate=True) ** 2) for l in L]
+    elif len(sigma) != m:
+        raise ValueError('len(sigma) != len(L)')
 
     prox_cc_g = [gi.convex_conj.proximal for gi in g]
 
--- a/odl/operator/operator.py
+++ b/odl/operator/operator.py
@@ -739,8 +739,13 @@ class Operator(object):
                                       '`Operator.norm(estimate=True)` to '
                                       'obtain an estimate.')
         else:
-            from odl.operator.oputils import power_method_opnorm
-            return power_method_opnorm(self, **kwargs)
+            norm = getattr(self, '__norm', None)
+            if norm is not None:
+                return norm
+            else:
+                from odl.operator.oputils import power_method_opnorm
+                self.__norm = power_method_opnorm(self, **kwargs)
+                return self.__norm
 
     def __add__(self, other):
         """Return ``self + other``.
--- a/odl/solvers/iterative/iterative.py
+++ b/odl/solvers/iterative/iterative.py
@@ -23,7 +23,7 @@ __all__ = ('landweber', 'conjugate_gradient', 'conjugate_gradient_normal',
 # TODO: update all docs
 
 
-def landweber(op, x, rhs, niter, omega=1, projection=None, callback=None):
+def landweber(op, x, rhs, niter, omega=None, projection=None, callback=None):
     """Optimized implementation of Landweber's method.
 
     Solves the inverse problem::
@@ -45,6 +45,7 @@ def landweber(op, x, rhs, niter, omega=1, projection=None, callback=None):
         Number of iterations.
     omega : positive float, optional
         Relaxation parameter in the iteration.
+        Default: ``1 / op.norm(estimate=True) ** 2``
     projection : callable, optional
         Function that can be used to modify the iterates in each iteration,
         for example enforcing positivity. The function should take one
@@ -98,6 +99,9 @@ def landweber(op, x, rhs, niter, omega=1, projection=None, callback=None):
     if x not in op.domain:
         raise TypeError('`x` {!r} is not in the domain of `op` {!r}'
                         ''.format(x, op.domain))
+
+    if omega is None:
+        omega = 1 / op.norm(estimate=True) ** 2
 
     # Reusable temporaries
     tmp_ran = op.range.element()


aringh:
I don't mind these kind of default values where it is possible. To me it looks like it is only going to make life easier for a user, which is a good thing.

kohr-h:
I didn't read this until now, and funny enough I had started to write similar code. I think it's a great idea, the easier these methods are to use, the better.

Question: what's the rational for the choice of `tau`?

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -89,19 +91,28 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
-        l(x) = 0 \\text{ if } x = 0, \infty \\text{ else.}
+        l(x) = 0 \text{ if } x = 0, \infty \text{ else.}
 
-    To guarantee convergence, the parameters :math:`\\tau`, :math:`\\sigma_i`
+    To guarantee convergence, the parameters :math:`\tau`, :math:`\sigma_i`
     and :math:`L_i` need to satisfy
 
     .. math::
-       \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2 < 4
+       \tau \sum_{i=1}^n \sigma_i ||L_i||^2 < 4
```
kohr-h: `||` -> `\|`

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -128,11 +139,17 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
         raise ValueError('not all operators in `L` are linear')
     if not all(x in op.domain for op in L):
         raise ValueError('`x` not in the domain of all operators')
-    if len(sigma) != m:
-        raise ValueError('len(sigma) != len(L)')
     if len(g) != m:
         raise ValueError('len(prox_cc_g) != len(L)')
 
+    if tau is None:
+        tau = 1 / sum(l.norm(estimate=True) for l in L)
```
kohr-h: Prefer `for Li in L`

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -89,19 +91,28 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
     can be obtained by setting
 
     .. math::
-        l(x) = 0 \\text{ if } x = 0, \infty \\text{ else.}
+        l(x) = 0 \text{ if } x = 0, \infty \text{ else.}
```
kohr-h: Finally, no more annoying double backslashes!

kohr-h:
One more comment: I think it would make sense to break this out into a little helper (per solver) that does this computation, with the option to provide precomputed operator norms. Sometimes it's too expensive to use the default settings of the power iteration, and for that case the user should have a way to compute provide the norms themselves get back the default step sizes for those.
Alternatively, the `douglas_rachford_pd` itself could accept an optional `L_norms` sequence and use those norms instead of computing them itself.
In both cases, it could be flexible with `None` entries signalling "please compute".

adler-j:
> One more comment: I think it would make sense to break this out into a little helper (per solver) that does this computation

Exellent suggestion. In that case I'd go for `SOLVER_NAME_stepsize`, is that OK?

> with the option to provide precomputed operator norms

If we break this out into a utility, I'm OK with that. For the "core functions" we don't want to clog the interface.

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: We really should make an across the board change for this

adler-j:
> Question: what's the rationale for the choice of tau?

The rationall here is rather weak, but in general we want `\tau \approx \sigma`. In this case, this was a decent "average guess" that seems to work well. 

Perhaps @sbanert knows of some reference for a "best"  initial guess in some sense?

kohr-h:
> In that case I'd go for SOLVER_NAME_stepsize, is that OK?

Sounds good to me (keeping in mind that some methods only have a single such parameter).

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -48,12 +50,14 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
         Sequence of `Opeartor`'s with as many elements as ``g``.
-    tau : float
-        Step size parameter for ``f``.
-    sigma : sequence of floats
-        Step size parameters for the ``g_i``'s.
     niter : int
         Number of iterations.
+    tau : float, optional
+        Step size parameter for ``f``.
+        Default: Sufficient for convergence, see Notes.
```
adler-j: Should refer  `douglas_rachford_pd_stepsize`

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -89,19 +93,28 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
 
-    To guarantee convergence, the parameters :math:`\\tau`, :math:`\\sigma_i`
+    To guarantee convergence, the parameters :math:`\tau`, :math:`\sigma_i`
     and :math:`L_i` need to satisfy
 
     .. math::
-       \\tau \\sum_{i=1}^n \\sigma_i ||L_i||^2 < 4
+       \tau \sum_{i=1}^n \sigma_i \|L_i\|^2 < 4
+
+    An example of such a choice is
```
adler-j: This "example" should be removed from this function to avoid duplication.

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -118,21 +131,28 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
         raise ValueError('not all operators in `L` are linear')
     if not all(x in op.domain for op in L):
         raise ValueError('`x` not in the domain of all operators')
-    if len(sigma) != m:
-        raise ValueError('len(sigma) != len(L)')
     if len(g) != m:
         raise ValueError('len(prox_cc_g) != len(L)')
 
+    if tau is None or sigma is None:
+        tau_default, sigma_default = douglas_rachford_pd_stepsize(L, tau)
```
adler-j: We want to support the case `sigma` given but not `tau`. I'll implement this

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +244,61 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
     # The final result is actually in p1 according to the algorithm, so we need
     # to assign here.
     x.assign(p1)
+
+
+def douglas_rachford_pd_stepsize(L, tau=None):
+    r"""Compute defautl step sizes for the Douglas-Rachford solver.
```
adler-j: defautl

adler-j: Also, we need a `douglas_rachford_pd` link somewhere here

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +244,61 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+
+    Parameters
+    ----------
+    L : sequence of `Operator` or float
+        The operators or the norms of the operators that are used in the
+        `douglas_rachford_pd` method. For `Operator` entries, the norm
+        is computed with `power_method_opnorm`.
```
adler-j: `power_method_opnorm` -> `Operator.norm(estimate=True)`

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +244,61 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+        if np.isscalar(Li):
+            L_norms.append(float(Li))
+        elif isinstance(Li, Operator):
+            L_norms.append(Li.norm(estimate=True))
+        else:
+            raise TypeError('invalid entry {!r} in `L`'.format(Li))
+
+    if tau is None:
+        norm_sum = sum(L_norms)
+        tau = 1 / norm_sum if norm_sum != 0 else 1.0
```
adler-j: Frankly I'd prefer throwing an error in this case. Obviously something went wrong, no?

odl/solvers/nonsmooth/douglas_rachford.py
kohr-h: What's missing?

odl/solvers/nonsmooth/douglas_rachford.py
kohr-h: I guess so. I was just thinking a bit back and forth about how to handle operators that don't matter, which are those that are used with an indicator function. There you can take whatever step size you like and only take the rest of the ingredients into account. But it seems to be easiest to do that manually. So let's raise if there is a nonpositive norm anywhere.

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: Basically there are 4 use cases:

None, None
tau, None
None, Sigma
tau, sigma

To the best of my understanding, this version only handles these ones

None, None
tau, None
tau, sigma

right?

kohr-h: In the "None, sigma" case the method is run and you get the defaults (`tau=None` is OK for the helper), and then you set `tau` to its default and `sigma` is skipped. 
So to me it seems all cases are handled, or do I have this backwards?

adler-j: Lets say I pick sigma = [HUGE NUMBER], then this would still pick tau=[AVERAGE NUMBER], instead of tau=[TINY NUMBER], right?

So sure, it is handled, but I don't think we would have convergence?

sbanert:
> Question: what's the rationale for the choice of tau?  The rationall here is rather weak, but in general we want \tau \approx \sigma. In this case, this was a decent "average guess" that seems to work well. Perhaps @sbanert knows of some reference for a "best"  initial guess in some sense?

As far as I know the “optimal” choice of τ and σ is strongly problem-dependent (a bit like finding the regularisation parameter in an inverse problem). But if you know nothing, I think τ ≈ σ is a good point to start with.

odl/solvers/nonsmooth/douglas_rachford.py
kohr-h: Sure, it's always the case that `sigma` depends on `tau`, not the other way around. Programmatically speaking, all cases are handled, but one of them incorrectly (not guaranteeing convergence). In that sense you're right.
I guess you'd take something like `2 / sum(sigma)` for that case.

adler-j:
Also added this for PDHG, I guess we could merge this now (after review) and then make a new PR where we update the examples etc properly.

kohr-h:
Seems like you triggered PEP8 checks on a bunch of examples, now you gotta fix them :grin:

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -72,7 +78,7 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
     The mathematical problem to solve is
 
     .. math::
-       \min_x f(x) + \sum_{i=0}^n (g_i \Box l_i)(L_i x),
+       \min_x f(x) + \sum_{i=0}^n (g_i \Box l_tuplei)(L_i x),
```
kohr-h: This looks like an accident

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -301,6 +303,75 @@ def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
             callback(x)
 
 
+def pdhg_stepsize(L, tau=None, sigma=None):
+    r"""Default step sizes for `pdhg`.
+
+    Parameters
+    ----------
+    L : `Operator` or float
+        Operators or norm of the operators that are used in the `pdhg` method.
```
kohr-h: Singular

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -301,6 +303,75 @@ def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
...
+        If it is an operator `Operator`, the norm is computed with
```
kohr-h: double operator

adler-j:
Damn.

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -301,6 +303,75 @@ def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
...
+    To guarantee convergence, the parameters :math:`\tau`, :math:`\sigma`
+    and :math:`L` need to satisfy
+
+    .. math::
+       \tau \sigma \|L\|^2 < 1
+
+    This function has 4 options, :math:`\tau`/:math:`\sigma` given or not
+    given.
+
+    If :math:`\tau` nor :math:`\sigma` is given, they are chosen as:
```
kohr-h: neither ... nor are given

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +231,100 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+    To guarantee convergence, the parameters :math:`\tau`, :math:`\sigma_i`
+    and :math:`L_i` need to satisfy
+
+    .. math::
+       \tau \sum_{i=1}^n \sigma_i \|L_i\|^2 < 4.
+
+    This function has 4 options, :math:`\tau`/:math:`\sigma` given or not
+    given.
+
+    If :math:`\tau` nor :math:`\sigma` is given, they are chosen as:
```
kohr-h: neither ... nor are given

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +231,100 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+
+    Notes
+    -----
+    To guarantee convergence, the parameters :math:`\tau`, :math:`\sigma_i`
+    and :math:`L_i` need to satisfy
+
+    .. math::
+       \tau \sum_{i=1}^n \sigma_i \|L_i\|^2 < 4.
+
+    This function has 4 options, :math:`\tau`/:math:`\sigma` given or not
```
kohr-h: Maybe use an itemize list to improve readability? Like
```
- Neither :math:`\tau` nor :math:`\sigma` given:
   
   .. math::
       ...
```

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -18,13 +18,13 @@
...
 
-__all__ = ('pdhg',)
+__all__ = ('pdhg', 'pdhg_stepsize')
 
 
 # TODO: add dual gap as convergence measure
 # TODO: diagonal preconditioning
 
-def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
+def pdhg(x, f, g, L, niter, tau, sigma, **kwargs):
```
kohr-h: If you do this API change, please also switch the roles of `f` and `g`, see #1197

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: Great suggestion, fixing that.

adler-j:
I think this could use a re-review by now

odl/solvers/nonsmooth/douglas_rachford.py
kohr-h: Still valid

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +231,103 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+        .. math::
+            \tau = \frac{2}{\sum_{i=1}^n \sigma_i \|L_i\|^2}
+
+    - If only :math:`\tau` is given, :math:`\sigma` is set
+      to:
+
+        .. math::
+            \sigma_i = \frac{2}{n \tau \|L_i\|}
+
+    - If both are given, they are returned as is without further validation.
```
kohr-h: as-is

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +231,103 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+
+    This function has 4 options, :math:`\tau`/:math:`\sigma` given or not
+    given.
+
+    - If neither :math:`\tau` nor :math:`\sigma` are given, they are chosen as:
+
+        .. math::
+            \tau = \frac{1}{\sum_{i=1}^n \|L_i\|},
+            \quad
+            \sigma_i = \frac{2}{n \tau \|L_i\|}
```
kohr-h: Shouldn't this be `\|L_i\|^2`?

odl/solvers/nonsmooth/douglas_rachford.py
```diff
@@ -224,3 +231,103 @@ def douglas_rachford_pd(x, f, g, L, tau, sigma, niter,
...
+    - If only :math:`\sigma` is given, :math:`\tau` is set to:
+
+        .. math::
+            \tau = \frac{2}{\sum_{i=1}^n \sigma_i \|L_i\|^2}
+
+    - If only :math:`\tau` is given, :math:`\sigma` is set
+      to:
+
+        .. math::
+            \sigma_i = \frac{2}{n \tau \|L_i\|}
```
kohr-h: Square?

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
```diff
@@ -301,6 +303,77 @@ def pdhg(x, f, g, L, tau, sigma, niter, **kwargs):
...
+        .. math::
+            \tau = \frac{0.9}{\sigma \|L\|^2}
+
+    - If only :math:`\tau` is given, :math:`\sigma` is set
+      to:
+
+        .. math::
+            \sigma = \frac{0.9}{\tau \|L\|^2}
+
+    - If both are given, they are returned as is without further validation.
```
kohr-h: as-is

odl/solvers/nonsmooth/douglas_rachford.py
adler-j: In fact not, since the product is sigma*tau, both of which are `\|L_i\|`, hence `\|L_i\| * \|L_i\| = \|L_i\|^2`

adler-j: Oooor I gotta think here

adler-j:
I've now fixed the comments and done the f<->g change. Could really use a review of that.

odl/solvers/nonsmooth/primal_dual_hybrid_gradient.py
aringh: They should not be switched here, should they? All other (nonsmooth )solvers have this kind of signature:
```python
def douglas_rachford_pd(x, f, g, L, tau, sigma, niter, callback=None, **kwargs):
def admm_linearized(x, f, g, L, tau, sigma, niter, **kwargs):
def forward_backward_pd(x, f, g, L, h, tau, sigma, niter, callback=None, **kwargs):
```

odl/operator/oputils.py
```diff
@@ -133,8 +134,8 @@ def power_method_opnorm(op, xstart=None, maxiter=100, rtol=1e-05, atol=1e-08,
         `Operator.adjoint` must be defined (which implies that the
         operator must be linear).
     xstart : ``op.domain`` `element-like`, optional
-        Starting point of the iteration. By default, the ``one``
-        element of the `Operator.domain` is used.
+        Starting point of the iteration. By default an `Operator.domain`
+        element containing noise is used.
```
sbanert: Maybe this change makes the Python 2.7 doctest fail, and a non-random starting point should be used in the test?

sbanert:
A general remark: This code will probably fail if the operator is zero (or if the random element chosen by the power method is in the kernel, which is unlikely). Setting an operator to zero might not be useful for optimisation, but I’d prefer this code to work even in this case. Maybe, a division by zero should be caught?

odl/operator/oputils.py
kohr-h: Probably. We need to change the doctest a bit then.

kohr-h:
> A general remark: This code will probably fail if the operator is zero (or if the random element chosen by the power method is in the kernel, which is unlikely). Setting an operator to zero might not be useful for optimisation, but I’d prefer this code to work even in this case. Maybe, a division by zero should be caught?

Good point, we should catch that and use a reasonable replacement value.

kohr-h:
> In the file examples/solvers/pdhg_denoising_L2_HuberTV.py, following are the PEP8 issues :
> 
>     Line 47:21: E261 at least two spaces before inline comment

Can you fix this, and maybe add `-E741` to the [.pep8speaks.yml](https://github.com/odlgroup/odl/blob/dfc63836e2c3faaa2d817e391fb0a3052f079c87/.pep8speaks.yml#L16) file?

kohr-h:
> Can you fix this, and maybe add -E741 to the .pep8speaks.yml file to get rid of the second one?

Done in #1344

adler-j:
Are we ready to merge this after the tests?

PR #1287, opened 2018-02-11
Issue 564 functionals on product spaces

### LINKED ISSUE #564: Default functionals for ProductSpace
The default functionals from #246 #498 are only defined for `FnBase`. Implement these also for `ProductSpace`.

aringh:
To close #564. This includes

- [X] add test for power space,
- [ ] add test for product space.

The first one is completed, however it does not work for the `GroupL1Norm` (see #564). Currently that test is simply skipped. Do we want to solve the underlying problem? Or will the `GroupL1Norm` only be defined on `DiscreteLp`  spaces?

 For general product spaces several test fails.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -1244,7 +1244,8 @@ class KullbackLeiblerConvexConj(Functional):
         import scipy.special
 
         if self.prior is None:
-            tmp = -1.0 * (np.log(1 - x)).inner(self.domain.one())
+            tmp = self.domain.element(
+                -1.0 * (np.log(1 - x))).inner(self.domain.one())
         else:
             tmp = (-scipy.special.xlogy(self.prior,
                                         1 - x)).inner(self.domain.one())
@@ -1510,7 +1511,7 @@ class KullbackLeiblerCrossEntropyConvexConj(Functional):
     def _call(self, x):
         """Return the value in the point ``x``."""
         if self.prior is None:
-            tmp = (np.exp(x) - 1).inner(self.domain.one())
+            tmp = self.domain.element((np.exp(x) - 1)).inner(self.domain.one())
         else:
             tmp = (self.prior * (np.exp(x) - 1)).inner(self.domain.one())
         return tmp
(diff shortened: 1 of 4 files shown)


aringh:
Question related to some of the test fails: is there a reason for the difference between output from `numpy` operations if the element belongs to a `DiscreteLp` space or a `ProductSpace`?
```python
import odl
import numpy as np
space = odl.uniform_discr(0,1,10)
x_space = space.one()
np.exp(x_space)
```
gives output
``` 
uniform_discr(0.0, 1.0, 10).element(
    [ 2.71828183,  2.71828183,  2.71828183, ...,  2.71828183,  2.71828183,
      2.71828183]
)
```
while
```python
pspace = odl.ProductSpace(space, 2)
x_pspace = pspace.one()
np.exp(x_pspace)
```
gives output
```
array([[ 2.71828183,  2.71828183,  2.71828183, ...,  2.71828183,
         2.71828183,  2.71828183],
       [ 2.71828183,  2.71828183,  2.71828183, ...,  2.71828183,
         2.71828183,  2.71828183]])
```

aringh:
For general product spaces, some test fails just because `all_almost_equal` cannot be run on product spaces that are not power spaces. Can that be circumvented in a nice way, or does it simply mean that we cannot runt the same type of test on product spaces that are not power spaces?

adler-j:
What numpy version did you use when you ran these tests?

> For general product spaces, some test fails just because all_almost_equal cannot be run on product spaces that are not power spaces. Can that be circumvented in a nice way, or does it simply mean that we cannot runt the same type of test on product spaces that are not power spaces?

This sounds like a bug that we could fix rather easily

aringh:
I am running python 3.6 and numpy 1.12.1.
```bash
nomkl                     1.0                           0
numpy                     1.12.1             py36_nomkl_0  [nomkl]
```
I did not past the complete list of packages since it is rather long. But if you need anything else just let me know.

kohr-h:
Numpy version is irrelevant here since `ProductSpaceElement` does not implement `__array_ufunc__` (would be nice, but also complicated I think). However, it does implement `__array__`, but the `__array_wrap__` seems to have been forgotten 😮

kohr-h:
Should be an easy fix indeed

adler-j:
It seems like someone (named @adler-j) forgot to do that. I'll get it done.

Issue #1249, opened 2017-11-25
Change `space_elem == other` behavior?


kohr-h:
We differ from Numpy in the way we interpret `==`. Our use case has quite limited usability, and Numpy users will be surprised by the interpretation.
Another issue that comes with this difference is our current inability to do things like `x[x > 0] = 1` or so, which would be nice to have. Then, of course, `==` cannot be an exception.

Question: How about we change our tensors to behave like Numpy arrays in that respect?

adler-j:
I frankly vote yes for this. My suggestion is as follows: simply bypass of of these by making

```
def __eq__(self, other):
    return np.equal(self, other)
```

and equivalently for the others.

PR #1288, opened 2018-02-12
ENH/TST: add __array_wrap__ to ProductSpaceElement

aringh:
To fix the observed issue in #1287.

### DIFF
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -1234,36 +1234,27 @@ class ProductSpaceElement(LinearSpaceElement):
         try:
             iter(newreal)
         except TypeError:
+            # `newreal` is not iterable, assume it can be assigned to
+            # all indexed parts
             for part in self.parts:
                 part.real = newreal
             return
 
         if self.space.is_power_space:
             try:
+                # Set same value in all parts
                 for part in self.parts:
                     part.real = newreal
-            except (ValueError, TypeError):  # maybe need to catch more, e.g. `LinearSpaceTypeError`
+            except (ValueError, TypeError):
+                # Iterate over all parts and set them separately
+                for part, new_re in zip(self.parts, newreal):
+                    part.real = new_re
                 pass
         elif len(newreal) == len(self):
             for part, new_re in zip(self.parts, newreal):
                 part.real = new_re
         else:
             raise ValueError('invalid')  # better error
-
-#        if newreal in self.space.field:
-#            for part in self.parts:
-#                part.real.data[:] = newreal
-#        elif newreal in self.space.real_space or newreal in self.space:
-#            for part, newreal_subpart in zip(self.parts, newreal):
-#                part.real.data[:] = newreal_subpart.real.data[:]
-#        elif self.space.shape[0] == np.shape(newreal)[0]:
-#            for part, newreal_subpart in zip(self.parts, newreal):
-#                part.real.data[:] = np.real(newreal_subpart)
-#        else:
-#            raise ValueError('new real part does not the correct first '
-#                             'dimension, expected {} but got {}'
-#                             ''.format(self.space.shape[0],
-#                                       np.shape(newreal)[0]))
 
     @property
     def imag(self):
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -1192,7 +1192,9 @@ class ProductSpaceElement(LinearSpaceElement):
             for part, new_re in zip(self.parts, newreal):
                 part.real = new_re
         else:
-            raise ValueError('invalid')  # better error
+            raise ValueError(
+                'dimensions of the new real part does not match the space, '
+                'got element {} to set real part of {}'.format(newreal, self))
 
     @property
     def imag(self):
@@ -1210,6 +1212,63 @@ class ProductSpaceElement(LinearSpaceElement):
         """
         imag_part = [part.imag for part in self.parts]
         return self.space.real_space.element(imag_part)
+
+    @imag.setter
+    def imag(self, newimag):
+        """Setter for the imaginary part.
+
+        This method is invoked by ``x.imag = other``.
+
+        Parameters
+        ----------
+        newimag : array-like or scalar
+            Values to be assigned to the imaginary part of this element.
+
+        Examples
+        --------
+        >>> space = odl.ProductSpace(odl.cn(3), odl.cn(2))
+        >>> x = space.element([[1 + 1j, 2, 3 - 3j], [1 + 1j, 2]])
+        >>> zero = space.real_space.zero()
+        >>> x.imag = zero
+        >>> x
+        space.element([[1, 2, 3], [1, 2]])
+
+        Other array-like types:
+
+        >>> x.imag = 1.0
+        >>> x
+        space.element([[1 + 1j, 2 + 1j, 3 + 1j], [1 + 1j, 2 + 1j]])
+        >>> x.imag = [[2, 3, 4], [5, 6]]
+        >>> x
+        space.element([[1 + 2j, 2 + 3j, 3 + 4j], [1 + 5j, 2 + 6j]])
+        """
+        try:
+            iter(newimag)
+        except TypeError:
+            # `newimag` is not iterable, assume it can be assigned to
+            # all indexed parts
+            for part in self.parts:
+                part.imag = newimag
+            return
+
+        if self.space.is_power_space:
+            try:
+                # Set same value in all parts
+                for part in self.parts:
+                    part.imag = newimag
+            except (ValueError, TypeError):
+                # Iterate over all parts and set them separately
+                for part, new_im in zip(self.parts, newimag):
+                    part.imag = new_im
+                pass
+        elif len(newimag) == len(self):
+            for part, new_im in zip(self.parts, newimag):
+                part.imag = new_im
+        else:
+            raise ValueError(
+                'dimensions of the new imaginary part does not match the '
+                'space, got element {} to set real part of {}}'
+                ''.format(newimag, self))
 
     def conj(self):
         """Complex conjugate of the element."""
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -386,28 +386,39 @@ class ProductSpace(LinearSpace):
 
     @property
     def real_space(self):
-        """The space corresponding to this space's real dtype.
+        """Variant of this space with real dtype.
 
         Raises
         ------
-        NotImplementedError
-            If `dtype` is not a numeric data type.
-        TypeError
-            If `dtype` is not a real or complex type.
+        ValueError
+            Raised if the underlying spaces have different `dtype`, if `dtype`
+            is not a numeric data type, if `dtype` is not a real or complex
+            type, and if `dtype` is complex but cannot be mapped to a real
+            type.
         """
-        if not is_numeric_dtype(self.dtype):
-            raise NotImplementedError(
+        try:
+            current_dtype = self.dtype
+        except(AttributeError):
+            raise ValueError(
+                'dtypes of underlying spaces are not the same and therefore '
+                '`real_space` is not well-defined')
+
+        if not is_numeric_dtype(current_dtype):
+            raise ValueError(
                 '`real_space` not defined for non-numeric `dtype`')
 
-        if is_real_dtype(self.dtype):
+        if is_real_dtype(current_dtype):
             return self
-        elif is_complex_floating_dtype(self.dtype):
-            new_dtype = TYPE_MAP_C2R.get(self.dtype, 'None')
+        elif is_complex_floating_dtype(current_dtype):
+            new_dtype = TYPE_MAP_C2R.get(current_dtype, 'None')
+            if new_dtype is None:
+                raise ValueError('complex type `self.dtype` {} not possible '
+                                 'to map to a real type'.format(self.dtype))
             return self.astype(new_dtype)
         else:
-            raise TypeError('`real_space` is only defined for space with '
-                            'real or complex `dtype`, got {}'
-                            ''.format(self.dtype))
+            raise ValueError(
+                '`real_space` is only defined for space with real or complex '
+                '`dtype`, got {}'.format(current_dtype))
 
     @property
     def complex_space(self):
@@ -415,24 +426,35 @@ class ProductSpace(LinearSpace):
 
         Raises
         ------
-        NotImplementedError
-            If `dtype` is not a numeric data type.
-        TypeError
-            If `dtype` is not a real or complex type.
+        ValueError
+            Raised if the underlying spaces have different `dtype`, if `dtype`
+            is not a numeric data type, if `dtype` is not a real or complex
+            type, and if `dtype` is real but cannot be mapped to a complex
+            type.
         """
-        if not is_numeric_dtype(self.dtype):
-            raise NotImplementedError(
+        try:
+            current_dtype = self.dtype
+        except(AttributeError):
+            raise ValueError(
+                'dtypes of underlying spaces are not the same and therefore '
+                '`complex_space` is not well-defined')
+
+        if not is_numeric_dtype(current_dtype):
+            raise ValueError(
                 '`complex_space` not defined for non-numeric `dtype`')
 
-        if is_real_dtype(self.dtype):
+        if is_real_dtype(current_dtype):
             new_dtype = TYPE_MAP_R2C.get(self.dtype, 'None')
+            if new_dtype is None:
+                raise ValueError('real type `self.dtype` {} not possible to '
+                                 'map to a complex type'.format(self.dtype))
             return self.astype(new_dtype)
-        elif is_complex_floating_dtype(self.dtype):
+        elif is_complex_floating_dtype(current_dtype):
             return self
         else:
-            raise TypeError('`complex_space` is only defined for space with '
-                            'real or complex `dtype`, got {}'
-                            ''.format(self.dtype))
+            raise ValueError(
+                '`complex_space` is only defined for space with real or '
+                'complex `dtype`, got {}'.format(current_dtype))
 
     def astype(self, dtype):
         """Return a copy of this space with new ``dtype``.
@@ -455,7 +477,9 @@ class ProductSpace(LinearSpace):
             raise ValueError('`None` is not a valid data type')
 
         dtype = np.dtype(dtype)
-        if dtype == self.dtype:
+        current_dtype = getattr(self, 'dtype', None)
+
+        if dtype == current_dtype:
             return self
         else:
             return ProductSpace(*[space.astype(dtype)
@@ -1207,20 +1231,39 @@ class ProductSpaceElement(LinearSpaceElement):
         >>> x
         space.element([[2 + 1j, 3, 4 - 3j], [5 + 1j, 6]])
         """
-        if newreal in self.space.field:
+        try:
+            iter(newreal)
+        except TypeError:
             for part in self.parts:
-                part.real.data[:] = newreal
-        elif newreal in self.space.real_space or newreal in self.space:
-            for part, newreal_subpart in zip(self.parts, newreal):
-                part.real.data[:] = newreal_subpart.real.data[:]
-        elif self.space.shape[0] == np.shape(newreal)[0]:
-            for part, newreal_subpart in zip(self.parts, newreal):
-                part.real.data[:] = np.real(newreal_subpart)
+                part.real = newreal
+            return
+
+        if self.space.is_power_space:
+            try:
+                for part in self.parts:
+                    part.real = newreal
+            except (ValueError, TypeError):  # maybe need to catch more, e.g. `LinearSpaceTypeError`
+                pass
+        elif len(newreal) == len(self):
+            for part, new_re in zip(self.parts, newreal):
+                part.real = new_re
         else:
-            raise ValueError('new real part does not the correct first '
-                             'dimension, expected {} but got {}'
-                             ''.format(self.space.shape[0],
-                                       np.shape(newreal)[0]))
+            raise ValueError('invalid')  # better error
+
+#        if newreal in self.space.field:
+#            for part in self.parts:
+#                part.real.data[:] = newreal
+#        elif newreal in self.space.real_space or newreal in self.space:
+#            for part, newreal_subpart in zip(self.parts, newreal):
+#                part.real.data[:] = newreal_subpart.real.data[:]
+#        elif self.space.shape[0] == np.shape(newreal)[0]:
+#            for part, newreal_subpart in zip(self.parts, newreal):
+#                part.real.data[:] = np.real(newreal_subpart)
+#        else:
+#            raise ValueError('new real part does not the correct first '
+#                             'dimension, expected {} but got {}'
+#                             ''.format(self.space.shape[0],
+#                                       np.shape(newreal)[0]))
 
     @property
     def imag(self):
--- a/odl/space/pspace.py
+++ b/odl/space/pspace.py
@@ -1083,29 +1083,29 @@ class ProductSpaceElement(LinearSpaceElement):
 
     @property
     def real(self):
-        """Pointwise real part of the element."""
-        size = self.space.shape[0]
-        to_return = [None] * size
-        for (i, part) in zip(range(size), self.__parts):
-            to_return[i] = part.real
-        return self.space.element(to_return)
+        """Real part of the element."""
+        size = len(self)
+        real_part = [None] * size
+        for i, part in enumerate(self.parts):
+            real_part[i] = part.real
+        return self.space.element(real_part)
 
     @property
     def imag(self):
-        """Pointwise imaginary part of the element."""
-        size = self.space.shape[0]
-        to_return = [None] * size
-        for (i, part) in zip(range(size), self.__parts):
-            to_return[i] = part.imag
-        return self.space.element(to_return)
+        """Imaginary part of the element."""
+        size = len(self)
+        imag_part = [None] * size
+        for i, part in enumerate(self.parts):
+            imag_part[i] = part.imag
+        return self.space.element(imag_part)
 
     def conj(self):
-        """Pointwise complex conjugate of the element."""
-        size = self.space.shape[0]
-        to_return = [None] * size
-        for (i, part) in zip(range(size), self.__parts):
-            to_return[i] = part.conj()
-        return self.space.element(to_return)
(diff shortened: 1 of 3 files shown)


adler-j:
Lets merge this. For later, we need to fix the case of non-product spaces though. This should require the `__array_ufunc__` interface.

aringh:
The Doc build failed, but I don't understand why. Does anything need to be changed?

aringh:
I just rebased on master, and with commit eb9e7d42468bcd757b9fb6c343d544fb9f631a46 now in I hope that all tests should pass.

adler-j:
It seems the crash here is an actual bug and we should add `ProductSpaceElement.real` (and I guess `imag` and `conj()`).

aringh:
That is what I am trying to figure out, except you are faster than me ;-) 
```
In [1]: import odl
   ...: space = odl.rn(5)
   ...: pspace = odl.ProductSpace(space, 2)
   ...: x = space.one()
   ...: x_ps = pspace.one()
   ...: x.real
   ...: x_ps.real

  File "<ipython-input-1-140af8ed0ee3>", line 7, in <module>
    x_ps.real

AttributeError: 'ProductSpaceElement' object has no attribute 'real'
```
The weird thing is that the [build did not fail](https://travis-ci.org/odlgroup/odl/builds/340398388) before I rebased. But I do not see which commit in the last 24 hours that have changed this?

aringh:
I had two test fails in this commit when I run it on my local machine. However, it is in a part of the code I do not think I have changed. We will see what Travis says...

Anyway, the test fails are:
```
=================================== FAILURES ===================================
____________________________ test_is_numeric_dtype _____________________________

    def test_is_numeric_dtype():
        for dtype in numeric_dtypes:
>           assert is_numeric_dtype(dtype)
E           AssertionError: assert False
E            +  where False = is_numeric_dtype(dtype('S'))

odl/test/util/utility_test.py:34: AssertionError
______________________________ test_is_real_dtype ______________________________

    def test_is_real_dtype():
        for dtype in real_dtypes:
>           assert is_real_dtype(dtype)
E           AssertionError: assert False
E            +  where False = is_real_dtype(dtype('S'))

odl/test/util/utility_test.py:39: AssertionError
```
Does @kohr-h or @adler-j know why these fail? Versions of some relevant packages used on my local machine:
```
numpy                     1.11.2                   py35_0
pytest                    3.2.1                    py35_0  
pytest-cache              1.0                      py35_0  
pytest-pep8               1.0.6                    py35_0
```

odl/space/pspace.py
```diff
@@ -1081,6 +1081,32 @@ def ufuncs(self):
         """
         return ProductSpaceUfuncs(self)
 
+    @property
+    def real(self):
+        """Pointwise real part of the element."""
+        size = self.space.shape[0]
+        to_return = [None] * size
```
kohr-h: Can we call this `real_part` instead of `to_return`?

odl/space/pspace.py
```diff
@@ -1081,6 +1081,32 @@ def ufuncs(self):
         """
         return ProductSpaceUfuncs(self)
 
+    @property
+    def real(self):
+        """Pointwise real part of the element."""
+        size = self.space.shape[0]
```
kohr-h: `len(self)` should do.

odl/space/pspace.py
```diff
@@ -1081,6 +1081,32 @@ def ufuncs(self):
...
+        to_return = [None] * size
+        for (i, part) in zip(range(size), self.__parts):
```
kohr-h: `for i, part in enumerate(self.parts):`

odl/space/pspace.py
```diff
@@ -1081,6 +1081,32 @@ def ufuncs(self):
         """
         return ProductSpaceUfuncs(self)
 
+    @property
+    def real(self):
+        """Pointwise real part of the element."""
```
kohr-h: ~~Pointwise~~

kohr-h:
@aringh Weird that those tests fail. These are the lines that generate the dtypes to be tested:

https://github.com/odlgroup/odl/blob/53de0e9e978e4d63c87b1d49e5388f0566541e83/odl/test/util/utility_test.py#L18-L22

I have no idea why `np.dtype('S')` ends up in there. Can you check manually what you get on your machine with your Numpy version?

aringh:
I seem to have some extremely weird bug somewhere in my system... `np.sctypes.keys()` gives output 
`dict_keys(['float', 'int', 'uint', 'others', 'complex'])`, which looks fine. But then `np.sctypes['complex']` gives output
```
 dtype('S'),
 dtype('<U'),
 dtype('S'),
 dtype('<U'),
 dtype('S'),
 dtype('<U'),
 dtype('S'),
 dtype('<U'),
 dtype('S'),
 ...]
```
We can ignore this. I will update numpy, and if it doesn't work I will simply create a new environment.

odl/space/pspace.py
```diff
@@ -1081,6 +1081,32 @@ def ufuncs(self):
...
+        size = self.space.shape[0]
+        to_return = [None] * size
+        for (i, part) in zip(range(size), self.__parts):
+            to_return[i] = part.real
+        return self.space.element(to_return)
```
adler-j: This will be a problem since we want the real space, you need to implement `ProductSpace.real_space` and do
```python
return self.space.real_space.element(to_return)
```

odl/space/pspace.py
```diff
@@ -1081,6 +1081,32 @@ def ufuncs(self):
         """
         return ProductSpaceUfuncs(self)
 
+    @property
+    def real(self):
+        """Pointwise real part of the element."""
+        size = self.space.shape[0]
+        to_return = [None] * size
+        for (i, part) in zip(range(size), self.__parts):
+            to_return[i] = part.real
```
adler-j: Whats wrong with

```python
to_return = [part.real for part in self.parts]
```

odl/space/pspace.py
```diff
@@ -1064,6 +1081,32 @@ def ufuncs(self):
...
+        return self.space.element(to_return)
+
+    @property
+    def imag(self):
+        """Pointwise imaginary part of the element."""
+        size = self.space.shape[0]
+        to_return = [None] * size
+        for (i, part) in zip(range(size), self.__parts):
+            to_return[i] = part.imag
+        return self.space.element(to_return)
```
adler-j: See all points above

odl/space/pspace.py
```diff
@@ -1064,6 +1081,32 @@ def ufuncs(self):
...
+
+    def conj(self):
+        """Pointwise complex conjugate of the element."""
+        size = self.space.shape[0]
+        to_return = [None] * size
+        for (i, part) in zip(range(size), self.__parts):
+            to_return[i] = part.conj()
+        return self.space.element(to_return)
```
adler-j: See all points above

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,37 @@ def test_reductions():
...
+
+
+def test_real_imag_and_conj():
+    """..."""
+    space = odl.ProductSpace(odl.uniform_discr(0, 1, 3, dtype=complex),
+                             odl.cn(2))
+    x = noise_element(space)
+
+    # Test real
+    expected_result = space.element([np.real(x[0]), np.real(x[1])])
```
adler-j: use `x[0].real` instead

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,37 @@ def test_reductions():
...
+    assert x.real == expected_result
+
+    # Test imag
+    expected_result = space.element([np.imag(x[0]), np.imag(x[1])])
+    assert x.imag == expected_result
+
+    # Test conj. Note that ProductSpace does not implement asarray if
+    # is_power_space is false. Hence the construction below
+    expected_result = space.element([np.conj(x[0]), np.conj(x[1])])
+    assert all_almost_equal((x.conj())[0], expected_result[0])
```
adler-j: How about:

```
x_conj = x.conj()
assert all_almost_equal(x.conj[0], expected_result[0])
assert all_almost_equal(x.conj[1], expected_result[1])
```

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,37 @@ def test_reductions():
...
+    expected_result = space.element([np.real(x[0]), np.real(x[1])])
+    assert x.real == expected_result
+
+    # Test imag
+    expected_result = space.element([np.imag(x[0]), np.imag(x[1])])
+    assert x.imag == expected_result
+
+    # Test conj. Note that ProductSpace does not implement asarray if
+    # is_power_space is false. Hence the construction below
+    expected_result = space.element([np.conj(x[0]), np.conj(x[1])])
```
adler-j: Use `x[0].conj()`

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,37 @@ def test_reductions():
...
 
+def test_array_wrap_method():
+    """Verify that the __array_wrap__ method for NumPy works."""
+    space = odl.ProductSpace(odl.uniform_discr(0, 1, 10), 2)
+    x_arr, x = noise_elements(space)
+    y_arr = np.sin(x_arr)
+    y = np.sin(x)  # Should yield again an ODL product space element
+
+    assert all_equal(y, y_arr)
+    assert y in space
```
adler-j: Move this line to above the line before, since if `y not in space` we expect a fail, and this would be the natural order of testing it.

aringh:
This took quite some time to get it correct... but I hope that it is ok now. However, I have therefore note added setters yet, as requested by @kohr-h. I will try to do that.

aringh:
I think I would need some help: when implementing the setter for `.real`, should I care about if the element given to me have a non-zero imaginary part or not? Or should I just skip the imaginary part if it is none-zero?

aringh:
First try for setter for real part. If possible, could @adler-j or @kohr-h review it before I do the imaginary part? I think you will have comments :-p, and the imaginary part is same-same-but-different.

odl/space/pspace.py
```diff
@@ -1064,6 +1161,89 @@ def ufuncs(self):
...
+        """
+        if newreal in self.space.field:
+            for part in self.parts:
+                part.real.data[:] = newreal
+        elif newreal in self.space.real_space or newreal in self.space:
+            for part, newreal_subpart in zip(self.parts, newreal):
+                part.real.data[:] = newreal_subpart.real.data[:]
+        elif self.space.shape[0] == np.shape(newreal)[0]:
+            for part, newreal_subpart in zip(self.parts, newreal):
+                part.real.data[:] = np.real(newreal_subpart)
```
kohr-h: I wouldn't do all these checks, in particular don't extract the real part from the right-hand side. The backend will warn if such an assignment casts complex to float. IMO that's not quite enough since in most cases it's a sign of a bug, but we have to live with it. Taking the real part shadows such bugs.

So here just leave the RHS as it is.

odl/space/pspace.py
```diff
@@ -1064,6 +1161,89 @@ def ufuncs(self):
...
+        else:
+            raise ValueError('new real part does not the correct first '
+                             'dimension, expected {} but got {}'
+                             ''.format(self.space.shape[0],
+                                       np.shape(newreal)[0]))
```
kohr-h: Overall the logic seems a bit too complicated for the cases it handles, and it misses out one case. Here are the cases that I think should be distinguished:

1. RHS not iterable (likely a scalar) -> call each `part.real` setter with this RHS (don't check `field` or so)
2. Space is a power space and RHS an element of the base space -> call each `part.real` setter with this RHS (i.e., the same as in case 1, can be merged)
   The reason for this (and that it comes before the next case) is that Numpy arrays behave the same: implicit new axes are added *to the left*, where leftmost is the power of the power space in our case:
   ```python
   >>> a = np.zeros((2, 2))
   >>> a.real = [1, 2]
   >>> a
   array([[ 1.,  2.],
          [ 1.,  2.]])
   ```
   **Note:** This also needs to include *element-like* objects for the base space.
3. `len(rhs) == len(self)` -> call `part[i].real` setter with `rhs[i]`, i.e., pretty much the "standard" case.
4. Error

So an implementation of that logic would be:
```python
try:
    iter(newreal)
except TypeError:
    for part in self.parts:
        part.real = newreal
    return

if self.is_power_space:
    try:
        newreal = self.space[0].element(newreal)
    except (ValueError, TypeError):  # maybe need to catch more, e.g. `LinearSpaceTypeError`
        pass
    else:
        for part in self.parts:
            part.real = newreal
elif len(newreal) == len(self):
    for part, new_re in zip(self.parts, newreal):
        part.real = new_re
else:
    raise ValueError('invalid')  # better error
```

odl/space/pspace.py
```diff
@@ -381,6 +384,83 @@ def dtype(self):
...
+            If `dtype` is not a real or complex type.
+        """
+        if not is_numeric_dtype(self.dtype):
+            raise NotImplementedError(
+                '`real_space` not defined for non-numeric `dtype`')
+
+        if is_real_dtype(self.dtype):
+            return self
+        elif is_complex_floating_dtype(self.dtype):
+            new_dtype = TYPE_MAP_C2R.get(self.dtype, 'None')
```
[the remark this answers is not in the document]
aringh: My idea was indeed to just let `astype` raise the error after receiving `None`. But you are right, that might not be very helpful. I will raise a `ValueError` here (as well)

odl/space/pspace.py
```diff
@@ -381,6 +384,83 @@ def dtype(self):
...
+    @property
+    def real_space(self):
+        """The space corresponding to this space's real dtype.
+
+        Raises
+        ------
+        NotImplementedError
+            If `dtype` is not a numeric data type.
+        TypeError
+            If `dtype` is not a real or complex type.
```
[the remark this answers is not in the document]
aringh: I had to do some digging in the odl code to try to understand how this works, so indeed a lot of it is inspired or taken from `base_tensor.py`. I will do the change there as well.

odl/space/pspace.py
aringh: Your logic is much cleaner than mine, no doubt about that ;-) However, it passes all the current tests except when `newreal = [[0, 1 + 1j, 2], [3, 4]]`. Now, that a test for setting the real part crashes when the element we try to set it to is complex might seem logical. But what I do not like is that this is the ONLY test with complex values where it crashes. All other tests pass (with warnings of course). Should we care about this inconsistency, or just be happy that it indeed actually captures one error correctly and not just gives warnings in this case?

kohr-h: Well, this implementation can have a quite laid-back attitude and just distribute work down in the way that makes most sense. That also means that raising errors is up to lower layers, so I wouldn't worry about that too much.
Regarding warnings vs. errors, we've been more and more going along with Numpy on most parts, simply because they probably have thought through a bunch of things and solved issues we don't even know about... and it's convenient! 😄 

Anyway, what is this one error in the test?

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,110 @@ def test_reductions():
...
+    x = noise_element(space)
+    x_imag = x.imag
+
+    # Test for element in ProductSpace where imaginary part is zero
+    tmp = noise_element(space)
+    newreal = space.element(tmp.real)
+    x.real = newreal
+    assert x.real == newreal.real
+
+    # Test for element in ProductSpace with none-zero imaginary part
```
kohr-h: TBH, I wouldn't even test with complex right-hand sides. It's ultimately up to the backend (here Numpy) how to handle this, we don't really define ourselves what should happen in this case.

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,110 @@ def test_reductions():
...
+    expected_result = (newreal * space.one()).real
+    assert x.real == expected_result
+
+    # Assert that imaginary part has not changed in any of the tests
+    assert x.imag == x_imag
+
+    # Test that wrong first dimension do not pass
+    newreal = np.random.randn(5, 1)  # space has 5 entries in total
+    with pytest.raises(ValueError):
+        x.real = newreal
```
kohr-h: Can you parametrize this test? It's a lot of repetition, I think it wouldn't be too hard to make a fixture for this.

Also, there's no test for the power space case with broadcasting yet, that would be an important one.

odl/space/pspace.py
aringh: This fails
```python
import odl
space = odl.ProductSpace(odl.uniform_discr(0, 1, 3, dtype=complex), odl.cn(2))
x = odl.util.testutils.noise_element(space)
newreal = [[0, 1 + 1j, 2], [3, 4]]
x.real = newreal
```
and gives
```

  File "<ipython-input-3-ede4021b455c>", line 5, in <module>
    x.real = newreal

  File "[...]/odl/space/pspace.py", line 1249, in real
    part.real = new_re

  File "[...]/odl/discr/lp_discr.py", line 615, in real
    self.tensor.real = newreal

  File "[...]/odl/space/npy_tensors.py", line 1251, in real
    self.real.data[:] = newreal

TypeError: can't convert complex to float
```
while for example the following passes
```python
import odl
space = odl.ProductSpace(odl.uniform_discr(0, 1, 3, dtype=complex), odl.cn(2))
x = odl.util.testutils.noise_element(space)
newreal = odl.util.testutils.noise_element(space)
x.real = newreal
```

aringh: In fact, just chaining to a power space makes the first code snippet pass:
```
import odl
space = odl.ProductSpace(odl.uniform_discr(0, 1, 3, dtype=complex), 2)
x = odl.util.testutils.noise_element(space)
newreal = [[0, 1 + 1j, 2], [3, 4, 5]]
x.real = newreal
```

odl/test/space/pspace_test.py
```diff
@@ -900,5 +900,110 @@ def test_reductions():
...
+    space = odl.ProductSpace(odl.uniform_discr(0, 1, 3, dtype=complex),
+                             odl.cn(2))
+    x = noise_element(space)
+    x_imag = x.imag
+
+    # Test for element in ProductSpace where imaginary part is zero
+    tmp = noise_element(space)
+    newreal = space.element(tmp.real)
+    x.real = newreal
+    assert x.real == newreal.real
```
kohr-h: Looking ahead a bit: we've decided that we are going to become more like Numpy also for stuff like `x == y`, which means that code like this will break (see #1249).
I suggest that we proactively use `all_equal` for this case as well.

odl/space/pspace.py
kohr-h: For me all 3 versions work, which may be a matter of Numpy version. So my conclusion is that we should just let Numpy (or whatever backend we use) do what it does, and unless there's a fundamental difference we don't add new behavior. That also means less work for you 😉

aringh:
New try for the setter or  `real` from `ProductSpace`. Also rebased on master. If this is (at least almost) good to go, I will make a copy-past implementation for `.imag` as well ;-)

odl/space/pspace.py
```diff
@@ -381,6 +384,107 @@ def dtype(self):
...
+            Raised if the underlying spaces have different `dtype`, if `dtype`
+            is not a numeric data type, if `dtype` is not a real or complex
+            type, and if `dtype` is complex but cannot be mapped to a real
+            type.
+        """
+        try:
+            current_dtype = self.dtype
+        except(AttributeError):
+            raise ValueError(
+                'dtypes of underlying spaces are not the same and therefore '
```
adler-j: Well, technically I could imagine that a space of mixed complex64 and complex128 spaces would yield a space of mixed float32/64.

Wouldn't a easier implementation of all of this be to remove all logic and simply do:

```
real_spaces = [spc.real_space for doc in self]
return ProductSpace(*real_spaces)
```

Or does the logic add something to this?

odl/space/pspace.py
```diff
@@ -381,6 +384,107 @@ def dtype(self):
...
+            # Need to filter this out since Numpy iterprets it as 'float'
+            raise ValueError('`None` is not a valid data type')
+
+        dtype = np.dtype(dtype)
+        current_dtype = getattr(self, 'dtype', None)
+
+        if dtype == current_dtype:
+            return self
+        else:
+            return ProductSpace(*[space.astype(dtype)
```
adler-j: This is exactly what I meant above!

odl/space/pspace.py
aringh: You are absolutely right. I think this is what's called bootstrapping, no? :-)

odl/space/pspace.py
```diff
@@ -1064,6 +1123,158 @@ def ufuncs(self):
...
+        True
+        """
+        imag_part = [part.imag for part in self.parts]
+        return self.space.real_space.element(imag_part)
+
+    @imag.setter
+    def imag(self, newimag):
+        """Setter for the imaginary part.
+
+        This method is invoked by ``x.imag = other``.
```
aringh: Here, something like `x.imag = 1j` could to a user seem like a natural way of setting the imaginary part to `1`. However, in this case it will be set to `0`, since the imaginary part is a real number and thus set to the real part of `0 + 1j`. Although logical mathematically, it might be confusing before you think about it. Any suggestion on how this can be clarified? Does it need to be clarified?

odl/space/pspace.py
```diff
@@ -1064,6 +1123,158 @@ def ufuncs(self):
...
+        True
+        """
+        real_part = [part.real for part in self.parts]
+        return self.space.real_space.element(real_part)
+
+    @real.setter
+    def real(self, newreal):
+        """Setter for the real part.
+
+        This method is invoked by ``x.real = other``.
```
aringh: Related to the comment below, should we mention what happens with the imaginary part of an element which one tries to set as real part, i.e., if someone does something like `x.real = 1 + 1j`?

odl/space/pspace.py
adler-j: No, it does not need to be clarified since these are somewhat standard from e.g. numpy and other packages and hence easily googleable if anyone wants more info

odl/space/pspace.py
adler-j: In my opinion it should throw an error. Is there a reason it does not? Have you tested what numpy does, since we want to do the exact same.

adler-j:
```python
>>> arr = np.zeros(5)
>>> arr.imag = 3
TypeError: array does not have imaginary part to set
>>> arr.real = 2j
__main__:1: ComplexWarning: Casting complex values to real discards the imaginary part
```

this is what we expect basically

odl/space/pspace.py
kohr-h: Agreed, we should assume at least this basic amount of knowledge of complex numbers. I don't think `x.imag = 1j` should ever be expected to set the imaginary part to 1.

odl/space/pspace.py
kohr-h: I'd even say that the exact behavior depends on the backend, and that we shouldn't add logic that interferes with that. So IMO we pass on whatever we get and let the error or warning bubble up.

aringh:
@kohr-h and @adler-j both still require changes ;-) Good to go?

odl/space/pspace.py
```diff
@@ -381,6 +384,45 @@ def dtype(self):
...
             raise AttributeError("`dtype`'s of subspaces not equal")
 
+    @property
+    def real_space(self):
+        """Variant of this space with real dtype."""
+        return ProductSpace(*[space.real_space for space in self.spaces])
+
+    @property
+    def complex_space(self):
+        """The space corresponding to this space's complex dtype."""
```
kohr-h: Use same formulation as in `real_space`

odl/space/pspace.py
```diff
@@ -381,6 +384,45 @@ def dtype(self):
...
+        """
+        if dtype is None:
+            # Need to filter this out since Numpy iterprets it as 'float'
+            raise ValueError('`None` is not a valid data type')
+
+        dtype = np.dtype(dtype)
+        current_dtype = getattr(self, 'dtype', None)
+
+        if dtype == current_dtype:
+            return self
```
kohr-h: There's a bug here since 
```python
>>> np.dtype(float) == None
True
```
which means that `self` is returned if `dtype == np.dtype(float)` and no dtype of the current space can be determined. That's why I suggested `object`, not `None`, as `getattr` fallback (which assumes we will never support `object`).

odl/space/pspace.py
```diff
@@ -1006,6 +1048,23 @@ def __array__(self):
         """
         return self.asarray()
 
+    def __array_wrap__(self, array):
+        """Return a new product space element wrapping the ``array``.
+
+        Only available if `is_power_space` is True.
```
kohr-h: \`\`True\`\`
or alternatively \`\`is_power_space is True\`\`

odl/space/pspace.py
```diff
@@ -1064,6 +1123,158 @@ def ufuncs(self):
...
+            Values to be assigned to the real part of this element.
+
+        Examples
+        --------
+        >>> space = odl.ProductSpace(odl.cn(3), odl.cn(2))
+        >>> x = space.element([[1 + 1j, 2, 3 - 3j], [1 + 1j, 2]])
+        >>> zero = space.real_space.zero()
+        >>> x.real = zero
+        >>> x
+        space.element([[1j, 0j, -3j], [1j, 0]])
```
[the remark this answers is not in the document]
aringh: You are right... but it seems like doc-tests are not run for setters because it does not fail. And neither does
```python
>>> False
True
```
I will try to have a look at it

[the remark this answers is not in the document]
adler-j: This is apparently a "feature" and all the documentation should go into the getter

https://stackoverflow.com/questions/23341106/python-2-7-doctests-ignored-in-setter-method-of-a-class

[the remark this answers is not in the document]
adler-j: I'll make a general issue for this since we seem to have this error in many places.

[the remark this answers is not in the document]
aringh: So apparently it seems like the doc-strings of `.setter` is ignored, see [here](https://stackoverflow.com/questions/23341106/python-2-7-doctests-ignored-in-setter-method-of-a-class). I will simply put the corresponding doc-test in the property itself, while keeping the rest of the documentation in the setter

odl/space/pspace.py
adler-j: If this is never exposed externally, prefer e.g. `-1` as a fallback value since we know for sure we'll never support that.

kohr-h: -1 raises an error that we would need to catch. Alternatively we could use `np.void` which is about the only thing that compares to `False` with `np.dtype(object)`.

PR #1293, opened 2018-02-13
ENH: make pspace op matrix generation more robust

kohr-h:
This also invalidates the syntax `ProductSpaceOperator([A, B, C])` (with a 1D sequence), which I think was not valid in the first place, according to the documentation. It may break other stuff, let's see.

### DIFF
--- a/odl/test/operator/pspace_ops_test.py
+++ b/odl/test/operator/pspace_ops_test.py
@@ -10,35 +10,48 @@ from __future__ import division
 import pytest
 
 import odl
-from odl.util.testutils import all_almost_equal
+from odl.util.testutils import all_almost_equal, simple_fixture
 
 
-def test_pspace_op_init():
-    r3 = odl.rn(3)
-    I = odl.IdentityOperator(r3)
+base_op = simple_fixture(
+    'base_op',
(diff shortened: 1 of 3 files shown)


odl/test/operator/pspace_ops_test.py
```diff
@@ -10,35 +10,48 @@
...
-def test_pspace_op_init():
-    r3 = odl.rn(3)
-    I = odl.IdentityOperator(r3)
+base_op = simple_fixture(
+    'base_op',
+    [odl.IdentityOperator(odl.rn(3)),
+     odl.BroadcastOperator(odl.IdentityOperator(odl.rn(3)), 2),
+     odl.ReductionOperator(odl.IdentityOperator(odl.rn(3)), 2),
+     odl.DiagonalOperator(odl.IdentityOperator(odl.rn(3)), 2),
+     ],
```
adler-j: What does `ProductSpaceOperator(ProductSpaceOperator)` do?

kohr-h: That's a very good question. It fails since `pspace_op[0]` is a `ReductionOperator`, and that is caught in the conversion function. IMO that's OK.

PR #1302, opened 2018-02-19
DOC: Improve doc of property setters, closes #1297

### LINKED ISSUE #1297: Proper documentation of property.setter
As noticed by @aringh, `property.setter` methods do not really have any documentation, as is mentioned here:

https://stackoverflow.com/questions/23341106/python-2-7-doctests-ignored-in-setter-method-of-a-class

A side-effect of this is that doctests are only run on the getter.

So basically this means we should move any documentation from the setter into the getter.

adler-j:


odl/space/npy_tensors.py
```diff
@@ -1205,10 +1205,30 @@ def real(self):
...
+        Set the real part:
+
+        >>> space = odl.cn(3)
+        >>> x = space.element([1 + 1j, 2, 3 - 3j])
+        >>> zero = odl.rn(3).zero()
+        >>> x.real = zero
+        >>> x
+        cn(3).element([1j, 0j, -3j])
+
+        Can also set with other array-like types and broadcasting:
```
kohr-h: Use wording as in `lp_discr.py`

odl/space/npy_tensors.py
```diff
@@ -1262,10 +1264,30 @@ def imag(self):
...
+        cn(3).element([(1+0j), (2+0j), (3+0j)])
+
+        Other array-like types and broadcasting:
+
+        >>> x.imag = 1.0
+        >>> x
+        cn(3).element([(1+1j), (2+1j), (3+1j)])
+        >>> x.imag = [2, 3, 4]
+        >>> x
+        cn(3).element([(1+2j), (2+3j), (3+4j)])
```
kohr-h: This example has the best text headers IMO, use this style also in the other places.

adler-j:
Fixed the comments and the failing tests, merge after CI.

PR #1303, opened 2018-02-19
Issue 1296  slow backend

adler-j:


odl/tomo/operators/ray_trafo.py
```diff
@@ -125,8 +125,22 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
                 impl = 'astra_cuda'
             elif ASTRA_AVAILABLE:
                 impl = 'astra_cpu'
+                if reco_space.size >= 512 ** 2:
+                    warnings.warn(
+                        "The best available backend ('astra_cpu') may be too "
+                        "slow for volumes of this size. Consider installing "
+                        "e.g. 'astra_cuda'. This warning can be disabled by "
+                        "explicitly setting `impl='astra_cpu'`.",
+                        RuntimeWarning)
```
kohr-h: Why warn here? ASTRA is available, but not ASTRA CUDA, so probably there's little the user can do about it.

odl/tomo/operators/ray_trafo.py
```diff
@@ -340,8 +354,8 @@ def __init__(self, domain, geometry, **kwargs):
             - ``'skimage'``: scikit-image, only 2D parallel with square
               reconstruction space.
 
-            For the default ``None``, the fastest available back-end is
-            used.
+            For the default ``None``, the backend is chosen in the order listed
+            above, e.g. the fastest available back-end is used.
```
kohr-h: e.g.?
I'd use the old formulation and add ", tried in the above order."

odl/tomo/operators/ray_trafo.py
adler-j: The whole idea here is that they perhaps need to use another computer. I've had several users on laptops trying way to large problems, warning here lets them know that they should perhaps use a stationary computer instead.

odl/tomo/operators/ray_trafo.py
adler-j: That also works!

odl/tomo/operators/ray_trafo.py
kohr-h: Still I feel that forcing people to silence warnings by changing their code is not the way to go. If something runs way too slow they will notice, right? We can still place a prominent notice in the `RayTransform` doc, with a `note::` block, for example.

adler-j: Well, the whole reason I made this issue was basically a user report on *exactly* this problem. The user noticed that the code was running slowly, fine, but for a new user to go from that to realizing that "oh, it must be the backend of the ray transform" is highly nontrivial. For you/me, sure, but for someone new they might just expect that ODL simply is that slow, how would they know?

And sadly, noone reads the doc :/

kohr-h: If I read it correctly, the user report was for the case we silently fell back to `skimage`, and I don't debate that warning. But telling users that they should use ASTRA CUDA when a lot of them will say "I know I should, but I don't have a GPU!" goes too far IMO. 

Also, who says that 512^2 is "too big" for the CPU version? Maybe CPUs will get faster, or the CPU version of the projector gets a speed-up and the threshold needs to be adapted.

adler-j: The problem we're aiming at here is more the "corner case" of the GPU installation failing, which for anyone who has helped people with this is at least 50% of the cases. E.g. the user forgot to install CUDA before ASTRA. And frankly, how many users are running 512+ cases on a CPU on purpose?

W.r.t. the size, I picked it since my relatively decent computer took several seconds for this size, and 512 is far above "simple experimentation" level and closer to actual datasets. But sure, we could change it to 1024, which will stay far outside of CPU scope for at least several years.

kohr-h: > E.g. the user forgot to install CUDA before ASTRA.

Wouldn't that lead to a crash? Or does ASTRA handle that and report back that CUDA cannot be used? Not long ago it used to always say CUDA was available.



> And frankly, how many users are running 512+ cases on a CPU on purpose?

Alright, alright, I give in. Maybe it's not fun to wait for 5 seconds per forward projection anyway.. But then I at least need to complain about the message :grin:

odl/tomo/operators/ray_trafo.py
```diff
@@ -125,8 +125,22 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
                 impl = 'astra_cuda'
             elif ASTRA_AVAILABLE:
                 impl = 'astra_cpu'
+                if reco_space.size >= 512 ** 2:
+                    warnings.warn(
+                        "The best available backend ('astra_cpu') may be too "
+                        "slow for volumes of this size. Consider installing "
```
kohr-h: using

odl/tomo/operators/ray_trafo.py
```diff
@@ -125,8 +125,22 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+                        "e.g. 'astra_cuda'. This warning can be disabled by "
```
kohr-h: ... if your machine has an Nvidia GPU

adler-j:
Merged this without FFTW, would fix that later

kohr-h:
> would fix that later

New issue?

PR #1305, opened 2018-02-21
Bregman tv example

aringh:
A small Bregman-TV example in tomography.

However, I am not to convinced by the performance of the Bregman-TV algorithm on this example. Moreover, I would appreciate if anyone could double check that I have understood the idea behind iterative Bregman reconstruction correctly. Maybe @ozanoktem could have a look?

### DIFF
--- a/examples/solvers/bregman_tv_tomography.py
+++ b/examples/solvers/bregman_tv_tomography.py
@@ -1,91 +1,125 @@
 """Example of Bregman-TV in tomography.
 
-The example does five iterations in a Bregman-TV scheme, applied to a
-tomography problem.
+The example solves an inverse problem in tomography using Bregman iterations
+for a total variation (TV) regularized problem. To this end, let ``A`` denote
+the forward operator, ``g`` denote the data, ``D(Ax, g)`` denote data
+discrepancy functional, and ``alpha * R(x)`` denote the regularizing
+functional. In this particular example, the variational problem considered is
+the TV-problem
+
+    min_x  1/2 ||A(x) - g||_2^2 + lam || |grad(x)| ||_1.
+
+For a functional ``F`` the Bregman functional is defined as
+
+    D_F^p(x, z) = F(x) - F(z) - <p, x-z>,
+
+where ``p`` is a subgradient at the point ``x``, i.e., ``p in dF(x)``.
+
+The Bregman iterations can be written as
+
+    x_{k+1} in argmin D(Ax, g) + lam * D_R^{p_k}(x, x_k)
+    p_{k+1} - p_k in -(1/lam) * A^* dD(Ax_k, g).
+
+Note that the first step involves solving an optimization problem, which is
+(normally) done using an iterative method. This means that the solution method
+contains a set of inner and outer iterations.
+
+The Bregman iterations need to be stopped after a finite number of outer
+iterations. Otherwise the noise from ``g`` will start to deteriorate the
+result. In this example this is illustrated by doing a few outer iterations to
+many, and displaying the result after each outer iteration.
+
+For more details on Bregman iterations, see, e.g., [Bur2016].
+
+Referenses
+----------
+[Bur2016] Burger, M. *Bregman distances in inverse problems and partial
+differential equations*. In: Hiriart-Urruty, J B, Korytowski A, Maurer H,
+Szymkat M. Advances in Mathematical Modeling, Optimization and Optimal Control.
+Springer, 2016, pp 3--33.
+https://doi.org/10.1007/978-3-319-30785-5_2
+https://arxiv.org/abs/1505.05191
 """
 
 import odl
 import numpy as np
 
-# Discrete reconstruction space
+# Reconstruction space
 reco_space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20],
-                               shape=[300, 300], dtype='float32')
+                               shape=[128, 128], dtype='float32')
 
-# Make a parallel beam geometry with flat detector
-angle_partition = odl.uniform_partition(0, np.pi, 100)
-detector_partition = odl.uniform_partition(-30, 30, 558)
-geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+# Make a parallel beam geometry with flat detector, and create ray transform
+geometry = odl.tomo.parallel_beam_geometry(reco_space,  num_angles=100)
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry)
 
-# Create the ray transform
-ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
-
-
-# Create phantom
+# Create phantom, forward project to create sinograms, and add 10% noise
 discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
-
-# Create sinogram of forward projected phantom with noise
 noise_free_data = ray_trafo(discr_phantom)
 noise = odl.phantom.white_noise(ray_trafo.range)
-noise = noise * 1/noise.norm() * noise_free_data.norm() * 0.10
+noise *= 0.10 / noise.norm() * noise_free_data.norm()
 data = noise_free_data + noise
 
-
-# Do a FBP-reconstruction
-fbp_op = odl.tomo.fbp_op(ray_trafo, filter_type='Hann', frequency_scaling=0.4)
-fbp_reco = fbp_op(data)
-
-
-# Display phantom, FBP reconstruction, and sinograms
-discr_phantom.show(title='Phantom')
-fbp_reco.show(title='FBP reconstruction')
-data.show(title='Simulated data (Sinogram)')
-
-
-# Define components of the variational problem
-reg_param = 0.3
-gradient = odl.Gradient(reco_space)
-
-# Column vector of two operators
-op = odl.BroadcastOperator(ray_trafo, gradient)
-
-# Do not use the g functional, set it to zero
-g = odl.solvers.ZeroFunctional(reco_space)
-
-# l2-squared data matching
+# Components for variational problem: l2-squared data matching and isotropic
+# TV-regularization
 l2_norm = 0.5 * odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+gradient = odl.Gradient(reco_space)
+reg_param = 0.3
+l12_norm = reg_param * odl.solvers.GroupL1Norm(gradient.range)
 
-# Isotropic TV-regularization
-l1_norm = odl.solvers.GroupL1Norm(gradient.range)
+# Assemble functionals and operators for the optimization algorithm
+f = odl.solvers.ZeroFunctional(reco_space)  # No f functional used, set to zero
+g = [l2_norm, l12_norm]
+L = [ray_trafo, gradient]
 
-# Estimated operator norm, add 10 percent to ensure ||K||_2^2 * sigma * tau < 1
-op_norm = 1.1 * odl.power_method_opnorm(op)
+# Estimated operator norms, which are used to ensure that we fulfill the
+# convergence criteria of the optimization algorithm
+ray_trafo_norm = odl.power_method_opnorm(ray_trafo, maxiter=20)
+gradient_norm = odl.power_method_opnorm(gradient, maxiter=20)
 
-niter = 500  # Number of iterations in the inner loop
-tau = 1.0 / op_norm  # Step size for the primal variable
-sigma = 1.0 / op_norm  # Step size for the dual variable
+# Parameters for the optimization method; tuned in order to reduce the number
+# of inner iterations needed to solve the first step in the Bregman iterations
+niter_inner = 200
+tau = 0.01  # Step size for the primal variable
+sigma_ray_trafo = 45.0 / ray_trafo_norm ** 2  # Step size for dual variable
+sigma_gradient = 45.0 / gradient_norm ** 2  # Step size for dual variable
+sigma = [sigma_ray_trafo, sigma_gradient]
 
-# Create initial guess for the solver.
+# The reconstruction looks nice after about 5 outer iterations; set total
+# number of outer iterations to 7 to show what happens if one does to many
+niter_bregman = 7
+
+# Create initial guess and initial subgradient
 x = reco_space.zero()
+p = reco_space.zero()
 
-# This defines the Bregman iterations
-for breg_iter in range(5):
+# This defines the outer Bregman iterations
+for breg_iter in range(niter_bregman):
     print('Outer Bregman iteration: {}'.format(breg_iter))
 
-    # Define the bregman functional with base point in the previous iterate
-    bregman_func = reg_param * l1_norm.bregman(gradient(x))
+    # Create the affine part of the Bregman functional
+    constant = l12_norm(gradient(x))
+    linear_part = reg_param * odl.solvers.QuadraticForm(vector=-p,
+                                                        constant=constant)
 
-    # Combine functionals, order must correspond to the operator K
-    f = odl.solvers.SeparableSum(l2_norm, bregman_func)
-
-    # Used to print iteration number.
     callback_inner = odl.solvers.CallbackPrintIteration(step=50)
 
-    # Call the solver. x is updated in-place with the consecutive iterates.
-    odl.solvers.pdhg(x, f, g, op, tau=tau, sigma=sigma, niter=niter,
-                     callback=callback_inner)
+    # Inner iterations; x is updated in-place with the consecutive iterates
+    odl.solvers.forward_backward_pd(
+        x=x, f=f, g=g, L=L, h=linear_part, tau=tau, sigma=sigma,
+        niter=niter_inner, callback=callback_inner)
+
+    # Update the subgradient
+    p -= (1 / reg_param) * ray_trafo.adjoint(l2_norm.gradient(ray_trafo(x)))
 
     # Display the result after this iteration
     x.show(title='Outer Bregman iteration {}'.format(breg_iter),
            force_show=True)
-    x.show(title='Outer Bregman iteration {}, pixel values [0,1]'
-           ''.format(breg_iter), clim=[0, 1], force_show=True)
+
+# Create an FBP-reconstruction to compare with
+fbp_op = odl.tomo.fbp_op(ray_trafo, filter_type='Hann', frequency_scaling=0.4)
+fbp_reco = fbp_op(data)
+fbp_reco.show(title='FBP reconstruction')
+
+# Finally, also display phantom and sinograms
+discr_phantom.show(title='Phantom')
+data.show(title='Simulated data (Sinogram)')
--- a/examples/solvers/bregman_tv_tomography.py
+++ b/examples/solvers/bregman_tv_tomography.py
@@ -1,0 +1,91 @@
+"""Example of Bregman-TV in tomography.
+
+The example does five iterations in a Bregman-TV scheme, applied to a
+tomography problem.
+"""
+
+import odl
+import numpy as np
+
+# Discrete reconstruction space
+reco_space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20],
+                               shape=[300, 300], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+angle_partition = odl.uniform_partition(0, np.pi, 100)
+detector_partition = odl.uniform_partition(-30, 30, 558)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# Create the ray transform
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+
+
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+noise_free_data = ray_trafo(discr_phantom)
+noise = odl.phantom.white_noise(ray_trafo.range)
+noise = noise * 1/noise.norm() * noise_free_data.norm() * 0.10
+data = noise_free_data + noise
+
+
+# Do a FBP-reconstruction
+fbp_op = odl.tomo.fbp_op(ray_trafo, filter_type='Hann', frequency_scaling=0.4)
+fbp_reco = fbp_op(data)
+
+
+# Display phantom, FBP reconstruction, and sinograms
+discr_phantom.show(title='Phantom')
+fbp_reco.show(title='FBP reconstruction')
+data.show(title='Simulated data (Sinogram)')
+
+
+# Define components of the variational problem
+reg_param = 0.3
+gradient = odl.Gradient(reco_space)
+
+# Column vector of two operators
+op = odl.BroadcastOperator(ray_trafo, gradient)
+
+# Do not use the g functional, set it to zero
+g = odl.solvers.ZeroFunctional(reco_space)
+
+# l2-squared data matching
+l2_norm = 0.5 * odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+
+# Isotropic TV-regularization
+l1_norm = odl.solvers.GroupL1Norm(gradient.range)
+
+# Estimated operator norm, add 10 percent to ensure ||K||_2^2 * sigma * tau < 1
+op_norm = 1.1 * odl.power_method_opnorm(op)
+
+niter = 500  # Number of iterations in the inner loop
+tau = 1.0 / op_norm  # Step size for the primal variable
+sigma = 1.0 / op_norm  # Step size for the dual variable
+
+# Create initial guess for the solver.
+x = reco_space.zero()
+
+# This defines the Bregman iterations
+for breg_iter in range(5):
+    print('Outer Bregman iteration: {}'.format(breg_iter))
+
+    # Define the bregman functional with base point in the previous iterate
+    bregman_func = reg_param * l1_norm.bregman(gradient(x))
+
+    # Combine functionals, order must correspond to the operator K
+    f = odl.solvers.SeparableSum(l2_norm, bregman_func)
+
+    # Used to print iteration number.
+    callback_inner = odl.solvers.CallbackPrintIteration(step=50)
+
+    # Call the solver. x is updated in-place with the consecutive iterates.
+    odl.solvers.pdhg(x, f, g, op, tau=tau, sigma=sigma, niter=niter,
+                     callback=callback_inner)
+
+    # Display the result after this iteration
+    x.show(title='Outer Bregman iteration {}'.format(breg_iter),
+           force_show=True)
+    x.show(title='Outer Bregman iteration {}, pixel values [0,1]'
+           ''.format(breg_iter), clim=[0, 1], force_show=True)


ozanoktem:
I have together with Martin Benning looked at the example and Martin sees some issues that can explain the poor results.
 
1. It seems that the sub gradient in the Bregman iterates is not updated, probably it stays to zero. Consequently, assuming constant initialisation as a sub-gradient, the first Bregman iterate is according to theory, but not the following ones. The correct updates are of the form as in Algorithm 1, page 32 in [https://arxiv.org/abs/1801.09922](https://arxiv.org/abs/1801.09922).

2. It seems easiest to have a functionality to update the sub-gradient at the discretion of the user. It is unclear from the provided example whether this is possible as indicated above.

adler-j:
Is there an open reference implementation that we could look at to find differences?

ozanoktem:
Martin Benning will provide clean Matlab code illustrating the algorithm, I will make sure it is available.

aringh:
Great! Thank you for the feedback.

Now, correct me if I am wrong, but this how I have understood it: to approximate a solution to the problem `min f(x) + J(x)`, a solution which has some certain nice features, we use Bregman iterations and stop after a few iterations.

For the Bregman iterations each iterate `x_{k+1}` is obtained by solving `min f(x) + D_J^{p_k}(x, x_k)`, where `D_J^{p_k}` is the Bregman distance of the functional `J` in the base point `x_k` and using the subgradient `p_k`. As I have understood one want the subgradient `p_{k}` to be a subgradient of `J` in the base point `x_k`, correct?

In an iterative scheme I can compute the new subgradient `p_{k+1}` using the previous subgradient, as given in the algorithm on page 32. But if I have a way to evaluate the subgradient of the functional `J`, am I still forced to use this update rule?

What the current code does is simply to in each step of the Bregman iterations initialize a new Bregman functional at the new base point `x_k` (the code line `bregman_func = reg_param * l1_norm.bregman(gradient(x))`). This initialization of course includes computing a subgradient of the functional in the given base point. Since the subgradient operator is not necessarily single valued, this may of course differ from the one in the algorithm in page 32. But is there a reason why the algorithm performance should deteriorate in this case?

Or, have I misunderstood how the Bregman iterations work?

ozanoktem:
Attached MATLAB code is from Martin Benning and it is used to create the deconvolution Bregman example in Section 6.1 in [https://arxiv.org/abs/1801.09922](https://arxiv.org/abs/1801.09922). You only need to unzip the content of the attached file into a folder, add the sub-folders to the Matlab path (either manually or preferably with the file setpath.m) and run the Matlab live script deconvolution.mlx in the folder `Examples/Deconvolution.` This file also contains step-by-step comments.

[BregmanTV_example.zip](https://github.com/odlgroup/odl/files/1800120/BregmanTV_example.zip)

ozanoktem:
Dear @aringh ,

You are correct in assuming that you can select sub-gradients different from how its been done in, for example, [https://arxiv.org/abs/1801.09922](url). If the underlying functional is differentiable, you do not even need to select one at all, as the sub-differential is single-valued. However, many theoretical guarantees (such as Lemma 6.1 in [https://arxiv.org/abs/1801.09922](url)) rely on the specific sub-gradient-descent-type update. This could therefore explain why your results appear to be different compared to Bregman iterations with the conventional update. 
In addition to that, I'm wondering how you pick a specific sub-gradient. Do you use something like p = -div sign( nabla u), where sign is a vector-valued sign function that returns zero for each pixel where nabla u(x) = 0? 

Cheers,

Martin

mbenning:
Oops, I accidentally was still signed in as Ozan. Responses and complaints about the previous post have to be redirected to me ;-)

adler-j:
Is there any chance that we could put some effort into finishing this? I'd love to have a proper bregman example inside ODL.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,91 @@
...
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
+
+
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+noise_free_data = ray_trafo(discr_phantom)
+noise = odl.phantom.white_noise(ray_trafo.range)
+noise = noise * 1/noise.norm() * noise_free_data.norm() * 0.10
```
mehrhardt: small thing, but why not `noise *= (1/noise.norm() * noise_free_data.norm() * 0.10)`?

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,91 @@
...
+reco_space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20],
+                               shape=[300, 300], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+angle_partition = odl.uniform_partition(0, np.pi, 100)
+detector_partition = odl.uniform_partition(-30, 30, 558)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
+
+# Create the ray transform
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl='astra_cuda')
```
mehrhardt: I suppose these examples should not have CUDA hardcoded in there?

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,91 @@
...
+The example does five iterations in a Bregman-TV scheme, applied to a
+tomography problem.
+"""
+
+import odl
+import numpy as np
+
+# Discrete reconstruction space
+reco_space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20],
+                               shape=[300, 300], dtype='float32')
```
mehrhardt: to me this is way too "large" for a simple example.

mehrhardt:
Looks OK to me how you did it. Could be more elegant but there is no simple obvious choice how to do it. Why did you not implement the subgradient update as you quoted in your post?

aringh:
Thank you for the feedback! The reason I have not done it as I described above is because I am do not know how to do it :wink: 

The update of the subgradient as above will give me a subgradient that lives in the same space as the image (technically the dual space, but in this discretized setting it will effectively be something with the same discretization as the image). However, as the code is currently written, my Bregman functional is defined only for the L1-part of the regularize, and this means that the domain of this functional is the product space of the image space with itself. Therefore, the subgradient I need for the Bregman functional needs to live in this space, and I have not been able to figure out how to go from a subgradient `p`, which is a subgradient of the composition of the L1-norm and the spatial gradient by its construction, to a subgradient for only the L1-norm.

The reason for constructing the Bregman functional only for the L1-part is that I want to do operator splitting when I solve the nonsmooth problem.

I can relatively easy fix this example to a working Bregman-TV example, but that is _without_ using the `.bregman`-functionality. And that is not really what I would like, since if I cannot do an easy example with Bregman-TV using this functionality, why do we have it?

aringh:
Well, that's good! :smiley: Then maybe I can just change this to not use the `.bregman` -functionality. In that case it should not be to hard.

aringh:
So, I have:

1. squashed the three previous commits, since the two last ones were just minor adjustments.
2. changed the example to not use the `.bregman`-functionality, but instead I manually keep track of the linear part. After some tuning, I think that this is now an ok example.
3. rebased on master.

The example is now up for review.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Isotropic TV-regularization
+l1_norm = odl.solvers.GroupL1Norm(gradient.range)
+
+# Assembel functionals and operators for the optimization solver
+f = odl.solvers.ZeroFunctional(reco_space)  # Don't use f functional, set zero
+g = [l2_norm, reg_param*l1_norm]
+L = [ray_trafo, gradient]
+
+# Estimated operator norms, add 10 percent to ensure that we fulfill the
```
aringh: No longer add 10%, update comment according to code.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
+"""Example of Bregman-TV in tomography.
+
+The example does seven iterations in a Bregman-TV scheme, applied to a
```
aringh: Potentially add a longer description of the example? At least add a reference.

aringh:
This one is up for review if anyone is interested and have time :smiley:

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+noise_free_data = ray_trafo(discr_phantom)
+noise = odl.phantom.white_noise(ray_trafo.range)
+noise *= 1 / noise.norm() * noise_free_data.norm() * 0.10
+data = noise_free_data + noise
+
+# Do a FBP-reconstruction
```
adler-j: Comments like these are quite un-specific. It's obvious from the code that we're doing a FBP reconstruiction, so either scrap the comment or add more value, e.g. "Do a FBP-reconsturction to great initial guess", or something similar.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Define components of the variational problem
+reg_param = 0.3
+gradient = odl.Gradient(reco_space)
+
+# l2-squared data matching
+l2_norm = 0.5 * odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+
+# Isotropic TV-regularization
+l1_norm = odl.solvers.GroupL1Norm(gradient.range)
```
kohr-h: Maybe name this `l12_norm` to avoid confusion. Probably other examples have the same issue.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Assembel functionals and operators for the optimization solver
+f = odl.solvers.ZeroFunctional(reco_space)  # Don't use f functional, set zero
+g = [l2_norm, reg_param*l1_norm]
```
kohr-h: Spaces around `*`

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+L = [ray_trafo, gradient]
+
+# Estimated operator norms, add 10 percent to ensure that we fulfill the
+# convergence criteria of the optomization algorithm
+ray_trafo_norm = odl.power_method_opnorm(ray_trafo)
```
kohr-h: You may want to add `maxiter` here for CPU users.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+gradient_norm = odl.power_method_opnorm(gradient)
+
+# Set parameters for the optimization algorithm
+niter = 200  # Number of iterations in the inner loop
+tau = 0.01  # Step size for the primal variable
+sigma_ray_trafo = 45.0 / ray_trafo_norm**2  # Step size for the dual variable
+sigma_gradient = 45.0 / gradient_norm**2  # Step size for the dual variable
```
kohr-h: Spaces around `**`

kohr-h: Also the numbers are a bit "magical". Why not use the default stepsize helper? (It may not be implemented for forward-backward, though...)
Adding a parameter check would be nice for users.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+g = [l2_norm, reg_param*l1_norm]
+L = [ray_trafo, gradient]
+
+# Estimated operator norms, add 10 percent to ensure that we fulfill the
+# convergence criteria of the optomization algorithm
+ray_trafo_norm = odl.power_method_opnorm(ray_trafo)
+gradient_norm = odl.power_method_opnorm(gradient)
+
+# Set parameters for the optimization algorithm
+niter = 200  # Number of iterations in the inner loop
```
kohr-h: A more obvious name would be `niter_inner`

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+sigma = [sigma_ray_trafo, sigma_gradient]
+
+# Create initial guess for the solver.
+x = reco_space.zero()
+
+# Create initial subgradient
+p = reco_space.zero()
+
+# This defines the Bregman iterations
+for breg_iter in range(7):
```
kohr-h: Even though it's only in one place, making a variable `niter_bregman` makes it more clear / less magical.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+    # Used to print iteration number.
+    callback_inner = odl.solvers.CallbackPrintIteration(step=50)
+
+    # Call the solver. x is updated in-place with the consecutive iterates.
+    odl.solvers.forward_backward_pd(
+        x=x, f=f, g=g, L=L, h=linear_part, tau=tau, sigma=sigma, niter=niter,
+        callback=callback_inner)
+
+    # Update the subgradient
+    p -= 1/reg_param * ray_trafo.adjoint(l2_norm.gradient(ray_trafo(x)))
```
kohr-h: Spaces around `/`

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
+"""Example of Bregman-TV in tomography.
+
+The example does seven iterations in a Bregman-TV scheme, applied to a
+tomography problem.
+"""
+
+import odl
+import numpy as np
+
+# Discrete reconstruction space
```
kohr-h: ~Discrete~

examples/solvers/bregman_tv_tomography.py
kohr-h: an

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+data = noise_free_data + noise
+
+# Do a FBP-reconstruction
+fbp_op = odl.tomo.fbp_op(ray_trafo, filter_type='Hann', frequency_scaling=0.4)
+fbp_reco = fbp_op(data)
+
+# Display phantom, FBP reconstruction, and sinograms
+discr_phantom.show(title='Phantom')
+fbp_reco.show(title='FBP reconstruction')
+data.show(title='Simulated data (Sinogram)')
```
kohr-h: Maybe move these to the very end, together with the final Bregman result.

examples/solvers/bregman_tv_tomography.py
adler-j: Put a parenthesis here to make it more clear, e.g. `(1/reg_param)`

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+# convergence criteria of the optomization algorithm
+ray_trafo_norm = odl.power_method_opnorm(ray_trafo)
+gradient_norm = odl.power_method_opnorm(gradient)
+
+# Set parameters for the optimization algorithm
+niter = 200  # Number of iterations in the inner loop
+tau = 0.01  # Step size for the primal variable
+sigma_ray_trafo = 45.0 / ray_trafo_norm**2  # Step size for the dual variable
+sigma_gradient = 45.0 / gradient_norm**2  # Step size for the dual variable
+sigma = [sigma_ray_trafo, sigma_gradient]
```
adler-j: Have you tried using the default choices?

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+# Create the ray transform
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry)
+
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom with noise
+noise_free_data = ray_trafo(discr_phantom)
+noise = odl.phantom.white_noise(ray_trafo.range)
+noise *= 1 / noise.norm() * noise_free_data.norm() * 0.10
```
adler-j: Move 0.10 to the left

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,95 @@
...
+import numpy as np
+
+# Discrete reconstruction space
+reco_space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20],
+                               shape=[128, 128], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+angle_partition = odl.uniform_partition(0, np.pi, 100)
+detector_partition = odl.uniform_partition(-30, 30, 558)
+geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
```
adler-j: Simply go for `parallel_beam_geometry`

examples/solvers/bregman_tv_tomography.py
aringh: Since it is an iterative procedure inside an iterative procedure, I wanted to not have to  do to many inner iterations. Therefore I have tuned the numbers in order to be able to reduce `niter` to "only" 200. So the numbers are indeed a bit magical.

Taking something more standard as standard `tau = 0.5` and `sigma_i = 1 / op_norm**2` seems to demand many more inner iterations.

examples/solvers/bregman_tv_tomography.py
aringh: Yes, see answer to @kohr-h's comment on the same line

examples/solvers/bregman_tv_tomography.py
kohr-h: Ok, fair enough. That would definitely be worth mentioning in a comment.

examples/solvers/bregman_tv_tomography.py
aringh: I will move these to the end. But I will not make a final show of the Bregman recosntruction. Because the example is tuned so that something like 5 iterations is enough, but I do 7. This is on purpose, to illustrate that one needs to do early stopping when doing Bregman iterations. But maybe I should document that somewhere too.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+
+where ``p`` is a subgradient at the point ``x``, i.e., ``p in dF(x)``.
+
+The Bregman iterations can be written as
+
+    x_{k+1} in argmin D(Ax, g) + lam * D_R^{p_k}(x, x_k)
+    p_{k+1} - p_k in -(1/lam) * A^* dD(Ax_k, g).
+
+Note that the first step involves solving an optimization problem, which is
+(normally) done using in iterative method. This means that the solution method
```
kohr-h: ~in~ an

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
+"""Example of Bregman-TV in tomography.
+
+The example solves an inverse problem in tomography using Bregman iterations
+for a total variation (TV) regularized problem. To this end, let ``A`` denote
+the forward operator, ``g`` denote the given data, ``D(Ax, g)`` denote data
```
kohr-h: denote the data

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+
+import odl
+
+# Reconstruction space
+reco_space = odl.uniform_discr(min_pt=[-20, -20], max_pt=[20, 20],
+                               shape=[128, 128], dtype='float32')
+
+# Make a parallel beam geometry with flat detector
+geometry = odl.tomo.parallel_beam_geometry(reco_space)
```
kohr-h: How many angles does this use? Could be quite overkill. In the interest of CPU users, consider setting number of angles to a fixed value, 100-ish. That would also be more interesting as an inverse problem.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+
+# Create the ray transform
+ray_trafo = odl.tomo.RayTransform(reco_space, geometry)
+
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
```
kohr-h: Can you try to make larger blocks of code with comments on top? This feels like lots of comments that state the obvious. For instance, the phantom could easily be part of the "create data" block, and the reco space, geometry and ray trafo could be one block as well.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+
+# Create sinogram of forward projected phantom. Add 10% Guassian noise
```
kohr-h: Dot or no dot? I tend to do it like this for comments, i.e., no trailing dot, but it gets harder with 2 sentences. In those cases I resort to semicolon instead of full stop, i.e., "Do this thing; do a second thing"

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+reg_param = 0.2
+gradient = odl.Gradient(reco_space)
+
+# l2-squared data matching
+l2_norm = 0.5 * odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+
+# Isotropic TV-regularization
+l12_norm = odl.solvers.GroupL1Norm(gradient.range)
+
+# Assembel functionals and operators for the optimization solver
```
kohr-h: Typo "Assembel"
And "optimization solver" is a tautology, either just "solver", or "optimization method", or "optimizer"

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+# Create phantom
+discr_phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create sinogram of forward projected phantom. Add 10% Guassian noise
+noise_free_data = ray_trafo(discr_phantom)
+noise = odl.phantom.white_noise(ray_trafo.range)
+noise *= 0.10 / noise.norm() * noise_free_data.norm()
+data = noise_free_data + noise
+
+# Define components of the variational problem
```
kohr-h: This comment is for the whole section, but it looks like it only applies to `reg_param` and `gradient`. Another good reason to make larger blocks. If you need an intermediate comment in a block, I'd recommend doing it without blank line before, or, if it fits, on the same line:
```
# This is the summary of the block
some code running here
more code
some non-obvious code  # explained very briefly here
# This is a more elaborate comment on the next line
here's another non-obvious statement

# Next block starts here
```

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+gradient = odl.Gradient(reco_space)
+
+# l2-squared data matching
+l2_norm = 0.5 * odl.solvers.L2NormSquared(ray_trafo.range).translated(data)
+
+# Isotropic TV-regularization
+l12_norm = odl.solvers.GroupL1Norm(gradient.range)
+
+# Assembel functionals and operators for the optimization solver
+f = odl.solvers.ZeroFunctional(reco_space)  # Don't use f functional, set zero
```
kohr-h: set to zero

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+ray_trafo_norm = odl.power_method_opnorm(ray_trafo, maxiter=20)
+gradient_norm = odl.power_method_opnorm(gradient, maxiter=20)
+
+# Set parameters for the optimization algorithm. These are tuned in order to
+# reduce the number of inner iterations needed to solve the first step in the
+# Bregman iteration.
+niter_inner = 200  # Number of iterations in the inner loop
+tau = 0.01  # Step size for the primal variable
+sigma_ray_trafo = 45.0 / ray_trafo_norm**2  # Step size for the dual variable
+sigma_gradient = 45.0 / gradient_norm**2  # Step size for the dual variable
```
kohr-h: Spaces around `**` still missing

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+sigma = [sigma_ray_trafo, sigma_gradient]
+
+# Number of Bregman iterations
```
kohr-h: Obvious comment. Better explain why 7.

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+niter_bregman = 7
+
+# Create initial guess for the solver.
```
kohr-h: No dot

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+x = reco_space.zero()
+
+# Create initial subgradient
+p = reco_space.zero()
```
kohr-h: Do both `x` and `p` in one block

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+# This defines the Bregman iterations
+for breg_iter in range(niter_bregman):
+    print('Outer Bregman iteration: {}'.format(breg_iter))
+
+    # Create the affine part of the Bregman functional
+    constant = l12_norm(gradient(x))
+    linear_part = reg_param * odl.solvers.QuadraticForm(vector=-p,
+                                                        constant=constant)
+
+    # Used to print iteration number.
```
kohr-h: Obvious

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+    callback_inner = odl.solvers.CallbackPrintIteration(step=50)
+
+    # Call the solver. x is updated in-place with the consecutive iterates.
```
kohr-h: Again the issue with full stops, would be nice to have consistency

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,134 @@
...
+
+# Estimated operator norms. The norms are used to ensure that we fulfill the
+# convergence criteria of the optomization algorithm
+ray_trafo_norm = odl.power_method_opnorm(ray_trafo, maxiter=20)
+gradient_norm = odl.power_method_opnorm(gradient, maxiter=20)
+
+# Set parameters for the optimization algorithm. These are tuned in order to
+# reduce the number of inner iterations needed to solve the first step in the
+# Bregman iteration.
+niter_inner = 200  # Number of iterations in the inner loop
```
kohr-h: Obvious comment

examples/solvers/bregman_tv_tomography.py
aringh: It seems to use 285 angles 183 detector bins. I updated since @adler-j commented and told me to simplify to this :stuck_out_tongue_closed_eyes: Before it was 
```
angle_partition = odl.uniform_partition(0, np.pi, 100)
detector_partition = odl.uniform_partition(-30, 30, 558)
geometry = odl.tomo.Parallel2dGeometry(angle_partition, detector_partition)
```
This update is also why I in this commit updated the regularization parameter; the change in geometry gave a less interesting result with the previous parameter.

examples/solvers/bregman_tv_tomography.py
aringh: On purpose: I have only one character left on this line before I hit the 79 character limit :wink:. I will try to reformulate

examples/solvers/bregman_tv_tomography.py
kohr-h: Oh but you can give `num_angles` to the function, no need to go back to the old one.

aringh: Ah! I did not know :-)

aringh:
I will merge after the tests have run

kohr-h:
> I will merge after the tests have run

Can you squash the commits? We don't really have an `EX` label :wink:

aringh:
I know we don't, but I created one since I thought that "ENH" was a bit to ambitious for adding an example. But I can change :wink:

kohr-h:
> I know we don't, but I created one since I thought that "ENH" was a bit to ambitious for adding an example. But I can change wink

I think adding an example is a clear ENH. The other one, "update..." makes it a bit hard to find a good label because it's not very specific.
But please don't change it again :grimacing:

aringh:
I wanted to keep the two commits separate explicitly, since the first one is a try to do it using `.bregman`. It never really worked, but in the spirit of keeping some of the development history I though it could be good to keep it if someone wants to try it again. (The probability is very low that someone will find the commit, but anyway.)

examples/solvers/bregman_tv_tomography.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Create the affine part of the Bregman functional
+    constant = l12_norm(gradient(x))
+    linear_part = reg_param * odl.solvers.QuadraticForm(vector=-p,
+                                                        constant=constant)
+
+    callback_inner = odl.solvers.CallbackPrintIteration(step=50)
+
+    # Inner iterations; x is updated in-place with the consecutive iterates
+    odl.solvers.forward_backward_pd(
```
mehrhardt: Without the restrictions of ODL, I would relate `h` with the data fit. The update of the subgradient includes the gradient of the data fit at the last iterate. Thus, one could recycle these computations. It would be good if ODL had the option to do this. In particular for real size problems. This strategy would also be crucial for the linearized Bregman iterations.

kohr-h: That will all be possible when #1400 lands.

aringh: As of now, one can wrap both the data matching and the linear part of the Bregman functional in a `FunctionalQuadraticPerturb`, and pass all of that as the smooth part of the solver (i.e., as `h`). I guess it is something like that you were looking for?

However, before we have something like #1400 in place it will be hard (impossible?) to reuse the gradients computed by the solver.

PR #1308, opened 2018-03-01
ENH: Improve matrix_representation to also handle tensors

adler-j:
Some would call this a "vast simplification", doable because we now have proper indexing for product spaces etc.

This contains some "major" changes.

* For linear domains and ranges (e.g. 1d rn or uniform_discr) nothing changes.
* For domain/range with higher dimensional shapes, the function new computes a tensor of shape `range.shape + domain.shape`
* The above change made some previously usable modes (e.g. `rn(2) * rn(3)`) non-usable (since it does not have a proper shape), and I simply removed that feature.

### DIFF
--- a/odl/operator/oputils.py
+++ b/odl/operator/oputils.py
@@ -14,6 +14,7 @@ import numpy as np
 
 from odl.space.base_tensors import TensorSpace
 from odl.space import ProductSpace
+from odl.util import nd_iterator
 
 __all__ = ('matrix_representation', 'power_method_opnorm', 'as_scipy_operator',
            'as_scipy_functional', 'as_proximal_lang_operator')
@@ -31,6 +32,9 @@ def matrix_representation(op):
     -------
     matrix : `numpy.ndarray`
         The matrix representation of the operator.
+
+        The shape will be ``op.domain.shape + op.range.shape`` and the dtype
+        is the promoted (greatest) dtype of the domain and range.
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


adler-j:
Tests are now running (at least locally). Ready for review.

odl/util/utility.py
```diff
@@ -440,6 +442,26 @@ def is_string(obj):
...
 
 
+def nd_iterator(shape):
+    """Iterator over n-d cube with shape.
+
+    Parameters
+    ----------
+    shape : sequence of int
+        The number of points per axis
+
```
kohr-h:
```
Returns
-------
nd_iterator : generator
    Details go here
```

odl/operator/oputils.py
```diff
@@ -64,66 +69,33 @@ def matrix_representation(op):
```
kohr-h: Commenting here: Currently our example is much like a basic unit test and doesn't really educate users a lot. Maybe we can turn this into a more interesting example, for instance with a `Gradient` operator.
And then, now that the returned object is not necessarily a matrix any more, it would be good to show how to do the equivalent thing to `op(x)` with the matrix using `np.tensordot`.
Here's a suggestion:
```python
>>> space = odl.uniform_discr([0, 0], [2, 2], (2, 2))
>>> grad = odl.Gradient(space, pad_mode='symmetric')
>>> odl.matrix_representation(grad)  # this may be a bit long
array([[[[[-1.,  0.],
          [ 1.,  0.]],

         [[ 0., -1.],
          [ 0.,  1.]]],


        [[[ 0.,  0.],
          [ 0.,  0.]],

         [[ 0.,  0.],
          [ 0.,  0.]]]],



       [[[[-1.,  1.],
          [ 0.,  0.]],

         [[ 0.,  0.],
          [ 0.,  0.]]],


        [[[ 0.,  0.],
          [-1.,  1.]],

         [[ 0.,  0.],
          [ 0.,  0.]]]]])
>>> space = odl.uniform_discr([0, 0, 0], [2, 3, 4], (2, 3, 4))
>>> lapl = odl.Laplacian(space)
>>> lapl_mat = odl.matrix_representation(lapl)
>>> x = space.element(lambda x: x[0] ** 2 + 2 * x[1] ** 2 - x[2] ** 3)  # or sth else for less weird numbers
>>> lapl(x)
uniform_discr([ 0.,  0.,  0.], [ 2.,  3.,  4.], (2, 3, 4)).element(
    [[[   0.875,    2.25 ,   20.75 ,  159.625],
      [  -6.5  ,   -4.375,    1.875,  109.5  ],
      [ -47.125,  -33.75 ,  -15.25 ,  111.625]],
    
     [[  -9.125,   -5.75 ,   12.75 ,  149.625],
      [ -14.5  ,  -10.375,   -4.125,  101.5  ],
      [ -57.125,  -41.75 ,  -23.25 ,  101.625]]]
)
>>> np.tensordot(lapl_mat, x, axes=3)
array([[[   0.875,    2.25 ,   20.75 ,  159.625],
        [  -6.5  ,   -4.375,    1.875,  109.5  ],
        [ -47.125,  -33.75 ,  -15.25 ,  111.625]],

       [[  -9.125,   -5.75 ,   12.75 ,  149.625],
        [ -14.5  ,  -10.375,   -4.125,  101.5  ],
        [ -57.125,  -41.75 ,  -23.25 ,  101.625]]])
```

odl/operator/oputils.py
```diff
@@ -45,13 +50,34 @@ def matrix_representation(op):
...
-           [4, 5, 6],
-           [7, 8, 9]])
+    It also works with `ProductSpace`s and higher dimensional `TensorSpace`s.
+    In this case, the returned "matrix" will also be higher dimensional:
+
+    >>> space = odl.uniform_discr([0, 0], [2, 2], (2, 2))
+    >>> grad = odl.Gradient(space)
+    >>> tensor = odl.matrix_representation(grad)
+    >>> tensor.shape
+    (2L, 2L, 2L, 2L, 2L)
```
kohr-h: I can't imagine this will pass tests

adler-j: darn it.

adler-j: Did we ever decide on how to handle this type of problems?

kohr-h: I don't know. Did Numpy always use long ints for shape entries, irrespective of Python version? I guess so, because it's in the C code. Maybe we never had a doctest that looked at the shape of a Numpy array?
I think the only sensible approach here is to not print stuff that may differ between Py versions (same with strings).

PR #1310, opened 2018-03-11
BLD: add travis test with Numpy 1.10 and drop 1.9 support

kohr-h:
As long as we officially support the old version, we should also test it. I think that currently some part of the code doesn't work with 1.9


adler-j:
Except for the failing tests this should indeed go in!

kohr-h:
It seems like Numpy 1.9 does not allow to explicitly pass `out=None` to a ufunc. I'll fix it.

adler-j:
Frankly, we could also drop 1.9, its extremely old. Or is that the only fail?

kohr-h:
> Frankly, we could also drop 1.9, its extremely old. Or is that the only fail?

I think it would actually work if only Travis would build it..
In principle I don't mind dropping 1.9.

kohr-h:
Looking at the workaround code, it would actually make sense to drop 1.9. For all the pain it causes...

kohr-h:
Okay, I squashed everything, should work now.

PR #1315, opened 2018-03-23
ENH: add random ellipsoid phantom

kohr-h:
This might be useful.

### DIFF
--- a/odl/phantom/geometric.py
+++ b/odl/phantom/geometric.py
@@ -14,8 +14,8 @@ import numpy as np
 from odl.discr.lp_discr import uniform_discr_fromdiscr
 from odl.util.numerics import resize_array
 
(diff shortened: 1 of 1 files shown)


kohr-h:
I admit this is the lazy implementation. I'll make one that checks overlap with the coordinates only.

kohr-h:
I've done [some work](https://github.com/odlgroup/odl/compare/master...kohr-h:random_ellipsoids_better) on this a while ago, but it turned out that it's nontrivial to efficiently determine whether an ellipsoid intersects another one, apart from trivial situations. I've got one pending idea that's not implemented yet, which is to transform the coordinates such that the "query ellipsoid" becomes a ball -- this reduces the problem to determining the distance of an ellipsoid to the origin.

Other suggestions are welcome.

PR #1319, opened 2018-03-31
TST: make global fixtures local to the project, closes #1147

### LINKED ISSUE #1147: ODL pytest plugin interferes with foreign packages

kohr-h:
Simplest possible fix for the issue of potential fixture clashes across projects: Global fixtures that start with an underscore stay local to the project.


adler-j:
how about naming them `odl_SOMETHING`? To me it's more obvious "where they belong" that way

kohr-h:
Maybe my original post wasn't clear: the leading underscore prevents them from appearing at all outside ODL.

kohr-h:
> Aah so _ for fixtures means "project private", did not know that. Go ahead!

It's kind of hard to know since it seems to be undocumented behavior, see also [here](https://github.com/pytest-dev/pytest/issues/3366).
I'll check with a few pytest versions to make sure that it works reliably and then merge.

kohr-h:
Changed to `odl_<something>` naming scheme, and rebased. Merging after CI.

PR #1321, opened 2018-04-05
Add Mayo Reader

adler-j:
This adds a reader for Mayo clinic data.

Some examples:

recon

reference

It seems like the reconstruction is quite good as compared to the reference, so we should be "set".


odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -0,0 +1,48 @@
+"""Reconstruct Mayo dataset using FBP and compare to reference recon.
+
+Note that this example requires that Mayo has been previously downloaded and is
+stored in "mayo_folder".
```
kohr-h: I prefer the more correct term "directory" which can be abbreviated to the even shorter "dir". "folder" is so windows.

kohr-h: Also, this may be misunderstood as literally a directory called "mayo_folder". Thus, mention that it's the variable with that name which counts.

odl/contrib/datasets/README.md
```diff
@@ -21,6 +21,11 @@ Reference datasets with accompanying ODL geometries etc.
     CT data as provided by FIPS. The data is non-human and high resolution.
     * `walnut_data`
     * `lotus_root_data`
+
+
+    CT data as provided by Mayo Clinic. The data is human and high resolution. To access the data, see https://www.aapm.org/GrandChallenge/LowDoseCT/#registration
```
kohr-h: Maybe you can also write a word or two on what you have to do to get the data (no details, just that it requires registration because it's potentially sensitive). I often find it frustrating to just be pointed to a website, only to find out 10 minutes later that I can't get it easily.

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -0,0 +1,48 @@
...
+
+In this example we only use a subset of the data for performance reasons,
+there are ~48 000 projections in the full dataset.
+"""
+
+import odl
+from odl.contrib.datasets.ct.mayo import (projections_from_folder,
+                                          volume_from_folder)
+
+mayo_folder = 'E:/Data/MayoClinic'  # replace with your local folder
```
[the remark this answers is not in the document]
kohr-h: For the string: just a dot?
On the bigger point, I think this is the only sensible option.

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -0,0 +1,48 @@
...
+
+# Load reference reconstruction
+volume_folder = mayo_folder + '/Training Cases/L067/full_1mm_sharp'
+partition, volume = volume_from_folder(volume_folder)
+
+# Load projection data
```
kohr-h: a subset of the

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -0,0 +1,48 @@
...
+
+import odl
+from odl.contrib.datasets.ct.mayo import (projections_from_folder,
+                                          volume_from_folder)
+
+mayo_folder = 'E:/Data/MayoClinic'  # replace with your local folder
+
+# Load reference reconstruction
+volume_folder = mayo_folder + '/Training Cases/L067/full_1mm_sharp'
+partition, volume = volume_from_folder(volume_folder)
```
kohr-h: Weird to have a function called `volume_from_...` return something else *and* the volume.

For functions that return multiple things it's actually okay IMO to use a verb, here for example `load_reference`.

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -0,0 +1,48 @@
...
+# Calculate FBP reconstruction
+fbp_result = fbp(td_window * proj_data)
+
+# Compare the computed recon to reference reconstruction (coronal slice)
+ref = space.element(volume)
+fbp_result.show('Recon (coronal)', clim=[0.7, 1.3])
+ref.show('Reference (coronal)', clim=[0.7, 1.3])
+(ref - fbp_result).show('Diff (coronal)', clim=[-0.1, 0.1])
+
+# Also visualize sagittal slice (note that we only reconstructed a subset)
```
kohr-h: ~~reconstructed~~ used

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
+# Copyright 2014-2017 The ODL contributors
```
kohr-h: I guess new stuff can use 2018

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Tomographic datasets from Mayo Clinic.
```
kohr-h: the

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+
+In addition to the standard ODL requirements, this library also requires:
+
+    - tqdm
+    - dicom
+    - A copy of the Mayo dataset, see
+    https://www.aapm.org/GrandChallenge/LowDoseCT/#registration
```
kohr-h: This is a good list -- can you put it in the readme as well?

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+import odl
+import tqdm
+
+from dicom.datadict import DicomDictionary, NameDict, CleanName
+from odl.contrib.datasets.ct.mayo_dicom_dict import new_dict_items
+
+# Update the dicom dictionary with the extra Mayo tags
+DicomDictionary.update(new_dict_items)
+new_names_dict = dict([(CleanName(tag), tag) for tag in
+                       new_dict_items])
```
kohr-h: I don't think you need the outer list

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # Get the relevant file names
+    file_names = sorted([f for f in os.listdir(folder) if f.endswith(".dcm")])
+
+    if len(file_names) == 0:
+        raise ValueError('No dicom files found in {}'.format(folder))
+
+    file_names = file_names[proj_start:proj_end]
+
+    for file_name in tqdm.tqdm(file_names, 'loading projection data'):
```
kohr-h: Loading

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+        dataset = dicom.read_file(folder + '/' + file_name)
+
+        # Get some required data
+        rows = dataset.NumberofDetectorRows
+        cols = dataset.NumberofDetectorColumns
+        hu_factor = dataset.HUCalibrationFactor
+        rescale_intercept = dataset.RescaleIntercept
+        rescale_slope = dataset.RescaleSlope
+
+        # print(rescale_intercept, rescale_slope)
```
kohr-h: leftover

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # Load the array as bytes
+        data_array = np.array(np.frombuffer(dataset.PixelData, 'H'),
+                              dtype='float32')
+        data_array = data_array.reshape([rows, cols], order='F').T
+
+        # Rescale array
+        # rescaled_array = rescale_intercept + data_array * rescale_slope
```
kohr-h: `hu_factor` missing. But I think the code is pretty obvious, the comment is not really needed

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+    """Load geometry and data stored in Mayo format from folder.
+
+    Parameters
+    ----------
+    folder : str
+        Path to the folder where the Mayo dicom files are stored.
+    proj_start : int
+        Index of the first projection to use. Used for subsampling.
+    proj_end : int
+        Index of the final projection to use.
```
kohr-h: This reads as inclusive, but the implementation is exclusive. Also the value -1 needs explanation.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+
+    Returns
+    -------
+    geometry : ConeFlatGeometry
+        Geometry corresponding to the Mayo projector.
+    proj_data : `numpy.ndarray`
+        Projection data, normalized such that the reconstruction is g/cm^3.
```
kohr-h: will have unit ...

Generally great to mention this!

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+    return geometry, proj_data.asarray()
+
+
+def volume_from_folder(folder, slice_start=0, slice_end=-1):
+    """Load a volume from folder, also returns the corresponding partition.
+
+    Parameters
+    ----------
+    folder : str
+        Path to the folder where the dicom files are stored.
```
kohr-h: DICOM

odl/contrib/datasets/ct/mayo.py
```diff
@@ -0,0 +1,270 @@
...
+        # Get parameters
+        pixel_size = np.array(dataset.PixelSpacing)
+        pixel_thickness = float(dataset.SliceThickness)
+        rows = dataset.Rows
+        cols = dataset.Columns
+
+        # Get data array and convert to correct coordinates
+        data_array = np.array(np.frombuffer(dataset.PixelData, 'H'),
+                              dtype='float32')
+        data_array = data_array.reshape([rows, cols], order='F').T
```
kohr-h: This is a complicated way of writing
```python
data_array.reshape([cols, rows], order='C')
```

odl/contrib/datasets/README.md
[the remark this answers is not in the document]
adler-j: Latest trend in big data

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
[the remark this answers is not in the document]
adler-j: Ok, I'll change to that

odl/contrib/datasets/README.md
adler-j: I agree we should be explicit about that stuff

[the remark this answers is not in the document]
kohr-h: Haha, yes that was my idea as well

odl/contrib/datasets/ct/mayo.py
adler-j: Actually it is a name and I'm quite sure it should be written like this, did some googling and it seems to be proper style

odl/contrib/datasets/ct/mayo.py
adler-j: Both are pip installable, I'd say we just skip them for now since that would clog the readme

odl/contrib/datasets/ct/mayo.py
adler-j: Good catch. Whole thing could be written

```
NameDict.update((CleanName(tag), tag) for tag in new_dict_items)
```

odl/contrib/datasets/ct/mayo.py
adler-j: Comment was a leftover before I added the HU scaling

odl/contrib/datasets/ct/mayo.py
kohr-h: Oh, you're right, of course.

adler-j:
Will merge after CI

adler-j:
Getting "The command "if [[ "$BUILD_DOCS" == "true" ]]; then cd $TRAVIS_BUILD_DIR/doc/source && python generate_doc.py && cd -; travis-sphinx build -n -s $TRAVIS_BUILD_DIR/doc/source 2>/dev/null; fi" exited with 1.", anyone has any idea as to why?

aringh:
I got the same in #1340. I thought it was due to the new documentation, (using  r""" and removing double backslashes if I understood correctly), but I did not have time to investigate it further. However, you do not seem to have that, so maybe it is something else.

aringh:
I had a similar problem in #1267 (I think it was...) which resulted in a lot of commits where I removed the suppressing of the output from the build in Travis. I do not recall exactly what I did; after squashing it all became [this commit](https://github.com/odlgroup/odl/pull/1267/commits/eb9e7d42468bcd757b9fb6c343d544fb9f631a46). However, in the process I got a lot of help from @kohr-h

kohr-h:
See #1344

PR #1324, opened 2018-04-11
ENH: add derivative to some C-to-R operators, closes #1317

### LINKED ISSUE #1317: derivative of RealPart is missing/ not working
When defining an operator that should deal with complex values, one may need the derivative of the RealPart operator. But this does not seem to work. Below is a minimal example and the same issue also holds for ImagPart:

`import odl

space = odl.uniform_discr([-1, -1], [1,1], (500,500), dtype='complex')
    
x = space.one()

realPart_op = odl.operator.default_ops.RealPart(space)
realPart_op.derivative(x)`

kohr-h:


odl/operator/default_ops.py
```diff
@@ -957,6 +959,20 @@ def _call(self, x):
         """Return ``self(x)``."""
         return x.real
 
+    def derivative(self, x):
+        """Return the derivative operator.
+
+        Note that the returned operator (``self``) is the derivative in the
+        Gâteaux sense, but not in the Fréchet sense since it is not linear.
```
sbanert: Are you sure about the Fréchet and Gâteaux sense? As far as I have learned, even the Gâteaux derivative must by definition be linear in the direction. I would formulate it as follows:

Note that the returned operator (``self``) is the derivative only if the spaces are regraded as real vector spaces. The mapping is not differentiable as a mapping between complex vector spaces.

odl/operator/default_ops.py
```diff
@@ -1070,6 +1086,20 @@ def _call(self, x):
         """Return ``self(x)``."""
         return x.imag
 
+    def derivative(self, x):
+        """Return the derivative operator.
+
+        Note that the returned operator (``self``) is the derivative in the
+        Gâteaux sense, but not in the Fréchet sense since it is not linear.
+        Nevertheless, it has an `adjoint`.
```
sbanert: Same here.

odl/operator/default_ops.py
```diff
@@ -1325,13 +1355,129 @@ def __init__(self, space):
...
 
     def _call(self, x):
         """Return ``self(x)``."""
         return (x.real ** 2 + x.imag ** 2).ufuncs.sqrt()
 
+    def derivative(self, x):
+        r"""Return the derivative operator.
+
+        Note that the returned operator (``self``) is the derivative in the
+        Gâteaux sense, but not in the Fréchet sense since it is not linear.
```
sbanert: Same here.

odl/operator/default_ops.py
```diff
@@ -1325,13 +1355,129 @@ def __init__(self, space):
...
+        >>> c2 = odl.cn(2)
+        >>> op = odl.ComplexModulus(c2)
+        >>> op([3 + 4j, 2])
+        rn(2).element([ 5.,  2.])
+        >>> deriv = op.derivative([3 + 4j, 2])
+        >>> deriv.domain
+        cn(2)
+        >>> deriv.range
+        rn(2)
+        >>> deriv([2 + 3j, 4j])  # [5 / 5, 4 / 2]
```
sbanert: Is this really correct?

If I write z_1 = x_1 + i y_1, z_2 = x_2 + i y_2, then
f(z_1, z_2) = f(x_1, y_1, x_2, y_2) = [|z_1| \\ |z_2|], the Jacobian is (when regarded as a mapping ℝ^4 → ℝ²)
df(z_1, z_2) = [x_1/|z_1|, y_1/|z_1|, 0, 0 \\ 0, 0, x_2/|z_2|, y_2/|z_2|]
and for x_1 = 3, y_1 = 4, x_2 = 2, y_2 = 0, when multiplied with [2 \\ 3 \\ 0 \\ 4], I get [18/5 \\ 0]. What is my mistake?

sbanert:
One of the tests seems to fail, too.

odl/operator/default_ops.py
kohr-h: According to [Wikipedia](https://en.wikipedia.org/wiki/G%C3%A2teaux_derivative) linearity is not required, only existence of the limit for all points in question.

odl/operator/default_ops.py
kohr-h: Oh crap, I have forgotten a factor from the inner derivative. :hankey:

odl/operator/default_ops.py
sbanert: Yes, you are right, and the Hille–Phillips book works with the same convention. Still, if we regard ℂ^n as a real vector space (which is legitimate), then this is a Fréchet-differentiable mapping and the derivative is calculated correctly.

(By the way, the ``range`` method gives back a real vector space, and I do not know if the derivative of a mapping from a ℂ-vector space to a ℝ-vector space can really be defined.)

kohr-h: Thanks for checking a more waterproof source. Regarding the sanity of the approach, I think (to quote the [Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python)) the "practicality beats purity" principle applies here. For all practical purposes, code that deals with complex vector spaces should not have to explicitly cast to "pair of real". 
So IMO we should make it as correct as possible without interfering with other code for complex vectors.

kohr-h:
I'm actually considering to replace the complex modulus by the squared complex modulus since it has a much simpler derivative and adjoint. This operator can then be recovered by composing with `odl.ufunc_ops.sqrt(squared_modulus.range)`.
@adler-j you wrote this code IIRC, any big reason to include the square root?

odl/operator/default_ops.py
kohr-h: Yeah, I think it gets pretty hairy. After looking again at the Wikipedia text, I come to the conclusion that "proper" Gateaux derivatives in complex spaces requires the limit to be taken for arbitrary paths in the complex plane, which would be equivalent to Cauchy-Riemann, and thus complex differentiability. So in some sense we take the simple approach and look a bit the other way. Everything should make sense when we interpret all complex spaces as larger real spaces.

I'll mention the notions of differentiation in the docs.

adler-j: We have a whole file in the main doc here:

https://odlgroup.github.io/odl/math/derivatives_guide.html

It might be a good idea to update it with stuff like this, mostly for our own sake as a reference

ozanoktem:
Why is this not merged?

kohr-h:
> Why is this not merged?

A full review is required first, and I'll address @adler-j's comment about adding a note about complex spaces into our documentation on derivatives.

odl/operator/default_ops.py
```diff
@@ -1316,47 +1346,384 @@ def __init__(self, space):
...
+        .. math::
+            &M'(x): X(\mathbb{C}) \to X(\mathbb{R}), \\
+            &M'(x)(y) = \frac{\Re(x)\,\Re(y) + \Im(x)\,\Im(y)}{M(x)}.
+
+        It is linear when identifying :math:`\mathbb{C}` with
+        :math:`\mathbb{R}^2`, but not complex-linear.
+        """
+        op = self
+        x = self.domain.element(x)
+
```
adler-j: Do we want to special case x=0?

odl/operator/default_ops.py
```diff
@@ -1316,47 +1346,384 @@ def __init__(self, space):
...
-        >>> op.inverse(op([1, -2]))
-        rn(2).element([ 1.,  2.])
+        >>> op = odl.ComplexModulusSquared(r2)
+        >>> op([1, -2])
+        rn(2).element([ 1.,  4.])
+
+        The operator also works on other `TensorSpace`'s such as
+        `DiscreteLp`:
+
+        >>> r2 = odl.uniform_discr(0, 1, 2, dtype=complex)
```
adler-j: r2 is a bad name here

odl/operator/default_ops.py
```diff
@@ -1316,47 +1346,384 @@ def __init__(self, space):
...
+
+                .. math::
+                    \langle A x_1,\ A x_2 \rangle_{X(\mathbb{R})} \neq
+                    \langle A^*A x_1,\ x_2 \rangle_{X(\mathbb{C})}.
+                """
+                deriv = self
+
+                class ComplexModulusDerivativeAdjoint(Operator):
+
+                    def __init__(self):
```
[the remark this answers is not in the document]
kohr-h: Yes, I hadn't even considered that before encountering it in the functionals in another branch. Of course it's easier, I'll fix it here.

odl/operator/default_ops.py
```diff
@@ -1316,47 +1346,384 @@ def __init__(self, space):
...
+        rn(2)
+        >>> deriv([2 + 1j, 4j])  # [(3*2 + 4*1) / 5, (2*0 + 0*4) / 2]
+        rn(2).element([ 2.,  0.])
+
+        Notes
+        -----
+        The derivative of the complex modulus
+
+        .. math::
+            &M: X(\mathbb{C}) \to X(\mathbb{R}), \\
```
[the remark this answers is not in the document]
kohr-h: I'll give it a shot

odl/operator/default_ops.py
```diff
@@ -1316,47 +1346,384 @@ def __init__(self, space):
...
+
+                in particular
+
+                .. math::
+                    \langle A x_1,\ A x_2 \rangle_{X(\mathbb{R})} \neq
+                    \langle A^*A x_1,\ x_2 \rangle_{X(\mathbb{C})}.
+                """
+                deriv = self
+
+                class ComplexModulusDerivativeAdjoint(Operator):
```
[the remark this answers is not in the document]
kohr-h: I find it actually easier to read than the alternative way where the code implementing the class is somewhere else in the file or even in a different file. And these operators are really tiny, they fit on one screen.

odl/operator/default_ops.py
```diff
@@ -1316,47 +1346,384 @@ def __init__(self, space):
...
+
+                .. math::
+                    &M'(x): X(\mathbb{C}) \to X(\mathbb{R}), \\
+                    &M'(x)(y) = \frac{\Re(x)\,\Re(y) + \Im(x)\,\Im(y)}{M(x)}.
+
+                Thus, its adjoint can (formally) be identified as
+
+                .. math::
+                    &M'(x)^*: X(\mathbb{R}) \to X(\mathbb{C}), \\
+                    &M'(x)^*(u) = \frac{(\Re(x)\,u,\ \Im(x)\,u}{M(x)}.
```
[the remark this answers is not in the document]
kohr-h: I guess we could, but that gives us one extra copy and multiplication (from the complex embedding). I'm reluctant to sacrificing the speed and lower memory footprint here.

odl/operator/default_ops.py
kohr-h: That would be per component, and stuff like `x[np.isnan(x)] = 0` can add quite some overhead. I can do some benchmarking, but in principle I'm fine with NaN here. That also gives opportunity for debugging by filtering Numpy divide-by-zero errors.

kohr-h:
All remarks addressed I think (up to those that are in debate).

doc/source/math/derivatives_guide.rst
```diff
@@ -158,30 +155,89 @@ This is however not implemented for all functionals.
...
+
+.. math::
+    \langle f,\, \Re^*(g)\rangle_{L^2(\mathbb{C})} = \int \big[ \Re(f)(x)\, g(x) + \mathrm{i}\,\Im(f)(x)\, g(x) \big] \, \mathrm{d}x.
+
+Therefore, ODL takes the following pragmatic approach for complex <-> real operators:
 
-The subgradient is not explicitly implemented in odl, but is implicitly used in the proximal operators.
-See :ref:`proximal_operators` for more information.
\ No newline at end of file
+- Derivatives are taken in the real sense, but linearity is always set to `False` since this refers to linearity in the sense of the involved vector spaces.
```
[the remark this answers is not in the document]
kohr-h: Fixed in ce89155

kohr-h:
Good to go?

kohr-h:
> Looks good now, go ahead!

Still no approval :cry:

kohr-h:
> Still no approval cry

I'll just merge

PR #1326, opened 2018-04-13
Remove ascii and simple example

mehrhardt:
Simple PR to improve my SPDHG contribution. I removed the non-ascii characters as my spyder sometimes complained when importing. Moreover, I included a simple example to get started which does not compare algorithms and does not save anything on disk.


kohr-h:
You say when things are ready.

mehrhardt:
I am happy the way it is so you could merge. Many thanks!

kohr-h:
Let me test something.

@pep8speaks pep8ify

PR #1331, opened 2018-04-26
BUG: Correct OperatorRightVectorMult.derivative, closes #1330.

### LINKED ISSUE #1330: Bug in OperatorRightVectorMult.derivative
The current return statement is

``` 
return self.operator.derivative(x) * self.vector
``` 

but this missed pre-multiplying x by self.vector, it should be

``` 
return self.operator.derivative(x * self.vector) * self.vector
```

sbanert:


odl/operator/operator.py
```diff
@@ -2165,7 +2165,7 @@ def derivative(self, x):
         if self.is_linear:
             return self
         else:
-            return self.operator.derivative(x) * self.vector
+            return self.operator.derivative(self.vector * x) * self.vector
```
kohr-h: Apparently I wasn't fast enough :-)
Please first check if we have to multiply. This is a hot code path and we need to do our best to reduce wasted effort.

sbanert: I improved this by not checking for linearity, but if the derivative coincides with the operator itself. I checked that it works for ``odl.RealPart``, and all the tests pass.

sbanert: After discussion with @olivierverdier: What about setting the ``linear`` flag for the ``RealPart`` to ``True``?

sbanert: I hope, I got everything. Since the ``linear`` flag was used for other stuff in the RealPart and ImagPart operators, I introduced some additional flag for them.

doc/source/math/derivatives_guide.rst
```diff
@@ -238,6 +238,6 @@ but
 
 Therefore, ODL takes the following pragmatic approach for complex <-> real operators:
 
-- Derivatives are taken in the real sense, but linearity is always set to `False` since this refers to linearity in the sense of the involved vector spaces.
-- Even for formally non-linear derivative operators, an adjoint can be defined, which will also not be linear.
+- Derivatives are taken in the real sense. Linearity is set to `True` for an operator :math:`A: X \to Y` if :math:`A'(x) = A` for all :math:`x\in X`. This property can be used to optimize calculations with derivatives, since the derivative operator does not depend on the point. Linearity in the sense of complex vector spaces is currently not reflected by any flag in ODL.
```
kohr-h: Please put the sentences on one line each, RST will still output it as a paragraph. The reason is that it works better with Git's line based content tracking.

sbanert: Done.

sbanert:
I have not found any false statements in RealPart and ImagPart. However, I was missing an operator like ComplexConj, which would also belong to the category mentioned here.

kohr-h:
Yes, `ComplexConj` would be a good candidate for a follow-up PR :wink: 

I'll merge this in the meanwhile, LGTM

Issue #1328, opened 2018-04-25
Is it reasonable to change the implementation of Functional.range?


miallo:
1.  I remember from my courses in analytical geometry, that any norm of any vectorspace always maps to the real numbers. Is there a reason to return the range of a norm of a _complex space_ as ComplexNumbers?
```
>>> X = odl.uniform_discr(0, 1, 10, dtype='complex')
>>> L2NormSquared = odl.solvers.L2NormSquared(X)
>>> print(L2NormSquared.range)
ComplexNumbers
```
(the example holds true for `L1Norm` and `L2Norm` as well)
2.  And I have another problem: I want to create a functional that maps from the complex space to the real numbers (so I can minimise it). But the implementation of `odl.solvers.Functional` says that it maps to `space.field` (which for me would be the complex numbers). What can I do? Should I add an additional operator just to get the real part?
3.  And if so: when trying `odl.RealPart(l2NormSquared.range)` the result is `'ComplexNumbers' object has no attribute 'real_space'`. For now I could live without the functional, but in the future I would want to optimise over a parameter in one of the operators.
4.  And even when I try to work around the functional I have problems:
```
>>> f1 = (L2NormSquared.translated(someElement)) * forwardOp
>>> f = odl.ReductionOperator(f1, 2)
```
throws an error since `f1` is of type `FunctionalComp` and ` 'operators' must be a matrix of 'Operator' objects, '0' or 'None'` (<= by the way: shouldn't this rather read `... must be an array/list of ...`?)
5.  But most importantly: **Thank you for this nice module!**

aringh:
Hi @miallo! You are bumping into an age-old discussion/issue, namely #590.

Yes, a norm should map to the real numbers. However, the definition of a functional that we have used (at least so far) in `odl` is that it is a mapping from a vector space into the field of scalars in that space (see, e.g., https://en.wikipedia.org/wiki/Functional_(mathematics)). However, it is this that has caused all the head ache you now encounter.

Also: great that you like `odl` ^_^

miallo:
Thank you! I see... I must have been stuck in the "historical" point of view... 😁

Can you give me a hint then on how to extract the real part out of the ComplexNumbers()? Since I use the operator I cant just type `L2NormSquared.real` and as I mentioned in 3. I couldn't get `odl.RealPart` working as well...

Now the stupid physicist inside of me asks: Would something break if you set the range of a Functional by default to `space.field` and if `space.field != range` set `linear=False`?

[...]

Issue #1300, opened 2018-02-19
Should we keep `Functional.grad_lipschitz`?


adler-j:
The method `Functional.grad_lipschitz` is currently defined in quite some places, but never used. In addition to this, there seem to be some bugs and other issues. For example,

* `Functional.grad_lipschitz` implements a setter, which violates the imutability assumption of `Operator`s.
* `BregmanDistance` implements `grad_lipschitz`, but only after `_initialize_if_needed` has been called. Before that it returns `nan`.

My suggestion would be that we scrap this property and move the functionality into the `norm` method which is now added to operators in general. This would also fix the problem in `BregmanDistance` above. How do you feel about this?

kohr-h:
I'm all for it.

aringh:
It is ok for me.

PR #1276, opened 2018-01-29
WIP: improve repr by a lot

kohr-h:
- add helper `_separators` to determine string
  separators that keep a `repr` string within a
  given line width
- split repr string rendering into two functions,
  one to determine the parts to be printed and
  one to compile the final string (with separators)
- `attribute_repr_string` helper to class attributes
- `element_repr_string` helper to print space
  elements
- make repr printing of product space elements more
  consistent
- improve default repr of `Operator` (now actually
  good by default, but unspecific)
- use Numpy's print options as a single point to
  configure printing
- add doctests to lots of `__repr__` methods

Further miscellaneous changes:
- fix bug of `uniform_discr_fromdiscr` ignoring dtype
- sort imports using the `isort` tool (mode `-m 4`)
- add `asweighted` to base_tensors to make weighted
  and unweighted spaces compatible in tensor operators
- remove base classes for operator and adjoint in
  favor of inner adjoint classes (less code)
- make most default operators take `domain` and `range`
  and implement the corresponding adjoints/inverses etc.
- implement `noise_array` for product spaces that
  are not power spaces
- remove the `almost_equal` test util
- improve a number of doctests (use `odl.` calls,
  better names etc.)
- make comparison of arrays in tests faster
- implement broadcasting of shapes (M, N) and (N,)
  in product spaces
- remove some remaining doc glitches, mostly `FnBase`

TODO:
- [ ] go through rest of the library
- [ ] replace `almost_equal` by `pytest.approx`
- [ ] tests for pspace broadcasting
- [ ] make `uniform_discr_fromdiscr` respect weighting
- [ ] fix failing unit tests

PR #1333, opened 2018-04-27
WIP real-valued functional

miallo:
I started working on the implementation of the range of Functional as mentioned in #1328.

If I am not mistaken the old implementation of FunctionalRightVectorMult leads to the functional “forgetting” its  grad_lipschitz and linear properties.

I tagged some places in functional.py with “# HELP” where I am especially unsure about what I am doing.

I did NOT implement any kind of typecasting anywhere yet but just tried to make sure that the range property gets passed along.


odl/solvers/functional/functional.py
```diff
@@ -686,7 +693,12 @@ def __init__(self, func, vector):
                             ''.format(func))
 
         OperatorRightVectorMult.__init__(self, operator=func, vector=vector)
-        Functional.__init__(self, space=func.domain)
+        # HELP: before it was Functional.__init__(self, domain=func.domain).
+        # But doesn't this loose the information on grad_lipschitz and linear?
```
aringh: Yes, it seems like the previous implementation forgets the linearity and lipschitz constant for the gradient. However, wouldn't the latter change for this functional? Anyway, we are thinking of removing `grad_lipschitz` since it is not currently used anywhere. See #1300

odl/solvers/functional/functional.py
```diff
@@ -1020,11 +1033,11 @@ def __init__(self, func, quadratic_coeff=0, linear_term=None,
             raise ValueError(
                 "Complex-valued `constant` coefficient is not supported.")
         self.__constant = constant.real
-
+        #HELP: Need to check for range? The + should take care of it, right?
```
aringh: Since `FunctionalQuadraticPerturb` inherits from `Functional` the `super` call is going to execute the code in `Functional.__init__`. So if the test there is sufficient (which I think it is), then this should be fine.

odl/solvers/functional/functional.py
```diff
@@ -1151,10 +1164,10 @@ def __init__(self, left, right):
         if not isinstance(right, Functional):
             raise TypeError('`right` {} is not a `Functional` instance'
                             ''.format(right))
-
+        #HELP: Check ranges?
```
aringh: Should not be needed; it is done in `OperatorPointwiseProduct.__init__`.

odl/solvers/functional/functional.py
miallo: It has been since forever that I worked with subclasses... 😅 (Which is probably a good indicator for my terrible style of programming)

odl/solvers/functional/functional.py
```diff
@@ -1216,9 +1229,11 @@ def __init__(self, dividend, divisor):
 
         self.__dividend = dividend
         self.__divisor = divisor
-
+        
+        #HELP: Check ranges?
```
aringh: Yes, here you would have to check the range.

odl/solvers/functional/functional.py
```diff
@@ -1303,9 +1318,9 @@ def __init__(self, func):
         if not isinstance(func, Functional):
             raise TypeError('`func` {} is not a `Functional` instance'
                             ''.format(func))
-
+        #HELP: The range should be equal?!
```
aringh: Actually, I think that here is where it becomes tricky.... if the range is complex then how should one define the convex conjugate? This is related to #590. So my current best answer is: I am not sure of what the range of this functional should be to make most sense.

odl/solvers/functional/functional.py
aringh: Actually, let me be a bit more specific. What needs to be checked is that the range of the two functionals are the same. Since this only inherits from `Functional`, there is no check elsewhere that will make sure that this is the case. In the same way, there is already a check that the domains are the same.

odl/solvers/functional/functional.py
miallo: Now we could just check for `functional.range == RealNumbers` and otherwise throw an error

odl/solvers/functional/functional.py
```diff
@@ -64,7 +64,12 @@ def __init__(self, space, linear=False, grad_lipschitz=np.nan):
         # Cannot use `super(Functional, self)` here since that breaks
         # subclasses with multiple inheritance (at least those where both
         # parents implement `__init__`, e.g., in `ScalingFunctional`)
-        Operator.__init__(self, domain=space, range=space.field, linear=linear)
+#        print(range)
+        if range is None:
+            range = domain.field
+        elif range != domain.field:
+            linear = False
```
kohr-h: After some very recent discussions (see #1331), we decided for practical reasons to go with linearity in the sense "C = R^2". That would mean here that this `elif` case should go away.

odl/solvers/functional/default_functionals.py
```diff
@@ -453,18 +453,21 @@ class IndicatorLpUnitBall(Functional):
...
             Domain of the functional.
         exponent : int or infinity
             Specifies wich norm to use.
         """
-        super(IndicatorLpUnitBall, self).__init__(space=space, linear=False)
-        self.__norm = LpNorm(space, exponent)
+        super(IndicatorLpUnitBall, self).__init__(domain=domain, linear=False,
+                                                  range=range)
+        # HELP: The LpNorm should always have a range in the real numbers so
+        #       that it is comparable to 1, right?
```
kohr-h: Yeah, so you can override the `Functional` default before the `super` call:
```python
if range is None:
    range = RealNumbers()
```
And then see what happens :laughing:

odl/solvers/functional/functional.py
kohr-h: Yes so we have that issue already now and just decide not to do anything about it. So IMO let's not raise any exceptions yet, instead do the right thing for `RealNumbers` and otherwise do whatever we were doing all along. At some point this whole complex functionals thing needs a thorough analysis and solution (or maybe not?), but this is not the time.

odl/solvers/functional/default_functionals.py
miallo: Isn't this more like what we want to have: `self.__norm = LpNorm(domain, exponent, range=RealNumbers())`
Then the range of the class itself still can be complex if needed

kohr-h: Oh I thought you were in `LpNorm`, didn't realize it was the indicator. There, yes, I agree with your suggestion.

odl/solvers/functional/default_functionals.py
```diff
@@ -59,18 +59,18 @@ class LpNorm(Functional):
...
         ----------
-        space : `DiscreteLp` or `TensorSpace`
+        domain : `DiscreteLp` or `TensorSpace`
             Domain of the functional.
         exponent : float
             Exponent for the norm (``p``).
         """
         super(LpNorm, self).__init__(
-            space=space, linear=False, grad_lipschitz=np.nan)
+            domain=domain, linear=False, grad_lipschitz=np.nan, range=range)
```
kohr-h: So this is step 1: allow to manually set the range. I'm wondering about step 2: the sensible default for norms and such. What about
```python
if range is None and getattr(domain, 'is_complex', False):
    range = RealNumbers()
super(...)
```

odl/solvers/functional/default_functionals.py
```diff
@@ -500,6 +503,7 @@ def convex_conj(self):
         [BC2011] Bauschke, H H, and Combettes, P L. *Convex analysis and
         monotone operator theory in Hilbert spaces*. Springer, 2011.
         """
+        # HELP: I guess it needs `range`, too
```
kohr-h: Yes, you'll want to use `self.range` (it makes sense IMO, and since it's a `@property` you can't pass arguments).

odl/solvers/functional/functional.py
miallo: You are right and if I am not mistaken, it would change in a nontrivial way. My guess for the upper bound would be something like `func.grad_lipschitz * ||constant||_\infty`? Should I implement that or will it be scrapped anyways?

odl/solvers/functional/default_functionals.py
```diff
@@ -1768,16 +1777,20 @@ def __init__(self, operator=None, vector=None, constant=0):
...
             domain = vector.space
+            range = vector.space.field
 
         if (operator is not None and vector is not None and
                 vector not in operator.domain):
             raise ValueError('domain of `operator` and space of `vector` need '
                              'to match')
-
+        # HELP: operator.range == operator.domain to be able to evaluate .inner
+        #       => functional.range = operator.domain.field ?
```
kohr-h: True, this check is missing (or delayed until evaluation)

odl/solvers/functional/default_functionals.py
```diff
@@ -2154,9 +2168,10 @@ def __init__(self, space, outer_exp=1, singular_vector_exp=2):
         >>> norm(space.one())
         inf
         """
+        # HELP: range should be RealNumbers since comparable?
```
kohr-h: It's a norm, so yes

odl/solvers/functional/default_functionals.py
```diff
@@ -2262,8 +2277,10 @@ def __init__(self, functional, sigma=1.0):
         >>> l1_norm = odl.solvers.L1Norm(space)
         >>> smoothed_l1 = MoreauEnvelope(l1_norm)
         """
+        # HELP: Does this only makes sense, if 
+        #       functional.range == RealNumbers()? Otherwise raise Error?
```
kohr-h: No, there's a definition adapted for complex numbers, where the minimum is taken over the real part of the inner product only. Leave this as it is for now.

kohr-h: > Leave this as it is for now.

That is, do add the `range` parameter.

odl/solvers/functional/functional.py
miallo: Wait... This only works for linear operators, correct? So just ignore it and make it `NaN`?

odl/solvers/functional/default_functionals.py
```diff
@@ -223,6 +230,9 @@ def __init__(self, vfspace, exponent=None):
             0 and 1 are currently not supported due to numerical
             instability. Infinity gives the supremum norm.
             Default: ``vfspace.exponent``, usually 2.
+        range : `Field`, optional
+            Range of the functional. The default `None` will be treated as
+            `vfspace.field`.
```
adler-j: We need to decide on a default value here. Either we go for `RealNumbers` for all functionals that can only return real values anyway, or we go for `domain.field`. I'm personally not quite sure which convention is the best, but I guess the msot restrictive would be best, which indicates that we should go for `RealNumbers` whenever the functional is purely real valued.

odl/solvers/functional/default_functionals.py
```diff
@@ -453,25 +466,36 @@ class IndicatorLpUnitBall(Functional):
...
         self.__exponent = float(exponent)
+        self.__range = range
 
     @property
     def exponent(self):
         """Exponent corresponding to the norm."""
         return self.__exponent
 
+    @property
+    def range_(self):
```
adler-j: This is not needed. `IndicatorLpUnitBall` inherits from `Operator` and hence already has a `range` property, which is the one that should be used.

odl/solvers/functional/default_functionals.py
```diff
@@ -599,15 +627,18 @@ class L2Norm(LpNorm):
...
+    def __init__(self, domain, range=None):
         """Initialize a new instance.
 
         Parameters
         ----------
-        space : `DiscreteLp` or `TensorSpace`
+        domain : `DiscreteLp` or `TensorSpace`
             Domain of the functional.
+        range : `Field`, optional
+            Range of the functional. The default `None` will be treated as
```
adler-j: The standard notation is

```
Range of the functional. Default: `RealNumbers`
```

odl/solvers/functional/default_functionals.py
```diff
@@ -1547,7 +1599,7 @@ def proximal(self):
         See Also
         --------
         odl.solvers.nonsmooth.proximal_operators.\
-proximal_convex_conj_kl_cross_entropy :
+        proximal_convex_conj_kl_cross_entropy :
```
adler-j: This will break the link. It has to stay un-indented

odl/solvers/functional/default_functionals.py
```diff
@@ -1655,7 +1707,8 @@ def __init__(self, *functionals):
         domain = ProductSpace(*domains)
         linear = all(func.is_linear for func in functionals)
 
-        super(SeparableSum, self).__init__(space=domain, linear=linear)
+        super(SeparableSum, self).__init__(domain=domain, linear=linear,
+                                           range=functionals[0].range)
```
adler-j: I'm not sure about this choice. One would perhaps expect that real+complex should be complex, but with this implementation it depends on the order.

odl/solvers/functional/default_functionals.py
```diff
@@ -2239,10 +2307,10 @@ class MoreauEnvelope(Functional):
     References
     ----------
     .. _Proximal Algorithms: \
-https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf
+    https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf
     """
 
-    def __init__(self, functional, sigma=1.0):
+    def __init__(self, functional, sigma=1.0, range=RealNumbers()):
```
adler-j: Should be None

odl/solvers/functional/default_functionals.py
```diff
@@ -2364,6 +2441,7 @@ def __init__(self, space, gamma):
         >>> abs(huber_norm(x) - l1_norm(x)) < tol
         True
         """
+
```
adler-j: We typically do not add a newline after the docstring like this

adler-j (changes requested):
I did a quick review. Overall nice changes, but some stylistic stuff etc needs to be adressed.

The main issue with this is that it's a sweeping change but without any tests added. I'd prefer proper testing to be performed.

odl/solvers/functional/default_functionals.py
miallo: I understand your reasoning and personally I would agree with you. I just thought I went with the convention of operators, where the domain (including dtype) has to be the same. But you are certainly right, that this should not depend on the order of the ops. I will think about it

odl/set/sets.py
```diff
@@ -428,10 +428,7 @@ def __hash__(self):
 
     def element(self, inp=None):
         """Return a real number from ``inp`` or from scratch."""
-        if inp is not None:
-            return float(inp)
-        else:
-            return 0.0
+        return float(getattr(inp, 'real', 0.0))
```
miallo: Is this the way to go?

odl/solvers/functional/default_functionals.py
```diff
@@ -1702,13 +1684,21 @@ def __init__(self, *functionals):
...
                 isinstance(functionals[1], Integral)):
             functionals = [functionals[0]] * functionals[1]
+        if any(functional.domain is ComplexNumbers()
+                for functional in functionals):
+            range = ComplexNumbers()
+        elif any(functional.domain is RealNumbers()
+                for functional in functionals):
+            range = RealNumbers()
+        else:
+            range = Integers()
```
miallo: Should we even consider Integer functionals? Is there any application for them?

odl/solvers/functional/default_functionals.py
```diff
@@ -1684,11 +1684,11 @@ def __init__(self, *functionals):
         if (len(functionals) == 2 and
                 isinstance(functionals[1], Integral)):
             functionals = [functionals[0]] * functionals[1]
-        if any(functional.domain is ComplexNumbers()
+        if any(functional.range == ComplexNumbers()
```
miallo: Would it make sense to give all sets an order-parameter? Something like Integers().hirachy_number = 0, RealNumbers().hirachy_number = 1, ComplexNumbers().hirachy_number = 2? Then the _highest priority_ of range would be easy without all those comparisons

odl/solvers/functional/functional.py
```diff
@@ -48,23 +49,33 @@ class Functional(Operator):
     <http://odlgroup.github.io/odl/guide/in_depth/functional_guide.html>`_.
     """
 
-    def __init__(self, space, linear=False, grad_lipschitz=np.nan):
+    def __init__(self, domain, linear=False, grad_lipschitz=np.nan,
```
miallo: Should we change the order to `(self, domain, range, ...)`?
-  This might lead to incompatibilities with existing code like `Functional(space, False)`. Yes, I know that code like this is considered to be bad practise by many, but that probably doesn't stop users from doing so...
+  On the other hand it is the order of operator arguments (and more intuitive if you ask me).
-  For the `default_functionals` it can't be done since most have non-optional other arguments.

odl/set/sets.py
kohr-h: I wouldn't do it this way. The issue is that this maps anything which hasn't got a `.real` attribute to `0.0`. Core types like `int` or `bool` will play nicely with this, but anything else that could be mapped to a float wouldn't. For instance strings or custom classes that implement `__float__`, but not `real`.

So I'd prefer this:
```python
if inp is None:
    return 0.0
else:
    return float(getattr(inp, 'real', inp))
```

odl/solvers/functional/default_functionals.py
```diff
@@ -223,6 +229,8 @@ def __init__(self, vfspace, exponent=None):
             0 and 1 are currently not supported due to numerical
             instability. Infinity gives the supremum norm.
             Default: ``vfspace.exponent``, usually 2.
+        range : `Field`, optional
+            Range of the functional. Default: `vfspace.field`.
```
kohr-h: For this kind of thing you need to use double backticks. 
With single backticks, Sphinx will try to generate a hyperlink, which fails because Sphinx doesn't know the type of `vfspace`. Single backticks are usable for class names and glossary terms mostly, other things that are not referable need to go in double backticks.

odl/solvers/functional/default_functionals.py
```diff
@@ -324,8 +332,8 @@ def convex_conj(self):
 
     def __repr__(self):
         """Return ``repr(self)``."""
-        return '{}({!r}, exponent={})'.format(self.__class__.__name__,
-                                              self.domain,
+        return '{}({!r} -> {!r}, exponent={})'.format(self.__class__.__name__,
+                                              self.domain, self.range,
```
kohr-h: We try, whenever possible, to make `repr` return a string that is a valid Python expression. So please add `range` as a second argument.
Note that there's a quite huge overhaul of all this repr stuff going on, see #1276 for reference. You don't need to spend too much effort on this part.

odl/operator/operator.py
```diff
@@ -1106,8 +1105,14 @@ def __init__(self, left, right, tmp_ran=None, tmp_dom=None):
         rn(3).element([ 2.,  4.,  6.])
         """
         if left.range != right.range:
-            raise OpTypeError('operator ranges {!r} and {!r} do not match'
-                              ''.format(left.range, right.range))
+            if isinstance(left.range, Field) and \
```
kohr-h: Please no backslash line continuation. To split across lines, use parentheses:
```python
if (isinstance(left.range, Field) and
        isinstance(right.range, Field)):
    # do stuff
```

odl/set/sets.py
```diff
@@ -310,6 +310,25 @@ def field(self):
         """
         return self
 
+    def contains_set(self, other):
+        raise NotImplementedError('field {!r} does not have method '
+                                  '`contains_set`'.format(self))
```
kohr-h: This error would already be raised by the parent, wouldn't it?

odl/set/sets.py
```diff
@@ -310,6 +310,25 @@ def field(self):
...
+
+    def __add__(self, other):
+        if self.contains_set(other):
+            return self
+        elif other.contains_set(self):
+            return other
+        else:
+            raise ValueError('Fields {!r} and {!r} do not include '
+                             'each other'.format(self, other))
```
kohr-h: I'm skeptical towards this. It doesn't fulfill the rules for set addition, and just as a shortcut for something used in `OperatorSum`, it shouldn't introduce an API change for `Field`.

odl/solvers/functional/default_functionals.py
```diff
@@ -59,18 +60,23 @@ class LpNorm(Functional):
...
-        space : `DiscreteLp` or `TensorSpace`
+        domain : `DiscreteLp` or `TensorSpace`
             Domain of the functional.
         exponent : float
             Exponent for the norm (``p``).
+        range : `Field`, optional
+            Range of the functional. Default: `RealNumbers`.
         """
+        if range is None:
+            range = RealNumbers()
```
kohr-h: How about 
```python
def __init__(self, domain, exponent, range=RealNumbers()):
    ...
```
? I feel that the current variant is unnecessarily indirect. `None` is usually a proxy for `some_value_that_depends_on_other_stuff_or_must_be_computed`

odl/solvers/functional/default_functionals.py
```diff
@@ -2129,18 +2181,19 @@ class IndicatorNuclearNormUnitBall(Functional):
...
-        space : `ProductSpace` of `ProductSpace` of `TensorSpace`
+        domain : `ProductSpace` of `ProductSpace` of `TensorSpace`
             Domain of the functional.
         outer_exp : {1, 2, inf}, optional
             Exponent for the outer norm.
         singular_vector_exp : {1, 2, inf}, optional
             Exponent for the norm for the singular vectors.
-
+        range : `Field`, optional
+            Range of the functional. Default: `domain.field`.
```
kohr-h: Blank line needed here

odl/solvers/functional/functional.py
kohr-h: > Should we change the order to (self, domain, range, ...)?

IMO definitely. Mostly because `Operator` works the same way, and I agree that it's the natural order of arguments.



> This might lead to incompatibilities with existing code like Functional(space, False). Yes, I know that code like this is considered to be bad practise by many, but that probably doesn't stop users from doing so...

Well I haven't seen any direct subclassing of `Functional` out there in the wild, so I wouldn't be too concerned with breaking shitty code like this.



> For the default_functionals it can't be done since most have non-optional other arguments.

True, but I feel this should be the special case, not the rule.

odl/solvers/functional/functional.py
```diff
@@ -686,7 +698,12 @@ def __init__(self, func, vector):
                             ''.format(func))
 
         OperatorRightVectorMult.__init__(self, operator=func, vector=vector)
-        Functional.__init__(self, space=func.domain)
+        # TODO: Implement grad_lipschitz? If func.is_linear:
+        #           grad_lipschitz = func.grad_lipschitz * ||constant||_\infty
+        #       else: grad_lipschitz = np.nan
```
kohr-h: Nah, I wouldn't bother. The odds are that `grad_lipschitz` will be removed.

odl/solvers/functional/functional.py
```diff
@@ -929,8 +948,10 @@ def __init__(self, left, right):
             raise TypeError('`func` {} is not a `Functional` instance'
                             ''.format(right))
 
+        # HELP: What should happen if left.range != right.range?
```
kohr-h: IMO either nothing, or raise early whatever error adding the convex conjugates would raise. I tend towards not doing anything and letting `convex_conj` fail, at least if the error message is good enough to tell users what exactly went wrong.

Another way to play this would be to strictly require `RealNumbers` as `range` for both, since from the mathematical definition it wouldn't make sense otherwise. However, as in the convex conjugate there's probably an adapted definition for complex-valued functionals, so we probably shouldn't close the door here.

odl/solvers/functional/functional.py
```diff
@@ -1151,10 +1171,11 @@ def __init__(self, left, right):
         if not isinstance(right, Functional):
             raise TypeError('`right` {} is not a `Functional` instance'
                             ''.format(right))
+        range = left.range + right.range
```
kohr-h: See earlier comment, let's not add such cleverness.

odl/solvers/functional/functional.py
```diff
@@ -1211,14 +1232,14 @@ def __init__(self, dividend, divisor):
             raise TypeError('`divisor` {} is not a `Functional` instance'
                             ''.format(divisor))
 
-        if dividend.domain != divisor.domain:
-            raise ValueError('domains of the operators do not match')
+        range = dividend.range + divisor.range
```
kohr-h: Same here

odl/ufunc_ops/ufunc_ops.py
```diff
@@ -326,7 +326,7 @@ def __init__(self, field):
                              "".format(name))
 
         linear = name in LINEAR_UFUNCS
-        Functional.__init__(self, space=field, linear=linear)
+        Functional.__init__(self, domain=field, linear=linear)
```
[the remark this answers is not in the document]
kohr-h: It's just for clarity. 

If you write `Functional.__init__(domain, range, linear)` it's obvious what each of the arguments does, and keywords just add noise: `Functional.__init__(domain=domain, range=range, linear=linear)`.
On the other hand, something like `Functional.__init__(field, True)` doesn't convey any information about the purpose of the arguments, so here `Functional.__init__(domain=field, linear=True)` is much clearer.

But why is this related to tests?

odl/solvers/functional/default_functionals.py
adler-j: This only works if we want `RealNumbers` to be forever immutable. Are we sure about that? (Id be fine with it)

odl/solvers/functional/default_functionals.py
kohr-h: Actually, after looking at the failing doctest (output `L2Norm(rn(3) -> RealNumbers())` doesn't match expected output `L2Norm(rn(3))`, something like this), I'd suggest not changing anything in `__repr__`. That's suboptimal for the non-standard case but at least doesn't break the printout for the default case.

odl/set/sets.py
miallo: I hadn't thought about that... That is a nice solution!

odl/solvers/functional/default_functionals.py
kohr-h: When did we make that decision? I vaguely remember a discussion.

odl/solvers/functional/default_functionals.py
kohr-h: Yeah, we can make that clearer by adding `__slots__` to `RealNumbers` and `ComplexNumbers`.

odl/ufunc_ops/ufunc_ops.py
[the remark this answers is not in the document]
miallo: I stumbled across it in the test results and wanted to make sure that someone else takes over when I am done with implementing most of the stuff

odl/set/sets.py
miallo: I am adding fields, not sets and so I thought I went with the _Minkowski addition_ `A + B = {a + b | a in A, b in B}` (which at least for the implemented numbers is trivial). I actually haven't thought about other sets.

What would be your suggestion to approach this? At first I did something like `if isinstance(a.range, ComplexNumbers) or isinstance(b.range, ComplexNumbers): return ComplexNumbers(); elif ...` but that was rather unpleasant to look at...

miallo: I see... The spaces are homomorph but not isomorph. You were right all along...

kohr-h: :rofl: You went on quite a journey there, my mention of finite fields was more like an example.. In some sense you proved my minimal point that it's not trivial to implement set addition and keep it in line with mathematics for arbitrary fields.

But it doesn't seem to be necessary here anyway.

odl/set/sets.py
```diff
@@ -428,10 +428,10 @@ def __hash__(self):
 
     def element(self, inp=None):
         """Return a real number from ``inp`` or from scratch."""
-        if inp is not None:
-            return float(inp)
-        else:
+        if inp is None:
             return 0.0
+        else:
+            return float(getattr(inp, 'real', inp))
```
kohr-h: Besides the implementation, I'm wondering if we really want this implicit conversion. I'd actually prefer to let this fail for objects that don't implement `__float__` but do implement `real`.
There's also a weird side effect for objects that implement both `__float__` and `real`, namely that `real` is preferred over `__float__`. That doesn't seem quite right to me.
Other opinions, @aringh, @adler-j?

aringh: In this last case, should `ImagPart(RealNumbers().element(1))` give `0` or should it fail? The latter represents seeing `RealNumbers` and `ComplexNumbers` as two different entities, while the former somehow implicitly assumes that the real numbers are seen as embedded in the complex numbers, just with imaginary part 0. Which one is preferred?

kohr-h: I'd take Python (and NumPy) as a guide here:
```python
>>> x = 1.0
>>> x.imag
0.0
>>> x = np.array(1.0)
>>> x.imag
array(0.)
>>> float(1 + 1j)
...
TypeError: can't convert complex to float
>>> np.array(1 + 1j, dtype=float)
...
TypeError: can't convert complex to float
```

adler-j: Some quick trials gives:

```
>>> (1).imag
0
```

Hence, `ImagPart(RealNumbers().element(1))` should return 0.

odl/solvers/functional/default_functionals.py
```diff
@@ -1647,15 +1678,23 @@ def __init__(self, *functionals):
         >>> f_sum = odl.solvers.SeparableSum(l1, 5)
         """
         # Make a power space if the second argument is an integer
-        if (len(functionals) == 2 and
-                isinstance(functionals[1], Integral)):
+        if (len(functionals) == 2 and isinstance(functionals[1], Integral)):
             functionals = [functionals[0]] * functionals[1]
+        for func1 in functionals:
```
miallo: I guess
```Python
from functools import reduce
...
range = reduce((lambda x, func: x if func.range.contains_set(x) else func.range), functionals)
```
would be a more elegant way, but needs the import. Any comments?

miallo: Now that I think about it, I guess this is the way to do it (because it doesn't need the import):
```Python
range = RealNumbers()
for func in functionals:
    range = func.range if func.range.contains(range) else range
```
(Or maybe set `range = None` in the beginning?)

miallo: Or maybe it is reasonable to implement a method for the unification of sets? Something like
```Python
class Set {...
    def unified_with(self, other):
        """This is the default implementation for the unification of two sets
        It only supports equal sets
        """
        if type(self) == type(other):
            return other
        else:
            raise NotImplementedError('Cannot get the unified field for {} and'
                                      '{}'.format(self, other))
```
And the Number implementation
```Python
    def unified_with(self, other):
        """Returns the unified Field of `self` and `other`"""
        return other if other.contains_set(self) else self
```
That way new fields could easily implement their own method.
If we would end up doing so you might want to discuss the name...

kohr-h: I'd suggest the simplest possible solution:
```python
range = RealNumbers()
for func in functionals:
    if func.range.contains(range):
        range = func.range
```

odl/operator/default_ops.py
```diff
@@ -182,7 +182,7 @@ class LinCombOperator(Operator):
         LinCombOperator(a, b)([x, y]) == a * x + b * y
     """
 
-    def __init__(self, space, a, b):
+    def __init__(self, space, a, b, rangeType=None):
```
miallo: Oh, that wasn't meant to be pushed... I used it for a mapping-oparator from R2 to C...

kohr-h:
> Merged from odl main branch

Two things:

1. We don't do merge from master, please rebase instead. That keeps a nice and clean history. (If you need to know how it works, just ask.)
2. We have [some guidelines](https://odlgroup.github.io/odl/dev/gitwash/development_workflow.html#the-commit-message) regarding commit messages that we try to follow ourselves (though sometimes we get sloppy, too :wink:). It makes the list of commits much more digestible if you see at a glance which commit does what.
   Short version:
   - Prefix with `BUG:`, `MAINT:`, `STY:`, `TST:` etc.
   - Use imperative style ("Fix bug", not "Fixed bug")
   - Fit first line in ~50 characters, then go wild after a blank line

I guess this PR is gonna end up as one big squashed commit since it's been going a bit back and forth (no issue with that), so we'll likely fix it then.

adler-j:
So what's the status of this? I'd love for this to get finalized!

miallo:
> So what's the status of this?

As I mentioned somewhere before: I have never written any tests in python and I would rather have someone else do them since I am not 100% confident with what is to be expected for the default values (expecially for things like `IndicatorLpUnitBall.convex_conj`). Otherwise I hope I found most of the 'obvious' bugs for now...

adler-j:
I'm currently very busy leading up to the vacation, but I can do that after summer. Thanks again for the contribution!

Issue #1323, opened 2018-04-11
OperatorTest errors


dksim:
It appears the odl.diagnostics.OperatorTest class has a few bugs

**Examples:**
```
>>> import odl
>>> space = odl.uniform_discr([0,0],[1,1],[5,5])
>>> geometry = odl.tomo.parallel_beam_geometry(space, num_angles = 2, det_shape= 2)
>>> ray_trafo = odl.tomo.operators.RayTransform(space, geometry)
```
**1. Failure to estimate norm (raises ValueError)**
```>>> optest = odl.diagnostics.OperatorTest(ray_trafo)```
**2. Linear operator fails derivative and adjoint tests**
```
>>> ray_trafo.is_linear
True
>>> optest._derivative_convergence
```
> Verifying that derivative is a first-order approximation
> x=One             p=Gaussian       , error=0.0006087356400996563
> x=One             p=Grad 0         , error=0.00024304196855373788
> x=One             p=Grad 1         , error=0.00024307337744078404
> x=One             p=Grad all       , error=0.00048608393545274375
> x=Cube            p=One            , error=0.00026358338048604085
> x=Cube            p=Grad 0         , error=0.00024304196524427385
> x=Cube            p=Grad all       , error=0.00017199506603023416
> x=Gaussian        p=One            , error=0.00026358338048604085
> x=Gaussian        p=Gaussian       , error=0.0003335309723047023
> x=Gaussian        p=Grad 0         , error=0.00024304196524427385
> x=Gaussian        p=Grad 1         , error=0.00024307337413134527
> x=Gaussian        p=Grad all       , error=0.0007702715463741245
> x=Grad 0          p=One            , error=0.0002635833800723579
> x=Grad 0          p=Grad 0         , error=0.00038513577318706227
> x=Grad 0          p=Grad 1         , error=0.00038510436430001614
> x=Grad 0          p=Grad all       , error=0.0001420938050470074
> x=Grad 1          p=One            , error=0.0002635833800723579
> x=Grad 1          p=Gaussian       , error=0.00029464676943609776
> x=Grad 1          p=Grad all       , error=0.00017199506561655118
> x=Grad all        p=Grad 0         , error=0.00024304196524427385
> x=Grad all        p=Grad 1         , error=0.00024307337413134527
> x=Grad all        p=Grad all       , error=0.0001420938050470074
> error = inf_c ||A(x+c*p)-A(x)-A'(x)(c*p)|| / c
[...]

Issue #717, opened 2016-11-15
What do we mean by derivative?


adler-j:
As noticed in PR #668, we don't have a definition of derivative that we use consequently throughout the code. In many (most?) places we mean the Frechet derivative (linear approximation), but in some places we mean the more general Gateux derivative (directional derivative).

How do we want to deal with this? We basically have two options:

* Demand that all operators obtained by `op.derivative`  are linear, which implies that we only use the Frechet derivative.
* Allow `op.derivative` to be non-linear, i.e. allow the Gateux derivative.

ozanoktem:
I strongly advise we use the notion of Fréchet derivative and ensure the result is a linear operator. Since Fréchet differentiability implies Gâteaux differentiability, and the Fréchet and Gâteaux derivatives agree, the difference is when an operator is not Fréchet differentiable.

adler-j:
Well what If the user wants to differentiate something like the absolute value function? Should we strictly disallow that, or should we allow the Gateux derivative in this case?

ozanoktem:
> Well what If the user wants to differentiate something like the absolute value function? Should we strictly disallow that, or should we allow the Gateux derivative in this case?
Is there a method, like an optimization scheme, that is based on the Gâteaux derivative and not on the Fréchet derivative? What I mean is, what is the use case where you need a Gâteaux derivative of a non-Fréchet differentiable operator?

chongchenmath:
Old topic. In general, the Gâteaux derivative is calculable, but Fréchet is not. We need to calculate Gâteaux to get Fréchet if it is Fréchet differentiable. From numerical side, the Gâteaux is more straightforward and is always what we do.

adler-j:
> Is there a method, like an optimization scheme, that is based on the Gâteaux derivative and not on the Fréchet derivative? What I mean is, what is the use case where you need a Gâteaux derivative of a non-Fréchet differentiable operator?

Well easiest example is simply solving something like

`min_x  ||Ax-b||_2^2 + ||x||_1`

with steepest descent and line search. Sure you expect it to be bad, but it would be nice if it at least runs.

ozanoktem:
[...]

Issue #526, opened 2016-08-25
Operator diagnostics too strict for single precision


kohr-h:
The desired accuracies of calculations in the operator diagnostics are hard-coded for double precision, and the test suite gives errors in single precision for operators that are actually correct.

There should be a way to adapt the accuracy to the used precision.

adler-j:
There should be a parameter for the user to use as well where this can be set.

Issue #340, opened 2016-04-05
Scrap class structure of diagnostics


adler-j:
I think the diagnostics could very well be changed to use a functional approach, i.e. instead of

```python
odl.diagnostics.OperatorTest(my_operator).run_tests()
```

we could have

```python
odl.diagnostics.run_operator_tests(my_operator)
```

kohr-h:
Why not?

PR #1334, opened 2018-04-27
WIP: Issue 1323  optest

kohr-h:
This is an attempt at improving our `diagnostics` module. It's work in progress, and I'm curious what others think about the output and the workflow.

Here's what I think is better:
- The `OpTest` class is gone, instead everything is implemented with plain functions.
- No more dependency on execution order. Before, several methods depended on `norm()` being executed before, and it would also fail for nonlinear operators.
- The verbosity levels are more fine-grained. One can choose all the way from `'DEBUG'` which prints A LOT, to `'QUIET'`, which only prints the number of failures.

Here's an example:
```python
>>> S = odl.ScalingOperator(odl.rn(3), 2.0)
>>> odl.diagnostics.check_operator(S, verbosity='QUIET')
properties: 0 failed
norm: 0 failed
linearity: 0 failed
adjoint: 0 failed
>>> odl.diagnostics.check_operator(S, verbosity='INFO')
<output see gist>
```
Output is [here](https://gist.github.com/kohr-h/08524c118b986ec1ec6a2d696807ad68)

**TODO:**
- The returned dictionary is mostly only `{'num_failed': <number of failed tests>}`. Maybe some more details would be nice, like number of tests run, elapsed time, or even a list of checks, each with timing and result. 
  Input on this is appreciated!
- Mention the line width configurabiltiy in the docs
- Do the space tests as well
- Adapt `examples` in all spaces where defined
- Add option for log file

Ping @adler-j @aringh @sbanert 

Related issues: #1323, #526, #717, #340

### DIFF
--- a/odl/diagnostics/operator.py
+++ b/odl/diagnostics/operator.py
@@ -1,4 +1,6 @@
-﻿# Copyright 2014-2017 The ODL contributors
+﻿# coding: utf-8
+
+# Copyright 2014-2017 The ODL contributors
 #
 # This file is part of ODL.
 #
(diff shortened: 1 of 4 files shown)


kohr-h:
Bump. Would be nice to get some feedback.

odl/diagnostics/operator.py
```diff
@@ -9,380 +11,1012 @@
...
+    log('-' * log_linewidth, level='INFO')
+
+    for (name_x, x), (name_y, y) in samples(op.domain, op.domain):
+        op_x = op(x)
+        op_y = op(y)
+        op_x_y = op(x + y)
+
+        denom = opnorm * (x.norm() + y.norm())
+        if denom == 0:
+            denom = 1.0
```
adler-j: warning?

odl/diagnostics/operator.py
```diff
@@ -9,380 +11,1012 @@
...
-
-    def adjoint(self):
-        """Verify that `Operator.adjoint` works appropriately.
-
-        References
-        ----------
-        Wikipedia article on `Adjoint
-        <https://en.wikipedia.org/wiki/Adjoint>`_.
-        """
+            opnorm = op.norm(estimate=False)
```
adler-j: There is a utility that does this (in this file)

PR #1338, opened 2018-05-03
DOC: Correct dead link to commit message acronyms, closes #1337.

### LINKED ISSUE #1337: Dead link to commit message acronyms
By accident, I found that the link under item 1 in https://odlgroup.github.io/odl/dev/gitwash/development_workflow.html#the-commit-message is dead.

sbanert:


kohr-h:
Definitely :+1: But it seems the base URL for the NumPy doc changed, so we probably have a bunch more dead links, and also the [intersphinx link](https://github.com/odlgroup/odl/blob/master/doc/source/conf.py#L62) could be dead. Could you check that?

PR #1339, opened 2018-05-03
ENH: Add is_real and is_complex to pspace and fspace

adler-j:


odl/test/space/pspace_test.py
```diff
@@ -79,6 +79,8 @@ def test_emptyproduct():
     spc = odl.ProductSpace(field=reals)
     assert spc.field == reals
     assert spc.size == 0
+    assert spc.is_real
```
adler-j: Could use some guidance here. How should we handle the "empty" space? Go by dtype if exists? Always true? Always false?

adler-j:
See #1348 instead.

Issue #793, opened 2017-01-07
Remove "not a Hilbert space" restriction?


kohr-h:
We currently disable inner products and adjoints when the space exponent is not equal to 2. However, both can be interpreted in a generalized sense in Banach spaces.

- The inner product can be generalized to the *dual product* `<.,.>: L^p x L^(p') -> F`, similar with other dual product spaces `X x X^*`.
- Instead of the adjoint, one can use the *dual* of an operator `A: L^p -> L^q` as the operator `A^*: L^(q') -> L^(p')`, in general `A: X -> Y`, `A^*:Y^* -> X^*`.

In fact, when dealing with Lebesgue spaces, the formal definitions of these quantities are most of the time identical with the Hilbert space variants. And it can be very useful e.g. in optimization when these concepts can be applied "without thinking" to more general spaces.

There are basically 3 of ways to deal with this:
1. Be strict with the terms and disallow "abuse" for more general cases (current approach)
2. Be less strict and allow combinations of Lebesgue spaces with mutually conjugate exponents in inner products, and allow adjoints on the conjugate exponent variants of domain and range.
3. Allow anything.

Here's how I think about this:
(1) feels like an arbitrary and very limiting restriction IMO, without a compelling reason.
(2) is still mathematically sane but requires a bit more code - in particular something like a `dual_space` property would be needed for that. But we also have `tangent_bundle` so I don't think that would be a big deal to add.
(3) is the lazy approach, and that might bite us in the long run if things really stop making sense.

Opinions?

adler-j:
Well 2 is obviously the way to go, but it adds quite some complexity if users want to add a new norm etc, then they also have to somehow be able to provide the dual norm. With that said, adding it would be trivial, and adding "dual space" would i.m.o. make perfect sense.

A minor addition is that we should make all "obvious" identifications, i.e. the dual space of a hilbert space is itself, and the dual of the dual is (in most cases anyway) the space itself. Perhaps we also need a "is_reflexive" flag.

kohr-h:
> A minor addition is that we should make all "obvious" identifications, i.e. the dual space of a hilbert space is itself, and the dual of the dual is (in most cases anyway) the space itself. Perhaps we also need a "is_reflexive" flag.

We should make obvious identifications, right. Regarding reflexivity, I'm not sure what to do ideally. Can we ever represent the part of the bidual that is not contained in the original space? Like, functionals of functionals? That would really require a whole new machinery without really adding much.

[...]

PR #1340, opened 2018-05-04
Update the Bregman functional implementation

### LINKED ISSUE #1312: Implementation of Bregman distances
As the discussion in #1267 goes, we may need to revise the way Bregman distances work in ODL. The discussion should continue here, please fill in the details.

aringh:
Closes #1312

### DIFF
--- a/odl/solvers/functional/functional.py
+++ b/odl/solvers/functional/functional.py
@@ -225,18 +225,16 @@ class Functional(Operator):
         """
         return FunctionalTranslation(self, shift)
 
-    def bregman(self, point, subgrad=None):
+    def bregman(self, point, subgrad):
         """Return the Bregman distance functional.
 
         Parameters
         ----------
         point : element of ``functional.domain``
             Point from which to define the Bregman distance.
-        subgrad : element of ``functional.domain`` or `Operator`, optional
-            A subgradient of ``functional`` in ``point``, or an operator
-            which that takes an element in ``functional.domain`` and returns a
-            subdifferential of the functional in that point.
-            Default: ``functional.gradient``.
+        subgrad : element of ``functional.domain``
+            A subgradient of ``functional`` in ``point``. If it exists,  a
+            valid option is ``functional.gradient(point)``.
 
         Returns
         -------
@@ -245,25 +243,15 @@ class Functional(Operator):
 
         Notes
         -----
-        Given a functional :math:`f`, a point :math:`y`, and a (sub)gradient
-        :math:`p \in \partial f(y)`, the Bregman distance functional
-        :math:`D_f^p(\cdot, y)` in a point :math:`x` is given by
-
-        .. math::
-            D_f^p(x, y) = f(x) - f(y) - \\langle p, x - y \\rangle.
-
-
-        For mathematical details, see `[Bur2016]`_.
+        For mathematical details, see
+        `[Bur2016] <https://arxiv.org/abs/1505.05191>`_. See also the Wikipedia
+        article: https://en.wikipedia.org/wiki/Bregman_divergence
 
         References
         ----------
-        Wikipedia article: https://en.wikipedia.org/wiki/Bregman_divergence
-
         [Bur2016] Burger, M. *Bregman Distances in Inverse Problems and Partial
         Differential Equation*. In: Advances in Mathematical Modeling,
         Optimization and Optimal Control, 2016. p. 3-33.
-
-        .. _[Bur2016]:  https://arxiv.org/abs/1505.05191
         """
         return BregmanDistance(self, point, subgrad)
 
@@ -1334,7 +1322,7 @@ class FunctionalDefaultConvexConjugate(Functional):
 
 
 class BregmanDistance(Functional):
-    """The Bregman distance functional.
+    r"""The Bregman distance functional.
 
     The Bregman distance, also refered to as the Bregman divergence, is similar
     to a metric but satisfies neither the triangle inequality nor symmetry.
@@ -1348,20 +1336,19 @@ class BregmanDistance(Functional):
     :math:`D_f^p(\cdot, y)` in a point :math:`x` is given by
 
     .. math::
-        D_f^p(x, y) = f(x) - f(y) - \\langle p, x - y \\rangle.
+        D_f^p(x, y) = f(x) - f(y) - \langle p, x - y \rangle.
 
 
-    For mathematical details, see `[Bur2016]`_.
+    For mathematical details, see
+    `[Bur2016] <https://arxiv.org/abs/1505.05191>`_. See also the Wikipedia
+    article: https://en.wikipedia.org/wiki/Bregman_divergence
+
 
     References
     ----------
-    Wikipedia article: https://en.wikipedia.org/wiki/Bregman_divergence
-
     [Bur2016] Burger, M. *Bregman Distances in Inverse Problems and Partial
     Differential Equation*. In: Advances in Mathematical Modeling, Optimization
     and Optimal Control, 2016. p. 3-33.
-
-    .. _[Bur2016]:  https://arxiv.org/abs/1505.05191
     """
 
     def __init__(self, functional, point, subgrad=None):
@@ -1370,16 +1357,12 @@ class BregmanDistance(Functional):
         Parameters
         ----------
         functional : `Functional`
-            Functional on which to base the Bregman distance. If the
-            optional argument `subgrad` is not given, the functional
-            needs to implement `functional.gradient`.
+            Functional on which to base the Bregman distance.
         point : element of ``functional.domain``
             Point from which to define the Bregman distance.
-        subgrad : element of ``functional.domain`` or `Operator`, optional
-            A subgradient of ``functional`` in ``point``, or an operator
-            which that takes an element in ``functional.domain`` and returns a
-            subdifferential of the functional in that point.
-            Default: ``functional.gradient``.
+        subgrad : element of ``functional.domain``
+            A subgradient of ``functional`` in ``point``. If it exists,  a
+            valid option is ``functional.gradient(point)``.
 
         Examples
         --------
@@ -1388,7 +1371,9 @@ class BregmanDistance(Functional):
         >>> space = odl.uniform_discr(0, 1, 10)
         >>> l2_squared = odl.solvers.L2NormSquared(space)
         >>> point = space.one()
-        >>> Bregman_dist = odl.solvers.BregmanDistance(l2_squared, point)
+        >>> subgrad = l2_squared.gradient(point)
+        >>> Bregman_dist = odl.solvers.BregmanDistance(
+        ...     l2_squared, point, subgrad)
 
         This is gives squared L2 distance to the given point, ||x - 1||^2:
 
@@ -1406,42 +1391,11 @@ class BregmanDistance(Functional):
                              ''.format(point, functional.domain))
         self.__point = point
 
-        if subgrad is None:
-            # If the subgradient is not given, assign the gradient of the
-            # functional
-            try:
-                self.__subgrad = self.__functional.gradient
-                self.__subgrad_is_operator = True
-            except NotImplementedError:
-                raise NotImplementedError(
-                    '`subgradient` not given, and the `functional` {} does '
-                    'not implement `functional.gradient`'.format(functional))
-        else:
-            # Check that given subgradient is either an element in the domain
-            # or an operator that maps from the domain to itself
-            if isinstance(subgrad, Operator):
-                if subgrad.domain != self.__functional.domain:
-                    raise ValueError('`subgrad.domain` {} is not the same as '
-                                     '`functional.domain` {}'
-                                     ''.format(self.__functional.domain,
-                                               subgrad.domain))
-                if subgrad.range != self.__functional.domain:
-                    raise ValueError('`subgrad.range` {} is not the same as '
-                                     '`functional.domain` {}'
-                                     ''.format(self.__functional.domain,
-                                               subgrad.range))
-                self.__subgrad_is_operator = True
-
-            elif subgrad in functional.domain:
-                self.__subgrad_is_operator = False
-
-            else:
-                raise TypeError(
-                    '`subgrad` {} must either be an instance of ``Operator``, '
-                    'or an element in `functional.domain` {}'
-                    ''.format(subgrad, functional.domain))
-
-            self.__subgrad = subgrad
+        if subgrad not in functional.domain:
+            raise TypeError(
+                '`subgrad` must be an element in `functional.domain`, got '
+                '{}'.format(subgrad))
+        self.__subgrad = subgrad
 
         self.__is_initialized = False
 
@@ -1460,10 +1414,7 @@ class BregmanDistance(Functional):
 
     @property
     def subgrad(self):
-        """The subgradient used to define the Bregman distance.
-
-        It is either an operator or an element in the domain of the functioal.
-        """
+        """The subgradient used to define the Bregman distance."""
         return self.__subgrad
 
     @property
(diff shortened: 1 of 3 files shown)


odl/solvers/functional/functional.py
```diff
@@ -232,10 +232,10 @@ def bregman(self, point, subgrad=None):
...
         point : element of ``functional.domain``
             Point from which to define the Bregman distance.
-        subgrad : `Operator`, optional
-            A subgradient operator. Operator which that takes an element in
-            `functional.domain` and returns a subdifferential of the functional
-            in that point.
+        subgrad : element of ``functional.domain`` or `Operator`, optional
+            A subgradient of ``functional`` in ``point``, or an operator
+            which that takes an element in ``functional.domain`` and returns a
+            subdifferential of the functional in that point.
```
kohr-h: I think we're still not on the same page. What I had in mind was this:
```
subgrad : element of ``functional.domain``, optional
    A subgradient of ``functional`` in the given ``point``.
    The default for ``None`` is ``functional.gradient(point)``,
    which implies that this parameter is mandatory if
    ``functional.gradient`` is not implemented.
```
So no gradient operator anywhere, except for the default subgradient.

Just a nitpicking remark: subgradient = element, subdifferential = set. Some confusion about that here.

odl/solvers/functional/functional.py
```diff
@@ -245,12 +245,12 @@ def bregman(self, point, subgrad=None):
...
+        Given a functional :math:`f`, a point :math:`y`, and a (sub)gradient
+        :math:`p \in \partial f(y)`, the Bregman distance functional
+        :math:`D_f^p(\cdot, y)` in a point :math:`x` is given by
 
         .. math::
-            D_f(x, y) = f(x) - f(y) - \\langle \partial f(y), x - y \\rangle.
+            D_f^p(x, y) = f(x) - f(y) - \\langle p, x - y \\rangle.
 
 
         For mathematical details, see `[Bur2016]`_.
```
kohr-h: Can you make that an inline link,
```
 `[Bur2016] <https://arxiv.org/abs/1505.05191>`_
```
?

A bit below in `References`, the "Wikipedia article" is just lying around there, I think it would be better to have that in the text above.

odl/solvers/functional/functional.py
```diff
@@ -1343,12 +1343,12 @@ class BregmanDistance(Functional):
...
-    Given a functional :math:`f`, which has a (sub)gradient :math:`\partial f`,
-    and given a point :math:`y`, the Bregman distance functional
-    :math:`D_f(\cdot, y)` in a point :math:`x` is given by
+    Given a functional :math:`f`, a point :math:`y`, and a (sub)gradient
+    :math:`p \in \partial f(y)`, the Bregman distance functional
+    :math:`D_f^p(\cdot, y)` in a point :math:`x` is given by
 
     .. math::
-        D_f(x, y) = f(x) - f(y) - \\langle \partial f(y), x - y \\rangle.
+        D_f^p(x, y) = f(x) - f(y) - \\langle p, x - y \\rangle.
```
kohr-h: Convert to no-double-backslash version

odl/solvers/functional/functional.py
```diff
@@ -1364,21 +1364,21 @@ class BregmanDistance(Functional):
...
-    def __init__(self, functional, point, func_subgrad=None):
+    def __init__(self, functional, point, subgrad=None):
         """Initialize a new instance.
 
         Parameters
         ----------
         functional : `Functional`
             Functional on which to base the Bregman distance. If the
-            optional argument `func_subgrad` is not given, the functional
+            optional argument `subgrad` is not given, the functional
```
kohr-h: Double backticks needed

odl/solvers/functional/functional.py
```diff
@@ -1364,21 +1364,21 @@ class BregmanDistance(Functional):
...
         point : element of ``functional.domain``
             Point from which to define the Bregman distance.
-        func_subgrad : `Operator`, optional
-            A subgradient operator. Operator which that takes an element in
-            `functional.domain` and returns a subdifferential of the functional
-            in that point.
+        subgrad : element of ``functional.domain`` or `Operator`, optional
+            A subgradient of ``functional`` in ``point``, or an operator
+            which that takes an element in ``functional.domain`` and returns a
+            subdifferential of the functional in that point.
```
kohr-h: See earlier comment

odl/solvers/functional/functional.py
kohr-h: I'd say that's okay since most of the time it will be called through the `bregman` method, and there it's not outrageous if the creation takes some time.

kohr-h: An alternative that I would also find okay would be to make the parameter mandatory and just write in the docs that `functional.gradient(point)` is a valid choice if implemented.

mehrhardt: I am also in favour of no intelligence and make the `subgrad` mandatory.

aringh:
Updated accordingly. Now `subgrad` is a mandatory input argument, and it only allows for elements that are in `functional.domain`.

aringh:
>> Updated accordingly. Now subgrad is a mandatory input argument, and it only allows for elements that are in functional.domain.

> This is a good point. Not sure how (or if) ODL wants to handle it, but in Banach spaces the subgradient would be an element of the dual of functional.domain.

This is of course correct. However, so far we (as in @adler-j and @kohr-h which are the real developers ;-)) have not found a nice way to represent dual spaces, see #793. This also lead to the [this disclaimer](https://github.com/odlgroup/odl/blob/master/odl/solvers/functional/functional.py#L41-L48) in the base class `functional`.

I say we update this if we solve the other issue. If so, then we should also update `convex_conj`, since this functional would map from the dual space into the field of scalars (real numbers)

mehrhardt:
Sure, I don't think it is very urgent.

aringh:
Just for later reference: I tried to investigate the build fail here, by just removing `2>/dev/null` from `.travis.yml` in order to get more output. However, the build did not complete in over 3 hours, so I decided to just dropped that commit and reverted back.

odl/solvers/functional/functional.py
```diff
@@ -1462,15 +1432,15 @@ def initialized(self):
     def _initialize_if_needed(self):
         """Helper function to initialize the functional."""
         if not self.__is_initialized:
-            self.__func_subgrad_eval = self.func_subgrad(self.point)
+
             self.__constant = (-self.functional(self.point) +
-                               self.__func_subgrad_eval.inner(self.point))
+                               self.__subgrad.inner(self.point))
```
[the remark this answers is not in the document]
kohr-h: Absolutely correct. The `__` attributes should be used during `__init__` at least until the class is finalized (`super()` has been called). After that the "official" attributes are preferred since that allows subclasses to customize their behavior.

odl/solvers/functional/functional.py
```diff
@@ -1388,7 +1371,9 @@ def __init__(self, functional, point, subgrad=None):
         >>> space = odl.uniform_discr(0, 1, 10)
         >>> l2_squared = odl.solvers.L2NormSquared(space)
         >>> point = space.one()
-        >>> Bregman_dist = odl.solvers.BregmanDistance(l2_squared, point)
+        >>> subgrad = l2_squared.gradient(point)
+        >>> Bregman_dist = odl.solvers.BregmanDistance(
```
kohr-h: Even though it's a name, I'd say the rule to write variables in lower case overrules.

kohr-h:
I think now there are only a few minor things to fix, otherwise it looks very good.

kohr-h:
After rebasing, the doc build error should be gone.

kohr-h:
Oops

odl/solvers/functional/functional.py
```diff
@@ -1450,9 +1420,9 @@ def point(self):
...
     @property
-    def func_subgrad(self):
-        """The subgradient operator used to define the Bregman distance."""
-        return self.__func_subgrad
+    def subgrad(self):
+        """The subgradient used to define the Bregman distance."""
+        return self.__subgrad
 
     @property
     def initialized(self):
```
kohr-h: Can you please also remove this property? I don't think it should be public.

odl/solvers/functional/functional.py
```diff
@@ -1462,15 +1432,15 @@ def initialized(self):
     def _initialize_if_needed(self):
```
kohr-h: I suggest that we change this pattern from an additional initializer to a method that computes the value if needed and returns it. Like this:

```python
def __init__(self, ...):
    ...
    # Set to `None` to signal "not initialized"
    self.__constant = None

def _constant(self):
    if self.__constant is None:
        self.__constant = ...  # compute it
    return self.__constant

# Rest of the methods use `_constant()`
```

This makes it clearer that really the initialization is over after `__init__` (which it should for an immutable object like `Operator` or `Functional`), and that methods that alter state only do so in a caching manner, not to do stuff that's actually relevant for the functioning of the class.

odl/solvers/functional/functional.py
aringh: Just to log my own thinking: I left it because I though that one might be interested in checking which subgradient was used to define the distance.

Ok, will remove.

odl/solvers/functional/functional.py
aringh: Sure, such a pattern makes more sense. Before the method also evaluated the subgradient in the given point, which was moved out of `__init__` in order to not make initialization heavy.

aringh: I was just about to implement this when I realized that this is probably not the way we want to do it.

The entire functional is basically an encapsulation and factory for `FunctionalQuadraticPerturb`, and to initialize this functional we of course need the constant. So either this is initialized in `__init__`, which then becomes heavier, and then we can skip both patterns since everything already exists after `__init__`. Or we should keep the current pattern, since what it actually does is to set `self.__bregman_dist = FunctionalQuadraticPerturb`.

Which way is preferred?

odl/solvers/functional/functional.py
aringh: If this is removed, how should the comment by @mehrhardt be handled?

More precisely: should I completely removed this function, or should I just remove the `@property` flag and rename it to `_subgrad`? If I should completely remove it, should I work with `self.__subgrad` in the rest of the functional?

kohr-h: We're talking about the `initialized` property, which I think is a bit weird to have, since I would think that once the object has been constructed and initialized, i.e., `__init__` is over, it should be initialized. And it's also not necessary for the current implementation, it could easily be changed to the `... is None` check.

odl/solvers/functional/functional.py
kohr-h: Yes, my bad. But in that case we can change the pattern for the functional itself from creating an instance and caching it to creating it on the fly, like this:

```python
def __init__(self, ...):
    ...
    self.__constant = None

def _constant(self):
    if self.__constant is None:
        self.__constant = ...  # compute it
    return self.__constant

def _quad_perturb(self, const):
    return FunctionalQuadraticPerturb(
        self.functional, linear_term=-self.subgrad,
        constant=const)

def _call(self, x):
    return self._quad_perturb(self._const())(x)
 
@property
def convex_conj(self):
    return self._quad_perturb(self._const()).convex_conj
 
@property
def proximal(self):
    # Optimization! Proximal ignores the constant
    return self._quad_perturb(0).proximal
 
@property
def gradient(self):
    try:
        grad = self.functional.gradient
    except NotImplementedError:
        raise NotImplementedError(
            '`self.functional.gradient` is not implemented for '
            '`self.functional` {}'.format(self.functional))

    grad = grad - ConstantOperator(self.subgrad)
    return grad
```
If at some point we conclude that this creation of operators is too expensive (which I doubt will happen anytime soon) we can still choose to add caching, or more precisely, memoization.

aringh: I many of the nonsmooth solvers we create a new proximal operator in each iteration. Not storing the `FunctionalQuadraticPerturb` would slow down iterations in these solvers when using Bregman functionals since we would have to recompute the constant and reinitialize the `FunctionalQuadraticPerturb` in each iteration.

I would prefer to either do all computations in `__initi__`, which would then be more expensive and take longer time, or to use the current pattern that initializes the `FunctionalQuadraticPerturb` first time it is needed.

odl/solvers/functional/functional.py
aringh: Ah! Sorry, my bad. I was looking at the part of the code that was marked in colors :sweat_smile:

odl/solvers/functional/functional.py
kohr-h: > I many of the nonsmooth solvers we create a new proximal operator in each iteration. Not storing the FunctionalQuadraticPerturb would slow down iterations in these solvers when using Bregman functionals since we would have to recompute the constant and reinitialize the FunctionalQuadraticPerturb in each iteration.

Well yes, but the same holds for the current implementation.



> I would prefer to either do all computations in __initi__, which would then be more expensive and take longer time

That's a valid option.



> or to use the current pattern that initializes the FunctionalQuadraticPerturb first time it is needed.

I would call that a software anti-pattern: to create an invalid object and only initialize (and thereby make it valid) it when first used. That's why I suggested to put all the caching in one place (`_constant()`) and have all methods go through the accessors that cache.

mehrhardt:
Although @kohr-h does not like it, I kind of agree with you @aringh that it would be good to make the ODL objects as lightweight as possible and only compute things when needed. Similarly it would be good if there is a way of recycling computations, i.e. one can give additional arguments for the computations like J(x0) for D_J(x, x0) but also access J(x0) if one wants to use it for something else. Of course all of this needs to happen in a way that does not distract the "simple" use of ODL without any of these.

kohr-h:
> Although @kohr-h does not like it, I kind of agree with you @aringh that it would be good to make the ODL objects as lightweight as possible and only compute things when needed.

I don't understand where our views conflict. If you look at the solution I proposed, it doesn't compute the constant or the quadratic perturbation functional before it's really needed. The only real difference between the version before the current and my suggestion is that the computation of the constant is put in one specific place, which simplifies its usage and doesn't leave the object in a partly initialized state.
So yes, I agree that it's good to have a lightweight implementation that performs expensive computations on demand, and also to have an implementation that makes it easy to use in the right way (also internally). That's what I tried to advocate.

That said, I won't boycott an alternative implementation as long as it's sound. I might just file a PR later on to restart the discussion :stuck_out_tongue_winking_eye:

mehrhardt:
Then it was a misunderstanding. What about the option to provide or extract `J(x0)` for the Bregman distance `D_J(x, x0)`?

odl/solvers/functional/functional.py
```diff
@@ -1343,43 +1339,37 @@ class BregmanDistance(Functional):
...
-
     [Bur2016] Burger, M. *Bregman Distances in Inverse Problems and Partial
     Differential Equation*. In: Advances in Mathematical Modeling, Optimization
     and Optimal Control, 2016. p. 3-33.
-
-    .. _[Bur2016]:  https://arxiv.org/abs/1505.05191
     """
 
-    def __init__(self, functional, point, func_subgrad=None):
+    def __init__(self, functional, point, subgrad=None):
```
mehrhardt: I think the `=None` should be removed, no?

kohr-h: Correct, I overlooked that

kohr-h:
> Then it was a misunderstanding. What about the option to provide or extract J(x0) for the Bregman distance D_J(x, x0)?

For the external access part, that's doable by simply splitting the cached constant attribute into two cached attributes like `func_at_point` and `subgrad_inner_point`.

Regarding pre-computed values of `J(x0)` I'm not so sure. It seems like making the interface more cluttered, and there should be a good reason for it. How often do you think that would be useful? How likely is it that the point changes but the subgradient doesn't?
We could make this work by an "interface for the brave" by doing this:
```python
class BregmanDistance(...):
    ...
    def func_at_point(self):
        # Do the caching thing
        return self._func_at_point  # only one underscore!

dist = BregmanDistance(...)
dist._func_at_point = my_precomputed_value
```

mehrhardt:
I don't know how often this is needed. Perhaps we should not include for now as there might not be an immediate use case. I was just thinking of allowing ODL an option to create objects with nearly 0 overhead, as `J(x0)` can be costly.

kohr-h:
Agreed, I think we can skip this for now, given that there's a relatively simple way of doing it without changing the (public) API.

aringh:
I have done the last little update. The interface is kept clean, with no extra options of providing anything. As mentioned above, anyone how wants to add to this can create a PR later ;-)

I will merge after the tests.

Issue #1342, opened 2018-05-08
Possibility to turn off color bars in 2D image plots?


aringh:
Should we add a possibility to turn off the color barn for the `.show` method for discrete data? One way to do it is to add a bolean flag which is used [here](https://github.com/odlgroup/odl/blob/master/odl/util/graphics.py#L412), e.g., `if method == 'imshow' and show_2d_color_bar:`.

I have a vague recollation of a similar discussion before, but could not find anything in the issue history. The reason that I ask is since I have found a case when this would be nice :-p, but maybe it is to specific and not generic need.

adler-j:
Overall I'm in favor of being careful about adding too many "extras" to these functions. They are after all simply intended to be a quick baseline to see what is happening.

With that said, this sounds quite reasonable and reasonably easy to handle. I'd suggest try it out and see how it looks.

aringh:
Closed via #1343

PR #1343, opened 2018-05-09
Add posibility to turn off color bar in 2d plot

aringh:
Example on how #1342 could be implemented. I found no tests for `odl.util.graphics`, but a minimal working example that shows how it works is given by

```python
import odl
space = odl.uniform_discr([0,0], [1,1], [256,256])
x = odl.util.noise_element(space)
x.show()
x.show(show_2d_color_bar=False)
```

which gives the following images
figure_1
figure_3

Is this desirable? Or is it only adding unnecessary complexity to plotting?

### DIFF
--- a/odl/util/graphics.py
+++ b/odl/util/graphics.py
@@ -160,6 +160,11 @@ def show_discrete_data(values, grid, title=None, method='',
     axis_fontsize : int, optional
         Fontsize for the axes. Default: 16
 
+    colorbar : bool, optional
+        Argument relevant for 2d plots using ``method='imshow'``. If ``True``,
+        include a colorbar in the plot.
+        Default: True
+
     kwargs : {'figsize', 'saveto', ...}, optional
         Extra keyword arguments passed on to display method
         See the Matplotlib functions for documentation of extra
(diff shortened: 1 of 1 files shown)


aringh:
They fail is due to doc build fail. This is observed also in (at least) #1321 and #1340

kohr-h:
There's been an update do `travis-sphinx` on PyPI which seems to have broken things. Need to test it locally with more output.

kohr-h:
Remove the `"<1.7"` part from the Sphinx install command, then everything should be fine. The new version of `travis-sphinx` doesn't work with Sphinx 1.6.x, possibly due to a patch by me :roll_eyes:

kohr-h:
> Remove the "<1.7" part from the Sphinx install command, then everything should be fine. The new version of travis-sphinx doesn't work with Sphinx 1.6.x, possibly due to a patch by me roll_eyes

After rebasing, the error should be fixed.

odl/util/graphics.py
```diff
@@ -160,6 +160,11 @@ def show_discrete_data(values, grid, title=None, method='',
     axis_fontsize : int, optional
         Fontsize for the axes. Default: 16
 
+    colorbar : bool, optional
+        Argument relevant for 2d plots using ``method`` 'imshow'. If the plot
```
kohr-h: 2 quick comments:
- \`\`method\`\` 'imshow' -> \`\`method='imshow'\`\`
- "If" can't start a main clause, I guess "Whether" would be more correct. But generally, explanations like "If \`\`True\`\`, include a colorbar in the plot." are preferred.

PR #1344, opened 2018-05-09
BLD: fix travis doc build and pep8speaks name warning

kohr-h:


kohr-h:
Not a code change, so I'll wait to see that the Travis build succeeds and merge then.

PR #1347, opened 2018-05-15
Proximal of simplex constraint

mehrhardt:
I needed this myself and volunteered to properly implement it in ODL. Please have a look. I think this is worth to be included. It is used for instance in blind deconvolution problems.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -18,9 +18,11 @@ from odl.space import ProductSpace
 from odl.operator import (Operator, ConstantOperator, ZeroOperator,
                           ScalingOperator, DiagonalOperator, PointwiseNorm)
 from odl.solvers.nonsmooth.proximal_operators import (
+    proj_simplex,
     proximal_l1, proximal_convex_conj_l1,
     proximal_l1_l2, proximal_convex_conj_l1_l2,
     proximal_l2, proximal_convex_conj_l2, proximal_l2_squared,
+    proximal_linfty,
     proximal_huber,
     proximal_const_func, proximal_box_constraint,
     proximal_convex_conj_kl, proximal_convex_conj_kl_cross_entropy,
@@ -35,6 +37,8 @@ __all__ = ('ZeroFunctional', 'ConstantFunctional', 'ScalingFunctional',
            'IndicatorZero', 'IndicatorBox', 'IndicatorNonnegativity',
            'IndicatorLpUnitBall', 'IndicatorGroupL1UnitBall',
            'IndicatorNuclearNormUnitBall',
+           'IndicatorSimplex',
+           'IndicatorSumConstraint',
            'KullbackLeibler', 'KullbackLeiblerCrossEntropy',
            'QuadraticForm',
            'SeparableSum', 'MoreauEnvelope')
@@ -114,6 +118,8 @@ class LpNorm(Functional):
             return proximal_l1(space=self.domain)
         elif self.exponent == 2:
             return proximal_l2(space=self.domain)
+        elif self.exponent == np.inf:
+            return proximal_linfty(space=self.domain)
         else:
             raise NotImplementedError('`proximal` only implemented for p=1 or '
                                       'p=2')
@@ -2191,6 +2197,228 @@ class IndicatorNuclearNormUnitBall(Functional):
                                          self.domain,
                                          self.__norm.outernorm.exponent,
                                          self.__norm.pwisenorm.exponent)
+
+
+class IndicatorSimplex(Functional):
+    r"""Indicator functional of simplex.
+
+    Notes
+    -----
+    The functional :math:`F` is given by:
+
+    .. math::
+        F(x)
+        =
+        \begin{cases}
+            0 & \text{if } x_i \geq 0 \forall i \text{ and } \sum_i x_i = r \\
+            +\infty & \text{else.}
+        \end{cases}
+
+    where `r` is the diameter.
+    """
+
+    def __init__(self, space, diameter=1, sum_rtol=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        diameter : positive float, optional
+            Diameter `r` of simplex.
+        sum_rtol : float
+            Relative tolerance for sum comparison.
+
+        Examples
+        --------
+        Example where a point lies outside the unit simplex ...
+
+        >>> space = odl.rn(3)
+        >>> ind_simplex = IndicatorSimplex(space)
+        >>> x = space.one()
+        >>> ind_simplex(x)
+        inf
+
+        ... and one where it lies inside the unit simplex.
+
+        >>> x /= x.ufuncs.sum()
+        >>> ind_simplex(x)
+        0
+        """
+        super(IndicatorSimplex, self).__init__(
+            space=space, linear=False, grad_lipschitz=np.nan)
+        self.diameter = float(diameter)
+
+        if sum_rtol is None:
+            if space.dtype == 'float64':
+                sum_rtol = 1e-10
+            else:
+                sum_rtol = 1e-6
+        self.sum_rtol = sum_rtol
+
+    def _call(self, x):
+        """Return ``self(x)``."""
+
+        sum_constr = abs(x.ufuncs.sum() / self.diameter - 1) <= self.sum_rtol
+
+        nonneq_constr = x.ufuncs.greater_equal(0).asarray().all()
+
+        if sum_constr and nonneq_constr:
+            return 0
+        else:
+            return np.inf
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional.
+
+        The indicator functional is not differentiable over the entire domain.
+        """
+
+        raise NotImplementedError('Not implemented')
+
+    @property
+    def proximal(self):
+        """Return the `proximal factory` of the functional."""
+
+        domain = self.domain
+
+        class ProximalSimplex(Operator):
+            """Proximal operator implemented by the algorithm of [D+2008].
+
+            [D+2008] Duchi, J., Shalev-Shwartz, S., Singer, Y., and Chandra, T.
+            *Efficient Projections onto the L1-ball for Learning in High
+            dimensions*. ICML 2008, pp. 272-279.
+            http://doi.org/10.1145/1390156.1390191
+            """
+
+            def __init__(self, sigma):
+                self.sigma = sigma
+                super(ProximalSimplex, self).__init__(
+                    domain=domain, range=domain, linear=False)
+
+            def _call(self, x, out):
+
+                # projection onto simplex
+                proj_simplex(x, self.radius, out)
+
+        return ProximalSimplex
+
+    @property
+    def convex_conj(self):
+        """The convex conjugate."""
+
+        raise NotImplementedError('Not yet implemented')
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r})'.format(self.__class__.__name__, self.domain)
+
+
+class IndicatorSumConstraint(Functional):
+
+    """The indicator functional of a unit sum constraint.
+
+    Notes
+    -----
+    The functional :math:`F` is given by:
+
+    .. math::
+        F(x)
+        =
+        \\begin{cases}
+            0 & \\text{if } \sum_i x_i = 1 \\\\
+            +\\infty & \\text{else.}
+        \\end{cases}
+    """
+
+    def __init__(self, space, sum_value=1, sum_rtol=None):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        sum_value : float
+            Desired value of the sum constraint.
+        sum_rtol : float
+            Relative tolerance for sum comparison.
+
+        Examples
+        --------
+        Example where a point does not have unit sum ...
+
+        >>> space = odl.rn(3)
+        >>> ind_sum = IndicatorSumConstraint(space)
+        >>> x = space.one()
+        >>> ind_sum(x)
+        inf
+
+        ... and one where it does.
+
+        >>> x /= x.ufuncs.sum()
+        >>> ind_sum(x)
+        0
+        """
+        super(IndicatorSumConstraint, self).__init__(
+            space=space, linear=False, grad_lipschitz=np.nan)
+
+        if sum_rtol is None:
+            if space.dtype == 'float64':
+                sum_rtol = 1e-10
+            else:
+                sum_rtol = 1e-6
+        self.sum_rtol = float(sum_rtol)
+        self.sum_value = float(sum_value)
+
+    def _call(self, x):
+        """Return ``self(x)``."""
+
+        if abs(x.ufuncs.sum() / self.sum_value - 1) <= self.sum_rtol:
+            return 0
+        else:
+            return np.inf
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional.
+
+        The indicator functional is not differentiable over the entire domain.
+        """
+
+        raise NotImplementedError('Not implemented')
+
+    @property
+    def proximal(self):
+        """Return the `proximal factory` of the functional."""
+
+        domain = self.domain
+
+        class ProximalSum(Operator):
+            """Proximal operator."""
+
+            def __init__(self, sigma):
+                self.sigma = sigma
+                super(ProximalSum, self).__init__(
+                    domain=domain, range=domain, linear=False)
+
+            def _call(self, x, out):
+
+                offset = 1 / x.size * (self.sum_value - x.ufuncs.sum())
+                out.assign(x)
+                out += offset
+
+        return ProximalSum
+
+    @property
+    def convex_conj(self):
+        """The convex conjugate."""
+
+        raise NotImplementedError('Not yet implemented')
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r})'.format(self.__class__.__name__, self.domain)
 
 
 class MoreauEnvelope(Functional):
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -37,6 +37,8 @@ __all__ = ('combine_proximals', 'proximal_convex_conj', 'proximal_translation',
            'proximal_box_constraint', 'proximal_nonnegativity',
            'proximal_l1', 'proximal_convex_conj_l1',
            'proximal_l2', 'proximal_convex_conj_l2',
+           'proximal_linfty',
+           'proj_simplex', 'proj_l1',
            'proximal_l2_squared', 'proximal_convex_conj_l2_squared',
            'proximal_l1_l2', 'proximal_convex_conj_l1_l2',
            'proximal_convex_conj_kl', 'proximal_convex_conj_kl_cross_entropy',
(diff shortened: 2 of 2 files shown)


mehrhardt:
I haven't done any ODL testing yet. If you give me some instructions I could give it a try.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
     return ProximalL1L2
 
 
+def proximal_linf(space):
+    """Proximal operator factory of the linf-norm.
+
+    Function for the proximal operator of the functional ``F`` where ``F``
+    is the linf-norm::
+
+        ``F(x) =  \sup_i |x_i|``
```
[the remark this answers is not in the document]
mehrhardt: It is the l_\infty norm so it should be sup. Maybe the name of the functional is misleading?

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+            return 0
+
+        return np.inf
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional. The indicator functional is not
+        differentiable over the entire domain.
+        """
+
```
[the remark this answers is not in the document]
mehrhardt: Difficult. One the one hand I do agree as this is mostly true. On the other hand these are sources for errors that are very difficult to detect. For most algorithms the user should not call the gradient of an indicator any ways. Either it exists and is non-informative or it simply does not exist. Thus, I tend to not define a gradient here.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
...
+    """
+
+    if out is None:
+        out = x.space.element()
+
+    # sort values in descending order
+    x_vec = x.asarray().flatten()
+    x_sor = np.sort(x_vec)[::-1]
+
+    # TODO: explain a little what gets computed here
```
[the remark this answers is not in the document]
mehrhardt: maybe one should remove the TODO and just refer to the paper?

mehrhardt:
Is there a way to automatically find a non-ascii character? I can't find what Travis complains about.

odl/solvers/nonsmooth/proximal_operators.py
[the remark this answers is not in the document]
adler-j: I'd say the name is missleading in that case (or perhaps I need to be educated, but in that case the doc needs to be slightly fleshed out)

odl/solvers/functional/default_functionals.py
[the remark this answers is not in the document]
adler-j: Ok, lets leave it as is until someone has an issue with it!

odl/solvers/nonsmooth/proximal_operators.py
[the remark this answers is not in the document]
adler-j: Works for me!

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
...
+    [D+2008] Duchi, J., Shalev-Shwartz, S., Singer, Y., and Chandra, T.
+    *Efficient Projections onto the L1 -ball for Learning in High dimensions*.
+    ICML 2008, pp. 272–279. http://doi.org/10.1145/1390156.1390191
+
+    See Also
+    --------
+    proj_simplex : projection onto simplex
+    """
+
+    class ProxSimplex(Operator):
```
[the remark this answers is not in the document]
adler-j: Indeed, I'd try doing that.

Overall the split on ODL is mostly for historical reasons, we could probably move lots of stuff around behind the scenes.

adler-j:
No idea about how to find the non-ascii (at least in a whole package) but travis is quite specific:

    Non-ASCII character '\xe2' in file /home/travis/build/odlgroup/odl/odl/solvers/nonsmooth/proximal_operators.py on line 1532, but no encoding declared; see 

perhaps it's easiest to just let travis figure this out for now.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
                                          self.__norm.pwisenorm.exponent)
 
 
+class IndicatorSimplex(Functional):
+
+    """The indicator functional of a simplex in any dimension.
```
kohr-h: Can you use the `r"""` docstring style here please? Saves a lot of double backslashes.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        radius : positive float, optional
+            Radius `r` of simplex.
+        """
+        super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
+        self.radius = radius
```
kohr-h: `float(radius)` so you know it's a float next time.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+
+    def __call__(self, x):
+        """Return whether x is in the unit simplex."""
```
kohr-h: You overloaded the wrong method, use `_call` instead.
The docstring is also a bit misleading since I would assume that it returns a boolean and not 0 or infinity. Actually, """Return \`\`self(x)\`\`.""" is enough here.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+
+        r = self.radius
+        if (x.ufuncs.sum() == r and x.ufuncs.greater_equal(0).asarray().all()):
+            return 0
+
+        return np.inf
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional. The indicator functional is not
```
kohr-h: One line (or split)

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+        differentiable over the entire domain.
+        """
+
+        raise NotImplementedError('No yet implemented')
```
kohr-h: The "yet" doesn't seem to reflect your intention :wink:

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
...
+    --------
+    proj_l1 : projection onto l1-ball
+    """
+
+    if out is None:
+        out = x.space.element()
+
+    # sort values in descending order
+    x_vec = x.asarray().flatten()
+    x_sor = np.sort(x_vec)[::-1]
```
kohr-h: If you do `x_vec.sort()` it sorts in-place (returning `None`), which is slightly faster and likely consumes less memory.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
...
+        ``{ x \in X | x_i \geq 0, \sum_i x_i = r}``
+
+    with `r` being the radius.
+
+    Parameters
+    ----------
+    space : `LinearSpace`
+        Space / domain `X`.
+    radius : positive float, optional
+        Radius `r` of simplex.
```
kohr-h: \`X\` and \`r\` are dead links.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
...
+    out[:] = np.maximum(x - x_avrg[i], 0)
+
+    return out
+
+
+def proximal_simplex(space, radius=1):
+    """Proximal operator factory of the characteristic function of simplex.
+
+    Function for the proximal operator of the functional ``F`` where ``F``
+    is
```
kohr-h: I'd leave the `F` out of the game and just write
```
The proximal operator of the simplex indicator is

.. math::
    \iota_{S^{n-1}_r}(x) = ...
```
Also note that we use the term "indicator function" otherwise, not "characteristic function".

odl/solvers/nonsmooth/proximal_operators.py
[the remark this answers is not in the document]
kohr-h: Yes, for the relatively simple cases we should in the future just inline the class in the `functional.proximal` property. No need for an extra indirection.
We should even consider removing these factories altogether and put the parameters in the `__init__` of the prox operators. The functional property can then forward the arguments in the correct way.

The current situation is only because the initial solvers interface used the prox operators directly, and therefore they had to have a uniform interface.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,231 @@ def _call(self, x, out):
...
+            ----------
+            sigma : positive float
+                Step size parameter
+            """
+            super(ProximalLInf, self).__init__(
+                domain=space, range=space, linear=False)
+            self.sigma = float(sigma)
+
+        def _call(self, x, out):
+            """Apply the operator to ``x`` and stores the result in ``out``."""
```
kohr-h: Not really the best English, but I guess it's copy & paste.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+        """The convex conjugate."""
+
+        raise NotImplementedError('Not yet implemented')
+
+    def __repr__(self):
+        """Return ``repr(self)``."""
+        return '{}({!r})'.format(self.__class__.__name__, self.domain)
+
+
+class IndicatorSum(Functional):
```
kohr-h: This name is too broad for my taste. It should at least use the word "constraint".

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        """
+        super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
+
+    def __call__(self, x):
+        """Return whether x has unit sum."""
+
+        if x.ufuncs.sum() == 1:
```
kohr-h: With this implementation the value will practically always be infinity. A problem? Maybe consider allowing some tolerance.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+
+        return np.inf
+
+    @property
+    def gradient(self):
+        """Gradient operator of the functional. The indicator functional is not
+        differentiable over the entire domain.
+        """
+
+        raise NotImplementedError('No yet implemented')
```
kohr-h: Same as above

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+            def __init__(self, sigma):
+                self.sigma = sigma
+                super().__init__(domain=domain, range=domain, linear=False)
+
+            def _call(self, x, out):
+
+                # projection onto simplex
+                proj_simplex(x, self.radius, out)
+
+        return ProximalSimplex
```
kohr-h: This should be the full class from the other file. If things get too long, you can outsource it again.

odl/solvers/functional/default_functionals.py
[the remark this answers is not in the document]
adler-j: `ODL_I_AM_AN_ENGINEER`

odl/solvers/nonsmooth/proximal_operators.py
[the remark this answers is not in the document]
mehrhardt: I changed the documentation slightly to call "linf" "l_\infty" in the text. As "linfimum" does not make much sense in this context I suppose "linf" is fine here.

odl/solvers/functional/default_functionals.py
mehrhardt: What do you mean exactly?

odl/solvers/functional/default_functionals.py
mehrhardt: Good idea. How do we achieve that? It is also not clear how much tolerance to allow.

odl/solvers/functional/default_functionals.py
mehrhardt: I agree but I now chose to remove the class from the other file and only have it here inline.

odl/solvers/nonsmooth/proximal_operators.py
mehrhardt: How do I improve on this?

aringh:
For finding non-ascii, I found [this](https://sites.google.com/site/opensourceconstriubtions/ettl-martin-1/tutorials/how-to-detect-non-ascii-characters-using-the-linux-command-line-interface) which might help.

odl/solvers/functional/default_functionals.py
kohr-h: The summary line should be one line. If more explanation is necessary, that should be in a new block of text, separated by a blank line.
We're largely following [this convention](https://www.python.org/dev/peps/pep-0257/).

Of course, one can argue about the usefulness of summary lines like "Gradient operator of the functional." But I have yet to see better solutions.

odl/solvers/nonsmooth/proximal_operators.py
kohr-h: Use double-ticks: \`\`X\`\` or \`\`r\`\`
They are merely treated as inline code by Sphinx (the doc generator) and not interpreted as links to class names, glossary terms or similar.

odl/solvers/functional/default_functionals.py
kohr-h: You could use the built-in `round` function: `round(x.ufuncs.sum(), ndigits=6) == 1`. 

The tolerance level is a bit arbitrary, agreed, but since we're comparing against 1, the error is also "normalized". So any value that "works" for practical purposes (i.e., single precision arithmetic) should be okay.

As a guidance, consider the following experiment:
```python
>>> x = np.random.uniform(0, 1, size=int(1e9))
>>> y = x.astype(np.float32)
>>> xsum = x.sum()
>>> x /= xsum
>>> y /= xsum
>>> abs(1 - x.sum())
1.5543122344752192e-15
>>> abs(1 - y.sum())
8.106231689453125e-06
```
This is a relatively large example, but it shows the order of error that you'd have to expect from *stable* computations. So maybe it makes sense to define the required precision based on the data type.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
                                          self.__norm.pwisenorm.exponent)
 
 
+class IndicatorSimplex(Functional):
+
+    r"""The indicator functional of a simplex in any dimension.
```
adler-j: I'm not sure about the style we follow, but I think we skip the "The" here, right? Also "in any dimension" is basically implicit in ODL, we usually try to only mention if there is a restriction :)

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,169 @@ def __repr__(self):
...
+
+    Notes
+    -----
+    The functional :math:`F` is given by:
+
+    .. math::
+        F(x)
+        =
+        \begin{cases}
+            0 & \text{if } x_i \geq 0 \forall i \text{ and } \sum_i x_i = r \\
```
adler-j: Isn't this a non-convex functional? Wouldn't it be better to only require that the sum is less than or equal to r?

mehrhardt: This is convex as the simplex is a convex set. For deconvolution it is very important to have the sum equal to r in order to get non-trivial solutions.

mehrhardt:
I believe I made all required changes, cleaned up the commits and removed the non-ascii characters. If Travis is happy, I am good to go.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,197 @@ def __repr__(self):
                                          self.__norm.pwisenorm.exponent)
 
 
+class IndicatorSimplex(Functional):
```
kohr-h: Add the functional to `__all__`, to make it available under `odl.solvers` directly.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,197 @@ def __repr__(self):
...
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        diameter : positive float, optional
+            Diameter `r` of simplex.
+        """
+        super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
+        self.diameter = float(diameter)
+
+    def _call(self, x, ndigits=None):
```
kohr-h: :+1: for exposing this as a user option. Unfortunately, this operator can't be instantiated since it's not allowed to add arguments to `_call`:
```python
>>> ind = odl.solvers.functional.default_functionals.IndicatorSimplex(odl.rn(3))
...
ValueError: bad signature '_call(self, x, ndigits=None)': output parameter must be called 'out', got 'ndigits'
Possible signatures are ('[, **kwargs]' means optional):

_call(self, x[, **kwargs])
_call(self, x, out[, **kwargs])
_call(self, x, out=None[, **kwargs])
_call(self, x, *, out=None[, **kwargs])

Static or class methods are not allowed.
```
If you really want to add it here, you can use `**kwargs` in the signature and then do `ndigits = kwargs.pop('ndigits', None)`, then proceed as now.

But to me it seems better to add it to the initializer, because that exposes it to users. In `_call`, no-one will find it, and things like `OperatorComp` won't be able to use it.

It might also be easier to convey the meaning of the parameter by using a relative tolerance, e.g. `sum_rtol`, and then comparing the normalized sum to 1: `abs(x.ufuncs.sum() / self.diameter - 1) <= self.sum_rtol`. (`rtol` is a standard name in NumPy).

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2198,197 @@ def __repr__(self):
...
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        space : `DiscreteLp` or `FnBase`
+            Domain of the functional.
+        """
+        super().__init__(space=space, linear=False, grad_lipschitz=np.nan)
+
+    def __call__(self, x, ndigits=None):
```
kohr-h: Still overriding the wrong method, and the same issue with `ndigits` as above.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,166 @@ def _call(self, x, out):
...
+    See Also
+    --------
+    proj_l1 : projection onto l1-norm ball
+    """
+
+    if out is None:
+        out = x.space.element()
+
+    # sort values in descending order
+    x_sor = x.asarray().flatten().sort()[::-1]
```
kohr-h: This doesn't work because `sort()` returns `None`.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,166 @@ def _call(self, x, out):
...
+                domain=space, range=space, linear=False)
+            self.sigma = float(sigma)
+
+        def _call(self, x, out):
+            """Return ``self(x)``."""
+
+            radius = 1
+            proj_l1(x, radius, out)
+            out *= -1
+            out += x
```
kohr-h: This will give wrong results when `x is out`. In this step, you need the original `x`. Something like this is needed:
```python
if x is out:
    x = x.copy()
```
Furthermore, you can use `out.lincomb(-1, out, 1, x)` to do this in one step.

mehrhardt:
Everything cleaned up, tests passed, doctests implemented. Anything else needed?

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2199,225 @@ def __repr__(self):
...
+        --------
+        Example where a point does not have unit sum ...
+
+        >>> space = odl.rn(3)
+        >>> ind_sum = IndicatorSumConstraint(space)
+        >>> x = space.one()
+        >>> ind_sum(x)
+        inf
+
+        ... and one where it does.
```
kohr-h: Here we need a blank line so the docs are produced correctly.

odl/solvers/functional/default_functionals.py
```diff
@@ -2193,6 +2199,225 @@ def __repr__(self):
...
+
+            def __init__(self, sigma):
+                self.sigma = sigma
+                super(ProximalSum, self).__init__(
+                    domain=domain, range=domain, linear=False)
+
+            def _call(self, x, out):
+
+                offset = 1 / x.size * (self.sum_value - x.ufuncs.sum())
+                out.assign(x + offset)
```
kohr-h: Not a must, but you can avoid a copy by doing
```python
out.assign(x)
out += offset
```

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1441,6 +1443,167 @@ def _call(self, x, out):
...
+            super(ProximalLInfty, self).__init__(
+                domain=space, range=space, linear=False)
+            self.sigma = float(sigma)
+
+        def _call(self, x, out):
+            """Return ``self(x)``."""
+
+            radius = 1
+            proj_l1(x, radius, out)
+            out.lincomb(-1, out, 1, x)
```
kohr-h: We still have the alignment problem here. If `x is out`, `x` won't be the original `x` anymore after the `proj_l1` step, but you need the original `x` in the final step. So you need to check for that situation and retain a copy of the original if necessary:
```python
if x is out:
    x = x.copy()
```
This will de-alias `x` and `out`.

mehrhardt:
@kohr-h, please keep being "annoying" :) This just improves my code!

kohr-h:
Thanks a lot @mehrhardt!

I'll make an issue about the missing tests.

PR #1351, opened 2018-05-18
BUG: Handle list of length one appropriately for SamplingOperator.adj…

### LINKED ISSUE #1350: Adjoint of SamplingOperator error in 1D
The adjoint of the SamplingOperator is not working for me. 

ODL version.
'0.6.1.dev0'

Minimal example
````
import numpy as np
import odl

U = odl.uniform_discr(0, 100, 100) 
u = U.one()

idx = [2,3,4]

E = odl.SamplingOperator(U,idx)

E.adjoint(E(u))
```

adler-j:
…oint, closes #1350

### DIFF
--- a/odl/operator/tensor_ops.py
+++ b/odl/operator/tensor_ops.py
@@ -998,8 +998,14 @@ def _normalize_sampling_points(sampling_points, ndim):
     elif ndim == 1:
         if isinstance(sampling_points, Integral):
             sampling_points = (sampling_points,)
-        sampling_points = [np.array(sampling_points, dtype=int, copy=False,
-                                    ndmin=1)]
+        sampling_points = np.array(sampling_points, dtype=int, copy=False,
+                                   ndmin=1)
+
+        # Handle possible list of length one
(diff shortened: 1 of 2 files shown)


kohr-h:
A test that checks the reported bug code would be nice.

adler-j:
I guess I can get some time off to add tests for this class, but it's quite an effort. Perhaps just merge this and I'll make an issue about it. That way our user can get the fix asap.

kohr-h:
But I guess this one test case would be pretty simple, no? That would at least ensure that we don't resurrect the bug when changing the code again.

adler-j:
I guess I could, but this class has no real tests at all (its all doctests)

kohr-h:
I know, what's wrong with just adding one test function for only this to the `tensor_ops_tests.py` (I guess)? That also makes pretty obvious that there are tests missing :-)

adler-j:
So I added the requested test now, all OK?

kohr-h:
> So I added the requested test now, all OK?

Way more than I expected, but all the better.

PR #1352, opened 2018-05-18
Add license statement for "Cambridge" images

mehrhardt:
This is the requested license statement. The images are licensed CCBY4.0 as indicated in the ODL file and the license file in the folder next to the images.


adler-j:
Thanks for the quick fix! @aringh will like this :D

aringh:
Yes! Thank you @mehrhardt

Issue #1354, opened 2018-05-25
Easy way to run tests in odl.contrib?


aringh:
As it is now, the tests in `odl.contrib` are not run by travis (neither when one runs `py.test` in the terminal). Is there an easy way to run all the test in the contrib-package? Could one add a flag to pytest so that it also runs the test in the contrib-package?

**Edit**: individual tests can be run by
```bash
py.test path/to/test/test_myfun.py
```
but for example
```bash
py.test odl/contrib/
```
runs no test.

kohr-h:
How about
```
pytest odl/contrib/<something>/test/
```
? Does that run the tests?

adler-j:
We've intentionally decided to not have a "run all contrib tests" since that is basically guaranteed to fail since any one user will likely not have all dependencies installed at once. Some contrib packages might even be OS specific. I'm sure @kohr-h's approach should work though.

aringh:
Unfortunately I don't get @kohr-h's approach to work. For example:

```bash
pytest odl/contrib/fom/test/
============================= test session starts ==============================
platform linux -- Python 3.6.4, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
rootdir: [...]/odl/odl, inifile: pytest.ini
plugins: odl-0.6.1.dev0, pep8-1.0.6
collected 0 items                                                              

========================= no tests ran in 0.00 seconds =========================
```

I understand the arguments against having something that runs all contrib test. However, one issue with not being able to run all tests easily is that I think we have test fails we do not know about in the contrib package. For example, I get one test fail when I run the tests for unsupervised foms.

aringh:
So, I did not get the suggestion by @kohr-h to work. I get
```
$ pytest odl/contrib/fom/test/
============================= test session starts ==============================
platform linux -- Python 3.6.1, pytest-3.0.5, py-1.4.33, pluggy-0.4.0
rootdir: [...]/odl/odl, inifile: pytest.ini
plugins: odl-0.6.1.dev0, pep8-1.0.6
collected 0 items 

========================= no tests ran in 0.00 seconds =========================
``` 
However, I get the following to work
```
$ pytest odl/contrib/fom/test/*
============================= test session starts ==============================
platform linux -- Python 3.6.1, pytest-3.0.5, py-1.4.33, pluggy-0.4.0
rootdir: [..]/odl/odl, inifile: pytest.ini
plugins: odl-0.6.1.dev0, pep8-1.0.6
collected 14 items 

[...]

PR #1355, opened 2018-05-28
TST: update tests unsupervised foms.

aringh:
Increase dimension of spaces to reduce risk of random test fails.

### DIFF
--- a/odl/contrib/fom/test/test_unsupervised.py
+++ b/odl/contrib/fom/test/test_unsupervised.py
@@ -23,7 +23,7 @@ def test_estimate_noise_std_constant_1d():
 
 def test_estimate_noise_std_normal_1d():
     """Verify ``estimate_noise_std(N(0, 1)) == 1`` in 1d."""
-    img = np.random.randn(10)
+    img = np.random.randn(1000)
     result = odl.contrib.fom.estimate_noise_std(img)
     expected = np.std(img)
     assert pytest.approx(result, abs=0.2) == expected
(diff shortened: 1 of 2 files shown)


aringh:
I didn't seem to take to long on my laptop. My experience was that if we don't increase the sizes, then we need to increase tolerances up to around 0.4 to not get random fails to often. And that kind of tolerance seemed a little bit ridiculous.

I think that the reason for that noone has noticed this (or at least reported it) is that it is in the `contrib` package. In connection to the discussion in #1354, since this will probably only be run manually by anyone how explicitly wants to test if this works as expected, to me it doesn't seem to bad even if the runtime would be as much as a second.

adler-j:
0.19 is *a lot* of time. We have 5000 tests, at that runtime it would take 20 minutes to run them :/

aringh:
I know. But I still claim that if the tests are _never_ run unless one explicitly does `pytest odl/contrib/fom/test/test_unsupervised.py`, those 0.19 seconds is not what takes time. If I do take the time to run these tests, I therefore prefer that they actually do not fail. Of course, we could increase tolerance levels in the test, but as I wrote you need to increase them quite a lot it seems.

Go or no go?

adler-j:
Oh, this is in contrib? I'm a slow thinker at times...

Just go ahead!

adler-j:
Just some closing words: For contrib tests may easily take more time if we need, we should probably even add more tests to this...

PR #1356, opened 2018-05-29
Adding correctly the shearlet inverse adjoint operator

arsenal9971:
Hi, today working with odl and shearlab I realized that the definition of the adjoint of the inverse of the Shearlet operator was not correct. Both adjoint of the inverse and inverse of the adjoint are the same, so I assumed as the beginning that I could call subclasses defined in other classes. This was not the case.

Now it is fixed.


arsenal9971:
Now I am still need to do some fixes to the inverse adjoint in order to have type of odl.Operator

aringh:
Hi @arsenal9971! It looks like there is something weird with the commit history in the PR. I guess it is only the three last commits (383d95c, 57db2de and b618355) which contains the updates.

I assume that what has happened is that you still have the full history in your local repository, while @kohr-h squashed all commits in the previous PR into one single commit: 397fb3ea. Therefore, when you pushed this you got all the old commits coming along.

Being a git-novice, I would just have `pull`ed `master`, created a new branch, `cherry-pick`ed  the three commits, and pushed the new branch to create a new PR. However, @kohr-h or @adler-j might have other suggestions?

arsenal9971:
Hi @aringh, indeed only the last three commits contains the updates (fyi I still need to do a last fix, so it would be soon a fourth commit). About that, I don't really know how to fix this git issue, I actually took care to pull upstream/origin to maintain up to date my fork with the upstream repo.

kohr-h:
Yeah, cherry-picking is likely the easiest option:
```
git fetch origin
git checkout origin/master
git checkout -b fix_shearlab_invadj
git cherry-pick 0e78ea^..master
```
The last command cherry-picks the range of commits from *one before* 0e78ea (that's the `^`) until current `master`.

kohr-h:
Closing in favor of #1363

PR #1353, opened 2018-05-25
New reco pipline branch

aringh:
This replaces the PR #1188. I think I have addressed all open issues in that PR, and the branch is also rebased on master.

PR #1280, opened 2018-02-01
Reconstruction pipeline

adler-j:
TODO:

- [x] [#1184](https://github.com/odlgroup/odl/pull/1184) "TST: Add general tests for supervised FOMs"
- [x] [#1185](https://github.com/odlgroup/odl/pull/1185) "ENH: Pull request of group 1 (Workshop day, work on FOMs)"
- [x] [#1187](https://github.com/odlgroup/odl/pull/1187) "Optimize parameters"
- [x] <s>[#1188](https://github.com/odlgroup/odl/pull/1188)</s> [#1357](https://github.com/odlgroup/odl/pull/1357) "Reco pipeline mob"

After these are in we can review and merge to master.

Edited by @aringh

PR #1357, opened 2018-05-29
Newer reco pipeline

aringh:
Previously I did not look carefully, so when I closed #1188 and opened #1353, I rebased on master and wanted to merge straight to master. But I do not know if there are any commits in #1280 that would then be dropped... so now I took the same commits as in #1188 (plus the two extra), and rebased on #1280.

So: which one of this and #1353 do we want to keep?

### DIFF
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -19,7 +19,7 @@ __all__ = ('mean_squared_error', 'mean_absolute_error',
 
 def mean_squared_error(data, ground_truth, mask=None,
                        normalized=False, force_lower_is_better=False):
-    """Return mean squared L2 distance between ``data`` and ``ground_truth``.
+    r"""Return mean squared L2 distance between ``data`` and ``ground_truth``.
 
     See also `this Wikipedia article
     <https://en.wikipedia.org/wiki/Mean_squared_error>`_.
@@ -51,7 +51,7 @@ def mean_squared_error(data, ground_truth, mask=None,
     The FOM evaluates
 
     .. math::
-        \mathrm{MSE}(f, g) = \\frac{\| f - g \|_2^2}{\| 1 \|_2^2},
+        \mathrm{MSE}(f, g) = \frac{\| f - g \|_2^2}{\| 1 \|_2^2},
 
     where :math:`\| 1 \|^2_2` is the volume of the domain of definition
     of the functions. For :math:`\mathbb{R}^n` type spaces, this is equal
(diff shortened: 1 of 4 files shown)


aringh:
I will poke this again :smile: I think it would be good to have this in place next week for the meeting.

1. Keep this PR or #1353 ?
2. If the above answer is this PR, then a quick review by @sbanert or @adler-j would be nice. Otherwise, a review of #1353 would be needed.

sbanert:
The answer to the first question is clear: Keep this one.

adler-j:
If possible, make or into 1353

adler-j:
So basically, just fix the bugs here and we'll click merge

odl/contrib/fom/supervised.py
```diff
@@ -310,21 +331,21 @@ def range_difference(data, ground_truth, mask=None, normalized=False):
...
-            \\big(\\max(g) - \\min(g) \\big)
+            \big(\max(f) - \min(f) \big) -
+            \big(\max(g) - \min(g) \big)
             \Big|}{
             \Big|
-            \\big(\\max(f) - \\min(f) \\big) +
-            \\big(\\max(g) - \\min(g) \\big)
+            \big(\max(f) - \min(f) \big) +
+            \big(\max(g) - \min(g) \big)
             \Big|}
```
sbanert: Same as above.

odl/contrib/fom/supervised.py
```diff
@@ -460,7 +481,7 @@ def false_structures_mask(foreground, smoothness_factor=None):
     unique = np.unique(foreground)
     if not np.array_equiv(unique, [0., 1.]):
         raise ValueError('`foreground` is not a binary mask or has '
-                          'no true values {!r}'.format(unique))
+                         'no true values {!r}'.format(unique))
```
sbanert: Is it ``truth values``?

aringh: I think it is `true values`. The mask is suppose to indicate which parts of the image that should be de-emphasized (the foreground), and this error is returned if the values in `foreground` does not take exactly the values 0 and 1, i.e., nothing else than 0 and 1 and at least one pixel with each value. But that means that the error message is wrong anyway, because a mask of only ones (so all values are `true`) would also return this error.

aringh:
I think the test fails are due to #1344 or other old incompatibilities with travis, since this is based on an old commit.

In any case, this code only makes changes to `contrib` and these test are not run by travis. So it cannot possibly be any change in this code that breaks the tests. I say let's merge this, and then rebase the other branch and merge that too.

sbanert:
Great work!

PR #1362, opened 2018-06-12
DOC/TST: Add rounding in doctest of power_method_opnorm. Closes #1361.

### LINKED ISSUE #1361: Doctest of power_method_opnorm randomly fails
See for instance [here](https://travis-ci.org/odlgroup/odl/builds/390389954). Better to use approximate equality.

sbanert:
I rounded the result in the doctest to three decimals to obtain a stable result. Maybe this is not the most elegant thing to do in a doctest, but it seems to solve the problem reliably.

### DIFF
--- a/odl/operator/oputils.py
+++ b/odl/operator/oputils.py
@@ -154,11 +154,12 @@ def power_method_opnorm(op, xstart=None, maxiter=100, rtol=1e-05, atol=1e-08,
 
     Examples
     --------
-    Verify that the identity operator has norm 1:
+    Verify that the identity operator has norm 1 (within a precision of 0.001):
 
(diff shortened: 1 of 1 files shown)


adler-j:
Agree with @kohr-h, very nice way to format this, but the comment is overly specific

sbanert:
I wanted to explain the magic number 3 without having the user look up the definition of the `round` function, but since this is rather basic, I leave it out.

kohr-h:
> I wanted to explain the magic number 3 without having the user look up the definition of the round function, but since this is rather basic, I leave it out.

Then why not `round(result, ndigits=3)`? No more explanation necessary.

sbanert:
That’s even better, and I did not think about giving the keyword. Thank you!

PR #1358, opened 2018-05-29
Fix shearlab invadj

arsenal9971:
Here is the cherry picking of the fix

PR #1363, opened 2018-06-13
Shearlab inv adj new

arsenal9971:
Finally, lets see if it works

### DIFF
--- a/odl/contrib/shearlab/shearlab_operator.py
+++ b/odl/contrib/shearlab/shearlab_operator.py
@@ -74,12 +74,7 @@ class ShearlabOperator(odl.Operator):
             """
 
             def __init__(self):
-                """Initialize a new instance.
-                Parameters
-                ----------
-                op : `ShearlabOperator`
-                    The operator which this should be the adjoint of.
-                """
+                """Initialize a new instance."""
                 self.mutex = op.mutex
                 self.shearlet_system = op.shearlet_system
                 super(ShearlabOperatorAdjoint, self).__init__(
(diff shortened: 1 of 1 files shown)


adler-j:
So, before I go ahead here. We can close both #1358 and #1356?

odl/contrib/shearlab/shearlab_operator.py
```diff
@@ -171,8 +179,51 @@ def _call(self, x):
...
+                    odl.contrib.shearlab.ShearlabOperator
+                    """
+
+                    def __init__(self):
+                        """Initialize a new instance.
+                        Parameters
+                        ----------
+                        op : `ShearlabOperator`
+                              The operator which this should be the
+                              inverse of the adjoint of.
```
kohr-h: Remove the Parameters section

kohr-h:
> So, before I go ahead here. We can close both #1358 and #1356?

Okay, since the other two are littered with off-topic discussions about git, let's go with this one. I'll close the other two.

odl/contrib/shearlab/shearlab_operator.py
aringh: As pointed out in #1358, this also goes for lines 78, 117, and 164 (although these lines are not changed in these commits).

arsenal9971:
I already did the change, push to my branch `fix_shearlab_invadj`, and then  `git push -f origin kohr-h/shearlab_inv_adj_new:fix_shearlab_invadj.` But I don't see the change in this pull request.

kohr-h:
Then I guess your branch has the update. It doesn't make any more sense to push *my* branch again to `origin/fix_shearlab_invadj` (which you apparently did), because it doesn't have any changes. 

What you instead need to do is `git push origin fix_shearlab_invadj` or more explicitly `git push fix_shearlab_invadj:fix_shearlab_invadj` (*your* local `fix_shearlab_invadj` to the remote one in `origin` with the same name).

PR #1369, opened 2018-06-14
TST: update power method test. Closes #1364

### LINKED ISSUE #1364: Random fail for power method test
See [here](https://travis-ci.org/odlgroup/odl/jobs/391637933#L1224), where the code changes for the branch has nothing to do with the power method.

(Note that it is not the same issue as #1361, which was a doc-test for the same method.)

aringh:
Now the test is also up to date with the comment, given that default behavior of power method was changed in 2b3469641b.

### DIFF
--- a/odl/test/operator/oputils_test.py
+++ b/odl/test/operator/oputils_test.py
@@ -171,12 +171,13 @@ def test_power_method_opnorm_nonsymm():
     true_opnorm = 6
 
     # Start vector (1, 1) is close to the wrong eigenvector
-    opnorm_est = power_method_opnorm(op, maxiter=50)
+    xstart = odl.rn(2).element([1, 1])
+    opnorm_est = power_method_opnorm(op, xstart=xstart, maxiter=50)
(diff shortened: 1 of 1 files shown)


odl/test/operator/oputils_test.py
```diff
@@ -171,7 +171,8 @@ def test_power_method_opnorm_nonsymm():
     true_opnorm = 6
 
     # Start vector (1, 1) is close to the wrong eigenvector
-    opnorm_est = power_method_opnorm(op, maxiter=50)
+    xstart = odl.rn(2).element([1, 1])
+    opnorm_est = power_method_opnorm(op, xstart=xstart, maxiter=50)
```
sbanert: I noticed that the order of the keyword arguments is different in lines 175 and 180. Since the order in the definition of `power_method_opnorm` is `op`, `xstart`, `maxiter`, maybe line 180 should be adapted to this. Otherwise, this looks good to me.

aringh:
Updated accordingly. Will merge after tests are completed.

adler-j:
Great fix!

PR #1370, opened 2018-06-14
Added Callback.final() and saveto to CallbackShowConvergence

miallo:
Now a callback can perform a final task after the optimizer finished. This is used for `CallbackShowConvergence` to be able to save the figure after the iteration finished.

The function name `final`is probably not optimal but this was the best I could come up with...

And two minor typos where corrected

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -2295,7 +2295,7 @@ class Huber(Functional):
     .. math::
         F(x) = \\int_\Omega f_{\\gamma}(||x(y)||_2) dy
 
-    where :mth:`||\cdot||_2` denotes the Euclidean norm for vector-valued
+    where :math:`||\cdot||_2` denotes the Euclidean norm for vector-valued
     functions which reduces to the absolute value for scalar-valued functions.
     The function :math:`f` with smoothing :math:`\\gamma` is given by
 
--- a/odl/solvers/iterative/iterative.py
+++ b/odl/solvers/iterative/iterative.py
@@ -277,6 +277,9 @@ Conjugate_gradient_on_the_normal_equations>`_.
         if callback is not None:
             callback(x)
 
+    if callback is not None:
+        callback.final()
+
 
 def exp_zero_seq(base):
     """Default exponential zero sequence.
@@ -386,6 +389,9 @@ def gauss_newton(op, x, rhs, niter, zero_seq=exp_zero_seq(2.0),
 
         if callback is not None:
             callback(x)
+
+    if callback is not None:
+        callback.final()
 
 
 def kaczmarz(ops, x, rhs, niter, omega=1, projection=None, random=False,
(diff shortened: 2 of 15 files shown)


miallo:
@kohr-h Yes, I agree that `done()` might be a better name.

But what exactly do you mean with 
> this would work worse, depending on what final() does.

? Of course if you write something destructive in there then a warmstart won't work. But the default for all but `ShowConvergence` is just `pass`.
The only thing one could think about is if one should overwrite the saved figure to make the warmstart for the `CallbackShowConvergence` in combination with `saveto` work (the default implementation of `matplotlib.savefig` does not overwrite the file). But I guess that the `saveto`-feature is more common if you don't run it with a warmstart condition, but rather if you want to save the results of a script (e.g. if you start different jobs with different parameters and want to save the convergence to optimize the parameters - at least that was my use-case).

kohr-h:
> But what exactly do you mean with
> 
>     this would work worse, depending on what final() does.
> 
> ? Of course if you write something destructive in there then a warmstart won't work. But the default for all but ShowConvergence is just pass.

Well you can't assume that this is the way it always will be, and you cannot guess what users will do either. They can write whatever they want in `done()` of their own callback, and I'd expect them to be annoyed when they run for N iterations, do something else, then again N iterations, only to find that their `done()` method had already been called in between and has done something they wanted to do at the very end.

So I guess I mean that the optimization method shouldn't assume that the user is done when it exits.

miallo:
What about renaming it to `did_finish_optimization` instead? If the callback has a key that says `saveto` I recon the user will expect it to save something without having to manually call an additional method by hand (as is the case with `show`).

adler-j:
Is there a reason we don't do this using classical garbage collection instead? E.g. just implement `CallbackShowConvergence.__del__`.

Another, perhaps more pythonic, way would be to make `CallbackShowConvergence` a context manager.

kohr-h:
> What about renaming it to did_finish_optimization instead? If the callback has a key that says saveto I guess the user will expect it to save without having to manually call an additional method by hand (as is the case with show).

This whole things is a bit odd, to be honest. A callback is just a thing that is called in each iteration, and the user can decide to do things with it. It is not designed to know when things are "done", and the optimizer doesn't have that information either. Only the user knows when the right time has come to do the final thing, whatever that is.

Furthermore, the `saveto` thing is unspecific -- it doesn't convey *when* to save, or *what* (which boils down to the same). The context isn't clear, and the callback doesn't have that context, so the whole construction (at least when automated) becomes odd.

Actually, the whole code in `final()` is just one line, why not just put that as a hint in the documentation? Like
```
...
The figure object is available as `callback.fig` and can be used, e.g.,
to store the figure::

    callback.fig.savefig(path, **kwargs)

See `the matplotlib documentation
<https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html>`_
for options to ``savefig``.
"""

adler-j:
I agree with @kohr-h here. While useful, a sweeping change like this (touching upon all current and future solvers) needs to have a stronger motivation than saving one line. If this becomes a recurring problem we might consider adding this feature in the future.

Meanwhile, updating the documentation as suggested would be great.

PR #1371, opened 2018-06-15
BUG: Import scipy.optimize instead of scipy.

sbanert:
Probably, this has to be fixed in core ODL, too; particularly in the examples.

### DIFF
--- a/odl/contrib/param_opt/param_opt.py
+++ b/odl/contrib/param_opt/param_opt.py
@@ -9,7 +9,7 @@
 """Method(s) to find optimal reconstruction parameter(s) w.r.t. given FOM."""
 
 import numpy as np
-import scipy
+import scipy.optimize
 
 __all__ = ('optimal_parameters', )
 


sbanert:
No, not yet. Can @aringh, who has a conda environment with an old scipy version, check this?

aringh:
It does not break it on my machine, which have the following packages:
```
numpy                     1.12.1             py36_nomkl_0  [nomkl]
pytest                    3.0.5                    py36_0  
pytest-cache              1.0                      py36_0  
pytest-pep8               1.0.6                    py36_0
scipy                     0.19.0          np112py36_nomkl_0  [nomkl]
```

aringh:
I tried to have a look at release notes of scipy, but I find nothing about this change... maybe I missed it or I am not looking at the correct place.

Anyway, since it does not seem to break anything with `scipy 0.19.0`, which was released in March last year, I guess it is find. @kohr-h ?

sbanert:
I only found this commit, which dates back to 2013 and is tagged with `v1.1.0`.
https://github.com/scipy/scipy/commit/05b26a51a1c83a38e2098caa76a6ba5b9b14122b

adler-j:
So lets merge this then?

PR #1375, opened 2018-06-16
BUG: make DiscreteLpElement.show use correct interpolation, closes #1336

### LINKED ISSUE #1336: Interpolation not working or not showing
Not sure this is a bug or the desired behaviour, but the following example gives the same (visual) output in both cases. Could you comment on this?

```
import odl, numpy as np

space = odl.uniform_discr(-1, 1, 31, interp='nearest')
x = space.element(lambda x: np.exp(-x**2/2/0.25**2))
x.show()

space = odl.uniform_discr(-1, 1, 31, interp='linear')
x = space.element(lambda x: np.exp(-x**2/2/0.25**2))
x.show()
```

aringh:
Just implementing @kohr-h comment, this seems to solve the issue.

### DIFF
--- a/odl/discr/lp_discr.py
+++ b/odl/discr/lp_discr.py
@@ -1269,6 +1269,10 @@ class DiscreteLpElement(DiscretizedSpaceElement):
         """
         from odl.util.graphics import show_discrete_data
 
+        if 'interp' not in kwargs and not isinstance(self.space.interp, tuple):
+            # Otherwise different interp per axis
+            kwargs['interp'] = self.space.interp
+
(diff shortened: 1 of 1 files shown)


adler-j:
Interp can be a tuple in some cases where we have different interpolations per axis. We support that, but most graphing software does not. You can actually use "`interp_byaxis`" to make sure you always get the tuple. Otherwise, the results do look good, just gotta fix that small safeguard in case users want to override

mehrhardt:
@aringh this looks really good. Well done!

aringh:
Good to go?

PR #1376, opened 2018-06-16
BUG/TST: skip tests involving astra cuda if not available, closes #684

### LINKED ISSUE #684: ASTRA CUDA tests not skipped if only ASTRA CPU is installed

aringh:
Somewhat crude fix of the problem, that assumes that any installation of ASTRA always contains the CPU version.

I do not have an environment where I only have ASTRA CPU and not GPU. Could, e.g., @kohr-h check if it works as intended?

### DIFF
--- a/odl/test/tomo/backends/astra_setup_test.py
+++ b/odl/test/tomo/backends/astra_setup_test.py
@@ -292,9 +292,6 @@ def test_parallel_2d_projector():
 def test_parallel_3d_projector():
     """Create ASTRA 3D projectors."""
 
-    odl.tomo.astra_projector('nearest', VOL_GEOM_3D, PROJ_GEOM_3D, ndim=3,
-                             impl='cuda')
-
     # Run as a real test once ASTRA supports this construction
     with pytest.raises(ValueError):
         odl.tomo.astra_projector('nearest', VOL_GEOM_3D, PROJ_GEOM_3D,
@@ -305,6 +302,15 @@ def test_parallel_3d_projector():
                                  ndim=3, impl='cpu')
 
 
+@pytest.mark.skipif(not odl.tomo.ASTRA_CUDA_AVAILABLE,
+                    reason="ASTRA cuda not available")
+def test_parallel_3d_projector_gpu():
+    """Create ASTRA 3D projectors on GPU."""
+
+    odl.tomo.astra_projector('nearest', VOL_GEOM_3D, PROJ_GEOM_3D, ndim=3,
+                             impl='cuda')
+
+
 def test_astra_algorithm():
     """Create ASTRA algorithm object."""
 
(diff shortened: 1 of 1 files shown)


aringh:
The GPU version of ASTRA does not work, but the flag `odl.tomo.ASTRA_CUDA_AVAILABLE` does not seem to change to false, which thus cases the tests to crash.
```
$ CUDA_VISIBLE_DEVICES="" python
Python 3.6.1 |Continuum Analytics, Inc.| (default, May 11 2017, 13:09:58) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import odl
>>> odl.tomo.ASTRA_CUDA_AVAILABLE
True
```

adler-j:
What astra version is this? They should have fixed it

aringh:
1.8. I will try to upgrade

aringh:
I get the same behavior after upgrading to ASTRA 1.8.3

adler-j:
Thats weird, you should make a bug report with ASTRA then.

kohr-h:
Do we use the correct ASTRA function to check for GPU?

adler-j:
We use `astra.astra.use_cuda() `, that should be fine, right?

https://github.com/odlgroup/odl/blob/ac193c04d1bd50327b20de8f8ecbf303a32e5bcc/odl/tomo/backends/astra_cuda.py#L16

kohr-h:
That's fine.

However, the runtime check hasn't been released yet. The last release was [on Nov 6, 2017](https://github.com/astra-toolbox/astra-toolbox/releases), and the runtime check was added [3 days later](https://github.com/astra-toolbox/astra-toolbox/commit/0ad6cb01b9176e99c3e4bb5643ac6a5dd477aad6). Bad luck.

adler-j:
How about we still merge this? Then it's fixed for users who compile astra themselves and conda users in the future.

odl/test/tomo/backends/astra_setup_test.py
```diff
@@ -340,6 +346,16 @@ def test_astra_algorithm():
                                           proj_id, impl)
         astra.algorithm.delete(alg_id)
 
+
+def test_astra_algorithm_gpu():
+    """Create ASTRA algorithm object on GPU."""
+
```
aringh: The whole point of this PR was to safe-guard for non-cuda users. But apparently I forgot to add
```python
if not odl.tomo.ASTRA_CUDA_AVAILABLE:
    pytest.skip("ASTRA cuda not available")
```
at this point...

aringh: Now it is updated on the line below

aringh:
I am not sure that anyone have done a proper review of the code yet :sweat_smile:. But if the code looks fine, I am not against merging

aringh:
Updated accordingly and rebased on master

odl/test/tomo/backends/astra_setup_test.py
```diff
@@ -305,14 +302,25 @@ def test_parallel_3d_projector():
                                  ndim=3, impl='cpu')
 
 
+@pytest.mark.skipif(not odl.tomo.ASTRA_CUDA_AVAILABLE,
+                    reason="ASTRA cuda not available")
+def test_parallel_3d_projector_gpu():
+    """Create ASTRA 3D projectors on GPU."""
+
+    if not odl.tomo.ASTRA_CUDA_AVAILABLE:
```
adler-j: Leftover

PR #1377, opened 2018-06-17
Update copyright year to 2018

aringh:
From the root folder of `odl`, this was done with 
```
$ find . -type f -print0 | xargs -0 sed -i '' -e 's/Copyright 2014-2017/Copyright 2014-2018/g'
```
see, e.g., [this stack exchange question](https://superuser.com/questions/428493/how-can-i-do-a-recursive-find-and-replace-from-the-command-line). To check that nothing has been missed, I also ran
```
grep -rnw . -e "2017"
```
see, e.g., [this stack exchange question](https://stackoverflow.com/questions/16956810/how-do-i-find-all-files-containing-specific-text-on-linux).

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
 #
--- a/doc/source/generate_doc.py
+++ b/doc/source/generate_doc.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
 #
--- a/odl/__init__.py
+++ b/odl/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
 #
--- a/odl/contrib/__init__.py
+++ b/odl/contrib/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
 #
--- a/odl/contrib/datasets/__init__.py
+++ b/odl/contrib/datasets/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
 #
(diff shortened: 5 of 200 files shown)


aringh:
Running the second command for 2016, 2015, and 2014, I found the following
```bash
./examples/solvers/rosenbrock_minimization.py:1:# Copyright 2014-2016 The ODL development group
./doc/source/conf.py:117:copyright = u'2015-2017, ODL development group'
```
1) I thought copyright was removed from examples, or? Remove it from the Rosenbrock example?
2) I guess the second one should also be updated to 2018, or? However, all other copyright in the files say `2014-[...]`. Should this one also do that?

aringh:
True. Input from @ozanoktem or @kohr-h?

If we do not update the year in each file, should we not remove the years from the copyright statement of the files? Write something either like `Copyright The ODL contributors` or `Copyright from 2014 The ODL contributors`? The we could just update the line
```
./doc/source/conf.py:117:copyright = u'2015-2017, ODL development group'
```
ones a year.

kohr-h:
Generally, it seems like a year establishes copyright from that date on, with all the stuff like expiry attached. If you write a single year, that's where the clock starts. It may even be that only the original version published at that date is covered, but here my error bars are quite large.

If you add further years, the clock starts anew with the revisions from those years. I also assume that texts like "from 2014" with built-in auto-renewal are legally invalid (because otherwise everybody would do it).

All in all, I tend to agree with @adler-j's statement that renewal on demand is better in view of maintaining a meaningful git history. Also from a legal point of view, if a file has been updated last in 2017 and has copyright till 2017 in its header, that's okay since it's the last revision of that file.

Random sources I found by googling:
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated
http://smallbusiness.chron.com/year-copyright-statement-mean-62027.html

aringh:
So, there seems to be no general consensus online... [Here](https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/) they advocate "Year first published - last year changed", while [here](https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year) they argue that "Year first published" is enough. The only consensus seems to be that the notice itself is actually not needed for the material to be copyrighted, but it is good practice to have a statement.

So: lets not make "across the board" updates of the statement. But just as we have guidelines for code style, how to write documentation, etc., maybe we should have a policy for how to deal with copyright statements?

kohr-h:
When starting out, I went with [these instructions](https://www.gnu.org/licenses/gpl-howto.html) which I believe to be credible and legally sound (Gnu actually did all the lawyer stuff). Of course, I have no way of checking this, but whenever a random individual on the internet disagrees, I assume that the individual is wrong.

@aringh agreed, and good idea to write it down somewhere.

Issue #1311, opened 2018-03-11
Make TODO comments more specific


kohr-h:
During my repr improvement work, I've also gone though a bunch of TODO comments that just made me scratch my head. Stuff like `TODO: update docs`... okay, how? (That one was by me btw).

So I'd propose that we be more specific when writing comments like this, and make such comments easier to search. Here's my suggestion:

- Write `# TODO(<username>): ...` for generic comments on implementation details, suboptimal code, reasons for not implementing things etc. This way it's at least clear whom to ask what the comment was about.
- Write `# TODO(#<issue number>): ...` for TODOs that have an issue attached to them, for instance "do this when something gets available/fixed".
- Update  `# TODO(#<issue number>): ...` -> `# TODO(doable, #<issue number>): ...` when the issue has been closed in the meanwhile. This update should be done for every issue that gets closed, a simple text search brings up the candidates (see above).

PR #1380, opened 2018-06-19
ENH: add utility functions for repr printing

kohr-h:
This is a break-out from the monster PR #1276. It only adds the tools for better `repr` printing, thereby slightly changing how stuff works behind the scenes.
Follow-up PRs will apply this to the classes in ODL and fix some doc issues.

I've also started to implement #1311 here (more specific TODO comments).


kohr-h:
Failure is due to Numpy 1.14.0 being installed, needs fix

odl/util/utility.py
```diff
@@ -72,12 +78,80 @@ def indent(string, indent_str='    '):
...
+    <->This is line 1.
+    <->Next line.
+    <->And another one.
     """
-    return '\n'.join(indent_str + row for row in string.split('\n'))
+    return '\n'.join(indent_str + row for row in string.splitlines())
+
+
+def dedent(string, indent_str='   ', max_levels=None):
+    """Revert the effect of indentation.
```
adler-j: `IndentationOperator.inverse`!

(no, I'm not serious).

odl/util/utility.py
```diff
@@ -72,12 +78,80 @@ def indent(string, indent_str='    '):
...
+    <->Level 2.
+    <-><->Level 3.
+    """
+    lines = string.splitlines()
+
+    # Determine common (minumum) number of indentation levels, capped at
+    # `max_levels` if given
+    def num_indents(line):
+        nindents = 0
+        while True:
```
adler-j: This is a scary risk for an infinite loop. Perhaps some safety?

odl/util/utility.py
```diff
@@ -8,25 +8,31 @@
...
+__all__ = (
+    'array_str', 'dtype_str', 'dtype_repr', 'npy_printoptions',
+    'signature_string', 'signature_string_parts', 'repr_string',
+    'indent', 'dedent', 'attribute_repr_string', 'method_repr_string',
+    'is_numeric_dtype', 'is_int_dtype', 'is_floating_dtype', 'is_real_dtype',
+    'is_real_floating_dtype', 'is_complex_floating_dtype',
+    'real_dtype', 'complex_dtype', 'is_string', 'nd_iterator', 'conj_exponent',
+    'writable_array', 'run_from_ipython', 'NumpyRandomSeed',
+    'cache_arguments', 'unique',
+    'REPR_PRECISION')
```
adler-j: This file is getting out of hand. Perhaps we should split it (but leave that for another PR).

odl/util/utility.py
```diff
@@ -72,12 +78,80 @@ def indent(string, indent_str='    '):
...
+    >>> print(text)
+    <->This is line 1.
+    <->Next line.
+    <->And another one.
+    >>> print(dedent(text, '<->'))
+    This is line 1.
+    Next line.
+    And another one.
+
+    Multiple levels of indentation:
```
adler-j: IMO this is too many tests, hard to read. At least add some spacing

odl/util/utility.py
```diff
@@ -833,15 +995,436 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
...
+        # Next string might fit on same line
+        separators.append(', ')
+        cur_line_len += 1  # for the extra space
+    else:
+        # Use linebreak if string contains newline or doesn't fit
+        separators.append(',\n')
+        cur_line_len = indent_len
+
+    for i, s in enumerate(strings[1:-1]):
+
```
adler-j: Remove blankline

odl/util/utility.py
```diff
@@ -833,15 +995,436 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
...
+    return tuple(separators)
+
+
+def repr_string(outer_string, inner_strings, allow_mixed_seps=True):
+    r"""Return a pretty string for ``repr``.
+
+    The returned string is formatted such that it does not extend
+    beyond the line boundary if avoidable. The line width is taken from
+    NumPy's printing options that can be retrieved with
+    ``np.get_printoptions()``. They can be temporarily overridden
```
adler-j: numpy.get_printoptions creates a clickable link

odl/util/utility.py
```diff
@@ -833,15 +995,436 @@ def signature_string(posargs, optargs, sep=', ', mod=''):
...
+    using the `npy_printoptions` context manager. See Examples for details.
```
adler-j: This does not create a link

odl/util/utility.py
kohr-h: :scream:

kohr-h:
> But I feel that the code is somewhat abusing doctests to do full scale testing. I'd prefer doctests to stay small and perhaps move something to a real external test.

Well, here the intention is to document the behavior of the function as precisely as possible, and for that, unit tests are not a good vehicle. Doctests are much better suited for that. The functions here are not really user-facing, so I took the approach of maximizing documentation for developers.

odl/util/utility.py
kohr-h: It should, since it's one of our functions in the same module

odl/util/utility.py
kohr-h: ok

odl/util/utility.py
kohr-h: See #1386

kohr-h:
Comments addressed, rebased. Good to go?

kohr-h:
Bump. Good to go in?

PR #1383, opened 2018-06-20
BUG: Fix minor errors in mayo geometry definitions

adler-j:
It seems we had some minor (but notable in some cases) errors with the geometry descriptions.

### DIFF
--- a/odl/contrib/datasets/ct/mayo.py
+++ b/odl/contrib/datasets/ct/mayo.py
@@ -107,10 +107,7 @@ def load_projections(folder, indices=None):
     angles = [d.DetectorFocalCenterAngularPosition for d in datasets]
     angles = -np.unwrap(angles) - np.pi  # different defintion of angles
 
-    # Make a parallel beam geometry with flat detector
-    angle_partition = odl.nonuniform_partition(angles)
-
-    # Set minimum and maximum point
+    # Set minimum and maximum corners
     shape = np.array([datasets[0].NumberofDetectorColumns,
                       datasets[0].NumberofDetectorRows])
     pixel_size = np.array([datasets[0].DetectorElementTransverseSpacing,
@@ -126,30 +123,39 @@ def load_projections(folder, indices=None):
     src_radius = datasets[0].DetectorFocalCenterRadialDistance
     det_radius = (datasets[0].ConstantRadialDistance -
                   datasets[0].DetectorFocalCenterRadialDistance)
-
-    # Convert pitch and offset to odl defintions
     pitch = (pixel_size[1] * shape[1] * datasets[0].SpiralPitchFactor *
              src_radius / (src_radius + det_radius))
-    offset_along_axis = (datasets[0].DetectorFocalCenterAxialPosition -
-                         angles[0] / (2 * np.pi) * pitch)
 
     # Get flying focal spot data
     offset_axial = np.array([d.SourceAxialPositionShift for d in datasets])
     offset_angular = np.array([d.SourceAngularPositionShift for d in datasets])
     offset_radial = np.array([d.SourceRadialDistanceShift for d in datasets])
 
-    angles_offset = angles - offset_angular
-    src_rad_offset = src_radius + offset_radial
-    offset_x = (np.cos(angles_offset) * (-src_rad_offset) -
-                np.cos(angles) * (-src_radius))
-    offset_y = (np.sin(angles_offset) * (-src_rad_offset) -
-                np.sin(angles) * (-src_radius))
-    offset_z = offset_axial
+    if 1:
+        # Apply only mean of offsets
+        src_radius = src_radius + np.mean(offset_radial)
+        angles = angles - np.mean(offset_angular)
+        offset_along_axis = np.mean(offset_axial) * (
+            src_radius / (src_radius + det_radius))
+    else:
+        # TODO: Implement proper handling of flying focal spot
+        angles_offset = angles - offset_angular
+        src_rad_offset = src_radius + offset_radial
+        offset_x = (np.cos(angles_offset) * (-src_rad_offset) -
+                    np.cos(angles) * (-src_radius))
+        offset_y = (np.sin(angles_offset) * (-src_rad_offset) -
+                    np.sin(angles) * (-src_radius))
+        offset_z = offset_axial
 
-    # TODO: WE CURRENTLY IGNORE THE OFFSETS DUE TO FLYING FOCAL SPOT
-    source_offsets = np.array([offset_x, offset_y, offset_z]).T
+        source_offsets = np.array([offset_x, offset_y, offset_z]).T
+
+    # Convert offset to odl defintions
+    offset_along_axis = (offset_along_axis +
+                         datasets[0].DetectorFocalCenterAxialPosition -
+                         angles[0] / (2 * np.pi) * pitch)
 
     # Assemble geometry
+    angle_partition = odl.nonuniform_partition(angles)
     geometry = odl.tomo.ConeFlatGeometry(angle_partition,
                                          detector_partition,
                                          src_radius=src_radius,
@@ -245,8 +251,11 @@ def load_reconstruction(folder, slice_start=0, slice_end=-1):
     shape = np.array([rows, cols, len(volumes)])
     min_pt = (np.array(dataset.ImagePositionPatient) -
               np.array(dataset.DataCollectionCenterPatient))
-    min_pt[:2] += 0.5 * np.array(dataset.PixelSpacing)
-    max_pt = min_pt + voxel_size * np.array([rows - 1, cols - 1, 0])
+
+    max_pt = min_pt + voxel_size * np.array([rows, cols, 0])
+
+    # axis 1 has reversed convention
+    min_pt[1], max_pt[1] = -max_pt[1], -min_pt[1]
 
     min_pt[2] = -np.array(datasets[0].DataCollectionCenterPatient)[2]
     min_pt[2] -= 0.5 * pixel_thickness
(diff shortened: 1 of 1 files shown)


odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,28 +123,41 @@ def load_projections(folder, indices=None):
...
+        source_offsets = np.array([offset_x, offset_y, offset_z]).T
+
+    # Convert offset to odl defintions
+    offset_along_axis = (offset_along_axis +
+                         datasets[0].DetectorFocalCenterAxialPosition -
+                         angles[0] / (2 * np.pi) * pitch)
 
-    # TODO: WE CURRENTLY IGNORE THE OFFSETS DUE TO FLYING FOCAL SPOT
-    source_offsets = np.array([offset_x, offset_y, offset_z]).T
+    # Make a parallel beam geometry with flat detector
```
adler-j: Wrong comment

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,28 +123,41 @@ def load_projections(folder, indices=None):
...
-                np.cos(angles) * (-src_radius))
-    offset_y = (np.sin(angles_offset) * (-src_rad_offset) -
-                np.sin(angles) * (-src_radius))
-    offset_z = offset_axial
+    if 1:
+        # Apply only mean of offsets
+        src_radius = src_radius + np.mean(offset_radial)
+        angles = angles - np.mean(offset_angular)
+
+        # print(np.mean(offset_angular))
```
adler-j: Remove print

kohr-h:
I guess this isn't quite ready for review yet

adler-j:
Its reviewable, just some minor stuff to fix.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,30 +123,39 @@ def load_projections(folder, indices=None):
...
-    src_rad_offset = src_radius + offset_radial
-    offset_x = (np.cos(angles_offset) * (-src_rad_offset) -
-                np.cos(angles) * (-src_radius))
-    offset_y = (np.sin(angles_offset) * (-src_rad_offset) -
-                np.sin(angles) * (-src_radius))
-    offset_z = offset_axial
-
-    # TODO: WE CURRENTLY IGNORE THE OFFSETS DUE TO FLYING FOCAL SPOT
-    source_offsets = np.array([offset_x, offset_y, offset_z]).T
+    if 1:
```
kohr-h: I guess this is for development only?
If not, I'm not in favor of "parking" backup code like this. Better make an issue and post the old code there for reference. It's also in the git history anyway.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,30 +123,39 @@ def load_projections(folder, indices=None):
...
+        # Apply only mean of offsets
+        src_radius = src_radius + np.mean(offset_radial)
```
kohr-h: I don't get this. How could this ever work before if it's a vector? In the old version, `src_rad_offset` must have been a vector, and with it all 3 offsets as well.

adler-j: Basically the offsets are _weird_, so I decided to just go for their means. Before we did not do anything about it (src_rad_offset  was unused) which gave some slight errors, since apparently the "offsets" are not zero-centered.

odl/contrib/datasets/ct/mayo.py
adler-j: I'll remove it.

odl/contrib/datasets/ct/mayo.py
kohr-h: Okay. What I was puzzled about was that in the old version, `src_radius` became a vector (after adding `offset_radial`), then it was used to compute `offset_[xyz]`, each of which also was a vector then, and the final `source_offsets` stacked them to a matrix. I don't see how that made sense.

Anyhow, for some other PR.

adler-j:
With some further small fixes, it now actually looks like this does the right thing, and there are no magic numbers!

A note to future readers (e.g myself in a year): The tags "SpiralPitchFactor" and "TableFeedPerRotation" are slightly bugged and not to be trusted.

adler-j:
Could use a final review on this

odl/contrib/datasets/ct/mayo.py
```diff
@@ -233,25 +236,41 @@ def load_reconstruction(folder, slice_start=0, slice_end=-1):
         data_array = data_array.reshape([cols, rows], order='C')
         data_array = np.rot90(data_array, -1)
 
-        # Convert from CT numbers to densities
-        data_array /= 1024.0
+        # Convert from storage type to densities
+        # TODO: Optimize these computations
+        hu_values = (dataset.RescaleSlope * data_array +
```
aringh: Now I will reveal my lack of knowledge in this field: what is the difference between this `hu_values` and the `hu_factor` on line 57?

odl/contrib/datasets/ct/mayo.py
```diff
@@ -233,25 +236,41 @@ def load_reconstruction(folder, slice_start=0, slice_end=-1):
...
 
+    max_pt = min_pt + voxel_size * np.array([rows, cols, 0])
+
+    # axis 1 has reversed convention
+    min_pt[1], max_pt[1] = -max_pt[1], -min_pt[1]
+
+    if len(datasets) > 1:
+        slice_distance = np.abs(
+            np.array(datasets[1].DataCollectionCenterPatient)[2] -
+            np.array(datasets[0].DataCollectionCenterPatient)[2])
```
kohr-h: `np.array` shouldn't be necessary?

odl/contrib/datasets/ct/mayo.py
```diff
@@ -233,25 +236,41 @@ def load_reconstruction(folder, slice_start=0, slice_end=-1):
...
+        slice_distance = pixel_thickness
+
+    # Set min and max points. We need half voxel offsets since dicom uses
+    # midpoint and ODL uses the corners.
     min_pt[2] = -np.array(datasets[0].DataCollectionCenterPatient)[2]
-    min_pt[2] -= 0.5 * pixel_thickness
+    min_pt[2] -= 0.5 * slice_distance
     max_pt[2] = -np.array(datasets[-1].DataCollectionCenterPatient)[2]
-    max_pt[2] += 0.5 * pixel_thickness
+    max_pt[2] += 0.5 * slice_distance
```
kohr-h: What's the overall logic of computing the min/max in the z axis? Does it go from `-(patient z in dataset 0)` to 0, adjusted by half a slice distance in each direction?

odl/contrib/datasets/ct/mayo.py
```diff
@@ -127,29 +125,34 @@ def load_projections(folder, indices=None):
...
+    # Correct the angular sampling according to corrections
+    angles = angles - offset_angular
 
-    # TODO: WE CURRENTLY IGNORE THE OFFSETS DUE TO FLYING FOCAL SPOT
-    source_offsets = np.array([offset_x, offset_y, offset_z]).T
+    # TODO: Implement proper handling of flying focal spot
+    # Apply only mean of offsets
+    src_radius = src_radius + np.mean(offset_radial)
+    offset_along_axis = np.mean(offset_axial) * (
+        src_radius / (src_radius + det_radius))
```
kohr-h: Comment on why the multiplication with the magnification factor? (Which BTW would be a good name for a temp variable, to clarify things a bit.)

odl/contrib/datasets/ct/mayo.py
```diff
@@ -127,29 +125,34 @@ def load_projections(folder, indices=None):
...
     offset_radial = np.array([d.SourceRadialDistanceShift for d in datasets])
 
-    angles_offset = angles - offset_angular
-    src_rad_offset = src_radius + offset_radial
-    offset_x = (np.cos(angles_offset) * (-src_rad_offset) -
-                np.cos(angles) * (-src_radius))
-    offset_y = (np.sin(angles_offset) * (-src_rad_offset) -
-                np.sin(angles) * (-src_radius))
-    offset_z = offset_axial
+    # Correct the angular sampling according to corrections
```
kohr-h: This doesn't really say anything

odl/contrib/datasets/ct/mayo.py
```diff
@@ -127,29 +125,34 @@ def load_projections(folder, indices=None):
...
                   datasets[0].DetectorFocalCenterRadialDistance)
 
-    # Convert pitch and offset to odl defintions
-    pitch = (pixel_size[1] * shape[1] * datasets[0].SpiralPitchFactor *
-             src_radius / (src_radius + det_radius))
-    offset_along_axis = (datasets[0].DetectorFocalCenterAxialPosition -
-                         angles[0] / (2 * np.pi) * pitch)
+    # For unknown reasons, mayo does not include the tag
+    # "TableFeedPerRotation", which is what we want.
+    # Instead we manually compute the pitch
```
kohr-h: This is a great comment!

odl/contrib/datasets/ct/mayo.py
```diff
@@ -127,29 +125,34 @@ def load_projections(folder, indices=None):
...
-                np.cos(angles) * (-src_radius))
-    offset_y = (np.sin(angles_offset) * (-src_rad_offset) -
-                np.sin(angles) * (-src_radius))
-    offset_z = offset_axial
+    # Correct the angular sampling according to corrections
+    angles = angles - offset_angular
 
-    # TODO: WE CURRENTLY IGNORE THE OFFSETS DUE TO FLYING FOCAL SPOT
-    source_offsets = np.array([offset_x, offset_y, offset_z]).T
+    # TODO: Implement proper handling of flying focal spot
```
kohr-h: How about you write `# TODO(adler-j)`? Then people looking at this know whom to ask without doing a `git blame`.

odl/contrib/datasets/ct/mayo.py
kohr-h: Okay, maybe a comment could make that more obvious, I had to think about it for a second. Maybe also the review markup was a bit to blame.

adler-j:
Fixed comments, merge after CI.

PR #1387, opened 2018-06-24
Tests indicator sum and simplex

aringh:
Add `IndicatorSumConstraint` and `IndicatorSimplex` to general test suite for functionals.

Should we also add test separate in `default_functionals_test.py`?

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -121,8 +121,8 @@ class LpNorm(Functional):
         elif self.exponent == np.inf:
             return proximal_linfty(space=self.domain)
         else:
-            raise NotImplementedError('`proximal` only implemented for p=1 or '
-                                      'p=2')
+            raise NotImplementedError('`proximal` only implemented for p=1, '
+                                      'p=2, and p=inf')
 
     @property
     def gradient(self):
@@ -177,8 +177,8 @@ class LpNorm(Functional):
             return L2Gradient()
 
         else:
-            raise NotImplementedError('`gradient` only implemented for p=1 or '
-                                      'p=2')
+            raise NotImplementedError('`gradient` only implemented for p=1 '
+                                      'and p=2')
 
     def __repr__(self):
         """Return ``repr(self)``."""
@@ -2230,8 +2230,10 @@ class IndicatorSimplex(Functional):
             Domain of the functional.
         diameter : positive float, optional
             Diameter of the simplex.
-        sum_rtol : float
-            Relative tolerance for sum comparison.
+        sum_rtol : float, optional
+            Relative tolerance for sum comparison. If set to None, the default
+            is ``space.size`` times ``1e-10`` when ``space.dtype`` is
+            ``float64`` and ``1e-6`` otherwise.
 
         Examples
         --------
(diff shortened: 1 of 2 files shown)


odl/solvers/functional/default_functionals.py
```diff
@@ -2317,7 +2319,7 @@ def __repr__(self):
 
 class IndicatorSumConstraint(Functional):
 
-    """The indicator functional of a unit sum constraint.
+    r"""The indicator functional of a unit sum constraint.
```
adler-j: I think we typically try to avoid "The ..."

odl/solvers/functional/default_functionals.py
```diff
@@ -2341,8 +2343,10 @@ def __init__(self, space, sum_value=1, sum_rtol=None):
             Domain of the functional.
         sum_value : float
             Desired value of the sum constraint.
-        sum_rtol : float
-            Relative tolerance for sum comparison.
+        sum_rtol : float, optional
+            Relative tolerance for sum comparison. If set to None, the default
+            is ``1e-10`` when ``space.dtype`` is ``float64``, and ``1e-6``
```
adler-j: This should probably be scaled by the number of elements in the space

odl/test/solvers/functional/functional_test.py
```diff
@@ -131,7 +137,9 @@ def test_derivative(functional):
     the inner product of the gradient and the direction, if the gradient is
     defined.
     """
-    if isinstance(functional, odl.solvers.functional.IndicatorLpUnitBall):
+    if isinstance(functional, (odl.solvers.functional.IndicatorLpUnitBall,
```
adler-j: Create a global, e.g. "FUNCTIONALS_WITHOUT_DERIVATIVE" and use it instead of repeating yourself. That also fixes the comment in the line below that is now lacking

aringh:
Let's try to get this in :smiley:. Good to go?

odl/solvers/functional/default_functionals.py
```diff
@@ -2226,8 +2226,10 @@ def __init__(self, space, diameter=1, sum_rtol=None):
             Domain of the functional.
         diameter : positive float, optional
             Diameter `r` of simplex.
-        sum_rtol : float
-            Relative tolerance for sum comparison.
+        sum_rtol : float, optional
+            Relative tolerance for sum comparison. If set to None, the default
+            is ``space.size`` times ``1e-10`` when ``space.dtype`` is
+            ``float64`` and ``1e-6`` otherwise.
```
kohr-h: This becomes almost a bit too much backticked stuff mixed with regular English. I think this is a bit easier to comprehend:
```
Default:

- ``space.dtype == 'float64'``: ``1e-10 * space.size``
- Otherwise: ``1e-6 * space.size``
```

aringh:
Regarding the pep8 issue: this PR does not even touch these two lines of code. This was introduced in [this commit](https://github.com/odlgroup/odl/commit/b2e1152cb180e7d2281d3364c04d1502ba394f25#diff-50e0484cf75f430f06878917e0952f87), which was part of [this PR](https://github.com/odlgroup/odl/pull/1426), where it was argued that it is not a pep8 violation.

kohr-h:
> Regarding the pep8 issue: this PR does not even touch these two lines of code. This was introduced in this commit, which was part of this PR, where it was argued that it is not a pep8 violation.

Ignore it. Since it's in a docstring, it's not really a violation, and for long links I think it's better to go over the line limit than to create an ugly break with a backslash (which doesn't work with raw strings anyway).

PR #1389, opened 2018-06-24
MAINT: replace places with ndigits and remove almost_equal

kohr-h:
Boring PR, some renaming and replacement of `almost_equal` with `pytest.approx`.


odl/test/discr/lp_discr_test.py
```diff
@@ -1267,17 +1267,14 @@ def test_norm_rectangle(exponent):
...
+        assert discr_testfunc.norm() == pytest.approx(true_norm, rel=1e-2)
 
 
 def test_norm_rectangle_boundary(odl_tspace_impl, exponent):
     # Check the constant function 1 in different situations regarding the
     # placement of the outermost grid points.
     impl = odl_tspace_impl
 
-    if exponent == float('inf'):
-        pytest.xfail('inf-norm not implemented in CUDA')
```
aringh: Just curious, why did we have `xfail` on this one before and what has changed now?

odl/test/solvers/nonsmooth/douglas_rachford_test.py
```diff
@@ -92,7 +92,7 @@ def test_primal_dual_l1():
     x = space.zero()
     douglas_rachford_pd(x, f, g, L, tau=3.0, sigma=[1.0], niter=10)
 
-    assert all_almost_equal(x, data_1, places=2)
+    assert all_almost_equal(x, data_1, ndigits=2)
```
aringh: Why do we have `HIGH_ACCURACY` and  `LOW_ACCURACY` defined above, and anyway hard-code `ndigits` here?

odl/util/testutils.py
```diff
@@ -64,32 +66,14 @@ def dtype_places(dtype, default=None):
...
-    except TypeError:
-        return False
-
-    if np.isnan(a) and np.isnan(b):
-        return True
-
-    if np.isinf(a) and np.isinf(b):
-        return a == b
+def dtype_rtol(dtype, default=None):
+    """Return a relative tolerance for a given dtype.
```
aringh: No parameter list for this function?

odl/test/discr/lp_discr_test.py
kohr-h: The inf norm wasn't implemented on CUDA, and I guess we weren't (i.e. I wasn't) smart enough to xfail only in that case. Now we have no CUDA backend, and the new one will support it, so this restriction doesn't make any more sense.

odl/test/solvers/nonsmooth/douglas_rachford_test.py
kohr-h: Dunno, I just took the code as-is. Probably copy & paste.

odl/util/testutils.py
kohr-h: Hm yeah, I guess I was just too lazy :-)

odl/test/solvers/nonsmooth/douglas_rachford_test.py
kohr-h: I think here the number of expected correct digits depends on the problem and on the (few) iterations we run. I'll keep it like this.

odl/util/testutils.py
kohr-h: I removed the `r` from the function name since the tolerance can also be used as absolute. It also has a bit of doc, not the full `Parameters` section though. Should be good enough for an internal util.

kohr-h:
I'll merge when (if) CI is happy.

adler-j:
Honest question: How hard would it be to scrap all of these and simply have one large monolithic `approx(x, y)` that handles all corner cases and has a unified syntax?

kohr-h:
Topic for a new issue. Perhaps we can just subclass the pytest Approx class (or however it's named) and feed our default `rel` and `abs` parameters as we implement them now.

PR #1390, opened 2018-06-24
Small clean-up of TODOs

aringh:
Since the two commits are for different TODOs, they are probably easiest to review one at a time.


odl/test/solvers/smooth/smooth_test.py
```diff
@@ -34,6 +32,17 @@ def functional(request):
         scaling = odl.MultiplyOperator(space.element([1, 2, 3]),
                                        domain=space)
         return odl.solvers.L2NormSquared(space) * scaling
+    elif name == 'quadratic_form':
+        space = odl.rn(3)
+        # Symmetric and diagonally dominante matrix
```
kohr-h: dominant

odl/test/solvers/smooth/smooth_test.py
```diff
@@ -34,6 +32,17 @@ def functional(request):
...
+        matrix = odl.MatrixOperator([[7.0, 1, 2], [1, 5, -3], [2, -3, 8]])
```
kohr-h: At the cost of vertical space, matrices look nicer if the initializing nested lists are shown one row at a time.

odl/test/solvers/smooth/smooth_test.py
```diff
@@ -34,6 +32,17 @@ def functional(request):
...
+        vector = space.element([1.0, 2, 3])
```
kohr-h: Just 1

odl/test/solvers/smooth/smooth_test.py
```diff
@@ -34,6 +32,17 @@ def functional(request):
...
+
+        # Calibrate so that functional is zero in optimal point
+        constant = 1.0 / 4 * vector.inner(matrix.inverse(vector))
```
kohr-h: No need for the `.0` here either

odl/operator/default_ops.py
```diff
@@ -103,6 +104,25 @@ def inverse(self):
     def adjoint(self):
         """Adjoint, given as scaling with the conjugate of the scalar.
 
+        Examples
+        --------
+        In the real case, the adjoint is the same as the operator
```
kohr-h: `:` at the end

odl/operator/default_ops.py
```diff
@@ -103,6 +104,25 @@ def inverse(self):
...
+
+        >>> r3 = odl.rn(3)
+        >>> x = r3.element([1, 2, 3])
+        >>> op = ScalingOperator(r3, 2)
+        >>> op.adjoint(x) == op(x)
```
kohr-h: Don't rely on this behavior of `==`, it will change.

odl/operator/default_ops.py
```diff
@@ -103,6 +104,25 @@ def inverse(self):
...
+        True
+
+        In the complex case, the scalar is conjugated
```
kohr-h: `:`

odl/operator/default_ops.py
```diff
@@ -103,6 +104,25 @@ def inverse(self):
...
+
+        >>> c3 = odl.cn(3)
+        >>> x_complex = c3.element([1, 1j, 1-1j])
+        >>> op = ScalingOperator(c3, 1+1j)
+        >>> expected_op = ScalingOperator(c3, 1-1j)
+        >>> op.adjoint(x_complex) == expected_op(x_complex)
```
kohr-h: I'd just show the two side by side. This is more unit-test-like.

odl/operator/default_ops.py
```diff
@@ -340,11 +360,29 @@ def adjoint(self):
         >>> op2 = MultiplyOperator(3.0, domain=r3, range=r3)
         >>> op2.adjoint(x)
         rn(3).element([ 3.,  6.,  9.])
+
+        Multipy operator with complex space:
```
kohr-h: This reads as if you multiply an operator with a space. Maybe write "Multiplication operator"

odl/operator/default_ops.py
```diff
@@ -340,11 +360,29 @@ def adjoint(self):
...
-            return InnerProductOperator(self.multiplicand)
+            if isinstance(self.domain, RealNumbers):
+                return InnerProductOperator(self.multiplicand)
+            elif isinstance(self.domain, ComplexNumbers):
+                return InnerProductOperator(self.multiplicand.conjugate())
+            else:
+                raise TypeError(
+                    'adjoint only implemented for `RealNumbers` and '
+                    '`ComplexNumbers` when `domain` is a `Field`, got {}'
+                    ''.format(self.domain))
```
kohr-h: Too complicated. Just write `'adjoint not implemented for `domain` {!r}'.format(self.domain)`. Also it should be `NotImplementedError`, not `TypeError`.

odl/operator/default_ops.py
```diff
@@ -340,11 +360,29 @@ def adjoint(self):
...
+        elif is_complex_floating_dtype(self.domain):
```
kohr-h: This doesn't quite make sense. You probably meant `self.domain.dtype`, but simply do `self.domain.is_real` or `self.domain.is_complex`.

odl/operator/default_ops.py
aringh: How should I compare, `all_almost_equal`? But then it becomes more unit-test-like...

odl/operator/default_ops.py
aringh: How do I show them side by side? First one of them, with expected result underneath, and then the other?

I guess I could do the same just above.

odl/operator/default_ops.py
kohr-h: If it doesn't get too busy, just do
```python
>>> op(x)
(output)
>>> op.adjoint(x)  # the same
(output)
```

odl/operator/default_ops.py
kohr-h: Exactly. With "side" I meant "top/bottom" :wink:

aringh:
Updated accordingly and rebased on master. Good to go?

PR #1392, opened 2018-06-25
MAINT: fix link to Matthias' GH profile

kohr-h:


kohr-h:
Oh sorry, was too fast. Who would that be?

PR #1393, opened 2018-06-25
Minor updates in Pytorch interface

olivierverdier:
Mostly due to interface changes in Pytorch.

### DIFF
--- a/odl/contrib/torch/operator.py
+++ b/odl/contrib/torch/operator.py
@@ -102,7 +102,7 @@ class OperatorAsAutogradFunction(torch.autograd.Function):
             self.save_for_backward(input)
 
         # TODO: use GPU memory directly if possible
-        input_arr = input.cpu().numpy()
+        input_arr = input.detach().numpy()
         if any(s == 0 for s in input_arr.strides):
             # TODO: remove when Numpy issue #9165 is fixed
             # https://github.com/numpy/numpy/pull/9177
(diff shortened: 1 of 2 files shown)


odl/contrib/torch/operator.py
```diff
@@ -102,7 +102,7 @@ def forward(self, input):
             self.save_for_backward(input)
 
         # TODO: use GPU memory directly if possible
-        input_arr = input.cpu().numpy()
+        input_arr = input.detach().numpy()
```
kohr-h: If `input` is on the GPU, this fails. We need both `detach()` and `cpu()`.

olivierverdier: Do you mean `input.cpu().detach().numpy()`?

kohr-h: Yes, or `input.detach().cpu().numpy()`, whatever you like better. Perhaps a little helper `to_numpy` would make that less ugly.

odl/contrib/torch/operator.py
```diff
@@ -16,6 +16,12 @@
 from __future__ import division
 import numpy as np
 import torch
+import warnings
+from pkg_resources import parse_version
+
+if parse_version(torch.__version__) < parse_version('0.4'):
+    warnings.warn("This interface is designed to work"
```
adler-j: the strings will be concatenated like "workwith", you need to add a space after "work"

PR #1373, opened 2018-06-15
BUG: fix version check of pyFFTW regarding git revs

kohr-h:

PR #1402, opened 2018-06-28
PERF: replace slow parse_version by inlined fast one

kohr-h:
We once had the issue with the slowness of `from pkg_resources import parse_version`. Turns out that [the implementation of that function](https://github.com/pypa/setuptools/blob/5c816483500519a07a3db4364daaced379780eb7/pkg_resources/__init__.py#L120-L124) just consists of 4 lines, basically only calling into `packaging.version`. Fortunately, that module loads much faster (about 5 ms for me), so I copied over the function.

No more hacks to determine package versions.


doc/source/conf.py
```diff
@@ -7,12 +7,12 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 from __future__ import print_function
-from distutils.version import StrictVersion
 import sphinx
 import glob
 import os
 import sphinx_rtd_theme
 import sys
+from pkg_resources import parse_version
```
adler-j: `odl.util` or `parse_version`?

odl/trafos/backends/pyfftw_bindings.py
```diff
@@ -23,11 +23,8 @@
 except ImportError:
     PYFFTW_AVAILABLE = False
 else:
-    # Cheap way of getting the version, works with Git revision versions like
-    # '0.10.4+201.gdb78c2b'
-    _maj, _min, _patch = [int(n.split('+')[0])
-                          for n in pyfftw.__version__.split('.')[:3]]
-    if (_maj, _min, _patch) < (0, 10, 3):
+    from odl.util import parse_version
```
adler-j: Move to top

doc/source/conf.py
kohr-h: Here I think it's better to go for the standard library, since we have a potentially failing import of `odl` further down. It's only for documentation build anyway.

kohr-h:
> But is there no way around the licencing issue?

Well I can try to "reimplement" it, but even if I don't look at it for a week and start from scratch, it's quite likely that I end up with the exact same code.

kohr-h:
>> But is there no way around the licencing issue?
> 
> Well I can try to "reimplement" it, but even if I don't look at it for a week and start from scratch, it's quite likely that I end up with the exact same code.

Ah never mind. We can use [`packaging.version.parse`](https://packaging.pypa.io/en/latest/version/#packaging.version.parse) directly.

odl/space/entry_points.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
```
kohr-h: Actually I tried to update only if some other part of the file was changed as well. This one slipped

kohr-h:
Seems like `packaging` is a separate package, weird thing is that it's also included in `setuptools` in some weird way. I'll see if I can find another route to the required functionality without involving 3rd party packages.

kohr-h:
It seems to me that there are the following possibilities:

1. Keep as-is. Always a viable option.
2. Inline quite a bit of code from `pkg_resources`. Quite bad due to issues with license and maintenance.
3. Use the 3rd party package `packaging`.

To make option 3 more attractive: It's not really 3rd party since it's maintained by `pypa` together with `setuptools` and `pip`. And it's included in `setuptools` directly, so we use it one way or another.
And some version checks we make are really a bit ad-hoc and [do break sometimes](https://github.com/odlgroup/odl/pull/1373). So my favorite option is 3.

adler-j:
Go with 3 for now, when we've forgotten the `pkg_resources` implementation we could do something similar.

PR #1403, opened 2018-06-29
Remove STIR bindings

olivierverdier:
The rationale is that the external package [odlpet](https://github.com/olivierverdier/odlpet) takes over


adler-j:
Totally fine by me!

PR #1410, opened 2018-07-04
ENH: Add nbytes to pspace

adler-j:


odl/space/pspace.py
```diff
@@ -281,6 +281,11 @@ def __len__(self):
         """
         return len(self.spaces)
 
+    @property
+    def nbytes(self):
+        """Total number of bytes used by an element of this space in memory."""
```
kohr-h: Could be my formulation, but better would be "Total number of bytes in memory used by an element of this space."
The element property already does it like this.

Issue #1415, opened 2018-07-31
ASTRA scaling factor change in next release


wjp:
Hi,

A quick heads-up: in the next ASTRA release the scaling factor of the ASTRA BP code ODL uses has been changed by a factor `reco_space.cell_volume ** 2` (since commit astra-toolbox/astra-toolbox@9fa2ffdc5348f8f19de48d06a72b82bdc1ba8f22 ). This should only require changing

```
scaling_factor *= (src_radius ** 2 * det_px_area ** 2 / reco_space.cell_volume ** 2)
```

into

```
scaling_factor *= (src_radius ** 2 * det_px_area ** 2)
```

in `astra_cuda_bp_scaling_factor` for `ConeFlatGeometry`.

adler-j:
Just make sure we have a sharp version number so we can retain backwards compatibility. Should be an easy fix!

ozanoktem:
Somewhat off-topic, but not entirely: @wjp , what happened to the initiative for introducing native support for curved detectors (cylindrical and spherical) in ASTRA? Just curious.

wjp:
One more heads up:

I've done a complete overhaul of all 2D and 3D operator scaling, to get rid of some lingering inconsistencies in definition differences between 2D/3D and parallel/fan beam geometries, in particular for detector pixel scaling and oblique projections.

In the upcoming version, all projectors scale their output to match mathematical line integrals, regardless of detector pixel size, obliqueness, etc. (I assume ODL has realized this was the way to go a long time ago :-) )

Furthermore, I've made the approximation of the GPU cone beam BP density weighting more accurate (at hopefully limited computational cost since the extra factor could actually be folded into existing constants, it turns out), and also adopted the 2D fan beam GPU code to match. This density weighting is now enabled by default.

On top of this, the annoying distance scaling factor you had to use to correct the DensityWeighting in ODL is now handled automatically. The remaining adjoint scaling ODL has to do should now be limited to the different inner products on the ODL spaces vs the ASTRA spaces.

And finally, these changes are detectable by doing `astra.has_feature('projectors_scaled_as_line_integrals')` and `astra.has_feature('fan_cone_BP_density_weighting_by_default')`

The code is currently at https://github.com/wjp/astra-toolbox/tree/consistent_scaling .

Once this branch is merged (and the remaining Windows build issues are fixed...), we should be about ready for a 2.0.0 pre-release.

kohr-h:
[...]

PR #1416, opened 2018-08-17
ENH: Add a tomography example for the adupdates solver.

sbanert:
This pull request adds an example how to use the alternating dual updates method (#1243) and vector-valued stepsizes (#1166) in TV regularization for tomography.

### DIFF
--- a/examples/solvers/adupdates_tomography.py
+++ b/examples/solvers/adupdates_tomography.py
@@ -1,0 +1,148 @@
+r"""Total-variation regularized tomography example using the adupdates solver.
+
+This example solves a linear inverse problem of the form :math:`Ax = y`, where
+:math:`x \in \mathbb{R}^n` is the (unknown) original image to be reconstructed,
+:math:`y \in \mathbb{R}^m` is the noisy data and :math:`A \in \mathbb{R}^{m
+\times n}` is the measurement matrix describing the discretized physical model.
+
+To solve this problem, we first split the measurement matrix and the data into
+:math:`m` parts and solve the least squares problem
+
+..math::
+    \min_{x\in\mathbb{R}^n} D(x) := \sum_{j = 1}^m \| A_j x - y_j \|^2.
+
+To regularize, we add terms for the total variation and for a functional which
+guarantees that the solution will be pointwise nonnegative. The variational
+regularization problem we are solving is therefore
+
+..math::
+    \min_{x\in \mathbb{R}^n_+} D(x) + \sum_{i = 1}^d (\| \partial_{i, 1} x \|_1
+    + \| \partial_{i, 2} x \|_1.
+
+Here, :math:`\partial_{i, 1}` and :math:`\partial_{i, 2}` contain the even and
+odd components, respectively, of the discretized :math:`i`-th partial
+derivative, and :math:`d` is the dimension of the tomography problem. In this
+example, :math:`d = 2`. We solve the problem with the alternating dual updates
+method. For further details, see
+`[MF2015] <http://ieeexplore.ieee.org/document/7271047/>`_.
+
+References
+----------
+[MF2015] McGaffin, M G, and Fessler, J A. *Alternating dual updates
+algorithm for X-ray CT reconstruction on the GPU*. IEEE Transactions
+on Computational Imaging, 1.3 (2015), pp 186--199.
+"""
+
+import numpy as np
+import odl
+
+# The following parameters determine how to split the data (sinograms) and if
+# the solver should do the inner iterations in a fixed order or at random.
+SPLIT_METHOD = 'interlaced'  # How to split the data ('block' or 'interlaced')?
+SPLIT_NUMBER = 20     # How many pieces of data?
+RANDOM = True         # Choose the oder of the inner iterations at random?
+
+# --- Create simulated data (phantom) ---
+
+# Reconstruction space: Set of two-dimensional quadratic images.
+reco_space = odl.uniform_discr(min_pt=[-40.0, -40.0],
+                               max_pt=[40.0, 40.0],
+                               shape=[1024, 1024])
+phantom = odl.phantom.shepp_logan(reco_space, modified=True)
+
+# Create the forward operators. They correspond to a fully sampled parallel
+# beam geometry.
+geometry = odl.tomo.parallel_beam_geometry(reco_space)
+
+if SPLIT_METHOD == 'block':
+    # Split the data into blocks:
+    # 111 222 333
+    ns = geometry.angles.size // SPLIT_NUMBER
+    ray_trafos = [odl.tomo.RayTransform(reco_space,
+                                        geometry[i * ns:(i + 1) * ns])
+                  for i in range(SPLIT_NUMBER)]
+
+elif SPLIT_METHOD == 'interlaced':
+    # Split the data into slices:
+    # 123 123 123
+    ray_trafos = [odl.tomo.RayTransform(reco_space,
+                                        geometry[i::SPLIT_NUMBER])
+                  for i in range(SPLIT_NUMBER)]
+else:
+    raise ValueError('unknown data split`{}`: typo?'.format(SPLIT_METHOD))
+
+# Create the artificial data.
+data_spaces = [op.range for op in ray_trafos]
+noisefree_data = [op(phantom) for op in ray_trafos]
+data = [proj + 0.10 * np.ptp(proj) * odl.phantom.white_noise(proj.space)
+        for proj in noisefree_data]
+
+# Functionals and operators for the total variation. This is the l1 norm of the
+# (discretized) gradient of the reconstruction. For each of the dimensions
+# we create two functionals and two operators.
+
+# Start with empty lists ...
+tv_functionals = []
+tv_operators = []
+tv_stepsizes = []
+
+# ... and for each dimension of the reconstruction space ...
+reco_shape = reco_space.shape
+reco_dim = len(reco_shape)
+for dim in range(reco_dim):
+    # ... add two operators taking only the even and odd elements,
+    # respectively, in that dimension.
+    partial_der = odl.PartialDerivative(
+        reco_space, dim, pad_mode='order0')
+    all_points = list(np.ndindex(reco_shape))
+    even_pts = [list(p) for p in all_points if p[dim] % 2 == 0]
+    even_pts = np.array(even_pts).T.tolist()
+    odd_pts = [list(p) for p in all_points if p[dim] % 2 == 1]
+    odd_pts = np.array(odd_pts).T.tolist()
+    op1 = reco_space.cell_sides[dim] * odl.SamplingOperator(
+        reco_space, even_pts) * partial_der
+    op2 = reco_space.cell_sides[dim] * odl.SamplingOperator(
+        reco_space, odd_pts) * partial_der
+    tv_functionals += [odl.solvers.L1Norm(op1.range),
+                       odl.solvers.L1Norm(op2.range)]
+    tv_operators += [op1, op2]
+    tv_stepsizes += [0.5 / reco_shape[dim], 0.5 / reco_shape[dim]]
+
+# Functional and operator enforcing the nonnegativity of the image.
+nonneg_functional = odl.solvers.IndicatorNonnegativity(reco_space)
+nonneg_operator = odl.IdentityOperator(reco_space)
+nonneg_stepsize = 1.0
+
+# ... and the data fit functionals. The coefficient is a regularization
+# paratemeter, which determines the tradeoff between data fit and regularity.
+data_fit_functionals = [1.0 *
+                        odl.solvers.L2NormSquared(ds).translated(rhs)
+                        for (ds, rhs) in zip(data_spaces, data)]
+# In the stepsizes, we avoid the possible division by zero by adding a small
+# positive value. The matrix corresponding to the operator `op` has only
+# nonnegative entries, which ensures that the final results are positve.
+data_fit_stepsizes = [1.0 / (1e-6 + op(op.adjoint(ds.one())))
+                      for (ds, op) in zip(data_spaces, ray_trafos)]
+# Alternative choice without vector-valued stepsizes could be
+# data_fit_stepsizes = [1.0 / op.norm(estimate=True) ** 2 for op in ray_trafos]
+
+# Now we build up the ingredients of our algorithm:
+# Start at a zero image, ...
+x = reco_space.zero()
+
+# ... collect all the functionals, ...
+g = [nonneg_functional] + data_fit_functionals + tv_functionals
+
+# ... collect all the operators, ...
+L = [nonneg_operator] + ray_trafos + tv_operators
+
+# ... and collect all the inner stepsizes, which were chosen according to the
+# properties of the operators in `L`.
+inner_stepsizes = [nonneg_stepsize] + data_fit_stepsizes + tv_stepsizes
+
+odl.solvers.adupdates(x, g, L, stepsize=1.0, inner_stepsizes=inner_stepsizes,
+                      niter=5, random=RANDOM, callback=None,
+                      callback_loop=None)
+
+# Show the result within a window between zero and one.
+x.show(vmin=0.0, vmax=1.0)


sbanert:
I don’t understand the Travis failure. :(

kohr-h:
Seems to be [this issue](https://github.com/pytest-dev/pytest/issues/3825). Can you try older versions of pytest? Looks like we need to blacklist a few.

sbanert:
I can confirm that `pytest 3.6.1` works on my laptop until it hits `odl/test/tomo/backends/astra_cuda_test.py .[1]    5393 abort (core dumped)  pytest --pep8` (my computer does not have an Nvidia GPU), whereas `pytest 3.7.1` breaks with the recursion problem.

aringh:
Regarding the ASTRA issue: to me it looks like #684, which should have been fixed in #1376. But for the fix in #1376 to work, one need the ASTRA update that corresponds to the bug report astra-toolbox/astra-toolbox#109 that @adler-j mentioned. However, I think that this is still not in the latest release of ASTRA.

kohr-h:
You can check the ASTRA thing with a dev build from @wjp's conda channel:
```
conda install -c wjpalenstijn astra-toolbox=1.9.0dev
```

sbanert:
I'll try this later, since the pytest upgrade breaks my tests anyway. ;-)

sbanert:
Regarding the pytest recursion issue: It does not occur in version 3.6.4 (I tested in on my workstation), which is the newest version available via conda previous to 3.7.1, and according to the link @kohr-h posted, it should be fixed in version 3.7.1.

kohr-h:
Thanks for investigating, @sbanert!

> should be fixed in version 3.7.1

I guess you mean 3.7.2 (the upcoming release).

In the meanwhile, we should blacklist the affected versions 3.7.0 and 3.7.1. Not that it's a terribly important core dependency, but the deps that are pulled in by default should work. There are two places where we have control over the version. 

Here:
https://github.com/odlgroup/odl/blob/dd01b8530cb6908e52e22a5f744da6bc9acf6421/setup.py#L147

and here:
https://github.com/odlgroup/odl/blob/dd01b8530cb6908e52e22a5f744da6bc9acf6421/test_requirements.txt#L1

For the syntax to blacklist versions you can check the `numpy` line in the [requirements.txt](https://github.com/odlgroup/odl/blob/master/requirements.txt) file.

However, I suggest that we deal with it in a separate issue, since I think we might want to bail out or at least warn in `odl.test()` as well.

sbanert:
Regarding the ASTRA issue: After applying @kohr-h’s upgrade suggestion, the tests do no longer abort. However, I get 12 failures like this:
```python
__________________ test_projector[ geom='cone2d' - impl='astra_cpu' - angles='random' - in-place ] __________________

projector = RayTransform: uniform_discr([-20., -20.], [ 20.,  20.], (100, 100), dtype='float32') -> DiscreteLp(    FunctionSpace(I...
        [-29.7, -29.1, -28.5, ...,  28.5,  29.1,  29.7]
    ),
    rn((100, 100), dtype='float32', weighting=0.03651))
in_place = True

    def test_projector(projector, in_place):
        """Test Ray transform forward projection."""
    
        # TODO: this needs to be improved
        # Accept 10% errors
        places = 1
    
        # Create Shepp-Logan phantom
        vol = projector.domain.one()
    
        # Calculate projection
        if in_place:
            proj = projector.range.zero()
            projector(vol, out=proj)
        else:
            proj = projector(vol)
    
        # We expect maximum value to be along diagonal
        expected_max = projector.domain.partition.extent[0] * np.sqrt(2)
>       assert almost_equal(proj.ufuncs.max(), expected_max, places=places)
E       AssertionError: assert False
E        +  where False = almost_equal(33.670719, 56.568542494923804, places=1)
E        +    where 33.670719 = <bound method TensorSpaceUfuncs.max of <odl.util.ufuncs.TensorSpaceUfuncs object at 0x7ff89fbf6518>>()
E        +      where <bound method TensorSpaceUfuncs.max of <odl.util.ufuncs.TensorSpaceUfuncs object at 0x7ff89fbf6518>> = <odl.util.ufuncs.TensorSpaceUfuncs object at 0x7ff89fbf6518>.max
E        +        where <odl.util.ufuncs.TensorSpaceUfuncs object at 0x7ff89fbf6518> = DiscreteLp(    FunctionSpace(IntervalProd([  0.1072, -30.    ], [  6.1924,  30.    ]), out_dtype='float32'),\n    nonun...  11.42340565],\n     [ 20.49684906,  24.00613022,  24.0072403 , ...,  14.09786224,\n       12.93267536,  11.78796482]]\n).ufuncs

odl/test/tomo/operators/ray_trafo_test.py:217: AssertionError
```
Might these errors be due to #1415?

kohr-h:
> Might these errors be due to #1415?

Not exactly that, because we only look at forward projection here and check that we get close to the maximum ray intersection length. The change only affects the backprojection.
However, that's not to say that no other scaling factors changed :wink:

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,151 @@
...
+DIMENSIONS = 40.0     # Dimensions of reconstruction.
+RESOLUTION = 1024     # Discretization level of reconstruction space.
+SPLIT_METHOD = 'interlaced'  # How to split the data?
+SPLIT_NUMBER = 20     # How many pieces of data?
+NOISE_LEVEL = 5.0     # How much noise to add to the sinograms
+DATA_FIT_WEIGHT = 1.0  # Regularization parameter (coefficient of the data fit)
+EPSILON = 1e-6        # make sure that we do not divide by zero by adding this
+STEPSIZE = 1.         # Stepsize of the outer iterations
+NITER = 5             # Number of iterations
+RANDOM = True         # Choose the oder of the inner iterations at random?
```
kohr-h: While I appreciate the flexibility that comes with this parametrization, I think that this level of flexibility tends to scare off people. There are just too many choices. The script misses its purpose as an example (aimed at simplicity and comprehensibility) and rather comes as a command line utility in disguise (aimed at flexibility and full-featuredness).

On top of that, not all of the parameters are really needed. Some are just "for implementer's convenience" and could be handled in a better way. Examples:

- `DIMENSIONS`: Not needed. You can simply hardcode the extent of the reconstruction space and retrieve it later when you need it. That still gives you exactly one place to change it. (Apart from that, the name is slightly wrong: `HALF_DIMENSIONS` would be more accurate.)
- `RESOLUTION`: Apart from the fact that most people would not think of resolution in this way, you don't actually need it as a module constant, just like `DIMENSIONS`.

... and so on.

Ultimately, I think the only interesting knobs that I imagine somebody playing with this script would like to turn are related to the problem splitting, i.e., `SPLIT_NUMBER` AND `SPLIT_METHOD`, perhaps `RANDOM`. For the others: just do it as in the other examples and inline them.

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,151 @@
...
+    # 123 123 123
+    ray_trafos = [odl.tomo.RayTransform(reconstruction_space,
+                                        geometry[i::SPLIT_NUMBER])
+                  for i in range(SPLIT_NUMBER)]
+else:
+    raise ValueError('unknown data split`{}`: typo?'.format(SPLIT_METHOD))
+
+# Create the artificial data.
+data_spaces = [op.range for op in ray_trafos]
+data = [op(phantom) + NOISE_LEVEL * odl.phantom.white_noise(ds)
```
kohr-h: Here, for instance, there is no advantage of having `NOISE_LEVEL` as a constant above. The value is still entirely magical, and to change it, people will have to navigate to the top of the file.

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,151 @@
...
+    # ... add two operators taking only the even and odd elements,
+    # respectively, in that dimension.
+    partial_der = odl.PartialDerivative(
+        reconstruction_space, j, pad_mode='order0')
+    allpoints = list(np.ndindex(reco_shape))
+    sampling_points1 = [list(p) for p in allpoints if p[j] % 2 == 0]
+    sampling_points1 = np.array(sampling_points1).T.tolist()
+    sampling_points2 = [list(p) for p in allpoints if p[j] % 2 == 1]
+    sampling_points2 = np.array(sampling_points2).T.tolist()
+    op1 = 2 * DIMENSIONS / RESOLUTION * odl.SamplingOperator(
```
kohr-h: `2 * DIMENSIONS / RESOLUTION` looks very much like `reco_space.cell_sides[0]` to me

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,151 @@
...
+data_fit_functionals = [DATA_FIT_WEIGHT *
+                        odl.solvers.L2NormSquared(ds).translated(rhs)
+                        for (ds, rhs) in zip(data_spaces, data)]
+data_fit_stepsizes = [1.0 / (EPSILON + op(op.adjoint(ds.one())))
+                      for (ds, op) in zip(data_spaces, ray_trafos)]
+# Alternative choice without vector-valued stepsizes could be
+# data_fit_stepsizes = [1.0 / op.norm(estimate=True) ** 2 for op in ray_trafos]
+
+# Now we build up the ingredients of our algorithm:
+# Start at a black image, ...
```
kohr-h: Color only makes sense with a color map. Arrays are "zero", not "black".

sbanert:
Can I ignore the Python 2 failure? Otherwise, I think that I addressed all remarks.

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+# The following parameters determine how to split the data (sinograms) and if
+# the solver should do the inner iterations in a fixed order or at random.
+SPLIT_METHOD = 'interlaced'  # How to split the data ('block' or 'interlaced')?
+SPLIT_NUMBER = 20     # How many pieces of data?
+RANDOM = True         # Choose the oder of the inner iterations at random?
+
+# --- Create simulated data (phantom) ---
+
+# Reconstruction space: Set of two-dimensional quadratic images with
+# given resolution.
```
kohr-h: This seems to be a comment based on the old version

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+    ray_trafos = [odl.tomo.RayTransform(reconstruction_space,
+                                        geometry[i::SPLIT_NUMBER])
+                  for i in range(SPLIT_NUMBER)]
+else:
+    raise ValueError('unknown data split`{}`: typo?'.format(SPLIT_METHOD))
+
+# Create the artificial data.
+data_spaces = [op.range for op in ray_trafos]
+data = [op(phantom) + 5.0 * odl.phantom.white_noise(ds)
+        for (op, ds) in zip(ray_trafos, data_spaces)]
```
kohr-h: It's fine like this, but if you want, you can use a fixed *relative* amount of noise (for some notion of "amount"), e.g.
```python
noisefree_data = [op(phantom) for op in ray_trafos]
data = [proj + 0.05 * np.ptp(proj) * odl.phantom.white_noise(proj.space) for proj in noisefree_data]
```
(`np.ptp` is the one-function version of `max - min` that I didn't know until recently)

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+
+# Start with empty lists ...
+tv_functionals = []
+tv_operators = []
+tv_stepsizes = []
+
+# ... and for each dimension of the reconstruction space ...
+reco_shape = reconstruction_space.shape
+reco_dim = len(reco_shape)
+for j in range(reco_dim):
```
kohr-h: `j` doesn't convey much, I'd prefer this:
```python
for dim in range(reconstruction_space.ndim):
    # Do stuff with `dim` instead of `j`
```
We also use `axis` often in lieu of `dim` here.

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+    # ... add two operators taking only the even and odd elements,
+    # respectively, in that dimension.
+    partial_der = odl.PartialDerivative(
+        reconstruction_space, j, pad_mode='order0')
+    allpoints = list(np.ndindex(reco_shape))
```
kohr-h: Small thing, `all_points` reads nicer

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+    sampling_points1 = [list(p) for p in allpoints if p[j] % 2 == 0]
+    sampling_points1 = np.array(sampling_points1).T.tolist()
+    sampling_points2 = [list(p) for p in allpoints if p[j] % 2 == 1]
+    sampling_points2 = np.array(sampling_points2).T.tolist()
+    op1 = reconstruction_space.cell_sides[j] * odl.SamplingOperator(
+        reconstruction_space, sampling_points1) * partial_der
+    op2 = reconstruction_space.cell_sides[j] * odl.SamplingOperator(
+        reconstruction_space, sampling_points2) * partial_der
```
kohr-h: IMO the `reconstruction_space` name takes up a bit too much space, why not use the shorter `reco_space` as in the other examples?

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+for j in range(reco_dim):
+    # ... add two operators taking only the even and odd elements,
+    # respectively, in that dimension.
+    partial_der = odl.PartialDerivative(
+        reconstruction_space, j, pad_mode='order0')
+    allpoints = list(np.ndindex(reco_shape))
+    sampling_points1 = [list(p) for p in allpoints if p[j] % 2 == 0]
+    sampling_points1 = np.array(sampling_points1).T.tolist()
+    sampling_points2 = [list(p) for p in allpoints if p[j] % 2 == 1]
+    sampling_points2 = np.array(sampling_points2).T.tolist()
```
kohr-h: "1" and "2" don't convey meaning, use "even" and "odd"
Also, since the values are consumed immediately, shorter names are okay, e.g. `odd_pts` and `even_pts`.

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+To regularize, we add terms for the total variation and for a functional which
+guarantees that the solution will be pointwise nonnegative. The variational
+regularization problem we are solving is therefore
+
+..math::
+    \min_{x\in \mathbb{R}^n_+} D(x) + \sum_{i = 1}^d (\| \partial_{i, 1} x \|_1
+    + \| \partial_{i, 2} x \|_1.
+
+Here, :math:`\partial_{i, 1}` and :math:`\partial_{i, 2}` contain the even and
+odd components, respectively, of the discretized :math:`i`th partial
```
kohr-h: Not sure, but
```
:math:`i`th
```
may not be rendered correctly -- the backtick must be followed by a whitespace or punctuation, as far as I know.
```
:math:`i`-th
```
should work, though.

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+
+# ... and the data fit functionals. The coefficient is a regularization
+# paratemeter, which determines the tradeoff between data fit and regularity.
+data_fit_functionals = [1.0 *
+                        odl.solvers.L2NormSquared(ds).translated(rhs)
+                        for (ds, rhs) in zip(data_spaces, data)]
+# In the stepsizes, we avoid the possible division by zero by adding a small
+# positive value. The matrix corresponding to the operator `op` has only
+# nonnegative entries, which makes this ensures that the final results are
+# positve.
```
kohr-h: I guess you wanna delete "makes this"

examples/solvers/adupdates_tomography.py
```diff
@@ -0,0 +1,149 @@
...
+data_fit_stepsizes = [1.0 / (1e-6 + op(op.adjoint(ds.one())))
+                      for (ds, op) in zip(data_spaces, ray_trafos)]
+# Alternative choice without vector-valued stepsizes could be
+# data_fit_stepsizes = [1.0 / op.norm(estimate=True) ** 2 for op in ray_trafos]
+
+# Now we build up the ingredients of our algorithm:
+# Start at a zero image, ...
+x = reconstruction_space.zero()
+
+# collect all the functionals, ...
```
kohr-h: Above you also used `...` at the front, which made the continuation more obvious.

adler-j:
Just a general comment w.r.t the git flow here. The commit messages here are typically of this type: "MAINT: Remove algorithm parameters and use keyword arguments."

While this is _locally_ informative (e.g. me following this PR, its obvious what you mean), being locally informative is not really what we're after down the line. I'd personally prefer if there was at least a hint as to _where_ the change was done, e.g. "MAINT: Remove algorithm parameters and use kwargs in adupdates example." That way, one year from now when you're confused as to why things are as they are, it's a lot easier to scroll through the commits.

Another option (actually the best option if you had infinite time available) is that you use these "local" commit messages during development, and then before the branch is merged, you rebase the branch and merge commits appropriately, also updating the commit messages. For example, this whole PR could quite frankly be squashed into a single commit with all the intermediate commit messages ("MAINT: Add relative instead of absolute noise.") removed.

sbanert:
I can do, as you suggested, an interactive rebase before we merge. My reason is basically that it is easier to see what I changed after the reviews and that it’s easier to revert specific changes if someone disagrees with them.

kohr-h:
That's a very reasonable general workflow @sbanert.
In this case, though, the PR was already approved, so the only thing left to do was "whatever you want done before merge". When that's done I'd just merge it after having a last glance.

So please go ahead and take any final steps left to do, ping me, and I'll merge it.

PR #1417, opened 2018-08-21
TST: Blacklist pytest 3.7

sbanert:
This pull request blacklists those `pytest` versions which cause an infinite recursion (see the discussion in #1416). The `pytest` issue was fixed here https://github.com/pytest-dev/pytest/pull/3771 and should be included in version `3.7.2`.

### DIFF
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -649,7 +649,7 @@ def test(arguments=None):
         raise ImportError('ODL tests cannot be run without `pytest` installed.'
                           '\nRun `$ pip install [--user] odl[testing]` in '
                           'order to install `pytest`.')
-    if not version.parse(pytest.__version__) < version.parse("3.7*"):
+    if version.parse(pytest.__version__) >= version.parse("3.7"):
         raise RuntimeError('ODL tests cannot be run with `pytest` '
                            'version `3.7` or higher.\nRun `$ pip install '
                            '"pytest<3.7"` in order to install '
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -16,6 +16,7 @@ import sys
 import os
 import warnings
 from time import time
+from packaging import version
 
 from odl.util.utility import run_from_ipython, is_string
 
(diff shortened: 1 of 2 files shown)


kohr-h:
> have again no idea what’s going on here. This time, it’s definitely a non-blacklisted version of pytest (3.7.2), and still some errors occur in odl.util.

The weird thing is that only `odl.util` is affected. Without taking a detailed look, I suspect the quick fix for the infinite recursion bug since it mocks around with the collection of tests. The error occurs when `getattr(module, "pytest_plugins", [])` returns a module instead of a list (in our case, `module` is the `odl.util` module).
For now I agree, blacklist all of the 3.7 releases, but if possible, file a bug report on the pytest issue tracker. I'm sure there are not so many with the same setup as ours, so it might really be a corner case.

odl/util/testutils.py
```diff
@@ -648,6 +648,11 @@ def test(arguments=None):
         raise ImportError('ODL tests cannot be run without `pytest` installed.'
                           '\nRun `$ pip install [--user] odl[testing]` in '
                           'order to install `pytest`.')
+    if pytest.__version__ in ['3.7.0', '3.7.1']:
+        raise ImportError('ODL tests cannot be run with `pytest` '
```
kohr-h: Make this a `RuntimeError`

odl/util/testutils.py
```diff
@@ -648,6 +648,11 @@ def test(arguments=None):
         raise ImportError('ODL tests cannot be run without `pytest` installed.'
                           '\nRun `$ pip install [--user] odl[testing]` in '
                           'order to install `pytest`.')
+    if pytest.__version__ in ['3.7.0', '3.7.1']:
```
kohr-h: This doesn't cover the same range as the pip requirements (not 3.7.2 and not the possibly broken subsequent versions).
Suggestion:
```python
from packaging import version
...
# TODO(sbanert): adapt constraint when upstream issue is fixed
# See <link to pytest issue>
if not version.parse(pytest.__version__) < version.parse("3.7*"):
    raise RuntimeError(...)
```
This uses the faster-to-import `packaging.version`, and it includes dev versions like 3.7.0dev0.

odl/util/testutils.py
```diff
@@ -648,6 +648,11 @@ def test(arguments=None):
...
+        raise ImportError('ODL tests cannot be run with `pytest` '
+                          'versions `3.7.0` and `3.7.1`.\nRun `$ pip install '
+                          '[--user] odl[testing]` in order to install '
+                          '`pytest`.')
```
kohr-h: The helpful part of the error message isn't that helpful here, better `Run pip install "pytest<3.7"`

odl/util/testutils.py
kohr-h: I mean faster than `pkg_resources.parse_version`. The dependency on `packaging` has been added recently.

PR #1419, opened 2018-08-24
BUG: fix radii used in noise_power_spectrum (contrib)

kohr-h:
Small fix for the radii as used in `noise_power_spectrum`.

- The old version used the grid points, thereby not starting at 0 and not going all the way to the maximum radius (missing half a pixel). Now we use explicitly 0 as first bin edge and compute the last bin edge as the maximum distance of all the corners of the *domain* from the origin.
- In the old version, the radial spacing would be equal to the diagonal length of a cell. Now we use more bins, such that for square images, the 1D step is the same as the 2D step.
- The number of points that fall into a radial bin fluctuates quite a bit, so I added an option to increase the bin size, making the number of points in a bin as a function of radius look more like a ramp. See image below.

The PR also factors out the radial summing into a separate function since it's actually used by many FOMs.

```python
import odl
from odl.contrib.fom.util import radial_sum

space = odl.uniform_discr([-1, -1], [1, 1], (200, 200))
rsum_1 = radial_sum(space.one(), binning_factor=1)
rsum_2 = radial_sum(space.one(), binning_factor=2)
rsum_4 = radial_sum(space.one(), binning_factor=4)

fig, ax = plt.subplots()
ax.plot(rsum_1.space.grid.coord_vectors[0], rsum_1, label="binning 1")
ax.plot(rsum_2.space.grid.coord_vectors[0], rsum_2, label="binning 2")
ax.plot(rsum_4.space.grid.coord_vectors[0], rsum_4, label="binning 4")
ax.legend()
ax.set_title("Number of points in radial bins")
```
radial_binning


adler-j:
> The old version used the grid points, thereby not starting at 0 and not going all the way to the maximum radius (missing half a pixel). Now we use explicitly 0 as first bin edge and compute the last bin edge as the maximum distance of all the corners of the domain from the origin.
> In the old version, the radial spacing would be equal to the diagonal length of a cell. Now we use more bins, such that for square images, the 1D step is the same as the 2D step.

This sounds like a good change!

> The number of points that fall into a radial bin fluctuates quite a bit, so I added an option to increase the bin size, making the number of points in a bin as a function of radius look more like a ramp. See image below.

Wouldn't it be possible to simply normalize by the number of bins? E.g. compute it as you do here and divide by it later on.

> IMAGE

Why does changing the binning change the "slope"? I'd probably expect a scaling here

odl/contrib/fom/supervised.py
```diff
@@ -693,21 +694,11 @@ def noise_power_spectrum(data, ground_truth, radial=False):
...
-        n_bins = max(*nps.shape)
-        radial_nps, rad = np.histogram(r, weights=nps, bins=n_bins)
-
-        out_spc = odl.uniform_discr(0, np.max(r), n_bins,
-                                    impl=f1.space.impl,
-                                    dtype=f1.space.real_dtype,
-                                    axis_labels=['$\\^{r}$'])
-
-        return out_spc.element(radial_nps)
+        return radial_sum(nps)
```
adler-j: Great change!

odl/contrib/fom/util.py
```diff
@@ -393,3 +394,36 @@ def haarpsi_weight_map(img1, img2, axis):
     np.abs(img2_lvl3, out=img2_lvl3)
 
     return np.maximum(img1_lvl3, img2_lvl3)
+
+
+def radial_sum(image, binning_factor=1):
+    """Sum image values over concentric annuli.
```
adler-j: From the name of the funciton, I'd probably expect the cumulative sum instead (e.g. sum over circles)

odl/contrib/fom/util.py
```diff
@@ -393,3 +394,36 @@ def haarpsi_weight_map(img1, img2, axis):
...
+
+    Parameters
+    ----------
+    image : `DiscreteLp` element
+        Input data whose radial sum should be computed.
+    binning_factor : positive int or float, optional
```
adler-j: an int is a float, so just say float.

Also, the default value should be 1.0 to make this more clear

kohr-h:
> The total sum of the values is unchanged.

```python
>>> np.sum(rsum_1)
40000.0
>>> np.sum(rsum_2)
40000.0
>>> np.sum(rsum_4)
40000.0
```
Of course, the *integral* is scaled by the binning factor, which is also what you expect by looking at the graph.

kohr-h:
And sure enough, you can also do the binning afterwards, but only for integer values:
```python
>>> np.allclose(rsum_2, rsum_1.asarray().reshape(-1, 2).sum(axis=1))
True
```

adler-j:
> Because the value at r is the number of points that fall into the radial bin at r, and with larger bins you get more counts. The total sum of the values is unchanged.

I guess I mean that we should normalize by "# points / area". This should be approximately 1, but still correct for the discretization.

> Because the value at r is the number of points that fall into the radial bin at r, and with larger bins you get more counts. The total sum of the values is unchanged.

I guess this is even more reason to prefer the cumulative sum version of this :/ Is there a strong reason that we prefer to do integration over "spherical shells"?

kohr-h:
> I guess I mean that we should normalize by "# points / area". This should be approximately 1, but still correct for the discretization.

Hm, yeah, I guess we should do that in the functions that use radial integration. In the sense that we should clarify that the radial version is *integrated* over spheres (up to discretization).
That makes me wonder a little bit what people out there use in practice. Do they compensate for the point density variation?
Anyway, I guess I'll make a companion `sherical_int` that does the compensation.



> I guess this is even more reason to prefer the cumulative sum version of this :/ Is there a strong reason that we prefer to do integration over annuli?

Well I guess we can always transform one into the other by integration or differentiation, respectively. But in practice, the spheres (annuli / spherical shells in practice) version is the more useful form (computation of SSNR or FSC, to name some (in-)famous quantities).
Note that doing cumulative sums just happens to fix the situation for integrating `one()`.

kohr-h:
>> I guess I mean that we should normalize by "# points / area". This should be approximately 1, but still correct for the discretization.
> 
> Hm, yeah, I guess we should do that in the functions that use radial integration. In the sense that we should clarify that the radial version is integrated over spheres (up to discretization).
> That makes me wonder a little bit what people out there use in practice. Do they compensate for the point density variation?
> Anyway, I guess I'll make a companion sherical_int that does the compensation.

I've played around a little bit with it, and not totally unexpectedly you get quite erratic behavior around zero. The smaller the bin size (or the finer the discretization), the more you'll get a spike at 0. I don't think that's very useful.
In practice it doesn't matter anyway since the way radial power spectra are used is that you take the quotient of two of those, and then you divide out any normalization factor (or vector) anyhow.

I'll keep it at the sum version, and if anyone feels the need for a normalization, it's easy to do afterwards.

kohr-h:
I've added the option to change the bin size in `noise_power_spectrum`, other than that I don't think there's anything left to do here.

Good to go in?

kohr-h:
Small addition: Did the rename to `spherical_sum` as mentioned above.

odl/contrib/fom/supervised.py
```diff
@@ -760,7 +761,8 @@ def haarpsi(data, ground_truth, a=4.2, c=None):
     return (scipy.special.logit(numer / denom) / a) ** 2
 
 
-def noise_power_spectrum(data, ground_truth, radial=False):
+def noise_power_spectrum(data, ground_truth, radial=False,
+                         radial_binning_factor=2):
```
adler-j: radial_binning_factor should be 2.0 by default, to show it can be float

PR #1420, opened 2018-08-24
MAINT: avoid ambiguous list indexing, closes #1413

### LINKED ISSUE #1413: Deprecation of nested lists as indices
[In the release notes of NumPy 1.15.0 it is announced](https://docs.scipy.org/doc/numpy-1.15.0/release.html#deprecations) that expressions like `arr[[[1, 0], [0, 0]]]` will be interpreted differently than today. That is due to ambiguity -- does the user want this:
```python
>>>  arr = array([[1, 2, 3],
...               [4, 5, 6]])
>>> arr[array([1, 0]), array([0, 0])]
array([4, 1])
```
or this?
```python
>>>  arr = array([[1, 2, 3],
...               [4, 5, 6]])
>>> arr[array([[1, 0], [0, 0]])]
array([[[4, 5, 6],
        [1, 2, 3]],

       [[1, 2, 3],
        [1, 2, 3]]])
```
Currently NumPy does the first version, which is equivalent to the outer list being a tuple, but in the future it will switch to the second version. Since we always want the former, we should remove all occurrences of nested lists of indices and always use tuples.

kohr-h:
We had a few places where we explicitly created index lists that shouldn't be lists. The unit tests unearthed a whole bunch of them, but there might be more. 
I guess we can go ahead with this (assuming the tests go green) and fix other places that trigger warnings later on.

### DIFF
--- a/odl/space/base_tensors.py
+++ b/odl/space/base_tensors.py
@@ -968,8 +968,9 @@ numpy.ufunc.reduceat.html
 
         # Default to showing x-y slice "in the middle"
         if indices is None and self.ndim >= 3:
-            indices = (slice(None),) * 2
-            indices += tuple(n // 2 for n in self.space.shape[2:])
+            indices = tuple(
+                [slice(None)] * 2 + [n // 2 for n in self.space.shape[2:]]
+            )
 
         if isinstance(indices, (Integral, slice)):
             indices = (indices,)
--- a/odl/util/numerics.py
+++ b/odl/util/numerics.py
@@ -663,9 +663,9 @@ def _apply_padding(lhs_arr, rhs_arr, offset, pad_mode, direction):
                                  ''.format(axis, lr, pad_len, n_rhs))
 
         # Slice tuples used to index LHS and RHS for left and right padding,
-        # respectively.
-        lhs_slc_l, lhs_slc_r = list(working_slc), list(working_slc)
-        rhs_slc_l, rhs_slc_r = list(working_slc), list(working_slc)
+        # respectively; we make 4 copies of `working_slc` as lists
+        lhs_slc_l, lhs_slc_r, rhs_slc_l, rhs_slc_r = map(
+            list, [working_slc] * 4)
 
         # We're always using the outer (excess) parts involved in padding
         # on the LHS of the assignment, so we set them here.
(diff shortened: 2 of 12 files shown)


odl/space/base_tensors.py
```diff
@@ -968,8 +968,9 @@ def show(self, title=None, method='', indices=None, force_show=False,
 
         # Default to showing x-y slice "in the middle"
         if indices is None and self.ndim >= 3:
-            indices = (slice(None),) * 2
-            indices += tuple(n // 2 for n in self.space.shape[2:])
+            indices = tuple(
+                [slice(None)] * 2 + [n // 2 for n in self.space.shape[2:]]
```
adler-j: No need to create a list, the generator is enough

odl/util/numerics.py
```diff
@@ -742,13 +748,15 @@ def _apply_padding(lhs_arr, rhs_arr, offset, pad_mode, direction):
             bdry_slc_l[axis] = left_slc
             bdry_slc_r = list(working_slc)
             bdry_slc_r[axis] = right_slc
+            bdry_slc_l, bdry_slc_r = map(tuple, [bdry_slc_l, bdry_slc_r])
```
adler-j: Just write one per line

```
bdry_slc_l = tuple(bdry_slc_l)
bdry_slc_r = tuple(bdry_slc_r)
```

Slight DRY violation, but it makes the code much more readable and maintainable for non-pythonistas.

odl/space/base_tensors.py
kohr-h: I can't do `* 2` or `+` with a generator, that would require chaining and stuff. This is the simplest implementation, I think.

odl/util/numerics.py
kohr-h: I'll do the cases with 2, but the ones with 4 just feel wrong to put in 4 lines. If the functional version is not great, I can also do list comprehension:
```python
# Old
lhs_slc_l, rhs_slc_l, lhs_slc_r, rhs_slc_r = map(
    tuple, [lhs_slc_l, rhs_slc_l, lhs_slc_r, rhs_slc_r])
# New
lhs_slc_l, rhs_slc_l, lhs_slc_r, rhs_slc_r = [
    tuple(lst) for lst in [lhs_slc_l, rhs_slc_l, lhs_slc_r, rhs_slc_r]
]
```
Or I just write a little inline helper
```python
def to_tuple(*args):
    return tuple(tuple(arg) for arg in args)
```

odl/space/base_tensors.py
adler-j: Use parentheses

odl/util/numerics.py
adler-j: For 3 or more, I'm fine with the `map` approach, its just that for 2 it's barely any reduction.

kohr-h:
Alright, merging after CI.

kohr-h:
Build issue with anaconda and python 2.7 here as well, merging anyway.

PR #1421, opened 2018-08-27
ENH: Minor improvements to mayo reader

adler-j:
Wouldn't you know? Yet another minor bugfix-spree for Mayo. DICOM is not fun at all.

Anyway, this should be the best version so far, and I also slightly reduced the memory redundancy.

### DIFF
--- a/odl/contrib/datasets/ct/mayo.py
+++ b/odl/contrib/datasets/ct/mayo.py
@@ -36,7 +36,6 @@ __all__ = ('load_projections', 'load_reconstruction')
 
 def _read_projections(folder, indices):
     """Read mayo projections from a folder."""
-    projections = []
     datasets = []
 
     # Get the relevant file names
@@ -47,7 +46,9 @@ def _read_projections(folder, indices):
 
     file_names = file_names[indices]
 
-    for file_name in tqdm.tqdm(file_names, 'Loading projection data'):
+    data_array = None
+
+    for i, file_name in enumerate(tqdm.tqdm(file_names, 'Loading projection data')):
         # read the file
         dataset = dicom.read_file(folder + '/' + file_name)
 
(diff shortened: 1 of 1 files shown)


odl/contrib/datasets/ct/mayo.py
```diff
@@ -252,13 +255,15 @@ def load_reconstruction(folder, slice_start=0, slice_end=-1):
...
-    # Compute geometry parameters
     voxel_size = np.array(list(pixel_size) + [pixel_thickness])
     shape = np.array([rows, cols, len(volumes)])
-    min_pt = (np.array(dataset.ImagePositionPatient) -
-              np.array(dataset.DataCollectionCenterPatient))
 
-    max_pt = min_pt + voxel_size * np.array([rows, cols, 0])
+    # Compute geometry parameters
+    mid_pt = (np.array(dataset.ReconstructionTargetCenterPatient) -
+              np.array(dataset.DataCollectionCenterPatient))
```
kohr-h: Seems like DICOM has a lot of (potentially conflicting) redundant settings.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -141,6 +136,11 @@ def load_projections(folder, indices=None):
     # This increases the resolution of the reconstructions.
     angles = angles - offset_angular
 
+    # We correct for the mean offset due to ffz in the detector.
```
kohr-h: What is ffz?

odl/contrib/datasets/ct/mayo.py
```diff
@@ -59,20 +60,24 @@ def _read_projections(folder, indices):
...
+        proj_array += rescale_intercept
+        proj_array /= hu_factor
 
         # Store results
-        projections.append(data_array)
+        if data_array is None:
+            # We need to load the first dataset before we know the shape
+            data_array = np.empty((len(file_names), cols, rows), dtype='float32')
+
+        data_array[i] = proj_array[:, ::-1]
```
kohr-h: Nice change!

odl/contrib/datasets/ct/mayo.py
adler-j: Flying focal spot (I guess we all know why they use "z" instead. Anyway, I'll write it out.

odl/contrib/datasets/ct/mayo.py
adler-j: It's a nightmare

adler-j:
On python 2 we're getting:

```
1.93s$ conda create -n testenv python=$TRAVIS_PYTHON_VERSION nomkl
Fetching package metadata: ....
Solving package specifications: .
Error:  Dependency missing in current linux-64 channels: 
  - nomkl -> blas * openblas
You can search for this package on anaconda.org with
    anaconda search -t conda blas * openblas
You may need to install the anaconda-client command line client with
    conda install anaconda-client
The command "conda create -n testenv python=$TRAVIS_PYTHON_VERSION nomkl" failed and exited with 1 during .
Your build has been stopped.
```

Have you seen this before?

odl/contrib/datasets/ct/mayo.py
```diff
@@ -132,16 +132,20 @@ def load_projections(folder, indices=None):
     offset_angular = np.array([d.SourceAngularPositionShift for d in datasets])
     offset_radial = np.array([d.SourceRadialDistanceShift for d in datasets])
 
+    # TODO(adler-j): Implement proper handling of flying focal spot.
+    # Currently we do not fully account for it, merely making some "first
+    # order corrections".
```
kohr-h: :+1:

kohr-h:
> Have you seen this before?

No, but it's out of our control, so it might be an annoyance but not a blocker for a PR. My assumption is that Anaconda messed up their packages for Python 2.

adler-j:
Merged then!

PR #1426, opened 2018-09-01
Fixes before release

kohr-h:
A bunch of fixes that should go in before release. Mostly doc, but also some minor fixes for the slow tests.

### DIFF
--- a/odl/phantom/transmission.py
+++ b/odl/phantom/transmission.py
@@ -96,7 +96,7 @@ def shepp_logan_ellipsoids(ndim, modified=False):
 
     References
     ----------
-    .. _Shepp-Logan phantom: en.wikipedia.org/wiki/Shepp–Logan_phantom
+    .. _Shepp-Logan phantom: https://en.wikipedia.org/wiki/Shepp-Logan_phantom
     """
     if ndim == 2:
         ellipsoids = _shepp_logan_ellipse_2d()
@@ -148,7 +148,7 @@ def shepp_logan(space, modified=False, min_pt=None, max_pt=None):
 
     References
     ----------
-    .. _Shepp-Logan phantom: en.wikipedia.org/wiki/Shepp–Logan_phantom
+    .. _Shepp-Logan phantom: https://en.wikipedia.org/wiki/Shepp-Logan_phantom
     """
     ellipsoids = shepp_logan_ellipsoids(space.ndim, modified)
     return ellipsoid_phantom(space, ellipsoids, min_pt, max_pt)
--- a/odl/phantom/transmission.py
+++ b/odl/phantom/transmission.py
@@ -1,4 +1,4 @@
-﻿# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
 #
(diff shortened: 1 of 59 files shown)


kohr-h:
Regarding the PEP8 issues, these are all false positives. There are links in docstrings that go over 79 characters, and that's not a PEP8 violation. And even if it were, it's still better to violate than to split the link with a backslash. For raw strings (r""") that won't even work.

kohr-h:
546914c fixes the Travis build problem with Python 2.7.

odl/solvers/util/steplen.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, function, tau=0.5, discount=0.01, alpha=1.0,
         ----------
         function : callable
             The cost function of the optimization problem to be solved.
-            If function is not a `Functional`, the argument `dir_derivative`
+            If function is not a `Functional`, the argument ``dir_derivative``
```
adler-j: This looks like a hanging sentence here? Obviously from before.

odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
```diff
@@ -176,10 +176,24 @@ def test_proximal_defintion(functional, stepsize):
...
+    """Objective function of the convex conjugate problem."""
     return x.inner(y) - functional(x)
 
 
+def func_convex_conj_has_call(functional):
+    """Helper to determine whether a convex conjugate can be called."""
+    f_cconj = functional.convex_conj
+    if isinstance(f_cconj, FunctionalDefaultConvexConjugate):
+        return False
+    if (
```
adler-j: Write this as 

```
if (a or
    (b and
     c)):
```

kohr-h: It's only a and b in this case, and that results in the ugly extra indent of the second line (that I've always disliked):
```python
if (isinstance(f_cconj, odl.solvers.FunctionalTranslation) and
        isinstance(f_cconj.functional, FunctionalDefaultConvexConjugate)):
    # Stuff
```
I find this version much more readable since the statements line up.

kohr-h: An alternative would be to start on the first line and just put the last `):` on an extra line.

odl/phantom/transmission.py
```diff
@@ -392,7 +392,7 @@ def transposeravel(arr):
         # Bone
         materials[image > 1.75] = 7
 
-        return space.element(materials)
+        return space.element(materials.reshape(space.shape))
```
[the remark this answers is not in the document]
kohr-h: `materials` was just a flat array.

odl/solvers/util/steplen.py
kohr-h: I'll fix it.

adler-j: I'll trust your judgement on this, but two if statements following each other is typically read as independent. Even doing elif would make it read as related without changing the results

kohr-h: I'm gonna check, but I'm not entirely positive that the two are mutually exclusive.

adler-j: Hence the use of judgement :)

kohr-h: Well, since they return immediately anyway, it doesn't matter. I'll use elif then.

odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
kohr-h: I went with the current version, but without the line break after the opening parenthesis. Hence this:
```python
if (isinstance(f_cconj, odl.solvers.FunctionalTranslation) and
    isinstance(f_cconj.functional, FunctionalDefaultConvexConjugate)
):
    # Stuff
```

kohr-h:
Merging after CI

Issue #1422, opened 2018-08-29
Allow initial_param in optimal_parameters to be a bracket


adler-j:
As suggested by @olivierverdier, we should allow not only point estimates but also brackets.

See https://github.com/odlgroup/odl/pull/1280/files/0c3804428eb929c3c078b623b6c97d5806d458a3#r199210612

PR #1428, opened 2018-09-03
ENH: allow to pass brackets to univariate solvers

olivierverdier:
Proposed fix to #1422

### DIFF
--- a/odl/contrib/param_opt/param_opt.py
+++ b/odl/contrib/param_opt/param_opt.py
@@ -15,7 +15,7 @@ __all__ = ('optimal_parameters', )
 
 
 def optimal_parameters(reconstruction, fom, phantoms, data,
-                       initial=None, multivariate=True):
+                       initial=None, univariate=False):
     r"""Find the optimal parameters for a reconstruction method.
 
     Notes
@@ -59,8 +59,8 @@ def optimal_parameters(reconstruction, fom, phantoms, data,
         Initial guess for the parameters. It is
         - a required array in the multivariate case
         - an optional pair in the univariate case.
-    multivariate : bool, optional
-        Whether to use a multivariate solver (defaults to True)
+    univariate : bool, optional
+        Whether to use a univariate solver
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


odl/contrib/param_opt/param_opt.py
```diff
@@ -55,8 +55,12 @@ def optimal_parameters(reconstruction, fom, phantoms, data,
         True images.
     data : sequence
         The data to reconstruct from.
-    initial_param : array-like, optional
-        Initial guess for the parameters, default is zero.
+    initial : array-like or pair
+        Initial guess for the parameters. It is
+        - a required array in the multivariate case
+        - an optional pair in the univariate case.
+    multivariate : bool, optional
```
adler-j: I'd prefer if we had a `univariate` option, since it naturally feels like a "special case".

odl/contrib/param_opt/param_opt.py
```diff
@@ -55,8 +55,12 @@ def optimal_parameters(reconstruction, fom, phantoms, data,
...
+        Whether to use a multivariate solver (defaults to True)
```
adler-j: We typically don't mention the default value unless it's not obvious from the function arguments

odl/contrib/param_opt/param_opt.py
```diff
@@ -72,17 +76,15 @@ def func(lam):
     # Pick resolution to fit the one used by the space
     tol = np.finfo(phantoms[0].space.dtype).resolution * 10
 
-    initial_param = np.asarray(initial_param)
-
-    # We use a faster optimizer for the one parameter case
-    if initial_param.size == 1:
-        bracket = [initial_param - tol, initial_param + tol]
+    if not multivariate:
```
adler-j: Great change overall. This gives users the power to pick themselves, a much better way to do it!

adler-j:
Seems to be some remaining pep8 stuff then we go!

olivierverdier:
What is the remaining pep8 stuff, exactly?

aringh:
It seems to be that line 42 in  odl/contrib/param_opt/test/test_param_opt.py is to long (more than 79 characters). Easiest fix is probably to reformat line 41 and below like
```
    result = odl.contrib.param_opt.optimal_parameters(
        reconstruction, fom, phantoms, data, [.9, 1.1], univariate=True)
```

olivierverdier:
I had no idea that our very good friend @pep8speaks updated its comment instead of sending a new one, so I thought I was in the clear :cry:.  Anyway, fixed now.

adler-j:
Merged. Thank you very much for the fix!

PR #1430, opened 2018-09-07
Bump to 1.0.0dev0

kohr-h:
Since 1.0.0 will be next, I use the corresponding dev version for master.

The `meta.yaml` file is also up to date now, with some improvements to the way dependencies and tests are handled.


conda/meta.yaml
```diff
@@ -1,11 +1,11 @@
 package:
     name: odl
-    version: "0.6.1"
+    version: "1.0.0"
```
adler-j: Dev?

kohr-h: Hm, at the time I didn't use it, maybe because it wasn't supported? But it makes sense for uploading dev releases to a non-official channel, or for local testing. I'll add it.

PR #1432, opened 2018-09-08
Release notes 0.7.0

kohr-h:
The upcoming release notes, plus a tiny change.


kohr-h:
I was planning to write a more human-readable summary on the GH release page and keep this as a reference in chronological order (otherwise it gets really hard to keep logical consistency without forward and cross references).
I can add that to the rst document as well.

And pulling `contrib` stuff out to its own section is a good idea.

kohr-h:
Here we go. Unless there are complaints, I'll merge soon.

PR #1434, opened 2018-09-11
MAINT: Changes to TD window

adler-j:
* Change code to follow original article
* Fix bug with default `helical_geometry`
* Rename `n_half_rot` to the more standard `n_pi`

### DIFF
--- a/odl/tomo/analytic/filtered_back_projection.py
+++ b/odl/tomo/analytic/filtered_back_projection.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Copyright 2014-2018 The ODL contributors
 #
 # This file is part of ODL.
@@ -140,7 +141,7 @@ def tam_danielson_window(ray_trafo, smoothing_width=0.05, n_pi=1):
     Physics in Medicine & Biology 4 (1998), p 1015.
     https://dx.doi.org/10.1088/0031-9155/43/4/028
 
-    [PKGT2000] Proksa R, Kohler T, Grass M, Timmer J.
+    [PKGT2000] Proksa R, Köhler T, Grass M, Timmer J.
     *The n-PI-method for helical cone-beam CT*
     IEEE Trans Med Imaging. 2000 Sep;19(9):848-63.
     https://www.ncbi.nlm.nih.gov/pubmed/11127600
--- a/odl/tomo/analytic/filtered_back_projection.py
+++ b/odl/tomo/analytic/filtered_back_projection.py
@@ -99,7 +99,7 @@ def _fbp_filter(norm_freq, filter_type, frequency_scaling):
     return filt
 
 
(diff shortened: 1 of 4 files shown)


odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -163,19 +169,18 @@ def tam_danielson_window(ray_trafo, smoothing_width=0.05, n_half_rot=1):
...
-    theta = phi * 2
+    dx_axis = dx * src_radius / (src_radius + det_radius)
 
-    # Compute lower and upper bound
-    source_to_line_distance = src_radius + src_radius * np.cos(theta)
-    scale = (src_radius + det_radius) / source_to_line_distance
+    def Vn(u):
+        return (pitch / (2 * np.pi) *
+                (1 + (u / src_radius) ** 2) *
+                (n_pi * np.pi / 2.0 - np.arctan(u / src_radius)))
```
kohr-h: Not sure I wanna go read the paper to see what this function does. Is there some intuition behind it? A comment or a docstring would help here.

adler-j: Frankly, I dunno. The old code was me with pen and paper (I didn't even know the method existed and had a name at that time), trying to draw in 3D.

Interestingly, it seems that this change does not do anything to the output of the function, but merely harmonizes it with the article.

adler-j: I can add the general idea that this is the helix that the source travels on, projected onto the detector.

kohr-h: > I can add the general idea that this is the helix that the source travels on, projected onto the detector.

That'd be a great help!

kohr-h:
> isn't this more of a BUG rather than a MAINT?

Oh I see now that one of the commits has BUG. Never mind.

kohr-h:
> SyntaxError: Non-ASCII character '\xc3' in file ...

Why don't we add `# coding: utf-8` to all files we modify? It won't last for very long.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -138,6 +139,11 @@ def tam_danielson_window(ray_trafo, smoothing_width=0.05, n_half_rot=1):
     *Exact cone beam CT with a spiral scan*.
     Physics in Medicine & Biology 4 (1998), p 1015.
     https://dx.doi.org/10.1088/0031-9155/43/4/028
+
+    [PKGT2000] Proksa R, Kohler T, Grass M, Timmer J.
```
kohr-h: Disagree with replacing "Köhler" with "Kohler", it's wrong. Please use the unicode marker instead.

adler-j: You caught me :( I'll fix

kohr-h: Don't mess with German names :wink:

Issue #645, opened 2016-10-12
Reflection padding


kohr-h:
The next version of PyWavelets will get a `'reflect'` extension mode, which is the same as `'symmetric'`, but doesn't repeat the boundary value. This is actually [how we interpret symmetric padding](https://github.com/odlgroup/odl/blob/7e88a779be5feee6381dbc0a0664cf3d7ecaabd7/odl/util/numerics.py#L333-L335). I suggest that we rename our current symmetric mode to "reflect" and add the new symmetric mode like PyWavelets does it.

adler-j:
Have you had a look at how numpy handles these names, is it the same as PyWavelets?

kohr-h:
`numpy.pad` uses the same notion as `PyWavelets`. Not so surprising since there is a certain overlap of developers.

PR #1436, opened 2018-09-17
simplify WaveletTransform PyWavelets backend and add axes support

grlee77:
Most of this PR involves removing functions in `odl/trafos/backends/pywt_bindings.py` that are no longer necessary with PyWavelets 1.0. Functions to ravel/unravel coefficients were added to PyWavelets itself (see PyWavelets/pywt#341) and n-dimensional multilevel decomposition/reconstruction has been present via `wavedecn`/`waverecn` for a few years now.

The only new feature I added to the `WaveletTransform` operator is the ability to specify which `axes` the transform is being applied over. Previously it was always assumed that the DWT should be applied along all axes. Test cases have been added for this. If it is preferred, I can move the two commits related to this enhancement to a separate PR. It would also be fairly simple to allow per-axis choice of wavelet, but I haven't done that here.

Please also check the three new padding modes listed within `PAD_MODES_ODL2PYWT` to see if the naming on the ODL side is as desired.

All functions being tested within `pywt_bindings_test.py` were removed by this PR, so it may be preferable to just remove that file.  I added a few very basic tests of the remaining utilities, but I'm not sure they are really needed. 

These changes do require bumping the PyWavelets requirement to 1.0, which was only recently released, so I understand if you want to wait a bit before making that change. As an optional requirement, it is probably not a big deal to increase the version as it can be easily obtained via either pip or conda.


odl/trafos/wavelet.py
```diff
@@ -438,17 +467,42 @@ def __init__(self, range, wavelet, nlevels=None, pad_mode='constant',
...
-                wavelet=self.pywt_wavelet, mode=self.pywt_pad_mode)
+            coeffs = pywt.unravel_coeffs(coeffs,
+                                         coeff_slices=self._coeff_slices,
+                                         coeff_shapes=self._coeff_shapes,
+                                         output_format='wavedecn')
+            recon = pywt.waverecn(
+                coeffs, wavelet=self.pywt_wavelet, mode=self.pywt_pad_mode,
+                axes=self.axes)
+            recon_shape = self.range.shape
+            if recon.shape != recon_shape:
```
adler-j: This could use some information describing what happens

odl/trafos/wavelet.py
```diff
@@ -101,13 +100,29 @@ def __init__(self, space, wavelet, nlevels, variant, pad_mode='constant',
...
+        if self.impl not in _SUPPORTED_WAVELET_IMPLS:
+            raise ValueError("`impl` '{}' not supported".format(impl_in))
+
+        if axes is None:
+            axes = tuple(range(space.ndim))
+        elif np.isscalar(axes):
+            axes = (axes, )
+        elif len(axes) > space.ndim:
+                raise ValueError("Too many axes.")
+        self.axes = tuple(axes)
```
adler-j: No checks that axes is "reasonable"? E.g. contains proper integers.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -37,8 +31,9 @@
                       'order0': 'constant',
                       'order1': 'smooth',
                       'pywt_periodic': 'periodization',
-                      # Upcoming version of Pywavelets adds this
-                      # 'reflect': 'reflect'
+                      'reflect': 'reflect',
+                      'antireflect': 'antireflect',
```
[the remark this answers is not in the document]
grlee77: Maybe put the following link in a comment?
https://pywavelets.readthedocs.io/en/latest/ref/signal-extension-modes.html
There is a figure there that should make it pretty clear. There is also a table listing the corresponding names in Matlab's wavelet toolbox.

odl/trafos/wavelet.py
grlee77: Sure. Due to the decimation/upsampling by two, it is possible for signals with odd shape to have have a shape rounded up to the next even size after reconstruction:

e.g. in 1D, downsampling by two of a length 7 or 8 signal gives a length 4 signal in each band. Upon reconstruction, upsampling by two results in a length 8 signal. The extra sample needs to be discarded in the case that the original signal was only length 7.

There same behavior was also in the old implementation:
https://github.com/odlgroup/odl/blob/b06a6a2540ce398184d4adbed964ac0bad315de9/odl/trafos/backends/pywt_bindings.py#L591-L607

odl/trafos/wavelet.py
```diff
@@ -438,17 +467,42 @@ def __init__(self, range, wavelet, nlevels=None, pad_mode='constant',
...
             space=range, wavelet=wavelet, variant='inverse', nlevels=nlevels,
-            pad_mode=pad_mode, pad_const=pad_const, impl=impl)
+            pad_mode=pad_mode, pad_const=pad_const, impl=impl, axes=axes,
+            slices=slices, shapes=shapes)
 
     def _call(self, coeffs):
         """Return the inverse wavelet transform of ``coeffs``."""
+        if self._coeff_slices is None:
+            raise ValueError(
+                "self._coeff_slices not initialized.  Run the forward "
```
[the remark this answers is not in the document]
grlee77: I agree this is not ideal (however, in practice I think it is unlikely one would start with coefficients generated elsewhere and apply the inverse transform without having first done the forward transform). 

A simple way to fix this would be to just run a dummy transform during  `__init__`, but that has some computational overhead. Another alternative is to make a small utility function that can generate the appropriate slices without actually running the transform. I don't think there is a functio n for that in PyWavelets, but it shouldn't be too hard to implement if you prefer that route.

odl/trafos/wavelet.py
grlee77: Duplicate axes are also not allowed. Within `__call__` the axes will be checked by `wavedecn`. Would you prefer to replicate the checks highlighted at the link below within the `__init__` function here?:
https://github.com/PyWavelets/pywt/blob/53f9cdba2c7df4a080fb1a5fe81143a4190f898f/pywt/_multilevel.py#L334-L345

grlee77:
> My main question is: what happens to the range when axis is used? This should be documented

For example if I have an RGB image of shape (nx, ny, 3) and axes = (0, 1) this is just a 2D wavelet transform applied to each color channel. At the first transform level, neglecting potential boundary extension, each coefficient array has shape (nx/2, ny/2, 3) (no decimation was performed along non-transformed axes).

In other words it is just multiple len(axes)-dimensional transforms looped over the non-transformed axes.

Is that helpful?

grlee77:
All of the wavelets in PyWavelets are just 1D finite-impulse response filters so they can be applied along any dimension in a separable manner. So, using axes in the DWT is very similar to using axes in the FFT (e.g. numpy.fft.fftn).

There is also a "fully separable" variant of `wavedecn` implemented in PyWavelets 1.0. You can see the difference in the figure at the bottom of the following page:
https://pywavelets.readthedocs.io/en/latest/ref/2d-decompositions-overview.html

A potential future enhancement would be to add support for `fswavedecn`/`fswaverecn` variant.

grlee77:
Another case aside from multi-channel/multi-spectral data where axes support may be useful is for use with spatio-temporal data where one may want to use a different sparsifying transform in the temporal dimension than on the spatial ones.

grlee77:
The doctest failure on Python 2.7 seems to be a subtle issue that should probably be fixed in PyWavelets (see PyWavelets/pywt#427). 

Which order the various detail coefficients are stacked should not matter from a user perspective and will not prevent correct round-trip forward/inverse transforms, but could break things such as this doctest that assume a particular coefficient order in the raveled output. Most likely I will make a 1.0.1 release of PyWavelets in the near future, but for now we may need to not explicitly print the raveled coefficients in the doctest.

kohr-h:
> For example if I have an RGB image of shape (nx, ny, 3) and axes = (0, 1) this is just a 2D wavelet transform applied to each color channel. At the first transform level, neglecting potential boundary extension, each coefficient array has shape (nx/2, ny/2, 3) (no decimation was performed along non-transformed axes).
> 
> In other words it is just multiple len(axes)-dimensional transforms looped over the non-transformed axes.
> 
> Is that helpful?

I assume what @adler-j meant was that the docstring(s) should say a few words about what happens if you have an `axis` argument not equal to `None`. I agree, and I don't think it has to be more than one sentence. What is noteworthy is (1) that the decimation will only affect the included axes, (2) that the coefficient arrays will all have the excluded axes remain intact, and maybe (3) the order of axes if there's anything interesting to say (I'd expect no, since the order shouldn't change).

odl/trafos/backends/pywt_bindings.py
```diff
@@ -63,749 +63,66 @@ def pywt_pad_mode(pad_mode, pad_const=0):
...
-
-        Here, ``shape_aN`` is the shape of the N-th level approximation
-        coefficient array, and ``shape_Di`` is the shape of all
-        i-th level detail coefficients.
-
-    See Also
-    --------
-    pywt.dwt_coeff_len : function used to determine coefficient sizes at
-        each scaling level
+    coeff_slices : list
```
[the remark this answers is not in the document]
grlee77: I don't have a problem with that, except that a list is consistent with the return type from `pywt.ravel_coeffs`. I don't think I will change it on the PyWavelets side at this point for reasons of backwards compatibility.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -24,21 +25,20 @@
...
-           'pywt_wavelet', 'pywt_pad_mode', 'pywt_coeff_shapes',
-           'pywt_flat_coeff_size', 'pywt_max_nlevels',
-           'pywt_flat_array_from_coeffs', 'pywt_coeffs_from_flat_array',
-           'pywt_single_level_decomp', 'pywt_single_level_recon',
-           'pywt_multi_level_decomp', 'pywt_multi_level_recon')
+           'pywt_wavelet', 'pywt_pad_mode', 'precompute_raveled_slices')
 
 
+# A clear illustration of all of these padding modes is available at:
+# https://pywavelets.readthedocs.io/en/latest/ref/signal-extension-modes.html
```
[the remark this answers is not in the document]
grlee77: Okay. There is also some documentation in the init of `WaveletTransformBase` that I neglected to add the new edge modes to. 

I also just noticed that there is an error in the description of the mode `symmetric`.  In the text [here](https://github.com/odlgroup/odl/blob/553f9fd8eb619a248baefb995550b6e399a7b845/odl/trafos/wavelet.py#L81) it says that symmetric does not duplicate the outermost values, however in `PyWavelets` it does. The mode `reflect` in PyWavelets is the one that has the described behaviour. Should I change the dictionary so that the ODL name `symmetric` points to PyWavelets `reflect` for consistency with ODL convention or should I just fix the description in the docstring? If the former, what would we name the PyWavelets `symmetric` mode in that case? Matlab refers to these two cases as whole-sample symmetric (`symw`) and half-sample symmetric (`symh`).

odl/trafos/backends/pywt_bindings.py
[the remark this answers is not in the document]
kohr-h: That's fine then.

odl/trafos/backends/pywt_bindings.py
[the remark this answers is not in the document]
kohr-h: That was error on our side, see also #645. If you could fix that, it'd be great.

grlee77:
I added several commits to address the comments from the first round of reviews.

grlee77:
The rendering of the updated `mode` documentation in the wavelet operators looks like:

__init__mode

kohr-h:
> The rendering of the updated `mode` documentation in the wavelet operators looks like:

Perfect!



> I added several commits to address the comments from the first round of reviews.

I'm happy with the current state (apart from the doctest failure on Py2, which is due to random dict order). How about you @adler-j?

odl/trafos/wavelet.py
```diff
@@ -285,26 +373,63 @@ def __init__(self, domain, wavelet, nlevels=None, pad_mode='constant',
...
+        ...     domain=space, nlevels=2, wavelet='db2', pad_mode='symmetric')
+        >>> decomp = wavelet_trafo(np.ones(space.shape))
+        >>> decomp.shape
+        (387,)
+
+        References
+        ----------
+        .. _signal extension modes:
+           https://pywavelets.readthedocs.io/en/latest/ref/signal-extension-\
+    modes.html
```
kohr-h: Last comment: this link is likely broken due to the whitespace at the start of the second line. Best is to just not break the line in the link (we abandoned that elsewhere as well), let the line go over the PEP8 limit and ignore the pep8speaks complaint.

odl/trafos/backends/pywt_bindings.py
```diff
@@ -52,760 +52,101 @@ def pywt_wavelet(wavelet):
...
-modes.html
+    >>> import pywt
+    >>> data_shape = (64, 64)
+    >>> coeff_shapes = pywt.wavedecn_shapes(data_shape, wavelet='db2', level=3,
+    ...                                     mode='periodization')
+    >>> coeff_slices = precompute_raveled_slices(coeff_shapes)
+    >>> print(coeff_slices[0])  # approximation coefficients
+    slice(None, 64, None)
+    >>> print(coeff_slices[-1])  # first level detail coefficients
+    {'ad': slice(1024, 2048, None), 'da': slice(2048, 3072, None), 'dd': slice(3072, 4096, None)}
```
kohr-h: Maybe
```python
>>> d1_slices = coeff_slices[-1]  # first level detail coefficients
>>> d1_slices['ad'], d1_slices['da'], d1_slices['dd']
(slice(1024, 2048, None), slice(2048, 3072, None), slice(3072, 4096, None))
```

kohr-h: But the version after your edit is also fine. This function is somewhat low-level anyway, so the target audience here is more the developer than the user.

kohr-h:
I think all points from @adler-j's reviews are addressed, so I'll go ahead and merge.

kohr-h:
Thanks @grlee77!

PR #1439, opened 2018-09-24
MAINT: add Python 3.7 to identifiers in setup.py

kohr-h:
Trivial change


kohr-h:
Merging now since it's trivial

PR #1444, opened 2018-10-26
fix bug in prox of simplex constraint

mehrhardt:
The current version of the `IndicatorSimplex` could not evaluate its `prox`. Not sure how the bug was introduced and not flagged by tests. In any case, this is now fixed with this PR.


adler-j:
Merged! Great fix.

PR #1446, opened 2018-11-17
ENH: Add cuboid_phantom

adler-j:


adler-j:
Could this get a review please?

kohr-h:
I'll have a look

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+            offset_points = [vec * (xi - x0i)[..., None]
+                             for xi, vec, x0i in zip(subgrid, mat.T, [x0, y0])]
+            rotated = offset_points[0] + offset_points[1]
+            np.square(rotated, out=rotated)
+            max_distance = np.maximum(scales[0] * rotated[..., 0],
+                                      scales[1] * rotated[..., 1])
+        else:
+            # Calculate the points that could possibly be inside the volume
+            max_radius = np.sqrt([a_squared, b_squared])
+            idx, shapes = _getshapes_2d(center, max_radius, space.shape)
```
kohr-h: Why do you have to do this for non-rotated rectangles? Can't you just get the min and max per axis and use those directly as bounding boxes? In other words, can't you directly construct `slice` objects from the min and max and use those for indexing?

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+            subgrid = [g[idi] for g, idi in zip(grid, shapes)]
+            squared_dist = [ai * (xi - x0i) ** 2
+                            for xi, ai, x0i in zip(subgrid,
+                                                   scales,
+                                                   [x0, y0])]
+
+            # Parentheses to get best order for broadcasting
+            max_distance = np.maximum(squared_dist[0], squared_dist[1])
+
+        # Find the points within the ellipse
```
kohr-h: rectangle?

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+        inside = max_distance <= 1
+
+        # Add the ellipse intensity to those points
```
kohr-h: ditto

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+        If ``space`` is 2-dimensional, each row should contain the entries ::
+
+            'value',
+            'axis_1', 'axis_2',
+            'center_x', 'center_y',
+            'rotation'
+
+        The provided cuboids need to be specified relative to the
+        reference rectangle ``[-1, -1] x [1, 1]``.
+        The angles are to be given in radians.
```
kohr-h: How about 3D?

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+            new_max_pt = space.max_pt + (min_pt - space.min_pt)
+
+        Providing both results in a scaled version of the phantom.
+
+    Notes
+    -----
+    The phantom is created by adding the values of each cuboid. The
+    cuboid are defined by a center point
+    ``(center_x, center_y, [center_z])``, the lengths of its principial
+    axes ``(axis_1, axis_2, [axis_2])``, and a rotation angle ``rotation``
```
kohr-h: axis_3

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+
+    >>> space = odl.uniform_discr([-1, -1], [1, 1], [5, 5])
+    >>> cuboids = [[1.0, 1.0, 1.0, 0.0, 0.0, 0.0],
+    ...             [1.0, 0.6, 0.6, 0.0, 0.0, 0.0]]
+    >>> print(cuboid_phantom(space, cuboids))
+    [[ 0.,  0.,  1.,  0.,  0.],
+     [ 0.,  1.,  2.,  1.,  0.],
+     [ 1.,  2.,  2.,  2.,  1.],
+     [ 0.,  1.,  2.,  1.,  0.],
+     [ 0.,  0.,  1.,  0.,  0.]]
```
kohr-h: Something is not right with this example. It doesn't look like a cuboid to me.

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+def cuboid_phantom(space, cuboids, min_pt=None, max_pt=None):
+    """Return a phantom given by cuboids.
+
+    Parameters
+    ----------
+    space : `DiscreteLp`
+        Space in which the phantom should be created, must be 2- or
+        3-dimensional. If ``space.shape`` is 1 in an axis, a corresponding
+        slice of the phantom is created (instead of squashing the whole
+        phantom into the slice).
```
kohr-h: 3D is not supported, so this needs an update

odl/phantom/geometric.py
kohr-h: But since 3D isn't supported, remove 3rd axis stuff.

odl/phantom/geometric.py
```diff
@@ -703,6 +704,232 @@ def ellipsoid_phantom(space, ellipsoids, min_pt=None, max_pt=None):
...
+            # a boundary
+            for i in range(space.ndim):
+                if max_pt[i] in space.partition.cell_boundary_vecs[i]:
+                    snapped_max_pt[i] = max_pt[i]
+
+        tmp_space = uniform_discr_fromdiscr(
+            space, min_pt=snapped_min_pt, max_pt=snapped_max_pt,
+            cell_sides=space.cell_sides)
+
+        tmp_phantom = _phantom(tmp_space, ellipsoids)
```
kohr-h: This branch has obviously never been run, since it cannot work.

odl/phantom/geometric.py
kohr-h: OTOH, I can see why you want to use the same kind of logic. Maybe it's okay.

adler-j: I'll just leave this as is since the other optimization is relatively minor for most cases. This would only give a notable improvement in the case of slightly rotated and very elongated rectangles, which I'll just leave for now.

odl/phantom/geometric.py
adler-j: 3D is not currently supported, I'll update the rest to reflect this

odl/phantom/geometric.py
```diff
@@ -876,6 +1101,12 @@ def sigmoid(val):
     # Indicate proj axis 3D
     indicate_proj_axis(space).show('indicate_proj_axis 3d')
 
+    # cuboid phantom 2D
+    space = odl.uniform_discr([-1, -1], [1, 1], [300, 300])
+    ellipses = [[1.0, 0.6, 0.6, 0.0, 0.0, 1.5],
+                [1.0, 0.1, 0.3, 0.0, 0.0, 0.3]]
+    cuboid_phantom(space, ellipses).show('cuboid phantom 2d')
```
kohr-h: Use `cuboids` or `cubes` instead of `ellipses`.

PR #1449, opened 2019-01-15
Geometry for circular detectors

JevgenijaAksjonova:
FanFlatGeometry was renamed to FanGeometry and modified. An additional parameter det_curve_radius was added to allow the usage of curved detectors. A new class CircularDetector was implemented. It can be viewed as re-implementation of existing class CircleSectionDetector with different input arguments and more readable formulas. Next, it is planned to implement new classes CylindricalDetector and SphericalDetector and modify the class ConeFlatGeometry to allow for the usage of curved detectors.


sbanert:
There are still some issues remaining, @pep8speaks updates its post accordingly.

JevgenijaAksjonova:
Thanks, I did not notice. These problem are not in my code though (except for the last one). Its about splitting the line before or after an operator, it appears that the standards have change, for instance see https://github.com/PyCQA/pycodestyle/issues/513. Should I fix it so that it agrees with the new standards?

adler-j:
https://github.com/PyCQA/pycodestyle/issues/498 seems to be the latest one, I guess we should update according to this, could you help out with that @JevgenijaAksjonova ?

JevgenijaAksjonova:
Sure

JevgenijaAksjonova:
now it complains about the line break before =(

JevgenijaAksjonova:
> Finally I found the time to review this. Sorry for the delay.
> 
> The PR looks really good, and I've only got a bunch of minor comments.

Thanks for the review! I have fixed most of the issues, however I have some questions/comments, which I will write next to each issue

odl/tomo/geometry/conebeam.py
```diff
@@ -237,12 +253,15 @@ def frommatrix(cls, apart, dpart, src_radius, det_radius, init_matrix,
             determine the new vectors. If present, the third column acts
             as a translation after the initial transformation.
             The resulting ``det_axis_init`` will be normalized.
+        det_curve_radius : nonnegative float, optional
+            Radius of the detector curvature.
+            If ``None``, flat detector is used, otherwise must be positive.
```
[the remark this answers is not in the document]
JevgenijaAksjonova: I have removed the redundant >0 check, but maybe it is still good to have it documented here

odl/tomo/geometry/detector.py
```diff
@@ -897,6 +897,272 @@ def __str__(self):
         return repr(self)
 
 
+class CircularDetector(Detector):
```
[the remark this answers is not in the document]
JevgenijaAksjonova: There is a class called CircleSectionDetector, I did not dare to remove it, although I checked that it is not used anywhere on ODL. This class basically implements the same thing but in a different way and with different input parameters. Regarding the name, as the next step I am planning to implement CylindricalDetector and SphericalDetector in a similar way, so I thought the simply CircularDetector is more consistent with the other names

odl/tomo/geometry/conebeam.py
```diff
@@ -544,12 +568,21 @@ def __getitem__(self, indices):
...
+                           src_to_det_init=self.src_to_det_init,
+                           det_axis_init=self._det_axis_init_arg,
+                           translation=self.translation)
+
+
+class FanFlatGeometry(FanGeometry):
+
+    """Fan beam (2d cone beam) geometry with flat 1d detector.
+
+    FanGeometry with the default det_curve_radius = None """
```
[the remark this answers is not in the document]
JevgenijaAksjonova: I copied the general description of a class, however, most of the documentation is under the functions, for instance,  under the __init__(). Should I just copy it as well? I don't know how the documentation  process work, to be honest

odl/tomo/geometry/conebeam.py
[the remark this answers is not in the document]
kohr-h: Yes, absolutely.

odl/tomo/geometry/conebeam.py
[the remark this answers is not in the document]
kohr-h: Oh dear. I remember we had a discussion some time ago whether we should put the documentation of a class (constructor parameters, notes, etc.) in the class docstring or in the docstring of `__init__`. I was arguing for `__init__` back then, but this case is actually a good reason against it.

So yes, please do copy over the `__init__` docstring as well (without the first line).

As for the documentation process, the `sphinx` tool will collect the docstrings for classes and their methods and generate the HTML accordingly. The important detail here is that it will not add or somehow merge with the documentation of parent classes, but only look at the class itself.

odl/tomo/geometry/detector.py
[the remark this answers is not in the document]
kohr-h: Ah, that's why the name sounded awfully familiar to me :-)

You're right, it's not used currently. I guess the main difference to your version is that your angle is positive and 0 is at the leftmost pixel, whereas the existing one has positive and negative angles.
Anyhow, please remove the old one, we don't need that duplication.

odl/tomo/geometry/conebeam.py
```diff
@@ -119,17 +119,31 @@ def __init__(self, apart, dpart, src_radius, det_radius,
         >>> geom.det_axis_init
         array([ 1.,  0.])
 
+        Specifying curvature of the detector:
+
+        >>> apart = odl.uniform_partition(0, 2 * np.pi, 10)
+        >>> dpart = odl.uniform_partition(-np.pi / 2, np.pi / 2, 10)
+        >>> geom = FanBeamGeometry(apart, dpart, src_radius=1, det_radius=5,
+        ...                        det_curve_radius = 10)
```
kohr-h: In keyword arguments, no whitespace around `=`

odl/tomo/geometry/conebeam.py
```diff
@@ -329,7 +339,10 @@ def det_radius(self):
     @property
     def det_curve_radius(self):
         """Detector curve radius of this geometry."""
-        return self.__det_curve_radius
+        try:
+            return self.detector.radius
+        except AttributeError:
+            return None
```
kohr-h: You can write this a bit more compactly as `return getattr(self.detector, 'radius', None)`. It's a bit easier on the eye.

odl/tomo/geometry/conebeam.py
```diff
@@ -568,21 +581,34 @@ def __getitem__(self, indices):
         apart = part.byaxis[0]
         dpart = part.byaxis[1]
 
-        return FanGeometry(apart, dpart,
+        return FanBeamGeometry(apart, dpart,
```
kohr-h: I guess pep8speaks will complain about the indentation of the next lines

odl/tomo/geometry/detector.py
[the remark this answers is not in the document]
JevgenijaAksjonova: No, I think the functionality is the same at the end, and it is as similar as possible to flat detector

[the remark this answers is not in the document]
adler-j: It should be noted that the standard notation in ODL sets angle (or pixel) 0 to be the center of the detector. See https://odlgroup.github.io/odl/guide/geometry_guide.html for a quite good introduction. We should try to follow that style (and also once this is a bit more mature, add it to that documentation).

adler-j:
BTW @JevgenijaAksjonova you might want to have a look at e.g. https://docs.scipy.org/doc/numpy-1.15.0/dev/gitwash/development_workflow.html#writing-the-commit-message to get inspiration on how to write commit messages. It's not really a major worry but it helps keep the code organised.

odl/tomo/geometry/conebeam.py
```diff
@@ -114,17 +119,31 @@ def __init__(self, apart, dpart, src_radius, det_radius,
...
+        >>> dpart = odl.uniform_partition(-np.pi / 2, np.pi / 2, 10)
+        >>> geom = FanBeamGeometry(apart, dpart, src_radius=1, det_radius=5,
+        ...                        det_curve_radius=10)
+        >>> geom.src_position(0)
+        array([ 0., -1.])
+        >>> geom.det_refpoint(0)
+        array([ 0.,  5.])
+        >>> np.round(geom.det_point_position(0, np.pi / 6), 2)
+        # (0, 5) + 10 * (sin(pi/6), cos(pi/6) - 1)
+        array([ 5.  ,  3.66])
```
kohr-h: Does this test pass? The comment on the first output line may make it fail. If so, you need to put it on an input line, i.e., write `>>> # Your comment`

odl/tomo/geometry/conebeam.py
```diff
@@ -187,10 +206,17 @@ def __init__(self, apart, dpart, src_radius, det_radius,
         self.__src_to_det_init = src_to_det_init
         # `check_bounds` is needed for both detector and geometry
         check_bounds = kwargs.get('check_bounds', True)
-        detector = Flat1dDetector(dpart, axis=det_axis_init,
-                                  check_bounds=check_bounds)
+        if det_curve_radius is None:
+            detector = Flat1dDetector(dpart, axis=det_axis_init,
+                                      check_bounds=check_bounds)
+        else:
+            det_curve_radius = float(det_curve_radius)
```
kohr-h: This will be done by `CircularDetector`, so it's okay to just pass on the parameter.

odl/tomo/geometry/conebeam.py
```diff
@@ -544,14 +578,126 @@ def __getitem__(self, indices):
...
+    The initial detector axis can also be set explicitly:
+
+    >>> geom = FanFlatGeometry(
+    ...     apart, dpart, src_radius=1, det_radius=5,
+    ...     src_to_det_init=(1, 0), det_axis_init=(0, 1))
+    >>> np.allclose(geom.src_to_det_init, e_x)
+    True
+    >>> np.allclose(geom.det_axis_init, e_y)
+    True
+    """
```
kohr-h: @adler-j What do you think? Should we just drop this class and use `FanBeamGeometry` only? I'd be okay with that, given that this class doesn't really do anything.

odl/tomo/geometry/detector.py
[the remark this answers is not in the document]
kohr-h: Good call, I forgot about that although I wrote that myself :roll_eyes:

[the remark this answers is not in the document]
JevgenijaAksjonova: I think it is in the center. Maybe I am missing something, why do you think it is not?

kohr-h:
> BTW @JevgenijaAksjonova you might want to have a look at e.g. https://docs.scipy.org/doc/numpy-1.15.0/dev/gitwash/development_workflow.html#writing-the-commit-message to get inspiration on how to write commit messages. It's not really a major worry but it helps keep the code organised.

Less of a worry now after #1471

odl/tomo/geometry/conebeam.py
kohr-h: On my machine, this test indeed fails:
```
132         >>> np.round(geom.det_point_position(0, np.pi / 6), 2)
Expected:
    # (0, 5) + 10 * (sin(pi/6), cos(pi/6) - 1)
    array([ 5.  ,  3.66])
Got:
    array([ 5.  ,  3.66])
```
Together with 2 others:
```
/home/hkohr/git/odl/odl/tomo/geometry/conebeam.py:132: DocTestFailure
________________________ [doctest] odl.tomo.geometry.detector.CircularDetector.surface ________________________
730             shape ``(2,)``, otherwise ``param.shape + (2,)``.
731 
732         Examples
733         --------
734         The method works with a single parameter, resulting in a single
735         vector:
736 
737         >>> part = odl.uniform_partition(-np.pi / 2, np.pi / 2, 10)
738         >>> det = CircularDetector(part, axis=[0, 1], radius=2)
739         >>> np.allclose(det.surface(-np.pi / 2), [-2, -2])
Expected:
    True
Got:
    False

/home/hkohr/git/odl/odl/tomo/geometry/detector.py:739: DocTestFailure
_____________________ [doctest] odl.tomo.geometry.detector.CircularDetector.surface_deriv _____________________
792         --------
793         surface
794 
795         Examples
796         --------
797         The method works with a single parameter, resulting in a single
798         vector:
799 
800         >>> part = odl.uniform_partition(-np.pi / 2, np.pi / 2, 10)
801         >>> det = CircularDetector(part, radius=2)
UNEXPECTED EXCEPTION: TypeError("__init__() missing 1 required positional argument: 'axis'",)

  File "/home/hkohr/miniconda/envs/odl_py3/lib/python3.6/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)

  File "<doctest odl.tomo.geometry.detector.CircularDetector.surface_deriv[1]>", line 1, in <module>

TypeError: __init__() missing 1 required positional argument: 'axis'

/home/hkohr/git/odl/odl/tomo/geometry/detector.py:801: UnexpectedException
```

JevgenijaAksjonova: How do you run the tests?

JevgenijaAksjonova: Ok, I guess its doctests...

kohr-h: Yes, I cloned your branch and ran the tests. If you have `pytest` installed you can run the doctests in a specific module or a directory by
```
pytest odl/tomo/geometry/detector.py  # runs doctests in that module
pytest odl/tomo/geometry/  # runs doctests in all modules in that subdirectory
```

JevgenijaAksjonova: Oh, I see, thanks. It should be fixed now

JevgenijaAksjonova:
So, are there any other changes necessary?

odl/tomo/geometry/conebeam.py
kohr-h: Looking at this again, I think this class isn't needed any more. It shouldn't be lying around just as a name alias.

Can you remove it @JevgenijaAksjonova? That would mean to replace `FanFlatGeometry` with `FanBeamGeometry` everywhere, and removing the duplicate unit tests.

odl/tomo/geometry/detector.py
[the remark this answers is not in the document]
kohr-h: > 0 is at the leftmost pixel

That was what I wrote above. Is that wrong? Is 0 the center pixel? If yes, then we're good.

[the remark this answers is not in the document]
JevgenijaAksjonova: Yes, I had a different example before, but now I changed it to 

>>> part = odl.uniform_partition(-np.pi / 2, np.pi / 2, 10)
>>> det = CircularDetector(part, axis=[1, 0], radius=2)
>>> np.allclose(det.surface(0), [0, 0])

odl/tomo/geometry/conebeam.py
JevgenijaAksjonova: I am afraid that not all functionality that works for FanFlatGeometry works in general case, for instance FBP. In this case, I think we should live at as is for now.

odl/tomo/geometry/detector.py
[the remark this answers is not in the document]
kohr-h: Okay, very good.

odl/tomo/geometry/conebeam.py
kohr-h: That's true, but for FBP we could replace the `isinstance(geometry, FanFlatGeometry)` checks with `isinstance(geometry, FanBeamGeometry) and geometry.det_radius is None`.

JevgenijaAksjonova: Ok, I see

JevgenijaAksjonova: Regarding the tests: I am not sure what to do with test_fanflat_props(shift) and test_fancurved_props(shift). There are some lines which are duplicates, but in general these are two different scenarios.

JevgenijaAksjonova: Well I guess can leave just the test_fancurved_props(), but call it test_fanbeam_props() instead

kohr-h:
I've created a pull request into your branch that parametrizes the fan beam test to cover both cases of flat and curved detectors. You can just merge it, and then we can go ahead with this.

odl/tomo/backends/astra_cuda.py
```diff
@@ -372,7 +372,8 @@ def astra_cuda_bp_scaling_factor(proj_space, reco_space, geometry):
         if isinstance(geometry, Parallel2dGeometry):
             # Scales with 1 / cell_volume
             scaling_factor *= float(reco_space.cell_volume)
-        elif isinstance(geometry, FanFlatGeometry):
+        elif isinstance(geometry, FanBeamGeometry)\
+                and geometry.det_curve_radius is None:
```
kohr-h: Last little thing: Please use parentheses instead of backslash to continue the line.

odl/tomo/backends/astra_cuda.py
```diff
@@ -398,7 +399,8 @@ def astra_cuda_bp_scaling_factor(proj_space, reco_space, geometry):
...
```
kohr-h: Same here

JevgenijaAksjonova:
So, is everything OK? Can the code be merged?

odl/tomo/geometry/conebeam.py
```diff
@@ -307,6 +334,11 @@ def det_radius(self):
         """Detector circle radius of this geometry."""
         return self.__det_radius
 
+    @property
+    def det_curve_radius(self):
```
kohr-h: Last-minute comment, but part of API, thus important. I think it's better to use "curvature" instead of "curve" since the description of `det_radius` is described as "Detector circle radius", and this one as "Detector curve radius". That can be very confusing since a circle is a curve.

So please change `det_curve_radius` everywhere to `det_curvature_radius`. It is a bit long but much clearer, and also valid for 2D detectors.

After that I'll merge, I promise :crossed_fingers:

kohr-h:
Perfect, thanks a lot @JevgenijaAksjonova!

JevgenijaAksjonova:
Great! Thank you for the code review!

PR #1455, opened 2019-02-01
ENH: Make filter type strings case-insensitive. Closes #1454.

### LINKED ISSUE #1454: Make fbp_op filter_type case insensitive
We currently distinguish between e.g. `'Hann'` and `'hann'`, this is simply irritating and should be fixed.

sbanert:


sbanert:
I suspect that something is wrong with @pep8speaks.

adler-j:
Basically they changed a standard and now anyone who changes a file gets yelled at. Perhaps we should just make a PR that fixes this globally.

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -79,16 +80,16 @@ def _fbp_filter(norm_freq, filter_type, frequency_scaling):
     """
     if callable(filter_type):
         filt = filter_type(norm_freq)
-    elif filter_type == 'Ram-Lak':
+    elif filter_type.upper() == 'RAM-LAK':
```
adler-j: prefer `lower`. Uppercase is usually reserved for global constants and could be confusing.

sbanert: Ok. I thought about these strings more like global constants (or an enumeration type). But you're the boss. :smiley:

sbanert:
Regarding @pep8speaks: It complains about both breaks after a binary operator (W504) and breaks before an operator (W503). I think that the latter one is a bug.

odl/tomo/analytic/filtered_back_projection.py
adler-j: ODL is a collaborative effort, I try avoiding being a boss the best I can :)

Anyway, I guess the above explanation was not perfect honestly, but I've always seen people use `str.lower()` to normalize capitalization, I tried googling it but didn't see any official guideline on it so perhaps it is just convention.

adler-j:
Feel free to merge

kohr-h:
> Regarding @pep8speaks: It complains about both breaks after a binary operator (W504) and breaks before an operator (W503). I think that the latter one is a bug.

The W503 complains about the "current" way of breaking lines, right? Could you add it to the [`.pep8speaks.yml` ignore section](https://github.com/odlgroup/odl/blob/master/.pep8speaks.yml#L15) then please?

odl/tomo/analytic/filtered_back_projection.py
```diff
@@ -57,6 +57,7 @@ def _fbp_filter(norm_freq, filter_type, frequency_scaling):
                    callable}
         The type of filter to be used.
         If a string is given, use one of the standard filters with that name.
+        Strings are case-insensitive.
```
kohr-h: We never say that explicitly.

sbanert:
Everything done and rebased.

PR #1456, opened 2019-02-04
TST: rename pytest_plugins -> pytest_config

kohr-h:
The name `pytest_plugins` is a reserved name used internally by pytest, and we accidentally mess them up. See https://github.com/pytest-dev/pytest/issues/3899


kohr-h:
All the remaining complaints are for too long lines in `long_description`, I won't fix those.

PR #1461, opened 2019-02-09
Issue 1460  pytest params

kohr-h:


kohr-h:
I'll fix the PEP8 issues.

odl/test/largescale/tomo/analytic_slow_test.py
```diff
@@ -40,14 +40,22 @@
...
+projector_ids = [
+    " geom='{}' - impl='{}' - angles='{}' ".format(*p.values[0].split())
+    for p in projectors
+]
 
 # bug in pytest (ignores pytestmark) forces us to do this this
 largescale = " or not pytest.config.getoption('--largescale')"
-projectors = [pytest.mark.skipif(p.args[0] + largescale, p.args[1])
-              for p in projectors]
+projectors = [
```
adler-j: Is this truly needed? It's honestly hard to read.

odl/test/largescale/tomo/ray_transform_slow_test.py
```diff
@@ -49,14 +48,23 @@
...
+    " geom='{}' - impl='{}' - angles='{}' ".format(*p.values[0].split())
+    for p in projectors
+]
 
 
 # bug in pytest (ignores pytestmark) forces us to do this this
 largescale = " or not pytest.config.getoption('--largescale')"
-projectors = [pytest.mark.skipif(p.args[0] + largescale, p.args[1])
-              for p in projectors]
+projectors = [
```
adler-j: Same as above here

odl/test/largescale/tomo/analytic_slow_test.py
kohr-h: It ain't pretty, I agree. Maybe I'll just make a little function to do the dirty work.

Formatting-wise, I'm currently going with what the [`black`](https://github.com/ambv/black) tool spits out. It usually does a good job and makes code formatting a no-brainer. Sometimes it looks a bit like Lisp. Is that a bad thing? :stuck_out_tongue:

adler-j: Lets go ahead for now, but long term I'm not sure if it is a bit to verbose

kohr-h: Actually that code didn't work at all, as I found out. I changed the implementation, now I think it's much cleaner. Please check again.

odl/test/largescale/tomo/ray_transform_slow_test.py
```diff
@@ -11,52 +11,61 @@
...
-
-
-# bug in pytest (ignores pytestmark) forces us to do this this
-largescale = " or not pytest.config.getoption('--largescale')"
-projectors = [pytest.mark.skipif(p.args[0] + largescale, p.args[1])
-              for p in projectors]
+projectors = []
+projectors.extend(
+    (pytest.param(value, marks=skip_if_no_astra)
+     for value in ['par2d astra_cpu uniform',
```
adler-j: Prefer something other than "value", e.g. be specific.

kohr-h: Well, `values` is the name under which the input is stored in `pytest.param`. But it's a generic construct, so you're right, we can be somewhat more specific here.

PR #1465, opened 2019-02-11
BUG: avoid circular import when pytest isn't available

kohr-h:


odl/tomo/util/testutils.py
```diff
@@ -21,8 +21,9 @@
     import pytest
 
 except ImportError:
-    # Use the identity decorator (default of OptionalArgDecorator)
-    from odl.util import OptionalArgDecorator as identity
+
+    def identity(f, **kwargs):
```
adler-j: And there is no other way to do this properly? This feels a bit like a hack.

kohr-h: I'm not sure what you mean by "properly". We should have a function that can be used as a decorator and that does nothing. Actually, the `**kwargs` is wrong, there should be a little indirection.

kohr-h: I changed the code to accommodate for optional args. Both of these will work and keep `f` untouched:
```python
@identity
def f(x):
    return x

@identity(a=42)
def f(x):
    return x
```

adler-j: What I mean by properly is fixing the import :)

adler-j: There is no way to do that? E.g. just reorder and put testutils last?

odl/util/testutils.py
```diff
@@ -205,7 +205,12 @@ def is_subdict(subdict, dictionary):
     import pytest
 
 except ImportError:
-    from odl.util import OptionalArgDecorator as identity
+
+    def identity(*args, **kwargs):
```
adler-j:
```
@identity(apply=lambda x: x+1)
def f(x):
    return 2 * x
```

This would fail, but perhaps that's expected?

kohr-h: With the latest version it would "work" (do nothing).

odl/tomo/util/testutils.py
kohr-h: Maybe, but that import was overkill anyway. I'll leave it as it is currently.

kohr-h:
Any objections to this version?

odl/util/testutils.py
adler-j: I guess the danger is 

```
@identity(lambda x: x+1)
def f(x):
    return 2 * x
```

But perhaps we just leave that be

kohr-h:
> I guess the danger is
> 
>     @identity(lambda x: x+1)
>     def f(x):
>         return 2 * x
> 
> But perhaps we just leave that be

Yeah, it's really only intended as a drop-in so that the decorators don't crash anything if we cannot import `pytest`. The issue would be the same in the imported decorator anyway.

adler-j:
Ok, I guess that's a hard corner case.

PR #1470, opened 2019-02-13
Issue 1467  reduction operator separable sum

zickert:
The problem with creating a `ReductionOperator` from functionals boils down to `RealNumbers` not being a `LinearSpace` in ODL. I suppose we do not want to change that, so I simply added a reference to `SeparableSum` in the docstring of `ReductionOperator`.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -1661,7 +1661,6 @@ class SeparableSum(Functional):
                 isinstance(functionals[1], Integral)):
             functionals = [functionals[0]] * functionals[1]
 
-        # Check that all elements of `functionals` are indeed functionals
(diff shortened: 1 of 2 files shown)


odl/solvers/functional/default_functionals.py
```diff
@@ -1661,6 +1661,10 @@ def __init__(self, *functionals):
                 isinstance(functionals[1], Integral)):
             functionals = [functionals[0]] * functionals[1]
 
+        # Check that all elements of `functionals` are indeed functionals
```
kohr-h: The code already says this :-)

zickert: I removed that line.

kohr-h:
Thanks!

PR #1472, opened 2019-02-13
BLD: move metadata to setup.cfg

kohr-h:
This PR migrates all static settings from `setup.py` to `setup.cfg`. The latter is a highly underused and also [under-documented](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) feature of the Python packaging ecosystem. I think that the static configuration `setup.cfg` is where the static information belongs.

This version should do the same as the previous code in `setup.py`. I've removed the install extras like `pip install odl[fftw]` since I don't think it's very useful. To install everything might be more useful, so I added the `all` option. Not that I really need it though.

Another thing I removed are the `requirements.txt` files. That's debatable, of course, but the way we used them is not how they're intended to be used. Their main purpose was to allow reproducible installations by pinning versions (not for dependencies of dependencies, though). Those files will be superseded by more comprehensive approaches for dependency management, and at the moment we don't really need them.


kohr-h:
This one should be rather straightforward to test and review. Any chance to get that done sometime soon?

setup.cfg
```diff
@@ -1,11 +1,107 @@
...
+license = MPL-2.0
+license_file = LICENSE
+url = https://github.com/odlgroup/odl
+download_url = https://github.com/odlgroup/odl/archive/master.zip
+platforms = any
+keywords = research, development, mathematics, prototyping, imaging, tomography
+classifiers =
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers
+    Intended Audience :: Science/Research
```
aringh: I guess [these](https://pypi.org/classifiers/) are the classifiers. One could add `Education` to the list?

setup.cfg
```diff
@@ -1,11 +1,107 @@
...
+    Topic :: Scientific/Engineering
+    Topic :: Scientific/Engineering :: Mathematics
```
aringh: Also `Topic :: Scientific/Engineering :: Medical Science Apps.`?

setup.py
```diff
@@ -48,115 +39,4 @@ def run_tests(self):
...
-
-# Determine version from top-level package __init__.py file
-with open(os.path.join(root_path, 'odl', '__init__.py')) as f:
-    for line in f:
-        if line.startswith('__version__'):
-            version = line.strip().split()[-1][1:-1]
-            break
-
-
-long_description = """
```
aringh: This text will be lost. Feature or bug?

setup.cfg
kohr-h: Ah good, that's the new link. Thanks! I'll add it as a comment.

I'll also add `Education` -- in the broader sense of science education it seems applicable :wink:

setup.cfg
kohr-h: Yeah why not.

setup.py
kohr-h: Now the `README.md` file [will be picked](https://github.com/odlgroup/odl/blob/405a151a5daabdbb6b432e62cfeac189ec2930c3/setup.cfg#L5-L6) instead. That saves us from keeping both in sync.

kohr-h:
> I am not familiar with the "set-up"-part of the package, and I do not know what conventions there are for doing this.

Neither am I, but I tried to follow the instructions in the setuptools documentation. You can also check some of the effects by running `pip install -e .` and then looking at the contents of `odl.egg-info/PKG-INFO`.

kohr-h:
And thanks for the review @aringh!

kohr-h:
Done, plus a few more metadata items I found applicable. Will merge after CI.

PR #1473, opened 2019-02-18
MAINT: use title case in figure titles

kohr-h:
Boring PR that makes all figure titles adhere to title case.


kohr-h:
I'll use more "o"s next time in the text ;-)

PR #1478, opened 2019-03-07
 Cylindrical and Spherical detectors added, ConeBeamGeometry created

JevgenijaAksjonova:
by analogy to CircularDetecotor and FanBeamGeometry

### DIFF
--- a/odl/test/tomo/geometry/geometry_test.py
+++ b/odl/test/tomo/geometry/geometry_test.py
@@ -34,7 +34,7 @@ det_pos_init_3d_params = list(set(permutations([1, 0, 0]))
 det_pos_init_3d = simple_fixture('det_pos_init', det_pos_init_3d_params)
 axis = simple_fixture('axis', det_pos_init_3d_params)
 shift = simple_fixture('shift', [0, 1])
-curved_det = simple_fixture('curved_det', [False, True])
+detector_type = simple_fixture('detector_type', [0, 1, 2])
 
 
 # --- tests --- #
@@ -423,21 +423,21 @@ def test_parallel_beam_geometry_helper():
     assert geometry.det_partition.extent == pytest.approx(2 * rho)
 
 
-def test_fanbeam_props(curved_det, shift):
+def test_fanbeam_props(detector_type, shift):
     """Test basic properties of 2d fan beam geometries."""
     full_angle = 2 * np.pi
     apart = odl.uniform_partition(0, full_angle, 10)
     dpart = odl.uniform_partition(-np.pi / 2, np.pi / 2, 10)
     src_rad = 10
     det_rad = 5
-    curve_rad = src_rad + det_rad + 1 if curved_det else None
+    curve_rad = src_rad + det_rad + 1 if detector_type > 0 else None
     translation = np.array([shift, shift], dtype=float)
     geom = odl.tomo.FanBeamGeometry(apart, dpart, src_rad, det_rad,
                                     det_curvature_radius=curve_rad,
                                     translation=translation)
 
     assert geom.ndim == 2
-    if curved_det:
+    if detector_type > 0:
         assert isinstance(geom.detector, odl.tomo.CircularDetector)
     else:
         assert isinstance(geom.detector, odl.tomo.Flat1dDetector)
@@ -458,7 +458,7 @@ def test_fanbeam_props(curved_det, shift):
     # shift by 1 (=detector param) along the detector axis [0, 1] at that
     # angle. For curved detector, we have to take curvature of into account.
     # Global translation should come afterwards.
-    if curved_det:
+    if detector_type > 0:
         det_param = np.pi / 6
         dx = curve_rad * (1 - np.cos(det_param))
         dy = curve_rad * np.sin(det_param)
@@ -548,7 +548,7 @@ def test_fanbeam_frommatrix():
                                                    det_rad, sing_mat)
 
 
-def test_helical_cone_flat_props(shift):
+def test_helical_cone_beam_props(detector_type, shift):
     """Test basic properties of 3D helical cone beam geometries."""
     full_angle = 2 * np.pi
     apart = odl.uniform_partition(0, full_angle, 10)
@@ -557,11 +557,23 @@ def test_helical_cone_flat_props(shift):
     det_rad = 5
     pitch = 2.0
     translation = np.array([shift, shift, shift], dtype=float)
-    geom = odl.tomo.ConeFlatGeometry(apart, dpart, src_rad, det_rad,
+    if detector_type == 2:
+        curve_rad = [src_rad + det_rad + 1] * 2
+    elif detector_type == 1:
+        curve_rad = src_rad + det_rad + 1
+    else:
+        curve_rad = None
+    geom = odl.tomo.ConeBeamGeometry(apart, dpart, src_rad, det_rad,
+                                     det_curvature_radius=curve_rad,
                                      pitch=pitch, translation=translation)
 
     assert geom.ndim == 3
-    assert isinstance(geom.detector, odl.tomo.Flat2dDetector)
+    if detector_type == 2:
+        assert isinstance(geom.detector, odl.tomo.SphericalDetector)
+    elif detector_type == 1:
+        assert isinstance(geom.detector, odl.tomo.CylindricalDetector)
+    else:
+        assert isinstance(geom.detector, odl.tomo.Flat2dDetector)
 
     # Check defaults
     assert all_almost_equal(geom.axis, [0, 0, 1])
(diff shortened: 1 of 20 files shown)


adler-j:
Should this be ready for review? I could see if I can get some time to do that.

JevgenijaAksjonova:
Yes, I am waiting for a review. Thanks

ozanoktem:
Could we give @JevgenijaAksjonova som feedback on her revised code?

adler-j:
> Could we give @JevgenijaAksjonova som feedback on her revised code?

I'll try to but you are also most welcome to submit a review, its good exercise :)

ozanoktem:
> > Could we give @JevgenijaAksjonova som feedback on her revised code?
>
> I'll try to but you are also most welcome to submit a review, its good exercise :)

Thanks for the offer despite having a dense schedule. In principle I also agree that it would have been a good exercise, but as you know @adler-j , I have no breathing space until beginning of June due to evaluations and position/grant applications. I think it is unwise to wait until then to give @JevgenijaAksjonova som feedback, donation you agree @adler-j ? Perhaps @olivierverdier and @kohr-h  could help out here?

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -34,7 +34,7 @@
 det_pos_init_3d = simple_fixture('det_pos_init', det_pos_init_3d_params)
 axis = simple_fixture('axis', det_pos_init_3d_params)
 shift = simple_fixture('shift', [0, 1])
-curved_det = simple_fixture('curved_det', [False, True])
+detector_type = simple_fixture('detector_type', [0, 1, 2])
```
olivierverdier: Minor quibble: it would be easier for a later maintainer of the tests to have explicit names for the `detector_type`, rather than numbers. Even "flat", "curved1", "curved2" would be better than 0,1,2.

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,640 @@ def __str__(self):
...
+        self.__axes = axes / np.linalg.norm(axes, axis=1, keepdims=True)
+
+        self.__radius = float(radius)
+        if self.__radius <= 0:
+            raise ValueError('`radius` must be positive')
+
+        initial_axes = np.array([[0, -1, 0], [0, 0, 1]])
+        r1 = rotation_matrix_from_to(initial_axes[0], axes[0])
+        r2 = rotation_matrix_from_to(initial_axes[1], axes[1])
+        self.__rotation_matrix = np.matmul(r1, r2)
```
kohr-h: Are you sure this is correct?

First, if `axes` are not perpendicular, how can you ever rotate the `inital_axes` to `axes`? To avoid getting insanely complex geometries with sheared cylinders, I think we should ensure that the axes are perpendicular.

And for perpendicular `axes`, I would expect that one first has to rotate one of the two vectors to its final location, and then *around* that new axis to match the other one.

Please make sure with some with some checks (e.g. unit tests) that this part is correct.

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,640 @@ def __str__(self):
...
```
kohr-h: Same as above

odl/tomo/geometry/conebeam.py
```diff
@@ -797,9 +828,30 @@ def __init__(self, apart, dpart, src_radius, det_radius, pitch=0,
...
+            detector = Flat2dDetector(dpart, axes=det_axes_init,
+                                      check_bounds=check_bounds)
+        else:
+            if np.array(det_curvature_radius).shape == ():
+                detector = CylindricalDetector(dpart,
+                                               radius=det_curvature_radius,
+                                               axes=det_axes_init,
+                                               check_bounds=check_bounds)
+            elif np.array(det_curvature_radius).shape == (2,):
+                if det_curvature_radius[0] != det_curvature_radius[1]:
```
kohr-h: Oh now I get it. We don't actually support 2 different radii. In that case, I think the API is a bit awkward, and it would definitely be better to use a single float for spherical and `[r, None]` and `[r, float('inf')]` for cylinders.

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,640 @@ def __str__(self):
...
+        return self.__translation
+
+    def surface(self, param):
+        """Return the detector surface point corresponding to ``param``.
+
+        For parameters ``phi`` and ``theta``, the surface point is given by ::
+
+            surf = R * radius * ( cos(phi)*cos(theta),
+                                 -sin(phi)*cos(theta),
+                                           sin(theta)) + t
```
kohr-h: whitespace around `*`

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,640 @@ def __str__(self):
...
+        deriv = np.stack((deriv_phi, deriv_theta), axis=-2)
+        deriv = np.matmul(deriv, np.transpose(self.rotation_matrix))
+
+        if squeeze_out:
+            deriv = deriv.squeeze()
+
+        return deriv
+
+    def surface_measure(self, param):
+        """Return the arc length measure at ``param``.
```
kohr-h: surface area measure

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,640 @@ def __str__(self):
...
+
+        This is a constant function evaluating to (`radius`, `radius`)
```
kohr-h: Should be scalar

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -576,17 +588,39 @@ def test_helical_cone_flat_props(shift):
...
-    # Global translation should come last.
-    assert all_almost_equal(geom.det_point_position(np.pi / 2, [1, 1]),
-                            translation + [-det_rad, 1, 1 + pitch / 4])
+    # Here we expect to rotate the reference point to [-det_rad, 0, 0] and,
+    # in the case of flat detector, shift by (1, 1) (=detector param)
+    # along the detector axes ([0, 1, 0], [0, 0, 1]) at that angle.
+    # In the case of curved detectors, curvature is taken into account.
+    # In addition, everything is shifted along the rotation axis [0, 0, 1]
+    # by 1/4 of the pitch (since the pitch is the vertical distance after
+    # a full turn 2*pi). Global translation should come last.
```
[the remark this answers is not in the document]
JevgenijaAksjonova: There is only this one sentence about curved detectors, that "curvature is taken into account". Other things apply to all detectors.

I don't know if it makes sense to explain in words, how curvature is taken into account, since the formulas should speak for themselves...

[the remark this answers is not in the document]
ozanoktem: I guess it should be enough to state that "curvature is included in the model for the detector geometry".

odl/tomo/geometry/conebeam.py
```diff
@@ -624,6 +626,13 @@ def __init__(self, apart, dpart, src_radius, det_radius, pitch=0,
         det_radius : nonnegative float
             Radius of the detector circle. Must be nonzero if ``src_radius``
             is zero.
+        det_curvature_radius : nonnegative float or a sequence of
+            two nonnegative floats, optional
+            Radius or radii of the detector curvature.
+            If ``None``, a flat detector is used,
+            if one value, a cylindrical detector is used,
+            if a sequence of two elements, e.g. [r, r], a spherical detector
```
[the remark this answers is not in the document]
JevgenijaAksjonova: Yes, the cylindrical case is more common. I will implement the proposed alternative.

odl/tomo/geometry/conebeam.py
```diff
@@ -624,6 +626,13 @@ def __init__(self, apart, dpart, src_radius, det_radius, pitch=0,
...
+            is used.
```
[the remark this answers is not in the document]
JevgenijaAksjonova: I have implemented only r1=r2. It seems to be very easy to extend it given the formulas that you just have proposed. 
However, from practical point of view this is not yet needed.

odl/tomo/geometry/detector.py
JevgenijaAksjonova: Is it OK, to remove this function and rely on the default implementation in the Detector class?

odl/tomo/geometry/detector.py
JevgenijaAksjonova: You are right, I don't know what I was thinking here...

JevgenijaAksjonova: So, it should be like this
r1 = rotation_matrix_from_to(initial_axes[0], axes[0])
r2 = rotation_matrix_from_to(np.matmul(r1, initial_axes[1]), axes[1])
and it will work, if axes are perpendicular.

odl/tomo/geometry/detector.py
JevgenijaAksjonova: I am afraid I dont really understand what this function is for, should it measure cell area?

odl/tomo/geometry/conebeam.py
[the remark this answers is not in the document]
kohr-h: I agree, it seems like an uncommon case. Interface-wise I think we should offer the full flexibility so that this case can be added later. Currently we can just raise a `NotImplementedError`.

odl/tomo/geometry/detector.py
kohr-h: Yes, usually the default implementation should be okay. The purpose of this function would be to integrate over the manifold described by the geometry. Not sure whether we really ever need that, maybe it was just too easy to add :roll_eyes:

odl/test/tomo/geometry/geometry_test.py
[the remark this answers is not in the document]
JevgenijaAksjonova: Since this is just a comment to test case, I think it can stay as is..

JevgenijaAksjonova:
I hope that all the issues has been addressed now.

JevgenijaAksjonova:
Hi! 
As I understand some checks fail because of CI issue https://github.com/odlgroup/odl/issues/1500, right?
Can we proceed with merge?

odl/tomo/geometry/conebeam.py
```diff
@@ -845,6 +894,13 @@ def frommatrix(cls, apart, dpart, src_radius, det_radius, init_matrix,
             determine the new vectors. If present, the fourth column acts
             as a translation after the initial transformation.
             The resulting ``det_axes_init`` will be normalized.
+        det_curvature_radius : nonnegative float or a sequence of
+            two nonnegative floats, optional
+            Radius or radii of the detector curvature.
+            If ``None``, a flat detector is used,
+            if one value, a cylindrical detector is used,
+            if a sequence of two elements, e.g. [r, r], a spherical detector
+            is used.
```
kohr-h: Should be the same as in `__init__`, no?

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,528 @@ def __str__(self):
...
+
+        self.__radius = float(radius)
+        if self.__radius <= 0:
+            raise ValueError('`radius` must be positive')
+
+        initial_axes = np.array([[0, -1, 0], [0, 0, 1]])
+        r1 = rotation_matrix_from_to(initial_axes[0], axes[0])
+        r2 = rotation_matrix_from_to(np.matmul(r1, initial_axes[1]), axes[1])
+        self.__rotation_matrix = np.matmul(r2, r1)
+        self.__translation = (- self.__radius
```
kohr-h:
```suggestion
        self.__translation = (-self.__radius
```

odl/tomo/geometry/detector.py
```diff
@@ -894,6 +895,528 @@ def __str__(self):
...
+    def surface_deriv(self, param):
+        """Return the surface derivative at ``param``.
+
+        The derivative at parameters ``phi`` and ``h`` is given by ::
+
+            deriv = R * ((-radius * sin(phi), 0),
+                         (-radius * cos(phi), 0),
+                         (                 0, 1))
+
+        where R is a rotation matrix.
```
kohr-h:
```suggestion
        where ``R`` is a rotation matrix.
```

JevgenijaAksjonova:
Hi! I have made the changes and rebased.

kohr-h:
Thanks a lot @JevgenijaAksjonova!

JevgenijaAksjonova:
Thank you for the review!

ozanoktem:
Good work @JevgenijaAksjonova

PR #1481, opened 2019-03-13
MAINT: move coverage config to setup.cfg

kohr-h:
Little addition to #1472: Move the `coverage` configuration to `setup.cfg`, which helps us get rid of one top-level file. This is officially supported according to [the documentation](https://coverage.readthedocs.io/en/latest/config.html?highlight=setup.cfg).


setup.cfg
```diff
@@ -111,3 +112,55 @@ max-line-length = 79
...
     __pycache__
+
+[coverage:run]
+branch = False
+include =
+    odl/*
+omit =
+    */__init__.py
+    odl/diagnostics/*
+    odl/util/*
```
adler-j: Why do we omit this?

kohr-h: I don't know. My guess is that it's from a time when the utility code was really only for internal usage and rather small in scope. Mostly little helpers and testing stuff. I'll check what makes sense to exclude there.

kohr-h: Now more fine-grained. And `tests/` excluded, they used to drag down our coverage score.

kohr-h:
Some more last changes. What I can't get rid of is the `def` and similar lines showing up as not covered. According to [the coverage FAQ](https://coverage.readthedocs.io/en/v4.5.x/faq.html) this shouldn't happen if called from the command line.

But anyway, the reported coverage is probably closer to reality now :-)
I'll merge after the final CI run has finished.

kohr-h:
I got one last idea that seems to work locally. Let's see how we'll do.

kohr-h:
Okay, I've solved the problem. If I run `coverage` manually (not through `pytest-cov`) the `def` and `class` lines etc. are correctly reported as covered. So I've changed `.travis.yml` to spit out a temporary script to run the tests (basically `import odl; odl.test()`) and to run it with `coverage run`.

PR #1483, opened 2019-03-17
link prox of conjugate of linfty norm to proj_l1

mehrhardt:
This is a super small PR, just linking the existing projection onto l1 balls to the convex conjugate of the linfty norm.

I have seen that other prox operators include scalings `lambda` and shifts `g` but I left them out here for simplicity.

### DIFF
--- a/odl/solvers/nonsmooth/proximal_operators.py
+++ b/odl/solvers/nonsmooth/proximal_operators.py
@@ -1495,12 +1495,12 @@ def proximal_linfty(space):
 
 
 def proximal_convex_conj_linfty(space):
-    """Proximal operator factory of the Linfty norm/distance convex conjugate.
+    r"""Proximal operator factory of the Linfty norm/distance convex conjugate.
 
     Implements the proximal operator of the convex conjugate of the
     functional ::
 
-        F(x) = ||x||_\\infty
+        F(x) = \|x\|_\infty
 
     with ``x`` in ``space``.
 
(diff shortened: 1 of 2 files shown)


mehrhardt:
The issue that pep8speaks raised

> Checking updated PR...
> 
> In the file odl/solvers/functional/default_functionals.py:
> Line 1564:80: E501 line too long (88 > 79 characters)
> Line 1464:80: E501 line too long (88 > 79 characters)

is actually not part of my changes.

odl/solvers/nonsmooth/proximal_operators.py
```diff
@@ -1494,6 +1494,67 @@ def _call(self, x, out):
...
+    The convex conjugate :math:`F^*` of the functional
+
+    .. math::
+        F(x) = ||x||_\\infty.
+
+    is in the case of scalar-valued functions given by the indicator function
+    of the unit 1-norm ball
+
+    .. math::
+        F^*(y) = \\iota_{B_1} \\big( y \\big).
```
kohr-h: How about fixing the math formatting here? Would be better to have a raw string `r"""` and use single-backslash for TeX commands. Also, norm bars should be `\|`.

mehrhardt: Thanks for the feedback. I fixed the formatting. Any other comments?

kohr-h: No I don't have any other complaints :-)

Just an observation @mehrhardt: the email that you used for the git commits doesn't seem to be associated with your Github account. You may want to do that, otherwise these commits are credited to some Matthias J. Ehrhardt that is not you :-)

mehrhardt: Cheers, got this sorted out now.

PR #1484, opened 2019-03-19
ENH: make FOMs accept array-like inputs

kohr-h:
With this change, all supervised FOMs accept array-like input, instead of insisting on ODL space elements. Makes them much more useful. (And many don't even depend on space properties.)

Unit tests added, they pass, and the doctests still work as well.

### DIFF
--- a/odl/contrib/fom/supervised.py
+++ b/odl/contrib/fom/supervised.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2018 The ODL contributors
+# Copyright 2014-2019 The ODL contributors
 #
 # This file is part of ODL.
 #
@@ -29,16 +29,16 @@ def mean_squared_error(data, ground_truth, mask=None,
 
     Parameters
     ----------
-    data : `Tensor`
-        Input data to compare to the ground truth.
-    ground_truth : `Tensor`
-        Reference to compare ``data`` to.
+    data : `Tensor` or `array-like`
+        Input data to compare to the ground truth. If not a `Tensor`, a
+        default space with cell size 1 will be assumed.
+    ground_truth : `array-like`
+        Reference to which ``data`` should be compared.
     mask : `array-like`, optional
         If given, ``data * mask`` is compared to ``ground_truth * mask``.
     normalized  : bool, optional
         If ``True``, the output values are mapped to the interval
-        :math:`[0, 1]` (see `Notes` for details), otherwise return the
-        original mean squared error.
+        :math:`[0, 1]` (see `Notes` for details).
     force_lower_is_better : bool, optional
         If ``True``, it is ensured that lower values correspond to better
         matches. For the mean squared error, this is already the case, and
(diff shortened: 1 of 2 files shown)


odl/contrib/fom/supervised.py
```diff
@@ -201,22 +215,32 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False,
 
     The normalized variant takes values in :math:`[0, 1]`.
     """
-    l1_norm = odl.solvers.L1Norm(data.space)
+    try:
```
adler-j: Isn't this basically `odl.vector` with a check?

odl/contrib/fom/supervised.py
```diff
@@ -201,22 +215,32 @@ def mean_value_difference(data, ground_truth, mask=None, normalized=False,
...
+        space = data.space
```
adler-j: data = space.element(data)?

odl/contrib/fom/supervised.py
```diff
@@ -305,10 +339,11 @@ def range_difference(data, ground_truth, mask=None, normalized=False,
 
     Parameters
     ----------
-    data : `Tensor`
-        Input data to compare to the ground truth.
-    ground_truth : `Tensor`
-        Reference to compare ``data`` to.
+    data : `Tensor` or `array-like`
+        Input data to compare to the ground truth. If not a `Tensor`, a
+        default space with cell size 1 will be assumed.
```
adler-j: cell size 1 is not quite true, we create a simple R^n space, right?

odl/contrib/fom/supervised.py
```diff
@@ -790,16 +853,26 @@ def noise_power_spectrum(data, ground_truth, radial=False,
...
         domain ``[0, rmax]``, where ``rmax`` is the radius of the smallest
-        ball containing ``data.space.domain``. Its shape is ``(N,)`` with ::
+        ball containing ``space.domain``. Its shape is ``(N,)`` with ::
 
             N = int(sqrt(sum(n ** 2 for n in image.shape)) / binning_factor)
     """
-    ft = odl.trafos.FourierTransform(data.space, halfcomplex=False)
+    try:
+        space = data.space
+        assert isinstance(space, odl.DiscreteLp)
```
adler-j: Meeeh

odl/contrib/fom/supervised.py
kohr-h: True, forgot to update the wording.

odl/contrib/fom/supervised.py
kohr-h: Could very well be.

odl/contrib/fom/supervised.py
kohr-h: C'mon, it's `contrib` :-)

kohr-h: Actually no, it's caught in the next line and handled. So no meh

odl/contrib/fom/supervised.py
adler-j: An option is to change `odl.vector` to not cast anything if it is already a vector, then just call it here.

adler-j: Perhaps add a `odl.asvector` to be exactly consistent with numpy

odl/contrib/fom/supervised.py
kohr-h: We need `space` in the first place, no?

adler-j: I guess they could technically be different, no?

odl/contrib/fom/supervised.py
kohr-h: Nah, it's a crappy function anyway. I don't want to add more crappiness.

kohr-h: I've updated the code, should be better now.

odl/contrib/fom/supervised.py
kohr-h: Isn't `data.space.element(data) is data` always true?

kohr-h:
Fixes done, okay now?

kohr-h:
> **[adler-j](/adler-j) ** approved these changes [ 11 minutes ago](#pullrequestreview-216599956)

:tada:

PR #1459, opened 2019-02-05
Remove FunctionSpace and Discretization

kohr-h:
**Rough Overview:**
- `Discretization` and `DiscretizationElement` are gone, functionality has been moved onto `DiscreteLp` and `DiscreteLpElement`
- `FunctionSpace` and `FunctionSpaceElement` are gone, functionality for vectorization and evaluation has been broken into a bunch of functions that live in `odl.discr.discr_utils`.
  For instance, `FunctionSpace.element()` created a wrapper that unified the interface of functions with and without `out` argument. This is done in the function `make_func_for_sampling`.
- Support for non-vectorized functions in `DiscreteLp.element()` is removed.
  *Rationale:* Why did we ever support this?
- Some parts of `discretization.py` (now gone) are inlined in `lp_discr.py`, which makes it appear as if it had changed a lot. It hasn't.
- `discr_sequence_space` is gone.
  *Rationale:* lack of usefulness.

- ~`FunctionSpaceMapping` classes have simply been turned into functions and moved over to the `discr_utils` module~ done in #1520 
- ~`interp` is no longer a property of the space but up to each operator (which I consider to be the better approach)~ done in #1520 

Closes #1457

PR #1485, opened 2019-03-21
ENH: allow choice of ASTRA projector type

kohr-h:
I suggest we switch to the `linear` for 2D parallel beam projections. The CPU line kernel gives quite bad results in terms of interpolation artifacts (@wjp: is that to be expected? I didn't imagine it to be this severe.). For comparison a projection-backprojection cycle:

**Line kernel**
line

**Linear kernel**
linear

### DIFF
--- a/odl/tomo/backends/astra_cpu.py
+++ b/odl/tomo/backends/astra_cpu.py
@@ -25,25 +25,49 @@ try:
 except ImportError:
     pass
 
+__all__ = (
+    'astra_cpu_forward_projector',
+    'astra_cpu_back_projector',
+    'default_astra_proj_type',
+)
 
-__all__ = ('astra_cpu_forward_projector', 'astra_cpu_back_projector')
 
+def default_astra_proj_type(geom):
+    """Return the default ASTRA projector type for a given geometry.
 
-def _default_proj_type(geom):
-    """Return the default projector type for a given geometry."""
+    Parameters
+    ----------
+    geom : `Geometry`
+        ODL geometry object for which to get the default projector type.
+
+    Returns
+    -------
+    astra_proj_type : str
+        Default projector type for the given geometry.
+
+        In 2D:
+
+        - `ParallelBeamGeometry`: ``'linear'``
+        - `DivergentBeamGeometry`: ``'line_fanflat'``
+
+        In 3D:
+
+        - `ParallelBeamGeometry`: ``'linear3d'``
+        - `DivergentBeamGeometry`: ``'linearcone'``
+    """
     if isinstance(geom, ParallelBeamGeometry):
         return 'linear' if geom.ndim == 2 else 'linear3d'
     elif isinstance(geom, DivergentBeamGeometry):
         return 'line_fanflat' if geom.ndim == 2 else 'linearcone'
     else:
         raise TypeError(
-            'no default exists for {}, `proj_type` must be given explicitly'
+            'no default exists for {}, `astra_proj_type` must be given explicitly'
             ''.format(type(geom))
         )
 
 
 def astra_cpu_forward_projector(vol_data, geometry, proj_space, out=None,
-                                proj_type=None):
+                                astra_proj_type=None):
     """Run an ASTRA forward projection on the given data using the CPU.
 
     Parameters
@@ -57,11 +81,11 @@ def astra_cpu_forward_projector(vol_data, geometry, proj_space, out=None,
     out : ``proj_space`` element, optional
         Element of the projection space to which the result is written. If
         ``None``, an element in ``proj_space`` is created.
-    proj_type : str, optional
+    astra_proj_type : str, optional
         Type of projector that should be used. See `the ASTRA documentation
         <http://www.astra-toolbox.com/docs/proj2d.html>`_ for details.
         By default, a suitable projector type for the given geometry is
-        selected.
+        selected, see `default_astra_proj_type`.
 
     Returns
     -------
@@ -102,9 +126,9 @@ def astra_cpu_forward_projector(vol_data, geometry, proj_space, out=None,
     proj_geom = astra_projection_geometry(geometry)
 
     # Create projector
-    if proj_type is None:
-        proj_type = _default_proj_type(geometry)
-    proj_id = astra_projector(proj_type, vol_geom, proj_geom, ndim)
+    if astra_proj_type is None:
+        astra_proj_type = default_astra_proj_type(geometry)
+    proj_id = astra_projector(astra_proj_type, vol_geom, proj_geom, ndim)
 
     # Create ASTRA data structures
     vol_data_arr = np.asarray(vol_data)
@@ -131,7 +155,7 @@ def astra_cpu_forward_projector(vol_data, geometry, proj_space, out=None,
 
 
 def astra_cpu_back_projector(proj_data, geometry, vol_space, out=None,
-                             proj_type=None):
+                             astra_proj_type=None):
     """Run an ASTRA back-projection on the given data using the CPU.
 
     Parameters
@@ -145,11 +169,11 @@ def astra_cpu_back_projector(proj_data, geometry, vol_space, out=None,
     out : ``vol_space`` element, optional
         Element of the reconstruction space to which the result is written.
         If ``None``, an element in ``vol_space`` is created.
-    proj_type : str, optional
+    astra_proj_type : str, optional
         Type of projector that should be used. See `the ASTRA documentation
         <http://www.astra-toolbox.com/docs/proj2d.html>`_ for details.
         By default, a suitable projector type for the given geometry is
-        selected.
+        selected, see `default_astra_proj_type`.
 
     Returns
     -------
@@ -196,9 +220,9 @@ def astra_cpu_back_projector(proj_data, geometry, vol_space, out=None,
                          allow_copy=True)
 
     # Create projector
-    if proj_type is None:
-        proj_type = _default_proj_type(geometry)
-    proj_id = astra_projector(proj_type, vol_geom, proj_geom, ndim)
+    if astra_proj_type is None:
+        astra_proj_type = default_astra_proj_type(geometry)
+    proj_id = astra_projector(astra_proj_type, vol_geom, proj_geom, ndim)
 
     # Convert out to correct dtype and order if needed.
     with writable_array(out, dtype='float32', order='C') as out_arr:
--- a/odl/tomo/backends/astra_cpu.py
+++ b/odl/tomo/backends/astra_cpu.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 The ODL contributors
+# Copyright 2014-2019 The ODL contributors
 #
 # This file is part of ODL.
 #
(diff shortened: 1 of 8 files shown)


kohr-h:
Note: the images come from the [`ray_trafo_parallel_2d.py`](https://github.com/odlgroup/odl/blob/master/examples/tomo/ray_trafo_parallel_2d.py) example.

adler-j:
Is the linear kernel matched?

wjp:
ASTRA's CPU projectors all have matched FP and BP, yes.

wjp:
By the way, if you're evaluating 2D CPU projection kernels, I've just added a new 2D parallel beam CPU projector to astra called `distance_driven` ( from http://ieeexplore.ieee.org/abstract/document/1239600/ ), which is supposedly quite nice in terms of discretization artifacts. We're still evaluating its behaviour here.

kohr-h:
How about this now?

kohr-h:
Yes, I've done exactly that in the #1459 PR. I can move it into this one instead.

kohr-h:
I've pulled over the changes as mentioned, ready for review.

odl/tomo/backends/astra_setup.py
```diff
@@ -161,26 +170,29 @@ def astra_volume_geometry(reco_space):
...
-        if (not reco_space.partition.has_isotropic_cells and
-                not astra_supports('anisotropic_voxels_2d')):
+        if (
+            not vol_space.partition.has_isotropic_cells
+            and not astra_supports('anisotropic_voxels_2d')
+        ):
             raise NotImplementedError(
                 'non-isotropic pixels in 2d volumes not supported by ASTRA '
-                'v{}'.format(ASTRA_VERSION))
+                'v{}'.format(ASTRA_VERSION)
```
adler-j: Perhaps clarify from what version this is supported? E.g. v < x.y.z?

odl/tomo/backends/astra_setup.py
```diff
@@ -574,88 +588,90 @@ def astra_data(astra_geom, datatype, data=None, ndim=2, allow_copy=False):
...
-
-    # create config dict
+    astra_geom = astra_proj_geom['type']
+    if (
+        astra_geom == 'parallel_vec'
+        and not astra_supports('par2d_vec_geometry')
+    ):
+        raise ValueError(
+            "'parallel_vec' geometry not supported in the current version "
+            "of ASTRA"
```
adler-j: Please clarify versions here as well

odl/tomo/backends/astra_setup.py
```diff
@@ -574,88 +588,90 @@ def astra_data(astra_geom, datatype, data=None, ndim=2, allow_copy=False):
...
+    elif astra_geom in {'fanflat', 'fanflat_vec'}:
+        valid_proj_types = ['line_fanflat', 'strip_fanflat', 'cuda']
+    elif astra_geom in {'parallel3d', 'parallel3d_vec'}:
+        valid_proj_types = ['linear3d', 'cuda3d']
+    elif astra_geom in {'cone', 'cone_vec'}:
+        valid_proj_types = ['linearcone', 'cuda3d']
+    else:
+        raise ValueError('invalid geometry type {!r}'.format(astra_geom))
+
+    if astra_proj_type not in valid_proj_types:
```
adler-j: How could this even happen given the above?

odl/tomo/backends/astra_setup.py
```diff
@@ -574,88 +588,90 @@ def astra_data(astra_geom, datatype, data=None, ndim=2, allow_copy=False):
...
     proj_cfg['VolumeGeometry'] = astra_vol_geom
     proj_cfg['ProjectionGeometry'] = astra_proj_geom
     proj_cfg['options'] = {}
 
-    # Add the hacky 1/r^2 weighting exposed in intermediate versions of
-    # ASTRA
-    if (proj_type in ('cone', 'cone_vec') and
-            astra_supports('cone3d_hacky_density_weighting')):
+    # Add the hacky 1/r^2 weighting exposed in intermediate versions of ASTRA
+    # TODO(kohr-h): add non-hacky version when available
```
adler-j: Isn't this reputably less hacky now as per https://github.com/odlgroup/odl/issues/1415?

odl/tomo/backends/skimage_radon.py
```diff
@@ -9,104 +9,99 @@
...
     boundary values.
     """
-    min_x = skimage_range.domain.min()[1]
-    max_x = skimage_range.domain.max()[1]
+    min_x = skimage_proj_space.domain.min()[1]
+    max_x = skimage_proj_space.domain.max()[1]
 
-    def interpolation_wrapper(x):
+    def interpolator(x):
         x = (x[0], np.maximum(min_x, np.minimum(max_x, x[1])))
```
adler-j: `np.clip(x[1], min_x, max_x)`

odl/tomo/backends/skimage_radon.py
```diff
@@ -9,104 +9,99 @@
...
+        The volume to project.
     geometry : `Geometry`
-        The projection geometry to use
-    range : `DiscreteLp`
-        range of this projection (sinogram space)
-    out : ``range`` element, optional
-        An element in range that the result should be written to
+        The projection geometry to use.
+    proj_space : `DiscreteLp`
+        Space in which the projections (sinograms) live.
```
adler-j: Isn't this a bit too casually written?

odl/tomo/backends/astra_cpu.py
```diff
@@ -8,41 +8,60 @@
...
-        Space to which the calling operator maps
+        Space to which the calling operator maps.
     out : ``proj_space`` element, optional
         Element of the projection space to which the result is written. If
         ``None``, an element in ``proj_space`` is created.
+    proj_type : str, optional
+        Type of projector that should be used. See `the ASTRA documentation
+        <http://www.astra-toolbox.com/docs/proj2d.html>`_ for details.
+        By default, a suitable projector type for the given geometry is
+        selected.
```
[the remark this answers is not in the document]
kohr-h: I'll make it a public function.

odl/tomo/backends/astra_cuda.py
```diff
@@ -154,8 +153,10 @@ def create_ids(self):
                                  data=self.in_array,
                                  allow_copy=False)
 
-        self.proj_id = astra_projector('nearest', vol_geom, proj_geom,
-                                       ndim=proj_ndim, impl='cuda')
+        proj_type = 'cuda' if proj_ndim == 2 else 'cuda3d'
```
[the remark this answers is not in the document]
kohr-h: I'm not sure what you mean. These are the only two possible values for CUDA projectors.

odl/tomo/backends/astra_setup.py
kohr-h: Good suggestion, requires a new function though.

odl/tomo/backends/astra_setup.py
```diff
@@ -115,6 +115,11 @@
     # Linking instead of copying of GPU memory, see
     # https://github.com/astra-toolbox/astra-toolbox/pull/93
     'gpulink': '>=1.8.3',
+
+    # Distance-driven projector for parallel 2d geometry, will be in the
+    # next release, see
```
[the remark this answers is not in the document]
kohr-h: I'll write "next release after 1.8.3"

odl/tomo/backends/astra_setup.py
kohr-h: I haven't checked this value up to that point yet, only collected the valid options.

odl/tomo/backends/astra_setup.py
kohr-h: No, I think the hackiness lies in the fact that it's not a proper 1/r^2 weighting. But I've forgotten the details.

odl/tomo/backends/skimage_radon.py
kohr-h: Well, it's not really a public function. Actually, I'll remove these implementation details from the `odl.tomo` namespace.

odl/tomo/backends/astra_setup.py
adler-j: I guess the comment was:

> Furthermore, I've made the approximation of the GPU cone beam BP density weighting more accurate (at hopefully limited computational cost since the extra factor could actually be folded into existing constants, it turns out), and also adopted the 2D fan beam GPU code to match. This density weighting is now enabled by default.

odl/tomo/backends/astra_cuda.py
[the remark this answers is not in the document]
adler-j: Still, combining dimension with implementation feels a bit off to me, especially if there are e.g. more added in the future.

odl/tomo/backends/astra_setup.py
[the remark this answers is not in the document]
adler-j: Much better

odl/tomo/backends/astra_setup.py
kohr-h: I see. But I guess a true distance weighting would differ per location in the volume, which I don't think ASTRA implements. Likely it's only a function of detector position.

odl/tomo/backends/astra_cuda.py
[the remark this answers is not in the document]
kohr-h: I still don't get it. If I change anything about these strings the code will crash (exception raised in ASTRA). What change are you suggesting?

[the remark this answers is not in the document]
adler-j: Ah, then I guess this is related to my other suggestion of naming everything that is astra related `astra_*` to make it clear that it's their name.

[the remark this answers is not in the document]
kohr-h: Okay, I can call it `astra_proj_type`.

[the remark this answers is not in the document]
adler-j: I'm perfectly happy with that solution

odl/tomo/backends/astra_cpu.py
```diff
@@ -8,41 +8,60 @@
...
+except ImportError:
+    pass
+
 
 __all__ = ('astra_cpu_forward_projector', 'astra_cpu_back_projector')
 
 
-# TODO: use context manager when creating data structures
-# TODO: is magnification scaling at the right place?
+def _default_proj_type(geom):
```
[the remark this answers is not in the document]
kohr-h: Done this

odl/tomo/backends/astra_setup.py
kohr-h: I checked the quote, it refers to a work-in-progress branch, so we can be fairly sure but not 100 % that the next release will contain the improved weighting.

kohr-h:
> Overall looks good, but needs some minor changes.

Done that.



> Also needs to add tests.

Frankly I don't really know what to test. The tests would end up just copying what's in the implementation.
If at all, such tests should be in the ray transform test code, to make sure all versions run.

kohr-h:
Ping for re-review @adler-j

kohr-h:
Ping again. How about we merge this one after resolving the conflicts? I think the job is done here.

adler-j:
Merge after conflicts are resolved. Sorry about the delay, rather busy atm.

kohr-h:
> Merge after conflicts are resolved. Sorry about the delay, rather busy atm.

I can imagine that. Thanks for popping in here anyway :-)

I'll fix the conflicts and you keep the mouse pointer ready to click "approve"^^.

kohr-h:
I'll inspect the CI for errors other than the "normal" ones, after that I'd like to merge.

kohr-h:
Just the "usual" errors in CI, merging.

Issue #1437, opened 2018-09-21
Non-uniform FFT


ozanoktem:
In some imaging modalities, like MRI and phase contrast tomography, the forward problem involves taking the Fourier transform of a function in 2D/3D that is not sampled on a grid. The current Fourier transform in ODL lacks support for this case, thereby forcing the user to interpolate in Fourier space.  An alternative would be to integer a backend for the DFT that can directly handle this situation, and there are several non-uniform FFT (NUFFT) routines out there. 

One implementation of NUFFT that has python wrappers and includes routines for computing the adjoint is Flatiron Institute Nonuniform Fast Fourier Transform (FINUFFT) which is available at  https://github.com/flatironinstitute/finufft, see also https://finufft.readthedocs.io/en/latest for the documentation. This git repo shows it is an actively developed software package. The library is currently supported for unix/linux and also tested on Mac OSX.

Bearing in mind the above, it would be a nice feature if one could integrate FINUFFT with ODL.

kohr-h:
> Flatiron Institute Nonuniform Fast Fourier Transform (FINUFFT)
Is there any particular reason for suggesting that particular library?
I still find [NFFT](https://github.com/NFFT/nfft) more convincing. More contributors, more releases, more functionality, and above all, no limitation to dimensions 1 to 3 as in FINUFFT.

Other than that, agreed, would be a nice addition.

ozanoktem:
Backed by a large institution, active git repo, bindings to various languages are all good features.

kohr-h:
All good, but limitation to dimensions 1, 2, 3 is an anti-feature. Also benchmarks would be nice to have.

ozanoktem:
Agree that NFFT has its advantages, but in NFFT most wrappers are for MATLAB rather than python. At least, it seems like that.

kohr-h:
I had a closer look at their docs (FINUFFT, that is). They claim that they're faster than NFFT if the points are not close to uniform, and that they use less RAM. Could be interesting in any case.
The API is a bit crude, but it's small and shouldn't be too hard to wrap as an ODL operator.

I wouldn't be entirely sure how to package it up (new operator or existing operator?), but there's no reason why we shouldn't have (potentially) multiple NUFFT backends.

If anybody wants to go ahead with this, great.

As an aside, what should definitely part of an effort to wrap this library is to make a `conda` package, ideally on [conda-forge](https://conda-forge.org/). Having a wrapper for a library that can't be easily installed is a non-starter.

[...]

PR #1488, opened 2019-03-29
Non uniform fft

zaccharieramzi:
This introduces the non uniform fft operator with a NFFT backend (provided via `pynfft`) and solves #1437 . A few things I am not completely certain of however:
- [`pynfft`](https://github.com/pyNFFT/pyNFFT) doesn't seem to be maintained anymore.
- I am not completely sure about the redundancies between the travis config file and the setup config (I updated both for now).
- I did some basic unit tests. If you want more please ask! (I also didn't refactor them that much, but I saw it was not the main concern)
- I am not confident the way I handled `domain` and `range` for my operator is as intended.

I will be working on the docs when I have time (this weekend or Monday probably), but this was just in case you already had some preliminary remarks.


odl/test/trafos/fourier_test.py
```diff
@@ -858,5 +859,74 @@ def fhat(x):
...
+    im_size = 512
+    shape = [im_size] * 2
+    samples = np.array(np.where(np.ones(shape) >= 0), dtype=float).T
+    samples /= im_size
+    samples -= 0.5
+    image = np.random.normal(size=shape)
+    nfft = NonUniformFourierTransform(shape=shape, non_uniform_samples=samples)
+    res_nfft = nfft(image)
+    res_np_fft = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(image), norm="ortho"))
+    assert np.allclose(res_nfft, res_np_fft.flatten())
```
zaccharieramzi: Remark btw on this. I stuck to the form used everywhere else in the unit tests, but I stumbled upon [this](https://github.com/numpy/numpy/issues/9988) and I am wondering whether we should start using the more correct form `np.testing.assert_allclose`.

kohr-h:
Don't worry about the Py2 CI fail for now. I'll try to allocate some time to review tonight.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+    The normalization is inspired from pysap-mri, mainly this class:
+    https://github.com/CEA-COSMIC/pysap-mri/blob/master/mri/reconstruct/fourier.py#L123
+    """
+    def __init__(self, shape, non_uniform_samples, domain, range):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        shape : tuple
+            The dimensions of the data whose non uniform FFT you want to
```
kohr-h: The docs tend to have a more formal tone, no "you".

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        n_dim = len(self.shape)
+        if not all(len(sample) == n_dim for sample in non_uniform_samples):
+            raise ValueError('One of the samples in `non_uniform_samples` does'
+              ' not have the right dimension')
+
+        if not all(
+                self._is_sample_valid(sample) for sample in non_uniform_samples
+            ):
+            raise ValueError('One of the samples in `non_uniform_samples` is'
+              ' not a float between -0.5 and 0.5')
```
kohr-h: This check will be very expensive for a large number of samples (explicit iteration). In view of my comment above, this kind of check should not be necessary.

The faster and more straightforward way of ensuring sanity is to (1) convert the samples to a NumPy array and (2) check its `shape` (should be `(N, ndim)`) and `dtype` (no complex, convert integer to float).

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        shape : tuple
+            The dimensions of the data whose non uniform FFT you want to
+            compute
+        non_uniform_samples : iterable
+            List of the fourier space positions where the coefficients are
+            computed. Each position must be in [-0.5; 0.5[
+        """
+        super(NonUniformFourierTransform, self).__init__(
+            shape=shape,
+            non_uniform_samples=non_uniform_samples,
```
kohr-h: `samples` should suffice

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+            domain=discr_sequence_space(shape, dtype=np.complex128),
+            range=discr_sequence_space(
+                [len(non_uniform_samples)],
+                dtype=np.complex128,
+            ),
+        )
+        self.adjoint_class = NonUniformFourierTransformAdjoint
```
kohr-h: Why store this?

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        x : `numpy.ndarray`
+            The results of the non uniform FFT
+
+        Returns
+        -------
+        out : `numpy.ndarray`
+            Normalized results of the non uniform FFT
+        """
+        out = x / np.sqrt(self.nfft.M)
+        return out
```
kohr-h: Could be done without copying.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        ----------
+        x : `numpy.ndarray`
+            The data whose non uniform FFT you want to compute
+
+        Returns
+        -------
+        out_normalized : `numpy.ndarray`
+            Result of the transform
+        """
+        self.nfft.f_hat = np.array(x)
```
kohr-h: `x.asarray()` does not copy. If `nfft` destroys the input, then we need a copy.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        out = self.nfft.trafo()
```
kohr-h: Does NFFT support pre-allocated `out`? If yes, would be worth using.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        out_normalized = self._normalize(out)
```
kohr-h: `out /= np.sqrt(self.nfft.M)`, no function required.

setup.cfg
```diff
@@ -39,8 +39,10 @@ classifiers =
 packages = find:
 install_requires =
     future >=0.14
+    cython
     packaging >=15.0
     numpy >=1.10,!=1.14.0,!=1.14.1,!=1.14.2
+    pyNFFT @ git+ssh://git@github.com/ghisvail/pyNFFT.git@e2da0af#egg=pyNFFT
```
kohr-h: Please don't make this a core dependency. It should be optional. Same goes for Cython.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+            The dimensions of the data whose non uniform FFT you want to
+            compute
+        non_uniform_samples : iterable
+            List of the fourier space positions where the coefficients are
+            computed. Each position must be in [-0.5; 0.5[
+        """
+        super(NonUniformFourierTransform, self).__init__(
+            shape=shape,
+            non_uniform_samples=non_uniform_samples,
+            domain=discr_sequence_space(shape, dtype=np.complex128),
```
kohr-h: This is too restrictive. It should be possible to use this operator with any `DiscreteLp` type domain, perhaps even with `cn` type spaces. Also other `dtype`s should be supported, even if the underlying library only supports double precision. In particular real-to-complex transforms should work. They don't have to be optimized but at least run.

My suggestion is that as with other operators, `domain` (for the forward transform) or `range` (for the adjoint) are parameters to `__init__`, and the range is inferred from that and the number of samples.

Further, I think the range should simply be `cn(num_samples)`, not a `DiscreteLp` The former is more general and reflects more the idea that the samples are unstructured.

odl/test/trafos/fourier_test.py
```diff
@@ -858,5 +859,98 @@ def fhat(x):
...
+        shape=shape,
+        non_uniform_samples=samples,
+    )
+    nfft_adj(nfft_coeffs)
+
+
+def test_non_uniform_fourier_trafo_1d_res():
+    sig_size = 512
+    shape = [sig_size]
+    samples = np.array(np.where(np.ones(shape) >= 0), dtype=float).T
```
kohr-h: This is a complicated way of writing `np.arange(sig_size)[:, None].astype(float)`.

kohr-h:
> [`pynfft`](https://github.com/pyNFFT/pyNFFT) doesn't seem to be maintained anymore.

Well, it's @grlee77 and myself who are maintainers now, so that can be improved. Maybe worth reviving that PR of mine...



> I am not completely sure about the redundancies between the travis config file and the setup config (I updated both for now).

The developer builds are a bit peculiar as you found out yourself, so we need a little bit of duplication in `.travis.yml` (explicitly install the deps). No problem anyway.

> I did some basic unit tests. If you want more please ask! (I also didn't refactor them that much, but I saw it was not the main concern)

Well currently there's not so much extra functionality to test :wink: Might change soon.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        shape : tuple
+            The dimensions of the data whose non uniform FFT you want to
+            compute
+        non_uniform_samples : iterable
+            List of the fourier space positions where the coefficients are
+            computed. Each position must be in [-0.5; 0.5[
```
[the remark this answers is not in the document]
zaccharieramzi: Ok makes sense!

I don't understand what you mean by normalised **based on Nyquist**. Is [this](https://github.com/CEA-COSMIC/pysap-mri/blob/master/mri/reconstruct/utils.py#L78-L105) what you mean?

odl/trafos/fourier.py
zaccharieramzi: hmm this way I could mutualise a bit the code for the adjoint between the 2 classes and put it in the base class. Do you think it's overcomplexified?

setup.cfg
zaccharieramzi: so should I just have it in travis and add in the docs some instruction on how to install the library? (also some checks to verify that the library is installed when using the classes)

odl/trafos/fourier.py
zaccharieramzi: Ok I will try to look a bit more into that.
Like I said I was not sure I really understood how to play with the spaces. If you know where to start that would be cool.
I basically understand that `cn` is the space of complex tensors. But aside from that it's a bit shdady (and I just tried to replicate what I saw in other operators simplified).

However, isn't it a burden that the user has to figure which domain to use? I see that it's what has been done so far so let's not change that now. I am asking more out of curiosity for why it has been done like this.

odl/trafos/fourier.py
zaccharieramzi: I haven't tried, but if so, why is it worth using?

odl/trafos/fourier.py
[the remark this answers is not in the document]
kohr-h: Yes that's the idea. Note, however, that there are different definitions of Fourier transform, and they result in different notions of "frequency". We go with the mathematician's definition that makes the transform unitary.
Concretely, for us the bandwidth of a uniformly spaced signal is `pi / s`, where `s` is the grid spacing. [Here](https://odlgroup.github.io/odl/math/trafos/fourier_transform.html) are some details. That means that given frequencies should be interpreted as lying in `[-pi/s, pi/s[` and normalized accordingly. (Note that `s` can be different per axis.)

NB: The term "Nyquist" is not mentioned there, I just threw it in sloppily to summarize that whole complex here :-).

odl/trafos/fourier.py
kohr-h: In general it's perfectly fine to factor out common code. But here we're talking about two lines of code maybe? I think it's more readable to inline the code in the two classes. (Also makes storing an attribute obsolete, which is always good.)

odl/trafos/fourier.py
```diff
@@ -7,9 +7,11 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 """Discretized Fourier transform on L^p spaces."""
-
 from __future__ import print_function, division, absolute_import
+from collections.abc import Iterable
+
 import numpy as np
+from pynfft.nfft import NFFT
```
adler-j: This needs to be include guarded so that users can run ODL without it. Further, I would personally put the nfft in a different file for now.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
             sign=sign, tmp_r=self._tmp_r, tmp_f=self._tmp_f)
 
 
+class NonUniformFourierTransformBase(Operator):
+    """Non uniform Fast Fourier Transform.
```
adler-j: "Fast" is an implementation detail which we typically do not include in headers in ODL.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+
+    The normalization is inspired from pysap-mri, mainly this class:
```
adler-j: This line should probably come later (e.g. as a comment?)

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+    """
+    def __init__(self, shape, non_uniform_samples, domain, range):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        shape : tuple
+            The dimensions of the data whose non uniform FFT you want to
+            compute
+        non_uniform_samples : iterable
```
adler-j: Does this need to be array-like? Or is iterable enough?

odl/trafos/fourier.py
kohr-h: Because it saves a copy. `out = self.nfft.trafo()` allocates new memory for `out`, whereas (hypothetically) `self.nfft_trafo(out=out)` just updates `out`, which should be faster.
Caveat: the implementation may be crappy and the latter may actually be slower than the former. Ideally you would benchmark the two variants.

Silly example (run in IPython):
```python
>>> x = np.zeros(int(1e7))
>>> %timeit out = np.exp(x)
79.6 ms ± 174 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
>>> out = np.empty_like(x)
>>> %timeit np.exp(x, out=out)
55.8 ms ± 309 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

odl/trafos/fourier.py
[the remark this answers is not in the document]
adler-j: The most important thing here is that we make sure that this is equivalent to our regular fourier transform if the points are selected in a uniformly spaced manner

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        self.shape = shape
+        self._check_samples(non_uniform_samples)
+        self.non_uniform_samples = non_uniform_samples
+        self.nfft = NFFT(N=shape, M=len(non_uniform_samples))
+        self.nfft.x = non_uniform_samples
+        self.nfft.precompute()
+        self.adjoint_class = None
+
+    def _check_samples(self, non_uniform_samples):
+        """Check that samples are correct
```
adler-j: This should be elaborated on

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        Parameters
+        ----------
+        non_uniform_samples : iterable
+            List of the fourier space positions where the coefficients are
+            computed
+
+        Raises
+        ------
+        ValueError :
+            If the non_uniform_samples is not an iterable, of 0 length or if
```
adler-j: Prefer to move these details up top and then simply have "if the input is not valid" down here.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+            one of the locations is not normalized in [-0.5; 0.5[
+        """
+        if not isinstance(non_uniform_samples, Iterable):
+            raise TypeError('`non_uniform_samples` is not iterable.')
+
+        if len(non_uniform_samples) == 0:
+            raise ValueError('`non_uniform_samples` is empty')
+
+        n_dim = len(self.shape)
+        if not all(len(sample) == n_dim for sample in non_uniform_samples):
```
adler-j: This whole thing is rather slow. Is there ever a reason where we could not make do with a simple array of positions instead of arbitrary iterables?

adler-j: E.g. in 3d this would take forever to perform a look in raw python.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        """
+        floats = all(isinstance(coord, float) for coord in sample)
+        if floats:
+            normalized = all(-0.5 <= coord < 0.5 for coord in sample)
+        else:
+            normalized = False
+        return floats & normalized
+
+    def _normalize(self, x):
+        """Normalize the result of the non uniform FFT.
```
adler-j: Please elaborate (on the row below) in what sense the normalization should be done.

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+        ----------
+        x : `numpy.ndarray`
+            The data whose non uniform FFT adjoint you want to compute
+
+        Returns
+        -------
+        out_normalized : `numpy.ndarray`
+            Result of the adjoint transform
+        """
+        self.nfft.f = np.array(x)
```
adler-j: Prefer `np.asarray` which (potentially) avoids a copy.

setup.cfg
adler-j: Indeed. Please have a look at how this has been done for Astra to get an idea.

kohr-h: In `setup.cfg` you can add them to the [`all`](https://github.com/odlgroup/odl/blob/5b576d700d902ce8ce6528bd86a65166d403a43d/setup.cfg#L67-L74) part of `[options.extras_require]`.
In Travis, yes, running unittests for NFFT is desirable. Just explicitly install the stuff you need.

For runtime checking, you can pretty much clone the `PYFFTW_AVAILABLE` section.

Instructions in the docs would be nice. (Note that there's also a package on `conda-forge`, that should probably the go-to solution for user installation. With pyFFTW we've also had build issues in the past and moved away from `pip` on Travis.)

odl/trafos/fourier.py
kohr-h: > However, isn't it a burden that the user has to figure which domain to use? I see that it's what has been done so far so let's not change that now. I am asking more out of curiosity for why it has been done like this.

It's a bit of a burden, I agree. But what you get in return is, for instance, a handle on the discretization that tells you exactly how to normalize frequency samples. I'd argue that's a net win. With additional helper functions that tell people what frequency range corresponds to their discretized function space, this operator becomes easier to user IMO. Apart from that, the space is usually dictated by the application anyway, so there's often not an additional step to be taken.
If you look through the [`examples`](https://github.com/odlgroup/odl/tree/master/examples) you will see that basically all of them start by defining a space (e.g., where images live), followed by code that just makes use of that. It's a very common pattern.

> I basically understand that `cn` is the space of complex tensors. But aside from that it's a bit shdady (and I just tried to replicate what I saw in other operators simplified).

For our case, the simplest variant of it applies, namely the space of N-long complex vectors. That's because the result of NFFT is basically just a list of complex numbers, corresponding to the Fourier samples at the chosen locations.
`DiscreteLp`, on the other hand, implies spatial structure: some kind of domain with a (rectangular) grid on it, where functions are discretized. We don't have that kind of structure here.

> Like I said I was not sure I really understood how to play with the spaces. If you know where to start that would be cool.

No worries, that's usually the hardest part to wrap your head around :-)
A good starting point would be this:

- Add a `space` as an argument and ensure it's a `DiscreteLp`. Also make sure it's uniformly discretized (`space.is_uniform`).
- Remove `shape` and take it from `space` instead.
- Take the per-axis `cell_sides` to compute the max frequencies.
- As for the samples, convert them to `np.ndarray` of a reasonable `dtype`. Check that their shape makes sense.
- Add a function for normalization of the frequencies based on the above computed max frequencies. Ideally, that computation would not run during `__init__` but during the first call to the operator, caching the normalized frequencies.

That should already take you quite far.

odl/trafos/fourier.py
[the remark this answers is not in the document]
kohr-h: Right, up to the additional factor that comes from the type of discretization (nearest/linear/...). IMO that has only lead to confusion in the past and should be removed from the existing FT.

odl/trafos/fourier.py
zaccharieramzi: I see. But in this case isn't it a bit more relevant since this implementation results in approximation?

odl/trafos/fourier.py
```diff
@@ -7,9 +7,11 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 """Discretized Fourier transform on L^p spaces."""
-
 from __future__ import print_function, division, absolute_import
+from collections.abc import Iterable
```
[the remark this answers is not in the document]
zaccharieramzi: Sure!

odl/trafos/fourier.py
zaccharieramzi: Sure !

odl/trafos/fourier.py
zaccharieramzi: Do you think I should still keep a `check` function at this point? Basically there is only one check, and that is the dimension.

odl/trafos/fourier.py
zaccharieramzi: Sure !

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+    """
+    def __init__(self, shape, non_uniform_samples):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        shape : tuple
+            The dimensions of the data whose non uniform FFT you want to
+            compute
+        non_uniform_samples : iterable
```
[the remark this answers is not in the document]
zaccharieramzi: Sure !

odl/trafos/fourier.py
zaccharieramzi: yes yes it's not that much. I just can't have it in the `__init__` because of the infinite recursion but even so it's not that bad.

odl/trafos/fourier.py
zaccharieramzi: Sure !

odl/trafos/fourier.py
zaccharieramzi: Sure !

odl/trafos/fourier.py
zaccharieramzi: Sure !

odl/trafos/fourier.py
zaccharieramzi: Yes it needs to be array-like, I will update that!

odl/trafos/fourier.py
zaccharieramzi: I think I will just remove that function for now. With all the simplifications it doesn't really make sense now.

odl/trafos/fourier.py
zaccharieramzi: See comment above.

odl/trafos/fourier.py
zaccharieramzi: Sure!

odl/trafos/fourier.py
zaccharieramzi: Sure!

odl/trafos/fourier.py
```diff
@@ -1645,6 +1647,211 @@ def inverse(self):
...
+            domain=domain,
+            range=range,
+            linear=True,
+        )
+        self.shape = shape
+        self._check_samples(non_uniform_samples)
+        self.non_uniform_samples = non_uniform_samples
+        self.nfft = NFFT(N=shape, M=len(non_uniform_samples))
+        self.nfft.x = non_uniform_samples
+        self.nfft.precompute()
```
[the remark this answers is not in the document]
zaccharieramzi: Sure!

odl/test/trafos/fourier_test.py
zaccharieramzi: I liked how it related to the previous tests, but I see how it can be disturbing. I will change it for `test_non_uniform_fourier_trafo_2d_res` as well.

.travis.yml
```diff
@@ -41,19 +41,27 @@ install:
       else
         wget -q https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
       fi
+    # Necessary to install pyNFFT
+    - if [ $TRAVIS_PYTHON_VERSION == "3.6" ]; then
+        export CPLUS_INCLUDE_PATH=/opt/python/3.6.3/include/python3.6m;
```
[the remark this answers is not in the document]
zaccharieramzi: I don't understand what you mean.

odl/test/trafos/fourier_test.py
```diff
@@ -858,5 +859,98 @@ def fhat(x):
...
 
 
+# ---- Non-uniform Fourier Transform ---- #
+def test_non_uniform_fourier_trafo_2d_call():
+    im_size = 512
+    shape = [im_size] * 2
+    samples = np.array(
+        np.where(np.random.normal(size=shape) >= 0),
+        dtype=float,
+    ).T
```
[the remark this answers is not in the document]
zaccharieramzi: Because you want the first dimension to correspond to the number of samples and the second the number of dimensions in your data. With `np.where` you get 3 lists, each corresponding to one dimension of your data.

odl/trafos/fourier.py
adler-j: Ok, then this is ignorance on my end. We actually get different results from the fast vs "normal" nfft?

zaccharieramzi: Well up to a certain accuracy yes.
> The generalisation of the FFT is an approximative algorithm and has computational complexity O (N log N + log (1/ε) M), where ε denotes the desired accuracy

odl/trafos/fourier.py
zaccharieramzi: I don't get the whole picture most certainly, but without any context, wouldn't a more fair benchmark be:

```
In [1]: import numpy as np                                                                                                                     

In [2]: x = np.zeros(int(1e7))                                                                                                                 

In [3]: %timeit out = np.exp(x)                                                                                                                
66.5 ms ± 215 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit np.exp(x, out=np.empty_like(x))                                                                                                
67.3 ms ± 172 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Maybe it's because in this case there is no copy but not sure. But anyway, I don't think `pynfft` supports pre-allocated `out` so I don't think we should delve into this too much (neither from the function `trafo` nor by setting `nfft.f` in advance in the case of the direct transform).

odl/trafos/fourier.py
zaccharieramzi: `nfft` doesn't destroy the input (just tested it).

odl/trafos/fourier.py
zaccharieramzi: I put the nfft in a different file. I applied the mechanism you used for `pywt` and `pyfftw` to `pynfft`, even if the bindings file is way smaller. Tell me what you think about it.

setup.cfg
zaccharieramzi: I am having some trouble using conda for now on my local machine, so I can't really test the local user installation. For now I have put the solution using `pip` but if you really want me to dig into the forge solution locally I can try and do it (it works on travis so it should locally as well).

odl/trafos/fourier.py
kohr-h: No, it's fine to just inline such a simple check.

odl/trafos/fourier.py
kohr-h: You can refer to the classes in the `adjoint` property directly. If you worry about efficiency (overhead of creating new operators, shouldn't be too much here, though) you can always cache the adjoint.

odl/trafos/fourier.py
kohr-h: The two versions that you compare should be practically indistinguishable.

My point was that you can make use of `out` if the called function supports it, and if it's implemented well it will not copy and typically run faster.
Often that's interesting when an operation is part of a chain of operations. For instance, if you want to compute `log(x / y)` there are basically two possibilities (assuming it's okay to mutate `x`):

- `z = np.log(x / y)` -- more readable, less efficient
  Actually this is equivalent to
  ```python
  >>> tmp1 = np.divide(x, y)  # new array
  >>> z = np.log(tmp1)  # new array
  ```
- `x /= y; np.log(x, out=x)` -- less readable, more efficient
  For analogy, you can also write this as
  ```python
  >>> np.divide(x, y, out=x)  # no new array
  >>> np.log(x, out=x)  # no new array
  ```

You will see a lot of code like the latter spread throughout ODL.

NB: With NumPy [temporary elision](https://docs.scipy.org/doc/numpy/release.html#temporary-elision) things are no longer that clear-cut, but I don't know how much of a difference it makes.



> But anyway, I don't think `pynfft` supports pre-allocated `out`

Hm, that would be worth adding



> neither from the function `trafo` nor by setting `nfft.f` in advance in the case of the direct transform -this I just tested -

Thanks for checking!

.travis.yml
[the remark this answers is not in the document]
kohr-h: I guess what he meant was that it spells out the path to Python and the exact version number, which will likely break when a new patch version comes out.
Perhaps there is a Travis environment variable for this?

doc/source/getting_started/installing_extensions.rst
```diff
@@ -30,6 +30,23 @@ Astra is most easily installed using conda:
...
+
+You can install `pynfft` with the following commands (on Linux):
+
+.. code-block:: bash
+
+    $ sudo apt install -y libnfft3-dev
+    $ export CPLUS_INCLUDE_PATH=/opt/python/<$FULL_PYTHON_VERSION>/include/<$MAJOR_MINOR_PYTHON_VERSION>m;
+    $ pip install cython numpy git+https://github.com/ghisvail/pyNFFT.git
+
+For further instructions, check `the pynff GitHub page <https://github.com/pyNFFT/pyNFFT>`_.
```
kohr-h: little typo "pynff"

doc/source/getting_started/installing_conda.rst
```diff
@@ -119,6 +119,23 @@ further instructions.
...
+
+        $ sudo apt install -y libnfft3-dev
+
+  * Install the python backend `pynfft`_ by running:
+
+    .. code-bock:: bash
+
+        $ export CPLUS_INCLUDE_PATH=/opt/python/<$FULL_PYTHON_VERSION>/include/<$MAJOR_MINOR_PYTHON_VERSION>m;
+        $ pip install numpy cython
+        $ pip install git+https://github.com/ghisvail/pyNFFT.git
```
kohr-h: This is the old repo link

doc/source/getting_started/installing_extensions.rst
zaccharieramzi: True!

doc/source/getting_started/installing_conda.rst
zaccharieramzi: True!

.travis.yml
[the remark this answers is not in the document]
zaccharieramzi: I don't think there is a Travis env var for this (cf [this](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)). I see however that you can [specify the patch](https://github.com/travis-ci/travis-ci/issues/7796) in the python version. Maybe it would be a good idea to set it, since after all, potentially other things might break with another patch change.

odl/trafos/fourier.py
zaccharieramzi: Yes yes definitely it's [what I did](https://github.com/odlgroup/odl/pull/1488/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.py#diff-7533defd16313b3c1ec23dd40014de57R144).

odl/trafos/fourier.py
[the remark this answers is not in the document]
zaccharieramzi: So I implemented the normalization [here](https://github.com/odlgroup/odl/pull/1488/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.py#diff-7533defd16313b3c1ec23dd40014de57R71). I took into account the fact that we could want different bandwidth on each dimension.

However I tested only against `numpy` fft for the uniformly spaced manner, because it seemed weird to me to test againsta `odl` fft. Do you want to have a test confirming consistency?

odl/trafos/fourier.py
zaccharieramzi: Do you mean how the normalization should be done or why?

odl/trafos/fourier.py
zaccharieramzi: Ok I see exactly what you wanted me to do and I think I have implemented it.
However, I didn't take `cell_sides` but `min_pt` and `max_pt` which, I think, are way better suited in this case. I just don't know if they are deprecated.

The examples did help understand what was going on.

adler-j:
We should get this great contribution finalized so we can get it it, I'm looking forward to it. Do you know what's left @kohr-h?

kohr-h:
@adler-j Main thing is the way the operator domain is handled and how the frequencies are normalized. I'll make a PR into this.

kohr-h:
@zaccharieramzi Sorry for the delay, I was on holiday.
I've drafted a PR to your branch, with explanations added: https://github.com/zaccharieramzi/odl/pull/1. Let me know if you have questions.

zaccharieramzi:
Hi @kohr-h @adler-j ! I wanted to let you know that I will not be working on this PR anymore.
I actually found another way of getting the Non-uniform fast fourier transform in  TensorFlow, by translating [`torchkbnufft`](https://github.com/mmuckley/torchkbnufft) in tf. The result is available [here](https://github.com/mmuckley/torchkbnufft) and is installable by a simple `pip install tfkbnufft`, if you ever try it please give me some feedback I would be grateful.

I am closing this for now, but of course feel free to finish this in your own style. I just feel it will be less messy for you to have one less PR open.

kohr-h:
Thanks @zaccharieramzi for taking the trouble to do this "clean-up", and thanks for your contribution! We still have this PR as a good base to start from when we want to implement NUFFT later on.

PR #1490, opened 2019-04-15
ENH: catch warnings in valid codepaths of KL functionals

kohr-h:
Small fix that gets rid of the obnoxious warnings in the KL functionals.

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -1125,17 +1125,18 @@ class KullbackLeibler(Functional):
         # Lazy import to improve `import odl` time
         import scipy.special
 
-        if self.prior is None:
-            tmp = ((x - 1 - np.log(x)).inner(self.domain.one()))
-        else:
-            tmp = ((x - self.prior +
-                    scipy.special.xlogy(self.prior, self.prior / x))
-                   .inner(self.domain.one()))
-        if np.isnan(tmp):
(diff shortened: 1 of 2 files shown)


odl/solvers/functional/default_functionals.py
```diff
@@ -1258,17 +1259,18 @@ def _call(self, x):
...
+        with np.errstate(invalid='ignore'):
+            if self.prior is None:
+                res = -(np.log(1 - x)).inner(self.domain.one())
+            else:
+                xlogy = scipy.special.xlogy(self.prior, 1 - x)
+                res = -self.domain.element(xlogy).inner(self.domain.one())
+
+        if not np.isfinite(res):
             # In this case, some element was larger than or equal to one
             return np.inf
```
adler-j: This has to be wrong, right? The gradient is a space element.

kohr-h: I see what you mean, issue of display :-) The folded part is all the way from `KullbackLeibler.gradient` to `KullbackLeiblerCrossEntropy._call`.
So this code is part of `KullbackLeiblerCrossEntropy._call`.

adler-j: Ok, then I'm good!

kohr-h:
> Fix PEP and go!

Those are unfixable too long doc lines

    odl.solvers.nonsmooth.proximal_operators.proximal_convex_conj_kl_cross_entropy :

I'll leave them alone.

Issue #1500, opened 2019-05-18
Fix for CI issue with NumPy <1.13.3


kohr-h:
The new version 1.3 of SciPy requires NumPy 1.13.3 or higher. That breaks a couple of our CI pipelines that test for older versions of NumPy. How I see it we have these choices to unbreak our CI:

- Use `scipy <1.3` in CI
- Change package dependency to `scipy <1.3`
- Only test NumPy 1.13.3 and above in CI
- Change package dependency to `numpy >= 1.13.3`.

If we go for dropping old NumPy, I'd also like that we consider going to 1.14 directly. That version introduced improved (more compact) printing of arrays which we currently disable by setting `numpy.printoptions(legacy="1.13")`. At some point we'll have to switch anyway, I don't know how many versions will keep that option. Of course, that will require to change a bunch of doctests.

PR #1503, opened 2019-05-30
MAINT: remove BOMs

kohr-h:
Trivial PR that removes those annoying BOM unicode characters. Some tools get tripped by them.

(For the modified files, I've also updated the copyright years.)

### DIFF
--- a/examples/solvers/deconvolution_1d.py
+++ b/examples/solvers/deconvolution_1d.py
@@ -1,4 +1,4 @@
-﻿"""Example of a deconvolution problem with different solvers (CPU)."""
+"""Example of a deconvolution problem with different solvers (CPU)."""
 
 import numpy as np
 import matplotlib.pyplot as plt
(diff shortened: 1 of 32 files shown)


kohr-h:
CI failures again due to incompatible Numpy, see #1500

adler-j:
what really changed here? I'm only seeing copyright years and a single other change.

kohr-h:
> what really changed here? I'm only seeing copyright years and a single other change.

The BOM characters are removed. They are invisible and don't show up here. But if you look at the git diff, for instance, you see this:

    -<U+FEFF># Copyright 2014-2017 The ODL contributors
    +# Copyright 2014-2019 The ODL contributors

These marks are only useful in Python 2, and tend to be added by Windows editors. We usually don't need unicode support in source files, and if we do, we make it explicit by some other measure.

adler-j:
All good then, go ahead!

PR #1504, opened 2019-05-30
ENH: Use contextlib to make context managers

kohr-h:
Using `contextlib.contextmanager`, we can scrap the boilerplate stuff (`__enter__` and `__exit__`, `__init__`) and write context managers in a much more straightforward way with plain functions and `try...finally`. A side benefit is that their usage looks more like a function call than a class creation.

All of our current custom context managers have been ported.

### DIFF
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -11,6 +11,7 @@
 from __future__ import absolute_import, division, print_function
 from future.moves.itertools import zip_longest
 
+from contextlib import contextmanager
 import os
 import sys
 import warnings
@@ -36,8 +37,8 @@ __all__ = (
     'noise_array',
     'noise_element',
     'noise_elements',
-    'FailCounter',
-    'Timer',
+    'fail_counter',
+    'timer',
     'timeit',
     'ProgressBar',
     'ProgressRange',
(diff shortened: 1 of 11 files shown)


odl/util/testutils.py
```diff
@@ -459,80 +460,78 @@ def noise_elements(space, n=1):
...
+        *** FAILED 1 TEST CASE(S) ***
     """
 
-    def __init__(self, test_name, err_msg=None, logger=print):
-        self.num_failed = 0
-        self.test_name = test_name
-        self.err_msg = err_msg
-        self.fail_strings = []
-        self.log = logger
+    class _Counter(object):
```
adler-j: _FailCounter perhaps?

odl/util/testutils.py
```diff
@@ -459,80 +460,78 @@ def noise_elements(space, n=1):
...
+    tstart = None
+    try:
+        tstart = time()
+        yield
 
-    def __exit__(self, type, value, traceback):
-        time_str = '{:.3f}'.format(time() - self.tstart)
-        print('{:>30s} : {:>10s} '.format(self.name, time_str))
+    finally:
+        if tstart is not None:
```
adler-j: Could this ever not happen?

odl/util/utility.py
```diff
@@ -1533,41 +1534,37 @@ def pkg_supports(feature, pkg_version, pkg_feat_dict):
...
+    >>> rand_int == same_rand_int
+    True
+    """
+    do_seed = seed is not None
+    try:
+        if do_seed:
+            orig_rng_state = np.random.get_state()
+            np.random.seed(seed)
+        yield
+
```
adler-j: Remove blankline here

odl/util/testutils.py
kohr-h: Yeah I'm very cautious here. If `time()` fails without raising an exception, then we're in serious trouble anyway.

odl/util/testutils.py
kohr-h: Sure!

odl/util/testutils.py
kohr-h: Removed that precaution, `time` says it returns a float.

PR #1509, opened 2019-07-05
Flying Focal Spot for FanBeam and ConeBeam geometries

JevgenijaAksjonova:
Added possibility to arbitrary shift source positions. Since ASTRA allows to provide arbitrary source positions to projection and back-projection functions, it is possible now to do reconstruction with flying focal spot.

Implementation:
The sequence of unique shifts should be defined relative to the default source position and provided as an input to the geometry. Then this sequence is extended for all angles and returned when calling geometry.flying_focal_spot. However, one has to manually provide flying_focal_spot sequence to geometry.src_position() as an argument. This choice was made because FFS shifts correspond to discretization of angular partition and defining a FFS shift for any angle would introduce significant overhead in terms of work (moreover, in practice this is not needed).

### DIFF
--- a/odl/test/tomo/geometry/geometry_test.py
+++ b/odl/test/tomo/geometry/geometry_test.py
@@ -16,7 +16,6 @@ import numpy as np
 
 import odl
 from odl.util.testutils import all_almost_equal, all_equal, simple_fixture
-from odl.tomo.util import flying_focal_spot
 
 
 # --- pytest fixtures --- #
@@ -565,7 +564,9 @@ def test_fanbeam_flying_focal_spot(init1=None):
     shift2 = np.array([-2.0, 3.0])
     init = np.array([1, 0], dtype=np.float32)
 
-    ffs = partial(flying_focal_spot, apart=apart, shifts=[shift1, shift2])
+    ffs = partial(odl.tomo.flying_focal_spot,
+                  apart=apart,
+                  shifts=[shift1, shift2])
     geom_ffs = odl.tomo.FanBeamGeometry(
         apart, dpart,
         src_rad, det_rad,
@@ -750,7 +751,9 @@ def test_conebeam_flying_focal_spot():
     shift1 = np.array([2.0, -3.0, 1.0])
     shift2 = np.array([-2.0, 3.0, -1.0])
     init = np.array([1, 0, 0], dtype=np.float32)
-    ffs = partial(flying_focal_spot, apart=apart, shifts=[shift1, shift2])
+    ffs = partial(odl.tomo.flying_focal_spot,
+                  apart=apart,
+                  shifts=[shift1, shift2])
     geom_ffs = odl.tomo.ConeBeamGeometry(
         apart, dpart,
         src_rad, det_rad,
@@ -946,6 +949,42 @@ def test_helical_geometry_helper():
     mag = (3 + 9) / (3 + rho)
     delta_h = space.cell_sides[2] * mag
     assert geometry.det_partition.cell_sides[1] <= delta_h
+
+
+def test_source_detector_shifts():
+    """Test source-detector shift functions, e.g. flying focal spot.
+
+    See the `flying_focal_spot` documentation for the exact conditions.
+    """
+    n_angles = np.random.randint(1, 100)
+    apart = odl.uniform_partition(0, np.pi, n_angles)
+    part_angles = apart.meshgrid[0]
+
+    # shifts are periodic
+    def check_shifts(ffs, shifts):
+        i = 0
+        while i < part_angles.size:
+            j = min(len(ffs), i + len(shifts))
+            assert all_almost_equal(ffs[i:j], shifts[:(j - i)])
+            i = j
+
+    # shifts define ffs at partition points
+    n_shifts = np.random.randint(1, n_angles)
+    shift_dim = 3
+    shifts = np.random.uniform(size=(n_shifts, shift_dim))
+    ffs = odl.tomo.flying_focal_spot(part_angles, apart, shifts)
+    check_shifts(ffs, shifts)
+
+    shift_dim = 2
+    shifts = np.random.uniform(size=(n_shifts, shift_dim))
+    ffs = odl.tomo.flying_focal_spot(part_angles, apart, shifts)
+    check_shifts(ffs, shifts)
+
+    # shifts at other angles ar defined by nearest neighbor interpolation
+    d = np.random.uniform(-0.49, 0.49) * apart.cell_volume
+    shifts = np.random.uniform(size=(n_shifts, shift_dim))
+    ffs = odl.tomo.flying_focal_spot(part_angles + d, apart, shifts)
+    check_shifts(ffs, shifts)
 
 
 if __name__ == '__main__':
--- a/odl/test/tomo/geometry/geometry_test.py
+++ b/odl/test/tomo/geometry/geometry_test.py
@@ -550,8 +550,8 @@ def test_fanbeam_frommatrix():
                                                    det_rad, sing_mat)
 
 
-def test_fanbeam_flying_focal_spot(init1=None):
-    """Test the flying focal spot in 2d fan beam geometry."""
+def test_fanbeam_src_det_shifts(init1=None):
+    """Test the source/detector shifts in 2d fan beam geometry."""
     full_angle = np.pi
     n_angles = 2 * 6
     apart = odl.uniform_partition(0, full_angle, n_angles)
@@ -567,11 +567,10 @@ def test_fanbeam_flying_focal_spot(init1=None):
     ffs = partial(odl.tomo.flying_focal_spot,
                   apart=apart,
                   shifts=[shift1, shift2])
-    geom_ffs = odl.tomo.FanBeamGeometry(
-        apart, dpart,
-        src_rad, det_rad,
-        src_to_det_init=init,
-        src_shift_func=ffs)
+    geom_ffs = odl.tomo.FanBeamGeometry(apart, dpart,
+                                        src_rad, det_rad,
+                                        src_to_det_init=init,
+                                        src_shift_func=ffs)
     # angles must be shifted to match discretization of apart
     ang1 = -full_angle / (n_angles * 2)
     apart1 = odl.uniform_partition(ang1, full_angle + ang1, n_angles // 2)
@@ -600,6 +599,31 @@ def test_fanbeam_flying_focal_spot(init1=None):
                                     src_to_det_init=init)
     assert all_almost_equal(geom.det_refpoint(geom.angles),
                             geom_ffs.det_refpoint(geom_ffs.angles))
+
+    # However, detector can be shifted similarly as the source
+    def det_shift(angle):
+        return ffs(angle) / src_rad * det_rad
+    geom_ds = odl.tomo.FanBeamGeometry(
+        apart, dpart,
+        src_rad, det_rad,
+        src_to_det_init=init,
+        det_shift_func=det_shift)
+    det_rad1 = src_rad1 / src_rad * det_rad
+    det_rad2 = src_rad2 / src_rad * det_rad
+    geom1 = odl.tomo.FanBeamGeometry(apart1, dpart, src_rad, det_rad1,
+                                     src_to_det_init=init1)
+    geom2 = odl.tomo.FanBeamGeometry(apart2, dpart, src_rad, det_rad2,
+                                     src_to_det_init=init2)
+    dr1 = geom1.det_refpoint(geom1.angles)
+    dr2 = geom2.det_refpoint(geom2.angles)
+    dr = geom_ds.det_refpoint(geom_ds.angles)
+    assert all_almost_equal(dr[0::2], dr1)
+    assert all_almost_equal(dr[1::2], dr2)
+
+    # source positions are not affected
+    assert all_almost_equal(geom.src_position(geom.angles),
+                            geom_ds.src_position(geom_ds.angles))
+
 
 def test_helical_cone_beam_props(detector_type, shift):
     """Test basic properties of 3D helical cone beam geometries."""
@@ -738,14 +762,15 @@ def test_helical_cone_beam_props(detector_type, shift):
     assert repr(geom)
 
 
-def test_conebeam_flying_focal_spot():
-    """Test the flying focal spot in 3d cone beam geometry."""
+def test_conebeam_source_detector_shifts():
+    """Test source/detector shifts in 3d cone beam geometry."""
     full_angle = np.pi
     n_angles = 2 * 7
     apart = odl.uniform_partition(0, full_angle, n_angles)
     dpart = odl.uniform_partition([-1, -1], [1, 1], (10, 10))
     src_rad = 10
     det_rad = 5
+    pitch = 3
     # Source positions with flying focal spot should correspond to
     # source positions of 2 geometries with different starting positions
     shift1 = np.array([2.0, -3.0, 1.0])
@@ -754,11 +779,11 @@ def test_conebeam_flying_focal_spot():
     ffs = partial(odl.tomo.flying_focal_spot,
                   apart=apart,
                   shifts=[shift1, shift2])
-    geom_ffs = odl.tomo.ConeBeamGeometry(
-        apart, dpart,
-        src_rad, det_rad,
-        src_to_det_init=init,
-        src_shift_func=ffs)
+    geom_ffs = odl.tomo.ConeBeamGeometry(apart, dpart,
+                                         src_rad, det_rad,
+                                         src_to_det_init=init,
+                                         src_shift_func=ffs,
+                                         pitch=pitch)
     # angles must be shifted to match discretization of apart
     ang1 = -full_angle / (n_angles * 2)
     apart1 = odl.uniform_partition(ang1, full_angle + ang1, n_angles // 2)
@@ -772,10 +797,12 @@ def test_conebeam_flying_focal_spot():
     src_rad2 = np.linalg.norm(np.array([src_rad + shift2[0], shift2[1], 0]))
     geom1 = odl.tomo.ConeBeamGeometry(apart1, dpart, src_rad1, det_rad,
                                       src_to_det_init=init1,
-                                      offset_along_axis=shift1[2])
+                                      offset_along_axis=shift1[2],
+                                      pitch=pitch)
     geom2 = odl.tomo.ConeBeamGeometry(apart2, dpart, src_rad2, det_rad,
                                       src_to_det_init=init2,
-                                      offset_along_axis=shift2[2])
+                                      offset_along_axis=shift2[2],
+                                      pitch=pitch)
 
     sp1 = geom1.src_position(geom1.angles)
     sp2 = geom2.src_position(geom2.angles)
@@ -786,9 +813,39 @@ def test_conebeam_flying_focal_spot():
     # detector positions are not affected by flying focal spot
     geom = odl.tomo.ConeBeamGeometry(apart, dpart,
                                      src_rad, det_rad,
-                                     src_to_det_init=init)
+                                     src_to_det_init=init,
+                                     pitch=pitch)
     assert all_almost_equal(geom.det_refpoint(geom.angles),
                             geom_ffs.det_refpoint(geom_ffs.angles))
+
+    # However, detector can be shifted similarly as the source
+    coef = det_rad / src_rad
+    def det_shift(angle):
+        return ffs(angle) * coef
+    geom_ds = odl.tomo.ConeBeamGeometry(apart, dpart,
+                                        src_rad, det_rad,
+                                        src_to_det_init=init,
+                                        det_shift_func=det_shift,
+                                        pitch=pitch)
+    det_rad1 = src_rad1 / src_rad * det_rad
+    det_rad2 = src_rad2 / src_rad * det_rad
+    geom1 = odl.tomo.ConeBeamGeometry(apart1, dpart, src_rad, det_rad1,
+                                      src_to_det_init=init1,
+                                      offset_along_axis=shift1[2] * coef,
+                                      pitch=pitch)
+    geom2 = odl.tomo.ConeBeamGeometry(apart2, dpart, src_rad, det_rad2,
+                                      src_to_det_init=init2,
+                                      offset_along_axis=shift2[2] * coef,
+                                      pitch=pitch)
+    dr1 = geom1.det_refpoint(geom1.angles)
+    dr2 = geom2.det_refpoint(geom2.angles)
+    dr = geom_ds.det_refpoint(geom_ds.angles)
+    assert all_almost_equal(dr[0::2], dr1)
+    assert all_almost_equal(dr[1::2], dr2)
+
+    # source positions are not affected
+    assert all_almost_equal(geom.src_position(geom.angles),
+                            geom_ds.src_position(geom_ds.angles))
 
 
 def test_cone_beam_slanted_detector():
--- a/odl/test/tomo/geometry/geometry_test.py
+++ b/odl/test/tomo/geometry/geometry_test.py
@@ -549,6 +549,45 @@ def test_fanbeam_frommatrix():
                                                    det_rad, sing_mat)
 
 
+def test_fanbeam_flying_focal_spot():
+    """Test the flying focal spot in 2d fan beam geometry."""
+    full_angle = np.pi
+    apart = odl.uniform_partition(0, full_angle, 10)
+    dpart = odl.uniform_partition(0, 1, 10)
+    src_rad = 10
+    det_rad = 5
+    # Source positions with flying focal spot should correspond to
+    # source positions of 2 geometries with different starting positions
+    shift1 = np.array([2, -3])
+    shift2 = np.array([-2, 3])
+    init = np.array([1, 0], dtype=np.float32)
+    geom_ffs = odl.tomo.FanBeamGeometry(apart, dpart, src_rad, det_rad,
+                                        src_to_det_init=init,
+                                        flying_focal_spot=[shift1,
+                                                           shift2])
+    # angles must be shifted to match discretization of apart
+    ang1 = -full_angle / 20
+    apart1 = odl.uniform_partition(ang1, full_angle + ang1, 5)
+    ang2 = full_angle / 20
+    apart2 = odl.uniform_partition(ang2, full_angle + ang2, 5)
+
+    init1 = init + np.array([0, shift1[1]]) / (src_rad + shift1[0])
+    init2 = init + np.array([0, shift2[1]]) / (src_rad + shift2[0])
+    # radius also changes when a shift is applied
+    src_rad1 = np.linalg.norm(np.array([src_rad, 0]) + shift1)
+    src_rad2 = np.linalg.norm(np.array([src_rad, 0]) + shift2)
+    geom1 = odl.tomo.FanBeamGeometry(apart1, dpart, src_rad1, det_rad,
+                                     src_to_det_init=init1)
+    geom2 = odl.tomo.FanBeamGeometry(apart2, dpart, src_rad2, det_rad,
+                                     src_to_det_init=init2)
+
+    sp1 = geom1.src_position(geom1.angles)
+    sp2 = geom2.src_position(geom2.angles)
+    sp = geom_ffs.src_position(geom_ffs.angles, geom_ffs.flying_focal_spot)
+    assert all_almost_equal(sp[0::2], sp1)
+    assert all_almost_equal(sp[1::2], sp2)
+
+
 def test_helical_cone_beam_props(detector_type, shift):
     """Test basic properties of 3D helical cone beam geometries."""
     full_angle = 2 * np.pi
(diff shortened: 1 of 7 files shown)


JevgenijaAksjonova:
Hi! Thank you for the review @kohr-h! The main issue with a function of the type src_shift_func(angle, geometry) is that it is not clear how to define a shift for an arbitrary angle. Basically, FFS is defined for a sequence of angles that correspond to angular partition. One could use interpolation to define it for any angle, but it seems a bit strange to me to do it just for flexibility of the interface. Or maybe it is not strange? What do you think?

JevgenijaAksjonova:
Why is it necessary to define FFS for arbitrary angle?

kohr-h:
> Why is it necessary to define FFS for arbitrary an angle?

In practice it isn't, but you still had to do it because the methods require it:

https://github.com/odlgroup/odl/blob/7fef8ad009b842e8dee21cb1aca11728f8afce07/odl/tomo/geometry/conebeam.py#L405-L407

Other kinds of shifts would look similar.

JevgenijaAksjonova:
Hi! I have been thinking about this and I am not convinced that FFS should be a function of the rotation angle. Because I imagine that in practical CT scanners both FFS and rotation angle depend on time. Since we don't have a notion of time in the geometry, we could use the rotation angle as a proxy for time, but I don't see how this would make life easier for people who are trying to define arbitrary shifts.
On the other hand, providing FFS as a sequence implicitly defines it as a function of the time step (the moment when projection is made). Just to be clear - the sequence can be arbitrary long, so if one actually has a function FFS(angle), he can just pass a sequence FFS(geometry.angles).

Secondly, even if we define FFS(angle), I don't think that we should allow an arbitrary angle. Firstly, I was concerned only about introducing unnecessary complexity in the code. Now I am thinking that we actually don't know how the source moves between the shifts, e.g. if the speed is uniform, if it is not, doing linear interpolation is wrong and can lead to "silent errors". To circumvent this issue, we could define FFS only for angles that belong to discretization and rise "Not implemented" or "Invalid argument error", if the provided angle does not belong to angular partition.

kohr-h:
> Hi! I have been thinking about this and I am not convinced that FFS should be a function of the rotation angle. Because I imagine that in practical CT scanners both FFS and rotation angle depend on time. Since we don't have a notion of time in the geometry

That's a valid concern. However, note that geometries have a notion of "motion parameter", which usually is, but doesn't have to be, a rotation angle. It can be anything that parametrizes the part of the geometry that is not the detector coordinate. So it could easily be time. The `angles` property is there purely for convenience, since it's usually the same as `motion_params.coord_vectors[0]`.



> On the other hand, providing FFS as a sequence implicitly defines it as a function of the time step (the moment when projection is made). Just to be clear - the sequence can be arbitrary long, so if one actually has a function FFS(angle), he can just pass a sequence FFS(geometry.angles).

The problem with this definition of FFS is that it "sticks" to the time steps, i.e., if you decide to double the frequency of the projections being taken, the frequency of the FFS will also double. It's not a function of continuous time, but of discrete time points. If that's the correct behavior, then okay.



> Now I am thinking that we actually don't know how the source moves between the shifts, e.g. if the speed is uniform, if it is not, doing linear interpolation is wrong and can lead to "silent errors".

That's correct, and a very valid concern.



> To circumvent this issue, we could define FFS only for angles that belong to discretization and rise "Not implemented" or "Invalid argument error", if the provided angle does not belong to angular partition.

That suggestion is impractical, though, because you need to compare floating point values to do the check, and floating point numbers are practically never equal. Checking for "almost equal" is also tricky, in particular around 0. So raising errors for invalid input will inevitably generate annoyance.

The best thing I can think of here is to do nearest neighbor interpolation, i.e., to always "snap" to the known values. That gives correct results for the given angles, and still allows arbitrary input.

JevgenijaAksjonova:
Sorry, it took such a long time, I have added a function flying_focal_spot in a separate file, which just snaps to the known values. It is possible to provide this function as an argument to geometry.

jonasteuwen:
What is the status of this issue?

ozanoktem:
> What is the status of this issue?

@JevgenijaAksjonova, could you provide an update please?

kohr-h:
Oh dear. If I remember correctly this is just a final review short of being merged :grimacing:

I'll give it a fresh look asap.

jonasteuwen:
It would be very convenient, it appears that this new Mayo clinic data has it for some of the scans!

adler-j:
Oh did Mayo finally release their data, do you have a link? I was buggering them for years.

ozanoktem:
Mayo Clinic announced the public availability of 299 patient CT cases containing head, chest and abdomen exams from Siemens and GE scanners. All cases have routine dose and simulated low dose projection data, full dose images, and clinical data identifying all pathology.
 
They appreciate receiving a copy of any publications that make use of these data.
 
Detailed information and a link to the data library are available here: https://ctcicblog.mayo.edu/hubcap/patient-ct-projection-data-library/.

jonasteuwen:
I did manage to read their data in python, but they provide this FFS in some of them. Would be good if this gets implemented, then I can publish some code giving a baseline.

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,48 @@
...
+
+from __future__ import print_function, division, absolute_import
+import numpy as np
+from odl.discr import uniform_discr_frompartition
+
+__all__ = ('flying_focal_spot')
+
+
+def flying_focal_spot(angle, apart, shifts):
+    """Flying focal spot shifts in the geometry.
```
kohr-h, line 19:
```suggestion
    """Flying focal spot shifts for divergent beam geometries.
```

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,48 @@
...
+    ----------
+
+    angle : float or `array-like`
+        Angle(s) in radians describing the counter-clockwise
+        rotation of source and detector.
+
+    apart : 1-dim. `RectPartition`
+        Partition of the angle interval.
+
+    shifts : sequence of `array-like`'s
```
kohr-h, line 31:
```suggestion
    shifts : sequence of `array-like`
```

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    angle = np.array(angle, dtype=float, copy=False, ndmin=1)
+    angle_space = uniform_discr_frompartition(apart, interp='nearest')
+    angle_ind = angle_space.element(range(apart.size))
+    ind = angle_ind.interpolation(angle)
+
+    shifts = np.array(shifts, dtype=float, ndmin=2)
+    if shifts.shape[1] not in [2, 3]:
+        raise ValueError('Flying focal spot shifts {} must have '
+                         'shape (2,) or (3,)'.format(shifts))
```
kohr-h, line 45: Better show the shape of the given array in the error message.

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,48 @@
...
+    shifts : sequence of `array-like`'s
+        Each vectors in a sequence represent a subsequent shift
+        relative to the default source position. Vector elements
+        represent shifts along the following directions:
+        det_to_src, tangent to the rotation, rotation axis.
+    """
+    angle = np.array(angle, dtype=float, copy=False, ndmin=1)
+    angle_space = uniform_discr_frompartition(apart, interp='nearest')
+    angle_ind = angle_space.element(range(apart.size))
+    ind = angle_ind.interpolation(angle)
```
kohr-h, line 40: Needs to be adapted to the changes in #1459

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2014-2019 The ODL contributors
```
kohr-h, line 1:
```suggestion
# Copyright 2014-2020 The ODL contributors
```

odl/tomo/geometry/conebeam.py
```diff
@@ -1195,24 +1309,57 @@ def src_position(self, angle):
...
+                     + np.multiply.outer(src_shifts[:, 1], tangent))
+        center_to_src_init = center_to_src_init + ffs_shift
+        # broadcasting to perform matrix multiplication "manually",
+        # since existing numpy functions do cross product along the outer
+        # dimensions, which we don't need here
+        center_to_src_init = np.repeat(center_to_src_init, 3,
+                                       axis=-2).reshape(-1, 3, 3)
         # `circle_component` has shape (a, ndim)
-        circle_component = rot_matrix.dot(center_to_src_init)
+        circle_component = np.sum(rot_matrix * center_to_src_init, axis=-1)
```
kohr-h, line 1355: Can't we do that with `np.matmul` somehow? It behaves differently from `np.dot`.

JevgenijaAksjonova: I don't think so, since the arrays are 3 dimensional here

rot_matrix has dimension (n_angles, 3, 3)
center_to_src_init initially has dimension (n_angles, 3)

For each angles we need to multiply a vector by a matrix, but we don't need to "cross- multiply" vectors and matrices corresponding to different angles. Documentation for matmul says:
"Stacks of matrices are broadcast together as if the matrices were elements, respecting the signature (n,k),(k,m)->(n,m):"

So, it does not handle 3d arrays the way we need it here.

JevgenijaAksjonova: I agree that the way I did it looks clumsy, it took me quite some time to understand what I did there and why =D

kohr-h: Haha, no worries. If it works and there's no obvious easier way it's fine. I was just wondering.

JevgenijaAksjonova:
Hi! Are there more changes required?

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -549,6 +550,48 @@ def test_fanbeam_frommatrix():
                                                    det_rad, sing_mat)
 
 
+def test_fanbeam_flying_focal_spot():
+    """Test the flying focal spot in 2d fan beam geometry."""
+    full_angle = np.pi
+    apart = odl.uniform_partition(0, full_angle, 10)
+    dpart = odl.uniform_partition(0, 1, 10)
```
adler-j, line 557: This is a non-centered detector, rather weird but I guess it's OK for a test.

adler-j, line 557: Avoid using the same magic number everywhere in tests (here: 10) since it can hide bugs. It's usually recommended to use different prime numbers.

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -549,6 +550,48 @@ def test_fanbeam_frommatrix():
...
+    init = np.array([1, 0], dtype=np.float32)
+    geom_ffs = odl.tomo.FanBeamGeometry(
+        apart, dpart,
+        src_rad, det_rad,
+        src_to_det_init=init,
+        src_shift_func=lambda angle: flying_focal_spot(
+            angle, apart=apart,
+            shifts=[shift1, shift2]))
+    # angles must be shifted to match discretization of apart
+    ang1 = -full_angle / 20
```
adler-j, line 573: 20 is a magic number here, create a variable for it.

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -549,6 +550,48 @@ def test_fanbeam_frommatrix():
...
+    # Source positions with flying focal spot should correspond to
+    # source positions of 2 geometries with different starting positions
+    shift1 = np.array([2, -3])
+    shift2 = np.array([-2, 3])
+    init = np.array([1, 0], dtype=np.float32)
+    geom_ffs = odl.tomo.FanBeamGeometry(
+        apart, dpart,
+        src_rad, det_rad,
+        src_to_det_init=init,
+        src_shift_func=lambda angle: flying_focal_spot(
```
adler-j, line 569: This lambda is rather big, prefer creating a function outside or using functools.partial

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -549,6 +550,48 @@ def test_fanbeam_frommatrix():
...
+                                     src_to_det_init=init1)
+    geom2 = odl.tomo.FanBeamGeometry(apart2, dpart, src_rad2, det_rad,
+                                     src_to_det_init=init2)
+
+    sp1 = geom1.src_position(geom1.angles)
+    sp2 = geom2.src_position(geom2.angles)
+    sp = geom_ffs.src_position(geom_ffs.angles)
+    assert all_almost_equal(sp[0::2], sp1)
+    assert all_almost_equal(sp[1::2], sp2)
+
```
adler-j, line 593: Check that e.g. the detector positions are correct as well.

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -549,6 +550,48 @@ def test_fanbeam_frommatrix():
...
+
```
adler-j, line 594: Also add a test where this is actually used so we see that it interacts well with the backend.

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -686,6 +729,50 @@ def test_helical_cone_beam_props(detector_type, shift):
     assert repr(geom)
 
 
+def test_conebeam_flying_focal_spot():
```
adler-j, line 732: Same comments as above.

odl/test/tomo/geometry/geometry_test.py
```diff
@@ -549,6 +550,48 @@ def test_fanbeam_frommatrix():
...
+def test_fanbeam_flying_focal_spot():
+    """Test the flying focal spot in 2d fan beam geometry."""
+    full_angle = np.pi
+    apart = odl.uniform_partition(0, full_angle, 10)
+    dpart = odl.uniform_partition(0, 1, 10)
+    src_rad = 10
+    det_rad = 5
+    # Source positions with flying focal spot should correspond to
+    # source positions of 2 geometries with different starting positions
+    shift1 = np.array([2, -3])
```
adler-j, line 562: This is an array of ints, change to floats

odl/tomo/geometry/conebeam.py
```diff
@@ -159,6 +166,17 @@ def __init__(self, apart, dpart, src_radius, det_radius,
...
+        Specifying a flying focal spot:
+
+        >>> apart = odl.uniform_partition(0, 2 * np.pi, 4)
+        >>> geom = FanBeamGeometry(apart, dpart, src_radius=1, det_radius=5,
+        ...                        flying_focal_spot=[(0.1, 0), (0, 0.1)])
+        >>> geom.flying_focal_spot
+        array([[ 0.1, 0. ],
+               [ 0. , 0.1],
+               [ 0.1, 0. ],
+               [ 0. , 0.1]])
```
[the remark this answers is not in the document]
adler-j: Agree with holger on this

odl/tomo/geometry/conebeam.py
```diff
@@ -359,13 +386,30 @@ def angles(self):
         """Discrete angles given in this geometry."""
         return self.motion_grid.coord_vectors[0]
 
+    @property
+    def src_shift_func(self):
+        """Source shifts in the geometry."""
+        if self.__src_shift_func is None:
```
adler-j, line 392: This should happen in the constructor

odl/tomo/geometry/conebeam.py
```diff
@@ -408,16 +449,50 @@ def src_position(self, angle):
...
         center_to_src_init = -self.src_radius * self.src_to_det_init
+        # shifting the source according to ffs
+        tangent = np.array([self.src_to_det_init[1], -self.src_to_det_init[0]])
+        ffs_shift = (np.multiply.outer(flying_focal_spot[:, 0],
+                                       -self.src_to_det_init)
+                     + np.multiply.outer(flying_focal_spot[:, 1], tangent))
+        center_to_src_init = center_to_src_init + ffs_shift
+        # broadcasting to perform matrix multiplication "manually",
+        # since existing numpy functions do cross product along the outer
+        # dimensions, which we don't need here
```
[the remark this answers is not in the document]
adler-j: I think einsum is usually the most beautiful way to do it.

odl/tomo/geometry/conebeam.py
```diff
@@ -639,6 +716,11 @@ def __init__(self, apart, dpart, src_radius, det_radius,
             beam geometry.
         axis : `array-like`, shape ``(3,)``, optional
             Vector defining the fixed rotation axis of this geometry.
+        src_shift_func : a function of (angle), optional
+            Should return a 3 dimensional vector representing a shift
+            relative to the default source position. Vector elements
+            represent shifts along the following directions:
+            det_to_src, tangent to the rotation, rotation axis.
```
adler-j, line 723: For helical geometries the tangent to the rotation and the rotation axis are not orthogonal. Is this intended?

odl/tomo/geometry/conebeam.py
```diff
@@ -1024,6 +1125,15 @@ def angles(self):
         """Discrete angles given in this geometry."""
         return self.motion_grid.coord_vectors[0]
 
+    @property
+    def src_shift_func(self):
+        """Source shifts in the geometry."""
+        if self.__src_shift_func is None:
```
adler-j, line 1131: Move to constructor

odl/tomo/geometry/conebeam.py
adler-j: Use a einsum. I think this should be

```
np.einsum('bji, bi-> bj', rot_matrix, center_to_src_init)
```

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,46 @@
...
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Source and detector shifts for divergent beam geometries."""
+
+from __future__ import print_function, division, absolute_import
+import numpy as np
+from odl.discr import uniform_discr_frompartition
+from odl.discr.discr_utils import nearest_interpolator
+
+__all__ = ('flying_focal_spot')
```
adler-j, line 16: Missing `,` at the end, otherwise this is a big tuple of chars

odl/tomo/geometry/conebeam.py
```diff
@@ -21,6 +21,7 @@
 from odl.tomo.util.utility import (
     euler_matrix, is_inside_bounds, transform_system)
 from odl.util import array_str, indent, signature_string
+from odl.tomo.util.source_detector_shifts import flying_focal_spot
```
adler-j, line 24: What does this import do?

odl/tomo/geometry/conebeam.py
```diff
@@ -159,6 +167,22 @@ def __init__(self, apart, dpart, src_radius, det_radius,
...
         >>> np.allclose(geom.det_axis_init, e_y)
         True
+
+        Specifying a flying focal spot:
+
+        >>> apart = odl.uniform_partition(0, 2 * np.pi, 4)
+        >>> geom = FanBeamGeometry(
+        ...     apart, dpart,
+        ...     src_radius=1, det_radius=5,
+        ...     src_shift_func=lambda angle: flying_focal_spot(
```
adler-j, line 177: Use full paths in doctests

odl/tomo/geometry/conebeam.py
```diff
@@ -408,16 +452,48 @@ def src_position(self, angle):
...
         >>> np.allclose(points[1], [2, 0])
         True
+
+        Specifying flying focal spot:
+
+        >>> apart = odl.uniform_partition(0, 2 * np.pi, 4)
+        >>> geom = FanBeamGeometry(
+        ...     apart, dpart,
+        ...     src_radius=1, det_radius=5,
+        ...     src_shift_func=lambda angle: flying_focal_spot(
```
adler-j, line 462: Use full paths in doctests

odl/tomo/geometry/conebeam.py
```diff
@@ -766,6 +848,22 @@ def __init__(self, apart, dpart, src_radius, det_radius,
...
         >>> np.allclose(geom.det_axes_init, (e_y, e_z))
         True
+
+        Specifying a flying focal spot:
+
+        >>> apart = odl.uniform_partition(0, 2 * np.pi, 4)
+        >>> geom = ConeBeamGeometry(
+        ...     apart, dpart,
+        ...     src_radius=1, det_radius=5,
+        ...     src_shift_func=lambda angle: flying_focal_spot(
```
adler-j, line 858: Use full paths in doctests

odl/tomo/geometry/conebeam.py
```diff
@@ -1195,24 +1310,57 @@ def src_position(self, angle):
...
         >>> geom.src_position(np.zeros((4, 5))).shape
         (4, 5, 3)
+
+        Specifying flying focal spot:
+
+        >>> apart = odl.uniform_partition(0, 2 * np.pi, 4)
+        >>> geom = ConeBeamGeometry(
+        ...     apart, dpart,
+        ...     src_radius=1, det_radius=5,
+        ...     src_shift_func=lambda angle: flying_focal_spot(
```
adler-j, line 1320: Use full paths in doctests

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,46 @@
...
+    apart : 1-dim. `RectPartition`
+        Partition of the angle interval.
+    shifts : sequence of `array-like`
+        Each vectors in a sequence represent a subsequent shift
+        relative to the default source position. Vector elements
+        represent shifts along the following directions:
+        det_to_src, tangent to the rotation, rotation axis.
+    """
+    angle = np.array(angle, dtype=float, copy=False, ndmin=1)
+    interpolator = nearest_interpolator(np.arange(apart.size),
```
adler-j, line 36: Using a NN interpolator should be documented.

adler-j, line 36: We do support multiple dimensions in angle partitions, which is broken here. Document and assert correctness.

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,46 @@
...
+
+from __future__ import print_function, division, absolute_import
+import numpy as np
+from odl.discr import uniform_discr_frompartition
+from odl.discr.discr_utils import nearest_interpolator
+
+__all__ = ('flying_focal_spot')
+
+
+def flying_focal_spot(angle, apart, shifts):
```
adler-j, line 19: Could you add a test for this in particular?

odl/tomo/util/source_detector_shifts.py
```diff
@@ -0,0 +1,46 @@
...
+        represent shifts along the following directions:
+        det_to_src, tangent to the rotation, rotation axis.
+    """
+    angle = np.array(angle, dtype=float, copy=False, ndmin=1)
+    interpolator = nearest_interpolator(np.arange(apart.size),
+                                        apart.coord_vectors)
+    ind = interpolator(angle)
+
+    shifts = np.array(shifts, dtype=float, ndmin=2)
+    if shifts.shape[1] not in [2, 3]:
```
adler-j, line 41: Input validation should come before other work (e.g. creating the interpolator).

odl/test/tomo/geometry/geometry_test.py
JevgenijaAksjonova: I have a separate jupyter notebook, where I have verified visually that ray transform works as expected. Should I include this without numerical verification of the result? 

As a next step I was thinking to planning detector shifts (by analogy) and then it would be possible to do some numerical verification by defining equivalent geometries with different source and detector shifts.

odl/tomo/geometry/conebeam.py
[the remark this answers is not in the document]
JevgenijaAksjonova: geom.src_shift_func() is a property which corresponds to an input argument (a function)

geom.src_shift_func(geom.angles) is an application of this function to angles of this geometry.

So, I don't understand a problem

odl/tomo/util/source_detector_shifts.py
JevgenijaAksjonova: What do you mean? Tests in geometry_tests use this function, since it is the only implemented option

odl/tomo/geometry/conebeam.py
JevgenijaAksjonova: These 3 vectors must be orthogonal vector. The second vector would be tangent to the rotation if pitch was 0. Do you have any suggestions how to call it?

JevgenijaAksjonova:
Hi! Thank you for the feedback!
I have addressed all issues even though some of them are not marked outdated

odl/tomo/geometry/conebeam.py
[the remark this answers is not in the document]
adler-j: It corresponds, but it's not the same. I'd expect them to be the exact same, up to e.g. calling `np.array` on it. In general I'd expect this code to always work:

```
new_obj = Class(feature=old_obj.feature)
```

odl/test/tomo/geometry/geometry_test.py
adler-j: I'd really appreciate a numerical test (even a rough one) for code this complex.

odl/tomo/geometry/conebeam.py
adler-j: I'd just be a bit verbose here and explain what you mean.

odl/tomo/util/source_detector_shifts.py
adler-j: I agree, but they test the whole thing. It might be good to have a test for this in isolation to help track down errors.

adler-j: The other option, if this is not intended to be used externally, is to make this a private function in the cone beam geometry file.

adler-j:
Some of the seemingly outdated comments are still unresolved, could you check them and mark as resolved?

odl/tomo/geometry/conebeam.py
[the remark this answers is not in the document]
JevgenijaAksjonova: Yes, this should work.

The thing is that feature itself is a function. So, geom.src_shift_func returns a function, which is exactly the same as an input provided (input is also a function). The only thing I do in the code, is assigning default value if None is provided. 

I think Holgers comment there referred to something else and is outdated.

JevgenijaAksjonova:
Hi! Please check the latest updates.

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,138 @@ def test_shifted_volume(geometry_type):
...
+def test_source_detector_shifts_2d():
+    """Check that source/detector shifts are handled correctly.
+
+    We forward project a Shepp-Logan phantom and check that reconstruction
+    with flying focal spot is equal to a sum of reconstructions with two
+    geometries which mimic ffs by using initial offsets
+    (the detector must be large enough, not to be influenced by shifts)
+    """
+
+    # If no implementation is available, skip
```
adler-j, line 488: Redundant comment

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,138 @@ def test_shifted_volume(geometry_type):
...
+    if not odl.tomo.ASTRA_AVAILABLE:
+        pytest.skip(msg='ASTRA not available, skipping 2d test')
```
adler-j, line 490: The fact that the test is skipped is documented automatically, I'd prefer information on why it was skipped. E.g. astra is needed to run it.

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,138 @@ def test_shifted_volume(geometry_type):
     assert np.max(proj[3, 15:]) > 5
 
 
+def test_source_detector_shifts_2d():
```
adler-j, line 479: Very nice test!

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,138 @@ def test_shifted_volume(geometry_type):
...
+
+    # If no implementation is available, skip
+    if not odl.tomo.ASTRA_CUDA_AVAILABLE:
+        pytest.skip(msg='ASTRA_CUDA not available, skipping 3d test')
+
+    space = odl.uniform_discr([-1] * 3, [1] * 3, [10] * 3)
+    phantom = odl.phantom.shepp_logan(space)
+
+    full_angle = np.pi
+    n_angles = 2 * 7
```
adler-j: Why not just 14?

JevgenijaAksjonova: it must be even

JevgenijaAksjonova:
Hi, Jonas! The commit you checked was not the last one. I fixed some issues in ray_transform tests for 2d, now it works fine. 
However, in the 3d test there is a problem, the back-projected images are visually indistinguishable, but numerically, there is a gap between pixel values - 0.39.
I have checked the projection data and it looks that it differs only by a shift (as it is supposed to). I have also added an assertion on the total sum of the projected data for each angle. It it is equal.
However, the total sum of back-projected images is different, for the ffs geometry it is smaller. In fact the following rescaling 

im_combined =im_combined/ np.sum(im_combined) * np.sum(im)

reduces the gap to 0.00079. 

Do you have any ideas, what is wrong? Could this be because the detector is not perpendicular to src_det_axis?

adler-j:
How large is the relative error? 0.39 sounds like a lot (if the values should be ~1) so it might be worth investigating. Could you perhaps try to trigger it without the FFS somehow?

JevgenijaAksjonova:
Hi! 

I have added detector shifts and changed the ray_transform tests so that source and detector positions coincide. Even after this a 2% error remained in cone beam, while using astra 1.8.3. Simply rescaling one of the images to match the scale of the other solves the issue. (This is not needed for the latest ASTRA, there it works fine).

The detector shifts themself dont work as expected in 3D, this is related to issue #1567.

JevgenijaAksjonova:
Now I have added tests for the radon transform with detector shifts. It is ready for review

ozanoktem:
@kohr-h, @adler-j and/or @aringh , care to give this a final review? Perhaps we can push this to acceptance.

ThomasBudd:
Hi! Can I help with this somewhere? Also I have a question regarding the tam-danielsen window for ffs. To me it is not obvious how to mask out overlapping rays when the focal spot jumps between two different positions. Did you take care of this @JevgenijaAksjonova ? What way can this be solved? Maybe this was the source for the small error in 3d!

adler-j (approved):
Some minor comments that should be doable without further review.

Sorry for the late review again.

kohr-h (approved):

kohr-h:
I was absent from GH for a while, sorry for the delay. Go ahead after the minor things @adler-j mentioned. There's nothing left to fix that can't be fixed afterwards :-)

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,291 @@ def test_shifted_volume(geometry_type):
...
+    """Check that source shifts are handled correctly.
+
+    We forward project a Shepp-Logan phantom and check that reconstruction
+    with flying focal spot is equal to a sum of reconstructions with two
+    geometries which mimic ffs by using initial angular offsets and
+    detector shifts
+    """
+
+    if not odl.tomo.ASTRA_AVAILABLE:
+        pytest.skip(msg='ASTRA not available, skipping 2d test')
```
kohr-h, line 543:
```suggestion
        pytest.skip(msg='ASTRA required but not available')
```

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,291 @@ def test_shifted_volume(geometry_type):
...
+    dpart = odl.uniform_partition([-4] * 2, [4] * 2, [8 * d] * 2)
+    geom = odl.tomo.ConeBeamGeometry(apart, dpart, src_rad, det_rad)
+    k = 3
+    l = 2
+    shift = np.array([0, k, l]) * dpart.cell_sides[0]
+    geom_shift = odl.tomo.ConeBeamGeometry(apart, dpart, src_rad, det_rad,
+                                           det_shift_func=lambda angle: shift)
+
+    assert all_almost_equal(geom.angles, geom_shift.angles)
+    angles = geom.angles
```
kohr-h, line 642:
```suggestion
    angles = geom.angles

    assert all_almost_equal(angles, geom_shift.angles)
```

odl/test/tomo/operators/ray_trafo_test.py
```diff
@@ -475,5 +476,291 @@ def test_shifted_volume(geometry_type):
...
+
+def test_source_shifts_3d():
+    """Check that source shifts are handled correctly.
+
+    We forward project a Shepp-Logan phantom and check that reconstruction
+    with flying focal spot is equal to a sum of reconstructions with two
+    geometries which mimic ffs by using initial angular offsets and
+    detector shifts
+    """
+
```
kohr-h, line 675:
```suggestion
```

kohr-h:
Okay, now I just did the nit edits myself. Merging after CI.

ThomasBudd:
Is the issue solved why @JevgenijaAksjonova got some errors in 3d? I think this has to do with the implementation of the tam-danielsen window. When we have projections at different source positions it is much more complicated to find and mask out the overlapping rays and this leads to errors in the backprojection.

Issue #1501, opened 2019-05-29
FBP for 3D Euler geometry?


robtovey:
Hi,

As the heading suggests, I am trying to perform a FBP on data that is computed on an Euler geometry (2d grid of angles, no translations) but there are a couple of points where the machinery breaks straight away. 

There is also no example that I can see [here](https://github.com/odlgroup/odl/tree/master/examples/tomo) which suggests that someone has probably known about this hole and not gotten around to it yet. If someone could give me a couple of pointers then I'd be happy to do the leg work on this, I can't imagine that it is a huge amount of work but if I'm mistaken then please correct me!

As far as I can tell, there are two conceptual changes. An Euler geometry can have a 2D grid of angles and does not have a rotation axis. Other than this I think the implementation is exactly the same just with a couple of minor tweaks which I'm hoping someone else will see clearer than me. Any advice on the 'proper' way to deal this case and which lines are causing the problems I've listed below would be great.

My understanding is that the important function is [fbp_filter_op](https://github.com/odlgroup/odl/blob/master/odl/tomo/analytic/filtered_back_projection.py#L313) and the offending lines are:
- 367: an Euler geometry can have a 2D angular grid which doesn't have a `length`, can this be substituted by `volume`?
- 392: an Euler geometry doesn't have a rotation axis, can we just jump to `axes=[1,2]`?
- 420: if we don't have a `rot_dir` then `abs_freq` should be something like `sqrt(x[1]**2+x[2]**2)`?
- 450: generates an error I don't understand. 'operands could not be broadcast together with shapes (3,) (4,).' I guess this is also something to do with some Euler geometries having 2D angular domains? I think ran_shp needs to be length 4 too?
- 474: multiplying operators together gives an error that 'cell sides of domain and range differ significantly'. I don't really understand this one either but it is an issue with the zero-padded Fourier transform, the image and domain are no longer equal.

I have attached my partial solution in case that is of use, the lines I've changed are marked with `# HERE` flags. [filtered_back_projection.py.txt](https://github.com/odlgroup/odl/files/3233056/filtered_back_projection.py.txt)
Thanks again for your help.

adler-j:
This is a very interesting issue and I really liked the research you did. I'll skip the questions (for now), not because they are bad but because I need more background to be able to answer properly.

[...]

PR #1511, opened 2019-07-15
issue-1501__FBP-for-3d-Euler

robtovey:
This is an implementation of the 3D filtered back-projection suggested in the issue #1501. 

Firstly, an apology that I'm not hugely familiar with things on github so apologies for the mistakes that I have made/will make.

The two files are an update of the FBP code and an example which is very similar to the other examples on FBP but also compares against a least squares solution computed by Landweber iteration to check it's performing as expected.

Mathematically, the FBP is as expected, and follows the argument on #1501. The only extension in the code is to compute the Fourier transform of the kernel [straight from wiki, number 502](https://en.wikipedia.org/wiki/Fourier_transform#Formulas_for_general_n-dimensional_functions) and to realise that 2D Euler angles/spherical coordinates are a non-uniform parametrisation of the sphere. 

There are a couple of things of note:
1) There is a *weird* scaling factor needed but seems to be consistent. It is on line 411:
`fudge = 0.03 / ray_trafo.domain.cell_volume`
I see no reason for why the discretisation of the volume should have any influence on a Fourier scaling in data-space but maybe this is something weird coming from astra? 0.03 is also just a meaningless number.
2) I can't think of any quantitative tests to perform to verify accuracy. The accuracy of FBP is a continuum result and in 3D I think the cost of getting near this limit is probably too high. The example code includes a comparison with a Landweber iteration which should compute the discrete analogue. The plots show they are reasonably close and the error gets lower with more projections but there is a difference, part of this may be down to point 4 below.
3) I have updated documentation in both files. The only qualitative difference compared with other methods is that the FBP equation is only valid when you specify 2 Euler angles. The case with 3 is more complicated as you can have over-complete parametrisations of the sphere so I don't know what to do there but they are the same geometry. I have explained this as `Parallel3dEulerGeometry : Exact reconstruction, only implemented for 2d angle partitions.`
4) The scaling I used for the integral over the sphere is actually very natural and I wonder if it should/could be built into the operator to make the behaviour closer to the continuum model. In particular, sampling uniformly in spherical coordinates means you sample very densly at the poles. Thus, at the moment, doing a least-squares fit will heavily bias to fitting the data at the poles. I wonder if this it is the natural scaling to treat each projection equally (as is currently done) or keep the metric uniform on the sphere and have a non-uniform weighting? This is actually a difference between the FBP and Landweber iteration at the moment.

Anyway, I hope this is of interest and let me know what more I can do.


adler-j:
Thank you for this pull request! We're currently quite busy so review might take a while sadly. Would you be able to fix the PEP8 issues indicated by @pep8speaks?

robtovey:
No problem, I know the feeling!
Don't worry about rushing for my sake, I'll try and get pep happy by the end of the week then just get back to me whenever.
Thanks for taking the time to update me!

robtovey:
I'm closing this as stale :)
Vscode keeps reminding me that it still needs reviewers...

leftaroundabout:
@robtovey apologies on behalf of the old ODL team that this never got merged! As you will have noticed, ODL has been more or less in stasis for a long time. We are in the process "resurrecting" it though, and I think this PR is actually still relevant.

robtovey:
No problem, feel free to re-open at a later date if it becomes relevant again

I'm basically closing for two reasons:
- getting rid of that annoying vscode notification :)
- it's been so long now that I've forgotten what I implemented and why, I'm probably not in a place to help if you do decide to take it forward

Good luck with the resurrection!

Issue #1435, opened 2018-09-12
Operator Space->Field can't be part of a ProductSpaceOperator


kohr-h:
### Context
I stumbled upon this issue while trying to use a scalar-valued side constraint in an optimization problem.

### Problem
We currently have a few places with hard-coded checks `isinstance(space, LinearSpace)`, and one of them is the range of operators provided to `ProductSpaceOperator`. This excludes functionals (i.e., `Functional` instances and `Operator` instances whose `range` is a `Field`) in the first place.

### Workarounds
If that restriction is relaxed to `isinstance(space, (LinearSpace, Field))` -- as already done in a few places -- we get to the implicit `LinearSpace` assumptions. The most obvious one is in `ProductSpaceOperator._call`: it uses `op(input, out=out[i])` in the loop, which isn't allowed for `Functional` because its range elements are not assignable in place.

If we work around that by first checking `op.is_functional` and doing `out[i] = op(input)` for `True`, we get into the next issue: `ProductSpaceElement.__setitem__` hard-codes in-place semantics in the loop over the parts.
At this point, it becomes really tedious to do all those checks over and over, and to change the logic of complicated functions just to make them work with `Field` instead of `LinearSpace`.

### The actual problem
Looking at it from a different angle, the fundamental problem that we try to avoid by disallowing `op(input, out=field_elem)` is that `Field` elements are primitive Python types that cannot be assigned to in place.

### Solution?
Thus, a logical way out of this problem is to create a wrapper class, e.g., `RealNumber`, for those types, that does allow assignment and usage as `out` in operators. Of course, to make that class behave nicely, it needs to implement basically all the methods that `LinearSpaceElement` provides, and `Field` also needs to be extended correspondingly. It's not totally straightforward, but certainly doable.

So the question is: how do we want to resolve the original issue?

I feel that the current limitation is severe enough that we address it in *some* way, but I'm not entirely sure which way is the best. I'd be glad to hear opinions on this.

Ping @adler-j @aringh

aringh:
I think that this is more or less the same problem as #563, to which @adler-j create an example of a work-around for in PR #565 (which however was closed and not merged).

My thoughts on the issue: in general it would be nice if `Field` worked more like `LinearSpace`, since then it would be easier to really interpret `Functional` as a special instance of `Operator`. However, when it comes to how to implement this in the best way, I am not sure. What you describe sounds good to me, and it sounds like #565 may be reused for this purpose.

kohr-h:
[...]

PR #1512, opened 2019-07-18
ENH: remove variable from torch doctests, make functionals work

kohr-h:
- Adapt doctests to modern PyTorch, i.e., remove `Variable` and update printing. Now they pass again.
- Make functionals work as `Module`, handy for using custom loss functions.

### DIFF
--- a/odl/contrib/torch/operator.py
+++ b/odl/contrib/torch/operator.py
@@ -68,13 +68,13 @@ class OperatorAsAutogradFunction(torch.autograd.Function):
 
         Parameters
         ----------
-        input : `torch.tensor._TensorBase`
+        input : `torch.Tensor`
             Point at which to evaluate the operator.
 
         Returns
         -------
-        result : `torch.autograd.variable.Variable`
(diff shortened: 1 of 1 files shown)


odl/contrib/torch/operator.py
```diff
@@ -329,51 +310,39 @@ def forward(self, x):
...
-        [torch.FloatTensor of size 3x2x2]
+                [[ 4.,  8.],
+                 [ 5., 10.]]])
         """
         in_shape = x.data.shape
-        op_in_shape = self.op_func.operator.domain.shape
-        op_out_shape = self.op_func.operator.range.shape
+        op_in_shape = self.operator.domain.shape
+        if self.operator.is_functional:
+            op_out_shape = ()
```
adler-j: We could give `Field` a shape `()` to avoid these special cases, if we wanted to.

kohr-h: Yes, indeed. It's also related to #1435

PR #1516, opened 2019-07-19
ENH: torch wrapper overhaul with better API

kohr-h:
This is an improved implementation for the interface code with PyTorch. It follows the official way of extending PyTorch as detailed [here](https://pytorch.org/docs/stable/notes/extending.html#extending-pytorch) and [here](https://pytorch.org/docs/stable/autograd.html#function). (I hope that it also solves the memory leak issues I encountered.)

Main changes:
- `OperatorAsAutogradFunction` -> `OperatorFunction` and `OperatorAsModule` -> `OperatorModule`
- The looping over extra dimensions is now implemented in the `OperatorFunction`, i.e., `OperatorModule` has basically become a thin wrapper around it.
- `OperatorFunction`no longer *wraps* an `operator` but *applies* it to given input and takes care of the graph building behind the scenes. It also never needs to be instantiated, instead:

   ```python
   op = <some ODL operator>
   x = torch.tensor(...)
   y = OperatorFunction.apply(op, x)
   ```

   This exactly as in the ["Extending `torch.autograd`"](https://pytorch.org/docs/stable/notes/extending.html#extending-torch-autograd) documentation.

- Instead of storing stuff in the instance, `OperatorModule` now makes use of the context object `ctx` to pass information from forward to backward pass.

**TODO:**
- [x] Implement improved API
- [x] Adapt doctests
- [x] Adapt examples
- [x] Adapt tests

Closes: #1523


kohr-h:
Unit tests adapted and working, ready for review.

kohr-h:
Unless there are objections, I'll merge this rather soon.

kohr-h:
Just checked the failing code in #1523, works for me now. Will merge.

PR #1520, opened 2019-08-13
API: move 'interp' out of spaces

kohr-h:
In the hope of getting it merged faster, I created this PR that takes the minimal action to remove `'interp'` from the spaces (where I no longer think it makes sense) to corresponding operators.

Tests will follow.

Please review this version.


kohr-h:
Some additional context: Apparently the PR #1459 is too big to review, so I'm trying to break it down into smaller ones. This one is only concerned with moving the `interp` property from spaces to operators and functions that make use of it (really only a few).
This implies that all calls to `<space or element>.sampling(...)` and `<space or element>.interpolation(...)` have to be replaced by according function calls. These functions are pulled out of the `FunctionSpaceMapping` type operators, which were removed.

I haven't adapted the tests yet since that would mean moving a bunch of code around, blowing up the size of the changes substantially. That'll come as a second step.

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    >>> f = np.array([[1, 2, 3, 4],
+    ...               [5, 6, 7, 8]],
+    ...              dtype=float)
+    >>> interpolator = nearest_interpolator(f, part.coord_vectors)
+    >>> interpolator([1, 1])  # single point
+    5.0
+    >>> interpolator([[0.5, 2.0],
+    ...               [0.0, 4.5],
+    ...               [0.0, 3.0]])  # 3 points at once
+    array([ 4.,  5.,  3.])
```
aringh: I am not sure I understand how this interpolator works in higher dimensions... why is this not ``array([ 2.,  4.,  3.])``?

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+        )
+
+        return interpolator(arg, out=out)
+
+    return nearest_interp
+
+
+def linear_interpolator(f, coord_vecs):
+    """Return the linear interpolator for discrete values.
+    Parameters
```
aringh: Blank lines, also below

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    an array of values is returned:
+    >>> part = odl.uniform_partition(0, 2, 5)
+    >>> part.coord_vectors  # grid points
+    (array([ 0.2,  0.6,  1. ,  1.4,  1.8]),)
+    >>> f = [1.0, 2.0, 3.0, 4.0, 5.0]
+    >>> interpolator = linear_interpolator(f, part.coord_vectors)
+    >>> interpolator(0.3)  # 0.75 * 1 + 0.25 * 2 = 1.25
+    1.25
+    >>> interpolator([0.6, 1.3, 1.9])
+    array([ 2.  ,  3.75,  3.75])
```
aringh: It is hard to, from the above doc, understand why `1.9` gets interpolated to `3.75`. It must have to do with an implicit assumption of the behavior of `f` at the boundary. What is the assumption? Can the user modify this by passing extra arguments?

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+        return interpolator(arg, out=out)
+
+    return linear_interp
+
+
+def per_axis_interpolator(f, coord_vecs, schemes, nn_variants=None):
+    """Return a per axis defined interpolator for discrete values.
+    With this function, the interpolation scheme can be chosen for each axis
+    separately.
+    Parameters
```
aringh: Blank lines

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+
+    schemes_in = schemes
+    if is_string(schemes):
+        scheme = str(schemes).lower()
+        if scheme not in _SUPPORTED_INTERP_SCHEMES:
+            raise ValueError('`schemes` {!r} not understood'
+                             ''.format(schemes_in))
+        schemes = [scheme] * f.ndim
+    else:
+        schemes = [str(scm).lower() if scm is not None else None
```
aringh: ~~Check that they are all in `_SUPPORTED_INTERP_SCHEMES`?~~ Seems to be done in line 431. Make that clear here?

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+            # Make list with `nn_variants` where `schemes == 'nearest'`,
+            # else `None` (variants only applies to axes with nn
+            # interpolation)
+            nn_variants = [nn_variants if scm == 'nearest' else None
+                           for scm in schemes]
+            if str(nn_variants_in).lower() not in ('left', 'right'):
+                raise ValueError('`nn_variants` {!r} not understood'
+                                 ''.format(nn_variants_in))
+        else:
+            nn_variants = [str(var).lower() if var is not None else None
```
aringh: ~~No need to check that they are all either `left` or `right`?~~ Seems to be done in line 431. Make it clear here?

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+                           for var in nn_variants]
+
+    for i in range(f.ndim):
+        # Reaching a raise condition here only happens for invalid
+        # sequences of inputs, single-input case has been checked above
+        if schemes[i] not in _SUPPORTED_INTERP_SCHEMES:
+            raise ValueError('`interp[{}]={!r}` not understood'
```
aringh: Give allowed values in error message? Also below

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+                             ''.format(schemes_in[i], i))
+        if (schemes[i] == 'nearest' and
+                nn_variants[i] not in ('left', 'right')):
+            raise ValueError('`nn_variants[{}]={!r}` not understood'
+                             ''.format(nn_variants_in[i], i))
+        elif schemes[i] != 'nearest' and nn_variants[i] is not None:
```
aringh: Doc says it `has no effect` but here it raises an error. Clarify this in the doc?

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+            raise ValueError("variant '{}' not understood".format(variant_))
+        self.variant = variant_
+
+    def _evaluate(self, indices, norm_distances, out=None):
+        """Evaluate nearest interpolation."""
+        idx_res = []
+        for i, yi in zip(indices, norm_distances):
+            if self.variant == 'left':
+                idx_res.append(np.where(yi <= .5, i, i + 1))
+            else:
```
aringh: Specify that this is the `right` case, and that all others are rule out by above checks?

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    # For upper/lower out-of-bounds nodes, we need to set the
+    # lower/upper neighbors to the last/first grid point
+    edge = [idcs, idcs + 1]
+    edge[0][hi] = -1
+    edge[1][lo] = 0
+
+    return w_lo, w_hi, edge
+
+
+def _compute_linear_weights_edge(idcs, ndist):
```
aringh: Is this where the implicit assumption is made on what happens to the value at interpolation points that are, e.g., larger than the current largest grid point?

odl/tomo/backends/skimage_radon.py
```diff
@@ -31,23 +33,24 @@ def skimage_proj_space(geometry, volume_space, proj_space):
...
     Clip all points to fit within the bounds of the given space.
     """
     min_x = skimage_proj_space.domain.min()[1]
     max_x = skimage_proj_space.domain.max()[1]
 
-    def interpolator(x):
+    def interpolator(x, out=None):
         x = (x[0], np.clip(x[1], min_x, max_x))
-        return sinogram.interpolation(x, bounds_check=False)
+        interpolator = linear_interpolator(
```
aringh: I don't know why it was there in the first place, but why remove the `bounds_check=False`?

odl/tomo/backends/skimage_radon.py
```diff
@@ -90,9 +93,12 @@ def skimage_radon_forward_projector(volume, geometry, proj_space, out=None):
     if out is None:
```
aringh: Here you still use the `if out is None:`-construction. What is the difference with `none_context` used above?

odl/tomo/backends/skimage_radon.py
```diff
@@ -127,9 +133,12 @@ def skimage_radon_back_projector(sinogram, geometry, vol_space, out=None):
     skimage_range = skimage_proj_space(geometry, vol_space, sinogram.space)
 
     skimage_sinogram = skimage_range.element()
-    skimage_sinogram.sampling(
-        clamped_interpolation(skimage_range, sinogram), bounds_check=False
-    )
+    with writable_array(skimage_sinogram) as sino_arr:
```
aringh: As above: why remove `bounds_check = False`?

odl/deform/linearized.py
```diff
@@ -56,24 +67,24 @@ def linear_deform(template, displacement, out=None):
...
-    interpolation and an offset equal to half the distance between two
+    interpolation and an offset of half the distance between two
     points, 0.1, one gets the mean of the values.
 
-    >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
-    >>> disp_field_space = space.tangent_bundle
-    >>> template = space.element([0, 0, 1, 0, 0])
     >>> displacement_field = disp_field_space.element([[0, 0, 0, -0.1, 0]])
-    >>> linear_deform(template, displacement_field)
+    >>> linear_deform(template, displacement_field, interp='linear')
```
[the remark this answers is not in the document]
kohr-h: That's right. It's been a while, but I think I didn't think it was necessary to re-define it here to exactly the same value.

odl/deform/linearized.py
```diff
@@ -297,48 +347,71 @@ def __init__(self, displacement, templ_space=None):
...
         >>> print(op([0, 0, 1, 0, 0]))
         [ 0.,  0.,  1.,  1.,  0.]
 
         The result depends on the chosen interpolation. With 'linear'
-        interpolation and an offset equal to half the distance between two
+        interpolation and an offset of half the distance between two
         points, 0.1, one gets the mean of the values.
 
-        >>> space = odl.uniform_discr(0, 1, 5, interp='linear')
+        >>> space = odl.uniform_discr(0, 1, 5)
```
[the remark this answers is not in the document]
kohr-h: Good point. The re-definition became unnecessary.

odl/discr/discr_ops.py
```diff
@@ -80,16 +81,36 @@ def __init__(self, domain, range):
...
         """
-        if out is None:
-            return x.interpolation
-        else:
-            out.sampling(x.interpolation)
+        interpolator = per_axis_interpolator(
+            x, self.domain.grid.coord_vectors, self.interp
+        )
+
+        context = none_context if out is None else writable_array
```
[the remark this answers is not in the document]
kohr-h: The idea is to make the rest of the code unconditional, because it can deal with `out_arr is None` itself. Compare this version to the alternative:
```python
if out is None:
    return point_collocation(
        interpolator, self.range.meshgrid, out=out_arr
    )
else:
    with writable_array(out) as out_arr:
        return point_collocation(
            interpolator, self.range.meshgrid
        )    
```

As a side note, for Python 3.7 and higher, we can move to [`contextlib.nullcontext`](https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext) if we want, but I don't feel like introducing a backports module just for that :-).

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    """Turn interpolation type into a tuple with one entry per axis."""
+    interp_in = interp
+    if is_string(interp):
+        interp = str(interp).lower()
+        if interp not in _SUPPORTED_INTERP_SCHEMES:
+            raise ValueError('`interp` {!r} not understood'.format(interp_in))
+        interp_byaxis = (interp,) * ndim
+    else:
+        interp_byaxis = tuple(str(itp).lower() for itp in interp)
+        if len(interp_byaxis) != ndim:
```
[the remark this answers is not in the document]
kohr-h: Right, I forgot that one. Good spot.

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    return all(itp == interp_byaxis[0] for itp in interp_byaxis)
+
+
+def _normalize_interp(interp, ndim):
+    """Turn interpolation type into a tuple with one entry per axis."""
+    interp_in = interp
+    if is_string(interp):
+        interp = str(interp).lower()
+        if interp not in _SUPPORTED_INTERP_SCHEMES:
+            raise ValueError('`interp` {!r} not understood'.format(interp_in))
```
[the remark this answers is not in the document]
kohr-h: I guess so :-)

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    Examples
+    --------
+    Sample a 1D function:
+    >>> from odl.discr.grid import sparse_meshgrid
+    >>> domain = odl.IntervalProd(0, 5)
+    >>> func = make_func_for_sampling(lambda x: x ** 2, domain)
+    >>> mesh = sparse_meshgrid([1, 2, 3])
+    >>> point_collocation(func, mesh)
+    array([ 1.,  4.,  9.])
+    By default, inputs are checked against ``domain`` to be in bounds. This
```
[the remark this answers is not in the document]
kohr-h: Oh yes, that must have been a copy-paste error. (Sometimes I copy stuff from GH directly, which tends to swallow blank lines). I don't even think the doctest will work like that at all.

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    >>> point_collocation(func1, mesh)
+    array([[[ 0.,  0.],
+            [ 1.,  1.]],
+    <BLANKLINE>
+           [[ 0.,  0.],
+            [ 0.,  0.]],
+    <BLANKLINE>
+           [[ 4.,  5.],
+            [ 5.,  6.]]])
+    >>> array_of_funcs = [  # equivalent to `vec_valued`
```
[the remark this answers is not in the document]
kohr-h: Lots of them missing here, will fix.

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,812 @@
...
+    Parameters
+    ----------
+    f : numpy.ndarray
+        Function values that should be interpolated.
+    coord_vecs : sequence of numpy.ndarray
+        Coordinate vectors of the rectangular grid on which interpolation
+        should be based. They must be sorted in ascending order. Usually
+        they are obtained as ``grid.coord_vectors`` from a `RectGrid`.
+    variant : {'left', 'right'}, optional
+        Which neighbor to prefer at the midpoint between two nodes.
```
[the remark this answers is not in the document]
kohr-h: Yes. (Although for a choice, the convention is to put the default first. But it's not very obvious.)

odl/tomo/backends/skimage_radon.py
kohr-h: I think I removed it for debugging purposes, but it should go back in. We construct `x` to be in bounds, so we can skip the (potentially expensive) check.

odl/tomo/backends/skimage_radon.py
kohr-h: Right, I could use it here, too.

odl/tomo/backends/skimage_radon.py
kohr-h: Same as above (*I think*, need to check).

kohr-h:
Thanks for the partial review, @aringh. I'll try to get the changes done asap.

odl/discr/discr_utils.py
kohr-h: Don't take the doctests too seriously at the moment. Need fix anyway. (Means: probably it should be `[2, 4, 3]`).

odl/tomo/backends/skimage_radon.py
kohr-h: Turns out that the created interpolator doesn't do bounds checking, so no need for `bounds_check=False`.

odl/tomo/backends/skimage_radon.py
kohr-h: There's a slight difference, in that the interpolator returns a NumPy array, not a ``proj_space`` element, so we can do either this, or call `proj_space.element` on the result. I'll stick with this version for now.

odl/discr/discr_utils.py
kohr-h: I've made the check a bit more readable now.

odl/discr/discr_utils.py
kohr-h: Done

odl/discr/discr_utils.py
kohr-h: Doc was right here, fixed.

odl/discr/discr_utils.py
kohr-h: Clarified

odl/discr/discr_utils.py
kohr-h: Yes, I hoped that the comments explain it in a somewhat understandable way. What are you after specifically?

kohr-h:
Doctests now work again and make more sense, sorry for the noise. I'll do the unit tests next.

odl/tomo/backends/skimage_radon.py
kohr-h: Yes

odl/test/space/tensors_test.py
```diff
@@ -93,7 +93,7 @@ def _weighting_cls(impl, kind):
 weight_ids = [' weight=1.0 ', ' weight=0.5 ', ' weight=<array> ']
 
 
-@pytest.fixture(scope='module', params=weight_params, ids=weight_ids)
+@pytest.fixture(params=weight_params, ids=weight_ids)
```
kohr-h: Fix for https://github.com/pytest-dev/pytest/issues/6497

odl/test/discr/discr_utils_test.py
```diff
@@ -0,0 +1,265 @@
+# Copyright 2014-2020 The ODL contributors
+#
+# This file is part of ODL.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Unit tests for `discr_utils`."""
```
kohr-h: Basically the same tests as `discr_mappings_test.py` before, with simplifications.

odl/test/util/normalize_test.py
```diff
@@ -78,7 +78,7 @@ def seq_conv(request):
                  for axis in axes_conv_params]
 
 
-@pytest.fixture(scope="module", ids=axes_conv_ids, params=axes_conv_params)
+@pytest.fixture(ids=axes_conv_ids, params=axes_conv_params)
```
kohr-h: Fix for https://github.com/pytest-dev/pytest/issues/6497 (same above)

kohr-h:
It's been a while. Tests are fixed now, with only minor adaptions.

Changes since last review:

- Nearest neighbor variants removed. Never used, and resulted in flaky tests.
- Fix added for a PyTest issue when one of the parameters is a NumPy array. PyTest caches test parameters and uses `==` to check if it has seen a value already. That doesn't work for arrays.
  Caching is only used for `scope='module'` or higher, so the default `scope` still works.

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,818 @@
...
+
+    >>> part = odl.uniform_partition(0, 2, 5)
+    >>> part.coord_vectors  # grid points
+    (array([ 0.2,  0.6,  1. ,  1.4,  1.8]),)
+    >>> f = [1.0, 2.0, 3.0, 4.0, 5.0]
+    >>> interpolator = linear_interpolator(f, part.coord_vectors)
+    >>> interpolator(0.3)  # 0.75 * 1 + 0.25 * 2 = 1.25
+    1.25
+    >>> interpolator([0.6, 1.3, 1.9])
+    array([ 2.  ,  3.75,  3.75])
```
aringh: I still do not understand how the last component becomes `3.75`. The linear interpolation is between `(1.8, f(1.8)) = (1.8, 5.0)` and `(x, f(x))` for what `x` and what value of `f(x)`?

Since the partition is between `0` and `2`, I assume `x = 2`. But that would mean `f(2) = 2.5`, which I do not understand why...

**Edit:** This makes sense if `x = 2.2`, since the step size between points are `0.4`, and indeed we have zero boundary values, i.e., `f(2.2) = 0`. Is this correct? Then I do understand the examples below (line 429 and further) as well, with mixed linear and nearest interpolation.

odl/discr/discr_utils.py
aringh: This was related to my question above about how the linear interpolation is done for points outside of the actual grid. However, I fail to see the difference between `weight zero` and `value zero` - don't they both achieve the same result about an implicit boundary value of `0`?

(At least in 1D, but maybe there is something in higher dimensions I do not fully grasp)

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,818 @@
...
+            Type of expected input values in ``__call__``
+        interp : sequence of str
+            Indicates which interpolation scheme to use for which axis
+        """
+        super(_PerAxisInterpolator, self).__init__(
+            coord_vecs, values, input_type)
+        self.interp = interp
+
+    def _evaluate(self, indices, norm_distances, out=None):
+        """Evaluate linear interpolation.
```
aringh: `linear` or `per axis`?

odl/deform/linearized.py
```diff
@@ -33,6 +37,13 @@ def linear_deform(template, displacement, out=None):
     displacement : element of power space of ``template.space``
         Vector field (displacement field) used to deform the
         template.
+    interp : str or sequence of str
+        Interpolation type that should be used to sample the template on
+        the deformed grid. A single value applies to all axes, and a
+        sequence gives the iterpolation scheme per axis.
+
+        Supported values: ``'nearest'``, ``'linear'``
```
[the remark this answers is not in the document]
kohr-h: Well, nearest neighbor is still supported, just the "left" and "right" variants are gone. So this is still correct.
But I need to make sure to update the passages that say something like "one of the neighbors is selected" to "the right neighbor is selected".

odl/deform/linearized.py
```diff
@@ -56,24 +67,24 @@ def linear_deform(template, displacement, out=None):
     >>> disp_field_space = space.tangent_bundle
     >>> template = space.element([0, 0, 1, 0, 0])
     >>> displacement_field = disp_field_space.element([[0, 0, 0, -0.2, 0]])
-    >>> linear_deform(template, displacement_field)
+    >>> linear_deform(template, displacement_field, interp='nearest')
```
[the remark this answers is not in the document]
kohr-h: See other comment. Yes, doctests pass, too.

odl/discr/discr_utils.py
```diff
@@ -0,0 +1,818 @@
...
+
+import numpy as np
+
+from odl.util import (
+    is_string, is_valid_input_array, is_valid_input_meshgrid,
+    out_shape_from_array, out_shape_from_meshgrid)
+
+__all__ = (
+    'point_collocation',
+    'nearest_interpolator',
```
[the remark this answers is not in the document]
kohr-h: See other comment

odl/discr/discr_utils.py
kohr-h: Yes, that's what happens. Sorry for not clarifying. Maybe I should write a comment into the example that explains the behavior.

odl/discr/discr_utils.py
kohr-h: You have the right hunch: in nD, the weights are the tensor product of the 1D weights.
So for instance, in 2D, if a weight is zero in one dimension, the total weights will have a zero row or column due to that. On the other hand, setting a single value to 0 has a very different effect.

odl/discr/lp_discr.py
```diff
@@ -89,39 +79,15 @@ def __init__(self, fspace, partition, tspace, interp='nearest', **kwargs):
             raise ValueError('`fspace.scalar_out_dtype` does not match '
                              '`tspace.dtype`: {} != {}'
                              ''.format(fspace.scalar_out_dtype, tspace.dtype))
+        if partition.shape != tspace.shape:
+            raise ValueError(
+                "`partition.shape` must be equal to `tspace.shape`, but "
+                "{} != {}".format(partition.shape, tspace.shape)
+            )
```
kohr-h: Probably this was implicitly  checked in sampling and interpolation operators earlier. This version is better anyhow.

odl/discr/lp_discr.py
```diff
@@ -1257,6 +1211,9 @@ def show(self, title=None, method='', coords=None, indices=None,
             is that ``fig`` is the return value of an earlier call to
             this function.
 
+        interp : {'linear', 'nearest'}, optional
+            Interpolation type that should be used for the plot.
```
kohr-h: This was simply missing in the docstring.

odl/deform/linearized.py
```diff
@@ -134,8 +145,23 @@ def __init__(self, template, domain=None):
             ``domain[0].partition == template.space.partition``, so
             this option is useful mainly when using different interpolations
             in displacement and template.
+
             Default: ``template.space.real_space.tangent_bundle``
 
+        interp : str or sequence of str
+            Interpolation type that should be used to sample the template on
+            the deformed grid. A single value applies to all axes, and a
+            sequence gives the iterpolation scheme per axis.
```
kohr-h: Typo "iterpolation"

odl/test/discr/discr_utils_test.py
```diff
@@ -0,0 +1,265 @@
...
+    true_mg = [1 + 2j, 0 + 1j, 3 + 4j, 4 + 5j]
+    assert all_equal(interpolator(mg), true_mg)
+    interpolator(mg, out=out)
+    assert all_equal(out, true_mg)
+
+
+def test_nearest_interpolation_2d():
+    """Test nearest neighbor interpolation in 2d."""
+    coord_vecs = [[0.125, 0.375, 0.625, 0.875], [0.25, 0.75]]
+    f = np.array([0, 1, 2, 3, 4, 5, 6, 7], dtype="float64").reshape((4, 2))
```
aringh: Could this be given in the same form as `coord_vecs`? That would make it easier to verify that the tests do what I expect them to do :smiley:

odl/test/discr/discr_utils_test.py
```diff
@@ -0,0 +1,265 @@
...
+    interpolator = nearest_interpolator(f, coord_vecs)
+
+    # Evaluate at single point
+    val = interpolator([0.3, 0.6])  # closest to index (1, 1) -> 3
```
aringh: ~~Is this comment correct? Isn't the closest point `[0.25, 0.625]`, correspoding to index `(0,3)` and value `3`? Or am I reading something backwards?~~

I see, I was indeed reading it wrongly. Related to my above comment :wink:

odl/test/discr/discr_utils_test.py
```diff
@@ -0,0 +1,265 @@
...
+
+def test_nearest_interpolation_2d_string():
+    """Test nearest neighbor interpolation in 2d with string values."""
+    coord_vecs = [[0.125, 0.375, 0.625, 0.875], [0.25, 0.75]]
+    f = np.array(list('mystring'), dtype='U1').reshape((4, 2))
+    interpolator = nearest_interpolator(f, coord_vecs)
+
+    # Evaluate at single point
+    val = interpolator([0.3, 0.6])  # closest to index (1, 1) -> 3
+    assert val == u't'
```
aringh: What is the u for?

odl/test/discr/discr_utils_test.py
```diff
@@ -0,0 +1,265 @@
...
+    val = interpolator([0.3, 0.6])
+    l1 = (0.3 - 0.125) / (0.375 - 0.125)
+    l2 = (0.6 - 0.25) / (0.75 - 0.25)
+    true_val = (
+        (1 - l1) * (1 - l2) * f[0, 0]
+        + (1 - l1) * l2 * f[0, 1]
+        + l1 * (1 - l2) * f[1, 0]
+        + l1 * l2 * f[1, 1]
+    )
+    assert val == pytest.approx(true_val)
```
aringh: Confession: I only checked this one carefully - the checks below in `test_linear_interpolation_2d` I have not done explicitly. But they have the same structure, and therefore seems to be correct.

odl/test/discr/discr_utils_test.py
```diff
@@ -0,0 +1,265 @@
...
+    interp = ['linear', 'nearest']
+    f = np.arange(1, 9, dtype='float64').reshape((4, 2))
+    interpolator = per_axis_interpolator(f, coord_vecs, interp)
+
+    # Evaluate at single point
+    val = interpolator([0.3, 0.5])
+    l1 = (0.3 - 0.125) / (0.375 - 0.125)
+    # 0.5 equally far from both neighbors -> NN chooses 0.75
+    true_val = (1 - l1) * f[0, 1] + l1 * f[1, 1]
+    assert val == pytest.approx(true_val)
```
aringh: Again, I only did this one carefully. But the rest have the same structure and therefore seems correct too.

odl/discr/discretization.py
```diff
@@ -124,61 +78,27 @@ def tspace_type(self):
...
+            Input data to create an element from.
         order : {None, 'C', 'F'}, optional
             Storage order of the returned element. For ``'C'`` and ``'F'``,
             contiguous memory in the respective ordering is enforced.
             The default ``None`` enforces no contiguousness.
         kwargs :
-            Additional arguments passed on to `sampling` when called
-            on ``inp``, in the form ``sampling(inp, **kwargs)``.
-            This can be used e.g. for functions with parameters.
+            Additional keyword arguments passed on to the sampler when called
```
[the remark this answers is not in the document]
kohr-h: Probably not :grin: 
Will reword.

odl/test/discr/discr_utils_test.py
kohr-h: Yes, of course. This is close to the old version, but a product of laziness largely :grin:

odl/test/discr/discr_utils_test.py
kohr-h: > Confession

:rofl: 

I can relate. But these tests are really the same as the old ones in `discr_mappings_test`, so I'm not too worried about correctness.

odl/test/deform/linearized_deform_test.py
```diff
@@ -211,14 +208,14 @@ def test_fixed_templ_deriv(space):
...
     error = (fixed_templ_deriv_exact - fixed_templ_deriv_comp).norm()
     rlt_err = error / fixed_templ_deriv_comp.norm()
-    assert rlt_err < error_bound(space.interp)
+    assert rlt_err < error_bound(interp)
 
 
 # --- LinDeformFixedDisp --- #
 
 
 def test_fixed_disp_init():
```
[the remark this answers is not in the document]
kohr-h: Hm, good question. I'll check.

odl/discr/lp_discr.py
kohr-h: Exactly. And `saveto` and `clim` are just pass-through args to `matplotlib`, so they're not really "ours".

odl/test/deform/linearized_deform_test.py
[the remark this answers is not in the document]
kohr-h: Well, these init tests just test instantiation and basic properties of classes. All other tests do the first part as well, so it doesn't really make a lot of sense to run the whole parameter set here. In some sense, init tests are mainly there to help find basic errors in the class instantiation, something that's harder if a test fails that actually tries to test something else.

So I think it's fine like this.

odl/test/discr/discr_utils_test.py
kohr-h: To make sure it's unicode. Maybe the `==` comparison would also work without, but this seems safer for Python 2.

kohr-h:
Thanks for the review, @aringh! Remaining stuff is fixed, will merge after CI.

PR #1528, opened 2019-09-30
Add scaling factors for ASTRA >= 1.9.9.dev

wjp:
This adapts the ODL astra_cuda backend scaling factors to the new adjoint scaling in astra-1.9.9.dev (current astra-toolbox master since a few days ago, astra-toolbox/astra-toolbox@54af7e8e22a3f1c9d90b13291b28d39778c05564), as announced in #1415 .

It is probably possible to clean things up further now, using a feature flag for example, but I'll leave that to you.

### DIFF
--- a/odl/tomo/backends/astra_cuda.py
+++ b/odl/tomo/backends/astra_cuda.py
@@ -115,7 +115,8 @@ class AstraCudaProjectorImpl(object):
                     self.proj_space.shape)
 
             # Fix scaling to weight by pixel size
-            if isinstance(self.geometry, Parallel2dGeometry):
+            if (isinstance(self.geometry, Parallel2dGeometry) and
+                parse_version(ASTRA_VERSION) < parse_version('1.9.9.dev')):
                 # parallel2d scales with pixel stride
                 out *= 1 / float(self.geometry.det_partition.cell_volume)
 
@@ -374,7 +375,10 @@ def astra_cuda_bp_scaling_factor(proj_space, reco_space, geometry):
     scaling_factor /= (reco_space.weighting.const /
                        reco_space.cell_volume)
 
-    if parse_version(ASTRA_VERSION) < parse_version('1.8rc1'):
+    if parse_version(ASTRA_VERSION) >= parse_version('1.9.9.dev'):
+        scaling_factor /= float(reco_space.cell_volume)
+        scaling_factor *= float(geometry.det_partition.cell_volume)
+    elif parse_version(ASTRA_VERSION) < parse_version('1.8rc1'):
         if isinstance(geometry, Parallel2dGeometry):
             # Scales with 1 / cell_volume
             scaling_factor *= float(reco_space.cell_volume)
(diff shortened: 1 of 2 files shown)


odl/tomo/backends/astra_cuda.py
```diff
@@ -374,7 +375,10 @@ def astra_cuda_bp_scaling_factor(proj_space, reco_space, geometry):
     scaling_factor /= (reco_space.weighting.const /
                        reco_space.cell_volume)
 
-    if parse_version(ASTRA_VERSION) < parse_version('1.8rc1'):
+    if parse_version(ASTRA_VERSION) >= parse_version('1.9.9.dev'):
+        scaling_factor /= float(reco_space.cell_volume)
+        scaling_factor *= float(geometry.det_partition.cell_volume)
+    elif parse_version(ASTRA_VERSION) < parse_version('1.8rc1'):
```
adler-j: Prefer consistent directions (e.g. `>` everywhere or `<` everywhere), it's a bit hard to follow the if statements atm.

wjp: Fair enough. I've added a separate commit to make things chronological.

I didn't do this before because the version currently in master is non-chronological, which made for a confusing diff. This separate commit to clean this up is hopefully the best of both worlds.

odl/tomo/backends/astra_cuda.py
```diff
@@ -463,10 +464,10 @@ def astra_cuda_bp_scaling_factor(proj_space, reco_space, geometry):
             # missing factor src_radius ** 2 in the ASTRA BP with
             # density weighting.
             det_px_area = geometry.det_partition.cell_volume
-            scaling_factor *= (src_radius ** 2 * det_px_area ** 2 /
-                               reco_space.cell_volume ** 2)
-
-        # TODO: add case with new ASTRA release
+            scaling_factor *= (src_radius ** 2 * det_px_area ** 2)
+    else:
```
adler-j: Scary catch-all else statement. Is this supposed to catch all above something? In that case be explicit about that and lets have a else with an error message.

adler-j: Could we just write that explicitly instead? Just easier to read (and less likely to be broken by some weird mistake later on). After that we merge.

wjp: Sure. I've added explicit comments about the version ranges to all four cases for consistency.

adler-j:
Ready to merge?

odl/tomo/backends/astra_cuda.py
kohr-h: IMO the catch-all `else:` doesn't really do any harm. If it were replaced with an `elif ... >1.9.9dev` and we somehow screwed up such that none of the cases applied, the function would still return a (most likely wrong) value.

The very cautious version would be the `elif ... >1.9.9dev` case, plus an `else:` case that raises a `RuntimeError`. But maybe normal reasoning is good enough here.

wjp:
Since it is also related to the 1.9.9.dev upgrade, I've also added the fix for the issue identified in #1529 directly to this PR. This should now cover the full `consistent_scaling` branch merge on the ASTRA side.

PR #1531, opened 2019-10-29
Doc: update doc for KullbackLeibler function. Closes #1218

### LINKED ISSUE #1218: Missmatch in KullbackLeibler functional and doc
The documentation says:

equation

while the implementation seems to implement

equation

while this may seem like a minor difference, it throws of computations where the scale is relevant.

Anyway, either we need to update the doc, or the implementation.

aringh:
Good to go?

### DIFF
--- a/odl/solvers/functional/default_functionals.py
+++ b/odl/solvers/functional/default_functionals.py
@@ -1038,8 +1038,8 @@ class KullbackLeibler(Functional):
         F(x)
         =
         \begin{cases}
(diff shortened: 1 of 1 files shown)


aringh:
The two pep8-issues were not in parts of the code that this commit touches. Moreover, they are in the documentation and of the form
```
        See Also
        --------
        odl.solvers.nonsmooth.proximal_operators.proximal_convex_conj_kl_cross_entropy :
```
which I do not see how one could shorten.

PR #1540, opened 2020-01-27
Support `RayTransform` custom backends

adriaangraas:
At this point it is difficult from the user level to specify a different backend for the `RayTransform` operator. (Current backends are Scikit-image, ASTRA CPU and ASTRA CUDA). Loading a custom backend is necessary if you like to modify the way ASTRA is called, for instance.

In this PR I propose to move some backend logic out of the `RayTransformBase` and `RayTransform` classes and into new `RayTransformImplBase` class and implementation classes. Something like this was already in place for `impl='astra_cuda'`. With this PR a user can for instance subclass the ASTRA CPU implementation and feed its type into the `impl` argument (strategy design pattern).

```python
    class MyRayTransformImpl(tomo.backends.AstraCpuRayTransformImpl):
        def call_forward(...):
            my_custom_stuff()

    ray_trafo = odl.tomo.RayTransform(reco_space, geometry, impl=MyRayTransformImpl)
```

Any implementation must subclass `RayTransformImplBase`, essentially forcing the `call_forward()` and `call_backward()`. There are some optional methods in the base class, such as `supports_geometry(geom)` and `supports_reco_space()`. 

Behavior and API for `RayTransform` and `RayBackProjection` remains backward compatible, in particular you can still specify `'astra_cpu`' etc. to `impl`. However, `AstraCudaProjectorImpl` and `AstraCudaBackProjectorImpl` are merged into `AstraCudaRayTransformImpl`. A lot of unnecessarily duplicated code has been removed there, and I didn't expect a lot of users to directly rely on these classes.

I'll make proper docs and fix the tests :) but I first wanted to send this in for some feedback on the concept.


odl/tomo/operators/ray_trafo.py
```diff
@@ -285,9 +186,96 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+                        "This warning can be disabled by explicitly setting "
+                        "`impl='astra_cpu'`.",
+                        RuntimeWarning)
+                elif impl_type == SkimageRayTransformImpl:
+                    warnings.warn(
+                        "The best available backend ('skimage') may be too "
+                        "slow for volumes of this size. Consider using ASTRA. "
+                        "This warning can be disabled by explicitly setting "
+                        "`impl='skimage'`.",
+                        RuntimeWarning)
```
kohr-h: Why this abstraction leak? In principle, the warning text could be part of the implementation class, so `RayTransformBase` wouldn't have to know about specific impls.

odl/tomo/operators/ray_trafo.py
```diff
@@ -396,41 +397,10 @@ def __init__(self, domain, geometry, **kwargs):
             variant='forward', **kwargs)
 
     def _call_real(self, x_real, out_real, **kwargs):
```
kohr-h: I think `_call_real` and `_call_complex` should also go into the implementation. That gives opportunity for optimization.

odl/tomo/backends/astra_cpu.py
```diff
@@ -29,6 +30,7 @@
     'astra_cpu_forward_projector',
     'astra_cpu_back_projector',
     'default_astra_proj_type',
+    'AstraCpuRayTransformImpl',
```
kohr-h: We shouldn't publicly expose these.

odl/tomo/backends/impl.py
```diff
@@ -0,0 +1,73 @@
...
+from __future__ import absolute_import, division, print_function
+
+from abc import abstractmethod, ABC
+from odl.discr import DiscreteLp
+from odl.tomo.geometry import Geometry
+
+__all__ = ('RayTransformImplBase',)
+
+
+class RayTransformImplBase(ABC):
```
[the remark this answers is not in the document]
adriaangraas: Good to know!

odl/tomo/backends/impl.py
```diff
@@ -0,0 +1,73 @@
...
+                            ' {!r}'.format(proj_space))
+
+        self.geometry = geometry
+        self.reco_space = reco_space
+        self.proj_space = proj_space
+
+    @staticmethod
+    def can_handle_size(size):
+        """A very general way to check if the implementation is capable
+        handling reconstruction volumes of a given size."""
```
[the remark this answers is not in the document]
adriaangraas: I moved this now into the `__init__`.

odl/tomo/backends/impl.py
```diff
@@ -0,0 +1,73 @@
...
+        return True  # by default no assumptions are made
```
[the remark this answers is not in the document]
adriaangraas: Now irrelevant since the base class is removed.

odl/tomo/backends/skimage_radon.py
```diff
@@ -171,3 +174,47 @@ def skimage_radon_back_projector(sinogram, geometry, vol_space, out=None):
...
+
+    @staticmethod
+    def can_handle_size(size):
+        return not size >= 256 ** 2
+
+    @classmethod
+    def supports_geometry(cls, geometry):
+        if not isinstance(geometry, Parallel2dGeometry):
+            return TypeError("{!r} backend only supports 2d parallel "
+                             'geometries'.format(cls.__name__))
```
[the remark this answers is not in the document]
adriaangraas: This indeed smells and I didn't really like it as well. I moved the checking logic into the `__init__` as well. There it makes sense to raise errors. This makes it impossible for the composing class to check ahead of time if the backend supports the geometry, but I think cleaner code is way more preferable anyway.

odl/tomo/operators/ray_trafo.py
adriaangraas: The motivation is to keep the same error messages, and a backend class is unaware of the others. The abstraction doesn't leak I believe, `RayTransform` is just aware of special cases. I changed the error messages a bit and moved this into the `__init__` of the backends as well.

odl/tomo/operators/ray_trafo.py
adriaangraas: This I think is a very good point. I did this, but I'm not entirely sure that I picked the best solution yet. Since the `_call` is quite a big function I didn't want to duplicate it throughout the implementations. I added a `complexify()` function for now, let me know what you think.

odl/tomo/backends/astra_cpu.py
adriaangraas: Indeed! Thanks.

adriaangraas:
Hi @kohr-h, thanks for reviewing!

I think you have fair points here. Although I like programming to interfaces (more to open the class for extension, not necessarily to standardize), I agree that OOP machinery easily ends up with a lot of boilerplate code. I tried removing the `RayTransformImplBase`, and I think that goes fairly well with just a few concessions: there is small bit of code duplication in the `__init__` of each implementation, and I think it'd have been a good place to implement complex `_call`s. I added a few runtime checks in `RayTransform` verifying the existence of `call_forward` and/or `call_backward`. All in all I think it simplified the code quite a bit. 

I also removed the `RayTransformBase` class, as you suggested. The `RayTransform`, and `RayBackProjection` are now directly derived from `Operator`. I completely agree with the former solution being suboptimal: subclassing, to have two `variants`, while simultaneously having the same variance in the forward and `adjoint()` is very confusing. Also, I don't think subclassing was the right pattern there to begin with. I'm not sure how you'd design the inner `adjoint` class, but I thought that a very natural way was to compose `RayBackProjection` _using_ instead of _extending_ a `RayTransform`. I gave this a try as well, let me know what you think.

I've two direct questions:
1. Now that I moved warnings into the implementations I cannot suppress the `size` warning, [like here](https://github.com/odlgroup/odl/blob/9382c4e290facde061cfda9ee8aa639a63745412/odl/tomo/backends/astra_cpu.py#L279). I could either add a `verbose` flag or try to catch the warnings, but neither seem very appealing. Do you have a good solution here?
2. I used a `complexify` function to turn real-valued functions into complex-valued functions. Is there a nicer way, maybe using ODL utilities, to achieve this?

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,164 +161,119 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+                                               geometry.ndim))
+
+        # Cache for input/output arrays of transforms
+        self.use_cache = kwargs.pop('use_cache', True)
+
+        # Check `impl`
+        impl = kwargs.pop('impl', None)
+        impl_type, self.__cached_impl = self._check_impl(impl)
+        self._impl_type = impl_type
+        self.__impl = impl.lower() if isinstance(impl, str) else impl_type
```
kohr-h: This I find a bit worrying. We try very hard to make every stored attribute a fixed type, or `None` if it makes sense, in particular for those that back public properties. Never one of several possible types. That's a recipe for downstream bugs.
I'd suggest you keep `impl` to always be a `str` and look up the corresponding type in a dictionary when needed. I don't think there's a need to store the type in the class. The fewer attributes, the better.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,164 +161,119 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+
+        self.geometry = geometry
```
kohr-h: This should ultimately be a `@property`, too.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,164 +161,119 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-            super(RayTransformBase, self).__init__(
-                domain=proj_space, range=reco_space, linear=True)
+        super(RayTransform, self).__init__(domain=reco_space,
+                                           range=proj_space, linear=True)
 
-    @property
-    def impl(self):
-        """Implementation back-end for the evaluation of this operator."""
-        return self.__impl
+    def _check_impl(self, impl):
```
kohr-h: This could be a `@staticmethod`.

odl/tomo/operators/ray_trafo.py
```diff
@@ -499,73 +345,51 @@ def __init__(self, range, geometry, **kwargs):
...
-            return self._adjoint
-
-        kwargs = self._extra_kwargs.copy()
-        kwargs['range'] = self.domain
-        self._adjoint = RayTransform(self.range, self.geometry,
-                                     impl=self.impl,
-                                     use_cache=self.use_cache,
-                                     **kwargs)
-        return self._adjoint
+        return copy.copy(self._ray_trafo)
```
kohr-h: See my suggestion for an inner class. It's largely the same, just without copying (which I don't think is necessary anyway).

odl/tomo/backends/astra_cpu.py
```diff
@@ -11,11 +11,13 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+import warnings
 
 from odl.discr import DiscreteLp, DiscreteLpElement
-from odl.tomo.backends.astra_setup import (
+from odl.tomo.backends import (
```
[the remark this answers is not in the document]
adriaangraas: Thanks! Good catch, not sure why I changed that.

odl/tomo/backends/astra_cpu.py
```diff
@@ -200,8 +200,8 @@ def astra_cpu_back_projector(proj_data, geometry, vol_space, out=None,
                         "".format(vol_space.impl))
     if vol_space.ndim != geometry.ndim:
         raise ValueError('dimensions {} of reconstruction space and {} of '
-                         'geometry do not match'.format(
-                             vol_space.ndim, geometry.ndim))
+                         'geometry do not match'.format(vol_space.ndim,
+                                                        geometry.ndim))
```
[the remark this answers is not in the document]
adriaangraas: Hm, I generally like to not have a end parenthesis on a newline, saving a line. Other than that I agree.

odl/tomo/backends/astra_cpu.py
```diff
@@ -249,6 +249,65 @@ def astra_cpu_back_projector(proj_data, geometry, vol_space, out=None,
     return out
 
 
+class AstraCpu:
```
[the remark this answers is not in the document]
adriaangraas: Fixed that! I need to properly check docs later.

odl/tomo/backends/astra_cuda.py
```diff
@@ -59,9 +60,39 @@ def __init__(self, geometry, reco_space, proj_space):
...
+        # single-axis geometry -- this is broken in some ASTRA versions
+        if (isinstance(geometry, Parallel3dAxisGeometry) and
+                not astra_supports('par3d_det_mid_pt_perp_to_axis')):
+            req_ver = astra_versions_supporting(
+                'par3d_det_mid_pt_perp_to_axis')
+            axis = geometry.axis
+            mid_pt = geometry.det_params.mid_pt
+            for i, angle in enumerate(geometry.angles):
+                if abs(np.dot(axis,
+                              geometry.det_to_src(angle, mid_pt))) < 1e-4:
```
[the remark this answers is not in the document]
adriaangraas: Agreed!

odl/tomo/backends/astra_cpu.py
```diff
@@ -249,6 +249,65 @@ def astra_cpu_back_projector(proj_data, geometry, vol_space, out=None,
...
+    def __init__(self, geometry, reco_space, proj_space):
+        """Initialize a new instance.
+
+        Parameters
+        ----------
+        geometry : `Geometry`
+            Geometry defining the tomographic setup.
+        reco_space : `DiscreteLp`
```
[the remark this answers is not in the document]
adriaangraas: Changed this everywhere. Hope nobody is calling `ray_trafo = RayTransform(reco_space=...)`?

odl/tomo/backends/astra_cuda.py
```diff
@@ -165,70 +145,78 @@ def create_ids(self):
...
-            adata.delete(self.vol_id)
-            self.vol_id = None
-        if self.sino_id is not None:
-            adata.delete(self.sino_id)
-            self.sino_id = None
-        if self.proj_id is not None:
-            aproj.delete(self.proj_id)
-            self.proj_id = None
+    def call_forward(self, x, out, **kwargs):
+        return complexify(self._call_forward_real, self.reco_space) \
```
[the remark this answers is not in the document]
adriaangraas: Irrelevant now with the decorator.

odl/tomo/backends/skimage_radon.py
```diff
@@ -11,13 +11,19 @@
 from __future__ import division
 
 import numpy as np
+import warnings
 
-from odl.discr import uniform_discr_frompartition, uniform_partition
+from odl.tomo import Parallel2dGeometry, Geometry
+
+from odl.discr import uniform_discr_frompartition, uniform_partition, \
```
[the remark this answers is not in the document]
adriaangraas: Done.

odl/tomo/backends/astra_cpu.py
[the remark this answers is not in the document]
adriaangraas: I was thinking this might clash with the ASTRA naming for a _Projector_, which is either a forward projector or a backprojector, while our class does both. I now named this `<backend>Impl`. What do you think?

odl/tomo/operators/ray_trafo.py
```diff
@@ -19,33 +20,34 @@
...
-ASTRA_CPU_AVAILABLE = ASTRA_AVAILABLE
-_SUPPORTED_IMPL = ('astra_cpu', 'astra_cuda', 'skimage')
+    ASTRA_AVAILABLE, ASTRA_CUDA_AVAILABLE, SKIMAGE_AVAILABLE)
+from odl.tomo.backends.astra_cpu import AstraCpu
+from odl.tomo.backends.astra_cuda import AstraCuda
+from odl.tomo.backends.skimage_radon import Skimage
+from odl.tomo.geometry import Geometry
+
+# Backends that are implemented in ODL and can be specified via `impl`
+_SUPPORTED_IMPL = {
```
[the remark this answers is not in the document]
adriaangraas: Sounds good.

odl/tomo/operators/ray_trafo.py
```diff
@@ -19,33 +20,34 @@
...
     _AVAILABLE_IMPLS.append('skimage')
 
-
 __all__ = ('RayTransform', 'RayBackProjection')
 
 
-class RayTransformBase(Operator):
-
+class RayTransform(Operator):
     """Base class for ray transforms containing common attributes."""
```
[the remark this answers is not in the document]
adriaangraas: Fixed!

odl/tomo/operators/ray_trafo.py
```diff
@@ -19,33 +20,34 @@
...
 
-    def __init__(self, reco_space, geometry, variant, **kwargs):
+    def __init__(self, reco_space, geometry, **kwargs):
```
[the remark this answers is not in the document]
adriaangraas: And everywhere else :)

odl/tomo/backends/astra_cpu.py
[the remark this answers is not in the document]
kohr-h: Ah yes, that's fine. `Impl` sounds so obscure that nobody will dare to use these classes directly :wink:

odl/tomo/operators/ray_trafo.py
adriaangraas: I changed `RayTransform.impl` so that it always returns a `str`. The reason why we currently have this hassle of storing the type is that `RayTransform` has "knowledge" of how to instantiate the implementation. Normally I wouldn't like that, but it makes sense because `RayTransform` needs to reinitiate the backend when `use_cache=False`. I guess I can also change the class so that it will directly use `type(self.__cached_impl)`, and then we'd have to make sure that there is always a `self.__cached_impl` available.

odl/tomo/operators/ray_trafo.py
adriaangraas: I did this, but why is this exactly? For autocompletion?

odl/tomo/backends/astra_cpu.py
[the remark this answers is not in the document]
kohr-h: I guess `reco_space` was so long and hard to type that nobody bothered and just used a positional argument. Some of our own examples might do that, though.

odl/tomo/operators/ray_trafo.py
adriaangraas: Changed this.

odl/tomo/operators/ray_trafo.py
adriaangraas: I like this. Makes `RayTransform` really clean and expressive. I was thinking that perhaps `.adjoint` had to return a new class (as it was before), but I don't think that is sensible with an `@property`.

odl/tomo/backends/astra_cpu.py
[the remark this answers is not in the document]
kohr-h: That's a matter of taste, sure. For other projects I nowadays use the `black` code formatter,  and it prefers the format I show above. Sometimes it's useful when you have chained calls, like 

    np.some_function(
        very_long_expression_that_fills_a_whole_line_or_more
    ).astype("float32")

But never mind.

odl/tomo/backends/astra_cpu.py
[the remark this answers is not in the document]
adriaangraas: Thanks, I missed this.

odl/tomo/operators/ray_trafo.py
kohr-h: The main reason is to make our objects "publicly immutable". You can't assign to properties unless a setter is defined, whereas the plain attributes can be changed arbitrarily. It's not perfect but at least conveys an intention.

adriaangraas:
Thanks again @kohr-h, I don't mind the nitty comments :+1: 

Open remain mostly decisions from the unresolved conversations above. Is there anything else that would be good to add to this PR?

About the decorator: Although the syntax is really clean in the class I don't like that it implicitly passes `self` through the call. I would have liked it better when it was more restricted to its input, or more widely applicable. I'll give it a thought.

In a next update I'll try to have a look at the docs and tests!

odl/tomo/backends/util.py
```diff
@@ -0,0 +1,26 @@
+__all__ = tuple()
+
+
+def _add_default_complex_impl(fn):
+    """Wrapper to call a class method twice when it is complex."""
+
+    def wrapper(self, x, out, **kwargs):
+        if self.vol_space.is_real and self.proj_space.is_real:
+            fn(self, x, out, **kwargs)
```
kohr-h: You're correct, `self` has to be supplied. Same for the complex case below.

adriaangraas: Sharp! Real-time feedback :)

kohr-h:
> Open remain mostly decisions from the unresolved conversations above. Is there anything else that would be good to add to this PR?

I'll go through it in detail again to get a better picture. But one thing that comes to my mind is the `impl` string. Maybe instead of using `__class__.__name__`, the `...Impl` classes should have a class attribute `impl_name` or so, and register themselves with that name in a global dictionary in `ray_trafo.py`. That smells a bit like danger of circular imports, but I think if the code in `ray_trafo.py` knows nothing about the specific `Impl`'s it shouldn't even have to import any of those modules.

Do you think that's reasonable?

(Or maybe that's already the current status.)

odl/tomo/operators/ray_trafo.py
kohr-h: Hm, yes, that makes sense. I'll have a closer look.

I recognize this problem from classes like `DiscreteLp` that are composed of other classes. The moment you want to "convert" the containing class, e.g., change the `dtype`, it suddenly needs to know how to instantiate its components. An annoying issue, and I haven't found a good solution yet.

odl/tomo/backends/util.py
kohr-h: Yes, but now it's late enough and my sharpness declines rapidly :-)

odl/tomo/operators/ray_trafo.py
adriaangraas: Thought about this. To get "creation knowledge" outside of the class we'd need either DI (not very ODL-style) or push it creation off to some external factory class. A bit of an overkill. Since `RayTransform` really composes and not "aggregates" `Impl`s, I don't think it is too weird for `RayTransform` to instantiate the `Impl`s. However, I don't think I'm a big fan of the root of this problem, the `use_cache` methodology. Caching/recreating the `Impl` is a way to prevent excessive memory usage after retrieving the volume data from GPU, while guaranteeing a `DiscreteLp` element. But do we not need to give users more explicit control over what is on the GPU in some future anyway? Like neural networks, that often implement `.numpy()` methods. And, if CuPy makes it into ODL, some methods may return CuPy arrays which probably shouldn't be cast into Numpy arrays by default.

At the moment I find only one example using `use_cache`. I'd think that we could either leave cache out, and people that need this can still `del ray_trafo` or reset `ray_trafo.impl`. Or maybe we'll leave it in until we find a good solution to the problems above?

odl/tomo/operators/ray_trafo.py
adriaangraas: That makes a lot of sense, thanks!

adriaangraas:
Hi @kohr-h. Hm, I don't know about the global dictionary, sounds like code that is hard to read and debug. I can see that this allows users to register their only class to the global register, which will then automatically be used (if the priorities go right) by `RayTransform`. I think the expressive `RayTransform(impl=MyImpl())` is way less magical and achieves the same thing. Furthermore `RayTransform` is already unaware of the default implementations, because it reads from `_AVAILABLE_IMPLS`. Maybe I am missing something here?

odl/tomo/backends/util.py
```diff
@@ -4,18 +4,17 @@
...
+            return fn(self, x, out, **kwargs)
         elif self.vol_space.is_complex and self.proj_space.is_complex:
             result_parts = [
                 fn(self, x.real, getattr(out, 'real', None), **kwargs),
                 fn(self, x.imag, getattr(out, 'imag', None), **kwargs)
             ]
 
             if out is None:
-                out = range.element()
+                out = self.proj_space.element()
```
adriaangraas: I need to fix this, maybe write a test to cover this?

kohr-h: Yes, that's a good idea.

kohr-h:
> _Update_: I guess we could make `_AVAILABLE_IMPLS` and `_IMPL_STR2TYPE` public though, so that amending at runtime is possible.

That's what I mean. Two things, then: maybe as a public dictionary, the name should be a bit more verbose. And we don't need the two above, `_AVAILABLE_IMPLS` could just be `_IMPL_STR2TYPE.keys()`.

odl/tomo/backends/skimage_radon.py
```diff
@@ -171,3 +177,79 @@ def skimage_radon_back_projector(sinogram, geometry, vol_space, out=None):
     out *= scaling_factor
 
     return out
+
+
+class SkimageImpl:
```
kohr-h: Suggestion: `SkImageImpl` (first "I" capitalized)

odl/tomo/operators/ray_trafo.py
kohr-h: These are all good thoughts and important to keep in mind. For the moment, though, I think the current solution is (more than) acceptable. One of the points of this PR is to make the `RayTransform` extensible by using a common interface for the `Impl` classes, which makes it easy to instantiate them correctly in the caching logic.
Regarding future-proofness of the caching: ASTRA (as of now) works with host memory via copying or device memory via linking, and with the current mechanism we have the first case covered. At some point we'll modify the caching logic to account for existing GPU memory. That will require a bit of restructuring, but I don't see more cases than that.

odl/tomo/backends/skimage_radon.py
adriaangraas: Good!

odl/tomo/backends/util.py
adriaangraas: Fixed this and updated the complex test to cover this.

odl/tomo/operators/ray_trafo.py
adriaangraas: I agree that this is not for this PR :)

adriaangraas:
I've changed `_IMPL_STR2TYPE` to an `OrderedDict` named `RAY_TRAFO_IMPLS`. Maybe verboseness is good here when the variable goes public? I'm also fine with the other name. The last inserted item now automatically has the highest priority.
I didn't manage to remove the second global without introducing much extra code. `RayTransform` needs to have different lists in order to tell the user that an `impl` exists but is not installed. That list can remain private though.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-
-            return out
-
-        else:
-            raise RuntimeError('bad domain {!r}'.format(self.domain))
+        super(RayTransform, self).__init__(domain=vol_space,
+                                           range=proj_space, linear=True)
 
+    @staticmethod
+    def _check_impl(impl):
```
kohr-h: This guy does a lot more than mere checking. Maybe `_initialize_impl`?

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+        If a custom ``impl`` was provided this method returns a ``str``
+        of the type."""
 
-    """Adjoint of the discrete Ray transform between L^p spaces."""
+        return self.__impl
 
-    def __init__(self, range, geometry, **kwargs):
-        """Initialize a new instance.
+    def create_impl(self, use_cache=True):
+        """Fetches or instantiates implementation backend for evaluation.
```
kohr-h: Those two cases are perhaps best summarized by the name `get`.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-            Implementation back-end for the transform. Supported back-ends:
-
-            - ``'astra_cuda'``: ASTRA toolbox, using CUDA, 2D or 3D
-            - ``'astra_cpu'``: ASTRA toolbox using CPU, only 2D
-            - ``'skimage'``: scikit-image, only 2D parallel with square
-              reconstruction space.
-
-            For the default ``None``, the fastest available back-end is
-            used, tried in the above order.
+        bool : use_cache
```
kohr-h: Please mirror at the colon

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+            If ``True`` returns the cached implementation backend, if it
```
kohr-h:
```suggestion
            If ``True``, return the cached implementation backend, if it
```

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+            was generated in a previous call (or given with ``__init__``).
+            If ``False`` a new instance of the backend will be generated,
```
kohr-h:
```suggestion
            If ``False``, a new instance of the backend will be generated,
```

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-        astra_cpu_back_projector
-        AstraCudaBackProjectorImpl
-        skimage_radon_back_projector
-        """
-        domain = kwargs.pop('domain', None)
-        super(RayBackProjection, self).__init__(
-            reco_space=range, proj_space=domain, geometry=geometry,
-            variant='backward', **kwargs)
+    def _call(self, x, out=None, **kwargs):
+        """Forward projection
```
kohr-h:
```suggestion
        """Forward projection.
```

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
 
-    def _call_real(self, x_real, out_real, **kwargs):
-        """Real-space back-projection for the current set-up.
+        Parameters
+        ----------
+        x : DiscreteLpElement
+            A volume. Must be an element from `RayTransform.domain`.
```
kohr-h:
```suggestion
            A volume. Must be an element of `RayTransform.domain`.
```

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+        out : `RayTransform.range` element, optional
+            A DiscreteLpElement from `RayTransform.range`, to which the result
+            of the operator evaluation is written.
```
kohr-h:
```suggestion
        out : `RayTransform.range` element, optional
            Element to which the result of the operator evaluation is written.
```
I think this doc can be quite minimal, it's only for devs and does not deviate a lot from `Operator._call` anyway. I'd also be okay with a one-liner.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+        **kwargs
+            Arbitrary keyword arguments, passed on to the implementation
+            backend.
```
kohr-h:
```suggestion
            Extra keyword arguments, passed on to the implementation
            backend.
```
"Arbitrary" goes a bit too far IMO

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-                        self._astra_wrapper = astra_wrapper
-                else:
-                    astra_wrapper = self._astra_wrapper
 
-                return astra_wrapper.call_backward(x_real, out_real, **kwargs)
-            else:
-                # Should never happen
-                raise RuntimeError('bad `impl` {!r}'.format(self.impl))
+        return self.create_impl(self.use_cache) \
+            .call_forward(x, out, **kwargs)
```
kohr-h:
```suggestion
        return self.create_impl(self.use_cache).call_forward(
            x, out, **kwargs
        )
```

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+        **kwargs
+            Arbitrary keyword arguments, passed on to the implementation
+            backend.
 
-        This method also sets ``self._astra_backprojector`` for
-        ``impl='astra_cuda'`` and enabled cache.
+        Returns
+        -------
+        DiscreteLpElement
+            A sinogram, from the projection space `RayTransform.range`.
```
kohr-h: Not every ray transform results in a (single) sinogram. Maybe keep it generic, like "Result of the projection" or so.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-                                     **kwargs)
+        if self._adjoint is None:
+            # bring `self` into scope to prevent shadowing in inline class
+            ray_trafo = self
+
+            class RayBackProjection(Operator):
+                """Adjoint of the discrete Ray transform between L^p spaces."""
+
+                def _call(self, x, out=None, **kwargs):
+                    """Backprojection
```
kohr-h:
```suggestion
                    """Backprojection.
```

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+                        evaluation is written into this variable.
+                    **kwargs
+                        Arbitrary keyword arguments, passed on to the
+                        implementation backend.
+
+                    Returns
+                    -------
+                    DiscreteLpElement
+                        The reconstructed volume, an element in
+                        `RayBackProjection.range`.
```
kohr-h: See above

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+                    """
+
+                    return ray_trafo.create_impl(ray_trafo.use_cache) \
+                        .call_backward(x, out, **kwargs)
```
kohr-h:
```suggestion
                    return ray_trafo.create_impl(ray_trafo.use_cache).call_backward(
                        x, out, **kwargs
                    )
```

adriaangraas:
Thank you @kohr-h! Going to pick this up very soon (days).

ozanoktem:
I want to thank both @adriaangraas and @kohr-h for the effort in providing new functional to ODL. Your work is much appreciated.

kohr-h:
@adriaangraas I'm currently rebasing on master. There has been a rather big internal API change (#1459) that impacts this PR, so I think it's easiest if I deploy the necessary changes here so you can focus on the actual content of the PR. Is that okay? So don't worry about an upcoming force-push :wink:

kohr-h:
Okay, I pushed a couple of commits: the rebase, fixes of rebase accidents and some formatting updates. Please take it from here @adriaangraas

odl/tomo/backends/astra_cuda.py
```diff
@@ -10,117 +10,107 @@
...
 
-from builtins import object
 from multiprocessing import Lock
 
 import numpy as np
+import warnings
 from packaging.version import parse as parse_version
 
 from odl.discr import DiscreteLp
+from odl.tomo.backends.util import _add_default_complex_impl
```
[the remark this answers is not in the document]
adriaangraas: Thanks! I've now checked a box for sorting imports, I hope that is consistent with `isort`, otherwise I'll configure that.

odl/tomo/backends/astra_cuda.py
```diff
@@ -10,117 +10,107 @@
...
 
         self.geometry = geometry
-        self.reco_space = reco_space
-        self.proj_space = proj_space
+        self._vol_space = vol_space
+        self._proj_space = proj_space
 
         self.create_ids()
 
         # ASTRA projectors are not thread-safe, thus we need to lock ourselves
```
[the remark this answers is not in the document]
adriaangraas: Haha true, the scope of this PR was creeping a bit anyway :smile:

odl/tomo/backends/skimage_radon.py
```diff
@@ -11,13 +11,19 @@
 from __future__ import division
 
 import numpy as np
+import warnings
 
-from odl.discr import uniform_discr_frompartition, uniform_partition
+from odl.tomo import Parallel2dGeometry, Geometry
```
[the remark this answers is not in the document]
adriaangraas: Missed this, you mean to import directly via `odl.discr.discr_space`? Here we are in `odl.tomo.backends`?

odl/tomo/backends/util.py
```diff
@@ -0,0 +1,44 @@
...
+    ----------
+    fn : Callable
+        Function with signature ``fn(self, x, out=None, **kwargs)``.
+        ``self`` must be an object instance having ``self.vol_space`` and
+        ``self.proj_space``. These spaces must be both real or both complex.
+
+    Returns
+    -------
+    Callable
+        A wrapper function to be executed by a Python decorator.
```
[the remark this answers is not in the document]
adriaangraas: Ok! I've changed this a bit, let me know if this is what you meant.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+
+        # Cache for input/output arrays of transforms
+        self.use_cache = kwargs.pop('use_cache', True)
+
+        # Check `impl`
+        impl = kwargs.pop('impl', None)
+        impl_type, self.__cached_impl = self._check_impl(impl)
+        self._impl_type = impl_type
+        self.__impl = impl.lower() \
+            if isinstance(impl, str) else impl_type.__name__
```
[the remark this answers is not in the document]
adriaangraas: Agreed! Did your formatter fix this (is that `black`?). Maybe I should set that up as well...

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-                        self._astra_wrapper = astra_wrapper
+        else:
+            # User did specify `impl`
+            if isinstance(impl, str):
+                if impl.lower() not in _ALL_IMPLS:
+                    raise ValueError('`impl` {!r} not understood'.format(impl))
+
+                if impl.lower() not in RAY_TRAFO_IMPLS.keys():
+                    raise ValueError(
+                        '{!r} back-end not available'.format(impl))
```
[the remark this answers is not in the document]
adriaangraas: Maybe this would have been better as a class constant, but I'm also in favor of simpler code. Removed it.

odl/tomo/backends/skimage_radon.py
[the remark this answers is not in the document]
kohr-h: No, I mean the `tomo` imports. Maybe this is based on me misunderstanding the Python import system, but it seems to me that it would be easier to break import cycles when import paths are deeper.
The above version is actually a cyclic import, because it says "import the `odl.tomo`, subpackage, *which includes myself*, and give me its attributes `Parallel2dGeometry` and `Geometry`.

odl/tomo/operators/ray_trafo.py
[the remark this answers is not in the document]
kohr-h: I use `black` privately, but not on ODL - I just use its suggested style. Using the tool would be a huge time-saver, but also a huge commit that screws up Git history badly. Maybe we can do it at some point anyway :-)

odl/tomo/backends/skimage_radon.py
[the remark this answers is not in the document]
adriaangraas: I guess it makes sense not to import from your parent indeed. Wonder what happens though at import-time: if the parent is already imported, then Python may not re-import the parent again? :thinking:

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +158,217 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
-            elif data_impl == 'cuda':
-                if self._astra_wrapper is None:
-                    astra_wrapper = AstraCudaProjectorImpl(
-                        self.geometry, self.domain.real_space,
-                        self.range.real_space)
-                    if self.use_cache:
-                        self._astra_wrapper = astra_wrapper
+        else:
+            # User did specify `impl`
+            if isinstance(impl, str):
```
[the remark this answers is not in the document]
adriaangraas: Did that. Note that there is a `.lower()` required later in the class though. Was there ever a use case for this, curious about it.

odl/tomo/operators/ray_trafo.py
adriaangraas: For me that reminds me of the dumb getters/setters of many other languages, but that might not be a valid reason. Changed it.

odl/tomo/operators/ray_trafo.py
adriaangraas: I'll keep that in mind also for other docs.

odl/tomo/operators/ray_trafo.py
adriaangraas: Agreed. Changed at two places.

odl/tomo/operators/ray_trafo.py
adriaangraas: Recently tried to stop with the backslash habit, but with `get_impl` the line fits!

odl/tomo/operators/ray_trafo.py
adriaangraas: Sounds good.

odl/tomo/operators/ray_trafo.py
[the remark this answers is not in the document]
adriaangraas: A quick search finds that this is possible without destroying Git history, but at the cost that every developer needs to re-clone their repository. Seems reasonable.

odl/tomo/backends/util.py
```diff
@@ -0,0 +1,71 @@
...
+    complex calls into two individual real calls.
+
+    The wrapper will only work for methods of which the class provides a
+    `vol_space` and `proj_space`. Typically, this will then work as a decorator
+    on the method, e.g.:
+    ```
+        @_add_default_complex_impl
+        def call_forward(self, x, out=None, **kwargs):
+            # Code that should run for real input and output
+    ```
```
kohr-h, line 30:
```suggestion
    on the method, e.g. ::

        @_add_default_complex_impl
        def call_forward(self, x, out=None, **kwargs):
            # Code that should run for real input and output
```

Note that the docstrings are parsed as reStructuredText, not Markdown. So the triple-quote thing won't work.

odl/tomo/operators/ray_trafo.py
```diff
@@ -273,299 +163,225 @@ def __init__(self, reco_space, geometry, variant, **kwargs):
...
+        else:
+            # User did specify `impl`
+            if is_string(impl):
+                if impl.lower() not in RAY_TRAFO_IMPLS.keys():
+                    raise ValueError(
+                        'The {!r} `impl` is not found. This `impl` is either '
+                        'not supported, it may be misspelled, or external '
+                        'packages required are not available. Consult '
+                        '`RAY_TRAFO_IMPLS` to find the run-time available '
+                        'implementations.'.format(impl)
```
kohr-h, line 238: Great error message!

kohr-h:
> One little thing

And I just did it myself :-)

kohr-h:
Not waiting for CI. The last commit built fine, so I'll just merge it.

kohr-h:
Thanks a lot for all the work @adriaangraas!

adriaangraas:
And thanks a ton for all the guidance @kohr-h, I've also learned a lot!

odl/tomo/operators/ray_trafo.py
adriaangraas: Thanks :D

Issue #1135, opened 2017-09-20
Add "CreateMovie" callback


adler-j:
Creating movies like:

https://adler-j.github.io/images/mayo_fbp_animation.mp4

Can be quite illustrative and useful for several cases. It would be great if we had a callback that automated this.

kohr-h:
Great idea! Some inspiration available [here](http://jakevdp.github.io/blog/2012/08/18/matplotlib-animation-tutorial/).

adler-j:
Also posting the code I used to generate the above movie:

https://gist.github.com/adler-j/b3e935620af051a36abb41a50d788cf0

miallo:
I just scribbled this down, and there is lots of room for improvements, but if someone else wants to take this as a starting point:

```python
import sys
sys.path.append('/Users/michael/Uni/Master/odl')

import odl
from odl.solvers import Callback
import matplotlib
matplotlib.use('Agg')
import matplotlib.animation as animation
import matplotlib.pyplot as plt
import contextlib


class CallbackMovieWriter(Callback):
    """
    Callback for generating movies with matplotlib.
    use as context manager / like this:
    ```
    with CallbackMovieWriter().saving('movie.mp4') as movieCallback:
        odl.solver.adam(..., callback=movieCallback)
    ```
    """
    def __init__(self, codec='ffmpeg',
            processing_function=None,
            movieWriterSaving_kwargs=dict(dpi=100),
            animation_kwargs=dict(fps=20, metadata=None),
            step=1):

        self.movieWriter = animation.writers[codec](**animation_kwargs)

        self.processing_function = processing_function

        try:
             self.step = frozenset(int(i) for i in step)
             self.should_evaluate_at_step = lambda i: i in self.step
        except TypeError:
             self.step = int(step)
             self.should_evaluate_at_step = lambda i: i % self.step == 0
        self.iter = 0


    def __call__(self, x):
        if self.should_evaluate_at_step(self.iter):
            if self.processing_function is not None:
                local_x = self.processing_function(x)
            else:
                local_x = x

            self.im.set_array(local_x)
            self.movieContextManager.grab_frame()
        self.iter += 1

    @contextlib.contextmanager
    def saving(self, fig, im, outfile, dpi=200, *args, **kwargs):
        self.im = im
        with self.movieWriter.saving(fig, outfile, dpi, *args, **kwargs) as mwwrite:
            self.movieContextManager = mwwrite
            yield self



if __name__ == "__main__":
    import numpy as np
[...]

PR #1544, opened 2020-03-05
Added CallbackMovie

miallo:
Code for #1135

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -23,7 +23,7 @@ __all__ = ('Callback', 'CallbackStore', 'CallbackApply', 'CallbackPrintTiming',
            'CallbackPrintIteration', 'CallbackPrint', 'CallbackPrintNorm',
            'CallbackShow', 'CallbackSaveToDisk', 'CallbackSleep',
            'CallbackShowConvergence', 'CallbackPrintHardwareUsage',
-           'CallbackProgressBar', 'CallbackMovie')
+           'CallbackProgressBar', 'save_animation')
 
 
 class Callback(object):
@@ -1036,8 +1036,75 @@ class CallbackProgressBar(Callback):
                                    inner_str)
 
 
-class CallbackMovie(Callback):
+@contextlib.contextmanager
+def save_animation(filename,
+                   writer="ffmpeg",
+                   writer_kwargs=None,
+                   dpi=200,
+                   saving_kwargs=None,
+                   fig=None,
+                   step=1):
+    """Context manager for creating animations using `matplotlib.animation` [1]
+    [1] https://matplotlib.org/3.2.0/api/animation_api.html
 
+    Parameters
+    ----------
+    filename : str
+    writer : string
+        Available writers: `matplotlib.animation.writers.list()`, examples:
+        'ffmpeg' -> .mp4 file, 'pillow' -> .gif file
+    writer_kwargs : dict
+        Containing e.g. `fps` and a dict of `metadata` like
+        `title`, `author`, ...
+    saving_kwargs : dict
+        Containing e.g. `fps` and a dict of `metadata` like
+    fig : matplotlib figure, optional
+        Use the provided figures last imshow, otherwise create a new figure
+    step : positive int, list, optional
+        Number of iterations between frames, or
+        list of iterations where frames should be aquired.
+    """
+    if writer_kwargs is None:
+        writer_kwargs = {}
+    if saving_kwargs is None:
+        saving_kwargs = {}
+    import matplotlib.animation
+    writer = matplotlib.animation.writers[writer]
+    moviewriter = writer(**writer_kwargs)
+    iter = 0
+    if fig is None:
+        import matplotlib.pyplot
+        fig, ax = matplotlib.pyplot.subplots()
+        im = None
+    else:
+        try:
+            ax = fig.axes[-1]
+            im = ax.get_images()[-1]
+        except IndexError:
+            raise ValueError('Matplotlib figure fig={} must contain '
+                             'imshow plot'.format(fig))
+
+    class CallbackAppendMovieFrame(Callback):
+        def __call__(self, x):
+            # Now I remember why I wanted to create the figure outside of
+            # the contextManager, where I know about the dimensions of the axes
+            nonlocal im
+            if iter % step == 0:
+                if im is None:
+                    im = matplotlib.pyplot.imshow(x)
+                else:
+                    im.set_array(x)
+                moviewriter.grab_frame()
+
+    try:
+        with moviewriter.saving(fig, filename, dpi=dpi, **saving_kwargs):
+            yield CallbackAppendMovieFrame()
+    finally:
+        # Whatever needs to be finalized, closed, ...
+        pass
+
+
+class CallbackMovie(Callback):
     """Callback for generating movies with matplotlib.
     Must be used as context manager with the ``saving`` method
 
@@ -1051,14 +1118,14 @@ class CallbackMovie(Callback):
     ...     callback([[3,2],[1,0]])
 
     """
-    def __init__(self, codec='ffmpeg',
+    def __init__(self, writer='ffmpeg',
                  animation_kwargs=dict(fps=20, metadata={}),
                  step=1):
         """Initialize a new instance.
 
         Parameters
         ----------
-        codec : string
+        writer : string
             See documentation of matplotlib.animation.writers, examples:
             'ffmpeg' -> .mp4 file, 'pillow' -> .gif file
         animation_kwargs : dict
@@ -1070,7 +1137,7 @@ class CallbackMovie(Callback):
             list of iterations where frames should be aquired.
         """
         import matplotlib.animation
-        writer = matplotlib.animation.writers[codec]
+        writer = matplotlib.animation.writers[writer]
         self.moviewriter = writer(**animation_kwargs)
 
         try:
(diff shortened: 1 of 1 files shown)


miallo:
What is the minimum supported Python2 version? If it is 2.5 or below we need `from __future__ import with_statement`

kohr-h:
> What is the minimum supported Python2 version? If it is 2.5 or below we need `from __future__ import with_statement`

No, the minimum is 2.7. See here:

https://github.com/odlgroup/odl/blob/0d06b07c80edfd1acb945d6d86e1d65f6385f9d0/setup.cfg#L30-L35

and here:

https://github.com/odlgroup/odl/blob/0d06b07c80edfd1acb945d6d86e1d65f6385f9d0/setup.cfg#L46

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,96 @@ def __repr__(self):
...
+
+    """Callback for generating movies with matplotlib.
+    Must be used as context manager with the ``saving`` method
+
+    Examples
+    --------
+    >>> import matplotlib.pyplot as plt
+    >>> fig = plt.figure()
+    >>> _ = plt.imshow([[0,1],[2,3]])
+    >>> with CallbackMovie('pillow').saving(fig, "test.gif") as callback:
```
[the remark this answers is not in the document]
miallo: I just blindly copied the `matplotlib.animation` behaviour. No further reason

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,96 @@ def __repr__(self):
...
+    ...     callback([[0,1],[2,3]])
+    ...     callback([[3,2],[1,0]])
```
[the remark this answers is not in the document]
miallo: I thought that an example for the user would be good (especially with the `.saving` construct), it wasn't primarily intended to be a doctest ;)

miallo:
> My suggestion would be to pull the context manager out of the class and make it a free context manager that provides a callback when entered.

One thing to consider is: if users type `odl.solvers.Callback` and look at the autocomplete, they will not see this. I know it isn't the best reason, but this is how I discovered the ones I use.


> Whether the figure should be created outside or in the context manager I don't really know yet. I guess most of the time one would use this with the `agg` backend anyway that doesn't show anything on screen. Is there a specific reason you put it outside in the current implementation, @miallo?

I did this to easily set up the colormap, and x/y-limits inside the `plt.imshow`  but I agree that it isn't nice. And you can do it with `rcParams["image.cmap"] = 'viridis'`, so creating it is a good idea. It does make the handling easier and less error prone.

kohr-h:
> > My suggestion would be to pull the context manager out of the class and make it a free context manager that provides a callback when entered.
> 
> One thing to consider is: if users type `odl.solvers.Callback` and look at the autocomplete, they will not see this. I know it isn't the best reason, but this is how I discovered the ones I use.

But also not the worst :-)
In hindsight, I guess we over-optimized a bit for discoverability by naming everything `Callback...`. When names are a bit less uniform, people will look around more, perhaps even in the docs :+1: 

That said, I'd still advocate the standalone context manager.



> > Whether the figure should be created outside or in the context manager I don't really know yet. I guess most of the time one would use this with the `agg` backend anyway that doesn't show anything on screen. Is there a specific reason you put it outside in the current implementation, @miallo?
> 
> I did this to easily set up the colormap, and x/y-limits inside the `plt.imshow` but I agree that it isn't nice. And you can do it with `rcParams["image.cmap"] = 'viridis'`, so creating it is a good idea. It does make the handling easier and less error prone.

Ah I see. Indeed, global settings can be modified outside using matplotlib `rcParams` (or the `.rc()` function). But maybe it's really useful to have a handle outside to the figure and thereby also to its contents. If you agree, make it optional so users aren't forced to instantiate it but can if they want.

miallo:
The current implementation of the whole callback only works for (2D-)images - if it was used for 1D plots as well, there are a couple of things that had to be changed...


> Is there a specific reason you put [the figure] outside in the current implementation, @miallo?

Now I remember the main reason why I did this: I cannot create a `plt.imshow` instance without an array of the shape of the images. The workaround is in my last commit, but it isn't nice...

kohr-h:
Thanks for the update, looking very promising.

> The current implementation of the whole callback only works for (2D-)images - if it was used for 1D plots as well, there are a couple of things that had to be changed...

I think it's fine for now to only support the most relevant case. To make that clearer, the callback could raise a `NotImplementedError` if `x.ndim != 2` (or `not hasattr(x, 'ndim')`.


> > Is there a specific reason you put [the figure] outside in the current implementation, @miallo?
> 
> Now I remember the main reason why I did this: I cannot create a `plt.imshow` instance without an array of the shape of the images. The workaround is in my last commit, but it isn't nice...

Ah, that makes sense. I don't think the workaround is that bad, though. To make it even nicer, you could move the `im` logic fully into the callback, i.e., get the `ax` object outside and in each call 
```py
try:
    im = ax.get_images()[-1]
except IndexError:
    im = ax.imshow(x)
```
That would remove the necessity to overwrite a `nonlocal`.

---

BTW, we don't *have* to create a `Callback` class internally. The solvers also accept a plain Python function, so if you like you can scrap the class and just write the `__call__` part as a function that you yield instead of the class instance. Up to you.

miallo:
> BTW, we don't _have_ to create a `Callback` class internally. The solvers also accept a plain Python function, so if you like you can scrap the class and just write the `__call__` part as a function that you yield instead of the class instance. Up to you.

A `Callback` has the one advantage that you can easily combine it with an odl-operator by "multiplication"...

kohr-h:
> > BTW, we don't _have_ to create a `Callback` class internally. The solvers also accept a plain Python function, so if you like you can scrap the class and just write the `__call__` part as a function that you yield instead of the class instance. Up to you.
> 
> A `Callback` has the major advantage that you can easily combine it with an odl-operator by "multiplication" and that you can combine multiple callbacks with `&`...

Good point, let's stick to `Callback` then.

miallo:
I extracted `plot_in_figure` into a separate method if one wants to plot non 2D data. Is this unnecessary?

What else needs to be done?

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+        'ffmpeg' -> .mp4 file, 'pillow' -> .gif file
+    writer_kwargs : dict
+        Containing e.g. `fps` and a dict of `metadata` like
+        `title`, `author`, ...
+        Passed to `matplotlib.animation.writer`.
+    saving_kwargs : dict
+        Containing e.g. `fps` and a dict of `metadata` that may contain
+        title, artist, genre, subject, copyright, srcform, comment.
+        Passed to `matplotlib.animation.writer.saving`.
+    fig : matplotlib figure, optional
```
kohr-h:
```suggestion
    fig : matplotlib.figure.Figure, optional
```
This might be linked correctly by Sphinx.

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+        Matplitlib figure containing at least one imshow-object,
```
kohr-h: I don't think that's required anymore.

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+        otherwise create a new figure
```
kohr-h:
```suggestion
        For the default ``None``, a new figure is created.
```

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+    step : positive int, optional
+        Number of iterations between frames
+    """
+
+    if writer_kwargs is None:
+        writer_kwargs = {}
+    if saving_kwargs is None:
+        saving_kwargs = {}
+    import matplotlib.animation
```
kohr-h: Please gather lazy imports at the top of the function.

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+        Containing e.g. `fps` and a dict of `metadata` that may contain
+        title, artist, genre, subject, copyright, srcform, comment.
+        Passed to `matplotlib.animation.writer.saving`.
+    fig : matplotlib figure, optional
+        Matplitlib figure containing at least one imshow-object,
+        otherwise create a new figure
+    step : positive int, optional
+        Number of iterations between frames
+    """
+
```
kohr-h:
```suggestion
```
(no blank line)

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+    if writer_kwargs is None:
+        writer_kwargs = {}
+    if saving_kwargs is None:
+        saving_kwargs = {}
+    import matplotlib.animation
+    writer = matplotlib.animation.writers[writer]
+    moviewriter = writer(**writer_kwargs)
+    iter = 0
+    if fig is None:
+        import matplotlib.pyplot
```
kohr-h: It's almost standard to `import matplotlib.pyplot as plt`

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+        fig, ax = matplotlib.pyplot.subplots()
+    else:
+        ax = fig.axes[-1]
+
+    class CallbackAppendMovieFrame(Callback):
+        """Callback-helper for appending frames to an animation
```
kohr-h:
```suggestion
        """Callback for appending frames to an animation.
```

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
                                    inner_str)
 
 
+@contextlib.contextmanager
+def save_animation(filename,
```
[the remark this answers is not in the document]
kohr-h: Apparently it doesn't

odl/solvers/util/callback.py
```diff
@@ -1035,6 +1036,82 @@ def __repr__(self):
...
+                   writer="ffmpeg",
+                   writer_kwargs=None,
+                   dpi=200,
```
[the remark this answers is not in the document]
kohr-h: Actually it's a mandatory parameter in `saving`, so we keep it here.

kohr-h:
I pushed a few doc and default behavior improvements, after that I'll merge.

Issue #1548, opened 2020-04-01
Incorrect parameter formatting in documentation


kohr-h:
For a while now, the parameters of functions are not displayed correctly in the documentation. See for instance [here](https://odlgroup.github.io/odl/generated/odl.discr.grid.sparse_meshgrid.html#odl.discr.grid.sparse_meshgrid). It seems as if the colon between name and type of the parameter gets swallowed and the two get smashed together.

I saw the same issue in the doc window of the Spyder IDE, so I suspect that it has to do with a Sphinx update. Maybe numpydoc needs to be patched.

kohr-h:
This is a known issue in the Sphinx RTD theme: https://github.com/readthedocs/sphinx_rtd_theme/issues/746

A somewhat brutal fix that we could import: https://github.com/geopandas/geopandas/pull/1299

kohr-h:
Indeed fixed by #1549, so closing this for now. We should observe the sphinx-rtd-theme issue.

kohr-h:
The `sphinx_rtd_theme` issue has been closed as solved. We can remove our workaround as soon as a release is availble. I'll open a new issue.

PR #1549, opened 2020-04-01
Fix for parameter formatting issue

kohr-h:
Potential fix for #1548


kohr-h:
According to the build log, the doc build succeeds and static files are being copied. I'll merge to see whether it has the desired effect.

Issue #1514, opened 2019-07-18
pytest.config deprecated and now failing


kohr-h:
New trouble: we're getting tons of these kinds of errors in the tests now:
```
________ ERROR at setup of test_example[ray_trafo_parallel_2d_complex] _________
Error evaluating 'skipif' expression
    not pytest.config.getoption('--examples')
pytest.PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
```

The reason for this is a deprecation already quite a while back, documented [here](https://docs.pytest.org/en/latest/deprecations.html#pytest-config-global).

We need to change the implementation of conditional test skipping, e.g. to what is outlined in [this example](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option). Seems quite doable to me.

In the meanwhile, we can temporarily switch the errors back to warnings as shown [here](https://docs.pytest.org/en/latest/changelog.html#removals), but that only buys us time until pytest 5.1. We then need to restrict pytest to `<5.1` in our CI until the issue is solved.

kohr-h:
The temporary fix is in master now, but I'll keep this open until the issue is actually solved.

PR #1553, opened 2020-04-13
TST: Fix largescale pytest mark issue

kohr-h:
We change the command-line option to a version recommended in
the pytest docs:
https://docs.pytest.org/en/latest/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs

- A global `suite(name)` mark is registered, that takes a
  name and marks a test as belonging to a specific suite.
- These special suites are opt-in, i.e., not run by default.
- To enable a suite, one adds `-S name` to the pytest
  command-line options.
- Since function-scoped fixtures are handled stricter now,
  the workaround in tensors_test.py didn't work any longer
  (see https://github.com/pytest-dev/pytest/issues/6497).
- We thus set the minimum pytest version to 5.4.0, where the
  fix for the issue is implemented.

Closes: #1514

### DIFF
--- a/odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
+++ b/odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
@@ -282,4 +282,4 @@ def test_proximal_convex_conj_kl_cross_entropy_solving_opt_problem():
 
 
 if __name__ == '__main__':
-    odl.util.test_file(__file__, ['-S', 'slow'])
+    odl.util.test_file(__file__, ['-S', 'largescale'])
--- a/odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
+++ b/odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2019 The ODL contributors
+# Copyright 2014-2020 The ODL contributors
 #
 # This file is part of ODL.
 #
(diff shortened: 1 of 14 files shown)


odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
```diff
@@ -279,4 +282,4 @@ def test_proximal_convex_conj_kl_cross_entropy_solving_opt_problem():
 
 
 if __name__ == '__main__':
-    odl.util.test_file(__file__, ['--largescale'])
+    odl.util.test_file(__file__, ['-S', 'largescale'])
```
adler-j, line 285: Shouldn't `'-S largescale'` be more appropriate?

odl/test/test_doc.py
```diff
@@ -55,13 +55,10 @@
...
-def doc_src_file(request):
-    return request.param
+doc_src_file = simple_fixture("doc_src_file", doc_src_files)
 
 
-@pytest.mark.skipif("not pytest.config.getoption('--doctest-doc')",
-                    reason='Need --doctest-doc option to run')
+@pytest.mark.suite("doc_doctests")
 def test_file(doc_src_file):
     # FIXXXME: This doesn't seem to actually test the file :-(
```
adler-j, line 63: Probably should be TODO :)

odl/util/pytest_config.py
```diff
@@ -8,21 +8,30 @@
 
 """Test configuration file."""
 
-from __future__ import print_function, division, absolute_import
-import numpy as np
+from __future__ import absolute_import, division, print_function
```
adler-j, line 11: Unrelated, but python 2 is no longer supported by the community. We could remove support for it in ODL.

odl/test/largescale/solvers/nonsmooth/default_functionals_slow_test.py
kohr-h: That's indeed how you give it on the command line. Here it's just a list of arguments to the call, and from what I've seen it's more common to give it in "already pre-parsed" form, i.e., each argument as a separate element of the list.

odl/test/test_doc.py
kohr-h: Yeah, new issue I guess

odl/util/pytest_config.py
kohr-h: Yes definitely. I'll make an issue for that. Perhaps it's okay to say "this is the last commit that supports Python 2", rather than crafting a release for it.

PR #1556, opened 2020-04-15
Proximal to contrib and remove mrc

kohr-h:
- MRC is obsolete/incomplete, better to use the `mrcfile` package.
- ProxImaL shouldn't be in core as optional (and unmaintained, as it seems) dependency.


odl/contrib/solvers/operator/oputils.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2014-2020 The ODL contributors
```
adler-j, line 1: I would have preferred a better name for this folder. Perhaps something involving proximal?

kohr-h: I mirrored the original layout since we do that in other places, too. But sure, I'll give the module a better name.

PR #1563, opened 2020-05-23
[DO NOT MERGE] Fix for #1561 wrong operator norm if domain == range

### LINKED ISSUE #1561: Norm estimate off by 2x and doesn't converge to the right value
I found this surprising behavior in the norm estimate computation. The estimates are about a factor of two from the value computed by numpy. Running the power method for more iterations and tighter tolerances does not improve the estimate, which is also surprising because the power method should converge rather quickly. It's quite possible that I've misunderstood something fundamental. Thanks!

## Setup
```py
import numpy as np
import odl
np.random.seed(12)
X = np.random.normal(size=(100,100))
L = odl.MatrixOperator(X)
```
## Unusual behavior
```py
>>> L.norm(estimate=True)
10.00298085568247
>>> np.linalg.norm(X, ord=2)  # The expected value.
19.492578280435232
>>> odl.power_method_opnorm(L, maxiter=2)
10.81398255418922
>>> odl.power_method_opnorm(L, maxiter=1000, atol=1e-12, rtol=0.0)  # More iterations don't help
10.375771243811089
```

kohr-h:
Fixes #1561 

**Note:** I changed the default value of `maxiter` to `None`, leading to iteration till convergence by default. The old default of 100 was just arbitrary and didn't make sense for many use cases anyway.

### DIFF
--- a/odl/operator/oputils.py
+++ b/odl/operator/oputils.py
@@ -9,9 +9,11 @@
 """Convenience functions for operators."""
 
 from __future__ import absolute_import, division, print_function
+from future.utils import native
 
 import numpy as np
-from future.utils import native
+
+from odl.operator.operator import Operator
 from odl.space import ProductSpace
 from odl.space.base_tensors import TensorSpace
 from odl.util import nd_iterator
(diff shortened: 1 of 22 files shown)


kohr-h:
One failure in CI is due to an incompatibility of scikit-image with NumPy 1.13 (on the C API level). We get this error:

    RuntimeError: module compiled against API version 0xc but this version of numpy is 0xb

Not sure how to fix it. Maybe we need to consider deprecating 1.13 as well, or not use the newest scikit-image. Or just ignore the failures.

odl/operator/oputils.py
```diff
@@ -190,26 +195,15 @@ def power_method_opnorm(op, xstart=None, maxiter=100, rtol=1e-05, atol=1e-08,
         raise ValueError('`maxiter` must be positive, got {}'
                          ''.format(maxiter_in))
 
-    if op.domain == op.range:
```
adler-j, line 193: This whole PR could be changed to changing this line to `if op.adjoint is op:`, are we sure that's not a more sensible change?

kohr-h: Oh I didn't even think of that :man_facepalming: 

Yes, that makes sense. I was kind of regretting the change of behavior already...
Let me change it.

kohr-h:
> Regarding the bug I guess we need to disallow one of them and I'd probably prefer scikit-image for now.

I'm not sure how to fix it though. The last scikit-image release has been a while ago already, so I don't know why this issue is triggered now. I'll test keeping the skimage version back.

miallo:
> > Regarding the bug I guess we need to disallow one of them and I'd probably prefer scikit-image for now.
> 
> I'm not sure how to fix it though. The last scikit-image release has been a while ago already, so I don't know why this issue is triggered now. I'll test keeping the skimage version back.

#1572 will make sure that compatible versions of different modules are installed with the new dependency resolver (which is turned on by default in pip 20.3, which will be published in October 2020)

kohr-h:
Closing in favor of #1577

PR #1570, opened 2020-07-19
Fixed off-by-one error in CallbackProgressBar

miallo:
After the first iteration tqdm is called even if e.g. `step=10` and tqdm assumes that already 10 iterations have taken place (so actually it is an off-by-(`step-1`) error). This overestimates the iteration speed and the process bar is wrong as well.

### DIFF
--- a/odl/solvers/util/callback.py
+++ b/odl/solvers/util/callback.py
@@ -1014,10 +1014,9 @@ class CallbackProgressBar(Callback):
 
     def __call__(self, _):
         """Update the progressbar."""
+        self.iter += 1
         if self.iter % self.step == 0:
             self.pbar.update(self.step)
-
-        self.iter += 1
 
     def reset(self):
         """Set `iter` to 0."""


adler-j:
@kohr-h do you have any idea why the build would error?

miallo:
The difference between this test and the one from #1571 two days later is the scipy version (1.5.1 vs 1.5.2). In the release notes of 1.5.2 it says "#12487: BUG: optimize: incorrect result from approx_fprime". I suspect that rerunning the tests now would fix the problem, since as far as I can see, the callback is not even called in this test and even if it would, this callback just ignores the input and _does not mutate the data at all_. And in this test actually scipy does the optimization and not odl.

miallo:
Yes, that fixed this. And the reason, why the `NUMPY_VERSION=1.13.3` built failed is that scipy 1.5.2 is installed, but it [requires](https://github.com/scipy/scipy/blob/master/INSTALL.rst.txt#prerequisites) numpy>=1.14.5, which obviously is a conflict. To fix that, the latest compatible scipy version seems to be 1.3.3

PR #1571, opened 2020-07-21
Change DiscreteLP into DiscretizedSpace in TypeError messages

adriaangraas:
This was missed :)


miallo:
There is also a mention in an astra_cpu docstring

adriaangraas:
Thanks @miallo I did a search and found some others as well. Think these are all :smile:

kohr-h:
Thanks!

PR #1577, opened 2020-09-05
Fix wrong check for self-adjointness in power method

### LINKED ISSUE #1561: Norm estimate off by 2x and doesn't converge to the right value

kohr-h:
Fix for wrong check of self-adjointness in `power_method_opnorm`.

Closes #1561

Supersedes PR #1563

### DIFF
--- a/odl/operator/oputils.py
+++ b/odl/operator/oputils.py
@@ -190,7 +190,7 @@ def power_method_opnorm(op, xstart=None, maxiter=100, rtol=1e-05, atol=1e-08,
         raise ValueError('`maxiter` must be positive, got {}'
                          ''.format(maxiter_in))
 
-    if op.domain == op.range:
+    if op.adjoint is op:
         use_normal = False
         ncalls = maxiter
     else:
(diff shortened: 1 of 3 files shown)


kohr-h:
I accidentally put the branch in the main repo.

adler-j (approved):

PR #1581, opened 2020-09-20
Skip 3d source shift test if no ASTRA CUDA available

kohr-h:
Fix for a wrong skipping behavior in the 3D source shift test: it should have been skipped if ASTRA CUDA is not available, but was only skipped if ASTRA wasn't available at all.


kohr-h:
@pep8speaks please propose fixes

adler-j (approved):

PR #1585, opened 2020-11-04
Fix in _resize_discr for min_pt and max_pt of non-affected axes

### LINKED ISSUE #1497: Angle measurements of parallel beam CT geometry not shifted inwards at domain boundary
I am using odl 0.7.0 and would like to use an angle partition for a low dose parallel 2D geometry where measurements are taken every 5 degrees starting at 0 degrees, i.e. measurements at `[0, 5, …, 175]` degrees. To make sure that the measurements at the border (0 degrees and 175 degrees) are not shifted I guess I have to use `nodes_on_bdry=True` in the angle partition. However, creating a ray transform `ray_trafo` with this setup results in an exception when constructing the fbp with 

```
odl.tomo.analytic.filtered_back_projection.fbp_op(ray_trafo)
```

This does not happen when setting `nodes_on_bdry=False` in the angle partition. (However, then I think the measurements at the boundary are shifted inwards). I appended a minimum working example in a python notebook converted to pdf and would greatly appreciate any hints on the issue.

[Minimum working example.pdf](https://github.com/odlgroup/odl/files/3187478/Minimum.working.example.pdf)

jleuschn:
`min_pt` and `max_pt` were just copied for axes that are not affected by the resizing. This lead to inconsistent `cell_sides` if `nodes_on_bdry` did not match between the original discr and the resized discr (for the latter it is by default `False`, but can be configured via `kwargs`).

This should fix #1497.

Another MWE that works with this fix but does not without:
```
import numpy as np
from odl.discr import uniform_discr, ResizingOperator

d = uniform_discr([0.], [np.pi], (10,), nodes_on_bdry=[(True, False)])

r = ResizingOperator(d, ran_shp=(10,))
r_inv = r.inverse
```

### DIFF
--- a/odl/discr/discr_ops.py
+++ b/odl/discr/discr_ops.py
@@ -507,18 +507,16 @@ def _resize_discr(discr, newshp, offset, discr_kwargs):
     for axis, (n_orig, n_new, off, on_bdry) in enumerate(zip(
             discr.shape, newshp, offset, nodes_on_bdry)):
 
-        if not affected[axis]:
-            new_minpt.append(discr.min_pt[axis])
-            new_maxpt.append(discr.max_pt[axis])
-            continue
-
-        n_diff = n_new - n_orig
-        if off is None:
-            num_r = n_diff // 2
-            num_l = n_diff - num_r
+        if affected[axis]:
+            n_diff = n_new - n_orig
+            if off is None:
+                num_r = n_diff // 2
+                num_l = n_diff - num_r
+            else:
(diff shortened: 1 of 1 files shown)


adler-j (approved):

kohr-h:
Looks to me like the CI failures are unrelated and due to an issue with older NumPy versions and dependent libraries (pyFFTW)
(ノಠ益ಠ)ノ彡┻━┻.
Let's just merge the fix.

PR #1589, opened 2020-12-18
Maintenance of the Mayo clinic human CT dataset loader

pierocor:
This branch updates the loader and utility code for the Mayo clinic human CT dataset. In particular:
* Update dependency (dicom -> pydicom).
* Add minor fixes due to incongruent metadata and incompatibilities with ODL 1.0.0.
* Implement handling of the flying focal spot via the `src_shift_func` of the `ConeBeamGeometry`.


odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -1,29 +1,38 @@
...
 import odl
 from odl.contrib.datasets.ct import mayo
+from time import perf_counter
 
-mayo_dir = ''  # replace with your local folder
+# define data folders
+proj_folder = odl.__path__[0] + '/../../data/LDCT-and-Projection-data/' \
+             'L004/08-21-2018-10971/1.000000-Full dose projections-24362/'
+rec_folder = odl.__path__[0] + '/../../data/LDCT-and-Projection-data/' \
+             'L004/08-21-2018-84608/1.000000-Full dose images-59704/'
```
kohr-h, line 18: - I suggest to use `odl.contrib.datasets.util.get_data_dir()` to determine a storage location.
- What are the reasons to change the dataset directory? It would be good to have a small comment on top that says which data it is.
- Please use `dir` instead of `folder`. The latter is Windows lingo, and Python uses e.g. `is_dir()` to check if a `Path` object represents a directory.

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -1,29 +1,38 @@
...
 
-# Load reference reconstruction
-volume_folder = mayo_dir + '/Training Cases/L067/full_1mm_sharp'
-partition, volume = mayo.load_reconstruction(volume_folder)
+# Load projection data
+print("Loading projection data from {:s}".format(proj_folder))
+geometry, proj_data = mayo.load_projections(proj_folder,
+                                            indices=slice(16000, 19000))
```
kohr-h, line 23: Same question as above: why change the slice?

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
```diff
@@ -32,16 +41,27 @@
...
+fbp_result = fbp(td_window * flat_proj_data)
+stop = perf_counter()
+print('FBP done after {:.3f} seconds'.format(stop-start))
+
+fbp_result_HU = (fbp_result-0.0192)/0.0192*1000
+
+# Save reconstruction in Numpy format
+fbp_filename = proj_folder+'/fbp_result.npy'
+print("Saving reconstruction data in {:s}".format(fbp_filename))
+np.save(fbp_filename, fbp_result_HU)
```
kohr-h, line 54: Better not save data to disk in an example. And if so, not in the original dataset directory.
For an example, I think it's better to just display it. I like the HU-converted reconstruction.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -47,43 +51,48 @@ def _read_projections(folder, indices):
...
     file_names = file_names[indices]
 
-    data_array = None
-
     for i, file_name in enumerate(tqdm.tqdm(file_names,
                                             'Loading projection data')):
         # read the file
-        dataset = dicom.read_file(folder + '/' + file_name)
+        try:
+            dataset = pydicom.read_file(folder + os.path.sep + file_name)
```
kohr-h, line 58: The original version was wrong regarding path building. Please use the Python functionality intended for that. For now (still Py2 compatible 😒) that would be `path.join` with `from os import path`.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -47,43 +51,48 @@ def _read_projections(folder, indices):
...
+        except:
+            print("corrupted file: {}".format(file_name), file=sys.stderr)
+            print("error:\n{}".format(sys.exc_info()[1]), file=sys.stderr)
+            print("skipping to next file..", file=sys.stderr)
+            continue
```
kohr-h, line 63: I don't get this part. Why continue when we fail to read a file? We will have no clue what data will come out at the end.
I'd very much prefer this part to just fail if something cannot be read.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -47,43 +51,48 @@ def _read_projections(folder, indices):
...
-            # We need to load the first dataset before we know the shape
-            data_array = np.empty((len(file_names), cols, rows),
-                                  dtype='float32')
-
-        data_array[i] = proj_array[:, ::-1]
+        proj_array = proj_array.reshape([cols, rows])
+        data_array.append(proj_array[:, ::-1])
         datasets.append(dataset)
 
+    data_array = np.array(data_array)
```
kohr-h, line 87: Better use `np.stack`. If anything makes it impossible to build a properly-typed array, then `np.array` will give you one with `dtype=object` (which is horrible), whereas `np.stack` will fail (which is good).

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,74 +139,83 @@ def load_projections(folder, indices=None):
     # For unknown reasons, mayo does not include the tag
     # "TableFeedPerRotation", which is what we want.
     # Instead we manually compute the pitch
-    pitch = ((datasets[-1].DetectorFocalCenterAxialPosition -
-              datasets[0].DetectorFocalCenterAxialPosition) /
-             ((np.max(angles) - np.min(angles)) / (2 * np.pi)))
+    table_dist = datasets[-1].DetectorFocalCenterAxialPosition - \
+                 datasets[0].DetectorFocalCenterAxialPosition
```
kohr-h, line 143: Please use parentheses instead of backslash to break a line.
```suggestion
    table_dist = (datasets[-1].DetectorFocalCenterAxialPosition
                  - datasets[0].DetectorFocalCenterAxialPosition)
```

👍 for breaking the logic into two parts, makes it much clearer what's going on.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,74 +139,83 @@ def load_projections(folder, indices=None):
...
     # Create partition for detector
-    detector_partition = odl.uniform_partition(minp, maxp, shape)
+    detector_partition = odl.uniform_partition(det_minp, det_maxp, det_shape)
 
     # Convert offset to odl definitions
-    offset_along_axis = (mean_offset_along_axis_for_ffz +
-                         datasets[0].DetectorFocalCenterAxialPosition -
-                         angles[0] / (2 * np.pi) * pitch)
+    offset_along_axis = datasets[0].DetectorFocalCenterAxialPosition - \
+                        angles[0] / (2 * np.pi) * pitch
```
kohr-h, line 164:
```suggestion
    offset_along_axis = (datasets[0].DetectorFocalCenterAxialPosition
                         - angles[0] / (2 * np.pi) * pitch)
```

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,74 +139,83 @@ def load_projections(folder, indices=None):
...
 
     # Assemble geometry
     angle_partition = odl.nonuniform_partition(angles)
+
+    # Flying focal spot
+    src_shift_func = None
+    if use_ffs:
+        src_shift_func = lambda angle: odl.tomo.flying_focal_spot(
+                        angle, apart=angle_partition, shifts=shifts)
```
kohr-h, line 173: - I prefer the construct
  ```py
  if use_ffs:
      # something
  else:
      src_shift_func = None
  ```
  Yours looks a bit like C-style 😉 
- Minor nit (PEP8): rather than assigning a lambda, consider using `functools.partial`:
  ```py
   src_shift_func = partial(odl.tomo.flying_focal_spot, apart=angle_partition, shifts=shifts)
  ```
  I find it a bit easier on the eyes.

odl/contrib/datasets/ct/mayo.py
```diff
@@ -126,74 +139,83 @@ def load_projections(folder, indices=None):
...
+    proj_data : `numpy.ndarray`
+        Interpolated projection data on the flat rectilinear grid.
+    """
 
     # convert coordinates
-    theta, up, vp = ray_trafo.range.grid.meshgrid
-    d = src_radius + det_radius
-    u = d * np.arctan(up / d)
-    v = d / np.sqrt(d**2 + up**2) * vp
+    theta, up, vp = range_grid.meshgrid #ray_trafo.range.grid.meshgrid
```
kohr-h, line 206: Nice that you avoid the creation of a temporary operator!

odl/contrib/datasets/ct/mayo.py
```diff
@@ -239,7 +261,7 @@ def load_reconstruction(folder, slice_start=0, slice_end=-1):
 
     for file_name in tqdm.tqdm(file_names, 'loading volume data'):
         # read the file
-        dataset = dicom.read_file(folder + '/' + file_name)
+        dataset = pydicom.read_file(folder + os.path.sep + file_name)
```
kohr-h, line 264: Same comment as above about path building.

kohr-h (changes requested):
Thanks a lot for going though the trouble to keep the Mayo dataset working! Please have a look at my review comments. They should be rather straightforward to address.
Regarding functionality, I trust that the changes you made work for you, so they're kind of tested. After all, it's in `contrib`.

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
pierocor: The Mayo dataset cannot be downloaded directly from a url but it requires its own software. This prevents to use `odl.contrib.datasets.util.get_data(filename, subset, url)` for local caching. Would this work better?
```suggestion
# replace with your local directory
mayo_dir = ''
# define projection and reconstruction data directories
# e.g. for patient L004 full dose CT scan:
proj_dir = os.path.join(
    mayo_dir, 'L004/08-21-2018-10971/1.000000-Full dose projections-24362/')
rec_dir = os.path.join(
    mayo_dir, 'L004/08-21-2018-84608/1.000000-Full dose images-59704/')
```

odl/contrib/datasets/ct/examples/mayo_reconstruct.py
pierocor: It has been adjusted to the example considered (patient L004). In this case, the projection data counts ~32k images and the above slice choice reconstructs a central part of the volume.

PR #1633, opened 2024-02-01
Small compatibility fixes.

leftaroundabout:
This is mostly about the way ragged arrays are used in the scikit-based ray transform. Old numpy versions used to allow that automatically, but it is eschewed because it requires a object per row. Newer versions require explicitly requesting it.

A more idiomatic solution would be to avoid ragged arrays entirely.

### DIFF
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -565,7 +565,11 @@ class _Interpolator(object):
             Interpolated values. If ``out`` was given, the returned
             object is a reference to it.
         """
-        x = np.asarray(x)
+        if isinstance(x, tuple):
+            # Given a meshgrid, the evaluation will be on a ragged array.
+            x = np.asarray(x, dtype=object)
+        else:
+            x = np.asarray(x)
         ndim = len(self.coord_vecs)
         scalar_out = False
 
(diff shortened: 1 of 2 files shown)


odl/discr/discr_utils.py
```diff
@@ -565,7 +565,11 @@ def __call__(self, x, out=None):
             Interpolated values. If ``out`` was given, the returned
             object is a reference to it.
         """
-        x = np.asarray(x)
+        if isinstance(x, tuple):
```
JevgenijaAksjonova, line 568: Perhaps worth replacing by: if self.input_type == 'array':
as in line 576

JevgenijaAksjonova (changes requested):

odl/discr/discr_utils.py
JevgenijaAksjonova: sorry, here it should be  'meshgrid'

leftaroundabout: Honestly I can't see a real reason why there is a `self.input_type` at all and this isn't just always inferred from the input type to `__call__` instead.
On the other hand, having tuples signify meshgrid-input is definitely very obscure and prone to confusing errors, so you're right that this is not a good way to do it. Will change as suggested, and add some assertions instead.

JevgenijaAksjonova (approved):

PR #1634, opened 2024-02-01
Add bindings to ASTRA cylindrical detector geometries

wjp:
This adds bindings for the `CylindricalDetector` geometry to the astra_cuda tomography backend.

Known bug: there is a misinterpretion of the detector center if the detector partition isn't centered around 0, which is the case for quarter-pixel-shifted detectors. Fixing this needs to be coordinated with the Mayo CT data loader, which currently is assuming the presence of this bug.

### DIFF
--- a/odl/tomo/backends/astra_cuda.py
+++ b/odl/tomo/backends/astra_cuda.py
@@ -137,7 +137,10 @@ class AstraCudaImpl:
         elif proj_ndim == 3:
             # The `u` and `v` axes of the projection data are swapped,
             # see explanation in `astra_*_3d_geom_to_vec`.
-            astra_proj_shape = (proj_shape[1], proj_shape[0], proj_shape[2])
+            if self.geometry.det_curvature_radius is None:
+                astra_proj_shape = (proj_shape[1], proj_shape[0], proj_shape[2])
+            else:
+                astra_proj_shape = (proj_shape[2], proj_shape[0], proj_shape[1])
             astra_vol_shape = self.vol_space.shape
 
         self.vol_array = np.empty(astra_vol_shape, dtype='float32', order='C')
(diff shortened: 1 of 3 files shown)


odl/tomo/backends/astra_cuda.py
```diff
@@ -233,8 +236,14 @@ def _call_forward_real(self, vol_data, out=None, **kwargs):
             if self.geometry.ndim == 2:
                 out[:] = self.proj_array
             elif self.geometry.ndim == 3:
-                out[:] = np.swapaxes(self.proj_array, 0, 1).reshape(
-                    self.proj_space.shape)
+                # TODO: Find a way not to have to do rollaxis(0, 3) for
```
JevgenijaAksjonova, line 239: Should we use moveaxis instead? It is recommended in the docs:
https://numpy.org/doc/stable/reference/generated/numpy.rollaxis.html

odl/tomo/backends/astra_setup.py
```diff
@@ -339,6 +339,81 @@ def astra_conebeam_3d_geom_to_vec(geometry):
...
+       http://www.astra-toolbox.com/docs/geom3d.html#projection-geometries
+    """
+    angles = geometry.angles
+    vectors = np.zeros((angles.size, 13))
+
+    # Source position
+    vectors[:, 0:3] = geometry.src_position(angles)
+
+    # Center of detector in 3D space
+    # FIXME: This is not correct: det_point_position returns the zero-point of
```
JevgenijaAksjonova, line 385: Actually, det_point_position return the center of the detector, so this is correct. The problem is with geometry.det_axes(angles), because axes are "attached" to the zero-point and not to the center as ASTRA assumes.

JevgenijaAksjonova (changes requested):

PR #1635, opened 2024-02-10
create custom COO matrix representation for product operators

### LINKED ISSUE #1626: BroadcastOperator doesn't work with scipy > 1.8.1
Hello,

I'm reproducing the reconstruction example in https://odlgroup.github.io/odl/guide/pdhg_guide.html, and I found out that the BroadcastOperator raises an Error related to the format and sparsity of the data when computing the L operator. I'm on python3.10.1, scipy1.10 and numpy1.23.5 on an OSX.  I've tried downgrading scipy to the v1.8.1, which solves the issue with the Broadcasting operator, but other problems with the data types to construct the phantom appear instead. I've tried hacking scipy  commenting the lines 127-129 in "/scipy/sparse/_sputils.py":

        newdtype = np.dtype(dtype)
        # Commented the section below to fix problem with ODL's broadcasting operator
        #if newdtype == np.object_:
           # raise ValueError(
            #    "object dtype is not supported by sparse matrices"
            #)
And now the Broadcasting operator works.

Notice that scipy1.10.0 is installed with ODL.

paulhausner:
Since scipy version 1.9 the creation of product space operators fails due to API changes. See #1626.

A simple way to fix this is utilizing a custom structure for the coo matrices instead which has the same signature as scipy.

Tested using python version 3.11.7 and scipy 1.10.1

Fixes #1626

### DIFF
--- a/odl/util/graphics.py
+++ b/odl/util/graphics.py
@@ -436,7 +436,7 @@ def show_discrete_data(values, grid, title=None, method='',
                 plt.colorbar(mappable=csub, ticks=ticks, format=fmt)
             elif update_in_place:
                 # If it exists and we should update it
-                csub.colorbar.set_clim(minval, maxval)
+                csub.set_clim(minval, maxval)
                 csub.colorbar.set_ticks(ticks)
                 if '%' not in fmt:
                     labels = [fmt] * len(ticks)
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -323,7 +323,7 @@ def noise_array(space):
     """
     from odl.space import ProductSpace
     if isinstance(space, ProductSpace):
-        return np.array([noise_array(si) for si in space])
+        return np.array([noise_array(si) for si in space], dtype=object)
     else:
         if space.dtype == bool:
             arr = np.random.randint(0, 2, size=space.shape, dtype=bool)
--- a/odl/operator/pspace_ops.py
+++ b/odl/operator/pspace_ops.py
@@ -175,7 +175,8 @@ class ProductSpaceOperator(Operator):
             # scipy sparse matrix not supported (deprecated due to API changes)
             # keep now for backward compatibility
             print('Warning: scipy.sparse.spmatrix is deprecated.')
-            self.__ops = COOMatrix(operators.data, (operators.row, operators.col),
+            self.__ops = COOMatrix(operators.data,
+                                   (operators.row, operators.col),
                                    operators.shape)
 
         elif isinstance(operators, COOMatrix):
--- a/odl/operator/pspace_ops.py
+++ b/odl/operator/pspace_ops.py
@@ -15,6 +15,7 @@ import numpy as np
 from odl.operator.operator import Operator
 from odl.operator.default_ops import ZeroOperator
 from odl.space import ProductSpace
+from odl.util import COOMatrix
 
 
 __all__ = ('ProductSpaceOperator',
@@ -166,12 +167,20 @@ class ProductSpaceOperator(Operator):
                                 ''.format(range))
             if range.is_weighted:
                 raise NotImplementedError('weighted spaces not supported')
-
+        
         if isinstance(operators, scipy.sparse.spmatrix):
             if not all(isinstance(op, Operator) for op in operators.data):
                 raise ValueError('sparse matrix `operator` contains non-'
                                  '`Operator` entries')
+            # scipy sparse matrix not supported (deprecated due to API changes)
+            # keep now for backward compatibility
+            print('Warning: scipy.sparse.spmatrix is deprecated.')
+            self.__ops = COOMatrix(operators.data, (operators.row, operators.col),
+                                   operators.shape)
+        
+        elif isinstance(operators, COOMatrix):
             self.__ops = operators
+        
         else:
             self.__ops = self._convert_to_spmatrix(operators)
 
@@ -229,8 +238,9 @@ class ProductSpaceOperator(Operator):
     @staticmethod
     def _convert_to_spmatrix(operators):
         """Convert an array-like object of operators to a sparse matrix."""
+        
         # Lazy import to improve `import odl` time
-        import scipy.sparse
+        # import scipy.sparse
 
         # Convert ops to sparse representation. This is not trivial because
         # operators can be indexable themselves and give the wrong impression
@@ -279,8 +289,8 @@ class ProductSpaceOperator(Operator):
         # in `coo_matrix.__init__`
         data_arr = np.empty(len(data), dtype=object)
         data_arr[:] = data
-        return scipy.sparse.coo_matrix((data_arr, (irow, icol)),
-                                       shape=(nrows, ncols))
+        
+        return COOMatrix(data_arr, (irow, icol), (nrows, ncols))
 
     @property
     def ops(self):
@@ -382,7 +392,7 @@ class ProductSpaceOperator(Operator):
         data[:] = deriv_ops
         indices = [self.ops.row, self.ops.col]
         shape = self.ops.shape
-        deriv_matrix = scipy.sparse.coo_matrix((data, indices), shape)
+        deriv_matrix = COOMatrix(data, indices, shape)
         return ProductSpaceOperator(deriv_matrix, self.domain, self.range)
 
     @property
@@ -431,7 +441,7 @@ class ProductSpaceOperator(Operator):
         data[:] = adjoint_ops
         indices = [self.ops.col, self.ops.row]  # Swap col/row -> transpose
         shape = (self.ops.shape[1], self.ops.shape[0])
-        adj_matrix = scipy.sparse.coo_matrix((data, indices), shape)
+        adj_matrix = COOMatrix(data, indices, shape)
         return ProductSpaceOperator(adj_matrix, self.range, self.domain)
 
     def __getitem__(self, index):
@@ -1145,7 +1155,7 @@ class DiagonalOperator(ProductSpaceOperator):
         data = np.empty(n_ops, dtype=object)
         data[:] = operators
         shape = (n_ops, n_ops)
-        op_matrix = scipy.sparse.coo_matrix((data, (irow, icol)), shape)
+        op_matrix = COOMatrix(data, (irow, icol), shape)
         super(DiagonalOperator, self).__init__(op_matrix, **kwargs)
 
         self.__operators = tuple(operators)
--- a/odl/operator/pspace_ops.py
+++ b/odl/operator/pspace_ops.py
@@ -167,7 +167,7 @@ class ProductSpaceOperator(Operator):
                                 ''.format(range))
             if range.is_weighted:
                 raise NotImplementedError('weighted spaces not supported')
-        
+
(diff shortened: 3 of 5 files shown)


leftaroundabout:
Has somebody investigated yet whether switching to scipy's [`dok_matrix`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.dok_matrix.html) instead would also work?

leftaroundabout:
After some thinking (and talking with @JevgenijaAksjonova) I would go further and say it's probably most reasonable to not use a dedicated sparse-matrix class at all, but simply a dictionary associating index-tuples to operators. This makes `__getitem__` trivial. The other needed operations are also easy, though they will be somewhat less efficient than with something based on lists of indices and values.

Why I hesitate about making a dedicated class: this won't be a _proper_ sparse-matrix implementation unless a lot more work is put into it. One would in principle need to think about questions like how duplicate and out-of-order indices are to be handled. Getting this wrong could lead to strange and subtle bugs. With a plain `dict`, these issues don't arise.

Performance / memory concerns should not really matter here, assuming that `ProductSpace` is typically used with just a few spaces, accordingly just a few operators in the product operator, but these operators are so heavy as to amortize the overhead of the dict. That's a very different situation from `scipy.sparse`'s use cases.

Maybe I'm wrong and some users need products over lots of spaces, though that would sound more like some _tensor product space_.

ozanoktem:
One setting where one might end up with a constructing operators that acts on products over lots of spaces is in Kaczmarz method. The idea is to apply the forward operator on subsets of the data, which in tomography is typically data acquired when the source is in a specific position. Thus, one easily ends up with products over hundreds (if not thousands) of spaces. ODL has an implementation of Kaczmarz, see [here for the documentation](https://odlgroup.github.io/odl/generated/odl.solvers.iterative.iterative.kaczmarz.html) and [here for an example usage](https://github.com/odlgroup/odl/blob/master/examples/solvers/kaczmarz_tomography.py).

JevgenijaAksjonova:
ProductSpaceOperator is not used in the example with Kaczmarz method.

leftaroundabout:
@ozanoktem I get the point: the Kaczmarz method conceptually always inverts a large `BroadcastOperator`, which is under the hood a dense (!), single-column `ProductSpaceOperator`. Though, as Jevgenija said, the ODL Kaczmarz method does not currently express it this way, it makes sense to provide for that option, if only for convenience (as in the example, which uses the `BroadcastOperator` for a single forward pass).

In that case using a `dict`-based implementation does have a more significant overhead, compared to the cheap `RayTransform`s. I still tend to say that in cases where this matters something more fundamentally different should be done, and at any rate a COO sparse matrix is also not optimal for the dense `BroadcastOperator`, but it is certainly a better compromise.

There is still an alternative to be considered to having a custom COO class here: since the cases `BroadcastOperator` and `ReductionOperator` are always dense matrices, they could either just not use `ProductSpaceOperator` underneath, or `ProductSpaceOperator` could special-case `n=1` and `m=1` and use simple array storage for these.

All of that requires extra code though, so I tend to agree to going with the custom-class approach, then.

leftaroundabout:
Ok, the test discrepancies are purely down to the patch
```diff
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -323,7 +323,7 @@ def noise_array(space):
     """
     from odl.space import ProductSpace
     if isinstance(space, ProductSpace):
-        return np.array([noise_array(si) for si in space])
+        return np.array([noise_array(si) for si in space], dtype=object)
     else:
         if space.dtype == bool:
             arr = np.random.randint(0, 2, size=space.shape, dtype=bool)
```
I think it's more the test suite that's at fault here, than the code being tested.

paulhausner:
@leftaroundabout Overall, I agree with your assessment. I think the duplicate and out-of-order elements issues also arise with the scipy implementation so the suggested class-based approach should not break any existing code at least. Checks for duplicates and ordering could be implemented pretty cheaply here.

It might be misleading to call the class `COOMatrix` since as you pointed out it does not implement any proper matrix functionality (even though this might be of interest in the future) but maybe it should be renamed for now to avoid confusion.

leftaroundabout:
@paulhausner 

> It might be misleading to call the class `COOMatrix`

I think the name is ok, as long as there's some documentation as to the fact that this is merely a bare-bones container data structure with only one use case.

Then I would support merging this; the `pspace_test`s need to be fixed separately. I haven't found anything else that would get broken by the change.

ozanoktem:
Besides making sure unit tests pass, one should also ensure that the examples listed in the links below (many of these rely on constructing product space operators) actually work:
- https://github.com/odlgroup/odl/tree/master/examples/solvers
- https://odlgroup.github.io/odl/getting_started/first_steps.html
- https://github.com/odlgroup/odlworkshop/blob/master/code/part4_spectral_ct.ipynb

leftaroundabout:
I checked that the examples involving product space operators (`pdhg_denoising_tgv.py` and `pdhg_tomography_tgv.py`) run. That is, they make it through 10 iterations until crashing with an unrelated `AttributeError: 'Colorbar' object has no attribute 'draw_all'`; presumably that's https://github.com/odlgroup/odl/issues/1636.

Can't vouch for correctness of computations. For now, so much in ODL just doesn't work at all that I'm afraid we first need to squash the issues that prevent getting results at all, then when all the examples run and plotting works again we have to scrutinise the results again.

Thanks to the high-level `Operator` class I'm actually fairly optimistic that if those examples run some iterations without error then they probably are also correct.

odl/util/graphics.py
```diff
@@ -436,7 +436,7 @@ def show_discrete_data(values, grid, title=None, method='',
                 plt.colorbar(mappable=csub, ticks=ticks, format=fmt)
             elif update_in_place:
                 # If it exists and we should update it
-                csub.colorbar.set_clim(minval, maxval)
```
leftaroundabout, line 439: Should not be in this PR, this is subject of https://github.com/odlgroup/odl/pull/1638

odl/util/testutils.py
```diff
@@ -323,7 +323,7 @@ def noise_array(space):
     """
     from odl.space import ProductSpace
     if isinstance(space, ProductSpace):
-        return np.array([noise_array(si) for si in space])
+        return np.array([noise_array(si) for si in space], dtype=object)
```
leftaroundabout, line 326: Handled by https://github.com/odlgroup/odl/pull/1640

leftaroundabout (approved):
I think this should be merged. Although I'm not sure the custom class is really the best from a project-maintenance perspective, it is certainly close to what SciPy used to do, and it _just works_ for the tests and examples we currently have.

There are some things in here that don't really belong to the sparse matrix issue, those are addressed in https://github.com/odlgroup/odl/pull/1640 and https://github.com/odlgroup/odl/pull/1638. It's probably easiest to first merge those PRs and then rebase this one on top of them.

JevgenijaAksjonova:
> I think this should be merged. Although I'm not sure the custom class is really the best from a project-maintenance perspective, it is certainly close to what SciPy used to do, and it _just works_ for the tests and examples we currently have.
> 
> There are some things in here that don't really belong to the sparse matrix issue, those are addressed in #1640 and #1638. It's probably easiest to first merge those PRs and then rebase this one on top of them.

Could you please rebase this PR?

leftaroundabout:
Rebased version at https://github.com/leftaroundabout/odl/tree/feature/custom-coo-sparse

I have confirmed that this still succeeds the unit tests and relevant examples.

ozanoktem:
I can still not run the test example in [https://github.com/odlgroup/odl/blob/master/examples/solvers/pdhg_tomography_tgv.py](https://github.com/odlgroup/odl/blob/master/examples/solvers/pdhg_tomography_tgv.py), get error message `ValueError: object dtype is not supported by sparse matrices` when running  lines 74-75 in that example.

ozanoktem:
This code example for TV reconstruction using [Douglas-Rachford](https://odlgroup.github.io/odl/generated/odl.solvers.nonsmooth.douglas_rachford.douglas_rachford_pd.html?highlight=douglas%20rachford#odl.solvers.nonsmooth.douglas_rachford.douglas_rachford_pd) does not work either:

```
import numpy as np
import odl

# --- Set up the forward operator (ray transform) --- #
      
# Reconstruction space: discretized functions on the rectangle
# [-20, 20]x[-20,20] with 300 samples per dimension.
reco_space = odl.uniform_discr(
    min_pt=[-20, -20], max_pt=[20, 20], shape=[300, 300], dtype='float32')

# Create a discrete Shepp-Logan phantom (modified version)
phantom = odl.phantom.shepp_logan(reco_space, modified=True)
# Show phantom
phantom.show(title='Phantom');

# Data space: Make a fan- geometry with flat detector
# Source positions: 360 positions uniformly spaced in 0, 2pi
source_pos = odl.uniform_partition(0, 2 * np.pi, 360)
#  Detector: 500 detectorelements uniformly distributed on [-68,68]
detector_pixels = odl.uniform_partition(-68, 68, 500)
# Aquisition geometry
geometry = odl.tomo.geometry.conebeam.FanBeamGeometry(source_pos, 
                                                      detector_pixels, 
                                                      src_radius=40, 
                                                      det_radius=40)

# Forward operator: Ray transform 
fwd_op = odl.tomo.RayTransform(reco_space, geometry)

# --- Generate artificial data --- #

# Create projection data by calling the ray transform on the phantom
ideal_data = fwd_op(phantom)
data = ideal_data + 0.5 * odl.phantom.white_noise(fwd_op.range) * np.mean(ideal_data)
    
# Show data
data.show(title='Projection Data (Sinogram)');

# --- Reconstruction: Total variation (with Douglas-Rachford) --- #

# Assemble all operators into a list.
alpha = 0.5
grad = odl.Gradient(fwd_op.domain)
lin_ops = [fwd_op, grad]

# Create functionals for the l2 distance and l1 norm.
g_funcs = [odl.solvers.L2NormSquared(fwd_op.domain).translated(data),
           alpha * odl.solvers.L1Norm(grad.range)]

# Functional of the bound constraint 0 <= f <= 1
f = odl.solvers.IndicatorBox(fwd_op.domain, 0, 1)

# Find scaling constants so that the solver converges.
# See the douglas_rachford_pd documentation for more information.
opnorm_fwdOp = odl.power_method_opnorm(fwd_op, xstart=data)
opnorm_grad = odl.power_method_opnorm(grad, xstart=data)
sigma = [1 / opnorm_fwdOp**2, 1 / opnorm_grad**2]
tau = 1.0

# Solve using the Douglas-Rachford Primal-Dual method
x = fwd_op.domain.zero()
odl.solvers.douglas_rachford_pd(x, f, g_funcs, lin_ops,
                                tau=tau, sigma=sigma, niter=100)

# --- Show results --- #
x.show('TV (with Douglas-Rachford)', force_show=True);
```

leftaroundabout:
@ozanoktem 

> I can still not run the test example in https://github.com/odlgroup/odl/blob/master/examples/solvers/pdhg_tomography_tgv.py,

Were you using [the rebased version](https://github.com/leftaroundabout/odl/tree/feature/custom-coo-sparse)? The PDHG examply works fine for me on that branch.

The branch here does not include the other fixes yet.

PR #1637, opened 2024-02-13
MAINT: Skip the check for theta <= 1 in the PDHG

sbanert:
DOC: Update the PDHG docstring to latest research.

The Chambolle-Pock method converges even for $\theta > 1$, so the check that requires the contrary (and aborts with an error otherwise) should be removed. I also updated the docstring with a paper of ours that shows relaxed requirements on the step lengths.

Moreover, I removed certain statements that seem unnecesary from the docstring like the requirement that the spaces be finite-dimensional (for inifite-dimensional spaces, the convergence is weak, as is standard for these methods), and that the functionals be non-negative (this is definitely not needed).

I am not even sure if it is good to keep the check for non-negativity of $\theta$ in the code. If users try to experiment with negative $\theta$, why not? Our code should not break in this case. The check for $\theta \leq 1$ currently prevents us from using this method for numerical experiments for the paper [BUG2023].


ozanoktem:
Agree that the solver should in general not prevent a user to call it with parameter settings that does not guarantee convergence. Perhaps it is enough that the solver raises a warning when it is used in this way.

Issue #1636, opened 2024-02-12
Issue with using callback to show intermediate results in iterative solvers


ozanoktem:
Most solvers in ODL are iterative and my understanding is that setting a callback offers a way to pass instructions to such a solver that can be executed in each iterate. There now seems to be an issue with the way the callback-function in ODL interfaces with the show-function. More specifically, one can use callback to show intermediate reconstructions in each iterate by setting

`callback_settings = (odl.solvers.CallbackPrintIteration() & odl.solvers.CallbackShow())`

and then calling the ODL solver with `callback=callback_settings`. However, this yields an attribute error that allow the first iterate to be shown, but then iterates are terminated with the message 

> iter = 0
> iter = 1
> Traceback (most recent call last):
> 
>   File ~/opt/anaconda3/envs/ODL/lib/python3.9/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
>     exec(code, globals, locals)
> 
>   File ~/Library/CloudStorage/Dropbox/Teaching/KTH/SF3582 Inverse Problems/Laborations/issue_callback.py:42
>     odl.solvers.iterative.iterative.landweber(fwd_op, reco, data, niter=10, callback=callback_settings)
> 
>   File ~/odl/odl/solvers/iterative/iterative.py:120 in landweber
>     callback(x)
> 
>   File ~/odl/odl/solvers/util/callback.py:133 in __call__
>     p(result)
> 
>   File ~/odl/odl/solvers/util/callback.py:666 in __call__
>     self.fig = x.show(title, fig=self.fig,
> 
>   File ~/odl/odl/discr/discr_space.py:1515 in show
>     return show_discrete_data(values, part, title=title, method=method,
> 
>   File ~/odl/odl/util/graphics.py:439 in show_discrete_data
>     csub.colorbar.set_clim(minval, maxval)
> 
> AttributeError: 'Colorbar' object has no attribute 'set_clim'

Below is a small example demonstrating the issue. 

```
import numpy as np
import odl
      
# Reconstruction space: discretized functions on the rectangle
# [-20, 20]x[-20,20] with 300 samples per dimension.
reco_space = odl.uniform_discr(
    min_pt=[-1, -1], max_pt=[1, 1], shape=[256, 256], dtype='float32')

# Create a discrete Shepp-Logan phantom (modified version)
phantom = odl.phantom.shepp_logan(reco_space, modified=True)

# Data space: Parallel-beam geometry with flat detector
[...]

PR #1638, opened 2024-02-15
Compatibility with matplotlib-3.8.

leftaroundabout:
The update methods for colorbars have been removed in favour of more global updates.

Applying the change to the `clim`s simply on the `csub` was suggested by paulhausner in https://github.com/odlgroup/odl/pull/1635 and seems to work fine.

Actually updating the drawing of the colorbar (specifically, the ticks on it) did not work with the fix suggested in the Matplotlib documentation, but `canvas.draw_idle()` does seem to have the intended effect.

Fixes https://github.com/odlgroup/odl/issues/1636.


JevgenijaAksjonova (approved):
Looks good

Issue #1639, opened 2024-02-16
Failing unit tests for product spaces


leftaroundabout:
Running `odl/test/space/pspace_test.py` results in several failures, all of them due to this error:
```
        from odl.space import ProductSpace
        if isinstance(space, ProductSpace):
>           return np.array([noise_array(si) for si in space])
E           ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.

../../util/testutils.py:326: ValueError
```
Evidently this is an instance where the old behaviour relied on NumPy implicitly building ragged arrays (as arrays-of-arrays, i.e. dtype=object), which newer versions do not do anymore (cf. also https://github.com/odlgroup/odl/pull/1633).

leftaroundabout:
To me, the most sensible fix would seem to just not use `np.array` at all. Product spaces are inhomogenenous, and normally each constituent space is at least somewhat big. This is a natural use case for a simple _tuple_.

Alternatively, this can also be done with NumPy, explicitly asking for an array of objects by specifying `dtype=object`.

leftaroundabout:
Probably, object-arrays are actually the less painful option.
```diff
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -323,7 +331,7 @@ def noise_array(space):
     """
     from odl.space import ProductSpace
     if isinstance(space, ProductSpace):
-        return np.array([noise_array(si) for si in space])
+        return np.array([noise_array(si) for si in space], dtype=object)
     else:
         if space.dtype == bool:
             arr = np.random.randint(0, 2, size=space.shape, dtype=bool)
```
This causes many other failures, because the ubiquitous (almost-) equality check performed on the results doesn't work on `dtype=object`. This can be fixed thus:
```diff
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -146,7 +146,15 @@ def all_equal(iter1, iter2):
 
 
 def all_almost_equal_array(v1, v2, ndigits):
-    return np.allclose(v1, v2,
+    if v1.dtype is np.dtype(object) or v2.dtype is np.dtype(object):
+        if len(v1) != len(v2):
+            return False
+        components_close = True
+        for w1,w2 in zip(v1,v2):
+            components_close &= all_almost_equal(w1, w2, ndigits)
+        return components_close
+    else:
+        return np.allclose(v1, v2,
[...]

PR #1640, opened 2024-02-19
Product space unit tests compatible with modern NumPy

leftaroundabout:
As per https://github.com/odlgroup/odl/issues/1639, the unit tests for product spaces have been failing.

This is mostly due to the more restrictive way NumPy nowadays handles arrays with non-rectangular /tensor structure (basically, it _doesn't_).

The way most of these tests work is by comparing an ODL operation to its plain NumPy pendant, but specifically with product spaces this means in general ragged arrays.

This pull request introduces some special case handling, with arrays of dtype=object (in the outer layer only) as heterogeneous arrays, to get those tests working again.

### DIFF
--- a/odl/test/space/pspace_test.py
+++ b/odl/test/space/pspace_test.py
@@ -997,50 +997,62 @@ def test_real_imag_and_conj():
 
 
 def test_real_setter_product_space(space, newpart):
-    """Verify that the setter for the real part of an element works."""
+    """Verify that the setter for the real part of an element works.
+    What setting the real part means depends on the inputs; we perform a
+    recursive deconstruction to cover the possible cases."""
+
+    def verify_result(x, expected_result, recursion_limit=4):
+        if recursion_limit<=0:
+            return False
+        try:
+            # Catch scalar argument
+            iter(expected_result)
+        except TypeError:
+            return verify_result(x, expected_result * space.one(), recursion_limit-1)
+        if expected_result in space:
+            return all_equal(x.real, expected_result.real)
+        elif all_equal(x.real, expected_result):
+            return True
+        elif space.is_power_space:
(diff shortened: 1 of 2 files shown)


odl/test/space/pspace_test.py
```diff
@@ -997,50 +997,66 @@ def test_real_imag_and_conj():
...
+            return verify_result(x, expected_result * space.one(),
+                                 recursion_limit - 1)
+        if expected_result in space:
+            return all_equal(x.real, expected_result.real)
+        elif all_equal(x.real, expected_result):
+            return True
+        elif space.is_power_space:
+            return verify_result(x, [expected_result for _ in space],
+                                 recursion_limit - 1)
+
```
JevgenijaAksjonova, line 1020: Why recursion limit is 4? Should not it be 2?

leftaroundabout: Well, the `recursion_limit` is never supposed to be actually reached, it's only there to ensure that the deconstruction doesn't end up in some infinite recursion scenario (which would result in a very strange stack-overflow failure of the test suite, if not just a hung test driver).

In practice `recursion_limit=2` would also be enough here, yes, but it would be conceivable to unwrap nested product spaces, in which case that would not suffice. Maybe some larger number would make it clearer what's the intent?

JevgenijaAksjonova: Perhaps you can leave a comment saying that the limit us not supposed to be reached?

leftaroundabout: [done](https://github.com/odlgroup/odl/pull/1640/commits/67e5eee133b122c3f666e3c8bd16f85765bd8357)

JevgenijaAksjonova (approved):

Issue #1648, opened 2024-08-07
NumPy dtype=object problem in linear resampling operator


leftaroundabout:
The internally used array types, in particular `norm_distances` in [`ResamplingOperator`](https://github.com/odlgroup/odl/blob/60a854a8391d62addbbe1c617e076668ef14d523/odl/discr/discr_ops.py#L29) are currently inconsistent. These should mostly be `np.ndarray(dtype=float)` or similar, but some of them are in fact lists or NumPy arrays with `dtype=object`.

In older versions of NumPy, this kind of discrepancy merely caused silent performance degradation. In newer ones, adding to an object-array is actually an error, which is caught by the test suite:
```
___________________________________________ [doctest] odl.discr.discr_ops.Resampling.__init__ _______________________________________________________
064 
065         Apply the corresponding resampling operator to an element:
066 
067         >>> print(resampling([0, 1, 0]))
068         [ 0.,  0.,  1.,  1.,  0.,  0.]
069 
070         With linear interpolation:
071 
072         >>> resampling = odl.Resampling(coarse_discr, fine_discr, 'linear')
073         >>> print(resampling([0, 1, 0]))
UNEXPECTED EXCEPTION: numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('O') to dtype('float64') with casting rule 'same_kind'
Traceback (most recent call last):
  File "/home/justussa/miniconda3/envs/odl_olddeps/lib/python3.7/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)
  File "<doctest odl.discr.discr_ops.Resampling.__init__[8]>", line 1, in <module>
  File "/home/justussa/progwrit/python/odl/odl/operator/operator.py", line 694, in __call__
    out = self._call_out_of_place(x, **kwargs)
  File "/home/justussa/progwrit/python/odl/odl/discr/discr_ops.py", line 116, in _call
    interpolator, self.range.meshgrid, out=out_arr
  File "/home/justussa/progwrit/python/odl/odl/discr/discr_utils.py", line 172, in point_collocation
    out = func(points, **kwargs)
  File "/home/justussa/progwrit/python/odl/odl/discr/discr_utils.py", line 495, in per_axis_interp
    res = interpolator(x, out=out)
  File "/home/justussa/progwrit/python/odl/odl/discr/discr_utils.py", line 606, in __call__
    values = self._evaluate(indices, norm_distances, out)
[...]

PR #1649, opened 2024-08-08
Fixing issues related to dtype=object arrays in interpolation routines

leftaroundabout:
In old versions of NumPy, ODL relied on its capability to represent ragged arrays automatically as arrays of arrays (i.e., of objects). This was in particular used for meshgrids, which are a kind of discretization supported by the interpolation classes in `odl.discr`.

Current NumPy does not automatically convert to dtype=object anymore, and for good reasons: it is error-prone (shapes become ambiguous, whether to consider the nested array or just its outer structure) and performance / memory locality suffers. In https://github.com/odlgroup/odl/pull/1633, this was addressed by explicitly generating an object-array specifically for the meshgrid-specifying inputs, but further testing (https://github.com/odlgroup/odl/issues/1648) revealed that this was not sufficient: the `dtype=object` property would percolate into the interpolation calculations, and there cause new failures due to required implicit conversion (as well as performance degradation).

This PR goes into the details of the interpolation routines and ensures linear arrays are stored with primitive dtype. It fixes the discretization tests in NumPy-1.19, though there are still some implicit conversions that the even stricter numpy-1.26 does not accept, as well as different tests that currently fail for unrelated reasons.

### DIFF
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -1355,7 +1355,8 @@ def _make_dual_use_func(func_ip, func_oop, domain, out_dtype):
                             if shp and shp[0] == 1:
                                 res = res.reshape(res.shape[1:])
                         bcast_res.append(
-                            np.broadcast_to(res, scalar_out_shape).astype(scalar_out_dtype))
+                            np.broadcast_to(res, scalar_out_shape)
+                              .astype(scalar_out_dtype))
 
                     out_arr = np.array(bcast_res, dtype=scalar_out_dtype)
                 else:
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -933,10 +933,9 @@ def _func_out_type(func):
 
     return has_out, out_optional
 
-def _broadcast_nested_list(arr_lists, element_shape):
+def _broadcast_nested_list(arr_lists, element_shape, ndim):
     """ A generalisation of `np.broadcast_to`, applied to an arbitrarily
     deep list (or tuple) eventually containing arrays or scalars. """
-    ndim = len(element_shape)
     if isinstance(arr_lists, np.ndarray) or np.isscalar(arr_lists):
         if ndim == 1:
             # As usual, 1d is tedious to deal with. This
@@ -949,7 +948,7 @@ def _broadcast_nested_list(arr_lists, element_shape):
                 arr_lists = arr_lists.reshape(arr_lists.shape[1:])
         return np.broadcast_to(arr_lists, element_shape)
     else:
-        return [_broadcast_nested_list(row, element_shape) for row in arr_lists]
+        return [_broadcast_nested_list(row, element_shape, ndim) for row in arr_lists]
 
 
 def sampling_function(func_or_arr, domain, out_dtype=None):
@@ -1027,7 +1026,7 @@ def sampling_function(func_or_arr, domain, out_dtype=None):
             else:
                 raise TypeError('invalid input `x`')
 
-            bcast_results = _broadcast_nested_list(result, scalar_out_shape)
+            bcast_results = _broadcast_nested_list(result, scalar_out_shape, domain.ndim)
 
             # New array that is flat in the `out_shape` axes, reshape it
             # to the final `out_shape + scalar_shape`, using the same
@@ -1364,7 +1363,7 @@ def _make_dual_use_func(func_ip, func_oop, domain, out_dtype):
                 try:
                     out_arr = np.asarray(out)
                 except ValueError:
-                    out_arr = np.asarray(_broadcast_nested_list(out, scalar_out_shape))
+                    out_arr = np.asarray(_broadcast_nested_list(out, scalar_out_shape, ndim=ndim))
 
                 if out_arr.dtype != scalar_out_dtype:
                     raise ValueError(
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -20,6 +20,7 @@ import sys
 from builtins import object
 from functools import partial
 from itertools import product
+from warnings import warn
 
 import numpy as np
 
@@ -624,6 +625,8 @@ class _Interpolator(object):
             try:
                 xi = np.asarray(xi).astype(self.values.dtype, casting='safe')
             except TypeError:
+                warn("Unable to infer accurate dtype for"
+                  +" interpolation coefficients, defaulting to `float`.")
                 xi = np.asarray(xi, dtype=float)
 
             idcs = np.searchsorted(cvec, xi) - 1
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -933,6 +933,24 @@ def _func_out_type(func):
 
     return has_out, out_optional
 
+def _broadcast_nested_list(arr_lists, element_shape):
+    """ A generalisation of `np.broadcast_to`, applied to an arbitrarily
+    deep list (or tuple) eventually containing arrays or scalars. """
+    ndim = len(element_shape)
+    if isinstance(arr_lists, np.ndarray) or np.isscalar(arr_lists):
+        if ndim == 1:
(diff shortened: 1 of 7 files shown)


leftaroundabout:
Note about the linter comments: I personally disagree with many of pep8speaks' suggestions, but in particular E126 is also by others [considered over-pendantic](https://github.com/PyCQA/pycodestyle/issues/167) and does not actually follow from [the PEP8 style guide](https://peps.python.org/pep-0008/#indentation).

odl/discr/discr_utils.py
```diff
@@ -1332,12 +1339,12 @@ def dual_use_func(x, out=None, **kwargs):
             elif tensor_valued:
                 # The out object can be any array-like of objects with shapes
                 # that should all be broadcastable to scalar_out_shape.
-                results = np.array(out)
```
Emvlt, line 1335: This test relates to the function `func_tens_oop`, which checks that ODL broadcasts a list of float values to declare a numpy array to the appropriate numpy lingo. For now, what about adding a deprecation warning to this test?

odl/discr/grid.py
```diff
@@ -1111,8 +1111,8 @@ def uniform_grid_fromintv(intv_prod, shape, nodes_on_bdry=True):
 
     shape = normalized_scalar_param_list(shape, intv_prod.ndim, safe_int_conv)
 
-    if np.shape(nodes_on_bdry) == ():
-        nodes_on_bdry = ([(bool(nodes_on_bdry), bool(nodes_on_bdry))] *
```
Emvlt, line 1115: There should be a type hinting for the `nodes_on_bdry `argument

Emvlt (commented):
Excellent Work Done! ready for release :)

odl/discr/grid.py
leftaroundabout: I agree, but am unsure how best to do it &ndash; as I often am in Python, where there is no unambiguous "correct type".

A direct version would be
```
def uniform_grid(min_pt, max_pt, shape,
        nodes_on_bdry: Union[bool, List[Union[bool, Tuple[bool,bool]]]] =True):
    ...
```
That expresses fairly well the different ways the function can be called, but I reckon it looks daunting to the user. An alternative could be to define this union type separately:
```
BoundarySpecification = Union[bool, List[Union[bool, Tuple[bool,bool]]]]

def uniform_grid(min_pt, max_pt, shape, nodes_on_bdry: BoundarySpecification =True):
    ...
```
This has the advantage that the type signature remains short, though it may also be less useful. And arguably, if we make a definition at all then it would be consequent to make it a dedicated _class_, though that seems overkill here.

It is also debatable whether `List` is the right thing. The documentation says "sequence". I a way tuples would be more natural (both because they're also used for array shapes, and because the type of each element can be different and lists are more commonly understood as homogeneous containers), but I find
```
def uniform_grid(min_pt, max_pt, shape,
        nodes_on_bdry: Union[bool, Tuple[Union[bool, Tuple[bool,bool]], ...]] =True):
    ...
```
more confusing than the version with `List`. I'm also not a fan of `Iterable[...]`, that seems to suggest passing a lazily-generated sequence.

To be fair, the Sphinx documentation is actually pretty clear about what the `nodes_on_bdry` does and how it can be specified. Maybe a type hint causes more confusion than it clears up, and I should rather add some comments explaining what happens in the `if isinstance(nodes_on_bdry, bool):` and following lines?

odl/discr/discr_utils.py
```diff
@@ -621,6 +621,11 @@ def _find_indices(self, x):
 
         # iterate through dimensions
         for xi, cvec in zip(x, self.coord_vecs):
+            try:
+                xi = np.asarray(xi).astype(self.values.dtype, casting='safe')
+            except TypeError:
```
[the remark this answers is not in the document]
leftaroundabout: Done in 53c77602eeececf8b17532c45f166bb9dae0b43f.

odl/discr/grid.py
Emvlt: Here is how I understand the nodes_on_bdry argument: `nodes_on_bdry : Union[bool, Tuple[bool]]`. If a bool is provided, it sets all boundaries. If a tuple is provided it functions as follows:

1. check that the dimension of the tuple matches the dimension of the array: a`ssert len(nodes_on_bdry) == array.dim`
2. We then have an array describing:

- `nodes_on_bdry = (node_on_bdry_left,node_on_bdry_right)` in dimension 1
- `nodes_on_bdry = (node_on_bdry_left,node_on_bdry_right, node_on_bdry_top, node_on_bdry_bottom)` in dimension 2
- `nodes_on_bdry = (node_on_bdry_left,node_on_bdry_right, node_on_bdry_top, node_on_bdry_bottom, node_on_bdry_front, node_on_bdry_back)` in dimension 3

What do you think about that? :)

leftaroundabout: @Emvlt that's not how it currently works. Your examples should actually be written

- `nodes_on_bdry = ((node_on_bdry_left,node_on_bdry_right),)` in dimension 1
- `nodes_on_bdry = ((node_on_bdry_left,node_on_bdry_right), (node_on_bdry_top, node_on_bdry_bottom))` in dimension 2
- `nodes_on_bdry = ((node_on_bdry_left,node_on_bdry_right), (node_on_bdry_top, node_on_bdry_bottom), (node_on_bdry_front, node_on_bdry_back))`
```
So for example `nodes_on_bdry = ((True, False), (True, True))` for a 2D domain that includes the boundary everywhere except in the right boundary. If that were all, the type hint could be written as
```
Tuple[Tuple[Bool,Bool], ...]
```
or alternatively and IMO clearer
```
List[Tuple[Bool,Bool]]
```
However, the specification can also be shortened to `((True, False), True)`. Or it could be `(True, (False, True))` which means something different: all boundaries included except the top one. This makes it all more concise but leads to much messier types with a union nested inside a list.

I'm not sure if anybody really need this flexibility just to make their user code a few characters shorter. We could use a simplified type hint and then just remark in the docs that it can also be shortened by collapsing `(False, False)` to `False`.

Comment (leftaroundabout) [reply odl/discr/grid.py:1115]: I don't think we will come to a good decision here and then, let's leave the interface as it is for now. But I opened [an issue to discuss the use of type hints](https://github.com/odlgroup/odl/issues/1654) for the future.

File: odl/util/normalize.py [hunk 4]
```diff
@@ -278,11 +278,11 @@ def normalized_nodes_on_bdry(nodes_on_bdry, length):
...
     [(True, False), (False, False), (True, True)]
     """
-    shape = np.shape(nodes_on_bdry)
-    if shape == ():
-        out_list = [(bool(nodes_on_bdry), bool(nodes_on_bdry))] * length
-    elif length == 1 and shape == (2,):
-        out_list = [(bool(nodes_on_bdry[0]), bool(nodes_on_bdry[1]))]
+    if isinstance(nodes_on_bdry, bool):
+        return [(bool(nodes_on_bdry), bool(nodes_on_bdry))] * length
+    elif (length == 1 and len(nodes_on_bdry) == 2
```
JevgenijaAksjonova, line 283: bool(nodes_on_bdry) can be replaced by nodes_on_bdry

odl/discr/discr_utils.py
```diff
@@ -1339,12 +1339,12 @@ def dual_use_func(x, out=None, **kwargs):
...
                 # that should all be broadcastable to scalar_out_shape.
-                results = np.array(out)
-                if results.dtype == object or scalar_in:
+
+                if any(res.shape != scalar_out_shape for res in out) or scalar_in:
                     # Some results don't have correct shape, need to
                     # broadcast
                     bcast_res = []
-                    for res in results.ravel():
+                    for res in out:
```
JevgenijaAksjonova, line 1350: This loops only through the outer dimension. I think we should either remove the broadcasting all together or make it general.

leftaroundabout:
This PR has become messy through the corrections, and anyways covers multiple issues that are only coarsely related. I moved the changes to #1655 and #1656.

PR #1656, opened 2024-08-29
dtype=object array problems in function-on-grid evaluation

leftaroundabout:
_This is a cleaned-up version of parts of #1649._ See that PR for discussion.

In old versions of NumPy, ODL relied on its capability to represent ragged arrays automatically as arrays of arrays (i.e., of objects). This was in particular used for meshgrids, which are a kind of discretization supported by the interpolation classes in odl.discr.

Current NumPy does not automatically convert to `dtype=object` anymore, and for good reasons: it is error-prone (shapes become ambiguous, whether to consider the nested array or just its outer structure) and performance / memory locality suffers. In #1633, this was addressed by explicitly generating an object-array specifically for the meshgrid-specifying inputs, but further testing (#1648) revealed that this was not sufficient: the `dtype=object` property would percolate into the calculations for interpolation and grid generation, and there cause new failures due to required implicit conversion (as well as performance degradation).

This PR goes into the details of the grid generation routines and ensures linear arrays are stored with primitive dtype. It fixes the discretization tests up to NumPy-1.26.

### DIFF
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -119,7 +119,7 @@ def point_collocation(func, points, out=None, **kwargs):
     >>> ys = [3, 4]
     >>> mesh = sparse_meshgrid(xs, ys)
     >>> def vec_valued(x):
-    ...     return (x[0] - 1, 0, x[0] + x[1])  # broadcasting
+    ...     return (x[0] - 1., 0., x[0] + x[1])  # broadcasting
     >>> # For a function with several output components, we must specify the
     >>> # shape explicitly in the `out_dtype` parameter
     >>> func1 = sampling_function(
--- a/odl/discr/discr_utils.py
+++ b/odl/discr/discr_utils.py
@@ -924,6 +924,25 @@ def _func_out_type(func):
     return has_out, out_optional
 
 
+def _broadcast_nested_list(arr_lists, element_shape, ndim):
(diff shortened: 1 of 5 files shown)


odl/discr/discr_utils.py
```diff
@@ -1332,33 +1353,18 @@ def dual_use_func(x, out=None, **kwargs):
...
-                            # x[0] in a function.
-                            # Without this, broadcasting fails.
-                            shp = getattr(res, 'shape', ())
-                            if shp and shp[0] == 1:
-                                res = res.reshape(res.shape[1:])
-                        bcast_res.append(
-                            np.broadcast_to(res, scalar_out_shape))
+                try:
+                    out_arr = np.asarray(out)
+                except ValueError:
```
JevgenijaAksjonova, line 1358: Previously there was a check if scalar_in ==True. Now it is gone. I can't see why it was necessary....

JevgenijaAksjonova: Could it be because of the corner case when ndim=1?

leftaroundabout: Honestly I can't see why it was necessary in the old version either, but the new check `np.isscalar(arr_lists)` in line 930 certainly makes sure that scalars are handled correctly.

leftaroundabout:
I seem to have made a mistake while rebasing the changes, some tests are failing. I'll figure it out.

leftaroundabout:
There was just one more commit missing.

PR #1660, opened 2024-10-14
Backend/pytorch arrays

Emvlt:
New interface with ASTRA using Link interface:
1. New `astra_cuda_link` backend
2. New Link Class
3. PyTorch Autograd wrapper
4. Test functions
5. Benchmark package
6. Adapted PyTest functions
7. pytorch calls hotfixes
8. Type hinting hotfixes 

<!> Known issues:
1.  in-place does not work
10. 2D does not work


Emvlt:
We decided to not use the links interface during development due to changes in the ASTRA API and have updated the bindings, which will be released soon.

PR #1663, opened 2024-12-03
Support for immutable LinearSpace elements

leftaroundabout:
Up until now, ODL has had a strong emphasis on carrying out operations in-place. This is similar to traditional application-specific software like solvers written in Fortran, even when the user writes code that looks more functional/declarative.

While this is in many instances a good tradeoff between high-level ease of use and low-level performance, it is not suitable for all use cases.

- The pre-allocation philosophy makes ODL overly rigid. It requires that it is known beforehand what storage every element of a space will take up. This interferes with things like adaptive resolution and batching.
- Often it is not clear that it is performance-advantageous at all to modify in-place. If a result array is allocated ad-hoc just to write to it, nothing is gained. On the other hand, immutable objects can _avoid_ allocation by reusing objects in place, simply sharing references.
- While in-place updates are efficient in backends that merely wrap around arrays (like NumPy), this is not the case for more advanced backends. Of particular interest is PyTorch, which can build up a computation graph for automatic-differentiation purposes. Though PyTorch also supports in-place updates, these complicate the process and as a result make performance much worse than for out-of-place calculations.

This pull request allows spaces to specify whether their elements can be handled as mutable or immutable objects, and/or which way is more appropriate. Based on this, the general arithmetic operations will either pre-allocate memory and use it in place, or give back results in a RAII- / move-semantics fashion.

I have tested that this works by confirming that the test suite succeeds also when (contrivedly) selecting `in_place = NumOperationParadigmSupport.NOT_SUPPORTED` for `NumPyTensorSpace`. Not sure whether it would make sense to include this into the test suite.

The PR does not yet guarantee that operators and solvers follow the style indicated by `supported_num_operation_paradigms`, though to some extend they inherit the behaviour of the arithmetic primitives.

### DIFF
--- a/odl/space/npy_tensors.py
+++ b/odl/space/npy_tensors.py
@@ -889,6 +889,14 @@ class NumpyTensor(Tensor):
         """The `numpy.ndarray` representing the data of ``self``."""
         return self.__data
 
+    def _assign(self, other, avoid_deep_copy):
+        """Assign the values of ``other``, which is assumed to be in the
+        same space, to ``self``."""
+        if avoid_deep_copy:
+            self.__data = other.__data
+        else:
+            self.__data[:] = other.__data
+
     def asarray(self, out=None):
         """Extract the data of this array as a ``numpy.ndarray``.
 
--- a/odl/space/npy_tensors.py
+++ b/odl/space/npy_tensors.py
@@ -288,6 +288,8 @@ class NumpyTensorSpace(TensorSpace):
             # No weighting, i.e., weighting with constant 1.0
             self.__weighting = NumpyTensorSpaceConstWeighting(1.0, exponent)
 
+        self.__use_in_place_ops = kwargs.pop('use_in_place_ops', True)
+
         # Make sure there are no leftover kwargs
         if kwargs:
             raise TypeError('got unknown keyword arguments {}'.format(kwargs))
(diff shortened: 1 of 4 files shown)


odl/space/pspace.py
```diff
@@ -388,6 +389,50 @@ def dtype(self):
         else:
             raise AttributeError("`dtype`'s of subspaces not equal")
 
+    @property
+    def supported_num_operation_paradigms(self) -> NumOperationParadigmSupport:
```
Emvlt, line 393: If i understand correctly, the aim here is to count the spaces within the product space that prefer a type of operation over the other. Based on this count, we set the preferred type of operation of the entire product space. Could you please make that a tiny bit clearer in the comment? :)

odl/space/npy_tensors.py
```diff
@@ -564,7 +564,12 @@ def _lincomb(self, a, x1, b, x2, out):
         >>> result is out
         True
         """
-        _lincomb_impl(a, x1, b, x2, out)
```
Emvlt, line 549: This relates to issue #1678

odl/space/pspace.py
leftaroundabout: Done in df43691a

odl/space/pspace.py
```diff
@@ -618,6 +618,10 @@ def one(self):
 
     def _lincomb(self, a, x, b, y, out):
         """Linear combination ``out = a*x + b*y``."""
+        if out is None:
```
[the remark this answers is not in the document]
leftaroundabout: This is a good suggestion, however there are more classes that also do not return something from in-place `_lincomb` calls. To be completely consistent we would need to update all of them.

Since this is an internal method and all user-facing higher-level methods already work around it (expecting no return value in the in-place case) I would not prioritize this consistency for now.

Emvlt (approved):

PR #1667, opened 2024-12-06
It turns out `matplotlib-base` is not a package pip can track.

leftaroundabout:
It is only a Conda that splits up matplotlib into a base and a full part, which differ in the latter pulling in the Qt widget TK.

For pip this does not apply since is only installs Python modules anyway, and merely uses Qt if already present on the system.

Partially reverts 7443b3c036875e23cecd4a9fd3c0015441706114.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,6 @@ install_requires =
     packaging >=17.0
     numpy >=1.19, <1.27
     scipy >=1.1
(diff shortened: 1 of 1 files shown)


leftaroundabout:
This only reverts a single incorrect change I made in https://github.com/odlgroup/odl/pull/1659, so I'm merging it right away.

Issue #1670, opened 2024-12-07
PyTest-8 has changed keyword arguments


leftaroundabout:
This occurs only for skipped Astra tests, with `pytest==8.3.4` (and not with `pytest==7.4.4`):

```
________________________________________________________________________________________________________ test_anisotropic_voxels[ geometry='par2d' ] _________________________________________________________________________________________________________

geometry = Parallel2dGeometry(
    uniform_partition(0.0, 3.142, 100),
    uniform_partition(-30.0, 30.0, 100)
)

    def test_anisotropic_voxels(geometry):
        """Test projection and backprojection with anisotropic voxels."""
        ndim = geometry.ndim
        shape = [10] * (ndim - 1) + [5]
        space = odl.uniform_discr([-1] * ndim, [1] * ndim, shape=shape,
                                  dtype='float32')
    
        # If no implementation is available, skip
        if ndim == 2 and not odl.tomo.ASTRA_AVAILABLE:
>           pytest.skip(msg='ASTRA not available, skipping 2d test')
E           TypeError: skip() got an unexpected keyword argument 'msg'

pypitest.pyenv/lib/python3.12/site-packages/odl/test/tomo/operators/ray_trafo_test.py:362: TypeError
```

Issue #1671, opened 2025-01-08
Failing test on main branch


elma16:
When I run 

```
import odl
odl.test()
```

on google colab with a GPU, I get one failed test. In particular, this occurs on the operator sum test. 

```
______________________________ test_operator_sum[ dom_eq_ran=False ] _______________________________

dom_eq_ran = False

    def test_operator_sum(dom_eq_ran):
        """Check operator sum against NumPy reference."""
        if dom_eq_ran:
            mat1 = np.random.rand(3, 3)
            mat2 = np.random.rand(3, 3)
        else:
            mat1 = np.random.rand(4, 3)
            mat2 = np.random.rand(4, 3)
    
        op1 = MultiplyAndSquareOp(mat1)
        op2 = MultiplyAndSquareOp(mat2)
        xarr, x = noise_elements(op1.domain)
    
        # Explicit instantiation
        sum_op = OperatorSum(op1, op2)
        assert not sum_op.is_linear
>       check_call(sum_op, x, mult_sq_np(mat1, xarr) + mult_sq_np(mat2, xarr))

/usr/local/lib/python3.10/site-packages/odl/test/operator/operator_test.py:143: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

operator = OperatorSum(MultiplyAndSquareOp: rn(3) -> rn(4), MultiplyAndSquareOp: rn(3) -> rn(4))
x = rn(3).element([-2.2697122 , -0.73576414,  0.88103102])
expected = array([ 2.47990429,  2.6283014 ,  0.49551227,  2.94155397])

    def check_call(operator, x, expected):
        """Assert that operator(point) == expected."""
        # Out-of-place check
>       assert all_almost_equal(operator(x), expected)
E       assert False
E        +  where False = all_almost_equal(rn(4).element([ 0.26634276,  1.48797478,  0.20117861, -1.26019371]), array([ 2.47990429,  2.6283014 ,  0.49551227,  2.94155397]))
[...]

PR #1672, opened 2025-01-10
Use the preferred name of the message argument for skipped PyTests.

leftaroundabout:
Up to version 7.x, PyTest allows `skip(msg = ...)`, but from 8.0 it must be the more descriptive `skip(reason = ...)`.

Fixes https://github.com/odlgroup/odl/issues/1670, and also https://github.com/odlgroup/odl/issues/1671 (not the title problem, the follow-up).

I confirmed that the tests now work with both PyTest 7 and PyTest 8.


leftaroundabout:
I merge this now. The problem prevents running the tests with modern dependencies, the fix is simple and just follows PyTest's recommendations, it affects only the test suite, and nobody has voiced concerns.

PR #1676, opened 2025-04-24
Compatibility with NumPy-2

leftaroundabout:
NumPy version 2 makes some changes that broke ODL tests. The main ones concern copying semantics and dtype conversions.

These checks make the standard ODL test suite green when running against NumPy-2.2, and they also still succeed in NumPy-1.26.

The biggest change that needed to be made was the introduction of the `AVOID_UNNECESSARY_COPY` flag. It is a replacement for the paradigm of passing `copy=False` to the NumPy array conversion functions that has a too strict behaviour in NumPy-2.

### DIFF
--- a/examples/space/simple_rn.py
+++ b/examples/space/simple_rn.py
@@ -45,7 +45,7 @@ class SimpleRn(TensorSpace):
                                                        self.dim,))
         else:
             return self.element(np.array(
-                *args, **kwargs).astype(np.float64, copy=False))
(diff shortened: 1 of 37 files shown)


Emvlt:
Justus and Emilien reviewed the pull request. We modified some comments that J. already wrote and decided on a `try` for catching a possible drop of support for `np.core` in the future.

Emvlt (approved):

PR #1677, opened 2025-04-29
Relax tolerance for single-precision float constant-weighting test.

leftaroundabout:
The test succeeded most of the time, but since it depends on random initialisation there are some rare cases where the deviation is more than 1e-7 when using single-precision floats.
This is evidently not an ODL problem but simply a manifestation of the low precision on Float32 (the double precision test does not have the problem). So making the tolerance higher is appropriate.


Emvlt (approved):
We reviewed this change with J. and I have nothing to add.

PR #1681, opened 2025-05-21
Remove remnants of Python 2

leftaroundabout:
Python 3 has been ubiquitous for a long while now, and most of the newer developments around ODL cannot possibly be backported to Python 2. Yet, the ODL codebase still had some small special-case handling for Python 2 (which never trigger anymore) and mentioned distinctions between versions 2 and 3 in the documentation. Both provides at this point only useless clutter, so we should get rid of it.


leftaroundabout:
There are some more mentions of Python 2 yet; one I'm not sure about is this:
```
def cache_arguments(function):
    """Decorate function to cache the result with given arguments.

    This is equivalent to `functools.lru_cache`, with Python 3, and currently
    does nothing with Python 2 but this may change at some later point.

    Parameters
    ----------
    function : `callable`
        Function that should be wrapped.
    """
    try:
        from functools import lru_cache
        return lru_cache()(function)
    except ImportError:
        return function
```

In principle the whole `cache_arguments` could now be removed and replaced with `lru_cache` directly.

Emvlt (approved):
I looked through the changes and ran 'grep -rl Python2'  and found no other occurence of Python2 remants

PR #1683, opened 2025-05-26
Python array API support

Emvlt:


leftaroundabout:
Would it be a lot of work to put the weighting hierarchy in its own PR? I think there are a bunch of questions to be discussed there, and it should be checked thoroughly whether this really works, independent of the Python Array API generalizations.

What certainly needs clarification (if not changes) is that about the methods being "mutually exclusive". It makes some sense from the definition side (we don't want conflicting semantics between e.g. distance and norm). But surely, from the _use side_ we should then have all of them available, as far as mathematically possible? In the most typical use case (say, $L^2$) they can all be derived from the inner product, so it should be sufficient to provide only that, or alternatively the weights (which can then be used to implement first the inner product via array-API and then everything else from that).
But then there are also cases (Banach spaces) where there really is no inner product, and we'd need either weights and and exponent, or a custom norm, which also gives rise to a distance. In principle we could have _only_ a distance and nothing else &ndash; a metric space &ndash; but I'm not sure if that would ever crop up, given that the `TensorSpace` class already requires a vector-space structure.

Emvlt:
About the fact that the weighting refactoring should be in its own PR. I chose to add it here as for me, the python array API allows decoupling the backend classes (which will only add a few attributes to the Weighting) and the abstract class. As i am doing it for the TensorSpace and TensorSpaceElement classes i thought it would fit :) 
As for the behaviour, I am not adding anything: the classes are refactored but the functionality remain unchanged. I'm happy to open the discussion on the Banach spaces and agree to create a separate PR for that :)

odl/space/weighting.py
```diff
@@ -498,7 +498,7 @@ def _inner_default(x1, x2):
         # This could also be done with `np.vdot`, which has complex conjugation
```
leftaroundabout, line 498: This comment is now out of date. Fixed in d9575e856d7fcfaea09782cbba041952b3bf581c

odl/space/entry_points.py
```diff
@@ -28,22 +28,34 @@
 __all__ = ()
 
 IS_INITIALIZED = False
-TENSOR_SPACE_IMPLS = {'numpy': NumpyTensorSpace}
+TENSOR_SPACE_IMPLS = {
+    'numpy': NumpyTensorSpace
+    }
+AVAILABLE_DEVICES = {
```
leftaroundabout, line 34: This shouldn't really be needed now that `available_devices` is in `ArrayBackend`, right?

I think it would be better if everything that needs to know what devices are available for an `impl` goes through the corresponding `ArrayBackend` object instead of fiddling with another global dictionary (which would have to be kept in sync when new backends are added).

Emvlt: I agree that the devices must be only accessed through the backend. I used this variable to build the IMPL_DEVICE_PAIRS fixture as we loaded the backends but that's not clear, actually. I will move that in the pytest_config instead :)

odl/trafos/util/ft_utils.py
```diff
@@ -20,8 +20,8 @@
 from odl.set import RealNumbers
 from odl.util import (
     complex_dtype, conj_exponent, dtype_repr, fast_1d_tensor_mult,
-    is_complex_floating_dtype, is_numeric_dtype, is_real_dtype,
-    is_real_floating_dtype, is_string, normalized_axes_tuple,
+    is_complex_dtype, is_numeric_dtype, is_real_dtype,
+    is_floating_dtype, is_string, normalized_axes_tuple,
```
leftaroundabout, line 24: I'm not sure I like this change of naming. After all, complex numbers (at least the standard `complex64` and `complex128` are also based on floating point, so it's somewhat misleading to exclude them in something called `is_floating_dtype`.

Why not just leave those names as they are?

leftaroundabout: > as per NumPy's documentation, "There are 5 basic numerical types..."

but those are concrete fixed types. The `is_XYZ_dtype` don't refer to one particular type, but each discerns a whole bunch of types. And in the same way both `int` and `float` are numeric types (meaning for practical purposes that they support `+` and `*` and `-`), both `float` and `complex` are floating types, in the sense that they support `/` and `sqrt` etc..

The reason that excluding complex from `is_floating_dtype` doesn't affect the tests is probably that it then just omits some of the tests, but I don't think that's what we want. There is lots of stuff that should work both for `float` and for `complex`.

leftaroundabout:
> 2) Deprecating the test that relied on Numpy mechanisms to perform arithmetic operations between product spaces and lists. In my opinion, this should not be supported. It seems that nowhere in the codebase this behaviour is relied on. What does @leftaroundabout thinks about it?

I suppose you mean `test_unary_ops` and `test_operators`? (BTW commenting out a test case isn't really "deprecating" it)

Well, we certainly want `-x` and `x + y` and so on to work when `x` and `y` are elements of a `pspace`. This has nothing to do with NumPy, and it should be tested for both NumPy and PyTorch.

Emvlt:
Sure, deprecating is not the right word, you know what i mean though.

> **perform arithmetic operations between product spaces and lists** so not when `x` and `y` are elements of a pspace.

odl/trafos/util/ft_utils.py
Emvlt: Okay with that!

odl/array_API_support/array_creation.py
```diff
@@ -0,0 +1,126 @@
+from .utils import get_array_and_backend
+from numbers import Number
+import numpy as np
+
```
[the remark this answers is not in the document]
Emvlt: Let's consider `x`. We don't know if it's an `np.ndarray`, a `torch.Tensor `or a `LinearSpaceElement`.
We need to have the following machinery:
```
import odl

x_arr, backend = get_array_and_backend(x)
zeros_like = backend.array_namespace.zeros_like(x_arr)
```
With the array_creation module, we can have
```
import odl

zeros_like = odl.zeros_like(like_x_arr)
```
I think that the `get_array_and_backend `mechanism is a bit verbose for the user (but perfectly fine for us) and, most importantly, does not provide the documentation of the function in the IDE. I know you are a not a user of such features but i personnaly find it more user friendly.

In my opinion, our users overwhelmingly think in terms of arrays (just by personal experience+seeing the commits of non-core developpers) and will expect ODL to be lenient when it comes to the functions they are used to.

PR #1684, opened 2025-06-05
Avoid sporadic test failures due to nondeterminism

leftaroundabout:
@Emvlt and me have been running tests many times lately, to ensure the changes towards Python Array API are safe.

Besides the ones related to actual bugs, we also encountered some sporadic failures that happen only occasionally. These are due to randomness in either the test suite itself, or the methods being tested. 93e6d61 and 249a573 already addressed this partially, but not completely.

I have fixed all of these occurences that I encountered when running the entire test suite several hundred times. The tests in question I have run even more often.

There is no guarantee that there are not more, even rarer, nondeterminism problems, but at least they are now very unlikely to turn up in a debugging session or when a user attempts to get ODL running.

### DIFF
--- a/odl/test/discr/discr_ops_test.py
+++ b/odl/test/discr/discr_ops_test.py
@@ -259,8 +259,8 @@ def test_resizing_op_adjoint(padding, odl_tspace_impl):
         inner1 = res_op(elem).inner(res_elem)
         inner2 = elem.inner(res_op.adjoint(res_elem))
         assert inner1 == pytest.approx(
-            inner2, rel=space.size * dtype_tol(dtype)
-        )
+            inner2,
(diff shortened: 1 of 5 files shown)


Emvlt (approved):
This pull request addresses the problem of non-deterministic test in the test suite of ODL and raises a question about the `power_method_opnorm`'s initialisation (cf #1685). 
@leftaroundabout addressed the former by relaxing the tests to
- allow for an "unfortunate" initialisation of the random noise used, 
- accomodate for different floating-point and double-point precisions,
- account for boundaries of random integer generator
- run some solvers for longer to get the expected result
We refer the discussion about the latter in #1685

PR #1689, opened 2025-06-13
Tighten NumPy dependency to version 2.1

leftaroundabout:
This is necessary to take full leverage of the Python Array API, which is needed for ODL 1.0.

Users who want to build against NumPy-1.26 can stick to the 0.8.x versions, which will however not get the advantage of seemless multi-backend support.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - python >=3.7
     - setuptools <76.0
     - future >=0.16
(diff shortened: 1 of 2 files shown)


Emvlt (approved):
Moving to a NumPy version superior to 2.1 is necessary as we will rely heavily on the python array API for multi-backend compatibility.

Please see https://numpy.org/doc/stable/release/2.1.0-notes.html for the release notes about python array API compatibility.

PR #1692, opened 2025-07-12
fix(filter): preserve callable type in filter design

Kaiyokun:
Previously, `filter_type` in `_fbp_filter` was converted to a string before the callable check, causing the condition to always evaluate as false for callable inputs. This change preserves the original object type to ensure correct handling.

Key changes:
1. Removed the assignment `filter_type = str(filter_type).lower()` to preserve the original object type
2. Introduced `filter_type_in` to store the string representation exclusively for predefined filter comparisons
3. Maintained the existing filter type handling logic for strings ('ram-lak', 'shepp-logan', etc.)

The fix ensures:
- Callable filters (e.g., custom functions) are correctly detected by `callable()` and executed
- Predefined filter types continue to function via string matching
- Backward compatibility with existing filter type names

### DIFF
--- a/odl/tomo/analytic/filtered_back_projection.py
+++ b/odl/tomo/analytic/filtered_back_projection.py
@@ -77,19 +77,19 @@ def _fbp_filter(norm_freq, filter_type, frequency_scaling):
     ...                    filter_type='Hann',
     ...                    frequency_scaling=0.8)
     """
-    filter_type, filter_type_in = str(filter_type).lower(), filter_type
+    filter_type_in = str(filter_type).lower()
     if callable(filter_type):
         filt = filter_type(norm_freq)
-    elif filter_type == 'ram-lak':
+    elif filter_type_in == 'ram-lak':
         filt = np.copy(norm_freq)
-    elif filter_type == 'shepp-logan':
+    elif filter_type_in == 'shepp-logan':
         filt = norm_freq * np.sinc(norm_freq / (2 * frequency_scaling))
-    elif filter_type == 'cosine':
+    elif filter_type_in == 'cosine':
         filt = norm_freq * np.cos(norm_freq * np.pi / (2 * frequency_scaling))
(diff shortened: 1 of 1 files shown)


Emvlt:
This seems like a legit bug that passed through the tests.
@leftaroundabout what do you think about it? 
We also should put "writing a test suite for the analytic package of tomo" on our todo list.

Kaiyokun:
@leftaroundabout I absolutely agree that _​the `.lower()` workaround should never be needed_. I encountered this specific bug while reviewing the documentation for [`fbp_op`](https://odlgroup.github.io/odl/generated/odl.tomo.analytic.filtered_back_projection.fbp_op.html#odl.tomo.analytic.filtered_back_projection.fbp_op), particularly regarding the `filter_type` parameter:
> ​**filter_type**: optional  
> The type of filter to be used. The predefined options are, in approximate order from most noise sensitive to least noise sensitive: `'Ram-Lak'`, `'Shepp-Logan'`, `'Cosine'`, `'Hamming'`, and `'Hann'`. A callable can also be provided. It must take an array of values in [0, 1] and return the filter for these frequencies.

This led to confusion about the expected prototype for the custom callable. To investigate, I initially passed `print` just to log the arguments being supplied.

In my opinion, `filter_type` could be significantly improved. It should ideally be either:
-  A ​**standerd `enum.Enum` type**​ listing the predefined options, or
-  Even better, ​**predefined functions**​ representing each filter (e.g., `odl.tomo.analytic.filtered_back_projection.Ram_Lak`).

This approach would allow users to pass the function itself directly (e.g., `filter_type=Ram_Lak`). The added benefit is the ​**complete elimination**​ of the mixin-callable-string-matching `if-elif-else` logic within the implementation.

Of course, supporting a custom callable remains valuable. To make this clear and usable, the API documentation needs ​**more detailed explanation and an illustrative example**​ demonstrating exactly how to define and use such a function.

Issue #1693, opened 2025-08-12
ODL package executes user scripts during initialization when using common filenames


WarwickSong:
#### **Bug**
When importing the `odl` package, it automatically scans the current working directory and executes any script files with specific common names (e.g., `test.py`). This behavior leads to:
1. Unexpected execution of user scripts (not part of the ODL package).
2. Parameter parsing errors in Jupyter Notebook when scripts are run implicitly.
3. Inconsistent behavior between script execution (`python file.py`) and Jupyter Notebook.

**Technical Root Cause**  
The issue appears to originate from `odl/util/testutils.py:23`, where the line:  
```python
from future.moves.itertools import zip_longest
```  
triggers the `future` package's `__init__.py:8` to execute:  
```python
import_top_level_modules()
```  
This function attempts to import modules listed in `TOP_LEVEL_MODULES = ['builtins', 'copyreg', ..., 'test', ...]`. However, it **does not distinguish between standard library modules and user files** in the current working directory. For example:
- If a user has a file named `test.py` in their project, it will be imported as `test` due to the module resolution logic in `future/moves/__init__.py`.
- This leads to unintended execution of user code during ODL initialization.

#### **To Reproduce**
1. Create a script named `test.py` in your working directory (e.g., with a simple `print("Hello")`).
2. Run the following in Jupyter Notebook:
   ```python
   import odl
   ```
3. Observe that `test.py` is executed automatically, even though it's unrelated to ODL.

#### **Expected behavior**
The `odl` package should not execute user scripts during initialization unless explicitly configured to do so. The current behavior creates unexpected side effects and conflicts with user workflows.

#### **Actual behavior**
- User scripts with common names (e.g., `test.py`) are executed automatically.
- This causes parameter parsing errors in Jupyter Notebook due to unintended script execution.

#### **Environment**
- ODL version: 0.8.2
- Python version: 3.10.9
- Reproducible in Jupyter Notebook and IPython environments.

leftaroundabout:
After that change, your scenario does not result in code from the local script being executed anymore.

Thanks again for already narrowing down the problem, this made the fix very easy!

WarwickSong:
Thanks for the quick fix and clear explanation!
I see the change is in PR #1694 and hasn’t yet been merged into master.
Looking forward to seeing it included in the main branch, and appreciate your prompt response and clean solution.

PR #1694, opened 2025-08-12
Replace obsolete `future.moves` imports with the standard library version.

leftaroundabout:
These imports were once upon a time needed for compatibility with old Python versions, but this is not necessary anymore and anyways those old versions are also not compatible with other dependencies ODL now carries.

Besides general awkwardness, the `future.moves` modules also have strange side-effects, as reported in https://github.com/odlgroup/odl/issues/1693.

In Python 3, the simpler `import itertools` seems to work exactly the same. Test suite passes.

### DIFF
--- a/odl/util/testutils.py
+++ b/odl/util/testutils.py
@@ -20,7 +20,7 @@ from time import time
 import numpy as np
 from odl.util.npy_compat import AVOID_UNNECESSARY_COPY
 
-from future.moves.itertools import zip_longest
+from itertools import zip_longest
 
 from odl.util.utility import is_string, run_from_ipython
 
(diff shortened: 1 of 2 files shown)


leftaroundabout:
Ah, great - I hadn't noticed you already fixed this.

No reason not to merge it into `master` right now regardless, though? This is after all very straightforward and unrelated to the big changes coming up in 1.0.

leftaroundabout:
I wouldn't address all the `__future__` imports, we can do that later. But `future.moves` is actually a problem, so I'd fast-lane it.

Emvlt (approved):
This can be merged now and removing the "future" imports should be a priority 🫡

PR #1700, opened 2025-11-11
Fixing init and docs

Emvlt:
Three files are modified:

1. The ODL `__init__` which had its `__all__` synchronised with the new organisation of ODL folder and another way, more succint, to access the `tomo` functions
2. Added an `__init__` to the applications module
3. Removed a deprecated `pkg_resources` import

The documentation is also now showing the right modules


leftaroundabout:
This is addressing two subtle inconsistencies between the `__init__` files, the actual module structure, and the documented API. Both are fixed by the PR.

PR #1704, opened 2025-11-17
Dependency edit/pytest

Emvlt:
Making pytest a hard dependency of odl.

We decided to add pytest as a hard dependency for three reasons.
1) It is more robust than using if conditions inside the modules
2) It is lightweight so it should not make odl's import slower
3) It simplifies the new github test workflows


leftaroundabout:
After deciding against making the hard dependency, this is outdated.

PR #1707, opened 2025-11-19
Removing the __init__ from four ghost folders that were added in 6968…

Emvlt:
…0279b7762acaa4b591d1fa489a4df3b638db and b13196f662d2fa6a348eee54675b793d26593500


Emvlt:
I think there is also odl/solvers/functional/__init__.py which was added in the same commit but should not be there

PR #1708, opened 2025-11-24
Addition of a CLI option to only run np test for fast prototyping.

Emvlt:
It is quite convenient to be able to run np only tests on from within a venv that also has pytorch. 

The new functionnality can be invoked using 

`pytest odl/test --np`

Also edited some tests that were using a variable instead of a fixture

### DIFF
--- a/odl/core/util/pytest_config.py
+++ b/odl/core/util/pytest_config.py
@@ -17,7 +17,7 @@ from os import path
 import numpy as np
 
 import odl
-from odl.core.array_API_support import lookup_array_backend
+from odl.core.array_API_support import lookup_array_backend, check_device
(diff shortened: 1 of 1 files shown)


leftaroundabout (changes requested):
This is a really useful feature for testing most of ODL without having to wait a long time for lots of different devices etc..

Although the idea could be generalised (only one specific backend of choice / exclude one specific backend / only one specific device...) I think NumPy-only already offers most of the gains while being much simpler, so let's go forward with this. I would only give the flag a more descriptive name.

leftaroundabout (approved):
Useful feature, implemented in full generality and with good naming.

PR #1710, opened 2025-12-04
Documentation linting

Emvlt:
Opening a pull request for improved documentation and linting. 

This is quite tedious, as a lot of errors come from line-too-long, trailing-whitespaces (thanks Black!), f-strings encouraged used and unnecessary if/else conditions, but:

- It really helps for cleaning-up: i found that we had forgotten to remove the `__array_priority__` attribute of the LinearSpace and Operator and that several operations still relied on it behind our back.
- It makes it clear that function signature between Parent and Children class is not consistent.
- highlights the discrepancy between the new sparse module and the old COO-only sparse class (mostly used in pspace_ops) 
- shows that some functions could be broken down into smaller

### DIFF
--- a/odl/applications/tomo/__init__.py
+++ b/odl/applications/tomo/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2020 The ODL contributors
+# Copyright 2014-2025 The ODL contributors
 #
 # This file is part of ODL.
 #
@@ -7,8 +7,6 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 """Tomography-related operators and geometries."""
-
-from __future__ import absolute_import
 
 from .analytic import *
 from .backends import (


leftaroundabout:
I'm generally not a fan of linting, and particularly not of Black which in many cases makes the code less readable.

As I see it, there are two kinds of changes here that are actually worthwhile:

- Changing `.format` to f-strings
- Removing old-style `super(ClassName, self)` arguments

The rest of the changes are about 60% unnecessary shuffling around code that doesn't really make a difference, 20% mildly annoying tearing of layout, 5% really bad messing up the layout, and 15% mild improvement of readability.

Just my opinion.

leftaroundabout:
Would it be possible to run Black to _only_ apply the `.format` and `super` changes?

leftaroundabout:
This PR is obsolete now, the changes having (partially) been implemented in https://github.com/odlgroup/odl/pull/1715.

PR #1712, opened 2025-12-08
FBP testing function with different variations of parameters.

asalline:
This script runs a tests for Filtered Back-Projection with respect to different parameters, such as geometries, backends, and filters.


Emvlt (approved):
I have reviewed the changes and have succesfully ran the test suite on my machine, it's a green light from me! 
Thanks @asalline for these improvements :-)

PR #1716, opened 2026-01-05
Fixed a indent typo on the __repr__ of the Tensor object that created  a bug when calling the python built-in repr (None returned)

Emvlt:
This is minor


Emvlt:
Believe it or not, Numpy changed its API in 2.4 and the tests pull the 2.4 (We set >=2.3 in the requirements, which means that some user code might break).
I am pausing this PR and will propose modifications.

leftaroundabout:
I create two separate issues for those concerns.

